PR #120, opened 2011-06-09
Microformat tagging

petterreinholdtsen:
The microformat specification make it easier for search engines and automatic web page parsers to extract useful information from a web page.  As all FixMyStreet reports are geo located, it seem like a good idea
to make the coordinate available for search engines.  This should allow FixMyStreet reports to show up on Google Maps and Google Earth without any more work done from FixMyStreet.

This branch is a draft testing this approach.  Since a few minutes ago, it is used on www.fiksgatami.no,
to see if the reports show up on Google Maps as soon as Google starts the indexing of the site. :)

### DIFF
--- a/perllib/FixMyStreet/Map.pm
+++ b/perllib/FixMyStreet/Map.pm
@@ -64,7 +64,7 @@ sub display_map {
 # http://microformats.org/wiki/geo for the specification
     my ($q, %args) = @_;
     my $mapstr = $map_class->display_map(@_);
-    my $latitude = $args{latittude};
+    my $latitude = $args{latitude};
     my $longitude = $args{longitude};
     $mapstr .= "<div class=\"geo\"><abbr class=\"latitude\" title=\"$latitude\"/><abbr class=\"longitude\" title=\"$longitude\"/></div>\n";
     return $mapstr;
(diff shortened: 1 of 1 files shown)


dracos:
Since this was opened, microformats now have version 2 (with no examples of use in the wild on their wiki, though); to cover that, falling back to version 1, plus schema.org, and given we don't publicly show the co-ordinates (though the pin itself is performing a similar service), it looks like you'd have to have something like this:

```
<div itemscope itemtype="http://schema.org/GeoCoordinates" class="geo h-geo">
    <meta itemprop="latitude" content="[% pin.latitude %]" />
    <meta itemprop="longitude" content="[% pin.longitude %]" />
    <data class="p-latitude latitude" value="[% pin.latitude %]" />
    <data class="p-longitude longitude" value="[% pin.longitude %]" />
</div>
```

(I don't know what a data element is, it's not HTML5, but unless I've not read the docs right, they can't use the same hidden elements which is a shame.)

If we showed the co-ordinates somewhere on the page, this could become the slightly nicer:

```
<div itemscope itemtype="http://schema.org/GeoCoordinates" class="geo h-geo">
    <span class="p-latitude latitude" itemprop="latitude">[% pin.latitude %]</span>,
    <span class="p-longitude longitude" itemprop="longitude">[% pin.longitude %]</span>
</div>
```

Looks like Flickr, Google Maps, etc have all dropped support for displaying them; I think this is something we can leave and come back if anyone wants it in future.

PR #757, opened 2014-04-09
[Bromley] disambiguation for Beckenham

osfameron:
Also disambiguate "The Knoll" as we know that this will otherwise
resolve to "The Knole, Sevenoaks".

Fixes https://github.com/mysociety/FixMyStreet-Commercial/issues/495


dracos:
Not needed/ out of date.

PR #792, opened 2014-05-23
Refactor SendReport::Open311 to cobrand hooks

osfameron:
Avoid doing cobrand-specific logic within SendReport::Open311.
Instead:

```
- inflate a cobrand from the row's cobrand
  (complex cases such as UK may then defer to other cobrands
  depending on bodies_str)
- add a 'call_cobrand_hook' method to Cobrand::Base which
  calls a cobrand method *if* it is defined.
- the hooks defined are:
    open311_config:   set extra fields and $revert
    open311_pre_send: set endpoints and munge data
    open311_zero_result_on_fail: temporary behaviour required for
        West Berks
- create two dummy cobrands (West Berks and Cambridgeshire)
  to be able to dispatch these hooks to consistently.
```

These are tested in the new t/app/sendreport/open311.t, which does
not send any actual reports, but just checks that all the data is
munged into the expected format.  (tests under t/open311\* should
already test the actual sending)

**NB**: this is a fairly big change to the way we're doing things, and as there were no tests previously for this functionality, it would be really good to get a few pairs of eyes on this (because of having to restructure things to make them testable I can't simply run the new tests on the old code!)

### DIFF
--- a/bin/open311-populate-service-list
+++ b/bin/open311-populate-service-list
@@ -23,7 +23,8 @@ my ($opt, $usage) = describe_options(
 print($usage->text), exit if $opt->help;
 
 my $bodies = FixMyStreet::DB->resultset('Body')->search( {
-    id => { '!=', 2237 }, # XXX Until Oxfordshire does do so
+    # Until Oxfordshire does, and Bristol stops erroring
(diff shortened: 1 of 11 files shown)


pezholio:
From my limited understanding of Perl, this seems pretty sensible 👍

PR #849, opened 2014-07-27
Open311 endpoint candidate

osfameron:
NB: I expect travis to fail, as I'm not able to update the cpanfile.snapshot: Carton can't find LWP::Protocol::PSGI, though cpanm will happily do so.  Hopefully something silly I messed up.
Issuing PR anyway for code review, and to ask for help on ^^ :D


osfameron:
This Open311::Endpoint code was in fact merged in along with Warwickshire.

However, Matthew also pointed out that some of the dependencies make it awkward to provide support for some versions of Debian.

It may make sense to spin this out to its own distro and remove the code from FMS repo.

PR #873, opened 2014-08-13
DBIC Migrations

### LINKED ISSUE #856: Dev feature: DB Migrations and Fixture loading
After discussing with @davea, it would be good to:
- [ ] use a migration toolkit such as https://metacpan.org/pod/DBIx::Class::Migration to make it easier to make DB changes
- [ ] use a Fixtures tool (cp "factory boy/girl") like https://metacpan.org/pod/DBIx::Class::Fixtures to be able to dump and reload fixtures for testing/dev/demos

osfameron:
Please review!  If this, or something like this, gets merged in would also require

 - [ ] documentation in fxmystreet.org etc.
 - [ ] vagrant install script to use migrations
 - [ ] travis build script to use migrations

Fixes #856

Documented in bin/migrate

A sample second migration is available in the migration-test-moderation
branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ phonegap/www/js/config.js
 phonegap/Android/bin/
 phonegap/iPhone/FixMyStreet.xcodeproj/project.xcworkspace/xcuserdata
 phonegap/Android/gen/
+
+001-auto.sql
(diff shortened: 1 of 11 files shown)


dracos:
Looks like it did do for now.

PR #939, opened 2014-11-17
Issues/863 makemyisland changes

davewhiteland:
MakeMyIsland is effectively in production, so it would be nice to no longer be running this on its own branch


dracos:
No longer needed; moved to separate repository.

Issue #785, opened 2014-05-19
How can we change the "create new report" process to improve over-all report completion levels?


TomSteinberg:
Eg: User testing shows a couple of people who get lost in looking at old reports instead of succeeding in making a new one. An A/B test on hiding/showing the existing reports list would show whether this was a common problem, or just an accident of user tests.

Are there other things we can improve at the top of the reporting funnel? (ie: load homepage – find location – place marker on map)

dracos:
Closing as undecided. Will see if #1350 (and follow-on #1351) do anything, and perhaps look at #525/#724.

PR #947, opened 2014-11-20
AB test clicking a button on FMS.com homepage

zarino:
The latest suggestion from #785.

**Hypothesis:** Conversions can be increased by giving the reporting process a more clearly defined call to action.

**Method:** For 50% of users, replace the homepage location search box with a large button to “Report a problem”, and a smaller link to “view problems”. Clicking either of these elements does the same thing: fade the elements out, and the existing form in. Once it’s faded in, we auto-focus the form input to save a click, meaning the total number of clicks for the interaction is the same as the existing behaviour.

Preview animation

All set up in Google Analytics, and the experiment is already running. So just needs a code review (and no doubt commit squash) from @dracos, and then we can deploy!

/cc @TomSteinberg @wrightmartin

### DIFF
--- a/templates/web/fixmystreet.com/around/postcode_form.html
+++ b/templates/web/fixmystreet.com/around/postcode_form.html
@@ -1,0 +1,48 @@
+[%# Identical to parent, but with an extra div and javascript for our homepage A/B test %]
+
+<div id="front-main">
+    <div id="front-main-container">
+        [% INCLUDE 'around/intro.html' %]
+
+        <div id="homepage-ab-test">
+            <p><a id="cta-report">Report a problem</a></p>
+            <p><a id="cta-view">Or view problems in an area you know</a></p>
+        </div>
+
+        [%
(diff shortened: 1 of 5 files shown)


dracos:
"Once it’s faded in, we auto-focus the form input to save a click, meaning the total number of clicks for the interaction is the same as the existing behaviour." – the existing behaviour auto-focuses the form?

zarino:
> the existing behaviour auto-focuses the form

Oh! So it does.

osfameron:
Hah, I'd never noticed that either.  I wonder how many people do?  H

Hakim
-- 
*Developer*, mySociety Services
http://www.mysociety.co.uk

dracos:
Doesn't hurt if they don't, but I certainly use that knowledge whenever I use the site :) Anyway, merging & deploying this now.

zarino:
Thanks @dracos :+1:

Cc'ing @jacksonj04 – we might want to have a quick conversation about this once the new behaviour is deployed.

PR #970, opened 2014-12-17
Minor FixaMinGata fixes + s/Open Street Map/OpenStreetMap

jonkri:
:)


dracos:
Hi :) I have made some changes to this PR and placed them on the `jonkri-master` branch. Could you possibly have a look and check it all still works okay for you? The changes I've made are as follows:
* `councils_text_none.html` – I've made this a separate `fixamingata` template as I think that's clearer;
* Footer - I've updated it to use the `footer-marketing` template the parent does (and the text to fixmystreet.org not GitHub);
* `fill_in_details_form.html` – this can now be removed and smaller template overrides used for the logged out and tips (as with the parent templates), which means it gets a few other updates it was lagging behind, hope that makes sense. The only other thing you might want here is that when we added the "tips" to fixmystreet.com we decided we didn't need the sidebar that appears any more, do you want it removed also?
* CSS – I've kept only the one line that you changed, the other lines (and images) aren't needed to be copied;
* Minor commit renames

Let me know if you have any questions – if it's all okay, I'll then merge it in.

dracos:
This has been merged as #1021.

PR #975, opened 2014-12-19
Harrogate candidate

osfameron:
Currently deployed to http://harrogate.commercial.fixmystreet.staging.mysociety.org/

Please review if you get a chance @stevenday @davea @dracos (in theory Harrogate may ask for this to be deployed before Christmas -- happy to push back if anything makes you uncomforable)


stevenday:
Another minor comment, the little menu icon that displays the mobile menu on small screens is still yellow, rather than matching the blue of the theme:

screen shot 2015-01-05 at 10 49 14

perllib/FixMyStreet/Cobrand/Harrogate.pm
```diff
@@ -0,0 +1,246 @@
...
+sub enter_postcode_text {
+    my ($self) = @_;
+    return 'Enter a Harrogate district postcode, or street name and area';
+}
+
+# increase map zoom level so street names are visible
+sub default_map_zoom { return 3; }
+
+
+=head2 temp_update_contacts
```
stevenday: Looks like this docstring needs to move down a method?

perllib/FixMyStreet/Cobrand/Harrogate.pm
```diff
@@ -0,0 +1,246 @@
...
+        deleted => 0
+    })->update({
+        deleted => 1,
+        editor => 'automated script',
+        whenedited => \'NOW()',
+        note => 'Deleted by script as per requirements Dec 2014',
+    });
+}
+
+#sub get_geocoder {
```
stevenday: Could this be either un-commented or removed altogether?

templates/email/harrogate/submit.txt
```diff
@@ -0,0 +1,43 @@
...
+
+<?=$values['signature']?>
+
+This message was sent via FixMyStreet, a project of UKCOD, registered charity
+number 1076346. If there is a more appropriate email address for messages about
+<?=$values['category_footer']?>, please let us know by visiting <https://www.fixmystreet.com/contact>.
+This will help improve the service for local people. We
+also welcome any other feedback you may have.
+
+FixMyStreet is now available for full integration into council
```
stevenday: Is this needed in this footer? I'm not sure what the convention is, but I'd guess that something that mentions it's come from a cobrand would be better, so that whoever receives it knows they can contact us commercially if there's something wrong?

web/cobrands/harrogate/layout.scss
```diff
@@ -0,0 +1,183 @@
...
+    .ie6 body.authpage #site-logo,
+
+    body.twothirdswidthpage #site-logo,
+    .ie6 body.twothirdswidthpage #site-logo
+    {
+            position: relative;
+            top: 1.2em;
+            margin: 0;
+            padding: 0;
+            // background: url("/cobrands/harrogate/HBCLogo-cropped.png") 0 0 no-repeat;
```
stevenday: Can this be removed if it's not being used?

web/cobrands/harrogate/layout.scss
```diff
@@ -0,0 +1,183 @@
...
+        color: #fff;
+        @include border-radius(0em);
+    }
+}
+
+#main-nav {
+    ul#main-menu {
+        li {
+            span {
+                // color: $nav_fg;
```
stevenday: This `span` style can be removed if there's nothing in there..

stevenday:
The problem confirmation message page could do with a bit more padding or margin between the heading and the content I think:

screen shot 2015-01-05 at 11 32 05

stevenday:
The change password page seems to lose some of the header styling a bit:

screen shot 2015-01-05 at 11 34 44

stevenday:
Final comment, sorry! It looks like we could do with cadging the favicon from harrogate.gov.uk, if that's a thing we usually do for cobrands?

templates/email/harrogate/submit.txt
osfameron: good point, though as far as I can see, most existing cobrands, which *don't* specialise submit.txt at all, will get the same text.  So will probably leave this unless client requests a different wording.

osfameron:
Have fixed the little CSS fixes and code comments.

I've not changed the yellow icon, as it requires specialising and editing a CSS sprite (will open a ticket, as I don't think this is a blocker)

PR #980, opened 2014-12-30
Fixed show/hide pins and show/hide old for norwegian cobrand.

mhalden:
This updates map-OpenLayer.js for the current norwegian translation, so show/hide pins and show/hide old will work with Norwegian language.


dracos:
Thanks, sorry for the delay :)

PR #994, opened 2015-01-12
Use same handling for cron and non-cron email.

### LINKED ISSUE #988: SSL or authenticated SMTP don't work for cron emails
The SSL and authenticated SMTP options aren't implemented in the code that handles sending emails from cron hence mail sending fails.

dracos:
This means that e.g. SMTP authentication is used when set up by all
emails, not just non-cron ones. Fixes #988.

### DIFF
--- a/perllib/FixMyStreet/App.pm
+++ b/perllib/FixMyStreet/App.pm
@@ -10,7 +10,6 @@ use FixMyStreet;
 use FixMyStreet::Cobrand;
 use Memcached;
 use mySociety::Email;
-use mySociety::EmailUtil;
 use mySociety::Random qw(random_bytes);
 use FixMyStreet::Map;
 
@@ -348,9 +347,19 @@ sub send_email {
 }
 
 sub send_email_cron {
-    my ( $c, $params, $env_from, $env_to, $nomail, $cobrand, $lang_code ) = @_;
+    my ( $c, $params, $env_from, $nomail, $cobrand, $lang_code ) = @_;
 
-    return 1 if $c->is_abuser( $env_to );
+    my $first_to;
+    if (ref($params->{To}) eq 'ARRAY') {
+        if (ref($params->{To}[0]) eq 'ARRAY') {
+            $first_to = $params->{To}[0][0];
+        } else {
+            $first_to = $params->{To}[0];
+        }
+    } else {
+        $first_to = $params->{To};
+    }
+    return 1 if $c->is_abuser($first_to);
 
     $params->{'Message-ID'} = sprintf('<fms-cron-%s-%s@%s>', time(),
         unpack('h*', random_bytes(5, 1)), FixMyStreet->config('EMAIL_DOMAIN')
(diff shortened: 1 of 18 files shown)


perllib/FixMyStreet/App.pm
```diff
@@ -387,15 +396,16 @@ sub send_email_cron {
...
-        return 0;
-    } elsif (!$nomail) {
-        return mySociety::EmailUtil::send_email( $email, $env_from, @$env_to );
-    } else {
+    if ($nomail) {
         print $email;
         return 1; # Failure
+    } else {
+        my %model_args;
+        if (!FixMyStreet->test_mode && $env_from == FixMyStreet->config('CONTACT_EMAIL')) {
```
struan: eq not == here I think.

perllib/FixMyStreet/App.pm
```diff
@@ -348,9 +347,19 @@ sub send_email {
...
+    if (ref($params->{To}) eq 'ARRAY') {
+        if (ref($params->{To}[0]) eq 'ARRAY') {
+            $first_to = $params->{To}[0][0];
+        } else {
+            $first_to = $params->{To}[0];
+        }
+    } else {
+        $first_to = $params->{To};
+    }
+    return 1 if $c->is_abuser($first_to);
```
struan: On a side note, do we do all this so we don't send questionnaires or the like to people in the abuse table?

perllib/FixMyStreet/App.pm
```diff
@@ -387,15 +396,16 @@ sub send_email_cron {
...
-    } elsif (!$nomail) {
-        return mySociety::EmailUtil::send_email( $email, $env_from, @$env_to );
-    } else {
+    if ($nomail) {
         print $email;
         return 1; # Failure
+    } else {
+        my %model_args;
+        if (!FixMyStreet->test_mode && $env_from == FixMyStreet->config('CONTACT_EMAIL')) {
+            $model_args{mailer} = 'FixMyStreet::EmailSend::ContactEmail';
```
struan: Does this ever get used? It looks like it's always overridden.

struan:
Looks like this works. Few small comments one of which is me just being unsure about why something is doing what it is rather than a problem with the code.

perllib/FixMyStreet/App.pm
dracos: Whoops!

perllib/FixMyStreet/App.pm
dracos: Yes, and previously I don't think this was working, because `$env_to` was an arrayref, and it looks like `is_abuser` assumes it's a string. So this is just parsing out a To address to check (and To line can take a few forms annoyingly, and didn't seem worth parsing into an Email object just for that).

perllib/FixMyStreet/App.pm
dracos: If you're not running in test mode (as then you always want the mailer to be `Test`) if the from is the contact address, this sets the mailer to the one that has a sender envelope also of the contact address. It looks to be okay with a quick print debugging, where do you think it's overridden?

struan: OK, this is just me not understanding how this is working :(

PR #998, opened 2015-01-14
for ESCC, use Dynamics-friendly format for location

### LINKED ISSUE #997: format easting/northing as E/N single string  in report email for ease of cut-and-paste
Dynamics CRM accepts Easting/Northing as a single string with a slash delimiter, no spaces; council staff currently have to cut and paste twice to transfer location information from an incoming email notification in Dynamics (which then produces a map).

Either add it conditionally for ESCC who I saw doing this, or add it as an extra line on all default templates in case the receiving council staff are using the same Dynamics system?

davewhiteland:
Potentially close #997... but maybe this shouldn't just be ESCC?

Not sure about selecting the council here: not only do I not know if this is really the best way to do it, but should this be conditional at all since it may benefit anyone else using Dynamics without back-end integration? Does it break anything to either add it as a third line of easting/northing info?

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/Problem.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Problem.pm
@@ -303,10 +303,14 @@ sub send_reports {
             ( $h{easting}, $h{northing} ) = Utils::convert_latlon_to_en( $h{latitude}, $h{longitude} );
 
             # email templates don't have conditionals so we need to farmat this here
-            $h{easting_northing}                             #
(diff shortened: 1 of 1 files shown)


dracos:
I'd be happy switching it to always output on one line and we can then see if anyone minds/cares/notices (The submit-example.txt file should then be updated too.)

BenJam:
let's just do it for all local councils.

dracos:
Closed by b1a719a

Issue #978, opened 2014-12-30
Improve make_css bin spotting via symlink


dracos:
The script locates the gem-bin directory fine when run as e.g. `/var/www/fixmystreet.127.0.0.1.xip.io/fixmystreet/bin/make_css` but doesn't find it when run as `/home/fms/fixmystreet/bin/make_css` (where `fixmystreet` is a symlink there), as it's not getting the absolute path worked out correctly.

This doesn't matter currently as the `fms` user gets the `gem-bin` added to their `$PATH` anyway, but in case you're in a situation where that doesn't happen it should probably be fixed.

PR #999, opened 2015-01-15
Some internals improving

### LINKED ISSUE #852: document cron-wrapper for running the scripts in /bin
Although examples in the docs use it, I don't think we explain the generally useful `cron-wrapper` anywhere (maybe a `bin/README`, since perhaps it's going to catch out people thinking they can run scripts they find in there?), and although the

`Can't locate FixMyStreet/App.pm in @INC ...`

error does indicate the problem it's not obvious from there that `cron-wrapper` is the solution.

dracos:
Remove need for using `cron-wrapper` when running scripts, improve make_css when run via a symlink, and rename test-wrapper to run-tests. Fixes #852, #978.


osfameron:
Looks good! :+1:

PR #1000, opened 2015-01-15
Move site_title function to site-name web template

### LINKED ISSUE #979: Remove need for `site_title` in cobrand modle
It seems a bit much to require a `.pm` file to be created to simply change the site title. Emails already have this sorted with a `site-name.txt` template - we should have a similar template for the HTML that is used by things needing it. I don't think there are very many that would need changing:

- [x] `send_email_cron()` falls back to `site_title`
- [x] `Report.pm` - 403 non-public report message
- [x] Questionnaire email sending sets site_name, I think it shouldn't as this is set by `send_email_cron`
- [x] Admin config and index pages
- [x] Privacy and FAQ pages
- [x] fixmystreet `header.html` passes in `site_title` for &lt;title&gt;
- [x] fixmystreet `header_logo` and `opengraph` templates use `site_title`

dracos:
This matches the email template behaviour, and makes it much easier to
override without requiring a cobrand package file.

Fixes #979.


osfameron:
This looks good -- slightly confusing that there are now 2 site-titles (I note that Bromley has different versions for web and email, so this seems to be a useful feature).  Presumably this now needs documentation by @davewhiteland too.

dracos:
I was going to document it once it was merged.

PR #1007, opened 2015-01-21
Issues/989/jump to report

### LINKED ISSUE #989: Jump to a specific report
A commercial client have asked how they can jump to a specific report, e.g. if quoted by a customer.

I've previously wanted this in /admin/reports, but I think it would be a good idea to add to /dashboard, and will add this unless any objections.

(Yes, it's "easy" to hack the URL, but that's hard to explain to less technical users.)

osfameron:
Requested by a council client. Fixes #989

I've deployed to staging, so you can view on http://commercial.fixmystreet.staging.mysociety.org/dashboard if you are from_body.

Comments on the design (e.g. location of the jump-to box) very welcome!


dracos:
Out of date.

PR #1008, opened 2015-01-23
Expire cached geolocations after a week.

### LINKED ISSUE #684: Expire Bing cached geolocations
Bromley reported that "Henville Rd" wasn't resolving correctly.  There was a /data/vhost/www.fixmystreet.com/cache/bing/$HASH record for that term with no data, presumably because Bing had errored when it was originally retrieved.

   - [ ] we should expire any lookup that returns bad data?
   - [ ] we should expire old lookups (as Bing should, presumably, periodically get better, as badly geolocated searches are reported to fixed.)

dracos:
Refactor geoloction caching code into one function.
Fixes #684

### DIFF
--- a/perllib/FixMyStreet/Geocode.pm
+++ b/perllib/FixMyStreet/Geocode.pm
@@ -7,6 +7,11 @@
 package FixMyStreet::Geocode;
 
 use strict;
+use Digest::MD5 qw(md5_hex);
+use Encode;
+use File::Slurp;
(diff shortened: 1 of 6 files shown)


osfameron:
Looks good.  Have commented inline about confusing regexp that could do with tweak/comment.

Notice the ::G::Zurich doesn't call the new routine, is there a reason for that?  May be worth commenting in the commit.

PR #1009, opened 2015-01-23
Make sure all co-ordinates are stringified.

dracos:
This includes MapIt postcode lookups, geocoding, query parameters, tile
clicks. This way, we no longer need any "short" versions anywhere, and
the JSON response always uses a decimal point regardless of locale.

Fixes https://github.com/mysociety/fixmystreet-mobile/issues/190

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Alert.pm
+++ b/perllib/FixMyStreet/App/Controller/Alert.pm
@@ -441,11 +441,6 @@ sub determine_location : Private {
         $c->go('index');
     }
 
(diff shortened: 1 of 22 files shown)


jonkri:
Great! I look forward to seeing this merged. Thanks!

osfameron:
Some comments inline.  I don't really understand the intent of this code, so to give an enthusiastic bunny icon, I'd probably want to have a quick walk-through to explain it.  (But haven't spotten any obvious problems, though I think the change of return value for one of the controller actions could do with a bit of doc tweak).

PR #1015, opened 2015-02-07
Remove need to have to specify en-gb in LANGUAGES.

dracos:
Instead, add it if not present at the negotiation stage. Even if not negotiated, we currently need it to be present for the in_gb_locale() calls to work.


davewhiteland:
:+1:  I like removing that not least cos iirc a space between the two could cause problems too.

PR #1024, opened 2015-02-17
[Stevenage] tracking code

osfameron:
For release today hopefully

### DIFF
--- a/templates/web/stevenage/tracking_code.html
+++ b/templates/web/stevenage/tracking_code.html
@@ -1,0 +1,16 @@
+[% IF c.config.BASE_URL == "https://www.fixmystreet.com" %]
+
+<script>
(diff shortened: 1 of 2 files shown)


templates/web/stevenage/tracking_code.html
```diff
@@ -0,0 +1,16 @@
+[% IF c.config.BASE_URL == "https://fixmystreet.stevenage.gov.uk" %]
```
davea: I don't think this is right - the `BASE_URL` is still going to be `https://www.fixmystreet.com` as it's just a cobrand on the main site, not a full FMS instance in its own right. See [Oxfordshire](https://github.com/mysociety/fixmystreet/blob/master/templates/web/oxfordshire/tracking_code.html), for example.

davea:
Looks good :+1:

PR #1026, opened 2015-02-18
Order 'all reports' page by confirmation date

davea:
This PR:

 * Allows cobrands to specify their own `order_by` clause for the `/reports/[body]` page
 * Makes the Oxfordshire cobrand order reports by descending order of creation date

Closes mysociety/FixMyStreet-Commercial#681

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
@@ -53,6 +53,9 @@ sub pin_colour {
     return 'green' if $p->is_fixed || $p->is_closed;
     return 'red' if $p->state eq 'confirmed';
     return 'yellow';
+
+sub reports_ordering {
+    return { -desc => 'confirmed' };
 }
 
 1;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -47,5 +47,9 @@ sub users_can_hide { return 1; }
 
 sub default_show_name { 0 }
 
+sub reports_ordering {
+    return { -desc => 'created' };
```
dracos: Could we use `confirmed` rather than `created`? Created is really an internal date field not exposed externally, confirmed is the date the thing appears on the site and the date given as the "creation" of this report publicly. The order will normally be the same, but it's possible someone takes a while to confirm their report and then going by created order would look a bit wrong.

davea: Ooh, yes, thanks for the pointer. Will update accordingly.

PR #1027, opened 2015-02-18
Add differing pin colours to Oxfordshire cobrand

davea:
This also adds a new grey pin which is used by the Oxfordshire
cobrand for reports that aren't the Council's responsibility.


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -47,5 +47,14 @@ sub users_can_hide { return 1; }
 
 sub default_show_name { 0 }
 
+sub pin_colour {
+    my ( $self, $p, $context ) = @_;
+    return 'green' if ($p->is_fixed || $p->is_closed) && $p->state ne 'not responsible';
+    return 'red' if $p->state eq 'unconfirmed' || $p->state eq 'confirmed';
+    return 'grey' if $p->state eq 'not responsible';
```
dracos: If you put this return statement first, you wouldn't need the extra check on the green line :)

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -47,5 +47,14 @@ sub users_can_hide { return 1; }
 
 sub default_show_name { 0 }
 
+sub pin_colour {
+    my ( $self, $p, $context ) = @_;
+    return 'green' if ($p->is_fixed || $p->is_closed) && $p->state ne 'not responsible';
+    return 'red' if $p->state eq 'unconfirmed' || $p->state eq 'confirmed';
```
dracos: No unconfirmed report should be being shown on the site, so the first part of this line isn't needed.

dracos:
:+1: I wondered if they should start to fade out as they get old and don't receive updates (not here where a council will be updating them, but in general)

Issue #1018, opened 2015-02-10
Refactor to encapsulate operations that work on {extra} fields in Contact and Problem


osfameron:
## Problem 1: mismatching styles

While working on https://github.com/mysociety/FixMyStreet-Commercial/issues/665 we've had a few thorny issues with the {extra} field.

This is a NoSQL serialized field (using an old mySoc internal format called RABX for historical reasons) that tends to have one of two layouts:

 - AoH: list of fields required for a given category in format [ { name=>'field1', required => 1, ...}, { ... }, ... ]
 - Hash: particular data tagging the contact or problem with some particular data.

UK and generic FMS use the former scheme.  EmptyHomes and Zurich use the latter scheme.

If Open311 style contacts are used, the data will be dragged from the Contact's "fields" (AoH) into Problem's extra in AoH style.  Unfortunately, some of Zurich's new requirements suggest that we use this style... but Zurich's problems are in Hash format.  Boom!

## Problem 2: hard to manipulate

Because there are no routines to edit {extra} fields, we end up doing things like:

    my $extra = $row->extra;
    push @$extra, { ... };
    $row->extra( $extra );

which is somewhat horrible, and fragile (especially given that in Problem 1 above, we can't necessarily guarantee that the data is in the correct format.)  If you attempt to handle a field in the wrong way, you get an ugly Perl error, which can't be encapsulated into anything nicer.

## Proposed API for extra fields

This will be a Role that can be applied to problems, contacts, etc.

 It will handle "fields" in the following way:

 - if the layout is an array, then will use that array directly.
 - if the layout is a hash, then will use/create an entry called `_fields` as an array.
 - if there is no layout, then will check with `$c->cobrand` which layout to use

Methods will be:

 - `add_extra_field({ ... })`  - pushes to end
 - `get_extra_field('name')`  - searches for an entry with `name => 'name'`
 - `set_extra_field`({ ... }) - finds and updates an entry with `name => $record->{name}` or else pushes it
 - `set_extra_fields(...)` - replaces the entire fields entry with the supplied array
 - `get_extra_fields` - returns the existing fields as an array, to be iterated over

It will handle metadata as follows

 - if the layout is an array, then this is a noop.  e.g. will return undef or an empty hash as appropriate
 - if the layout is a hash, then will use that hash (ignoring the special `_fields` array)
 - if there is no layout, then will check with `$c->cobrand` which layout to use

Methods will be 

[...]

PR #1033, opened 2015-02-24
Roles::Extra to give an API to {extra} field #1018

osfameron:
A number of Zurich changes in https://github.com/mysociety/FixMyStreet-Commercial/issues/672 build on this, so would be useful to review first.

See #1018

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -636,8 +636,9 @@ sub setup_categories_and_bodies : Private {
             unless ( $seen{$contact->category} ) {
                 push @category_options, $contact->category;
 
-                $category_extras{ $contact->category } = $contact->extra
-                    if $contact->extra;
+                my $metas = $contact->get_extra_fields;
+                $category_extras{ $contact->category } = $metas
+                    if scalar @$metas;
 
                 $non_public_categories{ $contact->category } = 1 if $contact->non_public;
             }
@@ -870,8 +871,9 @@ sub process_report : Private {
             if $body_string && @{ $c->stash->{missing_details_bodies} };
         $report->bodies_str($body_string);
 
-        my @extra = ();
-        my $metas = $contacts[0]->extra;
+        my @extra;
+        # NB: we are only checking extras for the *first* retrieved contact.
+        my $metas = $contacts[0]->get_extra_fields();
 
         foreach my $field ( @$metas ) {
             if ( lc( $field->{required} ) eq 'true' ) {
@@ -894,7 +896,7 @@ sub process_report : Private {
 
         if ( @extra ) {
             $c->stash->{report_meta} = { map { $_->{name} => $_ } @extra };
-            $report->extra( \@extra );
+            $report->set_extra_fields( @extra );
         }
     } elsif ( @{ $c->stash->{bodies_to_list} } ) {
 
--- a/perllib/FixMyStreet/App/Controller/Tokens.pm
+++ b/perllib/FixMyStreet/App/Controller/Tokens.pm
@@ -66,10 +66,8 @@ sub confirm_problem : Path('/P') {
     # For Zurich, email confirmation simply sets a flag, it does not change the
     # problem state, log in, or anything else
     if ($c->cobrand->moniker eq 'zurich') {
-        my $extra = { %{ $problem->extra || {} } };
-        $extra->{email_confirmed} = 1;
+        $problem->set_extra_metadata( email_confirmed => 1 );
         $problem->update( {
-            extra => $extra,
             confirmed => \'ms_current_timestamp()',
         } );
 
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -4,6 +4,7 @@ use base 'FixMyStreet::Cobrand::Default';
 use DateTime;
 use POSIX qw(strcoll);
 use RABX;
+use Scalar::Util 'blessed';
 
 use strict;
 use warnings;
@@ -158,7 +159,8 @@ sub updates_as_hashref {
         $hashref->{update_pp} = $self->prettify_dt( $problem->lastupdate );
 
         if ( $problem->state eq 'fixed - council' ) {
-            $hashref->{details} = FixMyStreet::App::View::Web->add_links( $ctx, $problem->extra ? $problem->extra->{public_response} : '' );
+            $hashref->{details} = FixMyStreet::App::View::Web->add_links(
+                $ctx, $problem->get_extra_metadata('public_response') || '' );
         } elsif ( $problem->state eq 'closed' ) {
             $hashref->{details} = sprintf( _('Assigned to %s'), $problem->body($ctx)->name );
         }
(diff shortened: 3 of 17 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -889,8 +889,9 @@ sub process_report : Private {
             if $body_string && @{ $c->stash->{missing_details_bodies} };
         $report->bodies_str($body_string);
 
-        my @extra = ();
-        my $metas = $contacts[0]->extra;
+        my @extra;
+        # NB: we are only checking extras for the *first* retrieved contact.
+        my $metas = $contacts[0]->get_extra_fields($c);
```
dracos: contact `extra` is also looked up on line 652 (for use by the `category_extras` ajax).

perllib/FixMyStreet/Roles/Extra.pm
```diff
@@ -0,0 +1,170 @@
...
+    my ($self, $c, $key) = @_;
+    my $extra = $self->get_extra($c);
+
+    return unless ref $extra eq 'HASH';
+    return 1 unless exists $extra->{$key};
+    delete $extra->{$key};
+    $self->extra($extra);
+};
+
+=head2 set_extra_metadata
```
dracos: Typo.

perllib/FixMyStreet/Roles/Extra.pm
```diff
@@ -0,0 +1,170 @@
...
+
+=head2 get_extra
+
+Get the extra data.  If this is not set, then returns a {} or [] reference
+I<as appropriate>, e.g. depending on the Cobrand's expected layout.  This
+worrying about whether we should return a hash or array reference is the reason
+that we have to pass $c to Every Single Method (i.e. so that we can get access
+to C<$c-E<gt>cobrand>.
+
+(This is a bit of a wart, can we do better?)
```
dracos: As this doesn't need to worry about problem cobrand not matching site cobrand, you could perhaps access the cobrand on $self->cobrand (though it won't be a class). However, I wonder if it would be worth changing the other places in the codebase that assume extra to be an arrayref to use new functions and have get_extra always return a hashref. It would simplify much of the code of this file, and there only appears to be four or five places to do it. In fact, could you then also use an `after` on Problem's `extra` rather than a separate `get_extra` function, so that it always returned a hashref regardless of what old data might be in the database.

dracos:
Apart from the place in `New.pm` mentioned above, the other places that assume extra is an arrayref are: `Harrogate.pm` additional email function, `SendReport/Open311.pm` adding things to extra, `Open311.pm` looping over extra, and `PopulateServiceList.pm` setting extra. (Plus the manually run set up functions in Harrogate and EastSussex). If those used `get/set_extra_fields`, then the role could assume everything is always a hashref, I think, and that would prevent confusion in the rest of the codebase.

osfameron:
To make a minor change in ::DB::RS::Problem->send_reports, I might want to refactor to remove the need for passing $c to every method.  That would probably mean doing the refactor to change layout to hash for everything, which in turn means changing *all* uses of extra throughout codebase.  Will try to do this if I get time this week.

osfameron:
Unsure why red from Travis, I've rerun there, and build seems to be OK?

t/app/model/extra.t
```diff
@@ -0,0 +1,111 @@
+use strict;
+use warnings;
+use Test::More;
+use utf8;
+
+use FixMyStreet::App;
+use Data::Dumper;
+use DateTime;
+
+my $ c = FixMyStreet::App->new;
```
dracos: Strange space here.

t/app/controller/report_new_open311.t
```diff
@@ -169,7 +169,7 @@ foreach my $test (
         my $prob = $user->problems->first;
         ok $prob, 'problem created';
 
-        is_deeply $prob->extra, $test->{extra}, 'extra open311 data added to problem';
+        is_deeply $prob->extra, { _fields => $test->{extra} }, 'extra open311 data added to problem';
```
dracos: Should this use `get_extra_fields`?

perllib/Open311/PopulateServiceList.pm
```diff
@@ -266,7 +266,7 @@ sub _add_meta_to_contact {
     }
 
     if ( @meta ) {
-        $contact->extra( \@meta );
+        $contact->extra( { _fields => \@meta } );
```
dracos: Is there a reason this isn't `set_extra_fields`?

perllib/FixMyStreet/App/Controller/Tokens.pm
```diff
@@ -66,10 +66,8 @@ sub confirm_problem : Path('/P') {
     # For Zurich, email confirmation simply sets a flag, it does not change the
     # problem state, log in, or anything else
     if ($c->cobrand->moniker eq 'zurich') {
-        my $extra = { %{ $problem->extra || {} } };
-        $extra->{email_confirmed} = 1;
+        $problem->set_extra_metadata( $c, email_confirmed => 1 );
```
dracos: Presumably this `$c` should not be here. Please include a test along with the fix for this to show that this code is being tested.

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -619,9 +626,7 @@ sub admin_report_edit {
 
             # If they clicked the no more updates button, we're done.
             if ($c->req->param('no_more_updates')) {
-                my $extra = $problem->extra || {};
-                $extra->{subdiv_overdue} = $self->overdue( $problem );
-                $problem->extra( $extra );
+                $problem->set_extra_metadata_if_undefined( subdiv_overdue => $self->overdue( $problem ) );
```
dracos: The old code wasn't only if undefined (I don't know if this matters), presumably not, though I guess an edge case of a report going backwards in process then forwards again?

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -532,12 +541,10 @@ sub admin_report_edit {
...
         # Final, public, Update from DM
         if (my $update = $c->req->param('status_update')) {
-            $extra->{public_response} = $update;
-            $problem->extra( $extra );
+            $problem->set_extra_metadata(public_response => $update);
             if ($c->req->params->{publish_response}) {
                 $self->set_problem_state($c, $problem, 'fixed - council');
-                $extra->{closed_overdue} = $self->overdue( $problem );
-                $problem->extra( { %$extra } );
+                $problem->set_extra_metadata_if_undefined( closed_overdue => $self->overdue( $problem ) );
```
dracos: Similar to below comment.

dracos:
There are tests in `populate-service-list.t` that need updating, e.g. line 411 is setting `extra` directly rather than using `set_extra_fields`. And `SendReport/Open311.pm` could be broken because it calls row->extra directly in a number of places (this also requires a test to show this). And for completeness Report/New.pm emptyhomes section could use the new API.

perllib/Open311/PopulateServiceList.pm
osfameron: oh yes, `$contact->set_extra_fields(@meta)` doesn't even need the `if` (as an empty meta just sets `[]`)

PR #1036, opened 2015-02-26
Improve report print styles

zarino:
This was a request from a commercial customer (Zürich) but I've implemented the fix for all cobrands, since it's probably beneficial for everyone.

Before:

screen shot 2015-02-26 at 17 20 28

After:

screen shot 2015-02-26 at 17 20 07

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -889,8 +889,9 @@ sub process_report : Private {
             if $body_string && @{ $c->stash->{missing_details_bodies} };
         $report->bodies_str($body_string);
 
-        my @extra = ();
-        my $metas = $contacts[0]->extra;
+        my @extra;
+        # NB: we are only checking extras for the *first* retrieved contact.
+        my $metas = $contacts[0]->get_extra_fields($c);
 
         foreach my $field ( @$metas ) {
             if ( lc( $field->{required} ) eq 'true' ) {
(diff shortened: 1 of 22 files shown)


dracos:
There are some changes to the CSS in the commit that don't appear to be to do with the print styling. I see what the 10px removal is for easily enough, what is the other change for? It means the hover over the nav link goes a bit wrong, as it's now lower down. If that change is made, then is the `static` still needed in layout (with IE7 fix on top of that) or can that all be dropped at the same time? [Something to do with unclickable nav links, apparently.] Either way, this should probably be in a separate commit.

For the changes to apply to other cobrands as your comment says, do you need to apply the `screen and` to other `header.html` files?

zarino:
The changes in `web/cobrands/zurich/base.scss` are to stop the logo floating off to the top left of the page when you print:

screen shot 2015-03-03 at 17 01 21

It looked like styling specific to Zurich, so I didn't put it in the `@media print` block in `web/cobrands/sass/_base.scss` – maybe you'd prefer it in a `@media print` block in `web/cobrands/zurich/base.scss` to make it clearer?

zarino:
And yes, if other cobrands have overridden the `header.html` template, they'll need to add `screen and` to the `layout.scss` link to stop the print styles being trampled by `layout.scss`.

That feels like a step beyond this scope of the original ticket – and whoever does it should probably also check each cobrand they edit to make sure there aren't cobrand-specific issues that get in the way of the print layout (like the floating `#site-logo` in Zurich, for example).

dracos:
"maybe you'd prefer it in a @media print block in web/cobrands/zurich/base.scss to make it clearer?" - well, as I say it breaks the nav hover, so something needs to be changed :)

zarino:
> well, as I say it breaks the nav hover, so something needs to be changed :)

@dracos I've moved the `#site-logo` changes into an `@media print` block in `web/cobrands/zurich/base.scss`

PR #1038, opened 2015-03-02
Add event tracking to "problems nearby" tab

jacksonj04:
* Solves https://github.com/mysociety/research/issues/46

### DIFF
--- a/web/js/fixmystreet.js
+++ b/web/js/fixmystreet.js
@@ -207,6 +207,7 @@ $(function(){
     // we disable validation by default, and add it in the JS case.
     // For some reason, the removeAttr doesn't work if we place it at beginning.
     $('#mapForm').removeAttr('novalidate');
+
 });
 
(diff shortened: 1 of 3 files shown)


web/js/fixmystreet.js
```diff
@@ -207,6 +207,25 @@ $(function(){
...
+        var name = $(this).attr('data-clicktrack')
+        var callback = function(){
+            window.location.href = url
+        }
+        if(e.metaKey || e.ctrlKey){
+            callback = function(){}
+        } else {
+            e.preventDefault()
+        }
+        ga('send', 'event', 'action', 'click', name, {
```
dracos: Not everyone uses Google Analytics, so `ga` might not exist. It would be good if this could be a function that then the code added in posters.html could also reuse.

dracos:
Tabs do not exist any more.

PR #1041, opened 2015-03-04
Colour pins by status; size them according to zoom level

davea:
This PR implements the following behaviour on the Oxfordshire cobrand:

 * Colours map pins according to the report status.
 * Decreases the size of pins as the map is zoomed out to reduce clutter.

Fixes mysociety/FixMyStreet-Commercial#679.
Fixes mysociety/FixMyStreet-Commercial#683.


web/js/map-OpenLayers.js
```diff
@@ -90,6 +91,34 @@ function fms_markers_list(pins, transform) {
     return markers;
 }
 
+function fms_marker_size_for_zoom(zoom) {
+    // zoomOffset seems to change between pages, so check that the zoom level
+    // we're working with is in the expected range
```
dracos: Could you explain what this bit is for? If you added the zoomOffset, in the other call to `fms_marker_size_for_zoom` (line 115), would this be needed? Or could you drop the adding of zoomOffset from both calls and just have it here?

web/js/map-OpenLayers.js
```diff
@@ -90,6 +91,34 @@ function fms_markers_list(pins, transform) {
...
+    if (zoom < fixmystreet.zoomOffset) {
+        zoom += fixmystreet.zoomOffset;
+    }
+    var cobrand = $('meta[name="cobrand"]').attr('content');
+    if (cobrand === 'oxfordshire') {
```
dracos: I would be happy for this to apply to all cobrands, as it makes it better for all, unless there's a reason not to do so.

web/js/map-OpenLayers.js
```diff
@@ -144,6 +173,25 @@ function fixmystreet_onload() {
             backgroundHeight: 40,
             backgroundXOffset: -10,
             backgroundYOffset: -35
+        },
+        'small': {
+            externalGraphic: fixmystreet.pin_prefix + "pin-${colour}-small.png",
+            graphicWidth: 24,
+            graphicHeight: 32,
+            graphicXOffset: -12,
+            graphicYOffset: -32,
```
dracos: I think the offset numbers need looking at, as the bubbles don't appear in the right places like they do on the normal sized pins.

dracos:
Nice :) In terms of general commits, I would suggest (if it's okay to do this for all cobrands), merging all but the first commit together in one commit. If it's not okay, then I'd probably merge cfce3b4, 80ef35f, and the mini "setup" part of a960c3f together in one "new pins" commit, and then merge 90837f8 and the rest of a960c3f in a "using these new pins in Oxfordshire" commit. :rabbit:

web/js/map-OpenLayers.js
davea: No specific reason; I was taking a 'light touch' approach and trying not to affect other cobrands. I'll remove the Oxfordshire restriction.

web/js/map-OpenLayers.js
davea: Good catch; thanks. I think the offset numbers here are to do with positioning of the marker image relative the the point it's marking. It looks as though the bubble offset is set [elsewhere](https://github.com/mysociety/fixmystreet/blob/master/web/js/map-OpenLayers.js#L193) and will need to vary depending on the marker size in use.

web/js/map-OpenLayers.js
davea: `fixmystreet.map.getZoom()` seems to include `zoomOffset` on `/around` but not on `/reports`. I couldn't figure out what was going on and this fixed the problem. Will have another look into it and see if I can get rid of this.

web/js/map-OpenLayers.js
dracos: Oh yes, quite right, not as straightforward as I'd imagined, sorry.

web/js/map-OpenLayers.js
dracos: On `/reports` zoomOffset is 0 (because you can have any zoom level) whilst on `/around` it’s 13 (the minimum zoom level), but the adding of getZoom+zoomOffset should then be the same at the same resulting zoom level on both (ie. getZoom will return 3 on `/around` and 16 on `/reports` for the same map tiles).

web/js/map-OpenLayers.js
davea: If you're happy with my solution to this in 1448fdccbb3a3f744c0a030ea3a13afad565f617 I'll squash it into the previous commit and merge this PR.

PR #1042, opened 2015-03-04
Add expected response time to report page

davea:
* Adds a new `problem_response_days` to the `Default` cobrand that subclasses can override to provide an expected response time based on a problem's category.
 * OCC-provided response time figures are implemented in the `Oxfordshire` cobrand.
 * The response time is displayed at the top of the report when viewed in the Oxfordshire cobrand.

Fixes mysociety/FixMyStreet-Commercial#680.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Default.pm
+++ b/perllib/FixMyStreet/Cobrand/Default.pm
@@ -899,4 +899,3 @@ sub get_country_for_ip_address {
 }
 
 1;
-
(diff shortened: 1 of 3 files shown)


dracos:
I'd rename this commit to start with `[Oxfordshire]`. I realise it does add a default function, but nothing else uses it – actually, as this is just for Oxfordshire, I'm not sure it's worth having the function in `Default.pm`; one can always be added in future if it makes sense to do so, though I think it's probably something that won't ever require a default. Other than that, :+1:

davea:
Thanks, I'll make the changes you suggest.

PR #1043, opened 2015-03-04
Allow cobrands to specify report ordering on 'all reports' page

davea:
* Adds a new `reports_ordering` sub to the `Default` cobrand that's used to control ordering of reports in `Reports.load_and_group_problems`.
 * Orders reports by confirmed date in the `Oxfordshire` cobrand.

Fixes mysociety/FixMyStreet-Commercial#681.


dracos:
:+1:

Issue #682, opened 2014-03-03
Change Chelmsford to a city council


MyfanwyNixon:
Requested by them, and by a user - they received city status last year.

PR #1044, opened 2015-03-04
Oxfordshire: Full screen map design

davea:
This PR:

 * Uses a full-screen map on the `/around`, `/my`, and `/reports` pages in the Oxfordshire cobrand
 * Combines the 'around the map' and 'nearby' tabs on the report list into a single list
 * Adds filter fields atop the report list to filter by status/category. NB: These are non-functional placeholders in this PR.

Fixes mysociety/FixMyStreet-Commercial#687.


templates/web/oxfordshire/around/_report_banner.html
```diff
@@ -0,0 +1,12 @@
+<h1 class="big-green-banner">
+    <!-- The mobile interface uses this element to construct
+         the "Place pin on map" message at the top of the screen.
+         But we don’t want it to show up on desktop. -->
+</h1>
```
dracos: The contents of this `h1` are used if JavaScript is not available in the mobile interface. So it should not be empty. Presumably it is being hidden by the CSS on the desktop anyway?

templates/web/oxfordshire/around/tabbed_lists.html
```diff
@@ -0,0 +1,24 @@
...
+    </label>
+    <label>
+        about
+        <select>
+            <option>anything</option>
+            <option>bridges</option>
+            <option>carriageway defect</option>
+            <option>debris/spillage</option>
+            <option>…</option>
+        </select>
```
dracos: Is a server-side way of filtering this page going to be included here? If not, at the least these drop downs should only appear if their functionality is available. But really preferably the first option… :)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -463,7 +463,8 @@ $.fn.drawer = function(id, ajax) {
                 offset = -110;
             }
             if (cobrand == 'oxfordshire') {
-                offset = -13 * 16;
+                // Oxfordshire uses box-sizing:border-box and padding to work out height
```
dracos: Is this something we can do for the default `offset`?

web/cobrands/oxfordshire/layout.scss
```diff
@@ -196,18 +398,70 @@ input.green-btn{
...
+  }
+
+  img {
+    // the little chevron icon
+    vertical-align: -1px;
+    margin-left: 0.2em;
+  }
+}
+
+#skip-this-step {
```
dracos: Most of this section is in the base stylesheet.

dracos:
Looking good :-) Hope all my comments make sense, I think after a tidy this should be okay, but there are a number of points that would be good for improving and future-proofing.

A general note on commit messages – Oxfordshire specific commits should preferably start with `[Oxfordshire]`, and please do not refer to private issue numbers as they link to completely different issues in the public repository; make sure the commit message makes sense without any knowledge of the ticket, and if you wish to link, please use a full reference e.g. mysociety/fixmystreet-commercial#682 at the end of the commit message. As it is, @zarino's "placeholder report filters and styling, for #682" links to an issue on Chelmsford City Council :)

PR #1045, opened 2015-03-04
Filter map based on report status/category

davea:
This PR:

NB: This PR includes and relies on the functionality from PR #1044, which should probably be reviewed and merged first. I didn't want to create a PR against a PR...


dracos:
Some comments where I'm not sure why some things are present, or how it might be done more simply. I'd also squash quite a few of these together, but I'd probably wait for #1044 to be tidied/done first and the other issues dealt with before having a fresh look as to which. The adding of filtering by category/state server side all looks fine.

davea:
I'm closing this in favour of a new PR that will be opened against the branch in #1044, to keep the list of commits more focussed.

Issue #676, opened 2014-02-27
show list of alerts currently subscribed-to when looking at a user in the admin


davewhiteland:
Currently it's not easy to see if someone is subscribed to an alert or not, and I think we have a use case coming up where it will help us (Delhi clinic project).

PR #1047, opened 2015-03-09
Higher information density on Zürich admin pages

zarino:
@osfameron is going to check this over, and squash commits as necessary.

Changes the admin page from looking like this on a 1024×768 screen:

screen shot 2015-03-09 at 12 36 53

To looking like this:

screen shot 2015-03-09 at 12 36 49

screen shot 2015-03-09 at 12 41 18

screen shot 2015-03-09 at 12 37 45

Major changes:

* Two-column layout, with each column being `20em` wide (versus original single `27em` column)
* Smaller text size, smaller headings, and slightly shorter textareas
* Vertically stacked admin notes (the old table layout wouldn't fit in the new 20em width)
* Simplified display of location and reporter details in blue column (as per customer's mockup)

We could pretty up the form inputs, but we're into diminishing returns. I'm sceptical of the logic behind the separation of "Answers" and "Internal notes" (rather than just having "Notes" which can optionally be flagged as "I'm done and this is what you should tell the citizen") – But I'm going to trust that some user research went on here and I shouldn't change it.


osfameron:
I'm rebasing this onto mysociety/fixmystreet#676 as that includes all the form work that I've done.  Fixing conflicts as best as I can, but may have to refer to @zarino for anything that confuses me!

osfameron:
This pull is *SUPERSEDED* by https://github.com/mysociety/fixmystreet/pull/1061
(but keeping open in case it's useful to refer to and comment on the individual commits)

PR #1049, opened 2015-03-09
Zurich: 'photo required' functionality issues/commercial/665

osfameron:
Work for https://github.com/mysociety/FixMyStreet-Commercial/issues/665
Builds on https://github.com/mysociety/FixMyStreet-Commercial/issues/664, so PR is to that in first instance. 

(you should probably look at #1050 first, sorry for counterintuitive order of PR)

- Save 'photo_required' value in Contact->extra from admin edit form
 - Enforce per-category photo requirement on new reports
    A new step has been added to the photo upload process that
    ensures a photo is present if any of the categories chosen for
    the report require it.
    If the photo is missing an error is displayed to the user in the
    same manner as if the photo upload was invalid.
 - Add test of mandatory photo categories

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -370,6 +370,14 @@ sub update_contacts : Private {
         $contact->api_key( $c->req->param('api_key') );
         $contact->send_method( $c->req->param('send_method') );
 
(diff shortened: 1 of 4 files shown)


dracos:
:+1:

dracos:
This is in the right branch for now.

PR #1054, opened 2015-03-09
Zurich: Issues/commercial/663 zurich fields and categories

osfameron:
https://github.com/mysociety/FixMyStreet-Commercial/issues/663

Note that FMS::Cobrand::EastSussex and FMS::Cobrand::Harrogate still have temporary routines that generate contacts extra as arrays.  This shouldn't be an issue as 3b7adb471 will transparently upgrade them (and they most probably won't be rerun in any case.)

But TODO: refactor those cobrands to use setup-contacts script provided, for neatness.

### DIFF
--- a/bin/setup-contacts
+++ b/bin/setup-contacts
@@ -1,0 +1,204 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+require 5.8.0;
+use feature 'say';
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
+use FixMyStreet::App;
+
+my $moniker = $ARGV[0];
+
+my $c = FixMyStreet::App->new();
+my $cobrand = FixMyStreet::Cobrand->get_class_for_moniker($moniker)->new({ c => $c });
+$c->stash->{cobrand} = $cobrand;
+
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Base.pm
```diff
@@ -53,6 +53,230 @@ sub is_default {
...
+            key => 'contacts_body_id_category_idx'
+        }
+    );
+}
+
+sub setup_contacts {
+    my ($self, $description) = @_;
+
+    my $c = $self->{c};
+    die "Not a staging site, bailing out" unless $c->config->{STAGING_SITE}; # TODO, allow override via --force
```
dracos: Repeat of code in parent.

perllib/FixMyStreet/Cobrand/Base.pm
```diff
@@ -53,6 +53,230 @@ sub is_default {
 
 # NB: this Base class is for 'meta' features.  To add base methods for all cobrands,
 # you may want to look at FMS::Cobrand::Default instead!
+=head2 setup_contacts, ensure_bodies
```
dracos: This should be in Default.pm, not Base, as per the comment just above. Although I'm not sure it makes sense to be in Default either, as it's just for development; could this code live in the `setup-contacts` script, with just the cobrand-specific data in the cobrand file? It doesn't seem like it would be shared with anything.

dracos:
I'd squash the first two commits together (along with moving all the code around, maybe), looks okay.

PR #1055, opened 2015-03-09
Zurich: Issues/commercial/669 remove unnecessary confirmation email

osfameron:
https://github.com/mysociety/FixMyStreet-Commercial/issues/669

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -850,4 +850,13 @@ sub admin_stats {
     return 1;
 }
 
+sub problem_confirm_email_extras {
+    my ($self, $report) = @_;
+    my $confirmed_reports = $report->user->problems->search({
+        extra => { like => '%email_confirmed%' },
+    })->count;
+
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -850,4 +850,14 @@ sub admin_stats {
     return 1;
 }
 
+sub problem_confirm_email_extras {
+    my ($self, $c, $report) = @_;
```
dracos: I think C is $self->{c} here, so you shouldn't need to pass it in - do check (and by that, I mean write a test to show the email is being correctly written in the two cases :) ).

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -850,4 +850,14 @@ sub admin_stats {
...
+    my $user = $report->user or return;
```
dracos: `report->user` will always be set, I think.

dracos:
Looks fine, would be nice to have a test :)

PR #1056, opened 2015-03-09
[Zurich] all admin users can see stats (issues/commercial/673)

osfameron:
Admin CR4 https://github.com/mysociety/FixMyStreet-Commercial/issues/673

seems like admin_pages should be able to generate the menu, but
this change also needs template change.


dracos:
:+1:

PR #1058, opened 2015-03-09
[Zurich] use new larger tiles (issues/commercial/668)

osfameron:
- Refactor map tile size into a constant
 - Update map tiles to use new Zurich WMTS server
 - Restore correct top-left corner values for Zurich map


dracos:
Commit message - no need for "Restore correct top-left corner" given that was presumably squashed out :)

PR #1063, opened 2015-03-10
Stats19 checkbox on Collideoscope /around page

zarino:
This is the markup and styling for a checkbox that lets the user show or hide Stats19 data in the Collideoscope cobrand.

screen shot 2015-03-10 at 16 21 22

@osfameron needs to add the backend logic that makes the checkbox work.

@BenJam needs to decide, with our partners, whether "Stats19 reports" is the right wording to use. Maybe "reports from the Department for Transport" might be clearer?

screen shot 2015-03-10 at 16 22 19

### DIFF
--- a/perllib/FixMyStreet.pm
+++ b/perllib/FixMyStreet.pm
@@ -170,7 +170,8 @@ sub dbic_connect_info {
         # adding this should makeabunch of
         # $c->model('DB')->schema->storage->sql_maker->quote_char() 
         # etc. wrangling obsolete?
-        quote_names => 1, 
+        # BUT causes hilarious consequences, needs major test/refactor
+        ## quote_names => 1, 
     };
 
     return [ $dsn, $user, $password, $dbi_args, $dbic_args ];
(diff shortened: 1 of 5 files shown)


osfameron:
This requires:

 - [ ] the STATS19_IMPORT_USER clause added in #1062 probable needs to be added to `$cobrand->problems` (instead of recent_new)
 - [ ] Javascript on frontend to filter those reports out

osfameron:
Having discussed with @davea, doing this on the frontend seems quite tricky, and may be better to add filtering in back end.

BUT, as above brief analysis shows, this is a fair amount of work.

For the purpose of demoing the feature, the "Show Old" button already does something very similar to what we want e.g. shows all pins, and as the oldest pins will (for the moment) be Stats19 data, they are functionally equivalent. 

Am plumbing the checkbox into that functionality for now.  With a little more development time, we can make this do something very similar to the all_pins functionality, but which invokes the appropriate cobrand-specific filter criteria instead.

PR #1073, opened 2015-03-19
Oxfordshire: Filter reports on map by status & category

davea:
This PR:


davea:
The comments on #1045 apply here too.

davea:
Closing in favour of #1087, as the base branch for this PR has been merged into `oxfordshire-cobrand-updates`.

Issue #805, opened 2014-06-09
Move "name >= 5 characters" error check to cobrand


dracos:
It is not applicable in some places and so should be made optional

PR #1075, opened 2015-03-24
FixMyStreet 1.5.4

jxtsai:
哈囉 你好
很高興看到你在進行FixMyStreet 中文化工作， 我也正在Transifex 進行該專案 1.5.4版的中譯
（抱歉我還未熟悉github的操作）


dracos:
Hi, Thanks very much for your contribution :-) If I've understood your changes correctly, I think there are a number of things we will want to do in order to get your changes merged properly:
* The "name length"/space check in `Problem.pm` is not wanted by everyone – we already noticed this and raised an issue at #805, so this needs to be dealt with. I see this is why you have changed some templates/JS as well, which would not be needed if we fixed this properly;
* You only need to copy emails into `templates/email/tainan` if you want to translate them;
* You don't want to commit your conf/general.yml file, that's just a file for you to have locally in your installation, and it should remain listed in .gitignore.

I will get your translation included in master as soon as possible, and hopefully we can fix #805 properly so that you will not have to make the override changes you have made, but just set a configuration variable or something similar.

dracos:
The name length/space check has been moved to UK-only cobrands in #1172, which should fix your issues needed to make some manual changes. I hope to release 1.5.6 soon.
That was the only actual issue I see arising out of this, so I'm going to close it, please do let us know if we can help in any other way.

PR #1080, opened 2015-03-27
Improve display of council-specific information on individual report page

zarino:
Mostly for Oxfordshire cobrand, but I've included a minor improvement for everyone else.

/cc @davea


zarino:
I've checked this with the new full-screen layout merged in and – bar a few issues that I've fixed in https://github.com/mysociety/fixmystreet/commit/0f874fdf922da91bdca8306692de9f35a33f2728 – it all looks good!

PR #1086, opened 2015-04-07
Add "Report" button to mobile header, and use standard hamburger menu icon

### LINKED ISSUE #931: fixmystreet.com mobile navigation hides reporting link
As raised by @abimysoc in #922:

> I still think this is a site-wide issue with the mobile design, in that "report a problem" is often too well hidden.

And she's right. 

before

If you land on a page other than the homepage, it's hard to work out where to report your issue. This is partly because of the non-standard menu icon (a downwards pointing arrow – which only makes sense once you've clicked it and seen that it's pointing down to the menu in the footer), and also partly because, even with a better menu icon, hiding the main function of the site ("Report a problem") down in a menu in the footer seems counterintuitive.

Maybe something like this might be more obvious?

after

**Also:** If we do this, do we do it for the fixmystreet.com cobrand only, or for all sites?

zarino:
screenshot

Tested against all current cobrands, except Smidsy which hasn't yet been merged into master. As it stands, Smidsy will inherit the hamburger icon and the "Report" button by default, since it doesn't override `header_logo.html`.

The aim was to provide a top-level call to action, so visitors arriving on pages other than the homepage can report a problem in a single click.

Fixes #931.


dracos:
Great :)

PR #1087, opened 2015-04-08
[Oxfordshire] Filter reports on map by status & category

davea:
This PR:


osfameron:
Doesn't look like there are any tests?  Appreciate this is tricky to write with Javascript interaction, but presumably you can test the ajax side of it.  Let me know if you want any help or pointers with writing tests (tbh I find writing functional tests for FMS quite thorny sometimes and the existing examples aren't always easy to repurpose.)

osfameron:
Looks good in general!  There's a fair bit of repetition, but I realise that's largely falling out of how the codebase is structured.

Could do with squashing down (possibly quite ruthlessly, though keeping generic functionality and [Oxfordshire] commits separate).

Apart from that, I think looks good except lack of tests.

davea:
@dracos Could you give this a look when you've got a chance please?

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -221,6 +238,52 @@ sub check_location_is_acceptable : Private {
...
+
+
+    my $category = $c->req->param('category');
+    if ( $category ) {
+        my $count = $c->model('DB::Contact')->not_deleted->search(
+            {
+                body_id => [ keys %bodies ],
+                category => $category
+            }
+        )->count;
```
dracos: As you've just looked up all the categories above, could this be done looking at `@categories` rather than another DB call?

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -165,10 +165,27 @@ sub display_location : Private {
     $c->stash->{all_pins} = $all_pins;
     my $interval = $all_pins ? undef : $c->cobrand->on_map_default_max_pin_age;
 
+    my $states;
+    $c->stash->{filter_status} = $c->cobrand->on_map_default_status;
```
dracos: I think this default setting would be clearer as an `else` in the below `if` section.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -28,17 +28,38 @@ sub my : Path : Args(0) {
...
+    if ( $status eq 'all' ) {
+        $states = FixMyStreet::DB::Result::Problem->visible_states();
+        $c->stash->{filter_status} = 'all';
+    } elsif ( $status eq 'open' ) {
+        $states = FixMyStreet::DB::Result::Problem->open_states();
+        $c->stash->{filter_status} = 'open';
+    } elsif ( $status eq 'fixed' ) {
+        $states = FixMyStreet::DB::Result::Problem->fixed_states();
+        $c->stash->{filter_status} = 'fixed';
+    }
```
dracos: There look to be 4 or 5 copies of basically this code, could this be refactored into one function?

web/js/map-OpenLayers.js
```diff
@@ -29,6 +29,13 @@ function fixmystreet_update_pin(lonlat) {
             if ( lb.length === 0 ) { lb = $('#form_name').prev(); }
             lb.before(data.extra_name_info);
         }
+        // If the category filter appears on the map and the user has selected
+        // something from it, then pre-fill the category field in the report,
+        // if it's a value already present in the drop-down.
+        var category = $("#categories").val();
```
dracos: Idly wondering if this should be filter_categories or something to prevent possible confusion over a different category drop down in future.

web/js/map-OpenLayers.js
```diff
@@ -397,6 +420,16 @@ $(function(){
...
 
+    // Hide the pin filter submit button. Not needed because we'll use JS
+    // to refresh the map when the filter inputs are changed.
+    $(".report-list-filters [type=submit]").hide();
+
+    if (fixmystreet.page == "my" || fixmystreet.page == "reports") {
+        $(".report-list-filters select").change(function() {
+            $(this).closest("form").submit();
+        });
+    }
```
dracos: Nice :)

web/js/map-OpenLayers.js
```diff
@@ -575,6 +631,12 @@ OpenLayers.Control.Click = OpenLayers.Class(OpenLayers.Control, {
             document.getElementById('side-form').style.display = 'block';
         }
         $('#side').hide();
+        // Although it's now hidden, the category filter for the map is still
+        // posted to the server when the report is sent. The name clashes with
+        // the category select element in the report form, which can cause
+        // issues with the wrong/no category being used for the report.
+        // Work around this by renaming the field when it's not shown.
+        $("select#categories").attr("name", "category_filter");
```
dracos: Would it be nicer to rename the filter categories option permanently to avoid this clash?

dracos:
Some minor things, looks good :)

PR #1096, opened 2015-04-21
Simplify .update-img layout

### LINKED ISSUE #424: Update images sometimes shrunk
e.g. first update image on http://www.fixmystreet.com/report/325906 - looks to be the max-width coming into play in some way.

zarino:
Fixes #424.

Primary improvement is that images no longer shrink into oblivion when they're next to a very wordy update:

screen shot 2015-04-21 at 12 39 42

Using a float rather than table-cell styling also means we no longer have to work around IE7 table display support.

Works for short updates with images too:

screen shot 2015-04-21 at 12 46 00

### DIFF
--- a/templates/web/base/report/update.html
+++ b/templates/web/base/report/update.html
@@ -6,10 +6,11 @@
         [% INCLUDE meta_line %]
     </em></p></div>
 [% IF NOT update.whenanswered %]
+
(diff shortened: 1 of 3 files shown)


dracos:
Yay, good to have this fixed at last (and one more step along the improving the CSS path :) ).

PR #1105, opened 2015-05-08
[FMS.com] Unresponsive councils page and report form warning

### LINKED ISSUE #968: Changes to help with councils who are refusing to accept messages
1 - Write blog post on how councils are trapped by vendors, and what they have to do to cope

1.5 - Where we've had to delete all emails for a council, improve the user experience of being told the bad news.

2 - Write delicate emails directly to the council-staffer responsible. Ask for calls.

3 - Change the FMS email footer asking them to send us bounces.

4 - Send some personal mails to actual local users, encouraging them to write to the council sending this blog post and a personal message

5 - Rejudge

6 - Either write custom form posting scripts, or provide links

---

Example postcode covered by *no* FMS bodies: https://www.fixmystreet.com/around?pc=M50+2EQ

Example postcode where some categories (eg: abandoned vehicles, dog fouling) are covered by a body not accepting FMS reports: https://www.fixmystreet.com/around?pc=DE6+2AD

zarino:
Fixes #968.

**Things a developer now needs to do:**

- [x] Store which councils to show this for somewhere
- [x] Show the correct council name in the report form warning
- [x] Only show the warning if the council is unresponsive in the specific category selected
- [x] Show council name on unresponsive council page
- [x] Correct reporting/WriteToThem/Twitter links on unresponsive council page

**Screenshots:**

(Note that "your council" / "this council" is shown in place of the council name, until a developer can plug the right name in!)

screen shot 2015-05-08 at 12 26 03

screen shot 2015-05-08 at 12 25 08

screen shot 2015-05-08 at 12 24 52


wrightmartin:
@zarino @myf: can we fact check the '2%' figure?

Gemmamysoc:
So the '2%' figure is right. Actual is 2.3%. I'm just adding the links to the councils' own reporting systems to this document: https://docs.google.com/spreadsheets/d/1JvIHbg39L3w5_QIpU8nJjp_TxDctgCMVTT9tIEZ5AF4/edit#gid=0

dracos:
Changed it so that this message is loaded via the same mechanism that loads extra questions for a particular category (category_extras), which seemed a good fit. Added links, fixed the link colour, and another bug or two. There's no admin way to add to the list for which this will be shown, easiest will be command line script. But this  now is hopefully done from a code point of view; it works by looking at a boolean in the contact extra metadata unresponsive field.

dracos:
After testing, following things still to be done:
- [ ] Visited link colour
- [ ] Adding a Refused send method for those for which we want to show the message at once, rather than only for particular categories
- [ ] Switch to using "REFUSED" email address to signify a refused contact, rather than dev-only metadata setting

davea:
Quick discussion with @dracos to highlight that there's nothing in the report-sending code to avoid trying to send reports to `REFUSED` email addresses.

wrightmartin:
@zarino - I'm a bit worried the buttons on this screen don't look like buttons!

screen shot 2015-05-08 at 12 25 08

zarino:
@wrightmartin Hmm good point. We could add the arrows back in on small screens, although we'll either need to duplicate the image files at a smaller size, or maybe use `background-size`, if the positioning works out.

perllib/FixMyStreet/App/Controller/Static.pm
```diff
@@ -61,6 +61,20 @@ sub council : Global : Args(0) {
...
+    my $body = $c->stash->{body} = $c->model('DB::Body')->find({ id => $c->req->param('body') })
+        or $c->detach( '/page_error_404_not_found' );
+
+    my @contacts = $c->model('DB::Contact')->not_deleted->search( { body_id => $body->id } )->all;
+    my $any_unresponsive = 0;
+    foreach my $contact (@contacts) {
+        $any_unresponsive = 1 if $contact->email =~ /^REFUSED$/i;
+    }
+
+    $c->detach( '/page_error_404_not_found' ) unless $any_unresponsive;
```
davea: This seems to return 404 if the body's send method is `Refused` and no individual contacts are marked unresponsive.

t/app/controller/report_new.t
```diff
@@ -1462,6 +1462,78 @@ subtest "categories from deleted bodies shouldn't be visible for new reports" =>
...
+
+        my $report = $user->problems->first;
+        ok $report, "Found the report";
+        is $report->bodies_str, undef, "Report not going anywhere";
+
+        $contact3->update( { email => $old_email } );
+        $mech->delete_user($user);
+    };
+};
+
```
davea: Is it worth having a test for the `/unresponsive` page too?

templates/web/fixmystreet.com/static/unresponsive.html
```diff
@@ -0,0 +1,37 @@
+[% INCLUDE header.html
+    title = 'Unresponsive councils: What you can do', bodyclass = 'unresponsive-council fullwidthpage'
+%]
+
+<div class="unresponsive-council__header">
+  <h1>Bad news :(</h1>
+  <p><strong class="unresponsive-council-name">[% body.name %]</strong> no
+  longer accepts reports from FixMyStreet.</p>
```
davea: I think this is potentially misleading - the council might only refuse reports in certain categories but the wording makes it seem like they refuse *all* reports.

davea:
Changing a contact's email address to `REFUSED` in the admin is rejected with 'Please enter a valid email', guess that needs to be special-cased.

Apart from the other little bits, looks good!

t/app/controller/report_new.t
dracos: Yes, bad me, that would have spotted the thing you raise about it not actually working in one case! Although it doesn't look like it form the commit history, I was good and added new failing tests before fixing the code :)

davea:
Maybe some squashing? Otherwise :+1:

MyfanwyNixon:
> We’re sad about it too, but here’s some things you can do now:

Ultra-picky but instead of "here's some things you can do" can we have "here are some things you can do"?

PR #1110, opened 2015-05-21
Fix “All Reports” table headers to top of window on scroll

### LINKED ISSUE #50: Repeat headings on All Reports page or make them static/sticky/frozen

zarino:
Fixes #50.

Works in browsers all the way back to IE8. Degrades relatively gracefully in IE6–7 (position fixed doesn't work, so the table header row is just duplicated in place).

On all browsers, there’s a slight jump in cell widths when the floated header kicks in—possibly something to do with the percentage widths on the original table cells?—but I think we can live with it.

### DIFF
--- a/templates/web/base/reports/index.html
+++ b/templates/web/base/reports/index.html
@@ -44,4 +44,11 @@
 </tbody>
 </table>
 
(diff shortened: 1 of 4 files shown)


dracos:
Fixed the jump by using width instead of outerWidth.

PR #1125, opened 2015-06-16
[Oxfordshire] Cobrand improvements

davea:
This PR is the result of several smaller PRs that have all been reviewed, so everything here should be good to go.

Oxfordshire-cobrand-only changes:
 - Adds filtering by category & status to map pages.
 - Full-screen map layout
 - Order reports by confirmation date
 - Colour pins according to status
 - Include 'expected response time' when viewing report

Changes to all cobrands:
 - Map pin sizes vary depending on zoom level


templates/web/oxfordshire/reports/_problem-list.html
```diff
@@ -0,0 +1,19 @@
...
+
+[% BLOCK column %]
+    <ul class="issue-list-a">
+        [% IF problems %]
+            [% FOREACH problem IN problems %]
+                [% INCLUDE 'reports/_list-entry.html' %]
+            [% END %]
+        [% ELSE %]
+            <li class="empty">
+                <p>[% loc('There are no reports to show.') %]</p>
```
dracos: Happy for this to be in the main templates, I think, good to show a message when no reports :)

templates/web/oxfordshire/report/_main.html
```diff
@@ -0,0 +1,98 @@
+[% moderating = c.user && c.user.has_permission_to('moderate', problem.bodies_str) %]
```
dracos: Probably don't want to override all of this template if we can help it.

dracos:
Couple of template things, otherwise looks fine - some squashing perhaps?

stevenday:
Since Oxfordshire aren't ready to release these changes yet, I've split this into #1134 and #1135, to divide the map filtering feature from all of the Oxfordshire specific changes. The latter could probably be simplified quite a bit now, but I'll leave that as we'd like to release other branches based on this (Greenwich, Bromley) asap.

PR #1059, opened 2015-03-09
[Zurich] improve report print styles (issues/commercial/677)

osfameron:
- limit #site-logo changes to print media only

see #1036 for previous discussion (I couldn't see way of repointing the PR to zurich-2015-dev-candidate branch, so closed and opened this one)

PR #1057, opened 2015-03-09
[Zurich] text tweak (hint for "Photos" label) (issues/commercial/666)

osfameron:
https://github.com/mysociety/FixMyStreet-Commercial/issues/666

PR #1127, opened 2015-06-17
Zurich 2015 Updates

davea:
All commits here should have been reviewed already via smaller PRs:

 - #1054 
 - #1055 
 - #1056 
 - #1057 
 - #1058 
 - #1059
 - #1129
 - #1130
 - #1131 
 - #1144 
 - #1145 
 - #1216 
 - #1218
 - #1226 
 - #1227


davea:
This is ready for merging, I believe. Two things to consider when deploying:

 - Some of the DB migrations have been renamed since being deployed on the live ZWN site. We'll probably need to manually apply the new migrations from master when deploying next.
 - The .po files need updating and a couple of new strings (see #1213) translated by Zürich.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1370,36 +1478,27 @@ Rotate a photo 90 degrees left or right
 
 =cut
 
+# returns index of photo to rotate, if any
+sub _get_rotate_photo_param {
+    my ($self, $c) = @_;
+    my $key = first { /^rotate_photo/ } $c->req->parameters or return;
```
dracos: `$c->req->parameters` is a hashref, so I don't think this can work as is.

perllib/FixMyStreet/App.pm
```diff
@@ -318,15 +319,15 @@ sub send_email {
...
     my $content = $c->view('Email')->render( $c, $template, $vars );
 
     # create an email - will parse headers out of content
     my $email = Email::Simple->new($content);
-    $email->header_set( ucfirst($_), $vars->{$_} )
-      for grep { $vars->{$_} } qw( to from subject Reply-To);
-
-    return if $c->is_abuser( $email->header('To') );
+    $email->header_set( 'Subject', $vars->{subject} ) if $vars->{subject};
+    $email->header_set( 'Reply-To', $vars->{'Reply-To'} ) if $vars->{'Reply-To'};
```
dracos: I think this should be okay, as Reply-To only ever contains one item, and Email::Simple will take the first given an arrayref. For consistency, might want to move down to From/To conditionally, but shrug.

dracos:
Yay! :balloon: The main/only thing I think that needs looking out for any issues post merge on the main site is probably multiple photo support? A final go through the commit messages probably wise, as discussed :)

PR #1129, opened 2015-06-17
[Zurich] (671) Response templates for report admin

dracos:
For mysociety/FixMyStreet-Commercial#671.

### DIFF
--- a/db/schema_0033-response-templates.sql
+++ b/db/schema_0033-response-templates.sql
@@ -1,0 +1,8 @@
+create table response_templates (
+    id serial not null primary key,
+    body_id int references body(id) not null,
+    title text not null,
+    text text not null,
+    created timestamp not null default ms_current_timestamp(),
+    unique(body_id, title)
+);
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -824,6 +824,85 @@ sub report_edit : Path('report_edit') : Args(1) {
     return 1;
 }
 
+sub templates : Path('templates') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    $c->detach( '/page_error_404_not_found' )
+        unless $c->cobrand->moniker eq 'zurich';
+
+    my $user = $c->user;
+
+    $self->templates_for_body($c, $user->from_body );
+}
+
+sub templates_view : Path('templates') : Args(1) {
+    my ($self, $c, $body_id) = @_;
+
+    $c->detach( '/page_error_404_not_found' )
+        unless $c->cobrand->moniker eq 'zurich';
+
+    # e.g. for admin
+
+    my $body = $c->model('DB::Body')->find($body_id)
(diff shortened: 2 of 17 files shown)


db/schema_0033-response-templates.sql
```diff
@@ -0,0 +1,8 @@
+create table response_templates (
+    id serial not null primary key,
+    body_id int references body(id) not null,
+    title text not null,
+    text text not null,
+    created timestamp not null default ms_current_timestamp(),
```
dracos: This doesn't match the schema.

davea: I've added a fixup commit to remove `whenedited` from schema

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -824,6 +824,82 @@ sub report_edit : Path('report_edit') : Args(1) {
...
+    if ($c->req->method eq 'POST') {
+        if ($c->req->param('delete_template') eq _("Delete template")) {
+            $template->delete;
+        } else {
+            $template->title( $c->req->param('title') );
+            $template->text ( $c->req->param('text') );
+            $template->update_or_insert;
+
+            # log_edit currently has hard-coded types, do we need auditing on this?
+            # $c->forward( 'log_edit', [ $object_id, $object_type, $action_performed ] );
```
davea: This should either be turned into a `log_edit` call or removed

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -660,6 +660,48 @@ sub body {
...
+Similar to C<body> but returns a body object, either from C<external_body>
+or from C<bodies_str>.
+
+NB: For cobrands where C<bodies_str> is a comma-separated list, this may
+return the first retrieved body at random.  (This behaviour is currently
+intended for Zurich, which uses a single body.)
+
+=cut
+
+sub body_object {
```
davea: This doesn't ever seem to be used anywhere, and should probably have been removed when I updated the template to use `problem->response_templates`.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -824,6 +824,82 @@ sub report_edit : Path('report_edit') : Args(1) {
...
+
+    # e.g. for admin
+
+    my $body = $c->model('DB::Body')->find($body_id)
+        or $c->detach( '/page_error_404_not_found' );
+
+    $self->templates_for_body($c, $body);
+}
+
+sub template_edit : Path('templates') : Args(2) {
```
davea: Might be worth putting the `unless $c->cobrand->moniker eq 'zurich'` check on this and `templates_view` too, given that the HTML templates don't exist anywhere except the Zürich cobrand.

davea:
A few fixup commits to address some little bits, but I think this is looking OK.

davea:
Merged in 419b78c2c04fda6d886759cba3f10a2fd8f037f3.

PR #1130, opened 2015-06-17
[Zurich] (675) Attach images to emails

dracos:


perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -670,6 +670,32 @@ sub _admin_send_email {
...
+        my $id = $row->id;
+        my @attachments = map {
+            my $i = $index++;
+            {
+                body => $_->[1],
+                attributes => {
+                    filename => "$id.$i.jpeg",
+                    content_type => 'image/jpeg',
+                    encoding => 'base64',
+                        # quoted-printable ends up with newlines corrupting binary data
```
davea: this probably doesn't need explaining any more

t/app/helpers/send_email.t
```diff
@@ -33,16 +34,74 @@ my @emails = Email::Send::Test->emails;
...
+
+    ok $c->send_email( 'test.txt',
+        { to => 'test@recipient.com',
+          attachments => [
+             {
+                body => $data,
+                attributes => {
+                    filename => 'foo.gif',
+                    content_type => 'image/gif',
+                    encoding => 'quoted-printable',
```
davea: I wonder if it's worth changing this to `base64` so it matches what `munge_sendreport_params` actually uses.

davea:
I've made two fixup commits that refactor a test helper function to reduce duplicate code, and I think this is looking OK now.

PR #1131, opened 2015-06-17
[Zurich] (676) Time spent recording

dracos:
After 671/675.


davea:
A little fixup to remove some duplicated code, otherwise :+1:

PR #1132, opened 2015-06-18
[Greenwich] Cobranded front end

davea:
This PR adds a cobranded front end for Royal Borough of Greenwich.

It uses the map filtering work done in Oxfordshire's updates, so that's the base branch.


perllib/FixMyStreet/SendReport/Open311.pm
```diff
@@ -127,7 +127,7 @@ sub send {
             $revert = 1;
         }
 
-        if ($row->cobrand eq 'fixmybarangay') {
+        if ($row->cobrand eq 'fixmybarangay' || $row->cobrand eq 'greenwich') {
             # FixMyBarangay endpoints expect external_id as an attribute, as do Oxfordshire
```
stevenday: Presumably this comment should say Greenwhich, not Oxfordshire? (Or Oxfordshire should be in the if statement too?)

davea: That's a good point. It seems that Oxfordshire do require `external_id`, but it's handled elsewhere so I don't know why this comment mentions Oxfordshire. I'll update it to say Greenwich instead.

stevenday:
I've replaced this with #1136

PR #1135, opened 2015-06-29
[Oxfordshire] Cobrand improvements

stevenday:
Originally #1125, but I've rebased things so that oxfordshire-independent
changes like the filtering on map pages are in a separate branch so that this
doesn't hold up other work that relies on that feature.

Original PR description from @davea copied below:

This PR is the result of several smaller PRs that have all been reviewed, so everything here should be good to go.

Oxfordshire-cobrand-only changes:

Adds filtering by category & status to map pages.
Full-screen map layout
Order reports by confirmation date
Colour pins according to status
Include 'expected response time' when viewing report
Changes to all cobrands:

Map pin sizes vary depending on zoom level


davea:
Closing in favour of #1152.

PR #1136, opened 2015-06-29
[Greenwich] Cobranded front end

stevenday:
Originally from #1132, but now rebased onto the 'report-filtering-on-map'
branch so that it's separate from any changes to the Oxfordshire cobrand.


davea:
Merged in 9b7e4fdebeecbed6f600577177a86e3c216368c0.

PR #1111, opened 2015-05-26
Redo the Bromley Cobrand

stevenday:
Replaces the existing Bromley council cobrand with a new one that uses the
latest FMS styling. Including the new report list/filter sidebar.

Closes #721

PR #1137, opened 2015-06-29
Redo the Bromley Cobrand

### LINKED ISSUE #721: Add ability for cobrands to alter pin icons
Currently this is pretty hardcoded in the code/JS. Some way for a cobrand to provide a different image path would be good,

stevenday:
Originally from #1111 but rebased so that it sits on top of
"reporting-filtering-on-map" to be able to get the filtering updates without
relying on things from Greenwich or Oxfordshire.

Replaces the existing Bromley council cobrand with a new one that uses the
latest FMS styling. Including the new report list/filter sidebar.

Closes #721


stevenday:
This is replaced with https://github.com/mysociety/fixmystreet/pull/1155 which is based on master

PR #1139, opened 2015-07-07
Update DBIx::Class::EncodedColumn.

dracos:
This fixes a potential Unicode issue with the setting of encoded columns.


stevenday:
This needed a downgrade of MakeMaker?

dracos:
The newer version of EncodedColumn only depends upon an older version of MakeMaker (or I ran it with an older version and that just gets copied in, I've never really worked it out); the version of MakeMaker in the snapshot is the same (6.72).

PR #1143, opened 2015-07-10
Correcting some sentence structure / wordings on site

lowks:


dracos:
Hi, thanks for your contribution. Both the strings in your changes are correct English, as far as I am aware. Comparing "alert me to" and "alert me of" searches on Google, I get c. 5.5 million for the former, only 130k for the latter. The other string is an example of the present simple (similar to "The restaurant opens at 7pm tomorrow."); http://learnenglish.britishcouncil.org/en/english-grammar/verbs/talking-about-future has lots more information, here it's used on the front page as a list of actions that occur as to the steps of reporting a problem.

If they are not correct in your particular dialect/translation, then feel free to translate them appropriately for your language, of course. Does that make sense?

Also, if you do want to make changes to strings in the codebase in future, you need to make them to the actual code where the strings are used, not the translation files, as otherwise what will happen is the strings will appear in English for all languages (as they can't find a translation any more). Hope that's helpful.

PR #1144, opened 2015-07-14
[Zurich] Prevent invalid bodies_str (zurich-a)

davea:


davea:
Split the invalid bodies_str commit in two as it had some unrelated stuff in there, but otherwise looks good.

PR #1145, opened 2015-07-15
[Zurich] First part of 672/690 (new statuses and admin workflow changes)

davea:
This is the first batch of work for mysociety/FixMyStreet-Commercial#672 and mysociety/FixMyStreet-Commercial#672.


davea:
Lots of uses of `$c->req->params` that will need to be replaced with `$c->get_param` when rebased on master.

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -682,14 +762,94 @@ sub admin_report_edit {
...
+        push @states, {
+            state => 'fixed - council',
+            trans => _('Closed'),
+        };
+    }
+    elsif ($state =~/^(hidden|unconfirmed)$/) {
+        @states = grep { $_->{$state} } @states;
+    }
+    $c->stash->{states} = \@states;
+    $c->stash->{states_debug} = Dumper($state, \@states); use Data::Dumper;
```
davea: This isn't used anywhere so should be removed.

templates/web/zurich/admin/report_edit.html
```diff
@@ -120,106 +105,131 @@
...
 
-<li class="assignation">
-<label for="body_external">[% loc('Assign to external body:') %]</label>
-    <select name="body_external" id="body_external">
-        <option value="">--</option>
-    [% FOR body IN bodies %]
-        [% NEXT IF body.parent OR body.bodies OR body.deleted %]
-        <option value="[% body.id %]"[% IF body.id == problem.bodies_str %] selected[% END %]>[% body.name %]</option>
-    [% END %]
+[% IF problem.state == 'confirmed' %]
```
davea: This `IF` condition is a duplicate of the containing `IF`, so these `IF`/`END` tags can be removed.

templates/web/zurich/admin/report_edit.html
```diff
@@ -120,106 +105,131 @@
...
-      [%- ELSE -%]
-
-Freundliche Grüsse
+[% END %]
 
-Ihre Stadt Zürich
-    [%- END %]</textarea>
+[%# 3rd party messages sent for Extern/Wunsch states %]
+[% FOR s IN ['closed','investigating'] %]
+    [% IF problem.state == s %]
```
davea: `problem.state` can't be both `closed` and `investigating`, and the `s` value doesn't seem to be used within this `FOR` block, so this can probably be simplified to `IF problem.state == 'closed' OR problem.state == 'investigating'`.

templates/web/zurich/admin/report_edit.html
```diff
@@ -120,106 +105,131 @@
...
     </li>
-</ul>
+    </ul>
+    [% END %]
+[% END %]
 
-[% ELSIF problem.state == 'fixed - council' %]
+[%# Public response field shown for Ruckmeldung ausstehend/Extern/
+  # Zustandigkeit unbekannt/Wunsch/Nicht kontaktierbar/Unsichtbar %]
+[% FOR s IN ['planned','closed','unable to fix','investigating','partial','hidden'] %]
```
davea: Similarly to above, but with the increased number of possible states this would be better as a `SWITCH`.

PR #1150, opened 2015-07-21
Remove no longer used Essex/Oxfordshire handling.

### LINKED ISSUE #1149: Remove Essex County Council special code
Essex County Council problem reports are routed at the code level to special-case bodies (for east/mid/south/west), rather than by entering email addresses for the main council body in the admin interface. 

However, it looks like Essex have changed their system. The district-specific contacts are now bouncing - reports aren't being delivered. Default behaviour needs to be restored for this council.

dracos:
Fixes #1149


dracos:
Merged in 7f33b7f

PR #1151, opened 2015-07-21
[fixmystreet.com] Add deleted NI councils.

dracos:
Factor out template to correct cobrand.


dracos:
Merged in dee48ff (reordered IF as Carrick subset of Carrickfergus, doh).

PR #1153, opened 2015-07-23
Add facebook button to report pages

zarino:
screen shot 2015-07-23 at 15 19 16

For https://github.com/mysociety/fixmystreet-international/issues/9


davea:
Looks good!

I broke out the button HTML into its own template and committed it as mysociety/fixmystreet-international@2bebd69b10bbf6391c19fe443a13bbdd842ced7b.

PR #1155, opened 2015-07-24
[Bromley] Redo the Bromley Cobrand

stevenday:
Replaces the existing Bromley council cobrand with a new one that uses the
latest FMS styling. Including the new report list/filter sidebar.

Closes https://github.com/mysociety/FixMyStreet-Commercial/issues/721


stevenday:
@dracos - having got the client's approval for this, is there any chance you could take a look at the code at some point to make sure it's all ok?

templates/web/bromley/footer.html
```diff
@@ -1,49 +1,44 @@
+                    [% IF pagefooter %]
```
dracos: This footer ends up identical to the fixmystreet one - I don't think it needs adding in ba8ffd8

templates/web/bromley/header_logo.html
```diff
@@ -0,0 +1,62 @@
...
+            <input value="Search" type="submit" class="button__primary">
+        </form>
+    </div>
+</div>
+<div class="a-z-container ">
+    <nav class="main-a-z custom-side-nav">
+        <ul class="drop-down-nav">
+            <li class="first-child">
+                <a href="#" class="main-nav-title">A-Z</a>
+                <ul class="nav-hover-item nav-hover-item-active a-z" style="display: none;">
```
dracos: This display:none should be added by the JS (or ideally, CSS using a .js class or similar for instant effect), as currently I don't think this works without JavaScript (though the comment in the a-z JS implies it should).

web/cobrands/bromley/layout.scss
```diff
@@ -1,135 +1,270 @@
...
+body.fullwidthpage .container .content footer .tablewrapper {
+  background-color: $bromley_blue;
 }
-.ie6, .ie7, .ie8  {
-    .content {
-        margin: 0;
-        border: none;
+
+// Make the powered by FMS logo work in Old IE. FMS fixes this in IE8, but
+// it's broken in IE6 and 7 too.
```
dracos: Is this worth putting in core FMS if it's straightforward to do?

web/cobrands/bromley/layout.scss
```diff
@@ -1,135 +1,270 @@
...
+  // can't do that because they're in separate divs, so add a margin here
+  // instead.
+  margin-bottom: 20px;
 }
 
-// Perhaps fix map location (should be in central?)
-.ie6 #map_box, .ie7 #map_box {
-    right: -480px;
+// The reports_list.scss hides the big banner on the sidebar that tells people
+// what to do with the map (click on it). We don't want that, so put it back
```
dracos: It looks like the report_list assumes a different `click-the-map` class is used (see e.g. Oxfordshire). This should probably be looked at and fixed/consolidated to avoid future confusion.

templates/web/bromley/reports/_problem-list.html
```diff
@@ -0,0 +1,19 @@
+<section class="full-width">
```
dracos: This template is identical to the parent, and doesn't need to be added, I think.

templates/web/bromley/reports/_list-filters.html
```diff
@@ -0,0 +1,32 @@
+[% IF use_section_wrapper %]
```
dracos: Do we want to have a shared template between all these identical ones using this HTML?

templates/web/bromley/reports/_problem-list.html
dracos: (Ditto for Greenwich looks like)

templates/web/bromley/reports/_list-entry.html
```diff
@@ -0,0 +1,6 @@
+<li class="[% c.cobrand.pin_colour(problem) %]">
```
dracos: I note that this (and similar elsewhere it's used) has much less information than the upstream template – it loses distance away, last updated (which on all reports page can be quite important), state and the report's image. The last updated and image seem like they may be wanted, perhaps?

web/cobrands/bromley/layout.scss
```diff
@@ -1,135 +1,270 @@
...
+// Bromley's search form floats to the right too
+.main-search {
+  width: 35%;
+  margin: 10px 0 0 0;
+  float: right;
+  min-width: 287px;
+  clear: right;
+}
+
+// Bromley's A-Z nav gets the background coloured in on biggger screens to
```
dracos: Typo in comment :)

web/cobrands/bromley/layout.scss
```diff
@@ -1,135 +1,270 @@
...
+  background: #647890;
+  padding: 0;
+}
+
+// The link to open the A-Z nav gets hidden on big screens
+.drop-down-nav > li > a {
+  display: none;
+}
+
+// And the rest of the A-Z nav gets show with a horrible !important
```
dracos: What makes the `!important` necessary here?

web/cobrands/bromley/layout.scss
```diff
@@ -1,135 +1,270 @@
...
-        .content {
-            margin-top: 0;
-        }
-    }
+// Override the FMS main menu to give it a background colour
+#main-menu {
+  background-color: $bromley_blue;
+}
+// Override the selected element to match. We need to do this separately
+// because of FMS's selectors being quite specific
```
dracos: Ticket this? :)

web/cobrands/bromley/base.scss
```diff
@@ -1,60 +1,499 @@
...
+  }
 }
 
-// Colour tab to match colour scheme
+// Change the colour of the menu icon to black because of the changes to
+// #site-header
 #nav-link {
-  height: 4em; // to match #site-header
-  background-image: url(/cobrands/fixmystreet/images/menu-black.png);
+  background: transparent url(/cobrands/fixmystreet/images/menu-black.png) center center no-repeat;
```
dracos: After #1186, this can just be `$menu-image` in _colours.

dracos:
As above, great to see comments for all the styling reasons, should make it easier to make changes in future. If there's anything that you think could be changed in the main CSS to help simplify this, do ticket it (I think I commented on one obvious one, but there will be others, I'm sure). Hope comments all make sense, let me know if not. I think this also means bromley.css and bromley/position_map.js can be removed as no longer used.

templates/web/bromley/header_logo.html
stevenday: It turns out they actually do add it in the js, they were just being mean to non-js users. I've removed it now.

web/cobrands/bromley/layout.scss
stevenday: Sure, I've moved it to sass/_layout.scss, I don't think there's any harm in doing that.

templates/web/bromley/footer.html
stevenday: Have removed it completely.

web/cobrands/bromley/layout.scss
stevenday: I think the issue was that I hadn't added the custom `around/_report_banner.html` that oxfordshire and greenwich use in conjunction with the report filtering. Since that has some html and content differences to FMS, it doesn't seem like I can easily consolidate everything to use that (we'd need to edit the JS that changes the h1 on mobiles too), so instead I've changed bromley to use the same as the other three. This removes the need for this css.

templates/web/bromley/reports/_list-entry.html
stevenday: I've changed this template so that it contains more of the information again - you're right, the new filtering work seems to have dropped a bunch of stuff that there's no real need to. I think this is because it came from collideoscope originally, where a lot of the info isn't available or appropriate.

I looked into fixing this on a bigger scale, but things get a bit confusing between the _list-entry and _item templates, and the different types of list they can be in (.report-list, .report-list-a, .issue-list-a and so on). I'll ticket tidying it up better (perhaps along with including these filters as the default?).

web/cobrands/bromley/layout.scss
stevenday: Fixed, ta!

web/cobrands/bromley/layout.scss
stevenday: I'm not sure why, but it's necessary to make the nav display in IE8 and below - I took this from Bromley's CSS - I guess it's a quirk with the way they're doing the hide/show stuff :(

web/cobrands/bromley/layout.scss
stevenday: Done: https://github.com/mysociety/fixmystreet/issues/1191

stevenday:
@dracos - thanks for the review, I think I've addressed everything as you suggested. There's a couple of new commits to the core or other cobrands that it would be nice if you could double check: 297f201 and b0e6727, otherwise hopefully that's everything!

templates/web/bromley/around/_report_banner.html
```diff
@@ -0,0 +1,10 @@
+<h1 class="big-green-banner">
```
dracos: This file means the bromley check in `templates/web/fixmystreet/around/_report_banner.html` can be removed.

dracos:
`templates/web/greenwich/reports/_problem-list.html` can still be removed.

stevenday:
@dracos ok! Take 2. 

- I've reverted b0e6727 with c103940 (I'll rebase it away after) and added 0a87534 instead.
- I've added e5441088e94c8e45bb1b177f4d7163b3b616e52d to fix the issues with Oxfordshire's my page in 297f201 and make the css in `_report-filters.scss` more targetted. This looks ok across all three cobrands for me now, and I've added a comment as to why it's there for when we try to rationalise it.

dracos:
Great! Only thing I can see now is that the symlinks aren't relative from the directory they're in, so don't actually work, sorry.

PR #1158, opened 2015-07-27
Fix multiple body areas

davea:
This PR:

 - Fixes a bug in `App::get_param_list` that would cause a non-empty list to be returned if the param was missing from the request
 - Fixes a bug meaning only the first of the selected areas would be associated with the body when updating or creating it.


dracos:
Looks good, thanks for spotting this!

PR #1159, opened 2015-07-28
Add a custom FAQ page for Stevenage Borough Council

stevenday:
Most of the content is cribbed from other councils, with the exception of
their Out of Hours number which I found on their website.

Closes https://github.com/mysociety/FixMyStreet-Commercial/issues/661


dracos:
GitHub link should be HTTPS, otherwise looks fine.

PR #1160, opened 2015-07-28
Show admin notes in the Zurich admin

stevenday:
https://github.com/mysociety/FixMyStreet-Commercial/issues/706 notes that no
error messages are being shown in the zurich admin. This is because the
document body doesn't have the `show-admin-notes` class that stops them being
hidden by CSS. I've updated the `zurich/admin/header.html` template to fix
this.

Was this just an oversight, or was it intentional for some reason?

Fixes https://github.com/mysociety/FixMyStreet-Commercial/issues/706

### DIFF
--- a/templates/web/zurich/admin/header.html
+++ b/templates/web/zurich/admin/header.html
@@ -1,5 +1,5 @@
 [%
-    SET bodyclass = bodyclass || 'fullwidthpage';
+    SET bodyclass = bodyclass || 'fullwidthpage show-admin-notes';
     INCLUDE 'header.html' admin = 1, bodyclass = bodyclass _ ' admin';
 
     states = {


davea:
Just an oversight, I think. Nice fix! :+1: 

Please put this commit onto `zurich-2015-dev` too :)

dracos:
I think actually show-admin-notes class should probably be dropped.

dracos:
To expand, it looks like it was introduced in order to make admin notes optional on old installations that might not want them (so only new installs would have it in the header so that they would display). But it looks like all installations now have actual errors (e.g. this one, and the default "No results found" on some pages) using the classes that will be hidden unless the show-admin-notes is present. I think it should probably therefore be removed.

davewhiteland:
@dracos is correct, `show-admin-notes` was a CSS class that was my attempt to add the admin notes to cobrands that wanted them without them appearing on extant cobrands that didn't, without breaking out a new config variable. I think it's out of date now since we're displaying those error messages within the same classes. Poor separation of concerns there; so I approve of dropping it now it's served its (dubious anyway) purpose.

stevenday:
Ta! I've opened a new PR to fix it in Master - once that's merged, I'll get it into Zurich manually.

PR #1162, opened 2015-07-28
Drop .show-admin-notes class and associated code

### LINKED ISSUE #1161: Drop show-admin-notes CSS class and just show admin notes everywhere
From https://github.com/mysociety/fixmystreet/pull/1160 - this is a relic of an older situation that no longer applies.

stevenday:
As per @dracos and @davewhiteland's comments on https://github.com/mysociety/fixmystreet/pull/1160
this removes the show-admin-notes css and makes admin notes and errors display
all the time.

Closes #1161

This will need to be backported to Zurich to close https://github.com/mysociety/FixMyStreet-Commercial/issues/706 too.


dracos:
I don't think it needs the first display:block block as that was only there to counteract the default display:none, now removed; with that bit gone, looks good, thanks!

PR #1165, opened 2015-07-29
[East Sussex, Greenwich] Add contact email addresses

stevenday:
Overrides `contact_email` in the East Sussex and Greenwich cobrands so that emails from the
`/contact` page get sent to the council, rather than us.

~~This is hopefully very simple, I'm just making it a PR so that someone can
tell me if I've misunderstood all the implications of doing this (I have 4
more cobrands to update this on soon).~~

Fixes https://github.com/mysociety/FixMyStreet-Commercial/issues/619, https://github.com/mysociety/FixMyStreet-Commercial/issues/726

### DIFF
--- a/perllib/FixMyStreet/Cobrand/EastSussex.pm
+++ b/perllib/FixMyStreet/Cobrand/EastSussex.pm
@@ -48,16 +48,16 @@ Can run with a script or command line like:
 =cut
 
 use constant POTHOLE_SIZES => [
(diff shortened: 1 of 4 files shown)


dracos:
No, I think that's all there is to it :) Probably easiest to stick all the changes for this in one commit.

stevenday:
Ta, I've added greenwich's too, still waiting on Stevenage and Oxfordshire

PR #1166, opened 2015-07-29
[East Sussex] Fix menu overlapping intro, and highlighted item looking rubbish

stevenday:
Fixes https://github.com/mysociety/FixMyStreet-Commercial/issues/659 and a
related issue I noticed at the same time where the currently selected item in
the menu had different styling and so was smaller and out of line.

### DIFF
--- a/web/cobrands/eastsussex/base.scss
+++ b/web/cobrands/eastsussex/base.scss
@@ -59,7 +59,9 @@ h1.main {
     margin: 0 0 9px;
 }
 
-.mappage .application-nav, .twothirdswidthpage .application-nav {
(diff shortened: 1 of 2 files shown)


dracos:
The menu also overlaps on other pages, e.g. help page, alerts page, I think a more general solution is needed than just in `#front-main`, I'm afraid. However, thanks especially for including the comment in layout explaining what the CSS is doing there (note everyone – do this! All the time!)

stevenday:
@dracos - thanks for spotting that, I've added some extra css to account for those pages too (and removed an irregularity with the menu padding). Hopefully that's it dealt with everywhere.

dracos:
Looks good :)

PR #1167, opened 2015-07-30
[East Sussex] Stop the header scrolling off the screen on the map page

stevenday:
This is a bit of a hacky way to fix it, because the cobrand html/css is a bit
non-standard, but it does the trick I think.

Closes https://github.com/mysociety/FixMyStreet-Commercial/issues/542


dracos:
Does indeed seem to do the trick.

PR #1169, opened 2015-07-30
Fix recent photos on alerts page overruning footer.

### LINKED ISSUE #1168: Recent photos sidebar on alerts page runs over the footer
Moving this from https://github.com/mysociety/FixMyStreet-Commercial/issues/336 because it affects all FMS installations. e.g.

screen shot 2015-07-30 at 16 42 51

Should be as simple as removing the sticky-sidebar class, so PR to follow!

stevenday:
Closes #1168

### DIFF
--- a/templates/web/base/alert/index.html
+++ b/templates/web/base/alert/index.html
@@ -37,7 +37,7 @@ To find out what local alerts we have for you, please enter your [% c.cobrand.co
 </form>
 
 [% IF photos.size %]
-<div class="sticky-sidebar" id="alert_recent">
+<div id="alert_recent">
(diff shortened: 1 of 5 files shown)


dracos:
Yay. I think this could also remove the mentions of `alertindex` throughout (I assume the east sussex one won't apply any more either).

stevenday:
@dracos thanks! I've removed that class, and other instances of the `.sticky-sidebar` being used for the alert page sidebar I found via grep.

dracos:
Woo!

dracos:
Merged in b748677

PR #1170, opened 2015-07-30
Add some admin things

### LINKED ISSUE #517: add indication to admin where contact/categories have 'extra' information
Even if we don't add any capability to edit such information to admin, it's worth flagging up that such data exists (presumably displaying it (or revealing it when clicked)).

### LINKED ISSUE #580: in admin, lookup comment_user_id and show those details on body page
Using the admin to see which user was being associated with the updates coming in from Open311 -- body page shows a user ID but it would be nice if that was a lookup that showed the email address maybe

### LINKED ISSUE #601: Allow marking reports as sent in the admin
Currently you can mark a report as unsent in the admin, but you can't mark it as sent. It would be useful to have a way to do this in the admin to remove the need to go into the database and set the `whensent` column manually.

### LINKED ISSUE #638: add links to mapit pages from the admin (showing body's area)
On report admin pages, as well as showing the body ID (currently, as "For councils:") it would be handy to have a link to the MapIt page for those body IDs. This might also be useful on the admin pages for specific bodies too.

I think this should be pretty straightforward, because the "right" MapIt URL is available as the MAPIT_URL config var, and URL is of the form http://mapit.mysociety.org/area/2244.html
i.e. easy to construct knowing the body ID

Possible gotcha: might need to check if FMS is being picky about generation in MapIt, but I don't think we have a mechanism for that in place anyway…?

Use case for this:

Occasionally an FMS user disputes that the report has been sent to the correct council, and one cause of this is that they are unaware of the actual boundaries. So admin staff sometimes need to double check what's going on, and this usually entails a trip to MapIt… whose area pages (in .html format) are marvellously lovely and clear because they prettily show the boundary on the map.

A similar and *potentially* useful link on the report pages would be a link to MapIt off the coordinates. Currently the admin's report page shows the coordinates FMS is using for that point, which I suspect is only really useful in the context of MapIt anyway; and the URL again can be constructed (which is helpful because e.g., coords need to be the right way round, SRID might matter, etc).

### LINKED ISSUE #920: in admin, show metatadata for Open311 categories that have any
FMS can auto-populate body's contacts/categories with Open311's automatic service discovery, and this may include metatdata (such as "size of pothole?") -- but if metadata has been detected, it's not shown in the admin.

Should add this to the body page where the categories are listed.

The metadata is used, of course, when showing the form to the reporting user:
https://github.com/mysociety/fixmystreet/blob/master/templates/web/fixmystreet/report/new/category_extras.html
so a summary of that information could be included in the admin table.

### LINKED ISSUE #990: show user_body_permissions settings when inspecting users in the admin
Various levels of this, but just to show the 

  * if user has `user_body_permissions` settings, should show them:
    * in the /users summary table (since they are all the council users)
    * in the user_edit  page so you know when looking at the user that she/he is special

dracos:
This PR allows a body's external URL to be edited in the admin (no ticket). While it's there, it adds a mark as sent button (fixes #601); adds a link to a body's comment_user_id (fixes #580); adds mapit links to body/report pages (fixes #638); shows category extra data (fixes #517, fixes #920); lists report bodies more nicely on a report page, with links (no ticket); stars users who have moderate permission (fixes #990), and tweaks the display of body form a bit.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -425,7 +425,7 @@ sub update_contacts : Private {
 sub body_params : Private {
     my ( $self, $c ) = @_;
 
-    my @fields = qw/name endpoint jurisdiction api_key send_method send_comments suppress_alerts send_extended_statuses comment_user_id can_be_devolved parent deleted/;
+    my @fields = qw/name endpoint jurisdiction api_key send_method external_url/;
     my %defaults = map { $_ => '' } @fields;
     %defaults = ( %defaults,
         send_comments => 0,
@@ -436,7 +436,7 @@ sub body_params : Private {
         parent => undef,
         deleted => 0,
     );
-    my %params = map { $_ => $c->get_param($_) || $defaults{$_} } @fields;
+    my %params = map { $_ => $c->get_param($_) || $defaults{$_} } keys %defaults;
     return \%params;
 }
 
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -705,6 +705,13 @@ sub report_edit : Path('report_edit') : Args(1) {
 
         $c->forward( 'log_edit', [ $id, 'problem', 'resend' ] );
     }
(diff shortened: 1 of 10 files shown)


stevenday:
:+1: this all looks good!

PR #985, opened 2015-01-06
Add JavaScript for research survey

jacksonj04:
Add JavaScript for tracking users and performing logic, and page elements to send users to the survey.

The JavaScript is a compressed version of the full code found at https://gist.github.com/jacksonj04/a84f4b0d38d02b6cee6d/3439b2c8601ec08408de0eb0e801bd107d81df0e.

PR #1175, opened 2015-08-04
Remove Quant 2 survey elements.

jacksonj04:
This pull request removes the survey link added in #985.

NB: This leaves the hooks for included template fragments in place, since they will probably come in useful in future.


dracos:
Looks fine.

PR #1178, opened 2015-08-10
Improve stored procedure performance.

dracos:
Add an index on the radian values, and switch to a full bounding box
rather than a slower angle check.
Oh and make it STABLE, that's quite important.


dracos:
Merged in 6329d2b848c405e4e646a4111941e82fcb0f33cb (as is already live).

PR #1181, opened 2015-08-12
bin/make_css now follows symlinks

zarino:
The [fixmystreet-international](https://github.com/mysociety/fixmystreet-international) installation instructions make use of symlinks in, among other places, `web/cobrands/`. This means that `bin/make_css` needs to follow symlinks when looking for scss files to compile.

### DIFF
--- a/bin/make_css
+++ b/bin/make_css
@@ -20,7 +20,7 @@ fi
 
 DIRECTORY=$(cd "$DIR"/../web && pwd)
 
-DIRS=${@:-`find $DIRECTORY -name "*.scss" -exec dirname {} \; | uniq`}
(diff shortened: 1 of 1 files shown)


dracos:
Great, thanks! There's a typo make => makes in the commit body, and my only other suggestion would be perhaps rewording the commit message to better match others, e.g. "Follow symlinks during CSS compilation".

zarino:
@dracos: Commit title amended, as you suggested. Body's the same, since "instructions" is plural, meaning "make" not "makes" :-)

Shall I merge it?

dracos:
Ah yes, whoops, go for it :)

PR #1183, opened 2015-08-13
[Collideoscope] Proper stats19 filtering on the /around page

stevenday:
This PR:

On the FMS side:
- Adds a mechanism for cobrands to supply extra query parameters when getting
  problems on the map page. This requires changes to the /around and /ajax
  controller actions that produce lists of features and pins for the /around
  page, so that they call the cobrand for an extra hash and then insert it
  into the necessary queries. I've probably named this badly, need to document
  it somewhere, and might have missed a place where the params should be used.
  Thoughts welcome!

On the collideoscope side:
- Uses this new mechanism in the smidsy cobrand to filter by `external_body`
  and to lift the time interval restriction on problems, so that it can
  display all of the stats19 problems.
- Removes the old js code that connected the "show stats19 problems" checkbox
  to the 'show old pins' button as a hacky way of approximating it and
  replaces the checkbox with a link, because it's an easier way of making it
  work when there's no JS.
- Fixes a somewhat related bug where the sidebar wasn't getting updated with
  a list of problems that were on the map because it was missing an ID, and
  when that was fixed, another bug where the list generated didn't use quite
  the right styles because the standard FMS templates weren't being used.
- Adds some new js to make the stats19 link do things when it's clicked.

@Benjam, @zarino I've obviously taken a little liberty here with the styling.
I _could_ re-implement the checkbox in JS only if you think it matters? (Or
alternatively make it look like a button or something else appropriate). It
can't be a form element and work without JS though, because the whole sidebar
gets wrapped in a form and it will break.

Fixes https://github.com/mysociety/FixMyStreet-Commercial/issues/712

### DIFF
--- a/templates/web/smidsy/maps/osm-toner-lite.html
+++ b/templates/web/smidsy/maps/osm-toner-lite.html
@@ -11,5 +11,8 @@
 
 [% map_html = BLOCK %]
 [% INCLUDE maps/openlayers.html %]
-<script type="text/javascript">fixmystreet.show_stats19 = '[% show_stats19 %]';</script>
+<script type="text/javascript">
+    var fixmystreet = fixmystreet || {};
+    fixmystreet.show_stats19 = '[% show_stats19 | escape_js | html %]';
+</script>
 [% END %]
--- a/templates/web/smidsy/maps/osm-toner-lite.html
+++ b/templates/web/smidsy/maps/osm-toner-lite.html
@@ -1,0 +1,15 @@
+[% map_js = BLOCK %]
+<script type="text/javascript" src="[% version('/js/OpenLayers.fixmystreet.js') %]"></script>
+<script type="text/javascript" src="https://stamen-maps.a.ssl.fastly.net/js/tile.stamen.js?v1.3.0"></script>
+<script type="text/javascript" src="[% version('/js/map-OpenLayers.js') %]"></script>
+<script type="text/javascript" src="[% version('/js/map-toner-lite.js') %]"></script>
+<script type="text/javascript" src="[% version('/js/jquery.ba-hashchange.min.js') %]"></script>
+<!--[if lte IE 6]>
+  <link rel="stylesheet" href="/js/OpenLayers-2.13.1/theme/default/ie6-style.css" type="text/css">
+<![endif]-->
+[% END %]
+
+[% map_html = BLOCK %]
+[% INCLUDE maps/openlayers.html %]
+<script type="text/javascript">fixmystreet.show_stats19 = '[% show_stats19 %]';</script>
+[% END %]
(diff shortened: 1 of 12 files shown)


web/cobrands/smidsy/js/report.js
```diff
@@ -50,4 +50,40 @@ $(function() {
...
+        $link.html(hideText);
+      } else {
+        $link.data({'show-stats19': 1});
+        $link.html(showText);
+      }
+    };
+    if (window.fixmystreet.show_stats19 === '1') {
+      // Force a load of the pins with the stats_19 param set up on first
+      // load because FMS' default _onload function won't know about it.
+      window.fixmystreet.markers.protocol.options.params.show_stats19 = '[% show_stats19 %]';
```
davea: `window.fixmystreet.markers.protocol.options.params.show_stats19 = '1';` instead maybe? Don't think this file is processed as a template.

templates/web/smidsy/maps/osm-toner-lite.html
```diff
@@ -0,0 +1,15 @@
...
+<script type="text/javascript" src="[% version('/js/map-toner-lite.js') %]"></script>
+<script type="text/javascript" src="[% version('/js/jquery.ba-hashchange.min.js') %]"></script>
+<!--[if lte IE 6]>
+  <link rel="stylesheet" href="/js/OpenLayers-2.13.1/theme/default/ie6-style.css" type="text/css">
+<![endif]-->
+[% END %]
+
+[% map_html = BLOCK %]
+[% INCLUDE maps/openlayers.html %]
+<script type="text/javascript">fixmystreet.show_stats19 = '[% show_stats19 %]';</script>
```
davea: Seems a shame to have to override this whole template to get the flag into the JS. Could it be done in a smaller template?

There's a [XSS hole](http://smidsy.fixmystreet.staging.mysociety.org/around?pc=np203qn&show_stats19=1%27-alert(%27hello%27)-%27) here too, that value needs escaping or sanitising server-side.
Might be worth prefixing this line with `fixmystreet = fixmystreet || {}` to avoid a potential crash.

web/cobrands/smidsy/js/report.js
stevenday: woops, ta - I had this elsewhere in a template initially, obviously didn't check when I moved it.. Fixed in eb129b360821a50266a1a8092c4265dab68e0188

templates/web/smidsy/maps/osm-toner-lite.html
stevenday: Thanks! I've fixed the XSS in 405b0429827519c4831d3b6acc63063de8d0b5cf and added the fms var declaration as a precaution. 

I'm not sure about changing the template though  - this one is quite small already, and the js needs to be here (after the other templated map js is loaded in map/openlayer.html), it seems a fairly obvious place to put it.

PR #1185, opened 2015-08-17
bodies_str lookup refactoring

dracos:
Probably not the easiest refactor to follow, sorry, but hopefully makes sense nevertheless.

This moves "missing" handling to a separate column, as should have been done from the start (only affects 705 rows, thought it'd be a bit more), and simplifies and consolidates the body restriction db code. This speeds up body lookups tremendously in some places such as the dashboard, at the same time fixing a two-tier bug in it, yay.


stevenday:
Comments aside, and caveated by the fact that it took me quite a long time to follow, so I might not have grokked it totally - looks good! :mountain_bicyclist: :random_emoji:

PR #1186, opened 2015-08-18
Use variable for menu image.

dracos:
No need to override if it's set right from the start.


zarino:
Good idea!

Looks good to me.

Do you think you should also update https://github.com/mysociety/fixmystreet-international/blob/0ccd7c0aea5ba0cce13a3774b418874b81a9f857/web/cobrands/makemyisland/_colours.scss ?

dracos:
Yes, I have a commit ready for that after this is merged :)

PR #1193, opened 2015-08-28
Introduced loc labels for a text that is displayed in FMS council...

hcprieto:
instalations. This way, text can be translated to other languages through the standard localization process.
Useful for council installs outside the UK.

(I'm suggesting this micro-change as a way to, additionally, get familiar with the pull request process).


dracos:
Thanks for this; I'm happy for this to be translated if it's useful to you. We don't link to this anywhere yet, sorry, but note at http://mysociety.github.io/internationalization.html that we ask people to not concatenate strings that are going to be translated, as it makes it hard for languages that have differing word orders. In this example, you need to use the `tprintf` function to get what you want, something like `[% tprinf(loc("Confirm the report and %s will investigate"), c.cobrand.council_name) %]` (untested).

If you could get your first line of the commit message to fit in 50 chars, then expand in the body, that makes it easier to use various git tools. Hope that's helpful :)

hcprieto:
Thank you very much for the all the explanations.
This was my first pull request on github and the first time I am working with a perl project.
I had not read the internationalization page you mentioned, and wasn't sure how to include the var within the string. Now I checked all the docs, followed the links... and for sure, I will use tprintf in such cases.
Thanks for the recommendation on the commit message length too! :-)

I'll test the change with tprintf and find the way to resubmit the pull request.

hcprieto:
Ups! I didn't realice the updates would be automatically added to the pull request!! Sorry for that!!
And yes, I think I understand what you propose.
Please cancel this pull request. When I am ready to provide you with some valuable change proposals... I'll do as you propose, using an special branch.
I am not 100% sure right now how I'll do it isolate the changes... but for sure we will manage. :-)
Thanks again for all your support.

PR #1198, opened 2015-09-01
Use base URL in cobrand alert for no-body report.

dracos:
If a cobrand has a body restriction, then a report without any body
won't be shown on it, so we need to make sure links to the report (in
email alerts and RSS) are to the base URL, not the cobrand.


davea:
Would it be worth moving this logic into `Cobrand::Default::base_url_for_report`, or might other things break?

Apart from that, :+1:

dracos:
Yes, good idea, thanks – took me a couple of tries, but hopefully it might be better now.

davea:
Nice, I like it. :shipit:

PR #1199, opened 2015-09-01
Translate state column in admin index and lists.

### LINKED ISSUE #1179: Translation of report status
Is there a way to get the translation of report statuses in other languages? In the admin panel, it is shown in English.

dracos:
Fixes #1179.


davea:
Nice, I like the improvement to the grouping of states on the summary page.

:ship:

PR #1200, opened 2015-09-01
Show report details more clearly on questionnaire page

### LINKED ISSUE #1104: Questionnaire page hides the report details
If you've come to the Questionnaire page via the link in the email, chances are you won't immediately know which report you're being asked to evaluate. Especially if you're a heavy user and your report titles are generic like "Flytipped mattress" or "Pothole".

The Questionnaire page, by default, doesn't give you a single clue as to which report we're asking you about.

To see the report details, you have to click the "Your report" tab – which @paulfurley didn't notice until I studied the page with him a second time.

screenshot from 2015-05-01 14 18 39

Could we find a way of putting the report details right there, next to or above the questionnaire form, so you can see them both at the same time?

zarino:
Fixes #1104.

screen shot 2015-09-01 at 16 58 06

This continues to introduce the slightly lighter, more modern look from other pages like the unresponsive council page and the action confirmation pages.

But I held back on cleaning up *all* the typography and form styles on the page, because I wanted to keep this commit as simple as possible. I'm sure we'll revisit FMS.com form styling, in particular, at some point, so I think this can go live in the meantime.


templates/web/base/questionnaire/index.html
```diff
@@ -1,18 +1,37 @@
 [%
-    pre_map = INCLUDE 'report/_main.html';
+    SET bodyclass = 'mappage';
+    PROCESS "report/photo-js.html";
```
davea: Because you're including this from a template in the `base` cobrand, you'll need to move `photo-js.html` from the `fixmystreet` cobrand to `base` too.

templates/web/base/questionnaire/index.html
```diff
@@ -21,75 +40,60 @@
 
 <input type="hidden" name="token" value="[% token | html %]">
 
-<p>
-[% loc('The details of your problem are available on the right hand side of this page.') %]
-[% loc('Please take a look at the updates that have been left.') IF updates %]
```
davea: A test in [`questionnaire.t`](https://github.com/mysociety/fixmystreet/blob/22c74ad1b5eae2a4c356fb7d35adae762c60488d/t/app/controller/questionnaire.t#L316-L320) relies on this text being present to verify that updates are included on the questionnaire page. Is there some replacement text we can use on L318 to fix the test? ("Last update", or "Show all updates", maybe?)

davea:
If a report has no updates and a portrait photo, the photo can extend beyond the grey box and interfere with the "has this problem been fixed?" question:

screen shot 2015-09-02 at 10 29 46

zarino:
@davea @dracos – I've fixed those little issues, but the tests are still failing – supposedly because the "Would you like to receive another questionnaire" text is visible when it shouldn't be, but I can't tell why. :disappointed:

dracos:
That is a reportemptyhomes test (though the site is gone), and the exception for that showing that question (it never does for that cobrand) has been lost from the template. Have put it back.

templates/web/base/questionnaire/index.html
```diff
@@ -1,18 +1,37 @@
...
+        <a href="/report/[% problem.id %]">Show all updates</a>
+    </p>
+    <p class="questionnaire-report-reminder__last-update">&ldquo;[% add_links( updates.last.text ) %]&rdquo;</p>
+  [% END %]
+</div>
+
+<p>
+[% loc('An update marked this problem as fixed.') IF problem.is_fixed %]
+[% loc('Has this problem been fixed?') %]
+</p>
```
dracos: Is there a reason the error output, if any, appears between this question and its possible answers? Could this `p` be moved down to next to the answers?

templates/web/base/questionnaire/index.html
```diff
@@ -21,66 +40,56 @@
...
 [% END %]
 
 [% IF c.cobrand.moniker != 'emptyhomes' %]
-<div id="another_qn">
+<div class="js-another-questionnaire">
     <p>[% loc('Would you like to receive another questionnaire in 4 weeks, reminding you to check the status?') %]</p>
     <p>
         <input type="radio" name="another" id="another_yes" value="Yes"[% ' checked' IF another == 'Yes' %]>
-        <label for="another_yes">[% loc('Yes') %]</label>
+        <label class="inline" for="another_yes">[% loc('Yes') %]</label>
```
dracos: The p here is lacking the `.radio-segmented-control` so doesn't appear like the other questions.

PR #1208, opened 2015-09-04
Upgrade cpanm, bundle carton, improve gettext-extract, misc

### LINKED ISSUE #1202: Generating .po files is hard
It involves installing Locale::Maketext::Extract and running scripts with environment variables set. We should have a way to set this up and scripts that do the right thing.

dracos:
This upgrades cpanm and bundles carton, mostly so it's easy to use that in gettext-extract to create a local lib for running it. Fixes #1202


davea:
LGTM! :+1:

PR #1211, opened 2015-09-11
Right to Left layout

dracos:
Use some variables so that you can set a `$direction` variable in your `_colours.scss` file and have the layout flip. You will also need to add `dir="rtl"` in your `header.html`'s html element.

For testing, uncomment the lines in `fixmystreet/header.html` and `default/_colours.scss` (or put a `$direction` in whichever cobrand you want).

This PR also (came up as I was working on it, sorry):
* switches to using Modernizr to decide whether to show mobile map, rather than the width of the top border of the site-header (thanks past me!);
* updates the URL hash when menu nav button clicked;
* updates terminal title with CSS writing progress in `make_css_watch` (very handy during this coding!);
* adds an easy way to see the "Please check your email" pages for testing
* fixes padding on /alert form.


davea:
Looks good to me! Do you have any thoughts on how we can preserve this functionality for new features/improvements or is it as simple as just remembering to use `margin-#{$left}` instead of `margin-left` etc? Perhaps a git hook for stylesheets in the base cobrand, or is that overkill?

:+1:

PR #1212, opened 2015-09-14
New high-res OpenGraph image for all cobrands

### LINKED ISSUE #1184: Make a default image for opengraph
Related to #613

zarino:
Fixes #1184.

This is the image:

fms-og_image


wrightmartin:
:+1: lovely

dracos:
It'd be good if only the bits that different from the parent `header_opengraph.html` template could be changed, using another template, or perhaps only an IF statement would do (or a form of TRY that could fallback if image in cobrand directory does not exist...).

zarino:
From IRC:

```
[11:59:02] zarino says — there are 4 lines of header_opengraph.html that'll potentially change depending on which image a cobrand uses: the image url, mime type, height, and width
[11:59:49] zarino says — so would you prefer me to have a new include called header_opengraph_image.html or something, with those in?
[12:00:11] zarino says — and then include that into header_opengraph.html, with a fallback for themes that don't override it?
[12:00:34] matthew says — yep, sounds perfect
```

dracos:
Great :) Could it possibly be rebased on current master and squashed together before merge, as the branch started a while back?

PR #1214, opened 2015-09-15
Mock MapIt postcode call to fix broken tests.

dracos:
SW1A 1AA has moved, who'd have thought.


davea:
Huh, how about that. Looks good to me :+1:

PR #1217, opened 2015-09-18
Improve accessibility of map pages.

dracos:
Hide the map with aria-hidden, as all report links are duplicated in the main content. Fix skip map link on mobile (it was disappearing). Place delegate handler on body to prevent e.g. VoiceOver announcing everything underneath .container as clickable. Remove header report button rather than hide off-screen, as is a duplicate link to the previous.


davea:
:+1:

PR #1218, opened 2015-09-21
[Zurich] Further improvements to admin workflow (672/690)

davea:
Changes/improvements to admin workflow for problem closure after discussion with client.

### DIFF
--- a/t/cobrand/zurich.t
+++ b/t/cobrand/zurich.t
@@ -782,7 +782,7 @@ subtest "test stats" => sub {
         $mech->get_ok( '/admin/stats' );
         is $mech->res->code, 200, "superuser should be able to see stats page";
 
-        $mech->content_contains('Innerhalb eines Arbeitstages moderiert: 2'); # now including hidden
+        $mech->content_contains('Innerhalb eines Arbeitstages moderiert: 3');
         $mech->content_contains('Innerhalb von f&uuml;nf Arbeitstagen abgeschlossen: 3');
         # my @data = $mech->content =~ /(?:moderiert|abgeschlossen): \d+/g;
         # diag Dumper(\@data); use Data::Dumper;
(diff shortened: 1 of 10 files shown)


t/cobrand/zurich.t
```diff
@@ -515,18 +521,28 @@ subtest "external report triggers email" => sub {
...
+            button => 'publish_response',
             with_fields => {
                 body_external => $external_body->id,
                 external_message => $EXTERNAL_MESSAGE,
             } });
+        $report->discard_changes;
         $mech->get_ok( '/report/' . $report->id );
     };
-    $mech->content_contains('Extern');
+    is ($report->state, 'closed', 'Report was closed correctly') or return; # skipping remaining tests
```
davea: Not sure why the remaining tests would want to be skipped here instead of allowing the test to file, so `or return` should probably be removed.

templates/web/zurich/admin/report_edit.html
```diff
@@ -131,22 +133,20 @@
...
+        <option [% 'selected ' IF s.state == pstate %] value="[% s.state %]">[% s.trans %]</option>
+    [% END %]
+    </select>
 </p>
 
 <script type="text/javascript">
 $(function(){
 
-    $('.assignation select').change(function(){
+    $('.assignation__select').change(function(){
```
davea: Some of the logic in this and the `#state` change handler isn't quite right, but it's fixed in a later commit.

templates/web/zurich/admin/report_edit.html
```diff
@@ -241,38 +247,52 @@
...
         </select>
         <br>
-        [% IF problem.state == 'closed' %]
+        [% IF pstate == 'closed' %]
         <input type="checkbox" name="third_personal" id="third_personal" value="1"[% ' checked' IF problem.extra.third_personal %]>
         <label for="third_personal" class="inline">[% loc('Include reporter personal details') %]</label>
         [% END %]
         [% INCLUDE 'admin/response_templates_select.html' for='external_body' %]
-        <textarea name='external_message' id='external_body' cols=60 rows=5></textarea>
+        <textarea name='external_message' id='external_body' cols=60 rows=5>[% problem.get_extra_metadata('external_message') %]</textarea>
```
davea: Does this need to go through the `| html` filter to be made safe?

templates/web/zurich/report/updates.html
```diff
@@ -8,7 +8,7 @@ <h4 class="static-with-rule">[% loc('Updates') %]</h4>
                     [%# XXX following should honour zurich_closed_states instead? %]
                     [% IF problem.state == 'fixed - council'
```
davea: This should use `problem_has_public_response` instead.

PR #1220, opened 2015-09-21
Prevent potential recursion in session cookie.

### LINKED ISSUE #1077: Prevent $c->cobrand (without pre-set cobrand) from infinite recursion.
This can happen if you create a $c from a script, without setting cobrand.  The error message is confusing:

    Deep recursion on subroutine "Catalyst::Request::path" at ...
    Deep recursion on subroutine "Catalyst::Request::base" at ...
    Deep recursion on anonymous subroutine at accessor Catalyst::Request::base ...

Expected behaviour:

 - informative error message about what you've done wrong
 - OR default to first acceptable cobrand from ALLOWED_COBRANDS / default

dracos:
Unlikely to crop up, but this fixes #1077.


davea:
:+1:

PR #1222, opened 2015-09-21
Make sure Vagrantfile checks out git submodules.

### LINKED ISSUE #1197: Make sure submodules checked out in install script
If someone misses out the `--recursive` from the initial git clone and then sets up Vagrant, it all completes but doesn't run because it can't find code in the submodule. The Vagrantfile (or the install script? Vagrant only dev-usage of the script I imagine though, but might as well?) should make sure the submodule is checked out when run.

dracos:
Fixes #1197.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -29,6 +29,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :shell, :inline => <<-EOS
     # To prevent "dpkg-preconfigure: unable to re-open stdin: No such file or directory" warnings
     export DEBIAN_FRONTEND=noninteractive
+    # Make sure git submodules are checked out!
+    if [ ! -e fixmystreet/commonlib/.git ]; then
+      echo "Checking out submodules"
+      apt-get -qq install -y git >/dev/null
+      cd fixmystreet
+      git submodule --quiet update --init --recursive
+      cd commonlib
+      git config core.worktree "../../../commonlib"
+      echo "gitdir: ../.git/modules/commonlib" > .git
+      cd ../..
+    fi
     # Fetch and run install script
     wget -O install-site.sh --no-verbose https://github.com/mysociety/commonlib/raw/master/bin/install-site.sh
     sh install-site.sh --dev fixmystreet vagrant 127.0.0.1.xip.io


Vagrantfile
```diff
@@ -29,6 +29,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :shell, :inline => <<-EOS
     # To prevent "dpkg-preconfigure: unable to re-open stdin: No such file or directory" warnings
     export DEBIAN_FRONTEND=noninteractive
+    # Make sure git submodules are checked out!
+    git submodule --quiet update --init --recursive
```
davea: `git` isn't installed at this point on a fresh vagrant box, d'oh. :(

Vagrantfile
```diff
@@ -29,6 +29,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :shell, :inline => <<-EOS
     # To prevent "dpkg-preconfigure: unable to re-open stdin: No such file or directory" warnings
     export DEBIAN_FRONTEND=noninteractive
+    # Make sure git submodules are checked out!
+    apt-get -qq install -y git >/dev/null
+    git submodule --quiet update --init --recursive
```
davea: Alas, this still doesn't work because it needs to be run inside the `fixmystreet` directory. But even then... checking out the submodule from within the vagrant guest ends up breaking the git repo on the host because `git-submodule` uses absolute paths within its config files.

So I figured there must be a way to do the submodule check out within the vagrant box and then update the config paths to refer to absolute paths on the host. Turns out you can pass arguments to the provisioning script, so [here's a Vagrantfile](https://gist.github.com/davea/95954eb34c531734eeca) with an updated provisioning script that receives the absolute path from the host and rejigs the git config files to un-break the repo on the host. Should do the trick, both for repos that have been cloned recursively and those that haven't.

dracos: Sigh, sorry for wasting your time there, I thought when I pushed it that I had checked it earlier, but clearly I did not :-/ Thanks for coming up with the Vagrantfile; that'll break the submodule inside the guest, though :) Looking at http://stackoverflow.com/questions/17568543/git-add-doesnt-work along with https://github.com/git/git/commit/d75219b4a8a9c44520ddca234cde992498383b89 and https://github.com/git/git/commit/ea115a0d43d176f873855f9a7372376161dd8e65 the issue is the Vagrantfile uses a precise box, which is git 1.7.9; relative paths should still work, they were just made the default in 1.7.10, so switching to use them should let it work in both guest and host.

davea:
Nailed it! :+1:

PR #1223, opened 2015-09-21
Bring down page up to date, and work via proxy.

### LINKED ISSUE #1164: Down for maintenance page static assets broken on admin site
Which 404 because the server isn't serving them:

screen shot 2015-07-28 at 17 23 42

Should these be somewhere else, like S3?

dracos:
Fixes #1164.


davea:
Looks good! OOI how did you generate this? I'm guessing it was manually as opposed to having a special template that outputs a page suitable for storing as down.html?

:+1:

dracos:
Yeah, I downloaded the front page and switched the references to be absolute manually.

PR #1224, opened 2015-09-21
If a sign-in button explicitly clicked, use it.

### LINKED ISSUE #1091: Password manager can cause log in by password even if "Sign in (by email)" is clicked
As discussed with @dracos.  The fixes for #628 and #816 miss the case where:

 - password (but not name) is filled in
 - password is incorrect
 - you click "Sign in" button in the email section

Solution: If "email_sign_in" is passed in (e.g. that button was clicked explicitly, then we should do email login.

(If RETURN is pressed, as that seems to default to the first submit in the form, which is the password sign in, so will then fall back to checking if password was passed.)

dracos:
Fixes #1091.


davea:
LGTM! :star:

PR #1225, opened 2015-09-22
Add missing background-size to retina image.

### LINKED ISSUE #838: Map icon on homepage renders incorrectly on retina displays
It appears to load the @2x size image, but I can't see anywhere in the CSS that declares a background size to keep it at @1x size

fms

dracos:
Fixes #838.


davea:
Yep! :shipit:

PR #1226, opened 2015-09-22
[Zurich] Admin report printing improvements

davea:
Work for mysociety/FixMyStreet-Commercial#677 that improves the print styles for admin report pages.


templates/web/zurich/admin/report_edit.html
```diff
@@ -278,18 +250,18 @@
 [% END %]
 
 <p align="right" class="report-edit-action">
```
davea: Missed one!

PR #1231, opened 2015-09-24
Some translation/site name improvements

### LINKED ISSUE #348: i18n improvement in templates/web/default/alert/_list.html:25, templates/web/fixmystreet/alert/_list.html:28
The string "(alternatively the RSS feed can be customised, within" occurs at templates/web/default/alert/_list.html:25 and again at templates/web/fixmystreet/alert/_list.html:28, and another string is concatenated after it ("2km", "10km" etc.)

String concatenation is a nightmare for translators.  :)  The elements of the sentence won't necessarily come in the same order in other languages, or they might need to add (e.g.) suffixes to certain elements.

Ideally, for translation purposes, each string should be one "unit of meaning" — generally a full sentence, or a couple of short sentences.

Please could you change this to include a %s tag, rather than concatenate strings. Oh, and do the same everywhere else that strings are concatenated too.  :)  Ta.

### LINKED ISSUE #1205: Remove some "FixMyStreet" hard coded
e.g. contact/blurb.html and contact/index.html

dracos:
Fixes #1205 and fixes #348.


perllib/FixMyStreet/App.pm
```diff
@@ -169,6 +169,7 @@ sub setup_request {
 
     # append the cobrand templates to the include path
     $c->stash->{additional_template_paths} = $cobrand->path_to_web_templates;
+    $c->stash->{site_name} = Utils::trim_text($c->render_fragment('site-name.html'));
```
dracos: This should be further down, post language setting, just in case.

PR #1232, opened 2015-09-25
Streamline navigation menu CSS.

### LINKED ISSUE #1191: Make it easier to restyle the selected item in the menu
From https://github.com/mysociety/fixmystreet/pull/1155/files#r37307090 - whilst styling Bromley's cobrand in their blue colour, I found that the CSS selector for the main menu's selected item were pretty specific (lot's of tag names and id's) and it had to be set separately from the main menu itself. It'd be nice if that was easier.

dracos:
Reduce need for specificity. Also fix some bugs in Harrogate base menu
display, and couple of other cobrand tweaks. Fixes #1191.


stevenday:
Looks much better!

I wonder if, for consistency, `menu-item-selected` should also be replaced with `nav-menu__item--selected`, and perhaps a standard `nav-menu__item` class be added to all the `<li>`'s? (although the latter wouldn't have any default styles so perhaps not).

stevenday:
FYI I've tested this on the 4 or 5 council cobrands available to me locally, but not things fixamingata or fixmybarangay

dracos:
"I wonder if, for consistency, menu-item-selected should also be replaced with nav-menu__item--selected," – ah, I got rid of the CSS for that being needed at all, so I should drop it from the HTML (this is only Harrogate, I think?)

"and perhaps a standard nav-menu__item class be added to all the <li>'s? (although the latter wouldn't have any default styles so perhaps not)." – there actually is a tiny bit in the list-reset-soft. I wondered about that, but decided I'd stick with assuming the lists would have li and a/span under them and stick with that for the moment.

stevenday:
:+1:

PR #1234, opened 2015-09-28
Process update left as part of questionnaire.

dracos:
Treat an update left during a questionnaire the same as one left on a
report page, ie. pass it through cleanup_text. This will also make sure
updates left that are wholly whitespace are errored, or replaced with
the default text.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Questionnaire.pm
+++ b/perllib/FixMyStreet/App/Controller/Questionnaire.pm
@@ -234,6 +234,8 @@ sub process_questionnaire : Private {
 
     map { $c->stash->{$_} = $c->get_param($_) || '' } qw(been_fixed reported another update);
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Questionnaire.pm
```diff
@@ -234,6 +234,8 @@ sub process_questionnaire : Private {
 
     map { $c->stash->{$_} = $c->get_param($_) || '' } qw(been_fixed reported another update);
 
+    $c->stash->{update} = Utils::cleanup_text($c->stash->{update});
```
davea: The other update handling code includes `{ allow_multiline => 1 }` - should it be here too?

dracos: Yeah, might as well :)

PR #1240, opened 2015-09-30
Exclude deleted categories from the all reports page filters

### LINKED ISSUE #1239: Reports list page includes deleted categories
The new combined reports listing page, with filters, includes deleted categories, whereas other pages (e.g. the new report page) purposely exclude them. Was there a reason for that, or do they need to be removed from that page too?

For example, see: http://fix.royalgreenwich.gov.uk/reports/Royal+Borough+of+Greenwich - there's a "Testing" category which is marked as deleted in the admin.

stevenday:
Adds `is_deleted` to the ORM query that pulls out the list of categories for
the filters on the new combined reports page that Greenwich/Oxfordshire/Bromley
use. I've included a basic test too.

Closes #1239


davea:
Looks good! :shipit:

PR #1241, opened 2015-09-30
Don't set whole background style in issue-list css

### LINKED ISSUE #1238: Reports list sass clashes with base styling
e.g. on this page: http://fix.royalgreenwich.gov.uk/reports/Royal+Borough+of+Greenwich

Hover one of the report titles in the list on the left, and you get a tiled image of the marker pin, instead of just one.

cobrands/sass/_reports_list.scss sets a full `background` style for this element, including `no-repeat` but this is overriden by the incomplete `background` here: https://github.com/mysociety/fixmystreet/blob/2566ded7d3c012d8286e0da225c359369b3cf623/web/cobrands/sass/_base.scss#L887 (which in turn looks to be necessary because of this: https://github.com/mysociety/fixmystreet/blob/2566ded7d3c012d8286e0da225c359369b3cf623/web/cobrands/sass/_base.scss#L874). The fact that this just specifies a colour and no repeat, etc details means that we get the browser default (tiled in both directions) later on.

Could those two base styles just be background-colour instead? I'm not sure why we've not noticed this until now, something must have changed, but I can't see what.

stevenday:
The use of the background shorthand was causing cobrands which override this
to have unexpected styling issues because it only set a colour, not the other
background properties.

This commit makes it more specific by just using background-color.

Fixes #1238

### DIFF
--- a/web/cobrands/greenwich/base.scss
+++ b/web/cobrands/greenwich/base.scss
@@ -30,7 +30,7 @@ body.frontpage #site-logo,
 }
 
 .issue-list-a li, .list-a li, #user-meta p, #front-main #postcodeForm {
-    background: $greenwich_light_grey;
+    background-color: $greenwich_light_grey;
 }
 
 label[for=pc] {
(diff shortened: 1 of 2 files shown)


davea:
:+1:

Issue #1141, opened 2015-07-09
Use new filterable report list on fixmystreet.com map pages


stevenday:
As per: https://github.com/mysociety/FixMyStreet-Commercial/issues/400#issuecomment-120010233, https://github.com/mysociety/fixmystreet/pull/1134 added a new (optional) filtering feature which fms.com might benefit from.

dracos:
This should also hopefully as mentioned by @stevenday at https://github.com/mysociety/fixmystreet/pull/1155#discussion_r37619497 take the opportunity to tidy up `_list-entry`/_`item` templates, and types of list they can be in (.report-list, .report-list-a, .issue-list-a and so on).

dracos:
Front page list: `section.full-width ul.issue-list-a [report/_item.html]`
All reports list: `section.full-width ul.issue-list-a [reports/_list-entry.html]`
Your reports list: `ul.issue-list-a.full-width [report/_item.html]`
Your reports list (Greenwich/Oxfordshire): `ul.issue-list-a.full-width [reports/_list-entry.html]`
Around page list: `ul.issue-list-a [report/_item.html]`
Around page list (Greenwich/Oxfordshire): `ul.report-list [reports/_list-entry.html]`
Around page list (Bromley): `ul.issue-list-a [reports/_list-entry.html]`

`[reports/_list-entry.html]`: `[report/_item.html]` (with include_lastupdate)
`[reports/_list-entry.html]` (Greenwich/Oxfordshire): `li.pin_colour a.category h3/p`
`[reports/_list-entry.html]` (Bromley): `li.pin_colour a.category.text img.img span small`

`[report/_item.html]`: `li a.text img.img h4 small`
`[report/_item.html]` (Bromley): Same, span rather than h4

`.issue-list` class: List of updates on Your reports page, list of updates on report page.
`.issue-list-a` class: See list above, plus ward list.

dracos:
The report_list CSS, included by Bromley/Greenwich/Oxfordshire, treats `.report-list` and `.issue-list-a` the same, so I'm not sure of the purpose of the `report-list` class.

dracos:
Bromley low level templates are the same, so have combined them.

dracos:
Merged some base/fixmystreet templates, and added the filters to all cobrands. Then tidied it up so that front page has a different template from the map page three, which makes more sense.

PR #1245, opened 2015-10-06
Refactoring tidying

dracos:
This is preparatory work for #1141. It combines many templates, removes bit of unneeded stuff, and fixes a bug.


stevenday:
This looks good! Lots of red which is always nice :+1: 

I would say that the commit message for 4448e1594eaeceba3750ae9102965561d848ea57 is a little terse - it took me a while to understand everything that was going on there.

PR #1246, opened 2015-10-06
add page about surveys/questionnaires

### LINKED ISSUE #728: Explain questionnaire in documentation
As I've just realised that it doesn't actually tell you! (how it works page, I would assume, at the least.)

Anything else totally obvious and totally missing - put it here!

davewhiteland:
Main things:
* what are they?
* how do you get the results?
* how do you switch them off?

Closes #728

### DIFF
--- a/customising/cobrand-module.md
+++ b/customising/cobrand-module.md
@@ -70,3 +70,13 @@ have a body that only covered the state council administrative area.
     This is used to turn the full authority name returned by MapIt into a short
 name.
 
+* send_questionnaires
+
+    By default, FixMyStreet [sends questionnaires]({{site.baseurl}}running/surveys)
+    (sometimes called _surveys_) four weeks after a user submitted a report. The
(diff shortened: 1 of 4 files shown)


customising/cobrand-module.md
```diff
@@ -70,3 +70,13 @@ have a body that only covered the state council administrative area.
     This is used to turn the full authority name returned by MapIt into a short
 name.
 
+* send_questionnaires
+
+    By default, FixMyStreet [sends questionnaires]({{site.baseurl}}running/surveys)
+    (sometimes called _surveys_) four weeks after a user submitted a report. The
```
dracos: Should we be consistent in naming? The front end consistently calls them questionnaires, apart from the one admin page that calls them surveys. The documentation seems to prefer surveys; was there a reason for this?

dracos:
This was merged elsewhere but identically?

PR #1249, opened 2015-10-07
Add state/category filters to base cobrand

### LINKED ISSUE #1141: Use new filterable report list on fixmystreet.com map pages

dracos:
This removes the on map/nearby tabs, replacing them with one combined list. The styling is moved to the base, so that it is used on mobile too. A closed option is added to the existing dropdown, and the base/fixmystreet my templates are also combined. Fixes #1141.


templates/web/base/around/on_map_list_items.html
```diff
@@ -1,7 +1,14 @@
-[% IF on_map.size %]
-    [% FOREACH problem IN on_map %]
+[% all_reports = on_map.merge(around_map) %]
```
davea: Appending the lists end-to-end like this means the end result isn't sorted in a particularly useful way (not a new bug in this PR, I know!)
Seeing as `on_map` and `around_map` don't seem to be used on their own any more, is it worth doing this merge in the controller, and sorting the reports appropriately there?

davea:
Looks good, nice to get things tidied up and features into the main cobrand! :+1:

templates/web/base/around/on_map_list_items.html
dracos: It's not /that/ unuseful – it's the reports on the map first by last updated, followed by those off-map by distance, which presumably will confuse someone at some point but in general seems workable enough. I wondered about that, but I thought it might be all right (until someone minds) and clear enough to show that it is the template displaying these two lists together, rather than hold this down trying to come up with an appropriate sort ;)

davea: OK :+1:

PR #1250, opened 2015-10-07
Fix non-working scripts when CDPATH is set.

dracos:
If the CDPATH environment variable is set (so cd can look in multiple
places), cd prints out the resulting directory when used. This confuses
a command sequence used in a couple of places doing (cd && pwd) to get
a directory path. Make sure we ignore any output from cd.

### DIFF
--- a/bin/install_perl_modules
+++ b/bin/install_perl_modules
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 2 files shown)


davea:
Looks good! :+1:

PR #1252, opened 2015-10-08
Improve item list CSS.

dracos:
Generally, a rewrite of `issue-list-a` to `item-list--reports` or `item-list--wards` (two different uses) and `issue-list` to `item-list--updates`, with subsequent renames/changes to children.

For those cobrands using a different list showing a pin icon, use `item-list__item--with-pin` to stop needing to override existing classes.

Fix a variety of niggling issues e.g. ward list display, or overloading of the `.text` class.


templates/web/oxfordshire/front/_list-entry.html
```diff
@@ -0,0 +1,4 @@
+[%
+report = INCLUDE 'report/_item.html' no_fixed = 1;
+report.replace('item-list--reports__item', 'item-list--reports__item item-list__item--with-pin');
```
davea: Oh dear :wink: Would it be better or worse to pass in the extra class as a variable (or even a `with_pin = 1` switch) into the template, dya think? Might make the intention a bit clearer.

templates/web/greenwich/front/_list-entry.html
```diff
@@ -0,0 +1 @@
+../../oxfordshire/front/_list-entry.html
```
davea: I'm not massively keen on introducing dependencies between cobrands like this, as I know it'll catch me out in the future... Given it's a small template could it just be duplicated here?

davea:
The BEMification continues! The CSS changes look good, and everything seems fine in the various cobrands from the user's perspective too. Just a couple of comments above, but otherwise :+1:

templates/web/oxfordshire/front/_list-entry.html
dracos: Extra class makes sense, yeah.

templates/web/greenwich/front/_list-entry.html
dracos: Yep!

PR #1253, opened 2015-10-08
Fix lack of error logging in certain scenarios.

### LINKED ISSUE #404: Install script/nginx set up has no error logging
I don't know where the errors are going, but I can't find them anywhere.

dracos:
Upgrade Catalyst::Engine and Catalyst::Log to 5.90030 so that errors are logged via psgi.errors rather than printed to the (perhaps nulled) stderr. Fixes #404.
Also prevent Catalyst::View::TT from logging errors twice.

### DIFF
--- a/perllib/Catalyst/Engine.pm
+++ b/perllib/Catalyst/Engine.pm
@@ -1,0 +1,738 @@
+package Catalyst::Engine;
+
+use Moose;
+with 'MooseX::Emulate::Class::Accessor::Fast';
+
+use CGI::Simple::Cookie;
+use Data::Dump qw/dump/;
+use Errno 'EWOULDBLOCK';
+use HTML::Entities;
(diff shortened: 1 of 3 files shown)


perllib/Catalyst/Engine.pm
```diff
@@ -0,0 +1,738 @@
...
+            $i++;
+        }
+        $infos = join "\n", @infos;
+    }
+    else {
+        $title = $name;
+        $error = '';
+        $infos = <<"";
+<pre>
+(en) Please come back later
```
davea: Aha, so this is where this comes from!

davea:
Welp I'm not going to pretend there's any need to review the code in Engine.pm and Log.pm, nothing seems to have broken and the logging does only occur once now, so :+1:

PR #1254, opened 2015-10-09
Add Module::Pluggable to cpanfile snapshot.

dracos:
Module::Pluggable is marked as deprecated in Perl 5.18, to be removed
from core in a later version. cpanminus, which is called from carton,
ignores core modules that have been deprecated - running a deployment
install under 5.18 therefore caused a warning to be shown saying that
it could not find Module::Pluggable as no version was included in the
snapshot.


davea:
:shipit:

PR #1255, opened 2015-10-09
[Hart] Fix missing menu class, tidy remaining.

dracos:
The changes in 99610f1 missed out adding a class to the footer menu,
and a few subsequent changes to the CSS.


davea:
:+1:

Issue #1174, opened 2015-08-04
Add a health check page for monitoring


struan:
Should probably fail on the following conditions:

- Any reports unsent after N hours
- Unconfirmed email addresses in the contacts for councils with email send method

PR #1258, opened 2015-10-09
JSON status page

### LINKED ISSUE #1251: Add a simple json API for getting basic info about the site
Something like the Alaveteli https://www.whatdotheyknow.com/version.json - should give current stats on whatever we're tracking as transactions for FMS - and some basic version info would also be useful.

dracos:
Fixes #1251.
Would be good to do #1174 as part of this too, but no point in that holding up this which is fine alone.

### DIFF
--- a/templates/web/base/admin/index.html
+++ b/templates/web/base/admin/index.html
@@ -55,14 +55,16 @@ and to receive notices of updates.
 <h2>[% loc('Update breakdown by state') %]</h2>
 [% PROCESS states object=comments list=comments.keys.sort %]
 
-[% FOREACH category IN categories %]
-    [% IF loop.first %]
-    <h2>[% loc('Category fix rate for problems > 4 weeks old') %]</h2>
+[% IF categories.size %]
+    <h2 id="category-fix-rate">[% loc('Category fix rate for problems > 4 weeks old') %]</h2>
     <table>
         <tr><th>[% loc('Category') %]</th><th>[% loc('Total') %]</th><th>[% loc('Fixed') %]</th></tr>
-    [% END %]
+[% FOREACH category IN categories %]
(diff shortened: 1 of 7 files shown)


templates/web/base/admin/index.html
```diff
@@ -55,6 +55,7 @@
 <h2>[% loc('Update breakdown by state') %]</h2>
 [% PROCESS states object=comments list=comments.keys.sort %]
 
+[% IF categories.size %]
 [% FOREACH category IN categories %]
     [% IF loop.first %]
```
davea: This could be placed outside the `FOREACH` now, thanks to the `IF`.

templates/web/base/admin/index.html
```diff
@@ -64,5 +65,8 @@
         <tr><td>[% category.key %]</td><td>[% category.value.total %]</td><td>[% category.value.fixed / category.value.total * 100 | format('%.1f') %]%</td></tr>
     [% '</table>' IF loop.last %]
```
davea: Ditto comment above

templates/web/base/admin/index.html
```diff
@@ -64,5 +65,8 @@
...
 [% END %]
+[% ELSE %]
+<p><a href="?show_categories=1">[% loc('Category fix rate for problems > 4 weeks old') %]</a></p>
```
davea: Might be nice to include an anchor in the URL hash so the page scrolls directly to the table when the link is followed?

davea:
This is great, and I really like the nice side effect of cobrands getting their own status page with their own values.

A couple of things worth considering:

 - What if a cobrand doesn't want their status page to be visible? Should that be a flag set by the `Cobrand` module or left to Apache to control access?
 - A minor side effect of the 'unique users' calculation is that for cobrands with a very small number of reports is that multiple anonymous reports could potentially be attributed to the same user. e.g. if only a single user has made reports and they're all anonymous, it'll say '10 live problems; from 1 different user'. Unlikely to come up in practice but worth noting I suppose.

PR #1276, opened 2015-10-26
Add generic static route handler.

### LINKED ISSUE #1235: Generic static page route
Currently to add a new static template you also need a couple of lines of corresponding code in Static.pm. Have a generic route there that checks the template exists and uses it if so.

dracos:
This puts all static pages under /about, and lets you add new static
pages by just placing new HTML templates in your about template directory.
Fixes #1235


perllib/FixMyStreet/App/Controller/About.pm
```diff
@@ -0,0 +1,64 @@
...
+    }
+    $found{$page} = undef;
+    return $found{$page};
+}
+
+sub static_dirs {
+    my ($page, $dir_templates) = @_;
+    my @v = ("about");
+    push @v, "static" if -d "$dir_templates/static";
+    push @v, "faq" if -d "$dir_templates/faq" && $page =~ /faq/;
```
davea: Can't see the reason `faq` is special cased after all the template file renames elsewhere... Am I  missing something?

perllib/FixMyStreet/App/Controller/About.pm
```diff
@@ -0,0 +1,64 @@
...
+            foreach my $file ("$page-$lang_code.html", "$page.html") {
+                if (-e "$dir_templates/$dir_static/$file") {
+                    $found{$page} = "$dir_static/$file";
+                    return $found{$page};
+                }
+            }
+        }
+    }
+    $found{$page} = undef;
+    return $found{$page};
```
davea: Would returning `undef` here be more explicit?

davea:
The implementation of this looks fine, though I'm curious if the shift of all the URLs down one level to within `/about` is a deliberate decision or a consequence of the implementation? Given this allows for a more general way of serving static HTML pages, does it make sense to require new pages to be within `/about` if they might belong elsewhere in the URL hierarchy? Would it be much of a faff to put static pages at `/`?

(Curious that GitHub has got confused about the rename of `privacy.html` for `fixmystreet.com` and `fixmystreet-uk-councils`, not that it matters because the files are identical anyway...)

perllib/FixMyStreet/App/Controller/About.pm
dracos: I wanted it to be clear that it was caching the fact it wasn't found (and so returning that). Happy either way.

perllib/FixMyStreet/App/Controller/About.pm
dracos: I thought might as well have this in case someone updated the code but had their own templates that still had the FAQ in the old location, so they didn't break (for example our own `fixmystreet-international` repo which I haven't updated :) )

dracos:
It was a deliberate decision, so that static files had their own namespace that couldn't interfere elsewhere, and none of them (bar perhaps /faq, which should perhaps be /help anyway?) seemed important enough to warrant a top-level. But it wouldn't be much of a faff, I guess, it could be in the `default` fallback handler to check if there's a static file first before 404ing. Not sure.

perllib/FixMyStreet/App/Controller/About.pm
davea: Aha, makes sense! Might be worth putting a comment to that effect in there for future reference. In an ideal world we'd have some deprecation timeline for such things but that's another matter :)

davea:
OK cool, not worth worrying about moving things again then :) :+1:

PR #1283, opened 2015-11-18
[Zurich] Add extra problem metadata to CSV export

davea:
This PR fulfils a request from Zürich to add some extra information to the CSV downloaded from the admin stats page.

I've also included an updated translations file recently provided.

Closes mysociety/FixMyStreet-Commercial#753


dracos:
Does the translation update mean the confirmation email templates etc should be updated to match?

davea:
Yes, I'd forgotten those! Will update.

PR #1287, opened 2015-11-25
Drag-and-drop multiple photo upload, with image preview

### LINKED ISSUE #190: Display thumbnail of photo before submitting it
It would be good to do this as part of adding a nice image uploader (JS based that started uploading immediately, allowing you to fill in the form whilst it's going etc).

zarino:
Fixes #190:
- [x] Client side, with dropzone.js

Further work:
- [x] Tidy up report/new template duplication
- [x] Tidy up photoset error handling for multiple photos
- [x] Server side endpoint to receive images (subset of import code; should accept photo, save it as uploads currently do and return ID to put in `upload_fileid`).
- [x] https://github.com/enyo/dropzone/wiki/FAQ#how-to-show-files-already-stored-on-server (if error on whole form submission comes back from server)
- [x] https://github.com/enyo/dropzone/wiki/FAQ#how-to-show-an-error-returned-by-the-server (for error in individual upload)
- [x] Limit to image/* and check any other config
- [x] Move photo higher up form, so can be done first and then fill in form while uploading?
- [x] Have three fallback input file fields as Zurich currently do.
- [x] Fix tests

Currently looks like this:

drag

The Dropzone widget is only added in browsers that support it. Everything else will just get the current single file `<input>` field.

I've left some `console.log`s in so you can see when various Dropzone events happen.


abimysoc:
Could the photo upload box include some helpful hint text about what NOT to upload photos of? Car numberplates, for example. ;-)

zarino:
After talking to @dracos:

* Only show "drag and drop" wording on devices that can actually drag things from one window to another. Maybe the most logical way to detect this is by comparing `window.width` with `screen.width` and if they're the same, assuming dragging can't happen since the window covers the entire screen?
* Limit Dropzone uploads to 3, and make sure `.dz-max-files-reached` style is clear.

struan:
@zarino if I have my browser full screen I can still CMD-TAB to the Finder, select and start to drag the files I want and then CMD-TAB back to the browser. It's not a common use case but it is one.

templates/web/base/common_header_tags.html
```diff
@@ -8,6 +8,7 @@
 
 <script type="text/javascript" src="[% start %]/jslib/jquery-1.7.2.min.js"></script>
 <script src="[% start %][% version('/js/jquery.validate.min.js') %]" type="text/javascript" charset="utf-8"></script>
+<script type="text/javascript" src="[% start %]/js/dropzone.min.js"></script>
```
dracos: This needs a `version()` wrapper like the others.

zarino:
Ok, amendments made in 818032b4fe5ae58eaab9f2ef3a7cf4807dd8a842.

It now enforces a 3 file upload limit. And shows a cheeky error message when it is breached.

screen shot 2015-11-26 at 13 06 13

On FMS.com (only) we also list the things you should and shouldn't include in the photos. We should monitor whether this succeeds in decreasing user support requests, and if not, remove it, because it makes an already visually busy form even busier.

Given the difficulty of accurately detecting devices where the phrase "drag and drop" does or doesn't make sense (see https://github.com/Modernizr/Modernizr/issues/57#issuecomment-35831713 and https://github.com/Modernizr/Modernizr/pull/993#issuecomment-21233449) I think we're best just leaving the general message in place on all devices:

screen shot 2015-11-26 at 13 03 27

struan:
This looks good to me.

PR #1288, opened 2015-11-27
Reduce script memory usage

dracos:
This moves code around a bit in order for command line scripts not to need to initialize the full application. It also switches to Moo from Moose, we don't need anything complex, moves the code for the main bin sending scripts to their own files and a few other related bits.

I think the only potential issue is if something uses a module that was previously being imported elsewhere (somewhere within the FixMyStreet::App import) and now it doesn't know about it. Tests are fine, but I'm guessing some changed command line scripts may not be fully tested.


struan:
I think this looks ok and tests pass here too.

dracos:
Shouldn't have done 837aa695fd86a6b753fcd1c2f83b695bf5edad64 should have done 0c9e00d52568843b1f4c3e8b02b9e47eccceb6ea

PR #1290, opened 2015-12-02
Adds the ability to limit the all reports index page by cobrand

joeyisking:
Updates the all-reports.json file to include cobrands meaning that users can choose to limit the results if multiple cobrands are running on the same database


dracos:
Thanks very much for this :) I have merged it as 552805b - only to change a couple of spacing issues and shorten the first line of the commit message, hope that's helpful.

PR #1291, opened 2015-12-02
Adding the ability to save custom input fields from adding a new report

joeyisking:
Adds the ability to have custom input fields on the new reports page saved into the database

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -936,6 +936,13 @@ sub process_report : Private {
             $report->non_public( 1 );
         }
 
+        #Get a list of custom form fields we set and add them into @extra
+        if ($c->cobrand->form_extras){
+            foreach my $form_value ($c->cobrand->form_extras){
+                push @extra, { $form_value => $c->get_param_list($form_value)};
+            }
+        }
+        
         $c->cobrand->process_extras( $c, $contacts[0]->body_id, \@extra );
 
         if ( @extra ) {
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -936,8 +936,8 @@ sub process_report : Private {
             $report->non_public( 1 );
         }
 
-        #Get a list of custom form fields we set and add them into @extra
-        if ($c->cobrand->form_extras){
+        # Get a list of custom form fields we set and add them into @extra
+        if ($c->cobrand->form_extras) {
             foreach my $form_value ($c->cobrand->form_extras){
                 push @extra, { $form_value => $c->get_param_list($form_value)};
             }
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -936,6 +936,13 @@ sub process_report : Private {
             $report->non_public( 1 );
         }
 
+        #Get a list of custom form fields we set and add them into @extra
+        if ($c->cobrand->form_extras){
+            foreach my $form_value ($c->cobrand->form_extras){
+                push @extra, { $form_value => $c->get_param_list($form_value)};
```
dracos: I don't believe `get_param_list` is what you want here, as it returns a list value, which will potentially create extra fields in this hashref. e.g. if `$form_value` was `foo` and you had a URL with `foo=1&foo=2&foo=3` then this would end up with `{ foo => 1, 2 => 3 }` being pushed on to `@extra`. I imagine `get_param` should be good enough here, unless you had a specific reason otherwise?

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -936,6 +936,13 @@ sub process_report : Private {
             $report->non_public( 1 );
         }
 
+        #Get a list of custom form fields we set and add them into @extra
+        if ($c->cobrand->form_extras){
```
dracos: As per http://mysociety.github.io/coding-standards.html, please include spaces after the `#` and before the `{`s.

dracos:
Sorry for the delay in looking at this. In general, this looks fine, a couple of things for consideration:
* Looking at the existing `process_extra`s (e.g. in`Cobrand/EmptyHomes.pm`), would it be an idea to have `form_extras` include a required parameter for each field, to automatically error if the field is not present?
* `extra` can handle hashed data as well as a list (the list is mainly used by the Open311 sending of extra metadata); I'd have to have a little more think about whether this is better handled as hash entries e.g. `extra->{$form_value}`, rather than adding to the list. What/were are you using the extra data?

joeyisking:
- I don't currently have the fields being required but it would be an a good idea, however I'm super familiar with Perl or Catalyst so I'm not sure about the best way to do this.

- With the extra fields I was using it on the new report page to add a number of extra text boxes and then display them when the user viewed the form after submission.

dracos:
Hi, please would you look at this commit I have made to a branch: https://github.com/mysociety/fixmystreet/commit/59a9b246a880cc255c60c83e1c60d2e77b82dd62 and let me know if this meets your requirements, though in a slightly different way.

I have called the cobrand function `report_form_extras` rather than just `form_extras` in case we e.g. do the same thing for updates at some point; and its contents are a list something like `[ { name => 'address', required => 1 }, { name => 'passport', required => 0 } ]` (depending on what fields you want of course) so you can state whether fields are required or not.

It then stores these fields in the metadata part of the report's extra (which is a hash), rather than the 
fields part (which is a list), which is hopefully easier to then get information out of. So for example, if your report_form_extras was as above, then after someone had submitted your new report form, you could get the information out by using `$report->get_extra_metadata('address')` and `$report->get_extra_metadata('passport')` rather than having to loop over `get_extra_fields()`. I don't know how you're currently displaying the extra information, hopefully this is easier, but let me know if I've misunderstood something.

Let me know if you have any questions or comments; if this new way of doing it is okay with you, I will merge it soon.

dracos:
Slightly different approach in #1352 merged.

PR #1292, opened 2015-12-04
Allow more general filtering of Nearby

### LINKED ISSUE #1289: Nearby can't be filtered by $c->cobrand->problems
Nearby.pm doesn't filter using `$c->cobrand->problems` so if you have a custom query in the custom cobrand file It will incorrectly return the results on the `/around` page

dracos:
Fixes #1289, and makes it more consistent in general.


struan:
Minor test and small commit message quibbles aside this looks good.

PR #1293, opened 2015-12-05
Default the Google map view to hybrid

dracos:
A repeat of cc6a278; Google have now dropped version 3.20 of their Maps API, which we were using. Version 3.21 is the oldest available version, and only appears to work if you start off in hybrid view.

According to https://developers.google.com/maps/documentation/javascript/versions, in three months 3.21 will be dropped; 3.22 does not currently work with our OpenLayers code at all. Looking at http://dev.openlayers.org/releases/OpenLayers-2.13.1/examples/google-v3.html the maps do still appear if you first switch to the Satellite layer. So investigation will be needed to find out why that happens and if it can be worked around.


davea:
LGTM :+1:

PR #1298, opened 2015-12-09
Prettify admin/status numbers

dracos:
And front page number localisation.


struan:
Small query aside this looks fine.

PR #1304, opened 2016-01-04
Move alert page "recent photos" out of sidebar

### LINKED ISSUE #1168: Recent photos sidebar on alerts page runs over the footer

zarino:
Fixes #1168 (sidebar overlapping footer) by avoiding the sidebar entirely.

screen shot 2016-01-04 at 17 27 29

Photos are presented in a single row, with wrapping disabled. Grey background makes them recede a little, in an effort to make them less distracting.

I'm not convinced of the value of the photos without any anchoring text (lots of shots of potholes doesn't really serve as a glamorous advertisement for the alerts service) but in lieu of a major rethink for the alerts pages, this is a quick fix that avoids the footer overlap problem and reduces the number of pages using one of those smartypants fixed sidebars by two, which can only be a good thing.

### DIFF
--- a/templates/web/base/alert/index.html
+++ b/templates/web/base/alert/index.html
@@ -1,4 +1,4 @@
-[% INCLUDE 'header.html', title = loc('Local RSS feeds and email alerts'), bodyclass = 'twothirdswidthpage' %]
+[% INCLUDE 'header.html', title = loc('Local RSS feeds and email alerts'), bodyclass = 'fullwidthpage' %]
 
 <h1>[% loc('Local RSS feeds and email alerts') %]</h1>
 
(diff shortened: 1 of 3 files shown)


templates/web/base/alert/index.html
```diff
@@ -37,16 +37,16 @@
...
-  <aside>
-    <h2>[% loc('Some photos of recent reports') %]</h2>
-    [% FOREACH p IN photos;
-        photo = p.get_photo_params;
-    %]
-        <a href="/report/[% p.id %]"><img border="0" height="100"
-            src="[% photo.url_tn %]" alt="[% p.title | html %]" title="[% p.title | html %]"></a>
-    [% END %]
-  </aside>
+<h2>[% loc('Some photos of recent reports') %]</h2>
```
dracos: Is this missing the &lt;div class="alerts__nearby-activity"> the other page has?

zarino: The styling for `.alerts__nearby-activity` just adds a grey background, so I left it out on the alert index page because there's already an element immediately above with a similar grey background (the main postcode search form).

dracos: Thanks, just thought I'd check :)

PR #1307, opened 2016-01-12
Add ability for updates to have multiple photos.

### LINKED ISSUE #1300: Multiple photo support/nice UI for updates
Reports now have this, in #1287 (ticket #190), shouldn’t be too much work to adapt this to work in a similar fashion with updates as well.
- [x] Admin bugs displaying/removing multiple photo reports
- [x] Bug in list display of first of multiple photos
- [x] Treat updates same as reports multiple photo wise - uploading
- [x] Treat updates same as reports multiple photo wise - displaying
- [x] Remove unneeded old photo code (should fix dupe of shrinking code, for example).
- [x] Admin update show multiple photos

dracos:
Match what reports can have, with the same interface.
Along the way, fix some bugs to do with admin rotation/removal and cache busting in photo URLs,
combine all the report templates, and tidy up the use of photoset so in templates it's `object.photos` to get all its URLs and ID. Closes #1300.


struan:
The main issue I have, which applies to the multiple uploads for reports is that it fails silently if you upload a PNG, or, I assume, any other non JPG image. The report is accepted fine but no photo is attached and there's no error message.

Otherwise, slight niggle about the admin code, I think this looks good.

PR #1309, opened 2016-01-18
Clearer relationship between map pins and list items

### LINKED ISSUE #1094: The possibility of quickly mapping a report to a report pin without losing the summary context
In the report summary views, the only way to quickly map a report to a pin seems to be to click on the report title, thus navigating away from the summary view to the problem page.

We would be delighted to see that users could quickly find the pin for a given problem just by hovering the mouse pointer over the problem title, like so:

hover-problem

As you can see in the mock-up above, the appropriate pin changes its color when the problem is hovered. We think that such a feature would add to the usability of the view.

zarino:
Fixes #1094.

Hovering over map pins highlights the relevant item from the
sidebar beside the map. And hovering a sidebar item highlights
the relevant map pin.

pins

Clicking a map pin takes you directly to the relevant report
page, just like clicking an item in the sidebar.

This means the OpenLayers popup "balloon" no longer appears on
clicking a pin, and has been removed entirely.

To make customisation of the list item hover colour easier,
it has been abstracted out into a SCSS variable that cobrand
authors should override (as the fixmystreet.com cobrand does)
or leave at its default light grey colour.

### DIFF
--- a/web/cobrands/fixmystreet.com/_colours.scss
+++ b/web/cobrands/fixmystreet.com/_colours.scss
@@ -19,3 +19,6 @@ $col_click_map_dark: #4B8304;
 // The "fixed" message on a report page
 $col_fixed_label: #00BD08;
 $col_fixed_label_dark: #4B8304;
+
+$itemlist_item_background: #f6f6f6;
+$itemlist_item_background_hover: mix(#fff, $primary, 50%);
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -6,6 +6,9 @@ $heading-font: 'Museo300-display', MuseoSans, Helmet, Freesans, sans-serif !defa
 $image-sprite: '/cobrands/fixmystreet/images/sprite.png' !default;
 $menu-image: 'menu-white' !default;
 
+$itemlist_item_background: #f6f6f6 !default;
+$itemlist_item_background_hover: #e6e6e6 !default;
+
 @import "_mixins";
 @import "_report_list";
 
@@ -889,7 +892,7 @@ input.final-submit {
 
 .item-list__item {
   list-style: none;
-  background-color: #f6f6f6;
+  background-color: $itemlist_item_background;
   margin: 0.25em 0 0;
   padding: 0.5em 1em;
   display: block;
@@ -910,8 +913,8 @@ input.final-submit {
     display: block;
     padding: 0.5em 1em;
   }
-  a:hover, a:focus {
-    background-color: #e6e6e6;
+  a:hover, a:focus, &.hovered a {
+    background-color: $itemlist_item_background_hover;
     text-decoration: none;
   }
 }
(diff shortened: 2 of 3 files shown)


zarino:
Would appreciate your thoughts on this @dracos.

Two things I'm not happy with, but it felt like a lot of work to change:

* The items in the sidebar are currently separated by margin. This means, as you mouse up and down, the mouse passes through the 0.25em vertical gap between two items, all the pins unhide momentarily, and then hide again very quickly after, once you've entered the next item. Ideally, we'd mitigate this by removing the margin, and instead using padding. But it's complicated a bit by where the hover event is bound (to the `li` or the `a`). The spacing on those sidebar items is a bit janky anyway, so this felt like a separate ticket to me.
* The map pins should have `{ cursor: pointer }` to hint that clicking them will take you to a new page.

dracos:
Nice :) Does it scroll the sidebar so that the item is in view when you hover over a pin? I feel that's needed if not, otherwise you don't really know what you're hovering over if it's not the first few in the list.

* margin separation - you could have a debouncer of some sort on mouse out, perhaps? So it wouldn't clear for a short period and then if it had started a new one it wouldn't have to do anything.
* map pins - they're being plotted with SVG, so not sure where the CSS would go, but CSS is fine if you can get it to work :)

zarino:
@dracos I'm looking at adding that scroll-to-highlighted-sidebar-item functionality.

But in the meantime, I've spotted a bug and I can't work out what's causing it. Sometimes it happens immediately from pageload, other times you have to trigger it by zooming out and panning around a bit. Basically, the map pins flash. Take a look:

pins

I can see the `overFeature` event listener being called multiple times in quick succession, but I can't tell why it’s happening. Can you replicate it @dracos?

zarino:
@dracos I've added the scrolling behaviour, so items in the sidebar scroll into view when you hover over their map marker:

scroll

I've also (I think) fixed the marker flicker, but it seems to come back once you've zoomed out, and I have no idea why.

dracos:
Ooh, just thought - something in the sidebar that's out of view on the map should scroll the map ;) I can probably have a look at that when I have time to look at this in general.

zarino:
@dracos: Yeah, two things that stopped me doing that:

* I couldn't work out how to scroll the map to a marker (probably an OpenLayers command somewhere?)
* More annoyingly, we're going to compensate for the sidebar item you're hovering over suddenly shooting to the top of the list (because the list shows all the reports sorted by distance from the centre of the screen, so that sidebar item that used to be right at the bottom, will suddenly be right at the top).

dracos:
Agreed about the map scrolling, happy to leave. I've added cursor pointer when on hover and fixed clicking not working and no need for selectFeature override. I assume the debounce thing still needs doing, I didn't see that in any code, can take a look on Tuesday, otherwise looks good :)

dracos:
Debouncing (or whatever you want to call it) done, and I also found the flickering thing happening all the time at a certain window size, tracked it down to the BBOX strategy thinking it needed to refresh all the time when it didn't due to a rounding issue. Hopefully worked around; @zarino if you'd like to look over my changes, I can get them deployed on Tuesday.

zarino:
@dracos :+1: Looks good to me!

PR #1310, opened 2016-01-19
Pad internal update links so they are in view.

### LINKED ISSUE #1308: Internal update links hidden under fixed header
e.g. https://www.fixmystreet.com/report/123456#update_114289 looks like there's no update.

dracos:
Fixes #1308.

### DIFF
--- a/templates/web/base/report/update.html
+++ b/templates/web/base/report/update.html
@@ -28,6 +28,7 @@
                     <p class="meta-2"> [% INCLUDE meta_line %] </p>
                 </div>
             [% ELSE %]
+                <a name="update_[% update.id %]" class="internal-link-fixed-header"></a>
                 [% INCLUDE 'report/photo.html' object=update %]
                 <div class="item-list__update-text">
                     <div class="moderate-display">
@@ -44,7 +45,6 @@
                     [% END %]
 
                     <p class="meta-2">
-                        <a name="update_[% update.id %]"></a>
                         [% INCLUDE meta_line %]
                         [% mlog = update.latest_moderation_log_entry(); IF mlog %]
                             <br /> Moderated by [% mlog.user.from_body.name %] at [% prettify_dt(mlog.whenedited) %]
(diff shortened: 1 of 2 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -182,6 +182,11 @@ a {
   }
 }
 
+.meta-2 a[name] {
```
zarino: You moved the `<a>` outside of `<p class="meta-2">` in [update.html](https://github.com/mysociety/fixmystreet/blob/ca10ef870ca9525d3223fe0cfe9c1c78228cd3d6/templates/web/base/report/update.html) but it looks like you forgot to update this selector to match the new position.

Rather than concocting some selector like `item-list__update-wrap > a[name]` I would probably just add a class to the `<a>` and use that.

web/cobrands/sass/_base.scss
```diff
@@ -182,6 +182,11 @@ a {
...
+    padding-top: 5em;
+    margin-top: -5em;
```
zarino: I think you’ll also need `display: block` if you want this vertical margin rule to have any effect. (Vertical margin is ignored on inline elements.)

web/cobrands/sass/_base.scss
```diff
@@ -182,6 +182,11 @@ a {
   }
 }
 
+.meta-2 a[name] {
+    padding-top: 5em;
```
zarino: Maybe worth leaving a comment that this magic `5em` number is roughly the height of `.nav-wrapper-2` (4em @ 16px) plus the border-top on `.nav-wrapper-2` (4px), plus the padding-top on `.item-list__item` (0.5em @ 16px).

web/cobrands/sass/_base.scss
dracos: Ooh, it works without in Firefox, but as you say needed in Safari/Chrome.

web/cobrands/sass/_base.scss
dracos: Doh, ta. I've done this and also put it in layout rather than base as it doesn't need to apply to base!

zarino:
Hey @dracos. A few small comments. Thanks!

web/cobrands/sass/_base.scss
dracos: Thanks, good idea.

PR #1316, opened 2016-01-22
Add Facebook login

dracos:
This PR adds optional Facebook login to reporting/updating and general login. It first does some tidying of JSON modules, sets up db and config, then refactors the templates to add lots of subtemplates (without any changes), and then adds the new login using these templates. If a Facebook ID is specified the login form changes to have a Facebook login button and an email login button, the email login showing the existing form. Facebook login should work regardless of what is returned by Facebook (ie. outright refusal, only ID, ID+email), inserting into the existing flow of reporting/updating/logging in.

screen shot 2016-01-22 at 18 08 18


davea:
This works really well. It'd be interesting to see how many new users sign up/in with this versus the normal email flow if/when we go live with it on FMS.com

:+1:

PR #1324, opened 2016-02-02
SQL script to bulk-update reports from CSV

davea:
Zurich would like to update the detail text of a whole bunch of reports on the site. They've provided a CSV file with the report ID and the new detail text for the reports in question.
This script applies the new details to the database from the file 'report_updates.txt'.

### DIFF
--- a/bin/zurich/update_report_descriptions.sql
+++ b/bin/zurich/update_report_descriptions.sql
@@ -1,0 +1,23 @@
+-- Zurich would like to update the detail text of a whole bunch
+-- of reports on the site. They've provided a CSV file with the report id
+-- and the new detail text for the reports in question.
+-- This script applies the new details to the database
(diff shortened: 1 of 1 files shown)


bin/zurich/update_report_descriptions.sql
```diff
@@ -0,0 +1,27 @@
...
+UPDATE problem
+SET detail = report_updates.detail
+FROM report_updates
+WHERE problem.id = report_updates.id
+-- Only update a report if its detail field has actually changed:
+AND problem.id in (SELECT p.id
+                   FROM problem p, report_updates u
+                   WHERE p.id = u.id
+                   AND p.detail != u.detail
+                 );
```
dracos: Could this be `AND detail != report_updates.detail`, or was there a reason that didn't work?

dracos:
Looks good.

PR #1327, opened 2016-02-03
Better handle replies to bounce addresses.

dracos:
Auto unsubscribe alert bounces, forward on report bounces and alert replies to support, and send through to report creator non-bounce replies to their report (for systems that ignore both the From and Reply-To headers).


struan:
Looks good.

PR #1328, opened 2016-02-04
Consistent styling for updates on report page and /my page

### LINKED ISSUE #1312: Updates on /my have no padding
<a href="https://cloud.githubusercontent.com/assets/154364/12421122/8b187f1a-beb8-11e5-926b-2742d957b215.png">

</a>
See screenshot to the left.

zarino:
Fixes #1312.

Updates on the `/my` page now have the same styling as updates on the report page.

(At some point we might want these to share the same partial/template – but that felt like a bigger change than was necessary for this ticket. Thoughts, @dracos?)

screen shot 2016-02-04 at 11 54 14

### DIFF
--- a/templates/web/base/my/my.html
+++ b/templates/web/base/my/my.html
@@ -46,11 +46,21 @@
         <ul class="item-list item-list--updates full-width">
     [% END %]
 
-    <li>&ldquo;[% u.text | html %]&rdquo;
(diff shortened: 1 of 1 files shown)


dracos:
Great; yes, I think this is fine, they're sharing the styles and they're different enough currently (moderation, no need to anonymise/show creator name, etc).

PR #1329, opened 2016-02-04
Smaller image display for reports and updates

### LINKED ISSUE #1325: Individual report page needs to handle display of multiple photos
Reports can have multiple photos now. This needs to be displayed properly on the individual report page.

And updates can have multiple photos too. So whatever we come up with should work in the main report body and any updates below it.

zarino:
Displaying the images smaller makes the lines of text easier to read, and makes display of multiple images much better. People can still click the images to see them full size.

Fixes #1325.

Before:

u7yccoefd3

After:

qvt4cmmmxo

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -965,6 +965,28 @@ input.final-submit {
     margin: flip(0.3em 0 0.5em 1em, 0.3em 1em 0.5em 0); // gutter between text and floated image
     width: 33%;
 }
+.questionnaire-report-reminder .update-img {
+    float: $left;
+    margin: flip(0.1em 1em 1em 0, 0.1em 0 1em 1em);
+    width: 120px;
+}
+
+// A wrapper that's put around .update-imgs if there are more than 1 in a row
+.update-img-set {
+    text-align: center;
+    margin: 0 0 1em;
+
+    .update-img {
+        float: none;
+        display: inline-block;
+        margin: 0;
+        width: 31%;
+    }
+
+    .update-img + .update-img {
+        margin: flip(0 0 0 0.5em, 0 0.5em 0 0);
+    }
+}
 
 // fancybox gallery images have a magnifying glass in the corner
 .update-img {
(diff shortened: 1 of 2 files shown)


dracos:
The clearing can cause an issue with multiple portrait photos, e.g. 
screen shot 2016-02-05 at 16 41 47

dracos:
Is the issue that landscape at 33% width is much smaller than portrait at 33% width? Actually, given we have a max of three images, and they're now 33%ish each (and we could know server-side if there was one or more than one), could it perhaps have the images in a row above or below the text if there's multiple ones? Would that work/look good?

web/cobrands/sass/_base.scss
```diff
@@ -985,12 +986,7 @@ input.final-submit {
 }
 
 .problem-header {
-  margin-bottom:1em;
-}
-.problem-header .update-img {
-    float: $right;
-    margin-#{$left}: 0.5em;
-    margin-bottom: 0.5em;
+  margin-bottom: 1em;
```
dracos: As you asked in IRC but have gone offline, it looks like this margin bottom is to provide some extra separation in mobile view between the main problem description and the key tools links bar.

zarino:
If we *know* there are 3 images (which is roughly 2% of reports) then we could line them up horizontally, as you say:

screen shot 2016-02-09 at 11 05 36

However, since the photos can be different heights / aspect ratios, we'd probably need to clear after them, so that the text doesn't start riding up between the images.

Not sure what we'd do in the case of there being *two* images though. Looks odd with the text wrapping into the remaining 32% space:

screen shot 2016-02-09 at 11 13 01

But also looks odd with the text cleared after it.

screen shot 2016-02-09 at 11 14 23

Same if you float left instead:

screen shot 2016-02-09 at 11 14 53

I guess, if you *knew* there were two photos, rather than three, you could set them to 49% width instead, although then it looks like the images (250px) aren't actually wide enough to fill 50% the width of the sidebar:

screen shot 2016-02-09 at 11 19 39

*sigh*

dracos:
What if, for two, if they were kept at 33% but centred, any better? Two in a column at high width, a centred row at low width?

zarino:
Ok, this now looks like:

screen shot 2016-03-02 at 15 47 57

When there is only 1 photo, the photo floats to the side as before.

PR #1346, opened 2016-03-08
Remember user's last anonymous state.

dracos:
If a user is logged in, use their last report/update to set the default anonymity state.


davea:
:+1:

PR #1352, opened 2016-03-09
Add cobrand-specific custom reporting fields

dracos:
Taking the idea from #1291 (thanks), but adding optional required attribute, and storing the results in the extra metadata hash rather than the extra fields list.


dracos:
Looked over by @davea in IRC.

Issue #1341, opened 2016-02-29
FMS.com: A/B test more friendly positioning


zarino:
- [x] Less utilitarian colour scheme, perhaps with an image on homepage that connotes a community coming together to get stuff fixed
- [x] Change homepage title from "report, view or discuss local problems" to "get things fixed in your local community"
- [x] Change examples on homepage to "(blocked pavements, parks and playgrounds in need of maintenance)"
- [x] Change text in green banner on map page to "click map to request a fix"
- [x] Add user segment radio / select box to report form (@dracos)

zarino:
I've started this in [1341-friendly-ab-test](https://github.com/mysociety/fixmystreet/tree/1341-friendly-ab-test)

zarino:
@dracos: I've made the styling changes ([1341-friendly-ab-test](https://github.com/mysociety/fixmystreet/tree/1341-friendly-ab-test)). Can you do the wording, the radio/select box, and the A/B switching mechanic?

![screen shot 2016-03-02 at 11 12 24](https://cloud.githubusercontent.com/assets/739624/13458831/afe4cea4-e067-11e5-8ae1-f751fbe03caf.png)

PR #1353, opened 2016-03-09
A/B test friendly positioning

dracos:
For #1341. Will squash together before merging.


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -59,5 +59,26 @@ sub get_country_for_ip_address {
...
+    return 1 unless $self->{c}->user;
+
+    my $reports = $self->{c}->model('DB::Problem')->search({
+        user_id => $self->{c}->user->id,
+        extra => { like => '%gender%' }
+    });
+
+    while (my $report = $reports->next) {
+        my $gender = $report->get_extra_metadata('gender');
+        return 0 if $gender =~ /female|male|other/;
```
davea: Should `unknown` be included here? Currently a user who clicks 'prefer not to say' will be asked the question every time they make a report.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -59,5 +59,26 @@ sub get_country_for_ip_address {
...
+
+sub ask_gender_question {
+    my $self = shift;
+
+    return 1 unless $self->{c}->user;
+
+    my $reports = $self->{c}->model('DB::Problem')->search({
+        user_id => $self->{c}->user->id,
+        extra => { like => '%gender%' }
+    });
```
davea: It'd be marginally faster, for users with lots of reports, to use `order_by => { -desc => 'id' }` here.

templates/web/fixmystreet.com/report/new/extra_name.html
```diff
@@ -0,0 +1,40 @@
+[%# Same as parent, with added gender question %]
+
+[% IF c.cobrand.ask_gender_question %]
+<p>
+We’re conducting research to make sure our site is good for everyone;
+please help us out by providing your gender:
```
davea: We should add a note making it clear that their answer will not be shared with the council or displayed publicly. Perhaps a link to a longer explanation of why we're asking this on the FAQ or privacy page?

davea:
It's very yellow! The NIU theme really stuck with you eh, @zarino? :wink:

I think it could use a couple of little improvements, as per the inline comments, but otherwise :+1:

templates/web/fixmystreet.com/report/new/extra_name.html
```diff
@@ -0,0 +1,40 @@
...
+</p>
+
+<p class="radio-segmented-control" style="font-size: 80%">
```
davea: This is included on the update form too, but any answer given by the user is discarded. Perhaps this shouldn't be shown on the update form?

zarino:
Yellow is the colour of kings. (And queens.)

templates/web/fixmystreet.com/report/new/extra_name.html
dracos: Whoops! Thanks for spotting.

templates/web/fixmystreet.com/report/new/extra_name.html
dracos: Yes, good point.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
dracos: Thought they might change their mind in future, but you're right, it'd be annoying.

PR #1355, opened 2016-03-10
Angus cobrand & CRM integration

davea:
This PR:

 - Adds a cobrand for Angus Council in line with the angus.gov.uk design
 - Adds a new `Angus` send method to send reports to their Jadu-based CRM
 - Displays council-maintained streetlights as clickable markers on the new report map

For mysociety/FixMyStreet-Commercial#754


cpanfile
```diff
@@ -92,8 +92,9 @@ requires 'XML::Simple';
 requires 'YAML';
 
 feature 'uk', 'FixMyStreet.com specific requirements' => sub {
-    # East Hampshire
+    # East Hampshire & Angus
     requires 'SOAP::Lite';
+    requires 'LWP::Protocol::https';
```
dracos: I think with the more recent social login changes, this line shouldn't be necessary.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -654,7 +654,7 @@ sub response_templates {
 #     Note:   this only makes sense when called on a problem that has been sent!
 sub can_display_external_id {
     my $self = shift;
-    if ($self->external_id && $self->send_method_used && $self->bodies_str =~ /2237/) {
+    if ($self->external_id && $self->send_method_used && $self->bodies_str =~ /(2237|2550)/) {
```
dracos: This is the only hard-coded ID (now IDs) in Problem.pm; can we think of a way to factor it out to the cobrand while we're here? I guess passing `$c` through would be most straightforward, and one way that's already done around here (for passing down to body's URL? :-/ ).

perllib/FixMyStreet/SendReport/Angus.pm
```diff
@@ -0,0 +1,146 @@
...
+
+BEGIN { extends 'FixMyStreet::SendReport'; }
+
+use Try::Tiny;
+use Encode;
+use mySociety::Web qw(ent);
+
+sub get_auth_token {
+    my ($self, $authresult) = @_;
+    if ($authresult =~ m/<AuthenticateResult>\s*([^\s]+)\s*<\/AuthenticateResult>/) {
```
dracos: As XML::Simple is used elsewhere in the file (needs a `use` at the top?), it might be a bit nicer to use it to get the auth token/external ID? Though probably throws in its own issues with external XML...!

perllib/AngusSOAP.pm
```diff
@@ -0,0 +1,168 @@
+package AngusSOAP;
```
dracos: I realise this matches the existing one, but perhaps we should move that and this into a subdirectory.

t/app/sendreport/angus.t
```diff
@@ -0,0 +1,48 @@
...
+#             </AuthenticateResult>
+#             <success>
+#                     True
+#             </success>
+#             <message></message>
+#
+#     </AuthenticateResponse>
+# EOT
+# ;
+    is $angus->get_auth_token($authxml), 'TVRreUxqRTJPQzR5TlRVdU1qSjhNakF4Tmpvd01Ub3lNam94TlRvME16b3pPUT09VGhvdVNoYWx0Tm90UGFzcw==', 'token correct';
```
dracos: This test would presumably pass if the bit above was uncommented :)

For the more general PR comment, you could use/copy `t/open311/endpoint/Endpoint1.pm` to create a more general mock to return data from a fake endpoint.

templates/web/angus/footer.html
```diff
@@ -0,0 +1,30 @@
+                </div><!-- .content role=main -->
+            </div><!-- .container -->
+            [% IF c.cobrand.moniker == 'fixmystreet' AND bodyclass != 'mappage' %]
```
dracos: This bit will never happen.

templates/web/angus/header.html
```diff
@@ -0,0 +1,77 @@
...
+        <meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1">
+        <meta name="HandHeldFriendly" content="true">
+        <meta name="mobileoptimized" content="0">
+
+        [% INCLUDE 'header_opengraph.html' %]
+        [% SET start = c.config.ADMIN_BASE_URL IF admin;
+
+           # For clarity, the 'fixmystreet' moniker (for fixmystreet.com) puts
+           # it stylesheets under fixmystreet.com
+           IF c.cobrand.moniker == 'fixmystreet';
```
dracos: This can also be dropped :)
I guess adding more hooks to the parent header so this doesn't need to all be overridden is probably not worth it.

templates/web/angus/report/new/category_extras.html
```diff
@@ -0,0 +1,29 @@
+<div id="category_meta">
```
dracos: If the only difference here is the lack of `h4` (if I've diffed it right), hide it with CSS instead?

web/cobrands/angus/base.scss
```diff
@@ -0,0 +1,73 @@
...
+.item-list--reports__item, #user-meta p, #front-main #postcodeForm {
+//    background-color: $greenwich_light_grey;
+}
+
+label[for=pc] {
+    color: $nav_fg;
+}
+
+ol.big-numbers > li:before {
+    color: $primary;
```
dracos: This is in a few cobrands now, should we make a variable for it that can be overridden in colours?

templates/web/angus/post_header_extra.html
```diff
@@ -0,0 +1,14 @@
+<div class="wrapper wrapper__breadcrumb">
```
dracos: The use of .wrapper here is confusing to me. On a desktop map page for example, this now means that .wrapper__main_menu is shown, and so some special .nav-menu--breadcrumb display:none CSS has to be added (but OTOH .mappage .wrapper__breadcrumb is set display:none).
.wrapper is used to allow the navigation to move below/above at different widths; with a display:table. I'm not sure that it should be used for anything else.
The only thing I can see that it's doing is causing the floated menu to be contained, that is easily done with the cf class instead. Or please explain what else this is doing :)

web/cobrands/angus/base.scss
```diff
@@ -0,0 +1,73 @@
...
+#front-howto #front_stats {
+    color: $nav_fg;
+}
+
+
+.nav-menu--mysoc {
+  display: none;
+}
+
+.wrapper__breadcrumb, .wrapper__main_menu {
```
dracos: The change here in f866f85e92c8bb10f8a77143081a0e19d43e2600 is not totally followed through in that e.g. it was container__breadcrumb that used to be display:none in base, and it still has a (now doing nothing?) display:block in layout.scss.

web/cobrands/angus/layout.scss
```diff
@@ -0,0 +1,465 @@
+@import "_colours";
+@import "_fonts";
+@import "../sass/layout";
+@import "../sass/report_list_pins";
+
+$max_page_width: 1200px;
+
+/* from http://nicolasgallagher.com/micro-clearfix-hack/ */
+.clearfix:before, .clearfix:after { content: " "; display: table; }
```
dracos: We already have this in FMS core as `.cf` :)

templates/web/angus/header.html
```diff
@@ -0,0 +1,77 @@
...
+    <body class="[% bodyclass | html IF bodyclass %]">
+
+    <div class="wrapper">
+        <div class="table-cell">
+            <header id="site-header" role="banner">
+                <div class="container">
+                    [% INCLUDE 'header_logo.html' %]
+                    <a href="#main-nav" id="nav-link">Main Navigation</a>
+
+                    [% INCLUDE 'angus_search.html' %]
```
dracos: Nice fitting this in here :)

web/cobrands/angus/layout.scss
```diff
@@ -0,0 +1,465 @@
...
+      min-height: 4em;
+    }
+  }
+
+  #user-meta {
+    p {
+      color: $primary_text;
+    }
+
+    & + .wrapper + .wrapper + .container .content {
```
dracos: Would a generic sibling selector be okay here? Think that's supported from IE7 same as `+`.

web/cobrands/angus/position_map.js
```diff
@@ -0,0 +1,345 @@
...
+        select_feature_control.activate();
+
+        // Make sure the fault markers always appear beneath the street lights
+        streetlight_fault_layer.setZIndex(streetlight_layer.getZIndex()-1);
+
+        // Show/hide the streetlight layer when the category is chosen
+        $("#problem_form").on("change.category", "select#form_category", function(){
+            var category = $(this).val();
+            if (category == streetlight_category) {
+                streetlight_layer.setVisibility(true);
```
dracos: There's a race condition here, in that the visibility change causes the "Zoom in" message to be shown underneath the category box, then the server loads the column ID box which appears inbetween those two things, and the map zooms in automatically and the message changes anyway. It'd be good if the message didn't appear until after the zoom/column ID box appearing.

web/cobrands/angus/position_map.js
```diff
@@ -0,0 +1,345 @@
...
+                streetlight_fault_layer.setVisibility(true);
+                zoom_to_streetlights();
+            } else {
+                streetlight_layer.setVisibility(false);
+                streetlight_fault_layer.setVisibility(false);
```
dracos: The back button, which otherwise works to go back to the normal around page, is not removing the street light layer if it's present at the time.

dracos:
A few comments on code, CSS, and JavaScript, it looks very nice :)

perllib/FixMyStreet/DB/Result/Problem.pm
davea: It's tricky, isn't it, because the cobrand `$c->cobrand` is referring to might not be the `Cobrand` class for the body that the report was sent to. Is there a method for getting the correct `Cobrand` based on the problem's `bodies_str`?

dracos: Not sure that matters – this check only needs to be in the `UK.pm` cobrand, if I understand it right; it can then still check hard-coded body IDs, that's fine, but if it could call a function in `$c->cobrand` then those hard-coded IDs would only be in a place that they make sense in.

templates/web/angus/post_header_extra.html
davea: This was me getting confused with the purposes of existing CSS classes  - I didn't need to use `.wrapper` at all.

PR #1358, opened 2016-03-15
Updates to cpanfile to work with perl 5.20/5.22.

dracos:
The following modules had bugs that have been fixed for working in recent perls:
* List::MoreUtils
* Guard
* PadWalker
* aliased
* URI
* Convert::NLS_DATE_FORMAT

The CGI module was removed from core in 5.20, so also include it in the snapshot (I don't think it's actually used, but is a dependency).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ Makefile
 blib/
 inc/
 _Inline/
+*.pyc
 
 # International
 /fixmystreet-international
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,10 @@ notifications:
 
 language: perl
 perl:
-  - "5.14"
+  - "5.14" # wheezy, precise
+  - "5.18" # trusty
+  - "5.20" # jessie, vivid, wily
+  - "5.22" # stretch, xenial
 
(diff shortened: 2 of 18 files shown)


davea:
Well, installing the updated packages on Perl 5.14 inside a precise vagrant box doesn't seem to break anything...
I did try installing 5.22 with perlbrew but ran into problems with `install_perl_modules` and don't want to go too far down that rabbit hole for now.

Could we get Travis to run tests with 5.20/5.22?

dracos:
Good idea! That's thrown up another couple of modules to upgrade to suppress some warnings in 5.22, a regex change in 5.22 (`{` must now be escaped), and shown a couple of places the tests relied upon a certain ordering of a hash.

The remaining two failures are because, similarly, email headers are not appearing in the same order in different perl versions. I could just add another parse of the email text to fix that, but might try and bundle in simplifying the email module handling at the same time, qv https://github.com/mysociety/fixmystreet/blob/28144f5153a0a7f7bad9466c883bd5c147568028/perllib/FixMyStreet/Email.pm#L143-L145

PR #1359, opened 2016-03-15
Don't double-decode geocoded addresses.

dracos:
Perl 5.20 introduced a version of Encode that errors on decoding already
decoded content (rather than returning the same string). Whilst this is
in some way a bug in the code (although it exists because some versions
of FastCGI silently UTF-8 encode the content), in the Perl changelog,
this is covered only by the line: "Encode has been upgraded from version
2.49 to 2.60."


dracos:
Merged as d3d0ab6 (nicer commit message).

PR #1366, opened 2016-03-29
Refactor email

dracos:
The code was using Email::Simple for headers, then creating an on-the-wire email with nice encoded handling, then using Email::MIME if there were attachments. Switch to only using Email::MIME (slightly less nice encoded handling, but whatevs). A step to using same template handling for all email templates.

### DIFF
--- a/perllib/FixMyStreet/Email.pm
+++ b/perllib/FixMyStreet/Email.pm
@@ -1,6 +1,8 @@
 package FixMyStreet::Email;
 
+use Email::MIME;
 use Encode;
+use POSIX qw();
 use Template;
 use Digest::HMAC_SHA1 qw(hmac_sha1_hex);
 use mySociety::Email;
@@ -105,15 +107,10 @@ sub send_cron {
     $site_name = Utils::trim_text(Encode::decode('utf8', $site_name));
     $params->{_parameters_}->{site_name} = $site_name;
 
-    $params->{_line_indent} = '';
-    my $attachments = delete $params->{attachments};
-
-    my $email = mySociety::Locale::in_gb_locale { mySociety::Email::construct_email($params) };
-
-    $email = munge_attachments($email, $attachments) if $attachments;
+    my $email = mySociety::Locale::in_gb_locale { construct_email($params) };
 
     if ($nomail) {
-        print $email;
+        print $email->as_string;
         return 1; # Failure
     } else {
         my $result = FixMyStreet::EmailSend->new({ env_from => $env_from })->send($email);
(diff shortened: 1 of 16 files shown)


t/app/helpers/send_email.t
```diff
@@ -42,7 +42,7 @@ my $email = Email::MIME->new($email_as_string);
 my $expected_email_content =   path(__FILE__)->parent->child('send_email_sample.txt')->slurp;
 my $name = FixMyStreet->config('CONTACT_NAME');
 $name = "\"$name\"" if $name =~ / /;
-my $sender = $name . ' <' . FixMyStreet->config('DO_NOT_REPLY_EMAIL') . '>';
+my $sender = '"' . $name . '" <' . FixMyStreet->config('DO_NOT_REPLY_EMAIL') . '>';
```
davea: Is this duplicating the effort of line 44? For example if `CONTACT_NAME` is `Foo Bar`, will you end up with `""Foo Bar""` here?

perllib/FixMyStreet/Email.pm
```diff
@@ -121,41 +118,152 @@ sub send_cron {
...
+interpreted as the literal value of a header with the same name.
+
+=back
+
+If no Date is given, the current date is used. If no To is given, then the
+string "Undisclosed-Recipients: ;" is used. It is an error to fail to give a
+templated body, From or Subject (perhaps from the template).
+
+=cut
+sub construct_email ($) {
```
davea: I guess you included this here because other projects rely on the version in commonlib? Would it make sense to keep this within commonlib under a new name, then other dependent code can be refactored over time and eventually the old `construct_email` code removed? It'd cut down on duplication over the long run.

davea:
Looks good to me :+1:

t/app/helpers/send_email.t
dracos: Ah, in the test CONTACT_NAME will be "FixMyStreet" by default (it uses the example general.yml file), so this never did anything; it would indeed fail if it had a space in it (it remained with one set of quotes).

perllib/FixMyStreet/Email.pm
dracos: Looks like it's mainly just TheyWorkForYou and internal services/petitions that use it; this is the main use anywhere. I brought this in because making sure those other things had Email::MIME installed didn't seem worth the hassle, plus the next step would be to make this use FMS templates, something other things won't do and making this basically a totally different thing.

davea: OK, makes sense!

PR #1367, opened 2016-03-29
Switch list item heading from h4 to h3.

### LINKED ISSUE #1348: Fixes for WCAG 2.0 AA Rating
A lot of local government sites are or aspire to be WCAG 2.0 AA rated. We go to a great deal of effort to ensure that the user experience is accessible so let's carry that through. 

A quick check on http://achecker.ca/checker/index.php (which might not parse correctly, alternatives available) finds:
- [x] a minor header nesting issue on the homepage/report pages
- [x] file input on report pages missing a label

And from elsewhere:
- [x] Contrast of front page numbers

dracos:
This stops the front page header nesting skipping a level.

Needs checking if this affects any cobrands. Fixes #1348.

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -903,7 +903,7 @@ input.final-submit {
   margin-bottom: 0.5em;
 }
 
-.item-list__item--link {
+%item-list__item--link {
   padding: 0;
   a {
     color: #222222;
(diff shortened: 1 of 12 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -903,7 +903,7 @@ input.final-submit {
   margin-bottom: 0.5em;
 }
 
-.item-list__item--link {
+%item-list__item--link {
```
davea: Nice :+1:

davea:
The only two cobrands whose templates might need looking at are Zurich (has an `h4`) and Bromley (has a `span`).

A quick check of the cobrands looks good to me.

:+1:

dracos:
Thanks, I missed Zurich.

Issue #1369, opened 2016-03-30
Wrong body user default value for "fixed" problems


jonkri:
The "form_state" status drop down menu (visible for body users) defaults to "Open" when the problem is in fact fixed (whether by council or user). All the other problem states are selected correctly.

Without having done a particularly deep analysis of the problem (such as looking into where else in the code analogous problems might be present (except identifying the equivalent code in the bromley and eastsussex cobrands)), I found that a fix to this particular problem is to check for <code>"fixed - council"</code> and <code>"fixed - user"</code> (rather than just <code>"fixed"</code>) in the option element's selected criteria.

I will submit a patch with a suggested correction of the problem shortly.

PR #1370, opened 2016-03-30
Add suggested fix for #1369

jonkri:
Please see #1369 for details.


dracos:
Thanks for this :) I've moved the check into code to make the templates a bit nicer, commit 7d0362ca6a8db2909bbd550edc267f31c24bc42e, but the fix is the same.

PR #1373, opened 2016-04-01
[Oxfordshire] Open311 update timestamps should be in UTC

davea:
The timestamps returned from the database are in server-local time, not UTC.
Because FMS expects timestamps in Open311 responses to be UTC, the timestamps
of updates from OCC appear one hour ahead when DST is in effect.

This commit converts the timestamps to UTC before outputting them.

Fixes mysociety/FixMyStreet-Commercial#499


dracos:
:+1:

davea:
Thanks for the review. I'll leave this open until it's deployed on OCC's servers in case any further tweaks are required.

PR #1376, opened 2016-04-08
Cope with a '/' in body name in body slug.

### LINKED ISSUE #574: Cope with a '/' in body name in body slug
We're currently using the body name in `/reports/Xxxx` but this breaks if the body name (Xxxx) has a slash in it. Maybe use a slug to circumvent the problem?

note: encountered this in fixmindelo cobrand

dracos:
Fixes #574.

### DIFF
--- a/t/app/controller/reports.t
+++ b/t/app/controller/reports.t
@@ -12,6 +12,7 @@ $mech->create_body_ok(2514, 'Birmingham City Council');
 my $body_edin_id = $mech->create_body_ok(2651, 'City of Edinburgh Council')->id;
 my $body_west_id = $mech->create_body_ok(2504, 'Westminster City Council')->id;
 my $body_fife_id = $mech->create_body_ok(2649, 'Fife Council')->id;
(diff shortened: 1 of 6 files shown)


t/app/controller/reports.t
```diff
@@ -122,6 +123,9 @@ is scalar @$problems, 5, 'correct number of problems displayed';
 FixMyStreet::override_config {
     MAPIT_URL => 'http://mapit.mysociety.org/',
 }, sub {
+    $mech->get_ok('/reports/Electricity_Gas+Council');
```
davea: If this was a `follow_link_ok` it'd have the benefit of testing the href generation as well.

davea:
Not quite sure why but the 'List all reported problems' link on the body admin page redirects to `/reports` instead of showing the body if the name contains a slash.

The link uses the URL pattern `/reports/<mapit id>`, which according to [`Reports.pm`](https://github.com/mysociety/fixmystreet/blob/1d453871ac2bb988ddcae29a16b6628781626595/perllib/FixMyStreet/App/Controller/Reports.pm#L191-L192) isn't used.

davea:
👍

PR #1381, opened 2016-04-18
Improve two-tier council reporting workflow

davea:
This PR improves the user journey when making reports to a district council via a UK council cobrand.

 - The wording of the report confirmation message and email more clearly indicates where the report has gone, why, and how to follow its progress.
 - Viewing a report that was made via a cobrand on fms.com includes a link to the [FixMyStreet for Councils](https://fixmystreet.com/about/councils/) page, as well as a link to view that same report directly on the cobranded site.
 - UK council cobrands display all problems on the /around map, and the pins are coloured grey if the receiving council wasn't the same as that of the cobrand. This is in an attempt to reduce duplicate reports, and to stop valid existing reports being hidden in a council's area.

Closes mysociety/FixMyStreet-Commercial#500


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -11,7 +11,8 @@ sub council_url { return 'oxfordshire'; }
 sub is_two_tier { return 1; }
 
 sub base_url {
-    return FixMyStreet->config('BASE_URL') if FixMyStreet->config('STAGING_SITE');
+    my $self = shift;
+    return $self->next::method() if FixMyStreet->config('STAGING_SITE');
```
dracos: Should this also be done to the other council cobrands to which it applies?

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1038,6 +1027,31 @@ sub tokenize_user : Private {
...
+    my $token = $c->model("DB::Token")->create( {
+        scope => 'problem',
+        data => {
+            %$data,
+            id => $report->id
+        }
+    } );
+
+    my $template = 'problem-confirm.txt';
+    $template = 'problem-confirm-not-sending.txt' unless $report->bodies_str;
```
dracos: Good spot!

templates/web/base/tokens/confirm_problem.html
```diff
@@ -7,17 +7,30 @@
...
+            <a class="green-btn" href="[% c.cobrand.base_url_for_report( report ) %][% report.url %]">Go to FixMyStreet</a>
+        </p>
+     [% END %]
 
   [% ELSE %]
     <h2>[% loc('Thank you for reporting this issue!') %]</h2>
     [% IF report.bodies_str %]
       <p>[% loc('It’s on its way to the council right now.') %]</p>
     [% END %]
+    <p><a class="green-btn" href="[% c.cobrand.base_url_for_report( report ) %][% report.url %]">View report</a></p>
```
dracos: Will want to check this with @zarino who presumably deliberately made the title the link here, and might want this displayed a bit differently. This also needs i18n, and also the styling is wrong with fixmystreet.com styling – blue link, green background.

templates/web/base/report/_report_meta_info.html
```diff
@@ -0,0 +1,2 @@
+[% problem.meta_line(c) | html %]
+[%- IF !problem.used_map AND c.cobrand.moniker != 'emptyhomes' %]; <strong>([% loc('there is no pin shown as the user did not use the map') %])</strong>[% END %]
```
dracos: Can drop the emptyhomes exception here.

templates/email/default/problem-confirm.txt
```diff
@@ -11,7 +11,8 @@ website:
 If your email program does not let you click on this link, copy and paste it
 into your web browser and press return.
 [% IF c.cobrand.is_council && !c.cobrand.owns_problem( report ) %]
-Note that [% c.cobrand.council_name %] is not responsible for this type of problem.
+Please note that [% c.cobrand.council_name %] is not responsible for this type
```
dracos: Would squash this with parent.

templates/web/fixmystreet.com/report/_report_meta_info.html
```diff
@@ -0,0 +1,5 @@
+[% problem.meta_line(c) | html %]
+[% IF c.cobrand != problem.get_cobrand_logged AND problem.get_cobrand_logged.is_council %]
```
dracos: Does equality comparison work here? I think you need to compare the moniker or the class name.

templates/web/fixmystreet.com/report/_report_meta_info.html
```diff
@@ -0,0 +1,5 @@
...
+    using <a href="https://www.mysociety.org/services/fixmystreet-for-councils/">FixMyStreet for Councils</a>
```
dracos: This URL doesn't work - is it https://www.fixmystreet.com/about/council ?

templates/web/fixmystreet.com/report/_report_meta_info.html
```diff
@@ -1,5 +1,5 @@
 [% problem.meta_line(c) | html %]
 [% IF c.cobrand != problem.get_cobrand_logged AND problem.get_cobrand_logged.is_council %]
-    using <a href="https://www.mysociety.org/services/fixmystreet-for-councils/">FixMyStreet for Councils</a>
+    using <a href="https://www.fixmystreet.com/about/council">FixMyStreet for Councils</a>
```
dracos: Ah, squash this change into previous commit, then ;)

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -103,9 +98,35 @@ sub reports_body_check {
...
+    my $stats = {
+        fixed   => $fixed,
+        updates => $updates,
+        new     => $new,
+        recency => $recency,
+    };
+
+    return $stats;
+}
+
```
dracos: Could we factor out fetching $self->problems or $self->problems->to_body($self->council_id) rather than the whole function?

templates/web/base/report/_item.html
```diff
@@ -1,5 +1,5 @@
 <li class="item-list__item item-list--reports__item [% item_extra_class %]">
-<a href="[% c.uri_for('/report', problem.id ) %]">
+<a href="[% c.cobrand.base_url_for_report( problem ) %][% problem.url %]">
```
dracos: Might need changing in templates/web/base/maps/noscript_map.html too?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -31,11 +31,6 @@ sub restriction {
     return { cobrand => shift->moniker };
 }
 
-sub problems_restriction {
-    my ($self, $rs) = @_;
-    return $rs->to_body($self->council_id);
-}
```
dracos: I think it might be confusing to have any report viewable inside a cobrand, even though no links are made, it will be stumbled across/linked to incorrectly or similar. Can we keep the report page behaviour the same, perhaps `admin_problems` should remain `problems`, and instead we 'open up' the list requests only, is that silly?

templates/web/oxfordshire/reports/_list-entry.html
```diff
@@ -1,4 +1,5 @@
 [% INCLUDE 'report/_item.html'
     no_fixed = 1
+    include_sentinfo = 1
     item_extra_class = 'item-list__item--with-pin ' _ c.cobrand.pin_colour(problem)
```
dracos: The commit message of this commit doesn't match the contents.

dracos:
None of my "outdated diff" comments are outdated, thanks GitHub!
Looks good, a variety of small fixes.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -705,7 +705,20 @@ sub processed_summary_string {
...
-    my $body = $problem->body( $c );
+    my $body;
+    my $handler = $problem->get_cobrand_body_handler();
+    # If the report was sent to a cobrand that we're not currently on,
+    # include a link to view it on the responsible cobrand.
+    # This only occurs if the report was sent to a single body and we're not already
+    # using the body name as a link to all problem reports.
+    my $bodies_str_ids = $problem->bodies_str_ids;
+    if ( !mySociety::Config::get('AREA_LINKS_FROM_PROBLEMS') && scalar(@$bodies_str_ids) == 1 &&
+         $handler->is_council && $handler->moniker ne $c->cobrand->moniker ){
```
dracos: Given the is_council, I think it'd be preferable if this could be a function in UKCouncil.pm with a cobrand->can check here to call it.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -913,4 +926,37 @@ sub get_time_spent {
...
+e.g.
+    * if the problem was logged at fix.bromley.gov.uk it will return ::Bromley
+    * if a problem was logged at www.fixmystreet.com for the Bromley area, it
+      will *also* return ::Bromley
+
+=cut
+
+sub get_cobrand_body_handler {
+    my $self = shift;
+    my $cobrand_class = FixMyStreet::Cobrand->get_class_for_moniker( $self->cobrand );
```
dracos: Does this have to be via the moniker, or can we just use whatever we are on (as the corresponding code is in UK.pm)? I think this can be simplified, I mean.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -336,5 +336,20 @@ sub report_check_for_errors {
...
+sub get_body_handler_for_problem {
+    my ($self, $row) = @_;
+
+    foreach my $body (split(/,/, $row->bodies_str)) {
+        foreach my $avail ( FixMyStreet::Cobrand->available_cobrand_classes ) {
+            my $class = FixMyStreet::Cobrand->get_class_for_moniker($avail->{moniker});
+            my $cobrand = $class->new( ref $self ? { %$self } : {} );
+            next unless $cobrand->can('council_id');
+            # $c->log->debug( sprintf "Set body to '%s' and cobrand to '%s', which has council id %s", $body, $cobrand->moniker, $cobrand->council_id );
+            return $cobrand if int($body) eq $cobrand->council_id;
```
dracos: This check could possibly fail as body ID does not have to equal council (area) ID, see how e.g. the `owns_problem` function does it.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -336,5 +336,20 @@ sub report_check_for_errors {
     return %errors;
 }
 
+sub get_body_handler_for_problem {
+    my ($self, $row) = @_;
+
+    foreach my $body (split(/,/, $row->bodies_str)) {
```
dracos: bodies_str might be null.

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
davea: Good idea!

templates/web/base/tokens/confirm_problem.html
zarino: https://github.com/mysociety/fixmystreet-commercial/issues/500#issuecomment-212448459

perllib/FixMyStreet/Cobrand/UKCouncils.pm
davea: The way you suggest is what I'd originally done, but broke it out into `admin_problems`/`problems` after we chatted about it. Would a HTTP 301 for non-owned reports redirecting to FMS.com be OK here?

perllib/FixMyStreet/Cobrand/UK.pm
davea: I've refactored this to build a hash of cobrands by ID, but don't think I'm quite sure what you mean here.
Are you suggesting using Problem's `areas` instead of `bodies_str` for determining which cobrand to link to? The issue I found there was that the parent council's area ID is included in `areas` even if the report was sent to the district council, so you end up with a link to view the problem on the parent council's cobrand.

davea:
Thanks for the review! I've addressed most of the points you raised, and will do the various squashes/rewording of commits before merging.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
dracos: Thinking about it some more, perhaps changing Nearby/Around/etc would be better after all. I guess we have the following areas, for UK council cobrands:
* Admin, viewing all problems restricted to a cobrand;
* Around pages, viewing all problems;
* "Report" pages (either individual or list format), viewing cobrand-only problems - that's Report, Photo, Open311, My, JSON (think that's it).
In another scenario, you could imagine cobrands all being able to view all problems, the cobrand is just for rebranding purposes, or for a cobrand to only want to view its problems. So I guess adding another problems lookup for around pages might be easiest? Not sure of best names. problems, admin_problems, and map_problems? Though 'map_problems' would be problems suitable for showing on a map, not the ones to actually show on the current map, hmm.

There are other bugs in this area I've just discovered (e.g. /rss/problems on a cobrand doesn't respect any problems restriction currently).

perllib/FixMyStreet/Cobrand/UK.pm
dracos: No, not suggesting using `areas`. Your hash of cobrands is indexed by MapIt area ID. So rather than looping over split bodies_str and looking up via that, you should do the same as https://github.com/mysociety/fixmystreet/blob/f71c0242734c134ad7b9588f438f5e7d15038720/perllib/FixMyStreet/Cobrand/UKCouncils.pm#L118-L121 in order to compare the bodies' area IDs against your hash.

davea: Got it - 9b4bdbd7d4ae50115f1e9d857d45e5e5a22859d3 hopefully does the right thing :)

davea:
I've rebased this on master and tidied up the fixup commits. Still a bit of squashing to be done before merging, but I've left that for now as it might be easier to review.

dracos:
Looks like Harrogate and Stevenage need the `next` thing adding.

templates/web/base/tokens/confirm_problem.html
```diff
@@ -7,11 +7,19 @@
...
         <h2>Your issue is on its way to the council.</h2>
+        <p>Your reference for this report is [% report.id %], please quote it in any enquiries.</p>
     [% ELSE %]
-        <h2>Your issue is on its way to [% report.body %].</h2>
-        <p>Note that [% c.cobrand.council_name %] is not responsible for this type of problem.</p>
-    [% END %]
-    <p>Your reference for this report is [% report.id %], please quote it in any enquiries.</p>
+        <h2>Thank you for your report.</h2>
+        <p>
+            We don't handle this type of problem, so have passed it on to:
```
dracos: Curly apostrophe :)

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -868,4 +868,20 @@ sub get_time_spent {
...
+
+e.g. if a problem was logged at www.fixmystreet.com, this will be a
+FixMyStreet::Cobrand::FixMyStreet object.
+
+=cut
+
+sub get_cobrand_logged {
+    my $self = shift;
+    my $cobrand_class = FixMyStreet::Cobrand->get_class_for_moniker( $self->cobrand );
+    return $cobrand_class->new;
```
dracos: I think this would be simpler as a read-only lazy `has` attribute, see PhotoSet.pm.

templates/web/base/tokens/confirm_problem.html
zarino: ❤️

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -195,6 +195,7 @@ sub display_location : Private {
                 colour    => $colour,
                 id        => $p->id,
                 title     => $p->title_safe,
+                problem   => $p,
```
dracos: If we're passing this in, should the other attributes just hang off it? Probably not worth it.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -336,5 +336,36 @@ sub report_check_for_errors {
...
+        my $cobrand = $class->new({});
+        next unless $cobrand->can('council_id');
+        return $cobrand if $areas{$cobrand->council_id};
+    }
+
+    return $self->next::method;
+}
+
+sub show_link_to_council_cobrand {
+    my ( $self, $c, $problem ) = @_;
```
dracos: Don't need to pass in $c, it's in $self->{c}, I think?

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -658,7 +658,15 @@ sub can_display_external_id {
 
 sub duration_string {
     my ( $problem, $c ) = @_;
-    my $body = $problem->body( $c );
+    my $body;
+    if ( $c->cobrand->can('show_link_to_council_cobrand') &&
+         $c->cobrand->show_link_to_council_cobrand($c, $problem) ) {
+        my $base_url = $c->cobrand->get_body_handler_for_problem($problem)->base_url;
```
dracos: If the creation of the link string is moved into show_link_to_council_cobrand, then we don't need get_body_handler_for_problem in Default.pm, which is probably a bit tidier as it's all very specific at the moment.

dracos:
A few small things :+1:

PR #1388, opened 2016-04-21
Tidy up in-place moderation UI

zarino:
Part of mysociety/fixmystreet-commercial#731

Moderation UI is now better laid out into logical chunks.

screen shot 2016-04-21 at 13 40 11

There was one bug I couldn't work out, that @davea or @stevenday will need to fix:

- [x] Only show the "Revert to original title/text/photo" checkboxes when there is actually an original state to revert to

And a potential improvement that will make reverting better – again, something for @davea or @stevenday:

- [x] Clicking a "Revert" checkbox should show you a preview of what you'll be reverting to (ie: by replacing the text in the associated text input), in addition to "disabling" the text input.
- [x] For extra brownie points: *Unchecking* a "Revert" checkbox should re-populate the input with whatever its content was before.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -130,6 +130,17 @@ sub load_problem_or_display_error : Private {
     }
 
     $c->stash->{problem} = $problem;
+    if ( $c->user_exists && $c->user->has_permission_to(moderate => $problem->bodies_str) ) {
+        $c->stash->{problem_original} = $problem->find_or_new_related(
+            moderation_original_data => {
+                title => $problem->title,
+                detail => $problem->detail,
+                photo => $problem->photo,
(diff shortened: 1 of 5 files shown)


stevenday:
> Only show the "Revert to original title/text/photo" checkboxes when there is actually an original state to revert to

We went through this together and it's because this: https://github.com/mysociety/fixmystreet/blob/master/templates/web/base/report/_main.html#L28 is a bit naive - if the report has never been moderated `original` is always going to be empty, so the various if statements that check it won't work properly. We need to include something like this: https://github.com/mysociety/fixmystreet/blob/master/perllib/FixMyStreet/App/Controller/Moderate.pm#L60-L65 in the [Report display controller](https://github.com/mysociety/fixmystreet/blob/master/perllib/FixMyStreet/App/Controller/Report.pm#L45) (probably makes sense to move the logic about whether we're moderating or not there too).

PR #1391, opened 2016-04-27
Quieten Open311 errors

dracos:
This should stop the "read timeout" emails.
And also store the actual server error in the database for easier debugging.

### DIFF
--- a/perllib/FixMyStreet/SendReport/Open311.pm
+++ b/perllib/FixMyStreet/SendReport/Open311.pm
@@ -144,10 +144,11 @@ sub send {
             $self->success( 1 );
         } else {
             $result *= 1;
+            $self->error( "Failed to send over Open311\n" ) unless $self->error;
+            $self->error( $self->error . "\n" . $open311->error );
         }
     }
 
-    $self->error( 'Failed to send over Open311' ) unless $self->success;
 
     return $result;
 }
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/SendReport/Open311.pm
```diff
@@ -144,14 +144,11 @@ sub send {
             $self->success( 1 );
         } else {
             $result *= 1;
-            # temporary fix to resolve some issues with west berks
-            if ( $row->bodies_str =~ /2619/ ) {
-                $result *= 0;
-            }
+            $self->error( "Failed to send over Open311\n" ) unless $self->error;
+            $self->error( $self->error . "\n" . $open311->error );
```
dracos: Note: open311->error might not be set here, it's only set if non-200 response received (if bad 200 response received, it is `warn`ed but not set in the error attribute, I guess it probably should be).

davea: If it's not set will this cause the `uninitialised value` warning to be printed? Or is it simply an empty string if not set?

davea: Nevermind, I've just seen `default => ''` in `perllib/Open311.pm` 👍

davea:
👍

perllib/FixMyStreet/SendReport/Open311.pm
dracos: It should be the empty string, from `has error => ( is => 'rw', 'isa' => Str, default => '' );` so it should only add a newline at present.

PR #1392, opened 2016-04-27
Allow change of email in admin to existing entry.

### LINKED ISSUE #1207: Update email address in admin to existing address
Currently I think you can only change a user's email in the admin interface to an address not already present in the database; it should be able to spot and merge to an existing entry too.

dracos:
Fixes #1207


davea:
👍

PR #1406, opened 2016-05-19
Simplified layout logic for map/around page

### LINKED ISSUE #1326: Revisit mouse scroll behaviour on map
I feel the need to record that Google Maps and OpenStreetMap have now trained users to use their middle/scroll mouse button to zoom maps; so now, on FMS, I'm disappointed/bemused when doing that only causes the sidebar to scroll up and down.

### LINKED ISSUE #1350: Stop map being underneath content sidebar/header
Now the sidebar touches the edges around it, there's no need for the map to be stretched behind it covering the whole viewport. This causes a number of issues, such as things listed in the content that are actually underneath the sidebar so aren't visible. It should be possible to simplify what happens here, which sohuld help with #913 and #361 (if we still had balloons) and #1326.

zarino:
This changes the map page layout so that the map is no longer stacked behind the header and content sidebar.

As a handy side-effect, this also means users can use scrollwheels / scroll gestures to zoom the map (hooray!) and the scrollbar for the issue listing is now more obvious because it's inside the sidebar, rather than out on the far right edge of the window.

Technically, as well as making cobrands a bit easier to create (header height is much easier to track down and change), this also let us strip out all the complex JavaScript that handled the off-centre map, which is a relief.

A few non-obvious changes introduced by this PR:

* The yellow `#user-meta` box has been replaced by a "Your account" link in the nav bar, and "Sign out" and "Change password" buttons on the associated "Your account" page. The biggest impact this will have on the typical user is that the "Sign out" link is now one click further away. The main motivator behind this was that the user-meta box, in its previous location, was super hard to position, because it was neither in the header, nor the nav. It also cuts down on visual clutter during the reporting flow.
* There’s a new `.with-notes` class that gets added to various bits of the map page when a cobrand has a `#report-a-problem-sidebar` (ie: ArreglaMiBarrio, FiksGataMi, FixaMinGata, FixMindelo, Zurich). The `.with-notes` style makes the sidebar wider when reporting an issue, with a "middle" column displayed in a light blue background, containing the previous sidebar content. It results in a bit of wasted space in the ArreglaMiBarrio, FiksGataMi, FixaMinGata, and FixMindelo cobrands, but if the maintainers want to solve that, they can transition to using the inline form guidance that has been successful on FMS.com and the UK council cobrands.

A few things I've noticed don't work on my local VM:

* Clicking the map in FiksGataMi and FixaMinGata shows, what I think is, an error about bodies or authorities not being set up properly. I think this is just a config/database thing in my VM, so could someone else check?
* The JavaScript map doesn’t appear in IE7. But the `<noscript>` fallback doesn’t display either (because JavaScript *is* enabled). IE7’s pathetic debugging tools make this hard to track down, but I have a few theories as to what might have gone wrong:
   * We have removed some JavaScript that IE7 relied on?
   * Some JavaScript somewhere relies on a certain CSS style or DOM structure that we have changed, and while this works fine in modern browsers, it halts JavaScript execution on IE7?

Fixes #1350. Also fixes #1326.


abimysoc:
> The yellow #user-meta box has been replaced by a "Your account" link in the nav bar, and "Sign out" and "Change password" buttons on the associated "Your account" page

(ooh, yay!) What happens if you're not currently signed in?

zarino:
@abimysoc:

**Logged out:**

before

**Logged in:**

after

zarino:
And here’s what the "Your account" (previously: "Your reports") page looks like:

screen shot 2016-05-19 at 13 34 27

(The map is missing in this screenshot, because this test user has made no reports, so there are no pins to plot on a map.)

dracos:
I'd be happy for it to always say "Your account", which then asks you to sign in when clicked (as Your reports does now).

templates/web/base/my/my.html
```diff
@@ -7,14 +7,23 @@
...
 
-<p><a href="/auth/change_password">[% loc('Change password') %]</a></p>
+<p>[% c.user.name %] [% c.user.email %]</p>
+
+<p class="my-account-buttons">
+  <a href="/auth/change_password">[% loc('Change password') %]</a>
+  <a href="/auth/sign_out">[% loc('Sign out') %]</a>
+</p>
+
+<h2>[% loc('Your reports') %]</h2>
```
dracos: Nice, didn't realise there were two translations for this...

dracos:
Just a note to check how `with-notes` behaves without JS, and see if this needs changing at all.

web/cobrands/fixmystreet/position_map.js
```diff
@@ -1,21 +1,3 @@
...
-        $('#map_box').prependTo('body').css({
-            zIndex: 0, position: 'fixed',
-            top: 0, left: 0, right: 0, bottom: 0,
-            width: '100%', height: '100%',
-            margin: 0
-        });
-    }
-}
-
+function position_map_box() {}
```
dracos: This has been removed here, can the other position_map functions also similarly be removed? I assume they will be confusing things rather than helping now! Looks to be the following: angus/eastsussex/hart/oxfordshire/warwickshire – certainly the map on Oxfordshire goes 'wrong', appearing under the sidebar again.

And also the map_fix call has been removed, so all the map_fix definitions can be removed too.

web/js/map-OpenLayers.js
```diff
@@ -710,48 +697,29 @@ OpenLayers.Control.Click = OpenLayers.Class(OpenLayers.Control, {
         // Store pin location in form fields, and check coverage of point
         fixmystreet_update_pin(lonlat);
 
-        // Already did this first time map was clicked, so no need to do it again.
-        if (fixmystreet.page == 'new') {
-            return;
```
dracos: Was this removed so that the map recentres upon a reclick?
If so, then the pan should be moved higher up instead; this needs to be present because e.g. it breaks the mobile reporting if you reclick on the map, as multiple OK/Try again buttons are created.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -109,13 +72,10 @@ $(function(){
                     height: 'auto',
                     margin: 0
                 });
-                $('#fms_pan_zoom').css({ top: '2.75em' });
```
dracos: The removal of the different top for mobile/desktop means the navigation is appearing too high up on mobile view.

dracos:
The sidebar 'sliders' - "Get updates" on an around page, or "Wards of this council" on a body page, are broken – they slide up at the wrong width, then the map disappears, and when they slide back down they're narrower than on page load.

web/cobrands/sass/_layout.scss
```diff
@@ -264,40 +289,39 @@ body.mappage {
...
-    .nav-wrapper-2{
-      position: fixed;
-      background: $map_nav_bg;
-    }
-  }
-
-  #site-logo{
-    position: fixed;
-  }
+  #map_box {
```
dracos: #map_box only ever appears on a .mappage so this can be moved out to the top level, and/or the top level #map_box styling which is all overridden can be removed.

web/cobrands/sass/_layout.scss
```diff
@@ -876,40 +840,32 @@ textarea.form-error {
...
-#report-a-problem-sidebar {
-    position: absolute;
-    #{$left}: 29em;
-    width: 15em;
-    @include box-shadow(rgba(0, 0, 0, 0.6) 0px 0px 4px 1px);
+// If the form is .with-notes, limit the width of the inputs
+// to make space for the #report-a-problem-sidebar.
+#side-form {
+  .with-notes & {
+    width: 26em;
```
dracos: Should this be 27em to match the width without the sidebar?

web/cobrands/sass/_layout.scss
```diff
@@ -264,40 +289,39 @@ body.mappage {
...
+    padding: 1em 1em 3em 1em; // extra padding-bottom to clear the .shadow-wrap element
+    overflow: auto; // vertical scrollbar when list is taller than window
+    background-color: #fff; // since no longer in the flow inside .content
+    box-shadow: 0 0 20px rgba(0, 0, 0, 0.2);
+
+    &.with-notes {
+      width: 42em;
+      // TODO: Should have a bitmap image fallback for old browsers!!
+      background-image: -webkit-linear-gradient(90deg, transparent 28em, #E9F2FF 28em);
+      background-image: linear-gradient(90deg, transparent 28em, #E9F2FF 28em);
```
dracos: Should these be 29em to match the 26-27em change?

web/js/map-OpenLayers.js
```diff
@@ -710,48 +697,29 @@ OpenLayers.Control.Click = OpenLayers.Class(OpenLayers.Control, {
...
-        }
+
+        fixmystreet.map.updateSize(); // required after changing the size of the map element
+
+        lonlat.transform(
+            new OpenLayers.Projection("EPSG:4326"),
+            fixmystreet.map.getProjectionObject()
+        );
+        fixmystreet.map.panDuration = 100;
+        fixmystreet.map.panTo(lonlat);
```
dracos: I like this on desktop, I'm not sure about it on mobile width? Happy to be convinced :)

web/js/map-OpenLayers.js
```diff
@@ -710,48 +697,29 @@ OpenLayers.Control.Click = OpenLayers.Class(OpenLayers.Control, {
         // Store pin location in form fields, and check coverage of point
         fixmystreet_update_pin(lonlat);
 
-        // Already did this first time map was clicked, so no need to do it again.
-        if (fixmystreet.page == 'new') {
-            return;
+        // If there are notes to be displayed, add the .with-notes class
+        // to make the sidebar wider.
+        if($('#report-a-problem-sidebar').length){
+          $('#map_box, #map_sidebar').addClass('with-notes');
```
dracos: As this is only added here with a map click on around, this means that, with JS *on*, if the form is submitted and there's a server side error (e.g. bad password) and it returns the form for resubmission, the sidebar then moves inline (as it behaves without JS). I think this can be fixed server side - if we include the sidebar HTML early in the process when there might be a sidebar (so I think that's around and report/new), then it can know whether to put this class in.
As an aside, presumably the class could be added as a parent of both map_box and map_sidebar to save them both having to have it? Though if it's done server side, don't suppose that matters.

web/js/map-OpenLayers.js
```diff
@@ -547,7 +529,23 @@ OpenLayers.Control.PanZoomFMS = OpenLayers.Class(OpenLayers.Control.PanZoom, {
...
+        this.buttons = [];
+        this._addButton("panup", "north-mini.png");
+        this._addButton("panleft", "west-mini.png");
+        this._addButton("panright", "east-mini.png");
+        this._addButton("pandown", "south-mini.png");
+        this._addButton("zoomin", "zoom-plus-mini.png");
+        this._addButton("zoomworld", "zoom-world-mini.png");
+        this._addButton("zoomout", "zoom-minus-mini.png");
+        return this.div;
+    },
```
dracos: Can't be sure it's only this, but this ending comma will definitely be breaking the JavaScript in old IE.

web/js/map-OpenLayers.js
zarino: Yes to all of this!

I hadn't considered that there might be a way to do it server-side. Can we suck the contents of the sidebar template in, and assign it to a variable, and then check the variable for emptiness to see whether a sidebar has been declared? And then, obviously, print out the variable at the required point.

web/cobrands/zurich/layout.scss
```diff
@@ -187,6 +187,20 @@ body.mappage {
     #zurich-footer {
         display: none;
     }
+
+    #map_sidebar,
+    #map_box {
+      top: 10em;
+    }
+
+    #zurich-main-nav {
```
dracos: This means Zurich lose a navbar's worth of map height. Would it work if the nav-wrapper-2 height was kept at 117px, #map_box got 117px top instead of 10em, and zurich-main-nav had top of 117px, no right or left, width of 29em? That seems to better match the existing.

dracos:
Cobrand checking (ignoring the position_map problems mentioned previously, they'll need checking again after that's fixed):
* Angus – the fact the click map banner now butts up against the header looks a bit weird as they're the same colour.
* Hart – going quite a bit wrong on map page.
* Stevenage – ditto. I think Stevenage wants the `table-cell` behaviour this PR removes to have its header work, or at least putting that back fixes quite a lot of it.

dracos:
Phew, think that'll do for starters :-) Looking good, hope everything I've said makes sense.

(I've also rebased on master which has cut the files by 11 :) ).

zarino:
> (I've also rebased on master which has cut the files by 11 :) ).

\o/

Will take a look at these comments either later today, or Monday.

web/js/map-OpenLayers.js
zarino: Ahhh!! Maybe this is what was breaking IE7. Good spot!!

web/js/map-OpenLayers.js
dracos: Yes: `[% TRY %][% sidebar_html = PROCESS 'report/new/sidebar.html' %][% CATCH file %][% END %]` should do it, the you can use that in an `IF` and `[% sidebar_html %]` to print it out.

web/cobrands/angus/position_map.js
```diff
@@ -333,25 +333,4 @@ var add_streetlights = (function() {
...
-        });
-    } else {
-        $('#map_box').prependTo('body').css({
-            zIndex: 0, position: 'fixed',
-            top: 0, left: 0, right: 0, bottom: 0,
-            width: '100%', height: '100%',
-            margin: 0
-        });
-    }
-    add_streetlights();
```
dracos: Angus has this one extra special function, which I think can be put in here as a normal `$(add_streetlights);` instead.

dracos:
I've done some reordering/squashing, let me know what you think. I thought I might also squash together from"Absolutely position map pages" to "Fix "Get updates" / "Wards of this council" drawers" as they're all really the same thing, does that make sense?

I haven't yet checked all the cobrands, but a couple of issues:
* There are hardcoded left/rights, rather than the variable left/rights – e.g. `map_sidebar` is always on the left, but the JS drawer could currently appear on the right :)
* I don't understand the new sidebar JS initialization of variables, `var $sw = $sw || …`.

dracos:
Some cobrand checks new in!
* Harrogate: 10em header. As discussed, perhaps a general header height variable would be nice.
* Warwickshire: Generally broken, sadly, things like a class=container on wrapper look to need fixing.
* East Sussex: Map narrow, wrong header height; looks like this needs moving to "new"(!) HTML layout (ie remove three wrappers that do nowt).

zarino:
@dracos: Harrogate, Warwickshire, and East Sussex all fixed. Also `$mappage-header-height` is now a variable, for those cobrands that need to alter the header height and map_box/map_sidebar top offset.

web/cobrands/sass/_layout.scss
```diff
@@ -272,6 +273,7 @@ body.mappage {
       left: 0;
       right: 0;
       background: $map_nav_bg;
+      height: $mappage-header-height;
```
dracos: I think this might cause the navigation to break if it wraps onto two lines, e.g. http://www.fiksgatami.no/report/36972 at an appropriate width. `nav-wrapper-2` in this file further up has a `min-height` of 4em which I guess should be this variable, rather than adding this line here, does that work?

zarino: `$mappage-header-height` should really only apply to headers on map pages (otherwise we should change the variable name!) – so I've changed the `height` here to `min-height`, which will allow the nav to break onto multiple lines (https://github.com/mysociety/fixmystreet/pull/1406/commits/3850ec485da3ea27a18013f6d80c68bca04279b9#diff-d1a25ba8f0058d138eb8b66a0a772bd8R280)

dracos:
"I think those were all the fixes needed for RTL support?" Sadly not, `with-notes` appeared to the right of the form, rather than the left, and the background was one-way-specific. I changed:
```diff
@@ -370,8 +371,8 @@ body.mappage {
   .with-notes & {
     width: 42em;
     // TODO: Should have a bitmap image fallback for old browsers!!
-    background-image: -webkit-linear-gradient(90deg, transparent 29em, #E9F2FF 29em);
-    background-image: linear-gradient(90deg, transparent 29em, #E9F2FF 29em);
+    background-image: -webkit-linear-gradient(flip(90deg, 270deg), transparent 29em, #E9F2FF 29em);
+    background-image: linear-gradient(flip(90deg, 270deg), transparent 29em, #E9F2FF 29em);
   }
 }
 
@@ -859,9 +860,9 @@ textarea.form-error {
   // If the parent is .with-notes we know we have space to
   // float the sidebar content to the side of the form.
   .with-notes & {
-    float: right;
+    float: #{$right};
     width: 13em;
-    margin-right: -15em;
+    margin-#{$right}: -15em;
   }
```
And made some other fixes for Zurich:
```diff
@@ -27,8 +27,9 @@ body.mappage .content, body.frontpage .content {
     }
 }
 /* The header on a map page needs a shadow too */
-body.mappage .nav-wrapper-2 {
+body.mappage .nav-wrapper .nav-wrapper-2 {
     @include box-shadow(0 0 6px 1px #000);
+    z-index: 2; // One more than #zurich-main-nav so it's on top
 }
 /* Except on admin pages where there's an admin nav directly underneath it */
 body.mappage.admin .nav-wrapper-2 {
@@ -93,7 +94,7 @@ body.mappage #site-logo-text {
         border-top: none;
     }
     .nav-wrapper-3 {
-        height: 5em; // rough figure, enough to show the blue band, but less than .nav-wrapper-2 height
+        height: 6em; // rough figure, enough to show the blue band, but less than .nav-wrapper-2 height
         padding-top: 18px;
     }
 }
@@ -189,7 +190,7 @@ body.mappage {
     }
 
     #map_sidebar {
-        padding-top: 4em; // make space for the #zurich-main-nav
+        padding-top: 3em; // make space for the #zurich-main-nav
     }
 
     #zurich-main-nav {
@@ -236,6 +237,9 @@ body.twothirdswidthpage {
     }
 }
 
+#fms_pan_zoom {
+    top: 2em;
+}
 #fms_pan_zoom_zoomin {
     top: 0;
```

And I changed the `internal link-fixed-header` to use the new variable:
```diff
 .internal-link-fixed-header {
     display: block;
-    /* 5em is roughly the height of .nav-wrapper-2 (4em) plus the border-top on
-    .nav-wrapper-2 (4px), plus the padding-top on .item-list__item (0.5em). */
-    padding-top: 5em;
-    margin-top: -5em;
+    /* Roughly the height of .nav-wrapper-2 (mappage-header-height) plus the
+     * border-top on .nav-wrapper-2 (4px), plus the padding-top on
+     * .item-list__item (0.5em). */
+    padding-top: $mappage-header-height + 1em;
+    margin-top: -($mappage-header-height + 1em);
 }
```

PR #1412, opened 2016-05-24
Summary string template tidy

### LINKED ISSUE #694: Make it easier to customise report summary
Brought over from https://github.com/mysociety/fixmystreet/issues/515

It would be good if the [`processed_summary_string`](https://github.com/mysociety/fixmystreet/blob/1beeb0dd7ad089a3925a326ce54b5030497a6a34/perllib/FixMyStreet/DB/Result/Problem.pm#L667-L693) method was changed so it could be customised in the template. The use case is Oxfordshire had requested that we make the reference number "more obvious", which we weren't able to easily do at the time.

dracos:
This moves an old bit of HTML into a template, and adds a couple of new small templates so the main _main.html template doesn't have to be overridden by Oxfordshire. Fixes #694.


davea:
:shipit:

PR #1413, opened 2016-05-25
Use only one templating system for emails.

### LINKED ISSUE #1410: Move to one email template format
Now #1366 is done, we can much more easily drop one of the two email template formats we have for historical reasons.

dracos:
Historically, emails sent offline (alerts, questionnaires, etc) used a
different templating system from those sent by the website (e.g. login
emails), though the newer system was also being used for the site name
and signature of offline emails.

Fixes #1410.


davea:
Hooray for a comprehensive test suite that makes this sort of change relatively safe!

I was initially a bit scared of the massive diff, but actually the code changes are fairly small. Looks good!

👍

PR #1427, opened 2016-06-15
[UK Councils] Send correct confirm emails for updates

davea:
Some UK councils with Open311 integrations (e.g. Bromley) have a custom wording
in the confirmation email sent when updates are left on reports, to make the
user aware that the update is sent to the council in question.

Bromley noticed that some emails were being sent without this wording, leading
at least one user to contact the council directly about the report.

It turns out that although the email template contains an IF clause to use the
appropriate wording for Bromley (and Stevenage) reports, the incorrect template
file was being used when updates were made via the Bromley cobrand.

This commit solves the problem by introducing a new
`Cobrand::Default::path_to_email_templates` method, which is overridden by
`Cobrand::UKCouncils` to include the `templates/email/fixmystreet` path. Paths
returned by this method are used as the `additional_template_paths` param when
templating emails. A regression test is included.


dracos:
Looks good; do we want to rename templates/email/fixmystreet to fixmystreet.com to match the web templates a bit better and be clear which site it's applying to? (though I realise not identically as web templates have separate for uk councils).

dracos:
Also, I wonder if it works without the stringifys?

PR #1428, opened 2016-06-15
Some Open311 improvements

dracos:


davea:
Nice, XML code looks much cleaner now. 👍

PR #1429, opened 2016-06-15
[Bristol] Cobranded frontend

davea:
This PR adds a cobranded frontend for Bristol City Council.

What's interesting about this particular cobrand is that it uses BCC's own base map tiles, which are provided via WMTS. The code for the Zurich map was refactored to make it more reusable for easily changing the base map with only a few changes.

Fixes mysociety/FixMyStreet-Commercial#766.


web/cobrands/bristol/layout.scss
```diff
@@ -0,0 +1,244 @@
...
+      display: block;
+      border-bottom: solid 5px $g6;
+      border-top: solid 5px $bcc_red;
+      background-color: white;
+    }
+  }
+
+  // We have to override a lot of rules here to make the appearance
+  // of the logged-in UI consistent across pages
+  #user-meta p {
```
dracos: There is no user-meta any more.

PR #1430, opened 2016-06-15
[East Herts] Cobranded frontend

davea:
A nice frontend for FMS to make it fit in with www.eastherts.gov.uk


templates/web/eastherts/header.html
```diff
@@ -0,0 +1,66 @@
...
+           END %]
+        <link rel="stylesheet" href="[% start %][% version('/cobrands/' _ css_dir _ '/base.css') %]">
+        <link rel="stylesheet" href="[% start %][% version('/cobrands/' _ css_dir _ '/layout.css') %]" media="(min-width:48em)">
+        [% extra_css %]
+        <!--[if (lt IE 9) & (!IEMobile)]>
+            <link rel="stylesheet" href="[% start %][% version('/cobrands/' _ css_dir _ '/layout.css') %]">
+        <![endif]-->
+
+        <script src="[% start %][% version('/js/modernizr.custom.js') %]" charset="utf-8"></script>
+        <script src="[% start %][% version('/cobrands/fixmystreet/position_map.js') %]" charset="utf-8"></script>
```
dracos: This script no longer exists.

templates/web/eastherts/header.html
```diff
@@ -0,0 +1,66 @@
...
+    <body class="[% bodyclass | html IF bodyclass %]">
+
+    <div class="wrapper">
+        <div class="table-cell">
+            <header id="site-header" role="banner">
+                <div class="container">
+                    [% INCLUDE 'header_logo.html' %]
+                    <a href="#main-nav" id="nav-link">Main Navigation</a>
+                </div>
+                [% INCLUDE 'header_nav.html' %]
```
dracos: Worth putting this as a hook in base, to save having to override the whole header? Would this help do the same with any other cobrands?

davea: Yes, that would be pretty useful. Extracting the whole `<header>` element into its own template would be good, as I've had to override the whole thing in at least one cobrand.

Issue #1356, opened 2016-03-11
Improve IE7/8 workarounds and remove IE6 ones


zarino:
I just spotted a workaround for IE6 in [position_map.js](https://github.com/mysociety/fixmystreet/blob/396c497b96afe6396d034b83564b56bda8af7e9a/web/cobrands/fixmystreet/position_map.js):

```javascript
function position_map_box() {
    var $html = $('html');
    if ($html.hasClass('ie6')) {
        …
    } else {
        …
    }
}
```

And a quick search for `ie6` in the `*.scss` files returns a surprising number of hits.

In the last 6 months (10th September–10th March), 19 people used IE6 to visit fixmystreet.com. That’s less than 0.01% of visitors.

Do we still need these workarounds?

dracos:
Yes, I'd be happy to drop the IE6 things if also a little bit of work was put into improving IE7/8, only a few niggles but they're quite obvious (e.g. button background missing, navbar background missing, though the latter of those could be 'fixed' somewhat by #1350), and do have some thousands of sessions. Perhaps also checking which bits of IE7 workarounds are still needed (there's hardly any IE8 ones).

PR #1431, opened 2016-06-15
Drop IE6 support, fix IE7

dracos:
This drops IE6 special cases, most of the IE8 ones (they do not appear to be needed in my testing), and fixes the main display issue in IE7. I haven't done any testing of which of the IE7 special cases are needed or not, that is the only big class of things that could be done, but doesn't really seem worth it until such time as IE7 can be dropped.

For #1356.


zarino:
Looks good to me! 👍 

Tested home page, around page, and report pages in IE7 and IE8.

PR #1432, opened 2016-06-16
Load report details via Ajax from /around page

### LINKED ISSUE #1351: Pull in report on list pages but keep in list view
It might be nice for when you select a report on a list view that the report is highlighted on the map, and details of the report (and everything from a report page) are pulled in and displayed directly but staying within the list view, letting you more easily switch to another report (and behaves more like Google Maps qv #1326).

zarino:
Fixes #1351.

*Todo:*
- [x] On mobile, when map re-centres after clicking a report, the map appears to centre on a point about 200px *above* the selected marker. (why??)
- [x] Oxfordshire cobrand needs checking (now that `#site-header` is hidden/revealed by `.mobile-reporting-map`)
- [x] Back button doesn't work
- [x] `sub_map_links` don't get returned to their original state when you go from report detail page back to /around list.
- [x] Window resizing has non-working edge cases
- [x] Not always large marker when clicked on mobile
- [x] When you're viewing an individual report, clicking on the background of the map should return to the list of all reports, rather than starting a new report.
- [x] Page title needs to change (`document.title = "…"`) when you go between all reports view and individual report view, at the same time that the location path changes (`history.pushState`).
- [x] When making a new report, clicking the map again (to reposition the big greeen pin) correctly changes the lat/lon parameters in the URL with a new pushState. But cycling through those states later (with browser back button) doesn't reposition the green pin to match the lat/lon parameters in the URL.
- [x] "Back to all reports" link on individual report page needs styling.
- [x] History API fallback / polyfill for IE9 and below.
- [x] Angus has a 'hashchange' thing


dracos:
"when map re-centres after clicking a report, the map appears to centre on a point about 200px above the selected marker. (why??)" - you need to call updateSize when the map size changes, what it's doing is centring on the full screen map still (hence being lower than you would want).

"sub_map_links don't get returned to their original state when you go from report detail page back to /around list." - these links are in the HTML to begin with, not generated by JS (which only adds the permalink), so you need to get this information from the fetched page and store the original for switching back.

web/cobrands/sass/_base.scss
```diff
@@ -1306,6 +1310,42 @@ a:hover.rap-notes-trigger {
...
+  #fms_pan_zoom {
+    top: 2.75em; // make space for the semi-transparent "Place pin on map" bar
+  }
+
+  .container {
+    padding: 0; // map_box needs to be full width, so remove page gutter
+  }
+
+  #map_sidebar {
+    padding: 1em; // reinstate page gutter, but on sidebar, so map_box is unaffected
```
dracos: Should this be margin: 1em to prevent a big gap above the main header?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -38,118 +38,185 @@ function tabs(elem, indirect) {
...
-        if (last_type == type) { return; }
-        if (type == 'mobile') {
-            $html.addClass('mobile');
-            $('#map_box').css({ height: '10em' });
-            if (typeof fixmystreet !== 'undefined') {
-                fixmystreet.state_map = ''; // XXX
+    // A sliding drawer from the bottom of the page, small version
+    // that doesn't change the main content at all.
+    $.fn.small_drawer = function(id) {
+        var opened, $this = $(this), d = $('#' + id);
```
dracos: There is a reason `opened` was outside this function, and it's so it was a shared variable between the small drawers (so that clicking one when there's one open would close the open one). Moving it inside breaks this.

dracos:
Oxfordshire cobrand needs checking" - site-header needs to always be hidden in this cobrand.

web/cobrands/sass/_base.scss
zarino: I don't think so. Can't spot a big gap on the /around, /report, or /report/new sidebar.

zarino:
@dracos – I've styled up the "back" links, and given this another whirl in a bunch of different browsers. Looks really good.

The only thing missing is nice sharable/bookmarkable URLs in IE8–9. Since they don't support pushState, the URL isn't changing when the page content does. Might just be a case of including the [HTML5 History API](https://github.com/devote/HTML5-History-API) polyfill, in a conditional comment?

```html
<!--[if lte IE 9]>
  <script type="text/javascript" src="history.js"></script>
  <script type="text/javascript">
    var location = window.history.location || window.location;
  </script>
<![endif]-->
```

web/cobrands/sass/_base.scss
dracos: Hmm, I think I must have been thinking of this, but it doesn't seem that bad, so dunno:
2em gap at top

dracos:
I've fixed the Google issue mentioned in a recent commit (map-google was just missing a few lines map-OpenLayers had); and I've rebased this quite a bit to put all the function renaming/refactoring at the start, meaning future commits only show the actual code changes for the features being added. There were no differences at the end before/after rebase besides whitespace, so should be okay!

"The only thing missing" - ah, I'm afraid if only. Here's a new list from my testing:
- [ ] Around page, click to start a new report, hit refresh, click back. Weird halfway state ensues.
- [ ] On a report page on mobile, click the 'menu' button to go to the menu, then click back. Chaos ensues.
- [ ] IE polyfill as you mention
- [ ] (not needed but I'd like to do it) refactor the function names to be much better structured.

dracos:
Fixed Angus issue using hashchange (and removed hashchange entirely).
Will add IE9 fallback now.

I'd like @davea to take a quick peek at this now, given I've basically rewritten all the JS:
* `fixmystreet.set_up` is an object of functions called on document.ready, added to by both `fixmystreet.js` (they could be merged now, I'll probably do that now too).
* `fixmystreet.display` for state changes, `fixmystreet.resize_to` for window resizing, `fixmystreet.mobile_reporting` for full screen mobile UI.
* Google and OpenLayers share their code as much as possible now with any map-layer specific functions under `fixmystreet.maps`.
* Could still be some future work on e.g. the `fms_` functions left in e.g map-OpenLayers, either a closure or sticking them under `fixmystreet.maps`, unless I do that today.

dracos:
I've left today's tidyup of map private functions as a second commit for now, will squash in after review

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -493,14 +667,333 @@ $.fn.drawer = function(id, ajax) {
...
+
+$(function() {
+    window.cobrand = $('meta[name="cobrand"]').attr('content');
+    fixmystreet.original_title = document.title;
+
+    if (typeof variation !== 'undefined' && variation === 1) {
+        $('input[name=variant]').val(1);
+    }
+
+    $.each(fixmystreet.set_up, function(key, value) {
```
zarino: Tiny suggestion: Use more descriptive variables here? Like `(setup_name, setup_func)` or something? If nothing else, it'll make `value()` on the next line look a lot less mysterious.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -493,14 +667,333 @@ $.fn.drawer = function(id, ajax) {
...
+    $.ajax(reportPageUrl).done(function(html, textStatus, jqXHR) {
+        var $reportPage = $(html);
+        var $sideReport = $reportPage.find('#side-report');
+
+        if ($sideReport.length) {
+            $('#side').hide(); // Hide the list of reports
+            $('#side-report').remove(); // Remove any existing report page content from sidebar
+            $sideReport.appendTo('#map_sidebar'); // Insert this report's content
+
+            var found = html.match(/<title>([\s\S]*?)<\/title>/);
```
zarino: Any reason for the regex here, rather than pulling out out of the DOM we created further up? Like: `$reportPage.find('head title').text()`? I guess for speed, maybe?

dracos: `$reportPage` does not contain the head. See http://api.jquery.com/jQuery/#jQuery2: `As mentioned, jQuery uses the browser's .innerHTML property to parse the passed HTML and insert it into the current document. During this process, some browsers filter out certain elements such as <html>, <title>, or <head> elements. As a result, the elements inserted may not be representative of the original string passed.`

web/cobrands/fixmystreet/fixmystreet.js
dracos: Yes, will do, ta.

web/cobrands/fixmystreet/fixmystreet.js
zarino: Wow! Good to know!

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -3,153 +3,430 @@
...
-            $('#fms_pan_zoom').css({ top: '' });
-            $('span.report-a-problem-btn').css({ cursor:'' }).off('.reportBtn');
+        });
+    });
+  },
+
+  add_validation: function() {
+    // Map form doesn't work in some browsers with HTML5 validation and hidden form, so
+    // we disable validation by default, and add it in the JS case.
+    // For some reason, the removeAttr doesn't work if we place it at beginning.
```
zarino: As mentioned in IRC, this final line of the comment isn't required any more.

davea:
A few stream-of-consciousness things as I click around:

Not all the pins are removed when viewing an individual report:
screen shot 2016-07-08 at 16 54 24

Clicking a report in the sidebar that's currently off the map doesn't pan to it, so you don't know where it is.

dracos:
"Not all the pins are removed when viewing an individual report" – that's deliberate, more like Google Maps behaviour, and makes it easier to switch report or go back to around view, etc. (In longer term, an individual report page gone to directly could show other reports around it to make switching to around view easier too).
Oh yes, dur, regarding off-map sidebar click :)

dracos:
As well as hopefully fixing the off-map pin thing, I've fixed some other bugs I spotted for display/performance.

davea:
Clicking the big pin for the current report will reload that report, pushing another entry on to the back button stack and adding another 'Get updates' link to the map, presumably it should be a no-op?
screen shot 2016-07-11 at 10 00 33

Is it worth fading out everything except the current report pin when viewing a single report? In areas with lots of pins it can get quite crowded. (Also, pet peeve alert, I don't like the jittery flashing of so many pins when you move the cursor around the map)
screen shot 2016-07-11 at 10 08 34

web/cobrands/angus/position_map.js
```diff
@@ -318,18 +311,19 @@ var add_streetlights = (function() {
...
+    // Make sure the streetlights get hidden if the back button is pressed
+    fixmystreet.maps.display_around = (function(original) {
+        function hide_streetlights() {
+            streetlight_layer.setVisibility(false);
+            streetlight_fault_layer.setVisibility(false);
+            fixmystreet.markers.setVisibility(true);
+            original.apply(fixmystreet.maps);
+        }
+        return hide_streetlights;
+    })(fixmystreet.maps.display_around);
```
davea: Nice!

davea:
The sharing buttons don't get loaded properly (though they do work):
screen shot 2016-07-11 at 11 22 30

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -493,14 +666,335 @@ $.fn.drawer = function(id, ajax) {
...
+            // individual report's sub_map_links using map_controls().
+            if (!('originalSubMapLinks' in window)) {
+                window.originalSubMapLinks = $('#sub_map_links').detach();
+            }
+            fixmystreet.set_up.map_controls();
+
+            $sideReport.find('#key-tool-problems-nearby').addClass('.js-back-to-report-list');
+            fixmystreet.set_up.map_sidebar_key_tools();
+
+            fixmystreet.set_up.fancybox_images();
```
davea: `$.fancybox` doesn't exist at this point, does `report/photo-js.html` need to be included in the template for `/around`?

dracos: Ah yes, guess so. Theoretically could only pull it in at the point they click on a report, but probably not worth that.

davea:
Is there an intended difference now between 'Back to all reports' at the top and 'Problems nearby' at the bottom? They seem to take you to the same view of the map, but via a different mechanism. Could this be done the same way for both links?

Sorry for the nitpicks! Those aside, 👍

PR #1433, opened 2016-06-17
Add change email page, and connected purposes

### LINKED ISSUE #360: Allow users to change their own email
Comments from user:

"when post is not anon ... make user name a link ... to all other posts
by same user"

"add user icon"

dracos:
As well as the change email page, add CSRF tokens to more forms; add forgotten password help to general auth page, I imagine it'll be seen more frequently now it's Sign in in the navigation; and slightly nicer when you hit return on an auth page and e.g. FB login is on. Fixes #360

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -139,6 +139,10 @@ sub email_sign_in : Private {
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{facebook_id};
     $token_data->{twitter_id} = $c->session->{oauth}{twitter_id}
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{twitter_id};
+    if ($c->stash->{current_user}) {
+        $token_data->{old_email} = $c->stash->{current_user}->email;
+        $token_data->{r} = 'auth/change_email/success';
+    }
 
     my $token_obj = $c->model('DB::Token')->create({
         scope => 'email_sign_in',
@@ -146,7 +150,8 @@ sub email_sign_in : Private {
     });
 
     $c->stash->{token} = $token_obj->token;
-    $c->send_email( 'login.txt', { to => $good_email } );
+    my $template = $c->stash->{email_template} || 'login.txt';
+    $c->send_email( $template, { to => $good_email } );
     $c->stash->{template} = 'auth/token.html';
 }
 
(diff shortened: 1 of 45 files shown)


perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -444,6 +446,37 @@ sub change_password : Local {
...
+    my ($time, $hash) = split /-/, $token;
+    $c->stash->{csrf_time} = $time;
+    $c->detach('no_csrf_token')
+        unless $token
+            && $time > time() - 3600
+            && $token eq $c->forward('get_csrf_token');
+    delete $c->stash->{csrf_time};
+}
+
+sub no_csrf_token : Private {
```
davea: Should `csrf_time` be deleted from the stash here? Not sure if there are implications to leaving a user-supplied value in place.

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -444,6 +446,37 @@ sub change_password : Local {
...
+    $hash = encode_base64($hash, "");
+    my $token = "$time-$hash";
+    $c->stash->{csrf_token} = $token unless $c->stash->{csrf_time};
+    return $token;
+}
+
+sub check_csrf_token : Private {
+    my ( $self, $c ) = @_;
+
+    my $token = $c->get_param('token') || "";
```
davea: Should we do any validation on `$token`, e.g. max length, correct format etc?

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -182,12 +184,35 @@ sub token : Path('/M') : Args(1) {
...
+        # Were logged in as old_email, want to switch to email ($user)
+        if ($user->in_storage) {
+            my $old_user = $c->model('DB::User')->find({ email => $data->{old_email} });
+            if ($old_user) {
+                foreach (qw(Problem Comment Alert)) {
+                    $c->model("DB::$_")
+                        ->search({ user_id => $user->id })
+                        ->update({ user_id => $old_user->id });
+                }
+                $user->delete;
```
davea: There's a very unlikely crash here, if the user being deleted has some `admin_log` entries.

davea:
A few thoughts:
 - It might be worth sending an email to the old email address letting it know that the FMS account's email address has changed.
 - The wording of the email sent to the new address should more clearly say why the email was sent, what happens when the link is clicked, and what to do if they weren't expecting the email.
 - Merging the list of reports/updates made by two accounts feels a bit weird, but I'm not sure what to suggest instead. Maybe disallowing an existing email address to be used in the 'new address' field?

What do you think about the following scenario:

 1. Alice has made lots of anonymous reports on FMS, all via her account with address `alice@example.org`
 1. Bob suspects lots of the anonymous reports on FMS are from Alice, but can't be sure. He does know her email address, so he creates his *own* account `bob@evil.com`, then puts Alice's address `alice@example.org` in the 'new email address' field.
 1. Through confusion/coercion/a mistake, Alice clicks the link in the email she receives from FMS.
 1. Bob is now logged in as Alice and can see that the anonymous reports were in fact made by her.

Should all existing sessions for the old address be invalidated or some such?

perllib/FixMyStreet/App/Controller/Auth.pm
dracos: It's not used for anything once you're here; it's only deleted in the check in case there's a future call to get_csrf_token (which I don't think there is).

perllib/FixMyStreet/App/Controller/Auth.pm
dracos: Yeah, could do a basic check.

perllib/FixMyStreet/App/Controller/Auth.pm
dracos: Yep, will add, and UserBodyPermission too, ta.

dracos:
Going to have a check that you're logged in as the account that requested the change email, which should alleviate the above and a user missing error when logged in as the to-be-deleted account.

PR #1438, opened 2016-06-21
Remove default box-shadow from .content and aside

### LINKED ISSUE #1419: Unattractive shadows on light themes
Boxes like the "hello" box and the box to the right of the FAQ (there might be others, I'm not sure) have a dark shadow around them that, while looking OK against the dark background of FixMyStreet, in my opinion looks a bit bad against the lighter background:

shadow

I also think it looks bad on FixaMinGata, which has a light grey background.

zarino:
Fixes #1419.

The main change was removing `box-shadow` from `.content` and `.content aside` in `sass/_layout.scss`. All the rest is just removing all the places that cobrands overrode these shadows with their own `box-shadow: none`.

### DIFF
--- a/web/cobrands/oxfordshire/layout.scss
+++ b/web/cobrands/oxfordshire/layout.scss
@@ -37,24 +37,15 @@ body, body a {
   display: none;
 }
 
-// White background, so no shadow or margin needed.
-.content {
+// White background, so no margin needed.
+.content,
+.iel8 .content {
     margin: 0;
-    @include box-shadow(none);
-}
-.iel8 {
-    .content {
-        margin: 0;
-        border: none;
-    }
 }
 
 // Fix location of aside sidebar
 body.twothirdswidthpage {
     .content {
-        aside {
-            @include box-shadow(none);
-        }
         .sticky-sidebar {
             aside {
                 position: fixed;
@@ -111,7 +102,7 @@ body.mappage {
     background-color: #069b01; // picked from header.jpg
     background-position: 100% 40%;
     overflow: auto;
-    box-shadow: 0 0 5px rgba(0,0,0,0.3);
+    @include box-shadow(0 0 5px rgba(0,0,0,0.3));
 
     & > * {
       display: none;
@@ -276,9 +267,7 @@ h4.static-with-rule {
 
 .shadow-wrap {
     ul#key-tools {
-        -webkit-box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
-        -moz-box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
-        box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
+        @include box-shadow(0 0 1em 1em $oxfordshire_very_light_green);
         border-top-width: 2px;
     }
 }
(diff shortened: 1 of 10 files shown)


web/cobrands/oxfordshire/layout.scss
```diff
@@ -276,9 +272,7 @@ h4.static-with-rule {
 
 .shadow-wrap {
     ul#key-tools {
-        -webkit-box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
-        -moz-box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
-        box-shadow: 0em 0px 1em 1em $oxfordshire_very_light_green;
+        @include box-shadow(0 0 1em 1em $oxfordshire_very_light_green);
```
dracos: Any reason you didn't also do this for the box-shadow a few lines further down?

web/cobrands/zurich/layout.scss
```diff
@@ -10,7 +10,6 @@ $mappage-header-height: 7em;
 // White background, so no shadow or margin needed.
 .content {
     color: #3c3c3c;
-    @include box-shadow(none);
 }
 .iel8 {
     .content {
```
dracos: This content is overriding something that has been removed also.

web/cobrands/oxfordshire/layout.scss
```diff
@@ -40,7 +40,6 @@ body, body a {
 // White background, so no shadow or margin needed.
 .content {
     margin: 0;
-    @include box-shadow(none);
 }
 .iel8 {
     .content {
```
dracos: Similarly with the border here.

web/cobrands/zurich/layout.scss
dracos: And just below, body.mappage .content having anything (a box-shadow here) is meaningless now, as it only contains absolutely positioned map + sidebar?

web/cobrands/oxfordshire/layout.scss
zarino: For the buttons? Because the `box-shadow` value for those buttons contained a comma (an inset shadow, and a drop shadow) which `@mixin box-shadow()` couldn't handle.

But since you mentioned it, I've wrapped the `box-shadow` value in quotes, and upgraded the `experimental()` mixin (which `box-shadow()` uses) to handle whatever box-shadow value you happen to throw at it. :-)

web/cobrands/oxfordshire/layout.scss
zarino: Yes! Just spotted another in `sass/_layout.scss` too.

zarino:
@dracos – thanks for the review. I've removed those extraneous styles you spotted. Happy for me to squash and merge?

web/cobrands/oxfordshire/layout.scss
zarino: If we wanted a less hacky solution, and we're sure that FMS is running under Sass 3.2 or greater, we could let `@mixin box-shadow()` accept [a variable number of arguments](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#variable_arguments), and pass the resulting list into `@experimental` where it should just get magically turned into a comma-separated value.

I think. I've never tried it before.

dracos: Yes, due to the `make_css` improvements, FMS is pinned to libsass 3.3.6 which is on par with current sass (3.4.x).

web/cobrands/oxfordshire/layout.scss
```diff
@@ -37,24 +37,14 @@ body, body a {
...
 
-// White background, so no shadow or margin needed.
+// White background, so no margin needed.
 .content {
     margin: 0;
-    @include box-shadow(none);
-}
-.iel8 {
-    .content {
-        margin: 0;
```
dracos: This margin should presumably remain to override the` .iel8 .content` margin from sass.

zarino: Fixed!

web/cobrands/oxfordshire/layout.scss
zarino: Great. I've tried it out, and it works!

dracos:
Yep, I'd squash it all into one commit I think.

PR #1439, opened 2016-06-21
Reinstate missing margin before reporting form email input

### LINKED ISSUE #1418: Padding above grey area in report form when logged in
I noticed that there is no padding above the grey area in the problem reporting form:

fmg

It doesn't look as bad on FixMyStreet, but I think some padding would be nice there too:

fms

Cheers!

zarino:
`#form-box--logged-in-name` has no preceding heading, which meant the
grey box background was colliding with whatever content happened to go
before it.

We add some margin, and while we’re at it, remove the extra margin-top
on first-child labels (unnecessary, since .form-boxes have their own
padding) and tidy up the syntax of the .form-box rule set.

screen shot 2016-06-21 at 09 55 54

Fixes #1418.

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -313,27 +313,41 @@ label{
   margin: 0 -1em 0.25em;
   background: #eee;
   padding: 1em;
-  > input[type=text], input[type=email] {
-    margin-bottom:1em;
+
+  & > input[type=text],
+  & > input[type=email] {
+    margin-bottom: 1em;
   }
+
+  & > label:first-child {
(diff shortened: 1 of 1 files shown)


dracos:
Doesn't the HTML have an ID rather than a class? (So either the CSS needs changing, or the HTML.)

zarino:
@dracos – Good spot! Fixed. Happy for me to merge?

dracos:
Yep!

zarino:
Merged as 3f4948ceab38c184fbcf4521f3ec3351fccacce6.

PR #1446, opened 2016-06-29
[UK Councils] Clarify who receives messages from the contact form.

davea:
The default wording says messages do not go to a council, which is not true for
UK council cobrands.


dracos:
:+1:

PR #1447, opened 2016-06-29
[Zurich] Use new LV95/CH1903+ geocoder and base map tiles

davea:
- Use new `getLocation95` method for geocoding addresses
 - Replace Coordinates::CH1903 with Coordinates::CH1903Plus
 - Refactor Map::Zurich to use WMTSBase instead of duplicating code
 - Slightly refactor map templates to make using custom WMTS tiles easier
 - Use correct sizes for marker pins on maps
 - Disable geocoder cache on staging sites


davea:
This is now ready for review.

perllib/FixMyStreet/Geocode/Zurich.pm
```diff
@@ -67,7 +67,7 @@ sub string {
     my $cache_dir = FixMyStreet->config('GEO_CACHE') . 'zurich/';
     my $cache_file = $cache_dir . md5_hex($s);
     my $result;
-    if (-s $cache_file && -M $cache_file <= 7) {
+    if (-s $cache_file && -M $cache_file <= 7 && !FixMyStreet->config('STAGING_SITE')) {
```
dracos: Want to do this to the general cache function too?

dracos:
:+1:

perllib/FixMyStreet/Geocode/Zurich.pm
davea: Good idea, have done so.

PR #1451, opened 2016-07-01
Make sure map JS works through an admin proxy.

### LINKED ISSUE #1448: Admin: new report edit pages unnecessarily ask for authentication
After logging in to the FixMyStreet admin, the first time I click on an 'edit' link for a report causes my browser to prompt for username and password again (but it happily loads the page behind the dialogue box, and doesn't care if I dismiss the prompt without signing in). Not a browser issue - happens in Firefox 47 and Chrome 49 - and undoubtedly related to the new report editing pages as it only started happening when they went live.

dracos:
Tidy up use of absolute URLs when in the admin. Have the 404 handler spot static files in admin as a fallback. Fixes #1448.

- [x] Output validation_rules manually as separate JS files rather than template.


davea:
Looks good! 👍

PR #1455, opened 2016-07-06
Allow editing of category on admin report edit

### LINKED ISSUE #400: Allow editing of category(body) in admin interface
You should be able to change the category to one of the current categories for the location of the report (and if necessary, it then updates the body too to match).

Also include moving of pin?

dracos:
Might fix #400.


davea:
As mentioned in person, `problem->extra` isn't touched when changing the category which is probably fine but worth noting in case of future weirdness.

👍

PR #1456, opened 2016-07-08
[Bristol] Only show Open311 categories on Bristol cobrand

davea:
This adds a new Cobrand::Default::categories_restriction method that can be used
to limit what categories are available when making a new report on a per-cobrand
basis.

This is used by Cobrand::Bristol to only show Open311 categories when making a
new report on fixmystreet.bristol.gov.uk but allowing all categories on FMS.com.

Also removes the Hart::hidden_categories and uses the new categories_restriction
method to achieve the same thing.


dracos:
Nice bit of tidying :)

Issue #1278, opened 2015-10-29
Add option to include report images as attachments in body report email


dracos:
Zurich has code to attach images to the body report email at https://github.com/mysociety/fixmystreet/blob/a577e6cead169a5f28daf020ebbc86de6fbac6c2/perllib/FixMyStreet/Cobrand/Zurich.pm#L1012-L1036 – generalise this so that any cobrand can do it, via a configuration/cobrand variable.

jonkri:
+1

jonkri:
Like with #72, this would not just be useful when we have HTML e-mail, but also right now with text mail.

jonkri:
It's very nice to see progress on this! :)

What's the current status of this issue and #72, and HTML mail in general?

We will want to use HTML mail in FixaMinGata, as it looks better and makes printing of attachments easier than with text mails (at least under certain circumstances).

Cheers!

dracos:
This has been done as part of HTML email (though it does only include the first image of the report).

jonkri:
@dracos: I'm curious... How difficult would it be to attach a full version of the uploaded problem pictures in the report sent to the body? Thanks!

dracos:
The Zurich cobrand attaches the full size images using the `munge_sendreport_params` cobrand function, which I assume works! You could fetch different size versions by doing something similar, I imagine.

PR #1458, opened 2016-07-12
HTML emails

### LINKED ISSUE #1103: Follow-up "questionnaire" email could be much better
The email currently:

* Doesn't tell you where the problem was
* Doesn't show you the picture you uploaded with the report
* Requires you to load a page in your web browser and complete a whole form

The email *could* instead:

* Show you the location of the problem on a map (see https://github.com/mysociety/fixmystreet/commit/caa28b8701ff8de8d98ae98fbf133db0be8141c8 for example of embedding map in email)
* Show you your picture of the report, so you can differentiate it from others you might have reported around the same time
* Give you simple "Yes" and "No" buttons (like WriteToThem)

### LINKED ISSUE #1281: Consider HTML emails for various emails
Confirmation emails could be made clearer (do we want map/photo there, or too much weight?), questionnaire emails (see #1103), report emails to body (optional image #1278, map image?) - attachments as well as HTML version, possibly others.

zarino:
Fixes #1281. Fixes #1103. Provides one image in email, is that enough for #1278?

To do:

- [x] HTML markup and inline CSS styling
- [x] Default styling without FixMyStreet.com logo
- [x] Use site_name instead of "FixMyStreet" in templates
- [x] Create a new template for each of the required types of email 
   - [x] submit missing "more appropriate" message
- [x] Wire up the templates with real data
   - [x] Except alert emails which need some more server side work as they compose too much info in the code rather than the template
   - [x] Plural strings in some HTML email summaries as they involve number - reword?
- [x] Work out how to attach a map image, logo, report photo to the emails
   - [x] Create endpoint that returns static map for a report
   - [x] Generic inline imaging, done for map image + email logo
   - [x] Report photo (first definitely; all of them?)
   - [x] Map image only attached to online emails, questionnaire/submit lacking them
- [x] Testing that HTML part has the same link as text part
- [x] Cross-client testing in Litmus of confirmation email
- [ ] Manual testing of other emails (alert confirm, update confirm, area alert, update alert, change email, contact, login, moderated, questionnaire, submit)
- [ ] Do we mind HTML versions not having links to FB/Twitter/FAQ that text footer does

Screenshot of latest version:
screen shot 2016-07-20 at 17 50 41

screen shot 2016-07-21

Example of a council theme:
screen shot 2016-07-21 at 14 30 54


dracos:
I've added a temporary mechanism so that /about/email?id=N puts report N in the template as the `problem` variable.

The list of possible emails is:

1. Confirmation: alert-confirm.txt problem-confirm-not-sending.txt problem-confirm.txt update-confirm.txt
1. Questionnaire: questionnaire.txt
1. Email sent to council: submit.txt
1. Alerts (basically all the same, update might be slightly different): alert-problem-area.txt alert-problem-council.txt alert-problem-nearby.txt alert-problem-ward.txt alert-problem.txt alert-update.txt
1. Generic login: login.txt
1. problem-moderated.txt (sent to user when council moderates a report)
1. Change your email: change_email.txt
1. Emails sent as an auto-reply: reply-autoresponse
1. Contact form submission: contact.txt (comes to us, so probably not important)
1. Partial: partial.txt (not really used, not important)
1. Report sent: confirm_report_sent.txt (only used on SeeSomething so not important)

zarino:
Part way through Litmus testing. Here’s how it’s looking so far: https://litmus.com/checklist/public/8c3861b

Two main things we need to fix:

- [ ] Use an `` tag for the map, rather than a background image and padding
- [ ] Solve the super-narrow columns on Gmail native apps

templates/email/default/problem-confirm-not-sending.html
```diff
@@ -0,0 +1,33 @@
...
+%]
+
+<th style="[% td_style %][% primary_column_style %]" id="primary_column">
+  <h1 style="[% h1_style %]">Please confirm your&nbsp;report</h1>
+  <p style="[% p_style %]">Please click on the link below to confirm that you want your report to appear
+on [% site_name %], despite not being sent to the
+council.</p>
+  <p style="[% p_style %]">Your report will also appear on the [% site_name %] website.</p>
+  <p style="margin: 20px auto; text-align: center">
+    <a style="[% button_style %]" href="[% token_url %]">Yes, send my report</a>
```
davea: This button should say 'Publish my report' or similar, as it's not being sent anywhere.

dracos:
These emails currently only include the first image, inline; should they include all the images the user added to the report (perhaps only the submit email to the council? As other cases are just to help you know which report it is so one probably enough).

Other than that question,I think this is good to go with probably quite a bit more testing of the emails (not so much the visual as that's shared though that too, but more that each type actually has the right contents and works).

templates/email/default/alert-problem-ward.html
```diff
@@ -0,0 +1,19 @@
...
+email_columns = 1;
+
+PROCESS '_email_settings.html';
+
+INCLUDE '_email_top.html';
+
+%]
+
+<th style="[% td_style %][% only_column_style %]">
+    <h1 style="[% h1_style %]">New reports to [% area_name %] within {% ward_name %]</h1>
```
zarino: `{% ward_name %]` here should be `[% ward_name %]`

davea:
The formatting at the bottom of emails from the contact page is a bit bunched up:
screen shot 2016-07-21 at 12 00 41

zarino:
A few things I spotted this morning – not sure whether they're expected behaviour, or bugs. Perhaps @davea could investigate?

* Emails about [updates on a report](https://litmus.com/checklist/public/f66070b) should have the 2-column layout, with the update(s) in the primary column, and the original report details and map in the secondary column.
* Lots of the emails were lacking the official FixMyStreet logo – I assume because they aren't initiated by user action, and therefore don't know which cobrand’s logo to use? eg: [Email login](https://litmus.com/pub/f7645b6) / [Email change](https://litmus.com/pub/76a543f) / [Report moderation](https://litmus.com/pub/659187c) / [Contact form](https://litmus.com/pub/92d327a) / [Questionnaire](https://litmus.com/pub/46b1d8d)

dracos:
Have fixed contact email (except the links aren't clickable, fixing).

Two column alert-update hasn't previously existed, if it does, then that should work fine ;) I think logo is working okay, a lot of the test emails come from the default cobrand. Offline emails should use the cobrand that the report/update/alert was created on, so should be fine.

dracos:
Alert email now two column and pulling in map/photo okay.

PR #1463, opened 2016-07-18
Use normal authentication for controlling access to admin

davea:
This lays the groundwork for controlling individual features in the admin on a more fine-grained level with the UserBodyPermission model.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -49,12 +49,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # We want to be on port 3000 for development
         sed -i -r -e "s,^( *BASE_URL: .*)',\\1:3000'," fixmystreet/conf/general.yml
     fi
+    # Create a superuser for the admin
+    fixmystreet/bin/createsuperuser superuser@example.org password
     if [ $SUCCESS -eq 0 ]; then
         # All done
         echo "****************"
         echo "You can now ssh into your vagrant box: vagrant ssh"
         echo "The website code is found in: ~/fixmystreet"
         echo "You can run the dev server with: script/fixmystreet_app_server.pl [-d] [-r] [--fork]"
+        echo "Access the admin with username: superuser@example.org and password: password"
     else
         echo "Unfortunately, something appears to have gone wrong with the installation."
         echo "Please see above for any errors, and do ask on our mailing list for help."
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -75,7 +75,7 @@ sub index : Path : Args(0) {
 
     $c->forward('stats_by_state');
 
-    my @unsent = $c->model('DB::Problem')->search( {
+    my @unsent = $c->cobrand->problems->search( {
         state => [ 'confirmed' ],
         whensent => undef,
         bodies_str => { '!=', undef },
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -244,13 +244,15 @@ sub bodies : Path('bodies') : Args(0) {
         $c->forward('/auth/check_csrf_token');
 
         my $params = $c->forward('body_params');
-        my $body = $c->model('DB::Body')->create( $params );
-        my @area_ids = $c->get_param_list('area_ids');
-        foreach (@area_ids) {
-            $c->model('DB::BodyArea')->create( { body => $body, area_id => $_ } );
-        }
+        unless ( keys $c->stash->{body_errors} ) {
+            my $body = $c->model('DB::Body')->create( $params );
+            my @area_ids = $c->get_param_list('area_ids');
+            foreach (@area_ids) {
+                $c->model('DB::BodyArea')->create( { body => $body, area_id => $_ } );
+            }
 
-        $c->stash->{updated} = _('New body added');
+            $c->stash->{updated} = _('New body added');
+        }
     }
 
     $c->forward( 'fetch_all_bodies' );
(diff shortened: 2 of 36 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -32,10 +32,16 @@ sub begin : Private {
 
     $c->uri_disposition('relative');
 
-    if ( $c->cobrand->moniker eq 'zurich' || $c->cobrand->moniker eq 'seesomething' ) {
-        $c->detach( '/auth/redirect' ) unless $c->user_exists;
-        $c->detach( '/auth/redirect' ) unless $c->user->from_body;
+    # Must be logged in to see the admin, as a body user or a super user
+    unless ( $c->user_exists && ( $c->user->from_body || $c->user->is_superuser )) {
```
dracos: I think the default should perhaps (for now at least) be superuser only, with the cobrand allowing more if it wants (can look again once permissions are all up and set up I guess, and an admin interface for administering them exists :) ).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -73,7 +79,7 @@ sub index : Path : Args(0) {
 
     $c->forward('stats_by_state');
 
-    my @unsent = $c->model('DB::Problem')->search( {
+    my @unsent = $c->cobrand->admin_problems->search( {
```
dracos: Whoops, thanks!

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -392,5 +392,21 @@ sub link_to_council_cobrand {
...
 }
 
+=head1 admin_problems
+
+Further restricts the results of self->problems to those which match the current
+user's from_body, unless the user is a superuser.
+
+=cut
+
+sub admin_problems {
```
dracos: If you're on a council cobrand admin, you're already restricted to the `to_body` details. If you're on fixmystreet.com's admin, then I think that should probably be superuser only (which would be done by the previous change I imagine). If this is kept, I guess it should be in FixMyStreet.pm, but I'm not sure it's currently needed.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -295,7 +303,7 @@ sub body : Path('body') : Args(1) {
 
     $c->stash->{body_id} = $body_id;
 
-    $c->forward( 'check_for_super_user' );
+    $c->forward( 'check_for_super_user' ) if $c->cobrand->moniker eq 'zurich';
```
dracos: I think this means (currently) any from_body user could edit any body by knowing the URL. If above changes are made, then I think this still needs changing to make sure the body being dealt with is the appropriate one.

dracos:
Looks good :) Few changes to the defaults, nothing major.

We should add a 'createsuperuser' command I guess, and make sure this is clearly documented so people don't lose their admin access when they upgrade :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -313,7 +321,12 @@ sub body : Path('body') : Args(1) {
 
 sub check_for_super_user : Private {
     my ( $self, $c ) = @_;
-    if ( $c->cobrand->moniker eq 'zurich' && $c->stash->{admin_type} ne 'super' ) {
+
+    my $superuser = $c->user->is_superuser;
+    # Zurich currently has its own way of defining superusers
+    $superuser ||= $c->cobrand->moniker eq 'zurich' && $c->stash->{admin_type} eq 'super';
+
+    unless ( $superuser ) {
```
dracos: I guess this should technically be a 403?

t/cobrand/seesomething.t
```diff
@@ -17,7 +17,7 @@ $db->txn_begin;
 $db->resultset('Comment')->delete;
 $db->resultset('Problem')->delete;
 
-my $user = $mech->create_user_ok( $EMAIL );
+my $user = $mech->create_user_ok( $EMAIL, is_superuser => 1 );
```
dracos: Guess this user could also be a from_body user to test that, perhaps.

perllib/FixMyStreet/Script/CreateSuperuser.pm
```diff
@@ -7,10 +7,17 @@ use FixMyStreet;
...
 sub createsuperuser {
-    die "Specify a single email address to create a superuser or grant superuser status to." if (@ARGV != 1);
+    die "Specify a single email address and optionally password to create a superuser or grant superuser status to." if (@ARGV < 1 || @ARGV > 2);
 
-    my $user = FixMyStreet::DB->resultset('User')->find_or_create({ email => $ARGV[0] });
-    $user->update({ is_superuser => 1 });
+    my $user = FixMyStreet::DB->resultset('User')->find_or_new({ email => $ARGV[0] });
+    if ( !$user->in_storage ) {
+        die "Specify a password for this new user." if (@ARGV < 2);
+        $user->password($ARGV[0]);
```
dracos: This sets the password to the email :)

Vagrantfile
```diff
@@ -49,12 +49,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # We want to be on port 3000 for development
         sed -i -r -e "s,^( *BASE_URL: .*)',\\1:3000'," fixmystreet/conf/general.yml
     fi
+    # Create a superuser for the admin
+    cd fixmystreet
+    ./bin/createsuperuser superuser@example.org password
+    cd ..
```
dracos: Probably better as `(cd fixmystreet && bin/createsuperuser superuser@example.org password)`.
Although, shouldn't `fixmystreet/bin/createsuperuser superuser@example.org password` work?

perllib/FixMyStreet/Cobrand/SeeSomething.pm
```diff
@@ -60,6 +60,12 @@ sub allow_anonymous_reports { 1; }
 
 sub anonymous_account { return { name => 'Anonymous Submission', email => FixMyStreet->config('DO_NOT_REPLY_EMAIL') }; }
 
+# Any user with from_body set can view admin
+sub admin_allow_user {
+    my ( $self, $user ) = @_;
+    return 1 if $user->from_body;
```
dracos: Or superuser?

PR #1467, opened 2016-07-21
Refactor map pin/feature functions.

dracos:
Sorry for not doing this sooner (though would have missed /my then anyway!).


davea:
Nice bit of tidying 👍

PR #1470, opened 2016-08-02
Add 'inspect report' page

davea:
As part of the FixMyStreet for Councils work, this PR adds a new page that allows a council inspector to view and manage a problem by adding detailed notes, setting its priority etc.

This is still a work-in-progress, but I'm opening this PR to allow us to keep track of the page changes.

See mysociety/fixmystreetforcouncils#22, mysociety/fixmystreetforcouncils#39.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -764,12 +764,7 @@ sub report_edit : Path('report_edit') : Args(1) {
         }
         $problem->set_inflated_columns(\%columns);
 
-        if ((my $category = $c->get_param('category')) ne $problem->category) {
-            $problem->category($category);
-            my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
-            my $bs = join( ',', map { $_->body_id } @contacts );
-            $problem->bodies_str($bs);
-        }
+        $c->forward( '/admin/report_edit_category', [ $problem ] );
 
         if ( $c->get_param('email') ne $problem->user->email ) {
             my $user = $c->model('DB::User')->find_or_create(
@@ -810,6 +805,70 @@ sub report_edit : Path('report_edit') : Args(1) {
         $problem->discard_changes;
     }
 
+    return 1;
+}
+
+=head2 report_edit_category
+
+Handles changing a problem's category and the complexity that comes with it.
+
+=cut
+
+sub report_edit_category : Private {
+    my ($self, $c, $problem) = @_;
+
+    # TODO: It's possible to assign a category belonging to a district
+    # council, meaning a 404 when the page is reloaded because the
+    # problem is no longer included in the current cobrand's
+    # problem_restriction.
+    # See mysociety/fixmystreetforcouncils#44
+    # We could
+    #  a) only allow the current body's categories to be chosen,
+    #  b) show a warning about the impending change of body
+    #  c) bounce the user to the report page on fms.com
+    # Not too worried about this right now, as it forms part of a bigger
+    # concern outlined in the above ticket and
+    # mysociety/fixmystreetforcouncils#17
+    if ((my $category = $c->get_param('category')) ne $problem->category) {
+        $problem->category($category);
+        my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
+        my $bs = join( ',', map { $_->body_id } @contacts );
+        $problem->bodies_str($bs);
+    }
+}
+
+=head2 report_edit_location
+
+Handles changing a problem's location and the complexity that comes with it.
+For now, we reject the new location if the new location and old locations aren't
+covered by the same body.
+
+Returns 1 if the new position (if any) is acceptable, undef otherwise.
+
+NB: This must be called before report_edit_category, as that might modify
+$problem->bodies_str.
+
+=cut
+
+sub report_edit_location : Private {
+    my ($self, $c, $problem) = @_;
+
+    return 1 unless $c->forward('/report/new/determine_location');
+
+    if ( $c->stash->{latitude} != $problem->latitude || $c->stash->{longitude} != $problem->longitude ) {
+        delete $c->stash->{prefetched_all_areas};
+        delete $c->stash->{all_areas};
+        delete $c->stash->{fetch_all_areas};
+        delete $c->stash->{all_areas_mapit};
+        $c->forward('/council/load_and_check_areas');
+        $c->forward('/report/new/setup_categories_and_bodies');
+        my %allowed_bodies = map { $_ => 1 } @{$problem->bodies_str_ids};
+        my @new_bodies = @{$c->stash->{bodies_to_list}};
+        my $bodies_match = grep { exists( $allowed_bodies{$_} ) } @new_bodies;
+        return unless $bodies_match;
+        $problem->latitude($c->stash->{latitude});
+        $problem->longitude($c->stash->{longitude});
+    }
     return 1;
 }
 
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -278,16 +278,108 @@ sub delete :Local :Args(1) {
     return $c->res->redirect($uri);
 }
 
-sub map : Path('') : Args(2) {
-    my ( $self, $c, $id, $map ) = @_;
+=head2 action_router
 
-    $c->detach( '/page_error_404_not_found', [] ) unless $map eq 'map';
+A router for dispatching handlers for sub-actions on a particular report,
+e.g. /report/1/inspect
+
+=cut
+
+sub action_router : Path('') : Args(2) {
+    my ( $self, $c, $id, $action ) = @_;
+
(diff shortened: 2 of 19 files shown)


zarino:
Currently looks like this (in the Oxfordshire cobrand):

screen shot 2016-08-02 at 11 07 31

Because FMS doesn't use `box-size: border-box` by default, and the `.with-actions` sidebar is based on fixed widths, the gap between the two columns increases when you have invisible scroll bars – which isn't ideal, but probably outside the scope of this PR to fix:

screen shot 2016-08-02 at 11 11 28

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -278,17 +278,104 @@ sub delete :Local :Args(1) {
...
+            $c->forward( '/admin/log_edit', [ $id, 'problem', 'state_change' ] );
+        }
+        if ((my $category = $c->get_param('category')) ne $problem->category) {
+            $problem->category($category);
+            my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
+            my $bs = join( ',', map { $_->body_id } @contacts );
+            $problem->bodies_str($bs);
+        }
+
+        $problem->update;
```
zarino: https://github.com/mysociety/fixmystreetforcouncils/issues/22 suggests that the name and contact details for the inspector also need to be saved, as part of the Defect that gets sent to the contactor / work crew. Will that be available automatically, or do we need to add it as an attribute of the problem before we save?

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -278,17 +278,104 @@ sub delete :Local :Args(1) {
...
+        }
+
+        # Handle the state changing
+        my $old_state = $problem->state;
+        $problem->state($c->get_param('state'));
+        if ( $problem->is_visible() and $old_state eq 'unconfirmed' ) {
+            $problem->confirmed( \'current_timestamp' );
+        }
+        if ( $problem->state eq 'hidden' ) {
+            $problem->get_photoset->delete_cached;
```
dracos: Fine, just wondering if we should wrap state setting so that this guaranteed gets called whenever problem state is set, or whether that might introduce edge bugs.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -278,17 +278,104 @@ sub delete :Local :Args(1) {
...
+        }
+        if ( $problem->state ne $old_state ) {
+            $c->forward( '/admin/log_edit', [ $id, 'problem', 'state_change' ] );
+        }
+        if ((my $category = $c->get_param('category')) ne $problem->category) {
+            $problem->category($category);
+            my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
+            my $bs = join( ',', map { $_->body_id } @contacts );
+            $problem->bodies_str($bs);
```
dracos: This code looks to match admin's, so we should factor it out somewhere, I guess a model method on Problem?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -365,6 +365,41 @@ $.extend(fixmystreet.set_up, {
...
+                } else { // Unknown
+                    $link.html(translation_strings.geolocation_unknown);
+                }
+            }, {
+                enableHighAccuracy: true,
+                timeout: 10000
+            });
+        });
+    }
+  },
```
dracos: Can we factor these two together, or hoping gzip will handle most of it? ;)

templates/web/base/report/_main.html
```diff
@@ -88,7 +86,20 @@ <h1 class="moderate-display">[% problem.title | html %]</h1>
...
     </div>
+  [% END %]
+
+  [% IF can_moderate OR can_inspect %]
+    <p class="moderate-display">
+      [% IF can_moderate %]
+        <input type="button" class="btn moderate" value="moderate">
+      [% END %]
+      [% IF can_inspect %]
+        <a href="/report/[% problem.id %]/inspect" class="btn inspect">Inspect</a>
```
dracos: This has a different hover effect to when it is a button.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -278,17 +278,104 @@ sub delete :Local :Args(1) {
...
+to the current Problem in $c->stash->{problem}. Shows the 403 page if not.
+
+=cut
+
+sub check_has_permission_to : Private {
+    my ( $self, $c, $permission ) = @_;
+
+    my $bodies = $c->stash->{problem}->bodies_str;
+
+    $c->detach('/page_error_403_access_denied', [])
```
dracos: You need to supply an error message here, otherwise it's a bit odd getting the front page on its own :)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -365,6 +365,41 @@ $.extend(fixmystreet.set_up, {
     }
   },
 
+  report_inspect_geolocation: function() {
+    if (geo_position_js.init() && $('body.report_inspect').length) {
```
dracos: This class doesn't exist.

templates/web/base/report/_main.html
dracos: The button is present even if we're on the inspect page.

templates/web/base/report/_inspect.html
```diff
@@ -0,0 +1,77 @@
...
+        <p class="right">
+          <label for="problem_priority">Priority:</label>
+          <select name="priority" id="problem_priority">
+            <option value="" [% 'selected' UNLESS problem.get_extra_metadata('priority') %]>-</option>
+            [% FOREACH priority IN priorities %]
+              <option value="[% priority.key %]" [% 'selected' IF problem.get_extra_metadata('priority') == priority.key %]>[% priority.value %]</option>
+            [% END %]
+          </select>
+        </p>
+        <p class="left">
```
dracos: This one doesn't clear the report ID in the FMS cobrand at least.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -278,17 +278,104 @@ sub delete :Local :Args(1) {
...
+    $c->detach( '/page_error_404_not_found', [] );
+}
+
+sub inspect : Private {
+    my ( $self, $c, $id ) = @_;
+
+    $c->forward('/auth/get_csrf_token');
+    $c->forward( 'load_problem_or_display_error', [ $id ] );
+    $c->forward( 'check_has_permission_to', [ 'report_inspect' ] );
+    $c->forward( 'format_problem_for_display' );
```
dracos: Not sure if this is deliberate, updates aren't listed on the inspect page, but the update form still is. Guess it should be both, or neither.

templates/web/base/report/display.html
```diff
@@ -45,6 +45,8 @@
 [% END %]
 
   </div>
+
+  [% TRY %][% PROCESS second_column %][% CATCH file %][% END %]
```
dracos: Just to avoid the ugly TRY, you could have `[% second_column %]` here, and `[% second_column = BLOCK %]` in the other file.

templates/web/base/report/_inspect.html
```diff
@@ -0,0 +1,77 @@
+[% PROCESS 'admin/report_blocks.html'; # For the report state dropdown %]
+[% BLOCK second_column -%]
+  <div id="side-report-secondary">
+    <div class="problem-inspector-fields clearfix">
+      <form id="report_inspect_form" method="post" action="[% c.uri_for( '/report', problem.id, 'inspect' ) %]">
+        <p class="left">
+          <label for="problem_id">Report ID:</label>
```
dracos: Should mark all these for translation, if it's in base, I guess.

dracos:
Looking good :)

davea:
I've addressed all the points you raised @dracos, so if you get a chance for a re-review we can hopefully get this one merged.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -860,13 +860,19 @@ sub report_edit_location : Private {
 
     if ( $c->stash->{latitude} != $problem->latitude || $c->stash->{longitude} != $problem->longitude ) {
         $c->log->debug("lat/lon have changed");
+        $c->log->debug("stash->{bodies_to_list}: " . Dumper($c->stash->{bodies_to_list}));
+        delete $c->stash->{prefetched_all_areas};
+        delete $c->stash->{all_areas};
+        delete $c->stash->{fetch_all_areas};
+        delete $c->stash->{all_areas_mapit};
```
dracos: Hmm, do we need a nicer way of causing a 'fresh' lookup in some way?
Should it check whether load_and_check_areas errored at all (e.g. drag pin outside area entirely, I guess?)

Some debug stuff left in around here.

dracos:
The E/N updating doesn't work on fixmystreet.com cobrand as the OSGB JS isn't loaded, is that wanted?
The "Inspect" button is still there when you're already on the inspect screen, and still has a different hover from a normal button.
One inline comment, think that's it :)

Issue #881, opened 2014-08-28
Store and use zoom level when making report


dracos:
This would mean that future visitors to the problem report page would see the same thing the reporter saw when they made their report. This would be a way to fix the issue discussed in #875.

dracos:
As report pages are generally pulled in via ajax now, this wouldn't be used that often.

PR #1476, opened 2016-08-08
[UK] Record map zoom when clicking map to make new report

zarino:
Discussion around https://github.com/mysociety/fixmystreetforcouncils/issues/41 raised the question of what zoom level people are creating reports from.

This would be super handy data to collect.

We already track map clicks by sending a fake `/map_click` pageview to Google Analytics. Really this should be an event rather than a fake pageview – but we've got a goal already set up based on it being a pageview, so I'm introducing the event *alongside* the pageview. This way, we can record them both in tandem for a while, and then, later, change the Goal to monitor for the event, and remove the fake pageview.

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -746,6 +746,12 @@ OpenLayers.Control.Click = OpenLayers.Class(OpenLayers.Control, {
 
         if ( typeof ga !== 'undefined' && window.cobrand == 'fixmystreet' ) {
             ga('send', 'pageview', { 'page': '/map_click' } );
+            ga('send', {
+                hitType: 'event',
+                eventCategory: 'Map',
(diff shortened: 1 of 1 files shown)


zarino:
@dracos Does this look ok to you?

dracos:
That stores the zoom level at the point the map is clicked, not the zoom level at the point the report is made (they may zoom and drag the pin after the click, to reposition the pin more accurately you'd hope). You'd need to attach the event to the point the form is submitted (and only if it were successfully submitted), so on the check your email (not logged in), report confirmed (logged in) pages somehow. If it doesn't matter about multiple events, then the easier solution would be to attach it to pin drag and zoom changes as well as map clicks.

dracos:
Also, links with #881 (storing the zoom level server side).

PR #1477, opened 2016-08-09
Test tweaks

dracos:
This lets you run the test suite twice at once (postgres creation already handled this, auto using a different port, but it was using the same config YAML file, now unique per process ID).
Also fix a warning in Zurich test.


davea:
👍

PR #1482, opened 2016-08-10
Add view of planned list and add/remove buttons

dracos:
Does what it says on the tin.

Todo:
- [x] Decision on where to go post add/remove. I think going to list of planned reports upon removal is fine, not sure best place to go when adding. See point below.
- [x] Could use JS to have add/remove buttons work whilst staying on same page, would this be nice? Or then extra click to go back somewhere. But on adding, could be better as then you can go back to list of reports.

### DIFF
--- a/templates/web/oxfordshire/header.html
+++ b/templates/web/oxfordshire/header.html
@@ -42,6 +42,12 @@
                             <[% IF c.req.uri.path == '/my' OR ( c.req.uri.path == '/auth' AND c.req.params.r == 'my' ) %]span[% ELSE %]a href="/my"[% END
                              %]>[% c.user_exists ? loc("Your account") : loc("Sign in") %]</[% ( c.req.uri.path == '/my' OR ( c.req.uri.path == '/auth' AND c.req.params.r == 'my' ) ) ? 'span' : 'a' %]>
                         </li>
+                        [% IF c.user_exists AND c.user.has_permission_to('planned_reports', c.user.from_body.id) %]
+                        <li>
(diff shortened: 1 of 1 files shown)


zarino:
> decision on where to go post add/remove

I would suggest not going anywhere. As you mention, maybe using javascript to submit the form asynchronously, or maybe just submitting to the current page – either way you click the button, the report gets updated, and you are left on the same report page, except the "add to planned" button has now become "remove from planned".

Once a report has been added to the planner, it might be nice to give them a link they can follow to view their planner – either in the button itself (thus, a "planned" report would have two visible buttons: "Show in planner", and "Remove from planner") or as a temporary "This has been added to your planner – go visit it" sort of flash message.

dracos:
Added the JS (works nicely), and redirect goes back to the same page. Haven't added "Show planned reports" button as I wasn't sure if there might be a permanent place for this somewhere to be shown if the user has the right permission.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -16,17 +18,38 @@ Catalyst Controller.
...
+        { user_id => $c->user->id });
+    $c->forward('get_problems');
+    $c->forward('get_updates');
+    $c->forward('setup_page_data');
+}
+
+sub planned : Local : Args(0) {
+    my ( $self, $c ) = @_;
+
+    # TODO Only with permission to add/remove planned reports
```
davea: The `planned_reports` permission is available for this.

davea:
Looks awesome! 🚢

PR #1484, opened 2016-08-10
Fix "Problems nearby" link on /my and /reports.

dracos:
Problems nearby should load the /around page from those pages, rather
than act like "Back to all reports" but changing to an incorrect URL.


zarino:
👍

PR #1486, opened 2016-08-11
Add permissions restrictions/management to admin

davea:
- Adds `report_edit` and `user_edit` permissions which are required to edit reports or users via the admin
 - Cobrand admin doesn't allow editing of users from other cobrands
 - Setting a user's `from_body` is a simple 'staff' tickbox for non-superusers, which sets `from_body` to the same as the logged-in user.
 - Individual permissions can be granted/revoked via the user edit form, if the logged-in user has the `user_manage_permissions` permission.

For mysociety/fixmystreetforcouncils#35

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -920,7 +920,7 @@ sub users: Path('users') : Args(0) {
         my $search_n = 0;
         $search_n = int($search) if $search =~ /^\d+$/;
 
-        my $users = $c->model('DB::User')->search(
+        my $users = $c->cobrand->users->search(
             {
                 -or => [
                     email => { ilike => $isearch },
@@ -952,7 +952,7 @@ sub users: Path('users') : Args(0) {
         $c->forward('fetch_all_bodies');
 
         # Admin users by default
-        my $users = $c->model('DB::User')->search(
+        my $users = $c->cobrand->users->search(
             { from_body => { '!=', undef } },
             { order_by => 'name' }
         );
@@ -1120,7 +1120,13 @@ sub user_edit : Path('user_edit') : Args(1) {
 
     $c->forward('/auth/get_csrf_token');
 
-    my $user = $c->model('DB::User')->find( { id => $id } );
+    my $user = $c->cobrand->users->find( { id => $id } );
+    $c->detach( '/page_error_404_not_found' ) unless $user;
+
+    unless ( $c->user->is_superuser || ( $c->user->has_permission_to('user_edit', $c->user->from_body->id) ) ) {
+        $c->detach('/page_error_403_access_denied', []);
+    }
+
     $c->stash->{user} = $user;
 
     $c->forward('fetch_all_bodies');
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1148,10 +1148,18 @@ sub user_edit : Path('user_edit') : Args(1) {
         $user->name( $c->get_param('name') );
         $user->email( $c->get_param('email') );
         $user->phone( $c->get_param('phone') ) if $c->get_param('phone');
-        $user->from_body( $c->get_param('body') || undef );
         $user->flagged( $c->get_param('flagged') || 0 );
         # Only superusers can grant superuser status
         $user->is_superuser( ( $c->user->is_superuser && $c->get_param('is_superuser') ) || 0 );
+        # Superusers can set from_body to any value, but other staff can only
+        # set from_body to the same value as their own from_body.
+        if ($c->user->is_superuser) {
+            $user->from_body( $c->get_param('body') || undef );
+        } elsif ($c->get_param('body') eq $c->user->from_body->id) {
+            $user->from_body( $c->user->from_body );
+        } else {
+            $user->from_body( undef );
+        }
 
         unless ($user->email) {
             $c->stash->{field_errors}->{email} = _('Please enter a valid email');
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -669,6 +669,13 @@ sub report_edit : Path('report_edit') : Args(1) {
     $c->detach( '/page_error_404_not_found' )
       unless $problem;
 
+    unless (
+        $c->cobrand->moniker eq 'zurich'
+        || $c->user->has_permission_to(report_edit => $problem->bodies_str)
+    ) {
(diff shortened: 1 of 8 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -669,6 +669,13 @@ sub report_edit : Path('report_edit') : Args(1) {
     $c->detach( '/page_error_404_not_found' )
       unless $problem;
 
+    unless (
+        $c->cobrand->moniker eq 'zurich'
+        || $c->user->has_permission_to(report_edit => $problem->bodies_str)
+    ) {
+        $c->detach( '/page_error_404_not_found' );
```
dracos: A 403 perhaps?

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -637,7 +637,36 @@ List of names of pages to display on the admin interface
...
+         'update_edit' => [undef, undef],
+         'abuse_edit'  => [undef, undef],
+    };
+
+    # There are some pages that only super users can see
+    if ( $user->is_superuser ) {
+        $pages = {
+            %$pages,
+            'config' => [ _('Configuration'), 8],
+        }
```
dracos: This could be `$pages->{config} = [ _('Configuration'), 8];`

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -637,7 +637,36 @@ List of names of pages to display on the admin interface
 
 =cut
 
-sub admin_pages { 0 }
+sub admin_pages {
```
dracos: If we're having this in the Default, do we get rid of the 'default' in set_allowed_pages?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -50,6 +50,13 @@ sub updates_restriction {
     return $rs->to_body($self->council_id);
 }
 
+sub users_restriction {
+    my ($self, $rs) = @_;
+    # Council admins can only edit users who are members of the public or that
+    # same council, i.e. not users from other councils.
+    return $rs->search( { from_body => [ undef, $self->council_id ] } );
```
dracos: I think this means a from_body user can search for any non-body user.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -658,6 +658,36 @@ sub admin_allow_user {
     return 1 if $user->is_superuser;
 }
 
+=head2 available_permissions
+
+Grouped lists of permission types available for use in the admin
```
dracos: Perhaps note here which of these permissions are currently in use and which are future? e.g. report_edit_category doesn't appear to be used anywhere.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1149,14 +1155,33 @@ sub user_edit : Path('user_edit') : Args(1) {
...
             $user->from_body( $c->user->from_body );
         } else {
             $user->from_body( undef );
         }
 
+        if (!$user->from_body) {
+            # Non-staff users aren't allowed any permissions
+            $user->user_body_permissions->delete_all;
+        } elsif ($c->stash->{available_permissions} && $c->user->has_permission_to('user_manage_permissions', $user->from_body->id)) {
+            my @all_permissions = map { map { $_ } keys $_ } values $c->stash->{available_permissions};
```
dracos: Not sure what the inner map is doing here :)

dracos: Plus keys on a reference, should be `%$_`

dracos: Oh, and the values one too!

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1123,8 +1123,14 @@ sub user_edit : Path('user_edit') : Args(1) {
     my $user = $c->cobrand->users->find( { id => $id } );
     $c->detach( '/page_error_404_not_found' ) unless $user;
 
+    unless ( $c->user->is_superuser || ( $user->from_body && $c->user->has_permission_to('user_edit', $user->from_body->id) ) ) {
+        $c->detach('/page_error_403_access_denied', []);
+    }
+
     $c->stash->{user} = $user;
 
+    $c->stash->{available_permissions} = $c->cobrand->available_permissions;
```
dracos: If this is only set if the user_manage_permissions permission exists, that simplifies the other checks perhaps.

t/app/controller/admin_permissions.t
```diff
@@ -0,0 +1,226 @@
...
+            "permissions[user_assign_areas]" => undef,
+        } } );
+
+        is $user2->user_body_permissions->count, 0, 'user2 has had permissions removed';
+    };
+};
+
+$mech->log_out_ok;
+
+$mech->delete_user( $user );
```
dracos: If you put these lines in an `END`, they should run even if a test in this file causes it to die.

t/app/controller/admin_permissions.t
```diff
@@ -0,0 +1,226 @@
...
+my $dt = DateTime->new(
+    year   => 2011,
+    month  => 04,
+    day    => 16,
+    hour   => 15,
+    minute => 47,
+    second => 23
+);
+
+my $report = FixMyStreet::App->model('DB::Problem')->find_or_create(
```
dracos: There's a create_problems_for_body function in `$mech` if that's helpful :)

perllib/FixMyStreet/App/Controller/Admin.pm
davea: `available_permissions` is a nested hashref and I wanted to flatten it out to a simple list of permission names, the inner map seemed like the simplest way to do that. Is there a better approach?

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -634,6 +684,36 @@ sub admin_allow_user {
...
+    return {
+        _("Problems") => {
+            moderate => _("Moderate report details"),
+            report_edit => _("Edit reports"),
+            report_edit_category => _("Edit report category"), # future use
+            report_edit_priority => _("Edit report priority"), # future use
+            report_inspect => _("Markup problem details"),
+            report_instruct => _("Instruct contractors to fix problems"), # future use
+            contribute_as_another_user => _("Create reports/updates on a user's behalf"),
+            contribute_as_body => _("Create reports/updates as the council"),
```
dracos: Do we want planned reports as a permission?

PR #1488, opened 2016-08-15
Allow user to have an area assigned in admin

davea:
See mysociety/fixmystreetforcouncils#37.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -458,6 +458,14 @@ create table user_body_permissions (
     unique(user_id, body_id, permission_type)
 );
 
+create table user_planned_reports (
+    id serial not null primary key,
+    user_id int references users(id) not null,
+    report_id int references problem(id) not null,
+    added timestamp not null default current_timestamp,
+    removed timestamp
+);
+
 create table response_templates (
     id serial not null primary key,
     body_id int references body(id) not null,
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -30,7 +30,8 @@ create table users (
     is_superuser    boolean not null default 'f',
     title           text,
     twitter_id      bigint  unique,
-    facebook_id     bigint  unique
+    facebook_id     bigint  unique,
+    area_id         integer
 );
 
 -- Record details of reporting bodies, including open311 configuration details
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1134,6 +1134,7 @@ sub user_edit : Path('user_edit') : Args(1) {
     }
 
     $c->forward('fetch_all_bodies');
+    $c->forward('fetch_body_areas', [ $user->from_body ]) if $user->from_body;
 
     if ( $c->get_param('submit') ) {
         $c->forward('/auth/check_csrf_token');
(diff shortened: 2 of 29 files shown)


db/schema.sql
```diff
@@ -30,7 +30,8 @@ create table users (
     is_superuser    boolean not null default 'f',
     title           text,
     twitter_id      bigint  unique,
-    facebook_id     bigint  unique
+    facebook_id     bigint  unique,
+    area_id         integer
```
dracos: Discussed briefly in IRC, whether we do this, or add an extra column as with other tables. I guess an extra column is more extensible if we have other things in future, question is (given our current postgres version) do we need to query by area ID, or only look it up by user.

templates/web/base/admin/user-form.html
```diff
@@ -53,6 +53,25 @@
           </li>
         [% END %]
 
+        <li>
```
dracos: If there are no children areas, don't output this section?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1182,6 +1184,10 @@ sub user_edit : Path('user_edit') : Args(1) {
             }
         }
 
+        if ( $user->from_body && $c->user->has_permission_to('user_assign_areas', $user->from_body->id) ) {
+            $user->area_id( $c->get_param('area_id') || undef );
```
dracos: This doesn't check the value is a valid option.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1132,6 +1132,7 @@ sub user_edit : Path('user_edit') : Args(1) {
     $c->stash->{available_permissions} = $c->cobrand->available_permissions;
 
     $c->forward('fetch_all_bodies');
+    $c->forward('fetch_body_areas', [ $user->from_body ]) if $user->from_body;
```
dracos: Would this mean it wouldn't work for a superuser?

dracos: Ah yes, as how would it know which children to show. In which case, definitely shouldn't show the bit in the template if no children, as it will be confusing to superusers.

db/schema.sql
davea: Whilst we don't need to query by area right now, mysociety/fixmystreetforcouncils#38 makes it a likely requirement in the near future so I think this is best as its own column.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1195,7 +1195,9 @@ sub user_edit : Path('user_edit') : Args(1) {
         }
 
         if ( $user->from_body && $c->user->has_permission_to('user_assign_areas', $user->from_body->id) ) {
-            $user->area_id( $c->get_param('area_id') || undef );
+            my %valid_areas = map { $_->{id} => 1 } @{ $c->stash->{areas} };
+            my $new_area = $c->get_param('area_id');
+            $user->area_id( exists($valid_areas{$new_area}) ? $new_area : undef );
```
dracos: Don't need the exists here.

PR #1489, opened 2016-08-15
Fix invalid SQL being generated by moderation lookup

davea:
Not sure why this has only just caused issues, but it was causing a crash on the staging site due to invalid SQL:

`SELECT "me"."id", "me"."admin_user", "me"."object_type", "me"."object_id", "me"."action", "me"."whenedited", "me"."user_id", "me"."reason", "me"."time_spent" FROM "admin_log" "me" WHERE ( ( "action" = ? AND ( "me"."object_id" = ? AND "object_type" = ? ) ) ) ORDER BY "id desc"`

`DBI Exception: DBD::Pg::st execute failed: ERROR:  column "id desc" does not exist`

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/Comment.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Comment.pm
@@ -20,8 +20,8 @@ sub timeline {
 
     return $rs->search(
         {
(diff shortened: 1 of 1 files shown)


dracos:
:+1:

PR #1491, opened 2016-08-15
Performance tweaks step 2

dracos:
In Google Page Speed Insights, with manual minification of the main JS files, this moves the front page from 68/84 to 85/92ish (filmstrip from webpagetest.org, top is live site, bottom is my dev site):

filmstrip

(These are requests from the US, most of the initial delay is in that initial download.) and a report page from 58/77 to 85/86ish (ish because e.g. live site will have analytics that my dev site doesn't):

filmstrip-2

It also lets us turn on the Content-Security-Policy header for fixmystreet.com as all the JavaScript has been factored out.

Depends upon #1490


dracos:
It turns out there was a 'new Function' (purely to prevent an IE memory leak) in OpenLayers, that was caught by the CSP. Replaced manually with a null for the moment.

dracos:
Put GA back inline was only change from this.

PR #1492, opened 2016-08-16
JS History tweaks

dracos:
Couple of small things spotted during use; fixing the back link, stopping possible infinite loop and making the try again button a bit better.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1050,6 +1050,11 @@ $(function() {
         setup_func();
     });
 
+    // Have a fake history entry so we can cover all eventualities.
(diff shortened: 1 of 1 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1045,6 +1050,11 @@ $(function() {
         setup_func();
     });
 
+    // Have a fake history entry so we can cover all eventualities.
+    if ('replaceState' in history) {
+        history.replaceState({ initial: true });
```
dracos: Chrome required a second null argument here.

dracos:
The `history.pushState(null, null, reportListUrl);` also needed an initial of true setting so that it acted the same as initial page load.

PR #1494, opened 2016-08-17
[Harrogate,Zurich,Bristol] JavaScript and map fixes

dracos:
To fix various things due to moving of jquery to footer and removing inline map pin JS.
Remove some unneeded text.
Fix non-JS map tile generation.


davea:
The missing pin on non-JS Zurich maps aside, which isn't really a blocker, this looks good! 👍

dracos:
Not missing for me, so… :-)

PR #1495, opened 2016-08-18
Allow searching for reports with ref: prefix in postcode field

davea:
Searches the `id` and `external_id` fields.

This takes you directly to the report if only a single report matches, otherwise
a disambiguation page is shown.

For mysociety/fixmystreetforcouncils#16

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -38,6 +38,11 @@ sub index : Path : Args(0) {
     # Check if we have a partial report
     my $partial_report = $c->forward('load_partial');
 
+    # Check if the user is searching for a report by ID
+    if ( $c->get_param('pc') && $c->get_param('pc') =~ $c->cobrand->lookup_by_ref_regex ) {
+        $c->go('lookup_by_ref', [ $1 ]);
+    }
+
     # Try to create a location for whatever we have
     my $ret = $c->forward('/location/determine_location_from_coords')
         || $c->forward('/location/determine_location_from_pc');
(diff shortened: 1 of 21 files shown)


perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -38,6 +38,11 @@ sub index : Path : Args(0) {
     # Check if we have a partial report
     my $partial_report = $c->forward('load_partial');
 
+    # Check if the user is searching for a report by ID
+    if ( $c->get_param('pc') =~ /^\s*ref:(\d+)\s*$/ ) {
```
dracos: Could this also, for a UK cobrand, allow searching without `ref:` (as we don't have numeric zipcodes)?

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -289,6 +289,24 @@ sub map : Path('') : Args(2) {
     $c->res->body($image->{data});
 }
 
+sub lookup_by_ref : Private {
```
dracos: As this is part of a front page search (and only sets stash stuff on success), perhaps better in Around.pm? Shrug.

dracos:
I wondered if this should share more with the multiple location search results thing, given it's sort of the same type of thing, but seems fine like this :)

perllib/FixMyStreet/App/Controller/Around.pm
davea: Good idea, done in 9654c64296bd31e8d37ef8f6f59c897245ff1f20.

perllib/FixMyStreet/App/Controller/Report.pm
davea: Yep, have moved it.

davea:
I thought about that, but decided keeping it fairly small and separate was the way to go.

PR #1496, opened 2016-08-19
Fix WMTS-based map issues, and reporting as another user

dracos:


templates/web/base/report/new/form_user_loggedin.html
```diff
@@ -1,20 +1,29 @@
...
+[% ELSE %]
   [% can_contribute_as_another_user = c.user.has_permission_to("contribute_as_another_user", bodies.keys.join(",")) %]
   [% can_contribute_as_body = c.user.from_body AND c.user.has_permission_to("contribute_as_body", bodies.keys.join(",")) %]
-
   [% IF can_contribute_as_another_user OR can_contribute_as_body %]
+    [% INCLUDE form_as %]
+  [% END %]
+[% END %]
+
+[% BLOCK form_as %]
```
davea: This is now being shown to all users when clicking the map on `/around`. It just has a single 'Yourself' option for unprivileged users. It does disappear if you refresh the `/report/new?...` URL.

dracos: Whoops, I meant for the server to not output anything in that case, fix pushed.

davea: Looks good!

davea:
👍

dracos:
Squashed and merged at e51bb2a

PR #1497, opened 2016-08-19
Catch content changes with delegation/selectors.

dracos:
e.g. if a report is pulled in to the page via JavaScript, we need this
to run on the entry in the update form.


davea:
👍

PR #1500, opened 2016-08-24
Response template management

davea:
This extends the Zürich response template admin functionality to all cobrands, and adds:

 - Templates have an 'auto-response' flag, for automatically adding an update when a problem is left (future functionality)
 - Templates can be assigned to individual categories in addition to bodies
 - Managing templates requires the `template_edit` permission.

For mysociety/fixmystreetforcouncils#31

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -982,9 +982,8 @@ sub load_template_body : Private {
     my ($self, $c, $body_id) = @_;
 
     my $zurich_user = $c->user->from_body && $c->cobrand->moniker eq 'zurich';
-    my $has_permission = $c->user->from_body &&
-                         $c->user->from_body->id eq $body_id &&
-                         $c->user->has_permission_to('template_edit', $body_id);
+    my $has_permission = $c->user->has_body_permission_to('template_edit') &&
+                         $c->user->from_body->id eq $body_id;
 
     unless ( $c->user->is_superuser || $zurich_user || $has_permission ) {
         $c->detach( '/page_error_404_not_found' );
@@ -1212,7 +1211,7 @@ sub user_edit : Path('user_edit') : Args(1) {
     my $user = $c->cobrand->users->find( { id => $id } );
     $c->detach( '/page_error_404_not_found' ) unless $user;
 
-    unless ( $c->user->is_superuser || ( $c->user->has_permission_to('user_edit', $c->user->from_body->id) ) ) {
+    unless ( $c->user->is_superuser || $c->user->has_body_permission_to('user_edit') ) {
         $c->detach('/page_error_403_access_denied', []);
     }
 
@@ -1249,7 +1248,7 @@ sub user_edit : Path('user_edit') : Args(1) {
         # set from_body to the same value as their own from_body.
         if ( $c->user->is_superuser ) {
             $user->from_body( $c->get_param('body') || undef );
-        } elsif ( $c->user->has_permission_to('user_assign_body', $c->user->from_body->id ) &&
+        } elsif ( $c->user->has_body_permission_to('user_assign_body') &&
                   $c->get_param('body') && $c->get_param('body') eq $c->user->from_body->id ) {
             $user->from_body( $c->user->from_body );
         } else {
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -309,7 +309,7 @@ sub body : Path('body') : Args(1) {
         $c->forward('update_contacts');
     }
 
-    $c->forward('display_contacts');
+    $c->forward('fetch_contacts');
 
     return 1;
 }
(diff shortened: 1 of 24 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -254,6 +254,29 @@ sub has_permission_to {
     return $permission ? 1 : undef;
 }
 
+=head2 has_body_permission_to
```
dracos: Yay! I think there's probably a few other places this could be used.

templates/web/base/admin/templates_index.html
```diff
@@ -0,0 +1,11 @@
+[% INCLUDE 'admin/header.html' title=loc('Response Templates') -%]
+
+<ul>
+    [% FOR body IN bodies %]
+        <li>
+            <a href="/admin/templates/[% body.id %]/[% t.id %]">[% body.name %]</a>
```
dracos: You don't have a template ID at this point. Do we want to duplicate the body page here, or should it add a Templates link to the existing body page instead, and this page could be, dunno, latest created templates, or blank, or something exciting.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -834,80 +834,107 @@ sub categories_for_point : Private {
...
-            order_by => 'title'
-        }
-    );
+    unless ( $c->user->is_superuser || $zurich_user || $has_permission ) {
+        $c->detach( '/page_error_404_not_found' );
+    }
 
-    $c->stash->{response_templates} = \@templates;
+    # Regular users can only view their own body's templates
+    $body_id = $c->user->from_body->id unless $c->user->is_superuser;
```
dracos: I think it'd be better to redirect to the right page if they've come to a page for a different body (I think that's what this line would mean?).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -834,80 +834,107 @@ sub categories_for_point : Private {
...
-        $template = $body->response_templates->new({});
+        $template = $c->stash->{body}->response_templates->new({});
     }
     else {
-        $template = $body->response_templates->find( $template_id )
+        $template = $c->stash->{body}->response_templates->find( $template_id )
             or $c->detach( '/page_error_404_not_found' );
     }
 
+    $c->forward('display_contacts');
```
dracos: Maybe display_contacts should be renamed, it's fetching them really.

perllib/FixMyStreet/DB/Result/Contact.pm
```diff
@@ -11,29 +11,29 @@ use base 'DBIx::Class::Core';
...
+  { data_type => "text", default_value => "Other", is_nullable => 0 },
+  "deleted",
+  { data_type => "boolean", is_nullable => 0 },
+  "id",
+  {
+    data_type         => "integer",
+    is_auto_increment => 1,
+    is_nullable       => 0,
+    sequence          => "contacts_id_seq",
+  },
```
dracos: Wonder why the order changed here?

templates/web/base/admin/templates.html
```diff
@@ -0,0 +1,26 @@
+[% INCLUDE 'admin/header.html' title=tprintf(loc('Response Templates for %s'), body.name) -%]
```
dracos: Can the zurich version now be deleted?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -834,80 +834,107 @@ sub categories_for_point : Private {
...
             $template->delete;
         } else {
             $template->title( $c->get_param('title') );
-            $template->text ( $c->get_param('text') );
+            $template->text( $c->get_param('text') );
+            $template->auto_response( $c->get_param('auto_response') ? 1 : 0 );
             $template->update_or_insert;
+
+            my @live_contact_ids = map { $_->id } @live_contacts;
+            my @new_contact_ids = grep { $c->get_param("contacts[$_]") ? 1 : undef } @live_contact_ids;
```
dracos: You can just have `$c->get_param("contacts[$_]")` in the grep.

dracos:
Haven't had time to do in browser testing, but code review looks good :)

templates/web/base/admin/templates.html
davea: Not quite, because the zurich admin header doesn't put the `title` value in a <h1>.

perllib/FixMyStreet/DB/Result/Contact.pm
davea: Not sure, might be a Postgres version difference? The extra noise is a bit annoying.

templates/web/base/admin/templates_index.html
davea: Oops. I'd intended this to just be a list of links to bodies, have removed the `/[% t.id %]`.

perllib/FixMyStreet/DB/Result/Contact.pm
dracos: If nothing else has changed, can just silence the noise, it won't make any difference.

templates/web/base/admin/templates.html
dracos: Ah, perhaps just an IF in the base template, then? Given it's otherwise identical.

dracos:
Still not had time to do in browser testing, but fixups also look good :) A grep for `has_permission_to.*from_body` shows other places that could use `has_body_permission_to`.

templates/web/base/admin/templates.html
davea: 👍

PR #1505, opened 2016-08-30
FixaMinGata updates

jonkri:
Uses HTTPS and adds some styles.


dracos:
I think you want "Switch from HTTP to HTTPS in some places" rather than vice-versa :) To match previous commits, would you mind starting them with `[FixaMinGata]`? Actual content is all fine :-) If you can't reword them, no worries, I can merge as is :)

jonkri:
@dracos: Thanks for the review! :)

Does this look OK?

dracos:
Great, thanks! I've rebased as your branch was from quite a while ago, and merged this all in at 80a01c1. I made one change, sorry, I only spotted whilst merging, we no longer need `admin_base_url` in your cobrand on master because it defaults to `/admin` anyway :)

PR #1506, opened 2016-09-02
Add ability for reports to be inspected before being sent

davea:
- Adds an 'inspection required' boolean field to categories, managed via the admin
 - Skips sending reports that haven't been inspected if their category requires it
 - Adds button to inspect view allowing report to be marked as inspected

For mysociety/fixmystreetforcouncils#50

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -362,12 +362,18 @@ sub update_contacts : Private {
         $contact->api_key( $c->get_param('api_key') );
         $contact->send_method( $c->get_param('send_method') );
 
-        # Set the photo_required flag in extra to the appropriate value
+        # Set flags in extra to the appropriate values
         if ( $c->get_param('photo_required') ) {
             $contact->set_extra_metadata_if_undefined(  photo_required => 1 );
         }
         else {
             $contact->unset_extra_metadata( 'photo_required' );
+        }
+        if ( $c->get_param('inspection_required') ) {
+            $contact->set_extra_metadata( inspection_required => 1 );
+        }
+        else {
+            $contact->unset_extra_metadata( 'inspection_required' );
         }
 
         if ( %errors ) {
--- a/perllib/FixMyStreet/Cobrand/Default.pm
+++ b/perllib/FixMyStreet/Cobrand/Default.pm
@@ -860,25 +860,25 @@ sub get_report_stats { return 0; }
 sub get_body_sender {
     my ( $self, $body, $category ) = @_;
 
+    # look up via category
+    my $contact = $body->contacts->search( { category => $category } )->first;
     if ( $body->can_be_devolved ) {
-        # look up via category
-        my $config = $body->result_source->schema->resultset("Contact")->search( { body_id => $body->id, category => $category } )->first;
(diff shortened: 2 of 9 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -363,13 +363,19 @@ sub update_contacts : Private {
...
-        # Set the photo_required flag in extra to the appropriate value
+        # Set flags in extra to the appropriate values
         if ( $c->get_param('photo_required') ) {
             $contact->set_extra_metadata_if_undefined(  photo_required => 1 );
         }
         else {
             $contact->unset_extra_metadata( 'photo_required' );
         }
+        if ( $c->get_param('inspection_required') ) {
+            $contact->set_extra_metadata_if_undefined(  inspection_required => 1 );
```
dracos: I guess this is only place it's defined, but I think just being `set_extra_metadata` makes more sense.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -851,16 +851,16 @@ sub get_report_stats { return 0; }
 sub get_body_sender {
     my ( $self, $body, $category ) = @_;
 
+    # look up via category
+    my $contact = $body->result_source->schema->resultset("Contact")->search( { body_id => $body->id, category => $category } )->first;
```
dracos: `$body->contacts->search( { category => $category } )->first;` :)

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -851,16 +851,16 @@ sub get_report_stats { return 0; }
...
         } else {
-            return { method => $body->send_method, config => $body };
+            return { method => $body->send_method, config => $body, contact => $contact };
         }
     } elsif ( $body->send_method ) {
-        return { method => $body->send_method, config => $body };
+        return { method => $body->send_method, config => $body, contact => $contact };
     }
 
     return $self->_fallback_body_sender( $body, $category );
```
dracos: Should fallback_body_sender also have contact added to it?

perllib/FixMyStreet/Cobrand/Default.pm
dracos: If we're doing this all the time now, note that it is also then looked up again by Email and Open311 SendReports, so we should probably think of a way it only needs to be done once.

dracos:
Needs some tests :)

templates/web/base/report/_inspect.html
```diff
@@ -76,6 +76,17 @@
...
           <input type="submit" value="[% loc('Save changes') %]" name="save" />
         </p>
+        <p>
+          [% UNLESS problem.get_extra_metadata('inspected') %]
+            <input type="submit" value="[% loc('Save changes + send') %]" name="save_inspected" />
+          [% ELSE %]
+            [% IF problem.whensent %]
+              [% loc("<strong>Note:</strong> This report has been sent onwards for action. Any changes made won't be passed on.") %]
+            [% ELSE %]
+              [% loc("<strong>Note:</strong> This report hasn't yet been sent onwards for action. Any changes made may not be passed on.") %]
```
dracos: Is the text here correct?

PR #1511, opened 2016-09-05
Fix photo preview display after submission.

dracos:
If the form is submitted and reshown, due to an error or logging in, the
image preview is loaded from the server, rather than the client. We need
to get the image source data to get the Exif data out of it.


davea:
👍

PR #1514, opened 2016-09-06
Remove from site button fix, validation fixes/improvements

### LINKED ISSUE #503: Non word names can be entered for comments
e.g. '.. ..' is accepted by FMS.

### LINKED ISSUE #1493: Apply validation to 'name' on updates as we do original reports
currently only the original report is validated, one can leave an update with the name 'anon' but not create a report with that name.

### LINKED ISSUE #1508: Change 'remove from site' button to be report-based not cobrand
Currently the staff 'remove from site' button is displayed based upon the cobrand, rather than the report, which is confusing. I think doing it per report doesn't change any functionality for any current users, and is more consistent with e.g. the state dropdown of the update form.

### LINKED ISSUE #1512: Scrolling error into view no longer works in FF
As it now the map sidebar that scrolls, the validation window scrolling doesn't do anything. It needs to scroll the sidebar up to above the focussed element (as FF still only goes precisely to the focus, and the error is shown above it).

### LINKED ISSUE #1513: Validation can go bit awry due to duplicate ID use
All checked server side, so no actual issue, but e.g. FF only sets 'required' on the first ID it finds, and there are two if a report page has been pulled in via ajax (it doesn't set it on the right one, as it's second in the DOM).

dracos:
Shows remove button betterer (fixes #1508). Adds update name validation (fixes #503, fixes #1493). Fixes an issue scrolling error into view on Firefox (fixes #1512). Fixes issue with FF validation not marking all required fields as required due to duplicate ID use (fixes #1513).


davea:
I've always found it a bit scary that the remove report button doesn't have a confirmation step. Is this is good opportunity to wrap the click in a `confirm()` or am I being paranoid? :)

Looks good! 👍

davea:
👍

PR #1516, opened 2016-09-06
Add manage screen for editing priority/category.

dracos:
This is a cut-down version of the full inspect screen. There is a "manage" button if the user has either the edit category or edit priority permissions.

Not sure I like bits of this code-wise, e.g. hardcoded superuser for all permissions, but the alternative was more `can_*` variables in the template. Perhaps that would be simpler anyway, or 'inspect or edit these bits' could be given a single name ("manage" best I can think of) and wrapped in a function.
Anyway, does the job for the two perms given, plus allows pin moving, as that seemed related enough to either?

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -3,6 +3,7 @@ package FixMyStreet::App::Controller::Report;
 use Moose;
 use namespace::autoclean;
 use JSON::MaybeXS;
+use List::MoreUtils qw(any);
 
 BEGIN { extends 'Catalyst::Controller'; }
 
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -367,18 +368,19 @@ sub map : Private {
 
 =head2 check_has_permission_to
 
-Ensure the currently logged-in user has a particular permission that applies
+Ensure the currently logged-in user has particular permission(s) that applies
```
davea: This wording isn't clear if any or all of the given permissions are required (though the code is clear).

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -240,6 +240,27 @@ sub split_name {
     return { first => $first || '', last => $last || '' };
 }
 
+sub permissions {
+    my ($self, $body_id) = @_;
+
+    if ($self->is_superuser) {
+        # XXX Not great having this hardcoded here.
```
davea: Not a big fan of this hardcoding either. Could we pass in `$c` and build the hash from `$c->cobrand->available_permissions`?

templates/web/base/report/_inspect.html
```diff
@@ -1,4 +1,5 @@
 [% PROCESS 'admin/report_blocks.html'; # For the report state dropdown %]
+[% permissions = c.user.permissions(problem.bodies_str) %]
```
davea: This approach is great 💯

davea:
Looks good! Given #1517 refactors how the priority is stored, it makes sense to merge that first then rejig this on top.

PR #1517, opened 2016-09-07
Response priority management in admin

davea:
This PR upgrades the hard-coded `Oxfordshire::problem_response_priorities` method into a fully-fledged `ResponsePriority` model which can be managed via the admin. Priorities are assigned to bodies and categories, and Problems can optionally have a priority set.

For mysociety/fixmystreetforcouncils#66

### DIFF
--- a/db/schema_0045-response-priorities.sql
+++ b/db/schema_0045-response-priorities.sql
@@ -1,0 +1,20 @@
+BEGIN;
+
+CREATE TABLE response_priorities (
+    id serial not null primary key,
+    body_id int references body(id) not null,
+    name text not null,
+    deleted boolean not null default 'f',
+    unique(body_id, name)
+);
+
+CREATE TABLE contact_response_priorities (
+    id serial NOT NULL PRIMARY KEY,
+    contact_id int REFERENCES contacts(id) NOT NULL,
+    response_priority_id int REFERENCES response_priorities(id) NOT NULL
+);
+
+ALTER TABLE problem
+    ADD COLUMN response_priority_id int REFERENCES response_priorities(id);
+
+COMMIT;
--- a/perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
@@ -1,0 +1,107 @@
+package FixMyStreet::App::Controller::Admin::ResponsePriorities;
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+
+sub begin : Private {
+    my ( $self, $c ) = @_;
+
+    $c->forward('/admin/begin');
+}
+
+sub index : Path : Args(0) {
+    my ( $self, $c ) = @_;
+
+    my $user = $c->user;
+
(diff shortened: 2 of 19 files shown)


db/schema_0045-response-priorities.sql
```diff
@@ -0,0 +1,20 @@
...
+);
+
+CREATE TABLE contact_response_priorities (
+    id serial NOT NULL PRIMARY KEY,
+    contact_id int REFERENCES contacts(id) NOT NULL,
+    response_priority_id int REFERENCES response_priorities(id) NOT NULL
+);
+
+ALTER TABLE problem
+    ADD COLUMN response_priority_id int REFERENCES response_priorities(id);
```
dracos: You've probably already spotted this, this line needs adding in schema.sql.

perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
```diff
@@ -0,0 +1,125 @@
...
+    my ( $self, $c ) = @_;
+
+    $c->forward('/admin/begin');
+}
+
+sub auto : Private {
+    my ( $self, $c ) = @_;
+
+    $c->forward('/admin/auto');
+}
```
dracos: All `auto` functions in the chain are called (unlike `begin` or `end`) so this function shouldn't be needed, if it is, something must be going wrong somewhere :)

perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
```diff
@@ -0,0 +1,125 @@
+package FixMyStreet::App::Controller::Admin::ResponsePriorities;
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+use Path::Class;
```
dracos: Not sure any of these are used.

perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
```diff
@@ -0,0 +1,125 @@
...
+
+    my $has_permission = $c->user->has_body_permission_to('responsepriority_edit') &&
+                         $c->user->from_body->id eq $body_id;
+
+    unless ( $c->user->is_superuser || $has_permission ) {
+        $c->detach( '/page_error_404_not_found' );
+    }
+
+    # Regular users can only view their own body's priorities
+    if ( !$c->user->is_superuser && $body_id ne $c->user->from_body->id ) {
```
dracos: Don't think you can ever get to this if passing, due to the 404 check above. Don't mind whether it 404s or redirects, 404 probably easier :)

db/schema_0045-response-priorities.sql
dracos: General query - could priority have stayed as extra metadata?

davea: I thought about it, and similar to `users.area_id` thought it's likely we'll want to index/filter on this field in the near future.

dracos:
Would be good to have some tests of the admin pages. Tests for the front end inspect page showing/accepting the right things probably easier after merging this and then #1516.
🍨

PR #1518, opened 2016-09-07
[Oxfordshire] Display confirmed reports as received by the Council

davea:
This introduces a new `is_council_with_case_management` flag for Cobrands, that,
when active, considers any report that's been confirmed as having been received
by the Council in question.

See mysociety/fixmystreetforcouncils#65

### DIFF
--- a/templates/web/base/report/_council_sent_info.html
+++ b/templates/web/base/report/_council_sent_info.html
@@ -1,6 +1,6 @@
-[% IF problem.whensent || problem.can_display_external_id %]
+[% SET duration_clause = problem.duration_string(c) %]
+[% IF duration_clause || problem.can_display_external_id %]
     <p class="council_sent_info">
-    [% SET duration_clause = problem.duration_string(c) IF problem.whensent %]
     [%- IF problem.can_display_external_id %]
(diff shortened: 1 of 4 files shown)


templates/web/base/report/_council_sent_info.html
```diff
@@ -1,6 +1,7 @@
-[% IF problem.whensent || problem.can_display_external_id %]
+[% problem_sent = problem.whensent || c.cobrand.get_body_handler_for_problem(problem).is_council_with_case_management %]
```
dracos: To keep the template tidier, perhaps `duration_string` could return blank if there's no whensent (after your new bit of code there for case management), and then this template and the other could then print if it wasn't blank (so leaving the logic of whensent + whatever to the code).

davea: Good idea, I've fixed this up.

davea:
I've fixed up the template and made `Oxfordshire::is_council_with_case_management` dependent on `STAGING_SITE` so this won't affect the live site until we want it to.

Happy for me to merge?

dracos:
🎈

PR #1520, opened 2016-09-08
Oxfordshire photo guidance

zarino:
Fixes mysociety/fixmystreetforcouncils#6

screen shot 2016-09-08 at 11 48 37

Adds the photo guidance from FMS.com and also pluralizes the "Photos" label for all cobrands.


davea:
Is it worth adding the new guidance for all UK cobrands by putting it in `fixmystreet-uk-councils` instead of `oxfordshire`?

Also, don't forget the update form can have multiple photos too.

dracos:
Is it worth putting the guidance on all cobrands, in fact?

zarino:
Ok, the "Photos" pluralisation should now be in place wherever the dropzone component appears.

And I've moved the `after_photo.html` into `base`, so it should appear for every cobrand now. @davea can you think of any cobrands that won't want to show this?

dracos:
Now it's in base,it'll need marking for translation.

zarino:
@dracos: Done!

PR #1521, opened 2016-09-08
Require inspector to provide update when instructing report

davea:
This adds an update field to the bottom of the inspect form, allowing the
inspector to provide an update to be added to the report as it's sent.

See mysociety/fixmystreetforcouncils#64

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -4,6 +4,7 @@ use Moose;
 use namespace::autoclean;
 use JSON::MaybeXS;
 use List::MoreUtils qw(any);
+use Utils;
 
 BEGIN { extends 'Catalyst::Controller'; }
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -353,6 +357,18 @@ sub inspect : Private {
 
         if ($valid) {
             $problem->update;
+            if ( defined($update_text) ) {
```
dracos: This will be true if they have clicked save_inspected and not provided any update text.

dracos:
I see you moved valid further up, but it doesn't appear to use that - a server side check that update_text is set if that is a requirement? Otherwise, :+1: (with another test on top of #1516 please :)

davea:
I've fixed this so the server checks for an empty update, will add the test and merge after you've merged #1516.

PR #1524, opened 2016-09-09
Improve display of category extra data in admin.

dracos:
Allow get_extra_metadata to return all the data with no argument.

screen shot 2016-09-09 at 15 11 49


davea:
Nice! 🚢

BenJam:
👍

PR #1525, opened 2016-09-13
update devices mockup to remove scamsmart web address

wrightmartin:


dracos:
Taken into merged rebase.

PR #1527, opened 2016-09-14
Issues/forcouncils/57 trusted users

davea:
Adds UI to the user edit form allowing users to be given 'trusted' status to make reports that are sent to the external body/contractor without being inspected first.

This uses the `UserBodyPermission` model, meaning a user's trusted status is applied on a per-body basis. For superusers the admin UI is a multiselect for all bodies, but regular admin users can only assign trusted status for their own body.

This also removes the requirement for users to have a `from_body` set in `User::has_permission_to`, because trusted users don't have a `from_body`.

For mysociety/fixmystreetforcouncils#57

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -687,7 +687,7 @@ sub report_edit : Path('report_edit') : Args(1) {
 
     unless (
         $c->cobrand->moniker eq 'zurich'
-        || $c->user->has_permission_to(report_edit => $problem->bodies_str)
+        || $c->user->has_permission_to(report_edit => $problem->bodies_str_ids)
     ) {
         $c->detach( '/page_error_403_access_denied', [] );
     }
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1276,14 +1276,14 @@ sub user_edit : Path('user_edit') : Args(1) {
 
         if (!$user->from_body) {
             # Non-staff users aren't allowed any permissions or to be in an area
-            $user->user_body_permissions->delete_all;
+            $user->admin_user_body_permissions->delete;
             $user->area_id(undef);
             delete $c->stash->{areas};
             delete $c->stash->{fetched_areas_body_id};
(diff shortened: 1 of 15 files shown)


davea:
NB the changes in 0e6aece2288f4b35f57c9e1581dc06c849757c22 seem to have broken `t/app/controller/report_as_other.t` when run as part of the full test suite, though that file passes if run independently.

t/app/controller/admin.t
```diff
@@ -1181,6 +1181,7 @@ FixMyStreet::override_config {
                 phone => '',
                 flagged => undef,
                 is_superuser => undef,
+                trusted_bodies => undef,
```
dracos: Should it be in default_perms?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1272,7 +1272,9 @@ sub user_edit : Path('user_edit') : Args(1) {
 
         if (!$user->from_body) {
             # Non-staff users aren't allowed any permissions or to be in an area
-            $user->user_body_permissions->delete_all;
+            $user->user_body_permissions->search({
```
dracos: You could have a User method of admin_user_body_permissions that ignored the trusted one? That would also help with...

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1282,6 +1284,7 @@ sub user_edit : Path('user_edit') : Args(1) {
             $user->user_body_permissions->search({
                 body_id => $user->from_body->id,
                 permission_type => { '!=' => \@user_permissions },
+                permission_type => { '!=' => 'trusted' },
```
dracos: …this won't work because you've used the same key twice.

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -260,12 +260,11 @@ sub has_permission_to {
     my ($self, $permission_type, $body_id) = @_;
 
     return 1 if $self->is_superuser;
+    return 0 unless $body_id;
```
dracos: body_id should probably be renamed, as it's an array of one or more body IDs. (I think the `find` will still work, but should be checked.)

dracos:
Few small things :)

PR #1528, opened 2016-09-14
WIP: Restructure reporting form, standardize display of "extra questions"

zarino:
Fixes mysociety/fixmystreetforcouncils#7

To do:

- [x] Check reporting page and login page look ok in all the cobrands
- [x] Cross browser check
- [x] Screenreader check
- [x] “FixMyStreet is primarily for reporting physical problems” message? (UK council cobrands, plus a few others)
- [x] On FMS.com, reporting in a category that the council refuses, the yellow warning correctly appears below the category select box, but the "Public details" `.form-section-description` still says "These will be sent to `[council]` and also published online…" rather than just "These will be published online…". Same for the "Private details" description too.
- [x] The council name isn't inserted into the "Public details" `.form-section-description` when the form is loaded by javascript (ie: when you open the form by clicking the map). This is because `all_body_names` isn't set when the (hidden) form is rendered into the `around` page, and the map click javascript does nothing to update it.
- [x] Check for any remaining to-dos (`[%# TODO`)

screen shot 2016-09-14 at 15 56 25


dracos:
As well as updating the public/private details descriptions upon category/location change, I've taken the opportunity to put any unresponsive/don't have details message above the whole form, and only show the right body name once you pick a category. I think.

templates/web/fixmystreet-uk-councils/report/new/after_category.html
```diff
@@ -1 +0,0 @@
-<p><small>[% loc("Remember that FixMyStreet is primarily for reporting physical problems that can be fixed. If your problem is not appropriate for submission via this site remember that you can contact your council directly using their own website.") %]</small></p>
```
zarino: Oxfordshire (`templates/web/oxfordshire/report/new/after_category.html`) and Harrogate (`templates/web/harrogate/report/new/after_category.html`) still include this warning – should we remove them there too?

zarino:
This just got a 👍  from @wrightmartin, for the SCSS code and the general design changes.

templates/web/fixmystreet-uk-councils/report/new/after_category.html
dracos: I guess we leave it, and we can ask if they want to keep it. Oxfordshire's is specially worded anyway.

PR #1530, opened 2016-09-16
Some development bits

dracos:


bin/fixmystreet.com/fixture
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env perl
+
```
davea: Although the code mostly speaks for itself, could you add a comment describing this script and when it might be used?

PR #1532, opened 2016-09-20
Organize report admin actions into single control

zarino:
Fixes #89.

Adds a new segmented control, containing the admin actions the current user is permitted to perform (currently any number of the 3 actions: moderation, inspection, and shortlisting).

screen shot 2016-09-20 at 11 24 21

Tested in IE8-10, and modern browsers. And screenreaders.

### DIFF
--- a/templates/web/base/report/_main.html
+++ b/templates/web/base/report/_main.html
@@ -8,19 +8,25 @@
 <div class="problem-header clearfix" problem-id="[% problem.id %]">
 
 [% IF c.user.has_permission_to('planned_reports', problem.bodies_str) %]
-<form method="post" action="/my/planned/change" id="planned_form">
+<form method="post" action="/my/planned/change" id="planned_form" class="hidden-label-target">
     <input type="hidden" name="id" value="[% problem.id %]">
     <input type="hidden" name="token" value="[% csrf_token %]">
     <input type="hidden" name="change" value="[% IF c.user.is_planned_report(problem) %]remove[% ELSE %]add[% END %]">
-    <p><input type="submit"
-        data-remove="[% loc('Remove from planned reports') %]" data-add="[% loc('Add to planned reports') %]"
-        value="
+    <p><input
+        type="submit"
+        id="shortlist-report"
+        data-label-remove="[% loc('Remove from shortlist') %]"
+        data-label-add="[% loc('Add to shortlist') %]"
+        data-value-remove="[% loc('Shortlisted') %]"
+        data-value-add="[% loc('Shortlist') %]"
     [%~ IF c.user.is_planned_report(problem) ~%]
-        [% loc('Remove from planned reports') %]
+        value="[% loc('Shortlisted') %]"
+        aria-label="[% loc('Remove from shortlist') %]"
     [%~ ELSE ~%]
-        [% loc('Add to planned reports') %]
+        value="[% loc('Shortlist') %]"
+        aria-label="[% loc('Add to shortlist') %]"
     [%~ END ~%]
-    "></p>
+    ></p>
 </form>
 [% END %]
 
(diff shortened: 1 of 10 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -281,16 +281,25 @@ $.extend(fixmystreet.set_up, {
...
+                buttonLabel = $submit.data('label-remove');
+                buttonValue = $submit.data('value-remove');
             } else if (data.outcome == 'remove') {
-                $change.val('add');
-                $submit.val($submit.data('add'));
+                changeValue = "add";
+                buttonLabel = $submit.data('label-add');
+                buttonValue = $submit.data('value-add');
             }
+            $change.val(changeVal);
```
dracos: This should be changeValue

templates/web/base/report/_main.html
```diff
@@ -8,19 +8,25 @@
...
+    <p><input
+        type="submit"
+        id="shortlist-report"
+        data-label-remove="[% loc('Remove from shortlist') %]"
+        data-label-add="[% loc('Add to shortlist') %]"
+        data-value-remove="[% loc('Shortlisted') %]"
+        data-value-add="[% loc('Shortlist') %]"
     [%~ IF c.user.is_planned_report(problem) ~%]
-        [% loc('Remove from planned reports') %]
+        value="[% loc('Shortlisted') %]"
```
dracos: Should shortlisted look pushed in, like the radio segmented controls do, perhaps?

zarino:
@dracos – I've fixed the test to expect "Shortlist"/"Shortlisted" in place of the old button text, and also updated all the user-facing bits of UI to use the "Shortlist" terminology instead of "Planned reports".

The URLs, database fields, etc still use "`planned`" but I'm guessing it's safest to leave them as-is.

templates/web/base/report/_main.html
dracos: Then also Inspect could perhaps stay visible, also pushed in, when the inspect part is being shown, and clicking it could come out of inspect mode? Though might worry about losing info then I guess.

zarino: > Should shortlisted look pushed in, like the radio segmented controls do, perhaps?
  
Ooh that’s an interesting idea!

> Then also Inspect could perhaps stay visible, also pushed in, when the inspect part is being shown, and clicking it could come out of inspect mode? Though might worry about losing info then I guess.

We already have a "Cancel" button in the Inspection form, that just links back to `/report/:id`, losing any changes you’ve made in the form. Maybe a pressed-in button labelled "Inspect" is a little more dangerous, because it isn't labelled with a danger word.

web/cobrands/fixmystreet/fixmystreet.js
zarino: Done!

templates/web/base/report/_main.html
dracos: Yeah, agreed; I still think it would look nice to be there, pushed in, but clicking it do nothing perhap?

zarino: I guess the text could change on the pushed-in buttons – so pushing the "Shortlist" button turns it into a pushed-in, yellow "Shortlisted"; and pushing the "Inspect" button turns it into a pushed-in, yellow "Cancel" button.

Danger here is that, while "Shortlisted" makes sense as a label (since it's describing the current state of the stateful button – it's just a linguistic equivalent to the yellow, pushed-in styling), "Cancel" is an action rather than a state, so it doesn't match. The matching label for that button, I guess, would be "Inspecting", since that describes the state the report is currently in.

Bah.

zarino: Also, this is the point at which I ask: Why do we have an "Inspect" button at all? Why doesn't the "Inspector" sidebar *always* display if you have the `report_inspect` permission?

@BenJam @davea ^

zarino:
@dracos – I've just pushed a commit that refactors `.radio-segmented-control` into `.segmented-control--radio`, a special-case of `.segmented-control`. Makes the styling a bit simpler too.

PR #1533, opened 2016-09-20
Add user reputation for automatic report sending

davea:
This PR adds the ability for users to gain reputation if they make high-quality reports, and lose it if their reports are duplicates/otherwise rejected or removed from the site.

The purpose of reputation is to allow reports to be sent immediately without any need for inspection/instruction for those categories that require inspection. It operates in a similar way to the 'trusted' flag.

 - User's reputation is stored in the newly-added `extra` column.
 - Categories requiring inspection can have a reputation threshold set, below which reports must still be inspected before being sent.
 - When a report is inspected and sent on to be fixed the user gains a reputation point.
 - ~~When a report is rejected (e.g. has its state set to hidden/duplicate/not responsible) at the inspection stage the user's reputation is reduced by a point.~~
 - If a report is removed from the site entirely via the 'Remove from site' link, the user's reputation is decremented.

For mysociety/fixmystreetforcouncils#56.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -270,6 +270,8 @@ sub delete :Local :Args(1) {
     $p->lastupdate( \'current_timestamp' );
     $p->update;
 
+    $p->user->update_reputation(-1);
+
     $c->model('DB::AdminLog')->create( {
         admin_user => $c->user->email,
         object_type => 'problem',
@@ -315,6 +317,7 @@ sub inspect : Private {
 
         my $valid = 1;
         my $update_text;
+        my $reputation_change = 0;
 
         if ($permissions->{report_inspect}) {
             foreach (qw/detailed_location detailed_information traffic_information/) {
@@ -325,6 +328,7 @@ sub inspect : Private {
                 $update_text = Utils::cleanup_text( $c->get_param('public_update'), { allow_multiline => 1 } );
                 if ($update_text) {
                     $problem->set_extra_metadata( inspected => 1 );
+                    $reputation_change = 1;
                 } else {
                     $valid = 0;
                     $c->stash->{errors} ||= [];
@@ -362,6 +366,9 @@ sub inspect : Private {
         }
 
         if ($valid) {
+            if ( $reputation_change != 0 ) {
+                $problem->user->update_reputation($reputation_change);
+            }
             $problem->update;
             if ( defined($update_text) ) {
                 $problem->add_to_comments( {
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -378,4 +378,12 @@ sub is_planned_report {
     return $self->active_planned_reports->find({ id => $problem->id });
 }
 
+sub update_reputation {
+    my ( $self, $change ) = @_;
+
+    my $reputation = $self->get_extra_metadata('reputation') || 0;
+    $self->set_extra_metadata( reputation => $reputation + $change);
+    $self->update;
+}
+
 1;
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -26,20 +26,22 @@ __PACKAGE__->add_columns(
   { data_type => "text", is_nullable => 1 },
   "password",
   { data_type => "text", default_value => "", is_nullable => 0 },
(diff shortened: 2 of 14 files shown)


t/app/controller/report_inspect.t
```diff
@@ -131,6 +119,22 @@ FixMyStreet::override_config {
...
 
+FixMyStreet::override_config {
+    MAPIT_URL => 'http://mapit.mysociety.org/',
+    ALLOWED_COBRANDS => 'oxfordshire',
+}, sub {
+    subtest "test negative reputation" => sub {
+        my $reputation = $report->user->get_extra_metadata("reputation");
+
+        $mech->get_ok("/report/$report_id");
+        $mech->submit_form( button => 'remove_from_site' );
```
dracos: So this button wasn't being tested at all previously? Whoops!

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -378,4 +378,12 @@ sub is_planned_report {
     return $self->active_planned_reports->find({ id => $problem->id });
 }
 
+sub update_reputation {
+    my ( $self, $change ) = @_;
+
+    my $reputation = $self->get_extra_metadata('reputation') || 0;
+    $self->set_extra_metadata( reputation => $reputation + $change);
```
dracos: Tiny possibility of race condition here, I guess, but not sure it's worth doing anything about it.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -362,6 +374,10 @@ sub inspect : Private {
         }
 
         if ($valid) {
+            if ( $reputation_change != 0 && !$problem->get_extra_metadata('user_reputation_set') ) {
+                $problem->user->update_reputation($reputation_change);
+                $problem->set_extra_metadata( user_reputation_set => 1)
```
dracos: (After the changes to remove negative_states), do we mind if somehow the report got unhidden/uninspected and done again?

t/app/controller/report_inspect.t
```diff
@@ -6,8 +6,8 @@ use FixMyStreet::TestMech;
 
 my $mech = FixMyStreet::TestMech->new;
 
-my $brum = $mech->create_body_ok(2514, 'Birmingham City Council');
-my $oxon = $mech->create_body_ok(2237, 'Oxfordshire County Council');
+my $brum = $mech->create_body_ok(2514, 'Birmingham City Council', id => 2514);
+my $oxon = $mech->create_body_ok(2237, 'Oxfordshire County Council', id => 2237);
```
dracos: I wonder if we can fix the UK council cobrand to_body to not require us to do this, hmm...

t/app/controller/report_inspect.t
davea: Nope, I don't think so!

perllib/FixMyStreet/DB/Result/User.pm
davea: Hmm yes, though given the relative unimportance of the reputation value I think it's an acceptable risk.

perllib/FixMyStreet/App/Controller/Report.pm
davea: Hmm, do you mean that if a report is unhidden the reputation decrement should be undone? I think in the interests of keeping things simple each report should only ever have one effect on the user's reputation. Open to suggestions otherwise though.

t/app/controller/report_inspect.t
davea: Or alternatively change create_body_ok to use the first param as the body id unless otherwise set :)

perllib/FixMyStreet/App/Controller/Report.pm
dracos: No, I just meant do you need to bother setting/checking the user_reputation_set metadata, as it seems very unlikely that a report could have this happen twice anyway.

t/app/controller/report_inspect.t
dracos: I hope that we can get rid of the UK cobrand's need for body ID = area ID at some point, and doing that would certainly not persuade us to ;)

perllib/FixMyStreet/App/Controller/Report.pm
davea: Got it. That's a fair point, it probably doesn't matter enough to justify the extra user_reputation_set complexity, which I'll remove.

PR #1538, opened 2016-09-21
Some minor admin fixes

davea:
- Superusers couldn't view response priorities due to invalid `$c->forward` call
 - Navigation to other areas of admin was broken on response priority admin pages

Decided to do this as a PR instead of straight on to master as I wasn't sure that my `uri_for` fix was the best way to do it.

### DIFF
--- a/templates/web/base/admin/navigation.html
+++ b/templates/web/base/admin/navigation.html
@@ -2,7 +2,7 @@
     <p>
         <strong>[% tprintf(loc('%s admin:', "%s here is the site name"), site_name) %]</strong>
       [%- FOREACH link IN allowed_links %]
-        <a href="[% c.uri_for( link ) %]">[% allowed_pages.$link.0 %]</a>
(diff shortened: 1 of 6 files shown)


templates/web/base/admin/navigation.html
```diff
@@ -2,7 +2,7 @@
     <p>
         <strong>[% tprintf(loc('%s admin:', "%s here is the site name"), site_name) %]</strong>
       [%- FOREACH link IN allowed_links %]
-        <a href="[% c.uri_for( link ) %]">[% allowed_pages.$link.0 %]</a>
+        <a href="[% c.uri_for( '/admin/' _ link ) %]">[% allowed_pages.$link.0 %]</a>
```
dracos: This is fine :) uri_for makes it a working relative link. However, the response priority body lists (and two template lists too) have hard-coded "/admin…" which won't work if the admin is proxied.

PR #1539, opened 2016-09-26
[fixmystreet.com] Update privacy policy to allow Google Analytics demographics

jacksonj04:
Update the privacy policy for fixmystreet.com to allow us to switch on demographic tracking in Google Analytics. This is a direct lift of the [policy from TheyWorkForYou](https://www.theyworkforyou.com/privacy/), which we're already confident allows this sort of thing. As part of this it removes the table of Google cookies, adds a bit on opt-out tools, and retains the Google statement.

### DIFF
--- a/templates/web/fixmystreet.com/about/privacy.html
+++ b/templates/web/fixmystreet.com/about/privacy.html
@@ -75,7 +75,9 @@ the cookies and services that this site can use.
 
 <p>We use Google Analytics to collect information about how people use this
 site. We do this to make sure it’s meeting its users’ needs and to understand
-how we could do it better. Google Analytics stores information such as what
+how we could do it better.
+
+<p>Google Analytics stores information such as what
 pages you visit, how long you are on the site, how you got here, what you click
 on, and information about your web browser. IP addresses are masked (only a
 portion is stored) and personal information is only reported in aggregate. We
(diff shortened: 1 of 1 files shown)


templates/web/fixmystreet.com/about/privacy.html
```diff
@@ -73,50 +73,71 @@
...
-any other data held by Google. You may refuse the use of cookies by selecting
-the appropriate settings on your browser, however please note that if you do
-this you may not be able to use the full functionality of this website. By
-using this website, you consent to the processing of data about you by Google
-in the manner and for the purposes set out above.”</p>
-
-<p><a href="https://www.mysociety.org/privacy-online/">More general information
-on how third party services work</a></p>
+<p>We use Google Analytics to collect information about how people
+use this site. We do this to make sure it&rsquo;s meeting its
```
dracos: Could all these entities be the actual characters, please. And ideally, match the previous line width wherever possible so this diff is as small as necessary (that last part more for e.g. the Official Statement as I imagine that hasn't changed much, if at all).

templates/web/fixmystreet.com/about/privacy.html
```diff
@@ -73,50 +73,71 @@
 
 <h3>Measuring website usage (Google Analytics)</h3>
 
-<p>We use Google Analytics to collect information about how people use this
-site. We do this to make sure it’s meeting its users’ needs and to understand
-how we could do it better. Google Analytics stores information such as what
-pages you visit, how long you are on the site, how you got here, what you click
-on, and information about your web browser. IP addresses are masked (only a
```
dracos: Please could this section be retained, it appears to have been lost.

templates/web/fixmystreet.com/about/privacy.html
```diff
@@ -113,10 +122,24 @@
 the appropriate settings on your browser, however please note that if you do
 this you may not be able to use the full functionality of this website. By
 using this website, you consent to the processing of data about you by Google
-in the manner and for the purposes set out above.”</p>
-
-<p><a href="https://www.mysociety.org/privacy-online/">More general information
-on how third party services work</a></p>
+in the manner and for the purposes set out above.”
```
dracos: Keep the end `</p>` that was there before? :)

PR #1542, opened 2016-09-27
WIP: Mobile improvements for Inspect form

zarino:
Fixes https://github.com/mysociety/fixmystreetforcouncils/issues/36.

Includes:

* A new opt-in `.btn` class, and removal of most of the previous button styling, to make theming buttons and inputs easier. `.btn-primary` has been added as an alias of `.green-btn`, with an aim to eventually phasing out `.green-btn` and `.red-btn` entirely.
* More logical order for form elements in the Inspect form.
* More finger-friendly actions and form inputs on the Inspect form (eg: big button for “Set to my current location“, more spacing between Cancel and Save buttons).
* “Traffic management” is now a select box with pre-defined options (eg: yes / no), rather than a free text input.
* Unified “Extra details” input, in place of the previous 3 separate inputs: “traffic information”, “detailed location”, and “detailed information”.
* Fold-out section for extra metadata questions (eg: pothole sizes) and status-related actions (eg: resolving duplicates).
* "Public update" is now an opt-in field at the end of the main form – avoiding the confusion of a single form with two submit buttons.
* Button to “Navigate to this problem” with Google Maps (fixes https://github.com/mysociety/fixmystreetforcouncils/issues/81)
* Completely reworked `.with-actions` layout, using flexbox to set up much more robust columns, and allowing us to scroll Inspectors down to the Inspect form after they click the Inspect button (handy on mobiles). Might also fix the weird "scrollbar appearing outside the element" bug we saw on Safari in OS X with hidden scrollbars set in System Preferences.

Here’s an example of it in practice – ignore the “There are no reports to show” message after I mark the report as a duplicate – I'm just trying to get across the idea that the form contents will change, to display inputs or features related to the current category and status of the report.

form


dracos:
Pulling in the extra questions has thrown up that long labels don't seem to wrap and are making the sidebar far too wide.

dracos:
Another CSS thing - due to the width of the second column, we need to do something media-query wise when the browser width is wider than max 'mobile' display, but not wide enough to show both columns (and no map).

zarino:
@dracos I've addressed those three issues you brought up. Anything else to prevent this being merged?

templates/web/base/report/_inspect.html
```diff
@@ -1,49 +1,40 @@
...
-            <option value="" [% 'selected' UNLESS problem.response_priority_id %]>-</option>
-            [% FOREACH priority IN problem.response_priorities %]
-              <option value="[% priority.id %]" [% 'selected' IF problem.response_priority_id == priority.id %] [% 'disabled' IF priority.deleted %]>[% priority.name %]</option>
-            [% END %]
-          </select>
+      <div class="inspect-section">
+        <p>
+          <strong>[% loc('Report ID:') %]</strong>
+          [% problem.id %]
+          [%# XXX Not sure whether this problem_id is required when submitting form? %]
```
dracos: It doesn't look like it is.

zarino:
@dracos – I've merged the commits into more logical chunks. But all the editing as left the dates in a weird order. Should I be worried about that?

dracos:
No 🐤 (The commit, as opposed to author, dates are in a non-weird order, if you might want to use those instead by default.)

PR #1546, opened 2016-09-30
Show shortlist owner and allow taking.

dracos:
For mysociety/fixmystreetforcouncils#86 and mysociety/fixmystreetforcouncils#87.

Looks like this when a user with the `planned_reports` permission views a report that is on *another* user’s shortlist:

screen shot 2016-10-06 at 11 31 16


zarino:
Feels to me like an MVP for this wouldn't even involve a special "Take" button – just something to show whose shortlist the report is on. Then the act of "Taking" is really just the "Shortlisting" you’d do via the existing "Shortlist" button on the report page.

Something like this (super quick mockup but you get the idea):

screen shot 2016-10-05 at 17 34 37

It's not currently possible to shortlist a report from the area/list pages, so I don't see why this PR needs to introduce a way to "take" a shortlisted report on those pages right now. If we want to introduce one-click shortlisting from the area/list pages, we could do that in a separate PR 😄

zarino:
@dracos: I've given the shortlist message some styling to help it stand out in the sidebar. If/when we start referring to shortlists elsewhere on the site, we can start using the same icon, for consistency.

PR #1547, opened 2016-09-30
Allow multiple selections in report-list-filter UI

zarino:
To fix https://github.com/mysociety/fixmystreetforcouncils/issues/95

- [x] You can now select multiple statuses and categories in the UI
- [x] Update backend to work with multiple selections.
- [x] Fix test (probably will be by above)
- [x] Sort out vertical alignment across browsers
- [x] Sort out menu overflow in sidebar
   - https://github.com/mysociety/jquery-multi-select/pull/5

dropdowns

~~We should probably re-think the logic of the options here – for instance, should “all reports” be a fourth option in the status select box, or should the box just default to having its three options (“unfixed reports”, “closed reports”, “fixed reports”) ticked?~~

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -181,7 +181,7 @@ sub display_location : Private {
     my ( $on_map_all, $on_map, $nearby, $distance ) =
       FixMyStreet::Map::map_features( $c,
         latitude => $latitude, longitude => $longitude,
-        interval => $interval, category => $c->stash->{filter_category},
+        interval => $interval, categories => $c->stash->{filter_category},
         states => $c->stash->{filter_problem_states} );
 
     # copy the found reports to the stash
@@ -258,13 +258,11 @@ sub check_and_stash_category : Private {
     )->all;
     my @categories = map { $_->category } @contacts;
     $c->stash->{filter_categories} = \@categories;
-
-
-    my $category = $c->get_param('filter_category');
     my %categories_mapped = map { $_ => 1 } @categories;
-    if ( defined $category && $categories_mapped{$category} ) {
-        $c->stash->{filter_category} = $category;
-    }
+
+    my $categories = [ $c->get_param_list('filter_category', 1) ];
+    my @valid_categories = grep { $_ && $categories_mapped{$_} } @$categories;
+    $c->stash->{filter_category} = \@valid_categories;
 }
 
 =head2 /ajax
@@ -303,7 +301,7 @@ sub ajax : Path('/ajax') {
     my ( $on_map_all, $on_map_list, $nearby, $dist ) =
       FixMyStreet::Map::map_features($c,
           bbox => $bbox, interval => $interval,
-          category => $c->get_param('filter_category'),
+          categories => [ $c->get_param_list('filter_category', 1) ],
           states => $c->stash->{filter_problem_states} );
 
     # create a list of all the pins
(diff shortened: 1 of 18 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1223,11 +1223,16 @@ $(function() {
                     // location.href is something like foo.com/around?pc=abc-123,
                     // which we pass into fixmystreet.display.reports_list() as a fallback
                     // in case the list isn't already in the DOM.
+                    $('select[multiple] option').prop('selected', function() { return this.defaultSelected; }).trigger('change.multiselect');
```
zarino: Should this be `$('#filter_categories option, #statuses option')` ?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1223,11 +1223,16 @@ $(function() {
...
                     fixmystreet.display.reports_list(location.href);
                 } else if ('reportId' in e.state) {
                     fixmystreet.display.report(e.state.reportPageUrl, e.state.reportId);
                 } else if ('newReportAtLonlat' in e.state) {
                     fixmystreet.display.begin_report(e.state.newReportAtLonlat, false);
+                } else if ('filter_change' in e.state) {
+                    $('#filter_categories').val(e.state.filter_change.filter_categories);
+                    $('#statuses').val(e.state.filter_change.statuses);
+                    $('select[multiple]').trigger('change.filters').trigger('change.multiselect');
```
zarino: Should this be `$('#filter_categories, #statuses')` ?

dracos: Sure (well, $('#filter_categories').add('#statuses') is about 10 times faster :) )

web/cobrands/fixmystreet/fixmystreet.js
dracos: Will go with `$('#filter_categories').add('#statuses').find('option');`

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -258,13 +258,11 @@ sub check_and_stash_category : Private {
...
-
-    my $category = $c->get_param('filter_category');
     my %categories_mapped = map { $_ => 1 } @categories;
-    if ( defined $category && $categories_mapped{$category} ) {
-        $c->stash->{filter_category} = $category;
-    }
+
+    my $categories = [ $c->get_param_list('filter_category', 1) ];
+    my @valid_categories = grep { $_ && $categories_mapped{$_} } @$categories;
+    $c->stash->{filter_category} = \@valid_categories;
```
davea: I suppose renaming the request query param to `filter_categories` would run the risk of breaking bookmarks/links, but should the variable name used in the stash change to reflect its new expanded behaviour?

dracos: It's worse than that; filter_categories already exists, and is the list of categories to *show*.
Changing the request query param I'd be fine with, if not for that. I did try and change them all but got part way through and it all went horribly wrong so I rolled that bit back.

davea: Bah, nevermind then :)

dracos:
I have updated the multi select plugin, so this presumably needs the FMS CSS updating and any new "contained within" ID giving, and then that should be it?

zarino:
> this presumably needs the FMS CSS updating and any new "contained within" ID giving, and then that should be it?

Yes, I will do that right now.

PR #1548, opened 2016-10-04
WIP: Improved print view for lists of reports

zarino:
Fixes mysociety/fixmystreetforcouncils#21.

Todo:
- [x] Hide unneccessary items on list pages (eg: "Click the map" banner)
- [x] Insert hidden metadata into each list item, reveal on print
  - [x] Only show metadata if permission
  - [x] Not sure how to get the priority?
  - [x] Not sure how to get the extra category-related questions? (currently printed by a fairly complex loop around the `category_extras_fields` stuff in `_inspect.html`)
  - [x] Sort out the duplication of the `state_pretty` block
- [x] Check performance - not sure if checking permission is doing DB lookup each time
Currently looks like:

screen shot 2016-10-04 at 16 12 51

### DIFF
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -248,6 +248,15 @@ sub split_name {
     return { first => $first || '', last => $last || '' };
 }
 
+has body_permissions => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my $self = shift;
+        return [ $self->user_body_permissions->all ];
+    },
+);
+
 sub permissions {
     my ($self, $c, $body_id) = @_;
 
@@ -258,9 +267,7 @@ sub permissions {
 
     return unless $self->belongs_to_body($body_id);
 
-    my @permissions = $self->user_body_permissions->search({
-        body_id => $self->from_body->id,
-    })->all;
+    my @permissions = grep { $_->body_id == $self->from_body->id } @{$self->body_permissions};
     return { map { $_->permission_type => 1 } @permissions };
 }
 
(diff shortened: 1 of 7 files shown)


zarino:
@dracos / @davea – Done as much as I can on this – could you investigate the remaining points?

dracos:
Good news:
* I've included the extra details as a `dl` at the end; you'll want to move/style this :)
* I've sorted the duplication of `state_pretty`.
* I've fixed a template bug, and rebased a bit to get some other bug fixes.

Bad news:
* I'm afraid Chrome only prints one page of results; Safari only prints one result; Firefox prints no extra information at all :-/ 
* Putting the description in every item might lead to a large increase in page size, especially on around pages with lots of reports – can we do this only for planned report users, or only on `/my/planned`, or something? Note it's currently happening on e.g. the front page too, which probably isn't wanted.

zarino:
> I'm afraid Chrome only prints one page of results; Safari only prints one result; Firefox prints no extra information at all :-/

Weird. I'll investigate this today.

> Putting the description in every item might lead to a large increase in page size, especially on around pages with lots of reports – can we do this only for planned report users, or only on /my/planned, or something? Note it's currently happening on e.g. the front page too, which probably isn't wanted.

Limiting to just users with the planned reports permission sounds like a quick win here.

zarino:
@dracos Your two comments should now have been addressed.

By way of simplifying the print layout, I’ve removed the logo / branding from the top. This, combined with the right-floated map, saves a good deal of space. Do you think any cobrands (I'm thinking Zurich in particular here) will miss it?

dracos:
I think it'd be nice to have a logo, that wouldn't take up too much space and makes it easier to see where it's come from; reinstating `#site-header` doesn't put it back though, just an extra line appears? And Zurich does't have such a header (sometimes has a white blob?).

I think the right floated map needs some margin, otherwise e.g. photos butt up against it.

Looks very nice :)

zarino:
@dracos – The logo should be back now (ie: `#site-header` is `display:block` and `position:fixed`) and the map should have a bit of margin around it.

PR #1552, opened 2016-10-11
Add Bristol assets

dracos:
Todo:
- [x] HTTPS WFS server URL
- [x] Bug removing sidebar message when category changed to non-asset one; I'd added a check for layer visibility, so it was only called for one layer, but this means it's not being called when layer disappears
- [x] The above can probably be fixed by this one, which is tidying up the generalised JS to stop a few functions being defined one time per asset rather than once overall
- [x] Issue switching off the layer, sometimes the normal pin isn't reinstated (not new issue)
- [x] Sidebar needs to hide the attribute fields (usrn and asset_id).

bristolassets


web/cobrands/fixmystreet/assets.js
```diff
@@ -86,49 +87,50 @@ var add_streetlights = (function() {
         }
     }
 
-    function check_zoom_message_visiblity() {
```
davea: I always make that typo, the mistake is apparently invisble to me 😉

dracos:
I've updated 774a9ba so that the fields are hidden are also not checked for requirement (as you say, because there's nothing the user could do); the failure may then move backwards to the Open311 post but we'll see.

dracos:
"The "or pick a grit bin" looks a little lost here - could it be made bigger to draw the user's eye?" - found the style from Angus and moving to generic.

PR #1554, opened 2016-10-13
Let users hide their name on published reports and updates

### LINKED ISSUE #658: Allow users to make their reports anonymous, post-submission
Alternative to  #374 - allow users to change the anonymous state after making a report

zarino:
Fixes #658.

TODO:
- [x] Prototype
- [x] Get design feedback
- [x] Decide on non-javascript fallback
- [x] Make the buttons work

**How it looks for reports:**

hide-name

(Looks very similar for Updates – just swap the word "report" for "update")

### DIFF
--- a/perllib/FixMyStreet/App/Controller/My.pm
+++ b/perllib/FixMyStreet/App/Controller/My.pm
@@ -234,6 +234,38 @@ sub by_shortlisted {
     }
 }
 
+sub anonymize : Path('anonymize') {
+    my ($self, $c) = @_;
+    $c->forward('/auth/get_csrf_token');
+
+    my $object;
+    if (my $id = $c->get_param('problem')) {
+        $c->forward( '/report/load_problem_or_display_error', [ $id ] );
+        $object = $c->stash->{problem};
+    } elsif ($id = $c->get_param('update')) {
+        $c->stash->{update} = $object = $c->model('DB::Comment')->find({ id => $id });
+        $c->detach('/page_error_400_bad_request') unless $object;
+    } else {
+        $c->detach('/page_error_404_not_found');
+    }
+    $c->detach('/page_error_400_bad_request') unless $c->user->id == $object->user_id;
+    $c->detach('/page_error_400_bad_request') if $object->anonymous;
+
+    if ($c->get_param('hide') || $c->get_param('hide_everywhere')) {
+        $c->detach('/page_error_400_bad_request') unless $c->req->method eq 'POST';
+        $c->forward('/auth/check_csrf_token');
+        if ($c->get_param('hide')) {
+            $object->update({ anonymous => 1 });
+            $c->flash->{anonymized} = _('Your name has been hidden.');
+        } elsif ($c->get_param('hide_everywhere')) {
+            $c->user->problems->update({anonymous => 1});
+            $c->user->comments->update({anonymous => 1});
+            $c->flash->{anonymized} = _('Your name has been hidden from all your reports and updates.');
+        }
+        $c->res->redirect($object->url);
+    }
+}
+
 __PACKAGE__->meta->make_immutable;
 
 1;
--- a/templates/web/base/report/_main.html
+++ b/templates/web/base/report/_main.html
@@ -64,6 +64,10 @@
         [% INCLUDE 'report/_report_meta_info.html' %]
     </p>
 
+    [% IF anonymized ~%]
+        <p class="form-success">[% anonymized %]</p>
+    [% END ~%]
+
     [% INCLUDE 'report/_main_sent_info.html' %]
     [% mlog = problem.latest_moderation_log_entry(); IF mlog %]
         <p>[% tprintf(loc('Moderated by %s at %s'), mlog.admin_user, prettify_dt(mlog.whenedited)) %]</p>
--- a/templates/web/base/report/_report_meta_info.html
+++ b/templates/web/base/report/_report_meta_info.html
@@ -1,2 +1,5 @@
 [% problem.meta_line(c) | html %]
+[% IF c.user_exists AND c.user.id == problem.user_id AND !problem.anonymous %]
+    <small>(<a href="/my/anonymize?problem=[% problem.id | uri %]" class="js-hide-name">[% loc('Hide your name?') %]</a>)</small>
+[% END %]
 [%- IF !problem.used_map %]; <strong>([% loc('there is no pin shown as the user did not use the map') %])</strong>[% END %]
--- a/templates/web/base/report/update.html
+++ b/templates/web/base/report/update.html
@@ -24,7 +24,7 @@
             <div class="item-list__update-wrap">
             [% IF update.whenanswered %]
                 <div class="item-list__update-text">
-                    <p class="meta-2"> [% INCLUDE meta_line %] </p>
+                    <p class="meta-2">[% INCLUDE meta_line %]</p>
                 </div>
             [% ELSE %]
                 <a name="update_[% update.id %]" class="internal-link-fixed-header"></a>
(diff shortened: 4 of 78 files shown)


zarino:
Feedback in IRC:

> [09:31:36]  matthew says —	zarino: not sure why you'd want to remove name from 'all updates' but not all reports (or vice-versa) I guess
> [09:32:33]  struan says —	zarino: yay! that looks super useful.
> [09:32:35]  zarino says —	matthew: so maybe a "Hide my name across the whole site" that does reports and updates everywhere?
> [09:32:50]  Abi says —	zarino: exactly my feedback, yes please

davea:
It'd be good to only show the 'hide my name on all my reports/updates' link if the user actually has any other non-anonymous reports, would tidy it up a bit. Does the wording need clarifying to indicate it's only *previous* reports that will be made anonymous, not any future ones?

zarino:
> Does the wording need clarifying to indicate it's only previous reports that will be made anonymous, not any future ones?

I wonder whether clicking that button should also set the flag that unticks the "publish my name" checkbox by default on new report forms? Chances are, if a person wants to expunge their name from the entire site, they will want to keep it that way, so making sure their future reports are anonymous too (by default, at least) seems sensible.

dracos:
That will happen automatically, as the checkbox's default state is "whatever it was for their last report/update".

abimysoc:
> That will happen automatically, as the checkbox's default state is "whatever it was for their last report/update".

But they're probably using this anonymisation process because the last report/update they made *wasn't* anonymous...

zarino:
> But they're probably using this anonymisation process because the last report/update they made wasn't anonymous...

@abimysoc: Right, but once they’ve pressed the "Make all of my reports anonymous", the next time they load a reporting form, FMS will find their most recent report, discover it was made anonymously, and untick the "Show my name" checkbox. If you get what I mean.

templates/web/base/report/_main.html
```diff
@@ -60,6 +60,21 @@ <h1 class="moderate-display">[% problem.title | html %]</h1>
...
     </p>
 
+  [% IF c.user_exists AND c.user.id == problem.user_id AND !problem.anonymous %]
+    <!-- TODO: Do we want to ajax this in? -->
+    <!-- TODO: Should this work without javascript too? -->
+    <div class="box-warning hide-name-form">
+      [% IF problem.bodies_str %]
+        <p>[% tprintf(loc('Your name has already been sent to %s, but we can hide it on this page:'), problem.bodies_str) %]</p>
+      [% END %]
+        <!-- TODO: URL or form with button? -->
```
dracos: Anything with a side-effect should be a button, not a URL.

templates/web/base/report/_main.html
```diff
@@ -60,6 +60,21 @@ <h1 class="moderate-display">[% problem.title | html %]</h1>
         [% INCLUDE 'report/_report_meta_info.html' %]
     </p>
 
+  [% IF c.user_exists AND c.user.id == problem.user_id AND !problem.anonymous %]
+    <!-- TODO: Do we want to ajax this in? -->
+    <!-- TODO: Should this work without javascript too? -->
```
dracos: Yes to the latter :) I would put this in a separate template file (perhaps merged with the update one, it's similar enough) so that non-JS can have a link to a separate page that just shows that template.

templates/web/base/report/_report_meta_info.html
```diff
@@ -1,2 +1,6 @@
 [% problem.meta_line(c) | html %]
+[% IF c.user_exists AND c.user.id == problem.user_id AND !problem.anonymous %]
+  <!-- TODO: Should this link to a page with server-side form, for users without JavaScript? -->
```
dracos: For now, set it up as an `/about/` page (ie. http://fixmystreet.org/customising/static-pages/ ) and we can hook/move it from there.

templates/web/base/report/update.html
```diff
@@ -24,7 +24,24 @@ <h4 class="static-with-rule">[% loc('Updates') %]</h4>
             <div class="item-list__update-wrap">
             [% IF update.whenanswered %]
                 <div class="item-list__update-text">
-                    <p class="meta-2"> [% INCLUDE meta_line %] </p>
+                    <p class="meta-2">
+                        [% INCLUDE meta_line %]
```
dracos: This section is not necessary, there is no name shown for a questionnaire response (when update.whenanswered is true).

templates/web/base/report/update.html
```diff
@@ -45,10 +62,25 @@ <h4 class="static-with-rule">[% loc('Updates') %]</h4>
...
                         [% END %]
                     </p>
+                  [% IF c.user_exists AND c.user.id == update.user_id AND !update.anonymous %]
+                    <!-- TODO: Do we want to ajax this in? -->
+                    <!-- TODO: Should this work without javascript too? -->
+                    <div class="box-warning hide-name-form">
+                        <!-- TODO: URL or form with button? -->
+                        <a href="#">[% loc('Hide my name in this update') %]</a>
+                        <p>[% loc('Alternatively, we can hide your name on <strong>all of your updates</strong>:') %]</p>
+                        <!-- TODO: URL or form with button? -->
```
dracos: Same comments as for problems - separate joint template, link to separate page, form/buttons.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -438,6 +438,15 @@ $.extend(fixmystreet.set_up, {
     $('.js-contribute-as').change();
   },
 
+  hide_name: function() {
+      console.log('hide_name');
```
dracos: Debug lines.

zarino:
@dracos Ok, I think this branch is ready for a developer to come along and implement a page that handles the anonymisation form, to replace `/about/hide-my-name.html`.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -234,6 +234,35 @@ sub by_shortlisted {
     }
 }
 
+sub anonymize : Path('anonymize') {
```
davea: Should this controller require an HTTP `POST`, as it modifies (potentially lots of) reports?

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -234,6 +234,35 @@ sub by_shortlisted {
...
+        $c->stash->{update} = $object = $c->model('DB::Comment')->find({ id => $id });
+        $c->detach('page_error_400_bad_request') unless $object;
+    } else {
+        $c->detach('/page_error_404_not_found');
+    }
+    $c->detach('page_error_400_bad_request') unless $c->user->id == $object->user_id;
+    $c->detach('page_error_400_bad_request') if $object->anonymous;
+
+    if ($c->get_param('hide') || $c->get_param('hide_everywhere')) {
+        $c->forward('/auth/check_csrf_token');
```
davea: This has already been checked above.

dracos: It was got above (for displaying the form), but not checked :)

perllib/FixMyStreet/App/Controller/My.pm
dracos: It's also used to display the form, so I thought it didn't really matter (anyone malicious could try and construct a POST as easily as a GET, I mean), but yes it could check at the same point it checks the token. (Perhaps token check should always check POST, because you probably don't ever want a token in a GET…?)

dracos:
"Depending on the council's configuration, making all updates anonymous might affect those from Open311, causing confusion." – hopefully no-one is using a system Open311 user to leave updates, we should probably try and stop any that are, but yes, perhaps don't have the "all things" button for council users, that shouldn't be an issue. I'll have to remind myself of the possible combinations.
"As a user it'd be nice to see some kind of confirmation screen that my stuff has now been anonymized if that's the button I've clicked." – yeah, I thought taking them back to the page where they can see it is now anonymous would hopefully be enough, but it should probably show a message as well.

PR #1555, opened 2016-10-14
Fix list of response templates for a problem.

dracos:
Instead of returning all templates for the problem's body, return those
for the problem's category (or with no category).

Fixes mysociety/fixmystreetforcouncils#100

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -658,13 +658,15 @@ order of title.
 =cut
 
 sub response_templates {
-    my $problem = shift;
-    return $problem->result_source->schema->resultset('ResponseTemplate')->search(
+    my $self = shift;
+    return $self->result_source->schema->resultset('ResponseTemplate')->search(
         {
-            body_id => $problem->bodies_str_ids
+            'me.body_id' => $self->bodies_str_ids,
+            'contact.category' => [ $self->category, undef ],
         },
         {
-            order_by => 'title'
+            order_by => 'title',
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -659,12 +659,23 @@ order of title.
...
+            body_id => $problem->bodies_str_ids,
+            category => $problem->category
+        }, {
+            columns => [ 'id' ],
+        }
+    )->get_column('id')->all;
+    return $schema->resultset('ResponseTemplate')->search(
+        {
+            body_id => $problem->bodies_str_ids,
+            contact_id => [ \@cids, undef ],
```
davea: I'm not sure about including templates with no category set here. Is your intention that not ticking any category in the template admin form implicitly associate it with all categories? Would a 'select all' button in the template editor (and perhaps validation to make sure at least one category is selected) make this clearer?

dracos: Yes, I read it as selecting no categories was the same as not having made a filter, so applied to all, and would be a generic "applies anywhere" template. That seemed easier than having to select all the categories (a Select/Clear all would help there, yeah). I guess as you can't delete templates through the interface, no categories is a way to have it not shown, though; I think the best solution is probably some help text :) As either reading of nothing selected would be okay, depends what we want to do.

dracos:
"OOI would a similar query to that used in response_priorities apply here, instead of needing to perform two separate queries?" – yep, should no, I was struggling with basic SQL last week!

PR #1557, opened 2016-10-17
Admin category form improvements

davea:
Implements @zarino's suggestions for improving the category editor form in the admin:

 - Use the same template for adding/editing contacts
 - Hide Open311 API key etc. fields unless send method requires them
 - 'Note' is the last field on the form, and is relabelled 'Summarise your changes' for clarity
 - Required fields are marked as such in HTML, for client-side validation

screen shot 2016-10-17 at 16 11 09

Fixes mysociety/fixmystreetforcouncils#67 and mysociety/fixmystreetforcouncils#71.


web/js/fixmystreet-admin.js
```diff
@@ -23,9 +24,11 @@ $(function(){
         }
     }
 
-    if ($open311_only) {
-        $('#send_method').on('change', hide_or_show_open311);
-        hide_or_show_open311(true);
+    if ($('.admin-open311-only')) {
```
dracos: I think this needs a `.length`

templates/web/base/admin/contact-form.html
```diff
@@ -0,0 +1,137 @@
...
+      <p>
+        [% loc('Choose a <strong>category</strong> name that makes sense to the public (e.g., "Pothole", "Street lighting") but is helpful
+                to the body too. These will appear in the drop-down menu on the report-a-problem page.') %]
+        <br>
+        [% loc("If two or more bodies serve the same location, FixMyStreet combines identical categories into a single entry in
+                the menu. Make sure you use the same category name in the bodies if you want this to happen.") %]
+      </p>
+    </div>
+    <p>
+      <strong>[% loc('Category') %] </strong><input type="text" class="form-control" name="category" size="30" value="[% contact.category | html %]" required>
```
dracos: The category isn't editable on the edit page; it certainly would be nice to, somehow (issues with problem table notwithstanding), but I think here it will just then create a new contact with the new data given.

templates/web/zurich/admin/contact-form.html
```diff
@@ -0,0 +1,72 @@
+<form method="post" action="[% c.uri_for('body', body_id ) %]" enctype="application/x-www-form-urlencoded" accept-charset="utf-8" id="category_edit">
+    <p><strong>[% loc('Category:') %] </strong>[% contact.category | html %]
+    <input type="hidden" name="category" value="[% contact.category | html %]" >
+    <input type="hidden" name="token" value="[% csrf_token %]" >
+
+    <p><strong>[% loc('Email:') %] </strong>
+    <input type="text" class="form-control" name="email" value="[% contact.email | html %]" size="30">
+
+    <p>
+      [% IF c.cobrand.moniker != 'zurich' %]
```
dracos: If we're having a Zurich specific version, it doesn't need these bits :)

templates/web/zurich/admin/contact-form.html
```diff
@@ -0,0 +1,72 @@
...
+          [% loc('Reputation threshold:') %]
+          <input type="text" class="form-control" name="reputation_threshold" id="reputation_threshold"
+            value="[% contact.get_extra_metadata('reputation_threshold') | html %]" size="30">
+        </label>
+      </p>
+    [% END %]
+
+    <p><strong>[% loc('Note:') %] </strong><textarea class="form-control" name="note" rows="3" cols="40"></textarea>
+
+    [% IF body.can_be_devolved %]
```
dracos: Presumably this won't happen in Zurich either.

templates/web/base/admin/contact-form.html
davea: Good catch, thanks.

davea:
Thanks for the review - I've fixed up those various bits and it should be good to go now.

PR #1559, opened 2016-10-18
Add ‘description’ field to response priorities

davea:
- Adds an internal 'description' field to response priorities in the admin
 - Description is shown on the response priorities list table
 - Adds 'select all/select none' UI atop categories checklist (also for response templates)
 - Matches behaviour introduced in #1555 - empty category list means priority applies to all categories

For mysociety/fixmystreetforcouncils#66.


dracos:
:+1:

PR #1560, opened 2016-10-19
Resend report if changing category also changes body

davea:
Changing the category of a report (via the inspect/manage page or via the admin) might also mean the responsible body for that report has changed.

This PR adds functionality causing the report to be sent again if the body changes.

For mysociety/fixmystreetforcouncils#44.


dracos:
Code looks fine, but please could we have tests for the resending and the redirect :)

PR #1562, opened 2016-10-20
Have report page show inspect form by default

dracos:
This saves an inspector having to click a button to see the form.


dracos:
Yeah, I don't see an issue with having that, we can see if any confusion arises.

PR #1563, opened 2016-10-21
Assign users to categories

davea:
- Users can be assigned to one or more of their parent body's categories
- The 'all reports' link is prefiltered with the user's categories, if they have any, on the Oxfordshire cobrand

Left to do:
- [x] Prefilter /around when entering a postcode/location
- [x] Check other places in the site where it makes sense to prefilter

For mysociety/fixmystreetforcouncils#69

### DIFF
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -391,4 +391,19 @@ sub update_reputation {
     $self->update;
 }
 
+has categories => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my $self = shift;
+        return [] unless $self->get_extra_metadata('categories');
+        my @categories = $self->result_source->schema->resultset("Contact")->search({
+            id => $self->get_extra_metadata('categories'),
+        }, {
+            order_by => 'category',
+        })->get_column('category')->all;
+        return \@categories;
+    },
+);
+
 1;
--- a/templates/web/base/admin/user-form.html
+++ b/templates/web/base/admin/user-form.html
@@ -142,6 +142,10 @@
             <li>
               [% group.key %]
               <ul class="no-bullets">
+                <li>
+                  (<a href="#" data-select-all>[% loc('all') %]</a> /
+                  <a href="#" data-select-none>[% loc('none') %]</a>)
+                </li>
                 [% FOREACH permission IN group.value %]
                   <li>
                     <label class="inline">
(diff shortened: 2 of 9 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -391,4 +391,17 @@ sub update_reputation {
...
+sub categories {
+    my $self = shift;
+    return [] unless $self->get_extra_metadata('categories');
+    my @categories = $self->result_source->schema->resultset("Contact")->search({
+        id => $self->get_extra_metadata('categories'),
+    }, {
+        columns => [ 'category' ],
+        order_by => 'category',
+    })->all;
+    my $result = [ map( { $_->category } @categories ) ];
```
dracos: You can probably use get_column here, something like `my @categories = search({ id => …}, { order_by => 'category' })->get_column('category')->all; my $result = \@categories;`

templates/web/base/admin/user-form.html
```diff
@@ -71,6 +71,19 @@
...
+        [% IF contacts AND c.cobrand.moniker != 'zurich'%]
+          <li class="js-user-categories">
+            <div class="admin-hint">
+              <p>
+                [% loc(
+                  "Authorised staff users can be associated with the categories in which they operate.")
+                %]
+              </p>
+            </div>
+            [% INCLUDE 'admin/category-checkboxes.html' %]
```
dracos: Came in useful, this template!

templates/web/base/footer.html
```diff
@@ -25,7 +25,7 @@
                             %]span[% ELSE %]a href="[% base %]/my"[% END
                             %]>[% c.user_exists ? loc("Your account") : loc("Sign in") %]</[% c.req.uri.path == '/my' ? 'span' : 'a' %]></li>[%
                         %]<li><[% IF c.req.uri.path == '/reports'
-                            %]span[% ELSE %]a href="[% base %]/reports"[% END
+                            %]span[% ELSE %]a href="[% base %]/reports[% IF c.user_exists %]?filter_category=[% c.user.categories.join(",") | uri | html %][% END %]"[% END
```
dracos: As with the front page field you've added, do you want to only output this extra part of there are categories to include?

perllib/FixMyStreet/DB/Result/User.pm
davea: Ooh that's nicer, thanks!

PR #1564, opened 2016-10-21
Invert display of areas on body pages.

dracos:
This makes it easier to use the map, as the land outside the area is shaded, rather than the area itself.

It also makes zoom-to-extent/area_format work for bodies with multiple areas.

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -333,8 +333,26 @@ var fixmystreet = fixmystreet || {};
 
     function onload() {
         if ( fixmystreet.area.length ) {
+            var extent = new OpenLayers.Bounds();
+            var lr = new OpenLayers.Geometry.LinearRing([
+                new OpenLayers.Geometry.Point(20E6,20E6),
+                new OpenLayers.Geometry.Point(10E6,20E6),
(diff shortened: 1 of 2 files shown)


web/js/map-OpenLayers.js
```diff
@@ -333,23 +333,61 @@ var fixmystreet = fixmystreet || {};
...
+                new OpenLayers.Geometry.Point(0,-20E6),
+                new OpenLayers.Geometry.Point(10E6,-20E6),
+                new OpenLayers.Geometry.Point(20E6,-20E6),
+                new OpenLayers.Geometry.Point(20E6,0)
+            ]);
+            var loaded = 0;
+            var new_geometry = new OpenLayers.Geometry.Polygon(lr);
             for (var i=0; i<fixmystreet.area.length; i++) {
-                var area = new OpenLayers.Layer.Vector("KML", {
+                area = new OpenLayers.Layer.Vector("KML", {
```
zarino: Should this be `var area = …` ?

dracos: Doh, yes, I briefly tried it in a `$.each` instead of a for loop, so it was okay then, will fix!

PR #1567, opened 2016-10-25
Add simple way to return to "everything" in list filters

zarino:
Adds the features from https://github.com/mysociety/jquery-multi-select/pull/7 to FixMyStreet.

Fixes mysociety/fixmystreetforcouncils#115, and gives us a nice user-friendly "presets" interface we could add into #1563 too.

Looks like this in the Oxfordshire cobrand:

screen shot 2016-10-25 at 12 08 17


zarino:
@dracos – Could you check this over?

dracos:
🚂

PR #1568, opened 2016-10-26
Leave a public update when admin changes category.

### LINKED ISSUE #1544: Leave update when body user changes category
Would it make sense to provide body users with a drop down menu that they can use to select the category for a problem when making updates? This would allow body users to correct mistakes made by the reporting user (instead of the FixaMinGata admin having to do it), or save the user the hassle of having to create a new report.

If this would make sense, it would be nice if we could somehow allow the body user to also resend the e-mail, so that the report is being received by the correct entity.

Thanks!

dracos:
And add ability for admin users to italic updates, to be used by this.
Fixes #1544.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -841,6 +841,7 @@ sub report_edit_category : Private {
     my ($self, $c, $problem) = @_;
 
     if ((my $category = $c->get_param('category')) ne $problem->category) {
+        my $category_old = $problem->category;
         $problem->category($category);
         my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
         my @new_body_ids = map { $_->body_id } @contacts;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -849,6 +850,16 @@ sub report_edit_category : Private {
             $problem->whensent(undef);
         }
         $problem->bodies_str(join( ',', @new_body_ids ));
+        $problem->add_to_comments({
+            text => '/' . sprintf(_('Category changed from ‘%s’ to ‘%s’'), $category_old, $category) . '/',
```
davea: Guessing you mean to use `*` here instead of `/`

perllib/FixMyStreet/App/View/Web.pm
```diff
@@ -98,6 +99,24 @@ sub _space_slash {
...
+
+=cut
+
+sub markup_factory {
+    my ($c, @args) = @_;
+    my $update = $c->stash()->get('update');
+    my $user = $update ? $update->user : undef;
+    return sub {
+        my $text = shift;
+        return $text unless $user && ($user->from_body || $user->is_superuser);
```
davea: This (and the `$update` stuff above) is all quite specific to the changing category feature. Would it be better to handle this logic in the template instead? I'm just a bit worried that this new attractive looking `markup` filter might cause confusion because it won't actually work everywhere you might expect it to.

PR #1570, opened 2016-10-28
Make sure inspect form shown if report pulled in.

dracos:
And fix an issue where removed shortlisted users were shown on report page.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -179,6 +179,40 @@ fixmystreet.resize_to = {
   }
 };
 
+fixmystreet.geolocate = {
+    setup: function(success_callback) {
+        $('#geolocate_link').click(function(e) {
+            var $link = $(this);
+            e.preventDefault();
+            // Spinny thing!
+            if ($('.mobile').length) {
+                $link.append(' <img src="/cobrands/fixmystreet/images/spinner-black.gif" alt="" align="bottom">');
+            } else {
+                var spincolor = $('<span>').css("color","white").css("color") === $('#front-main').css("background-color") ? 'white' : 'yellow';
+                $link.append(' <img src="/cobrands/fixmystreet/images/spinner-' + spincolor + '.gif" alt="" align="bottom">');
+            }
+            geo_position_js.getCurrentPosition(function(pos) {
+                $link.find('img').remove();
+                success_callback(pos);
+            }, function(err) {
+                $link.find('img').remove();
+                if (err.code === 1) { // User said no
+                    $link.html(translation_strings.geolocation_declined);
+                } else if (err.code === 2) { // No position
+                    $link.html(translation_strings.geolocation_no_position);
+                } else if (err.code === 3) { // Too long
+                    $link.html(translation_strings.geolocation_no_result);
+                } else { // Unknown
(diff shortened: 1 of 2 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -179,6 +179,40 @@ fixmystreet.resize_to = {
...
+                var spincolor = $('<span>').css("color","white").css("color") === $('#front-main').css("background-color")? 'white' : 'yellow';
+                $link.append(' <img src="/cobrands/fixmystreet/images/spinner-' + spincolor + '.gif" alt="" align="bottom">');
+            }
+            geo_position_js.getCurrentPosition(function(pos) {
+                $link.find('img').remove();
+                success_callback(pos);
+            }, function(err) {
+                $link.find('img').remove();
+                if (err.code == 1) { // User said no
+                    $link.html(translation_strings.geolocation_declined);
```
zarino: When `#geolocate_link` is used on the Inspector page, this error message ("You declined; please fill in the box above") doesn't really make sense any more, since there’s no box to fill in. Do we need to make the error message configurable?

screen shot 2016-11-03 at 16 15 22

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -179,6 +179,40 @@ fixmystreet.resize_to = {
   }
 };
 
+fixmystreet.geolocate = {
+    setup: function(success_callback) {
+        $('#geolocate_link').click(function(e) {
+            var $link = $(this);
+            e.preventDefault();
+            // Spinny thing!
+            if($('.mobile').length){
```
zarino: Since we’ve taken the plunge and moved this chunk of code from its original location, might we also clean up the spacing around these the brackets in this `if`/`else` block?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -179,6 +179,40 @@ fixmystreet.resize_to = {
...
+            }else{
+                var spincolor = $('<span>').css("color","white").css("color") === $('#front-main').css("background-color")? 'white' : 'yellow';
+                $link.append(' <img src="/cobrands/fixmystreet/images/spinner-' + spincolor + '.gif" alt="" align="bottom">');
+            }
+            geo_position_js.getCurrentPosition(function(pos) {
+                $link.find('img').remove();
+                success_callback(pos);
+            }, function(err) {
+                $link.find('img').remove();
+                if (err.code == 1) { // User said no
```
zarino: JavaScript stickler inside me sort of expects these `==` to be `===` – wonder whether it would be confusing to change that in this PR?

dracos: I don't know if the error code is actually returned as a string, so perhaps best to leave it :)

web/cobrands/fixmystreet/fixmystreet.js
dracos: Yes! Hooray!

web/cobrands/fixmystreet/fixmystreet.js
dracos: Hmm, I think I'll ignore this for now, as it's an admin function that hopefully won't arise, and see.

PR #1572, opened 2016-11-01
Add sort order option to list pages.

dracos:
Needs a bit of design to fit in with the multiple ones, I think, but is up and running okay.

sortingfms

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -365,6 +365,8 @@ sub check_canonical_url : Private {
 sub load_and_group_problems : Private {
     my ( $self, $c ) = @_;
 
+    $c->forward('stash_report_sort', [ $c->cobrand->reports_ordering ]);
+
     my $page = $c->get_param('p') || 1;
     # NB: If 't' is specified, it will override 'status'.
     my $type = $c->get_param('t') || 'all';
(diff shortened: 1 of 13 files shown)


zarino:
@dracos I’ve tidied up the styling of the `select` so that it better matches the `.multi-select-button` above it. Annoyingly, in Safari, the select is given a grey "buttony" background, but if we stop it adding that, it also removes the dropdown arrows and padding, which we really want to keep. So I’ve left it as-is for now.

web/js/map-OpenLayers.js
```diff
@@ -297,6 +301,7 @@ var fixmystreet = fixmystreet || {};
         var qs = parse_query_string();
         var filter_categories = replace_query_parameter(qs, 'filter_categories', 'filter_category');
         var filter_statuses = replace_query_parameter(qs, 'statuses', 'status');
+        var filter_sort = replace_query_parameter(qs, 'sort', 'sort');
```
davea: Ditto the above comment about removing `filter_`.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -502,6 +504,27 @@ sub stash_report_filter_status : Private {
...
+
+    my $sort = $c->get_param('sort') || $default;
+    $sort = $default unless $sort =~ /^((updated|created)-(desc|asc)|comments-desc)$/;
+    $sort =~ /^(updated|created|comments)-(desc|asc)$/;
+    my $order_by = $types{$1} || $1;
+    my $dir = $2;
+    $order_by = { -desc => $order_by } if $dir eq 'desc';
+
+    $c->stash->{filter_sort} = $sort;
+    $c->stash->{filter_sort_order} = $order_by;
```
dracos: No need to apologies, you're quite correct! Renamed.

PR #1574, opened 2016-11-08
Version 2.0 bumph

dracos:
Make sure changelog is up to date, update translations.


BenJam:
minor comment: on two installs here the /admin/config page is printing 'unknown'. I cna create an issue but I know you'll see this.

dracos:
Aha, this happens if you’re running the default cobrand and haven’t overridden any templates. If you create an empty directory `templates/web/default` it should start working.

PR #1582, opened 2016-11-24
Add manual reordering of user's shortlist.

dracos:
Up/down/top buttons; JS will be added to allow nicer clicks/ dragging
(the code here can cope with an item moving to any point in the list).
- [x] Design (of list items in general)
- [x] Intercept up/down buttons on shortlist and move in JS/ AJAX request
- [x] Make add/take/remove buttons AJAX call the right script (same as button on report page).
- [x] Update add/remove/take buttons when clicked (half done)
- [ ] I find it confusing when you click to add/remove and the icon goes to the new hover state, which sort of implies that it hasn’t worked, when it has, iyswim.


zarino:
Icons added!

screen shot 2016-11-25 at 17 22 50

screen shot 2016-11-25 at 17 23 51

zarino:
Right now the styles are defined in `_report_list_pins.scss`, which is only included in a few cobrands’ CSS. Should we import `_report_list_pins.scss` for every cobrand?

dracos:
"Should we import _report_list_pins.scss for every cobrand?" – that seems needless if some cobrands don't use its (current) contents. Why not define indented in the general style?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -390,6 +397,82 @@ $.extend(fixmystreet.set_up, {
...
+        $item.insertBefore( $item.prev() );
+      } else if ('shortlist-remove' === whatUserWants) {
+          $submitButton.attr('class', 'item-list__item__shortlist-add');
+          $submitButton.attr('name', 'shortlist-add');
+          // TODO Change title/value to Add
+      } else if ('shortlist-take' === whatUserWants) {
+          $submitButton.attr('class', 'item-list__item__shortlist-remove');
+          $submitButton.attr('name', 'shortlist-remove');
+          // TODO Change title/value to Remove
+      } else if ('shortlist-add' === whatUserWants) {
```
davea: This action in this block is the same as the one above, can they be combined?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -390,6 +397,82 @@ $.extend(fixmystreet.set_up, {
...
+
+      $.ajax({
+        url: '/my/planned/change',
+        type: 'POST',
+        data: data
+      }).fail(function() {
+        // Undo the UI changes we made.
+        if ('shortlist-down' === whatUserWants) {
+          $item.insertBefore( $item.prev() );
+        } else if ('shortlist-up' === whatUserWants) {
```
davea: What about `shortlist-add`/`-remove`/`-take`?

dracos:
"Having the shortlist icon in every report list might be overkill - especially because it makes the click target for actually viewing the report smaller” – yeah, shall we have it only on /reports do you think?

dracos:
"The background colour of front page report lists is wrong now” - .item-list__item--with-pin used to have background: none that has gone, dunno if deliberately. (Though I can’t see background colour difference on my monitor!)

web/cobrands/fixmystreet/fixmystreet.js
dracos: shortlist-take doesn’t exist as a button any more, it’s only for a different icon, so yes, one can be removed :)

web/cobrands/fixmystreet/fixmystreet.js
dracos: Whoops, yes, there, and all this tidied up (it wasn’t working on around page either as that needs the ID putting back in the name).

dracos:
Programming issues hopefully fixed; will leave design ones to Zarino :)

zarino:
> The UI for shortlisting a report from the report detail page is now very different from the lists. Can we include the green star on the buttons to make it obvious it's the same action?

Done!

> "Having the shortlist icon in every report list might be overkill - especially because it makes the click target for actually viewing the report smaller”

The shortlist icons don’t take up *that* much space (at least, not until the sorting arrows are also added on the /shortlist page). I’d be inclined to leave them on all pages, if only for the consistency of knowing, if a report is shown, it’s always shown the same way. It would be weird to be able to add to the shortlist from one type of page, but not from another, almost identical page.

 > "The background colour of front page report lists is wrong now” - .item-list__item--with-pin used to have background: none that has gone, dunno if deliberately.

Yeah, I removed the background, to simplify the hover state logic. I’ve just pushed a fix for the erroneous background colour on the homepage list items.

PR #1587, opened 2016-12-01
Add 'state' to response templates

davea:
This allows response templates to be associated with a problem state in the admin, and changing a report's state when inspecting/managing a report will automatically prefill the update field with the appropriate response template for the new state, if one exists.

For mysociety/fixmystreetforcouncils#75

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -675,12 +675,20 @@ $.extend(fixmystreet.set_up, {
     // The inspect form submit button can change depending on the selected state
     $("#report_inspect_form [name=state]").change(function(){
         var state = $(this).val();
-        var $submit = $("#report_inspect_form input[type=submit]");
+        var $inspect_form = $("#report_inspect_form");
+        var $submit = $inspect_form.find("input[type=submit]");
         var value = $submit.attr('data-value-'+state);
         if (value !== undefined) {
             $submit.val(value);
         } else {
             $submit.val($submit.data('valueOriginal'));
+        }
+
+        // We might also have a response template to preselect for the new state
+        var $select = $inspect_form.find("select.js-template-name");
+        var $option = $select.find("option[data-problem-state='"+state+"']").first();
+        if ($option.length) {
+            $select.val($option.val()).change();
         }
     }).change();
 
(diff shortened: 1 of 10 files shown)


dracos:
How does the `auto_response` flag interact with this? Is that still necessary?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -980,9 +986,20 @@ $.extend(fixmystreet.set_up, {
...
+      $('.js-template-name').each(function() {
+          var $input = $('#' + $(this).data('for'));
+          $input.change(function() { $(this).data('dirty', !/^\s*$/.test($(this).val())); });
+      });
+
       $('.js-template-name').change(function() {
           var $this = $(this);
-          $('#' + $this.data('for')).val($this.val());
+          var $input = $('#' + $this.data('for'));
+          if (!$input.data('dirty')) {
```
dracos: Perhaps here could drop the `dirty` flag and use `if ($input[0].defaultValue != $input.val())` – this wouldn't have the 'empty field' functionality but otherwise I think is the same.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -661,6 +661,12 @@ $.extend(fixmystreet.set_up, {
         } else {
             $submit.val($submit.data('valueOriginal'));
         }
+
+        // We might also have a response template to preselect for the new state
+        var $option = $("#report_inspect_form select.js-template-name option[data-problem-state='"+state+"']").first();
```
dracos: Probably makes little/no difference, but could put `$('#report_inspect_form select.js-template-name')` in a variable first to use here and in if below.

davea:
Good question on the auto-responses - that functionality hasn't ever actually been implemented, so the answer is: it doesn't interact with it at all 😉

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -659,13 +659,20 @@ $.extend(fixmystreet.set_up, {
...
+        var $submit = $inspect_form.find("input[type=submit]");
         var value = $submit.attr('data-value-'+state);
         if (value !== undefined) {
             $submit.val(value);
         } else {
             $submit.val($submit.data('valueOriginal'));
         }
+
+        // We might also have a response template to preselect for the new state
+        var $option = $inspect_form.find("select.js-template-name option[data-problem-state='"+state+"']").first();
```
dracos: Still finding the select twice :)

web/cobrands/fixmystreet/fixmystreet.js
dracos: Yep, got you.

PR #1588, opened 2016-12-02
Offline inspection of shortlisted reports

dracos:
This PR adds some offline support for FixMyStreet:
* Static assets [1] are cached using an appcache manifest (inspectors use e.g. Lumia 650s with Edge), included in an iframe to prevent the HTML being cached (thanks to Lanyrd for that trick);
* When a report is shortlisted, or a visit to `/my/planned`, shortlisted reports are cached locally in localStorage along with their images and static maps [2];
* The fallback offline page gives a nice error (albeit we can't tell why it's being shown), and shows any offline reports present;
  
* Cached reports can be shown (the fallback offline page loads them from localStorage) with reduced functionality;
  
* The inspect form can be filled and submitted, and the submission is stored offline;
  
* A banner tells you are offline or if you have offline saved forms to submit, and clicking performs the uploads. [4]
  

For mysociety/fixmystreetforcouncils#20. It also fixes mysociety/fixmystreetforcouncils#123

TODO:
- [x] "Back to all reports" could go back to `/my/planned` by default (offline makes no difference, but when you go back online, bit odd);
- [x] [2] Progress messages on /my/planned as reports are cached, or when report is shortlisted;
- [x] Deal with CSRF somehow;
- [x] It'd nice to have a confirmation message when submitting a form when offline, so I know my input has been saved.
- [x] Could we indicate *which* problems on the shortlist have forms saved, to make it clearer which problems I've already dealt with? (It could read the form actions when printing the offline shortlist.)
- [x] Is the form supposed to redisplay any saved data? Unless I'm doing something wrong it doesn't seem to. (It could deserialize the most recent saved form…)
- [x] Allow saved timestamp for update comment.

Other footnotes:
* [1] CSS and JavaScript – I think CSS images are decorative and unnecessary for this functionality, and it would be much trickier to include them;
* [4] Detection of going online could auto-upload the forms rather than only update the banner (though we don't have background sync, so perhaps leaving to user is more straightforward/clear).


davea:
This is really awesome \o/

A few thoughts after playing around a bit:

 - It'd nice to have a confirmation message when submitting a form when offline, so I know my input has been saved.
 - Could we indicate *which* problems on the shortlist have forms saved, to make it clearer which problems I've already dealt with?
 - Is the form supposed to redisplay any saved data? Unless I'm doing something wrong it doesn't seem to.
 - The progress indication in the banner to upload saved reports when you go back online is a nice touch :)
 - It's probably more trouble than it's worth to make the update timestamps display the time the inspector submitted the offline form rather than when the server received them, right?

dracos:
Confirmation message – yes, the banner updates, but it should do something more obvious, definitely.
Which problems have saved forms – currently, it's just storing a list of forms, but that should be possible, yeah, it could read the form actions when printing the offline shortlist.
Redisplay – no, it just displays the report page from the cache each time. I'm not sure how to do this.
Update timestamps – if this is for the update comment, this should be fine, it can just store it in a hidden field and use that instead of the current timestamp? I guess potential for abuse is there, so perhaps only for people with the inspect permission.

dracos:
Okay, so now additionally: "Back to all reports" goes back to `/my/planned`; you should get a pogress message when it's caching reports; it should take you back to shortlist and give a confirmation message when you submit the inspect form; it should indicate reports with saved forms in the offline shortlist; it should use the offline saved time for the update subsequently created.

Only things really left are dealing with very old saved reports (ie. the CSRF token will have expired), and updating the form if viewed again with the latest serialised saved form data.

dracos:
It should probably also drop form submissions that fail to submit when back online otherwise there's no way of getting rid of them, and they're not going to magically start working again for whatever reason they've failed.

dracos:
It should now 'remember' saved form data if you go back to the same report. One thing left, CSRF tokens.

web/cobrands/fixmystreet/offline.js
```diff
@@ -0,0 +1,415 @@
...
+            $('#js-clear-localStorage').click(function() {
+                fixmystreet.offline.removeReports(fixmystreet.offlineData.getCachedUrls());
+                fixmystreet.offlineData.clearForms();
+                localStorage.removeItem('/my/planned');
+                alert('Offline data cleared');
+            });
+        }
+    }
+    fixmystreet.offlineBanner.make(true);
+} else {
```
davea: This doesn't seem to limit the offline preparation to logged-in inspectors, so everyone's browser will add the iframe on every page load.

PR #1590, opened 2016-12-13
Update priorities in inspect form on category change.

dracos:
Different categories may have a different list of priorities, so store them all and update as the category changes.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -312,6 +312,28 @@ sub inspect : Private {
     $c->stash->{categories} = $c->forward('/admin/categories_for_point');
     $c->stash->{report_meta} = { map { $_->{name} => $_ } @{ $c->stash->{problem}->get_extra_fields() } };
 
+    my %category_body = map { $_->category => $_->body_id } map { $_->contacts->all } values %{$problem->bodies};
+
+    my @priorities = $c->model('DB::ResponsePriority')->for_bodies($problem->bodies_str_ids)->all;
+    my $priorities_by_category = {};
+    foreach my $pri (@priorities) {
+        my $any = 0;
+        foreach ($pri->contacts->all) {
+            $any = 1;
+            push @{$priorities_by_category->{$_->category}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
+        }
+        if (!$any) {
+            foreach (grep { $category_body{$_} == $pri->body_id } @{$c->stash->{categories}}) {
+                push @{$priorities_by_category->{$_}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
+            }
+        }
+    }
+    foreach (keys %{$priorities_by_category}) {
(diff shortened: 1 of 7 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -573,6 +573,16 @@ $.extend(fixmystreet.set_up, {
...
+        // And update the associated priority list
+        var priorities = $("form#report_inspect_form " + selector).data('priorities');
+        var $select = $('#problem_priority'),
+            curr_pri = $select.val();
+        $select.find('option:gt(0)').remove();
+        $.each(priorities.split('&'), function(i, kv) {
+            kv = kv.split('=', 2);
+            $select.append($('<option>', { value: kv[0], text: decodeURIComponent(kv[1]) }));
+        });
+        $select.val(curr_pri);
```
davea: If there are no priorities found, this sets the label of the control to 'undefined':
screen shot 2016-12-15 at 15 20 43

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -312,6 +312,26 @@ sub inspect : Private {
...
+            foreach (@{$c->stash->{categories}}) {
+                push @{$priorities_by_category->{$_}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
+            }
+        }
+    }
+    foreach (keys %{$priorities_by_category}) {
+        $priorities_by_category->{$_} = join('&', @{$priorities_by_category->{$_}});
+    }
+
+    $c->stash->{priorities_by_category} = $priorities_by_category;
```
davea: There's some deduplication required here, as it's possible for the same priority to appear multiple times (e.g. `<p data-category="Drainage" class="hidden" data-priorities="2=Cat%201A&2=Cat%201A&2=Cat%201A&3=Cat%201B&4=Cat%204">`):
screen shot 2016-12-15 at 15 21 45

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -312,6 +312,26 @@ sub inspect : Private {
     $c->stash->{categories} = $c->forward('/admin/categories_for_point');
     $c->stash->{report_meta} = { map { $_->{name} => $_ } @{ $c->stash->{problem}->get_extra_fields() } };
 
+    my @priorities = $c->model('DB::ResponsePriority')->for_bodies($problem->bodies_str_ids)->all;
```
davea: I think this is causing some two-tier weirdness with the wrong priorities being shown if you change to a category in a different council:
priorities
Selecting the priorities by the report's location, not its `bodies_str`, would probably fix this.

web/cobrands/fixmystreet/fixmystreet.js
dracos: Empty string splitting to one element array gets me every time!

perllib/FixMyStreet/App/Controller/Report.pm
dracos: Hmm, yes, it should only add the priority that is on all categories, to those categories in the same body as the priority.

PR #1597, opened 2017-01-10
Hide 'Show my name publicly' when reporting on behalf of another user

pezholio:
For mysociety/fixmystreetforcouncils#140


davea:
Nice! I think it'd be clearer what's going on if the 'show my name publicly' tick box was always visible, and just simply unticked & disabled when reporting as another user. i.e., just the changes from 7d2b736d88938c8e846b3e00e9ce7cf4410f3825.

pezholio:
Yeah, good idea. I'll get on that 👍

pezholio:
@davea Happy to do this, but Github lets you squash and merge now - https://github.com/blog/2141-squash-your-commits

davea:
Alas, by way of self-hosting our repositories we can't take advantage of such niceties :)

pezholio:
Ah ha, no worries. Will squash the old fashioned way then 😄

PR #1600, opened 2017-01-10
Some fall out tidying

dracos:
This allows superusers to access `/my/planned` which was missing by mistake; and allows staging sites to turn off the appcache for easier debugging of things.

### DIFF
--- a/templates/web/base/admin/users.html
+++ b/templates/web/base/admin/users.html
@@ -26,7 +26,7 @@
         <td>[% PROCESS value_or_nbsp value=user.name %]</td> 
         <td><a href="[% c.uri_for( 'reports', search => user.email ) %]">[% PROCESS value_or_nbsp value=user.email %]</a></td> 
         <td>[% PROCESS value_or_nbsp value=user.from_body.name %]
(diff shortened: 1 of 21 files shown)


templates/web/base/admin/users.html
```diff
@@ -26,7 +26,7 @@
         <td>[% PROCESS value_or_nbsp value=user.name %]</td> 
         <td><a href="[% c.uri_for( 'reports', search => user.email ) %]">[% PROCESS value_or_nbsp value=user.email %]</a></td> 
         <td>[% PROCESS value_or_nbsp value=user.from_body.name %]
-            [% IF user.has_body_permission_to('moderate') %] * [% END %]
+            [% IF NOT user.is_superuser AND user.has_body_permission_to('moderate') %] * [% END %]
```
davea: Now we have multiple permissions, I wonder if this asterisk would be more use for identifying superusers?

dracos: Yeah, good idea.

dracos:
Merged at 940bd23 (altered config page to print out STAGING_FLAGS better).

PR #1601, opened 2017-01-11
Show user details for council report

pezholio:
For mysociety/fixmystreetforcouncils#136

PR #1603, opened 2017-01-13
Hide council user names for council updates

pezholio:
For mysociety/fixmystreetforcouncils#142 

In a similar way to #1601 this hides the user name when a council user has added an update. If a logged-in user has the `view_body_contribute_details` permission though, the user's name will show up in brackets.

This relies on #1601 to be merged too, so may be better adding on to that branch. They are distinct features though.

@dracos - what are your feels?


perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -264,6 +263,7 @@ sub meta_line {
     }
 
     if (!$update_state and $self->problem_state) {
+      my $state;
```
dracos: This is fine, just to note you could define and set it by combining with the line below.

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -6,6 +6,7 @@ package FixMyStreet::DB::Result::Comment;
 
 use strict;
 use warnings;
+use HTML::Entities;
```
dracos: We have a function in FixMyStreet::Template::html_filter that could be used here (only encodes the amp/gt/lt/'/" rathe than all entities, which means you don't need to worry whether it does the right thing for non-ASCII; it might do, I just haven't checked!).

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -223,4 +224,90 @@ __PACKAGE__->might_have(
...
+
+    if ($self->anonymous or !$self->name) {
+      $meta = sprintf( _( 'Posted anonymously at %s' ), Utils::prettify_dt( $self->confirmed ) )
+    } elsif ($self->user->from_body) {
+      my $user_name = encode_entities($self->user->name);
+      my $body = $self->user->body;
+      if ($body eq 'Bromley Council') {
+        $body = "$body <img src='/cobrands/bromley/favicon.png' alt=''>";
+      }
+      if (defined $c->user and $c->user->has_permission_to('view_body_contribute_details', $self->problem->bodies_str_ids)) {
```
dracos: First part of this could be $c->user_exists same as elsewhere.

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -223,4 +224,90 @@ __PACKAGE__->might_have(
...
+
+    my $update_state = '';
+
+    if ($self->mark_fixed) {
+      $update_state = _( 'marked as fixed' );
+    } elsif ($self->mark_open)  {
+      $update_state = _( 'reopened' );
+    } elsif ($self->problem_state) {
+      my $state;
+      $state = $self->problem_state_display;
```
dracos: You could combine the two lines above in one.

pezholio:
Cool - done! 👍

PR #1607, opened 2017-01-18
Close 'old' enquiries

pezholio:
For mysociety/fixmystreetforcouncils#51

This will mark all reports with the last update in 2015 as closed, and email the users with information on what problems have been closed, and the option to reopen them. Anything still open with a last update of 2014 or earlier gets closed with no email.

This could do with a code review, but would also be good to test against a corpus of real data before pointing it at live.

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -182,6 +182,8 @@ use Utils;
 use FixMyStreet::Map::FMS;
 use LWP::Simple qw($ua);
 use RABX;
+use URI;
+use URI::QueryParam;
 
 my $IM = eval {
     require Image::Magick;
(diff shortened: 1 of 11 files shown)


pezholio:
I think this is ready for review now :+1:

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -511,6 +513,30 @@ sub admin_url {
     return $cobrand->admin_base_url . '/report_edit/' . $self->id;
 }
 
+=head2 tokenised_url
+
+Return a url for this problem report that logs a user in
+
+=cut
+
+sub tokenised_url {
```
pezholio: I get what you're saying, but the issue there is that at the moment we've got dynamic attributes (namely `$user` and `$params`) that get passed in too. Is there a way round that?

templates/email/default/archive.html
```diff
@@ -0,0 +1,58 @@
...
+  <p style="[% p_style %]">
+    FixMyStreet is being updated in Oxfordshire to
+    improve how problems get fixed.
+  </p>
+  <p style="[% p_style %]">
+    As part of these updates, we are closing old reports that appear to be
+    resolved but remain open in the system.
+  </p>
+  <p style="[% p_style %]">
+    We noticed that you have [% report_count %] old reports on the system,
```
pezholio: Is there a Perl / Catalyst / tt2 way I can get around this?

templates/email/default/archive.html
```diff
@@ -0,0 +1,58 @@
+[%
```
pezholio: Hmmm... Moving the templates gives me `Template processing error: file error - archive.txt: not found` 😕

pezholio: Looking at the code `$cobrand->path_to_email_templates($lang_code)` seems to be returning `/templates/email/default` for some reason

pezholio: No wait, I got this. `FixMyStreet::override_config` is the key

templates/email/default/archive.html
pezholio: Ah, `[% nget('reports', 'report', report_count) %]` seems to do the trick 👍

Aliyadaulah:
hi please contact me at(aliyadaulah@gmail.com) for more about me from Aliya

PR #1610, opened 2017-01-21
fix #1604 strings possible to translate through gettext

p9s:
I added some fix, please review and merge

Thanks

PR #1611, opened 2017-01-21
fix #1604 strings possible to translate through gettext

### LINKED ISSUE #1604: Make "details" strings possible to translate through gettext
The following two files contain strings that are not translated through gettext:

* templates/web/base/report/new/form_report.html ("Public details")
* templates/web/base/report/new/form_user.html ("Private details")

Could these be made possible to translate?

Thanks!

p9s:


dracos:
Hi, I already merged your #1610 as 5277688 as mentioned there :-) If there's some issue, do explain further here or there.

PR #1615, opened 2017-01-27
Use span instead of div to wrap response template

davea:
Using a `div` inside a `p` isn't valid, so the DOM was being restructured by the
browser causing the 'save with public update' toggle on the inspect form to not
work.

Fixes mysociety/fixmystreetforcouncils#153


davea:
Yep, good thinking. I couldn't see any reason for the wrapper in Zürich or otherwise, so have gotten rid of it entirely.

PR #1620, opened 2017-01-30
Fix Shortlist filtering

pezholio:
For mysociety/fixmystreetforcouncils#146

This adds the JSON response to the controller action, and wraps the problem list in a `js-reports-list` container, so the list can be updated.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/My.pm
+++ b/perllib/FixMyStreet/App/Controller/My.pm
@@ -56,6 +56,10 @@ sub planned : Local : Args(0) {
     $c->forward('planned_reorder');
     $c->forward('/reports/stash_report_sort', [ 'shortlist' ]);
     $c->forward('get_problems');
+    if ($c->get_param('ajax')) {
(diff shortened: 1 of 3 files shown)


pezholio:
Ah, good catch. I'm sure there'll be a way round this. I'll have a peek 👍

pezholio:
This should be sorted now 👍

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -57,6 +57,7 @@ sub planned : Local : Args(0) {
     $c->forward('/reports/stash_report_sort', [ 'shortlist' ]);
     $c->forward('get_problems');
     if ($c->get_param('ajax')) {
+        $c->stash->{shortlist} = (!defined $c->get_param('sort') || $c->get_param('sort') eq 'shortlist');
```
dracos: You can just use the stash->{sort_key} here (which has already done param checking for you).

pezholio: Ah, nice 👍

PR #1622, opened 2017-01-31
Redirect to category-filtered reports view on login

davea:
As a little time-saver for users who have categories assigned, this PR will redirect them to their body's `/reports` page pre-filtered to their categories after login.

Fixes mysociety/fixmystreetforcouncils#69


davea:
In that case does it make sense for 'Sign in' in the nav bar to link to `/auth`? That way the correct redirect will happen.

dracos:
Yeah, I can't think of a problem with that :)

davea:
I got carried away and replaced the messy nav with something more reusable in b98d419a025177795ea850d3a1f8eac6dd2ac6a2. Before I carry on with that in the other cobrands, does it seem like a good approach?

davea:
I guess 6eee5b30233d9832d5a288c561a1684a8c736267 addresses the copying variables thing? I tried PROCESS but the `attr` and `liattr` variables were being carried through to the next `PROCESS navitem` call. Good point about the extra `| html`, don't think that's needed.

dracos:
Ah yes, looks good.

templates/web/base/main_nav_items.html
```diff
@@ -0,0 +1,38 @@
...
+    pc_uri = pc | uri;
+    suffix = "/list?pc=" _ pc_uri;
+  END;
+
+  PROCESS navitem uri='/alert' label=loc('Local alerts');
+~%]
+
+[%~ PROCESS navitem uri='/faq' label=loc('Help') ~%]
+
+[%~ PROCESS navitem uri='/privacy' label=loc('Privacy') liattrs='class="nav-menu__item--privacy"' ~%]
```
dracos: I think a variable should be used to show/not show this, rather than rely on CSS to hide it. I think the CSS is missing for Bromley and Stevenage anyway.

web/cobrands/angus/layout.scss
```diff
@@ -322,6 +322,10 @@ footer p.social a:active,footer p.social a:focus,footer p.social a:hover
   }
 }
 
+#main-nav {
```
dracos: Why not use omit-wrapper on the post_header_extra one instead?

templates/web/greenwich/header_logo.html
```diff
@@ -15,15 +15,4 @@
...
-        %]<li><[% IF c.req.uri.path == '/my' %]span[% ELSE %]a href="/my"[% END
-            %]>[% loc("Your reports") %]</[% c.req.uri.path == '/my' ? 'span' : 'a' %]></li>[%
-        %]<li><[% IF c.req.uri.path == '/reports' %]span[% ELSE %]a href="/reports"[% END
-            %]>[% loc("All reports") %]</[% c.req.uri.path == '/reports' ? 'span' : 'a' %]></li>[%
-        %]<li><[% IF c.req.uri.path == '/alert' %]span[% ELSE %]a href="/alert[% pc ? '/list?pc=' : '' %][% pc | uri %]"[% END
-            %]>[% loc("Local alerts") %]</[% c.req.uri.path == '/alert' ? 'span' : 'a' %]></li>[%
-        %]<li><[% IF c.req.uri.path == '/faq' %]span[% ELSE %]a href="/faq"[% END
-            %]>[% loc("Help") %]</[% c.req.uri.path == '/faq' ? 'span' : 'a' %]></li>
-    </ul>
+    [% INCLUDE "main_nav.html" body_name=c.cobrand.council_name hide_mysoc_link=1 ul_class="fms-menu-desktop" omit_wrapper=1 %]
```
dracos: Any reason to switch from ID to class here? You have `ul_id` to use.

templates/web/hart/footer.html
```diff
@@ -7,18 +7,7 @@
     <div class="nav-wrapper clearfix">
         <div class="main-menu-wrapper">
             <div class="main-menu">
-            <ul class="clearfix">
-                <li class="home"><[% IF c.req.uri.path == '/' %]span[% ELSE %]a href="/"[% END %] class="report-a-problem-btn"
-                    >[% "Report" %]</[% c.req.uri.path == '/' ? 'span' : 'a' %]></li>[%
-                %]<li><[% IF c.req.uri.path == '/my' OR ( c.req.uri.path == '/auth' AND c.req.params.r == 'my' ) %]span[% ELSE %]a href="/my"[% END
```
dracos: This tiny bit of functionality (also in Oxfordshire, others?) has been lost; I guess if it's /my or /auth (without any params) it could be a span rather than a link.

templates/web/bromley/footer.html
```diff
@@ -9,20 +9,7 @@
 
         <div class="nav-wrapper">
             <div class="nav-wrapper-2">
-                <div id="main-nav" role="navigation">
-                    <ul class="nav-menu nav-menu--main">
-                        <li><[% IF c.req.uri.path == '/' %]span[% ELSE %]a href="/"[% END
```
dracos: This one lacks the class all(?) the others have, but now it will have it - this makes a difference to display on e.g. /faq, so it either needs resetting in CSS or not including.

templates/web/base/main_nav_items.html
```diff
@@ -0,0 +1,38 @@
...
+    categories = c.user.categories.join(",") | uri;
+    suffix = "?filter_category=" _ categories;
+  END;
+
+  IF body_name;
+     body_name = body_name | uri;
+     suffix = "/${body_name}${suffix}";
+  END;
+
+  PROCESS navitem uri='/reports' label=loc('All reports');
```
dracos: For the span/a thing to work, uri needs to have the body name on it, not suffix.

templates/web/base/main_nav_items.html
dracos: Also if you use INCLUDE (sorry, I realise change here was probably my fault) then you shouldn't need to reset liattrs/attrs at the end of the navitem block.

templates/web/base/main_nav_items.html
```diff
@@ -0,0 +1,38 @@
...
+[%~ END ~%]
+
+[%~ IF c.user_exists AND c.user.has_body_permission_to('planned_reports') ~%]
+    [%~ PROCESS navitem uri='/my/planned' label=loc('Shortlist') ~%]
+[%~ END ~%]
+
+[%~
+  IF c.user_exists AND c.user.categories.size;
+    categories = c.user.categories.join(",") | uri;
+    suffix = "?filter_category=" _ categories;
```
dracos: I wonder if it would be clearer to pass suffix to all calls to navitem, so blank for most of them, rather than having a parent variable you have to clear.

davea: I was trying to keep the number of mandatory args to navitem down, hence clearing suffix at the end of the BLOCK. Unless there's a better way of having optional args when calling the BLOCK?

templates/web/base/main_nav_items.html
davea: The class only applies on desktop, the privacy link is visible on mobile. Does the reasoning for that still hold up, or should it just not be shown at all?

templates/web/greenwich/header_logo.html
davea: Good catch - this commit was from before the `ul_id` (mandated by Warwickshire's CSS) was added.

templates/web/base/main_nav_items.html
dracos: If it's currently not linked anywhere, I assume it's not a page that will have been adapted for that cobrand, and so that should presumably be maintained?

templates/web/base/main_nav_items.html
dracos: The nicest way I can think of would be to set a different var here, say tmp_suffix, then call INCLUDE navitem … suffix=tmp_suffix, then suffix wouldn't be set anywhere else, maybe?

davea: Yeah, I like that, will rejig it.

web/cobrands/angus/layout.scss
davea: Yep, forgot to go back to the Angus cobrand once I'd added the omit_wrapper option :)

Issue #1617, opened 2017-01-30
base_url_for_report should take account of language


dracos:
`base_url_for_report` is used in both front end and email, but does not take account of language, to perhaps stay in the right language in the front end, or take you to the right one in email. It should call `base_url_with_lang`.

dracos:
Fixed by #1627

PR #1627, opened 2017-02-02
Call base_url_with_lang in base_url_for_report

altinukshini:
Related issue #1617


perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -216,7 +216,7 @@ hashref.
 
 sub base_url_for_report {
     my ( $self, $report ) = @_;
-    return $self->base_url;
+    return base_url_with_lang($self);
```
dracos: Thanks for this – this should be `$self->base_url_with_lang` so that if the function is subclassed the correct one will be called.

dracos:
Merged, thank you :) I'll release 2.0.2 soon with this fix in.

PR #1613, opened 2017-01-24
Add My Areas page

pezholio:
This allows admins to apply multiple areas to a user, and allow a user with areas to see all reports in areas that they are assigned to.

fireshot capture 218 - your area __ fixmystreet_ - http___oxfordshire localhost_3000_my_areas

For mysociety/fixmystreetforcouncils#117

PR #1629, opened 2017-02-03
Filter by shortlisted problems

pezholio:
For mysociety/fixmystreetforcouncils#112

This allows inspectors to access any report listing page (such as /reports, or when #1613 is merged, /my/areas) and filter by if a problem is shortlisted or not.

The only issue I can see so far is that users will be able to click 'shortlisted' and 'unshortlisted', which may give an unexpected result (at the moment, it will default to shortlisted), but as people can click on 'unfixed' and 'fixed' at the moment too, that might be something to file for further improvements.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -378,6 +378,25 @@ sub load_and_group_problems : Private {
         non_public => 0,
         state      => [ keys %$states ]
     };
+    my $filter = {
+        order_by => $c->stash->{sort_order},
+        rows => $c->cobrand->reports_per_page,
+    };
+
+    if (defined $c->stash->{filter_status}{shortlisted}) {
+        $where->{'me.id'} = { '=', \"user_planned_reports.report_id"};
+        $where->{'user_planned_reports.removed'} = undef;
+        $filter->{join} = 'user_planned_reports';
+    } elsif (defined $c->stash->{filter_status}{unshortlisted}) {
+        my $shortlisted_ids = $c->cobrand->problems->search({
+            'me.id' => { '=', \"user_planned_reports.report_id"},
+            'user_planned_reports.removed' => undef,
+        }, {
+           join => 'user_planned_reports',
+           columns => ['me.id'],
+        })->as_query;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -378,6 +378,18 @@ sub load_and_group_problems : Private {
...
         state      => [ keys %$states ]
     };
+    my $filter = {
+        order_by => $c->stash->{sort_order},
+        rows => $c->cobrand->reports_per_page,
+    };
+
+    if (defined $c->stash->{filter_status}{shortlisted}) {
+        $where->{'me.id'} = { '=', \"user_planned_reports.report_id"};
+        $where->{'user_planned_reports.removed'} = { 'IS', \"NULL" };
```
davea: ` $where->{'user_planned_reports.removed'} = undef;` is a more idiomatic way of doing this.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -378,6 +378,18 @@ sub load_and_group_problems : Private {
...
+        $filter->{join} = 'user_planned_reports';
+    } elsif (defined $c->stash->{filter_status}{unshortlisted}) {
+        $where->{'me.id'} = { 'NOT IN', \"(SELECT problem.id FROM problem, user_planned_reports WHERE problem.id = user_planned_reports.report_id AND user_planned_reports.removed IS NULL)" };
```
davea: I like the approach of a subquery here. There's a `as_query` helper function you might want to use, as well as the `-not_in` operator, so you're leaning on the DB library instead of dropping to raw SQL. Something like
```
        my $shortlisted_ids = $c->cobrand->problems->search({
            'me.id' => { '=', \"user_planned_reports.report_id"},
            'user_planned_reports.removed' => undef,
        }, {
           join => 'user_planned_reports',
           columns => ['me.id'],
        })->as_query;
        $where->{'me.id'} = { -not_in => $shortlisted_ids };
```

should do the trick. There's some duplication of the query with the other clause in the parent `if`, so if you really wanted you could factor the common stuff out into a `shortlisted` sub in `ResultSet::Problem`, but 🤷‍♂️ .

PR #1631, opened 2017-02-06
Reorder list based on map view

pezholio:
Fixes mysociety/fixmystreetforcouncils#152

This loops through all the markers on a map, and, if they're visible, appends the associated list item to another `<ul>` element above the main `.item-list` element, effectively showing the visible elements first. I've added an extra `id` attribute to each item now to, as this is considerably faster than searching by `data` element.

### DIFF
--- a/templates/web/base/report/_item.html
+++ b/templates/web/base/report/_item.html
@@ -27,7 +27,7 @@
 [% END %]
 
 <li class="item-list__item item-list--reports__item [% item_extra_class %]"
-    data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]">
+    data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]" id="report-[% problem.id | html %]">
 <a href="[% c.cobrand.base_url_for_report( problem ) %][% problem.url %]">
     [% IF problem.photo %]
         <img class="img" height="60" width="90" src="[% problem.photos.first.url_fp %]" alt="">
--- a/templates/web/base/reports/_problem-list.html
+++ b/templates/web/base/reports/_problem-list.html
@@ -4,7 +4,7 @@
 
 [% BLOCK column %]
     [% IF reorder %]
-      <ul class="hidden item-list item-list--reports visible-reports"></ul>
+      <ul class="hidden item-list item-list--reports js-visible-reports"></ul>
     [% END %]
     <ul class="item-list item-list--reports">
         [% IF problems %]
--- a/templates/web/base/reports/_problem-list.html
+++ b/templates/web/base/reports/_problem-list.html
@@ -3,6 +3,7 @@
 %]
 
 [% BLOCK column %]
+    <ul class="hidden item-list item-list--reports visible-reports"></ul>
     <ul class="item-list item-list--reports">
         [% IF problems %]
             [% FOREACH problem IN problems %]
--- a/templates/web/base/reports/_problem-list.html
+++ b/templates/web/base/reports/_problem-list.html
@@ -3,7 +3,9 @@
 %]
 
 [% BLOCK column %]
-    <ul class="hidden item-list item-list--reports visible-reports"></ul>
+    [% IF reorder %]
+      <ul class="hidden item-list item-list--reports visible-reports"></ul>
+    [% END %]
     <ul class="item-list item-list--reports">
         [% IF problems %]
             [% FOREACH problem IN problems %]
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -119,6 +119,8 @@ var fixmystreet = fixmystreet || {};
           var visible = ''
           var $itemList = $('.item-list');
           var $visibleReports = $('.visible-reports');
+          if ($visibleReports.length == 0) return false;
+
           for (var i = 0; i < fixmystreet.markers.features.length; i++) {
               var feature = fixmystreet.markers.features[i];
               var $listItem = $itemList.find('#report-'+ feature.data.id);
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -125,10 +125,13 @@ var fixmystreet = fixmystreet || {};
               var feature = fixmystreet.markers.features[i];
               var $listItem = $itemList.find('#report-'+ feature.data.id);
               if (feature.onScreen()) {
-                  visible += $listItem[0].outerHTML; // This is faster than using .append()
+                  var clone = $($listItem[0].outerHTML);
+                  clone.attr('id', clone.attr('id') + '-visible');
+                  clone.removeClass('hidden');
+                  visible += clone[0].outerHTML; // This is faster than using .append()
                   $listItem.addClass('hidden');
               } else {
-                  $visibleReports.find('#report-'+ feature.data.id).remove();
+                  $visibleReports.find('#report-'+ feature.data.id + '-visible').remove();
                   $listItem.removeClass('hidden');
               }
           }
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -135,7 +135,7 @@ var fixmystreet = fixmystreet || {};
                   $listItem.removeClass('hidden');
               }
           }
-          if (visible == '') {
+          if (visible === '') {
             $visibleReports.addClass('hidden');
           } else {
             $visibleReports.removeClass('hidden');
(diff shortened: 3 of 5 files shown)


pezholio:
Quick example of how it should look:

avllisfo3o

pezholio:
Thoughts welcome on the UI here. I could append the visible items to the existing list container, but I think the border acts as a useful visual cue as to the state of the other reports.

pezholio:
I've added this functionality to the planned reports pages, but this messes with the reordering functionality. I'm also wondering if this is a useful thing to have at all on the shortlist page, given that there is likely to be fewer than c. 20 results on these pages

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,29 @@ var fixmystreet = fixmystreet || {};
...
+      reorder_visible: function() {
+          var visible = ''
+          var $itemList = $('.item-list');
+          var $visibleReports = $('.visible-reports');
+          for (var i = 0; i < fixmystreet.markers.features.length; i++) {
+              var feature = fixmystreet.markers.features[i];
+              var $listItem = $itemList.find('#report-'+ feature.data.id);
+              if (feature.onScreen()) {
+                  visible += $listItem[0].outerHTML; // This is faster than using .append()
+                  $listItem.addClass('hidden');
```
dracos: On an `/around` page, this hides all the reports, I'm afraid. (As there's no visible-reports div, so they all get hidden from the main div.)
I'm not sure this is appropriate behaviour for the around page, which is already doing "what's on the map" (plus a few outside to make sure something is shown), should it be restricted to the reports listing page only?

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,29 @@ var fixmystreet = fixmystreet || {};
...
 
+      reorder_visible: function() {
+          var visible = ''
+          var $itemList = $('.item-list');
+          var $visibleReports = $('.visible-reports');
+          for (var i = 0; i < fixmystreet.markers.features.length; i++) {
+              var feature = fixmystreet.markers.features[i];
+              var $listItem = $itemList.find('#report-'+ feature.data.id);
+              if (feature.onScreen()) {
+                  visible += $listItem[0].outerHTML; // This is faster than using .append()
```
dracos: This means each visible report is now in the DOM twice with the same ID. As long as nothing is relying on this (it's new here, so presumably not), I guess that's okay, but the worry is if something comes along and assumes it's unique in future.

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,29 @@ var fixmystreet = fixmystreet || {};
         fixmystreet.markers.redraw();
       },
 
+      reorder_visible: function() {
+          var visible = ''
+          var $itemList = $('.item-list');
+          var $visibleReports = $('.visible-reports');
+          for (var i = 0; i < fixmystreet.markers.features.length; i++) {
```
dracos: I assume marker order here is the same as sidebar list order, might be worth a comment confirming that.

templates/web/base/report/_item.html
```diff
@@ -27,7 +27,7 @@
 [% END %]
 
 <li class="item-list__item item-list--reports__item [% item_extra_class %]"
-    data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]">
+    data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]" id="report-[% problem.id | html %]">
```
dracos: Should whatever is using data-report-id be changed to use the ID instead?

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,29 @@ var fixmystreet = fixmystreet || {};
...
+              var $listItem = $itemList.find('#report-'+ feature.data.id);
+              if (feature.onScreen()) {
+                  visible += $listItem[0].outerHTML; // This is faster than using .append()
+                  $listItem.addClass('hidden');
+              } else {
+                  $visibleReports.find('#report-'+ feature.data.id).remove();
+                  $listItem.removeClass('hidden');
+              }
+          }
+          if (visible == '') {
```
dracos: Triple equals preferable here, if you're not going to have a boolean check.

web/js/map-OpenLayers.js
pezholio: Yeah, that was one of my concerns. I initially was selecting by `data-report-id`, but that was dog slow, which is why I added the ID. Wondering if I could append something to the ID when the element gets cloned and moved.

web/js/map-OpenLayers.js
pezholio: Ah, OK. I think you're right though. This behaviour is probably only appropriate on the `/reports` pages anyway

templates/web/base/report/_item.html
pezholio: As I've now changed the IDs to be unique in the list, I think it's safer to have the `data` element here too

templates/web/base/reports/_problem-list.html
```diff
@@ -3,7 +3,9 @@
 %]
 
 [% BLOCK column %]
-    <ul class="hidden item-list item-list--reports visible-reports"></ul>
+    [% IF reorder %]
+      <ul class="hidden item-list item-list--reports visible-reports"></ul>
```
dracos: More a musing than an issue, but is there a reason not to use an ID of, dunno, `js-visible-reports` here? To show it's for the JavaScript, and it's only being used once.

web/js/map-OpenLayers.js
```diff
@@ -125,10 +125,13 @@ var fixmystreet = fixmystreet || {};
               var feature = fixmystreet.markers.features[i];
               var $listItem = $itemList.find('#report-'+ feature.data.id);
               if (feature.onScreen()) {
-                  visible += $listItem[0].outerHTML; // This is faster than using .append()
+                  var clone = $($listItem[0].outerHTML);
```
dracos: Any reason not to use `$listItem.clone()`? :)

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,34 @@ var fixmystreet = fixmystreet || {};
         fixmystreet.markers.redraw();
       },
 
+      reorder_visible: function() {
+          var visible = ''
+          var $itemList = $('.item-list');
+          var $visibleReports = $('.visible-reports');
+          if ($visibleReports.length == 0) return false;
```
dracos: Please always have brackets with if statements (I don't think jslint will let this through anyway :) ).
At the start of this function, you could do `if fixmystreet.page != 'reports' { return; }` as that variable is always set, to save any DOM lookups at all.

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,34 @@ var fixmystreet = fixmystreet || {};
...
+
+          for (var i = 0; i < fixmystreet.markers.features.length; i++) {
+              var feature = fixmystreet.markers.features[i];
+              var $listItem = $itemList.find('#report-'+ feature.data.id);
+              if (feature.onScreen()) {
+                  var clone = $($listItem[0].outerHTML);
```
dracos: Any reason not to use `$listItem.clone()`? :)

templates/web/base/reports/_problem-list.html
```diff
@@ -3,6 +3,9 @@
 %]
 
 [% BLOCK column %]
+    [% IF reorder %]
+      <ul class="hidden item-list item-list--reports visible-reports"></ul>
```
dracos: More a musing than an issue, but is there a reason not to use an ID of, dunno, `js-visible-reports here`? To show it's for the JavaScript, and it's only being used once.

web/js/map-OpenLayers.js
pezholio: Oh, that's a better solution 👍

templates/web/base/reports/_problem-list.html
pezholio: Seems totally sensible and in line with the code style, so I'll do it 👍

templates/web/base/reports/_problem-list.html
```diff
@@ -4,7 +4,7 @@
 
 [% BLOCK column %]
     [% IF reorder %]
-      <ul class="hidden item-list item-list--reports visible-reports"></ul>
+      <ul class="hidden item-list item-list--reports js-visible-reports"></ul>
```
dracos: This could be an ID, rather than a class.

dracos:
This is no longer necessary, I'm afraid, due to #1689 `/reports` will now only show reports in view.

PR #1635, opened 2017-02-07
Fix crash on reports with empty bodies_str

davea:


davea:
I was, then my train arrived :) Will finish that up before merging.

PR #1636, opened 2017-02-07
Simplify open and inspect procedure for inspectors

pezholio:
For mysociety/fixmystreetforcouncils#150

This makes it somewhat easier for inspectors who might be reporting a new problem to open a new report and add the inspection details by autocompleting a bunch of stuff for them, and redirecting to the inspection page once the report has been made. The flow now looks like this:

jshz7qhxkz

And like this on mobile:

kir8cximhu

It doesn't quite address everything in mysociety/fixmystreetforcouncils#150, but goes a long way to reducing the friction.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -105,6 +105,7 @@ sub report_new : Path : Args(0) {
     # deal with the user and report and check both are happy
 
     return unless $c->forward('check_form_submitted');
+
     $c->forward('/auth/check_csrf_token');
     $c->forward('process_user');
     $c->forward('process_report');
@@ -493,7 +494,7 @@ Work out what the location of the report should be - either by using lat,lng or
 a tile click or what's come in from a partial. Returns false if no location
 could be found.
 
-=cut 
+=cut
 
 sub determine_location : Private {
     my ( $self, $c ) = @_;
@@ -515,7 +516,7 @@ sub determine_location : Private {
 Detect that the map tiles have been clicked on by looking for the tile
 parameters.
 
-=cut 
+=cut
 
 sub determine_location_from_tile_click : Private {
     my ( $self, $c ) = @_;
@@ -566,7 +567,7 @@ sub determine_location_from_tile_click : Private {
 Use latitude and longitude stored in the report - this is probably result of a
 partial report being loaded.
 
-=cut 
+=cut
 
 sub determine_location_from_report : Private {
     my ( $self, $c ) = @_;
@@ -1203,9 +1204,14 @@ sub redirect_or_confirm_creation : Private {
                 to => [ [ $report->user->email, $report->name ] ],
             } );
         }
-        $c->log->info($report->user->id . ' was logged in, showing confirmation page for ' . $report->id);
-        $c->stash->{created_report} = 'loggedin';
-        $c->stash->{template} = 'tokens/confirm_problem.html';
+        if ($c->user_exists && $c->user->has_body_permission_to('planned_reports')) {
+            $c->log->info($report->user->id . ' is an inspector - redirecting straight to report page for ' . $report->id);
+            $c->res->redirect( '/report/'. $report->id );
+        } else {
+            $c->log->info($report->user->id . ' was logged in, showing confirmation page for ' . $report->id);
+            $c->stash->{created_report} = 'loggedin';
+            $c->stash->{template} = 'tokens/confirm_problem.html';
+        }
         return 1;
     }
 
--- a/templates/web/base/report/new/category.html
+++ b/templates/web/base/report/new/category.html
@@ -5,7 +5,7 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category'>
+    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
         [%~ FOREACH cat_op IN category_options ~%]
         [% cat_op_lc = cat_op | lower =%]
         <option value='[% cat_op | html %]'[% ' selected' IF report.category == cat_op || category_lc == cat_op_lc || (category_options.size == 2 AND loop.last) ~%]
--- a/templates/web/base/report/new/category_wrapper.html
+++ b/templates/web/base/report/new/category_wrapper.html
@@ -1,7 +1,7 @@
 <div id="form_category_row">
 [% IF js %]
     <label for="form_category">[% loc('Category') %]</label>
-    <select class="form-control" name="category" id="form_category" required><option>[% loc('Loading...') %]</option></select>
+    <select class="form-control" name="category" id="form_category" data-role="[% c.user_exists AND c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" required><option>[% loc('Loading...') %]</option></select>
 [% ELSE %]
     [% IF category_options.size %]
         [% IF field_errors.category %]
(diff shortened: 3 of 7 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1206,10 +1207,16 @@ sub redirect_or_confirm_creation : Private {
                 to => [ [ $report->user->email, $report->name ] ],
             } );
         }
-        $c->log->info($report->user->id . ' was logged in, showing confirmation page for ' . $report->id);
-        $c->stash->{created_report} = 'loggedin';
-        $c->stash->{template} = 'tokens/confirm_problem.html';
-        return 1;
+        if ($c->user_exists && $c->user->has_body_permission_to('planned_reports')) {
```
dracos: Not sure $c->user_exists is needed here, as we are logged in if we get here, but no harm in leaving it :)

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1206,10 +1207,16 @@ sub redirect_or_confirm_creation : Private {
...
+            $c->log->info($report->user->id . ' is an inspector - redirecting straight to report page for ' . $report->id);
+            $c->res->redirect( '/report/'. $report->id );
+            return 1;
+        } else {
+            $c->log->info($report->user->id . ' was logged in, showing confirmation page for ' . $report->id);
+            $c->stash->{created_report} = 'loggedin';
+            $c->stash->{template} = 'tokens/confirm_problem.html';
+            return 1;
```
dracos: Better to have the one `return 1` underneath this if block?

templates/web/base/report/new/category.html
```diff
@@ -5,7 +5,7 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
+    <select class="form-control" name='category' id='form_category' data-role='[% c.user_exists AND c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
```
dracos: Out of interest, how come the second c.user here doesn't need this check? I'm guessing as there's no function call, it silently outputs instead?

templates/web/base/report/new/category_wrapper.html
```diff
@@ -1,7 +1,7 @@
 <div id="form_category_row">
 [% IF js %]
     <label for="form_category">[% loc('Category') %]</label>
-    <select class="form-control" name="category" id="form_category" required><option>[% loc('Loading...') %]</option></select>
+    <select class="form-control" name="category" id="form_category" data-role="[% "inspector" IF c.user_exists AND c.user.has_body_permission_to('planned_reports')  %]" required><option>[% loc('Loading...') %]</option></select>
```
dracos: I think both the data-roles could be the same? (Don't mind which :) )

dracos: And does it not need the corresponding `data-body`?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -403,6 +403,15 @@ $.extend(fixmystreet.set_up, {
                 $category_meta.empty();
             }
         });
+
+        if ($(this).data('role') == 'inspector') {
```
dracos: I guess this should preferably be in `staff.js` with a hook or something.

pezholio: Have you got an example of where you use this kinda pattern already?

templates/web/base/report/new/category.html
dracos: Not outdated, github!

web/cobrands/fixmystreet/fixmystreet.js
dracos: `fixmystreet.run` is used in set up for moderation etc., that could work here, though it is only one level of checking, but something like that. You need something like `if (fixmystreet.hooks && fixmystreet.hooks.category_change) { fixmystreet.hooks.category_change(args) }`.

Also, just thought - it probably only wants to do the setting if they're currently blank, to prevent removing anything already entered.

templates/web/base/report/new/category.html
```diff
@@ -5,7 +5,7 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category'>
+    <select class="form-control" name='category' id='form_category' data-role='[% c.user_exists AND c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
```
dracos: Out of interest, how come the second c.user here doesn't need this check? I'm guessing it's because there's no explicit function call (with brackets)?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -425,6 +425,8 @@ $.extend(fixmystreet.set_up, {
                 $category_meta.empty();
             }
         });
+
+        fixmystreet.run(fixmystreet.set_up.update_problem_fields($(this).data('role'), $(this).data('body'), args));
```
dracos: This will not work, because you are passing the results of a function to `fixmystreet.run` – the function by this point has already been called. You need to pass a function reference, be it just a name, or an anonymous function. I'd suggest just checking for `fixmystreet.hooks` as in my last review.

web/cobrands/fixmystreet/staff.js
```diff
@@ -340,5 +340,29 @@ $.extend(fixmystreet.set_up, {
             $input.val($this.val());
         }
     });
+  },
+
+  update_problem_fields: function(role, body, args) {
```
dracos: This should not be in `fixmystreet.set_up` as you do not want it to run at startup, which all functions in `set_up` are. That's why I suggested its own hooks object.

templates/web/base/report/new/category.html
pezholio: I don't think the first `c.user_exists` call is necessary either looking at it, as even when logged out, it still returns `true`. Not sure if that's the case with all the cobrands though

dracos: Are you sure? That seems very odd if so.
Having it there would (if it was the only use) prevent a needless lookup if not logged in, and can prevent an error in actual code, but yes, it's probably not needed in the template here.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -425,6 +425,10 @@ $.extend(fixmystreet.set_up, {
                 $category_meta.empty();
             }
         });
+
+        if (fixmystreet.hooks) {
```
dracos: This should be `if (fixmystreet.hooks && fixmystreet.hooks.update_problem_fields)` in case `hooks` later on gets defined elsewhere.

web/cobrands/fixmystreet/staff.js
```diff
@@ -342,3 +342,29 @@ $.extend(fixmystreet.set_up, {
     });
   }
 });
+
+fixmystreet.hooks = {
```
dracos: Could this set hooks if it doesn't exist, and then extend it? That way if more hooks are added in another file, nothing needs to be changed here. See how the existing `fixmystreet.set_up` does it in fixmystreet.js, might as well be consistent.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -128,6 +128,8 @@ function isR2L() {
   });
 })(jQuery);
 
+fixmystreet.hooks = fixmystreet.hooks || {};
```
dracos: What I meant was that this line would be in `staff.js` and then it would be self-contained and not overwrite any other hooks created in separate files. As we know this file comes before `staff.js` this doesn't matter, but sorry if I was unclear in what I was trying to explain.

pezholio: No worries :+1:

PR #1642, opened 2017-02-16
Clarify purpose of FMS4C contact form

### LINKED ISSUE #1553: Add additional notice to /councils contact page
We’ve had a few of these requests in the past few days. I think we need to add something to say ‘if you wish to report a problem please click here’ to take them to the main screen or possibly to ask for their location and jump into the process from there.

zarino:
* Tweak wording to make it clearer that the form is unsuitable for residents wanting to report street problems.
* Record HTTP Referer when form is submitted, so we can see which pages are leading people to the FMS4C page.
* Tidy up capitalisation elsewhere in headings on the page.
* Replace links to `/council` with `/about/council` to avoid an unnecessary redirect (thanks @davea!) and remove the `www.fixmystreet.com` prefix inside the fixmystreet.com cobrand templates.

Fixes #1553.

screen shot 2017-02-16 at 15 59 04

### DIFF
--- a/templates/web/fixmystreet.com/about/council.html
+++ b/templates/web/fixmystreet.com/about/council.html
@@ -27,7 +27,7 @@
       <div class="councils-features__grid">
         <div class="councils-features__feature">
           <div class="councils-features__thumb councils-features__thumb--excellence"></div>
-          <h3>Digital Excellence</h3>
+          <h3>Digital excellence</h3>
           <p>FixMyStreet was co-designed with council insiders. We made it into the most user-centred, accessible approach to non-emergency issue reporting available today. And now it’s the leader in its class.</p>
         </div>
         <div class="councils-features__feature">
@@ -114,7 +114,7 @@
       <h2>Works with</h2>
       <div class="councils-compatibility__grid">
         <div class="councils-compatibility__item">
-          <h3>Customer Management</h3>
+          <h3>Customer management</h3>
           <ul>
             <li><a href="http://www.kana.com/lagan-crm/customer-experience-management">Lagan</a></li>
             <li><a href="https://www.microsoft.com/en-gb/dynamics/default.aspx">Microsoft Dynamics</a></li>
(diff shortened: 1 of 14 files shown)


templates/web/fixmystreet.com/footer_extra.html
```diff
@@ -31,7 +31,7 @@ <h2 class="mysoc-footer__site-name">FixMyStreet</h2>
                             %]>[% loc("Local alerts") %]</[% c.req.uri.path == '/alert' ? 'span' : 'a' %]></li>
                     </ul>
                     <ul>
-                        <li><a href="https://www.fixmystreet.com/council">FixMyStreet for Councils</a></li>
+                        <li><a href="/council">FixMyStreet for Councils</a></li>
```
davea: Could this (and the others) be to `/about/council`, as that's where the `/council` redirect takes you?

templates/web/fixmystreet.com/about/council.html
```diff
@@ -245,20 +258,8 @@ <h2 class="councils-hero__title">Reliable fault-reporting and case management fo
...
-          <label><input name="extra.product" id="product" type="radio" value="case-management-integration">Case-management integration</label>
-          <label><input name="extra.product" id="product" type="radio" value="top-to-bottom-case-management">Top-to-bottom case-management</label>
-        </div>
-      </div>
-      <div class="form-group">
-        <label for="message">Message</label>
-        <textarea name="message" id="message" required> </textarea>
-      </div>
       <div class="form-group submit-group">
+        <input type="hidden" name="extra.referer" value="[% c.req.headers.referer %]">
```
davea: This needs a `| html` so the value is escaped.

templates/web/fixmystreet.com/footer_extra.html
zarino: I wondered about that! If we change these to `/about/council`, should we also change all the other instances of `www.fixmystreet.org/council` in all the other cobrands?

davea: Saving the web, one redirect at a time :) Yep, it wouldn't hurt 👍

zarino: Ok! Will pull out the two fixed URLs from the existing commit into a new separate commit, and then add in fixes for all the other URLs.

PR #1645, opened 2017-02-17
Send open reports regardless of current state

davea:
It's a relatively minor change that I don't **think** has any knock-on effects, but figured I'd PR it for a second opinion just to be safe 😃 

For mysociety/fixmystreetforcouncils#168

### DIFF
--- a/perllib/FixMyStreet/Script/Reports.pm
+++ b/perllib/FixMyStreet/Script/Reports.pm
@@ -28,7 +28,7 @@ sub send(;$) {
     my $base_url = FixMyStreet->config('BASE_URL');
     my $site = $site_override || CronFns::site($base_url);
 
-    my $states = [ 'confirmed', 'fixed' ];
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Script/Reports.pm
```diff
@@ -28,7 +28,10 @@ sub send(;$) {
     my $base_url = FixMyStreet->config('BASE_URL');
     my $site = $site_override || CronFns::site($base_url);
 
-    my $states = [ 'confirmed', 'fixed' ];
+    my $states = [
+        FixMyStreet::DB::Result::Problem::open_states(),
+        'fixed'
```
dracos: Can probably drop fixed in fact, given this matches nothing, and all fixeds are meant to be user/council.

dracos: There's another states list like this further down in the unsent section.

PR #1652, opened 2017-02-27
Shortlist multiple reports

pezholio:
For mysociety/fixmystreetforcouncils#59

This allows inspectors to shortlist multiple reports by zooming into a section of the map on the 'All Reports' page and clicking the `Shortlist all reports` button, which only appears on a zoom level of 14 or higher. This is a similar approach to being able to draw a rectangle on a map, but more mobile friendly and more in tune with the 'slippy map' way of doing things vs. old skool GIS system mapping.

Par example:

agva3yb4nm


web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,32 @@ var fixmystreet = fixmystreet || {};
...
+              for (var i = 0; i < fixmystreet.markers.features.length; i++) {
+                var feature = fixmystreet.markers.features[i];
+                if (feature.onScreen()) {
+                  features.push(feature.data.id);
+                }
+              }
+
+              $.post( "/my/planned/change_multiple", { ids: features, token: csrf }, function() {
+                $shortlistButton.addClass('hidden');
+              });
```
davea: What happens if this POST fails? Some kind of user feedback should be shown, whatever the outcome.

It'd be nice if the sidebar could be updated to show the now-shortlisted reports with the filled-in green star icon.

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -327,6 +327,11 @@ sub admin_user_body_permissions {
     });
 }
 
+sub is_inspector {
```
pezholio: Ah, OK. Happy to revert in that case 😄

templates/web/base/maps/openlayers.html
```diff
@@ -33,4 +33,6 @@
     [% IF map.copyright %]
     <div class="olControlAttribution" style="position: absolute;">[% map.copyright %]</div>
     [% END %]
-
+    [% IF c.user && c.user.is_inspector %]
+    <div id="fms_shortlist_all" class="hidden">Shortlist all visible</div>
```
pezholio: I guess it's to make it a bit more visible. The `sub_maps_links` part looks more related to map controls, whereas this is a specific action on reports.

web/js/map-OpenLayers.js
pezholio: Is an alert enough, given that it's likely to be an edge case?

pezholio:
Hmmmm... On reflection then, maybe the toolbar would be a better place. It would certainly look better on mobile too (and also, we could also add the fullscreen functionality in #1655 to make shortlisting on the go easier)

pezholio:
@davea I've added the button to the toolbar now. Just trying to dig in and work out why it's only showing 1 report as updating. Suspect it might be the fact that there's only 1 HTTP request being made?

pezholio:
@davea - Sorted than now too 👍

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,67 @@ var fixmystreet = fixmystreet || {};
         fixmystreet.markers.redraw();
       },
 
+      shortlist_multiple: function(ids, token) {
```
dracos: Now we've separated out the staff JS, I'd like this to be only in that staff JS please somehow :)

web/cobrands/sass/_base.scss
```diff
@@ -1493,6 +1493,27 @@ html.js #map .noscript {
       background-image:url($image-sprite);
       background-position: flip(right, -341px) -4070px;
     }
+    &#fms_shortlist_all {
```
dracos: Not necessarily now (though could be), but as we have factored out staff JS, should we also do the same with staff CSS?

templates/web/base/reports/body.html
```diff
@@ -27,6 +27,12 @@
 
 [% map_html %]
 
+[% IF c.user && c.user.has_body_permission_to('planned_reports') %]
+  <p id="sub_map_links">
+    <a id="fms_shortlist_all" class="hidden">Shortlist all visible</a>
```
dracos: Needs marking for translation

t/app/controller/my_planned.t
```diff
@@ -57,6 +57,31 @@ $mech->get_ok('/my/planned?ajax=1');
...
 
+subtest "POSTing multiple problems to my/planned/change adds all to shortlist" => sub {
+    my ($problem1, $problem2, $problem3) = $mech->create_problems_for_body(3, $body->id, 'New Problem');
+
+    # Grab CSRF token
+    $mech->get_ok($problem1->url);
+    my ($csrf) = $mech->content =~ /meta content="([^"]*)" name="csrf-token"/;
+
+    $mech->post_ok( '/my/planned/change_multiple', {
+            'ids[]' => [
```
dracos: The actual JS, as opposed to the test and server-side, is using `ids`, not `ids[]`? There's no need for `[]` as far as I can see.

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,67 @@ var fixmystreet = fixmystreet || {};
...
+              'lastupdate': $item.data('lastupdate')
+            });
+          }
+          $(document).trigger('shortlist-all', { items: items});
+        })
+        .fail(function(response) {
+          if (response.status == 400) {
+            // If the response is 400, then get a new CSRF token and retry
+            var csrf = response.responseText.match(/name="token" value="([^"]*)"/)[1];
+            fixmystreet.maps.shortlist_multiple(ids, csrf);
```
dracos: There is the possibility (if slight) of a loop here if the server were to keep returning 400 for some reason. Perhaps add a countdown flag so it won't retry for ever?

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,67 @@ var fixmystreet = fixmystreet || {};
         fixmystreet.markers.redraw();
       },
 
+      shortlist_multiple: function(ids, token) {
+        $.post("/my/planned/change_multiple", { ids: ids, token: token })
+        .done(function() {
+          var $itemList = $('.item-list'),
+              items = [];
+          $('.shortlisted-status').remove();
```
dracos: This class is only on an individual report page?

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,67 @@ var fixmystreet = fixmystreet || {};
...
+
+          for (var i = 0; i < ids.length; i++) {
+            var problemId = ids[i],
+                $item = $itemList.find('#report-'+ problemId)
+                $form = $item.find('form'),
+                $submit = $form.find("input[type='submit']" );
+
+            $item.attr('data-lastupdate', new Date().toISOString());
```
dracos: I'm not clear on what this is doing. `lastupdate` isn't usually updated on an add/remove shortlist, because the report hasn't changed. The lastupdate functionality should still operate okay without this, if the report is newly added, I think?

web/js/map-OpenLayers.js
```diff
@@ -115,6 +115,67 @@ var fixmystreet = fixmystreet || {};
...
+
+            $submit.attr('name', 'shortlist-remove')
```
dracos: Can this use the toggle_shortlist function?

templates/web/base/common_header_tags.html
```diff
@@ -6,6 +6,10 @@
     id="js-meta-data" data-page="[% page %]"
     data-cobrand="[% c.cobrand.moniker %]">
 
+[% IF csrf_token %]
+  <meta content="[% csrf_token %]" name="csrf-token" />
```
dracos: Switch the other things that could use this to it? e.g. offline.js has it once and staff.js line 109ish. And indeed the 400 handler in this PR :)

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -205,6 +205,20 @@ sub planned_change : Path('planned/change') {
     }
 }
 
+sub shortlist_multiple : Path('planned/change_multiple') {
+    my ($self, $c) = @_;
+    $c->forward('/auth/check_csrf_token');
+
+    my @ids = $c->get_param_list('ids[]');
```
dracos: Should be 'ids' from other comment.

web/js/map-OpenLayers.js
pezholio: I'm just following the conventions that were applied on a single update a la: https://github.com/mysociety/fixmystreet/blob/master/web/cobrands/fixmystreet/offline.js#L413-L416

dracos: That caches the lastupdate, yes, but it doesn't change it, as this code does?

t/app/controller/my_planned.t
pezholio: The IDs will be an array in the JS though, so server side, it's received as an array.

dracos: Sorry, could you explain this more? Perl doesn't do anything special with `[]` variables as far as I know, that's e.g. PHP?

dracos: You are passing `ids[]=1&ids[]=2` to the server, but the server is reading the `ids` variable, not the `ids[]` variable...

pezholio: So `ids` is an array on the client side, and running `$.post("/my/planned/change_multiple", { ids: ids, token: token })` will translate to POSTing `ids[]=1213,ids[]=3456` etc to `/my/planned/change_multiple`. Changing this to `ids` server side means `ids[]` doesn't get picked up.

dracos: Aha, I did not realise jQuery added `[]` (sigh!), thanks for explaining.

pezholio: Too much magic sometimes.

templates/web/base/common_header_tags.html
pezholio: Sorry, just picking this up. The lines that I think you're talking about (https://github.com/mysociety/fixmystreet/pull/1652/files#diff-edf8ca8647028bd9be9b2f3f92e67ab6L73 and https://github.com/mysociety/fixmystreet/pull/1652/files#diff-0c932d62c9dad1e14d7c9e8bde78d4f5R376) are dealing with `jqXHR` objects, so to parse these, I'd have to load the `responseText` into a jQuery object, which seems like overkill. I think a regex is probably fine here (unless I've got the wrong end of the stick!)

dracos: Yes, those changes (from a commit 22nd March) postdate my comment of 17th March :)

pezholio: Ah ha!

pezholio:
@dracos Reckon this is good to go now then?

web/cobrands/fixmystreet/staff.js
```diff
@@ -1,3 +1,12 @@
+function toggle_shortlist (btn, sw, id) {
```
dracos: Don't really want a global function, but not fussed for the moment.

web/cobrands/fixmystreet/staff.js
```diff
@@ -342,3 +342,68 @@ $.extend(fixmystreet.set_up, {
...
+      var $itemList = $('.item-list'),
+          items = [];
+
+      for (var i = 0; i < ids.length; i++) {
+        var problemId = ids[i],
+            $item = $itemList.find('#report-'+ problemId),
+            $form = $item.find('form'),
+            $submit = $form.find("input[type='submit']" );
+
+        $item.attr('data-lastupdate', new Date().toISOString());
```
dracos: Repeating from previous review:
dracos: I'm not clear on what this is doing. lastupdate isn't usually updated on an add/remove shortlist, because the report hasn't changed. The lastupdate functionality should still operate okay without this, if the report is newly added, I think?
pezholio: I'm just following the conventions that were applied on a single update a la: https://github.com/mysociety/fixmystreet/blob/master/web/cobrands/fixmystreet/offline.js#L413-L416
dracos: That caches the lastupdate, yes, but it doesn't change it, as this code does?

web/cobrands/fixmystreet/staff.js
```diff
@@ -342,3 +342,68 @@ $.extend(fixmystreet.set_up, {
...
+
+  show_shortlist_control: function() {
+    var $shortlistButton = $('#fms_shortlist_all');
+    if ($shortlistButton === undefined || fixmystreet.page != "reports" ) {
+      return;
+    }
+
+    if (fixmystreet.map.getZoom() >= 14) {
+      $shortlistButton.removeClass('hidden');
+      $shortlistButton.on('click', function() {
```
dracos: This will readd this listener every time the zoom goes above 14. It only needs to be added once on setup (or lazily on first time zoom hits 14), I think?

pezholio: Good point 👍

web/cobrands/fixmystreet/staff.js
pezholio: OK, I'll get rid of this line 😄

web/cobrands/fixmystreet/staff.js
```diff
@@ -366,5 +357,79 @@ $.extend(fixmystreet.hooks, {
...
+    });
+  },
+
+  show_shortlist_control: function() {
+    var $shortlistButton = $('#fms_shortlist_all');
+    if ($shortlistButton === undefined || fixmystreet.page != "reports" ) {
+      return;
+    }
+
+    $shortlistButton.on('click', function() {
```
dracos: This still adds this listener on every zoom change, no?

pezholio: Oh, of course it does!

web/cobrands/fixmystreet/staff.js
```diff
@@ -366,5 +357,79 @@ $.extend(fixmystreet.hooks, {
               $(this).data('autopopulated', false);
             });
         }
+    },
+
+    toggle_shortlist: function(btn, sw, id) {
```
dracos: Not sure it's strictly a hook, is it, but meh, it's better than global :)

web/cobrands/fixmystreet/staff.js
pezholio: I think adding the listener lazily is the best approach here

PR #1653, opened 2017-02-28
Only output appcache iframe for shortlist users

dracos:
First way of doing it left as first commit for interest.


dracos:
Merged as 2f4b341

PR #1655, opened 2017-03-01
Add fullscreen option to map

pezholio:
For mysociety/fixmystreetforcouncils#156

This adds a 'Show Fullscreen' option to the map on a `/report` page to make the map easier to use when on mobile:


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -869,6 +869,10 @@ $.extend(fixmystreet.set_up, {
         }
         $('#key-tools li:empty').remove();
         $('#report-updates-data').insertAfter($('#map_box'));
+        if (fixmystreet.page === 'report') {
+            $('#sub_map_links').append('<a href="#" id="show-fullscreen" class="expand">Expand map</span>');
```
dracos: These need internationalizing, see how other strings do it elsewhere :)

web/cobrands/sass/_base.scss
```diff
@@ -1496,6 +1497,48 @@ html.js #map .noscript {
       background-image:url($image-sprite);
       background-position: flip(right, -341px) -3936px;
     }
+    &.expand {
+      padding: 0.6em 1em 0.5em 1em;
+      &:after {
```
dracos: Not saying this is wrong, but why is this done differently to the other sub map link icons, with an `:after`, and not just an immediate background-image?
Also, note how the other icons do their positioning - you need to account for the LTR/RTL possibility (where you'd want the icon on the right in RTL).

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -869,6 +869,10 @@ $.extend(fixmystreet.set_up, {
...
+            $('#sub_map_links').append('<a href="#" id="hide-fullscreen" class="compress hidden">Collapse map</span>');
```
dracos: Would this be nicer as one link that changes class/text when clicked, rather than two? Guess it doesn't make much difference, but I think other similar uses in the code currently toggle the one link.

web/cobrands/sass/_base.scss
```diff
@@ -1559,7 +1602,7 @@ html.js #map .noscript {
   text-transform: uppercase;
   text-align: center;
   font-size: 0.875em;
-  .mobile-reporting-map & {
+  .mobile-reporting-map &, .report-fullscreen & {
```
dracos: This class never appears on a report page.

web/cobrands/sass/_base.scss
```diff
@@ -1656,7 +1699,7 @@ a:hover.rap-notes-trigger {
 // and make the map full screen to reduce distractions. JavaScript also
 // tweaks the text content of some of the map-related elements, to make
 // it more "app-like".
-.mobile-reporting-map {
+.mobile-reporting-map, .report-fullscreen {
```
dracos: I like how these share a lot, I wonder if we can clear it up a bit to split it out a bit more. So we'd have:
* one class that shows the map full screen, `map-fullscreen` perhaps;
* one class that 'locks' it, that's `only-map`;
* one class that does page-specific thing, so `mobile-reporting-map` (perhaps renamed something like map-reporting?) and its pan zoom top.

Then the report page would have `map-fullscreen only-map`, and the around page would have `map-fullscreen only-map map-reporting` (where it currently has mobile-reporting-map only-map). Does that make sense, or not worth it?

web/cobrands/sass/_base.scss
pezholio: Was wrestling with this for a while this morning (my CSS-fu is weak), and @zarino suggested doing it this way, as positioning relative to the right hand side is tricky to do with `background-position`. I think it's easier with the others as they use a sprite, but I decided to use an SVG as we seem to be moving in that direction for other stuff.

dracos: Fair enough! Hmm, not sure the LTR/RTL code has had to deal with an after/before swap before, I'll leave that with you to try and sort ;) There are many uses of the left/right and flip() mixins that hopefully make sense. `.expand` and `.close` can share most of their CSS too.

web/cobrands/sass/_base.scss
pezholio: Yeah, I like this, then if we want to use `map-fullscreen` anywhere else, it makes sense, without having to add another class

web/cobrands/sass/_base.scss
```diff
@@ -1496,6 +1497,48 @@ html.js #map .noscript {
...
+        @media (-webkit-min-device-pixel-ratio: 2), (min-resolution: 192dpi) {
+            background-image: url(/cobrands/fixmystreet/images/expand@2x.png);
+        }
+
+        @media all {
+            background-image: url(/cobrands/fixmystreet/images/expand.svg), none;
+        }
+        background-size: 16px 16px;
+        vertical-align: middle;
+        margin: flip(0 0 0 8px, 0 8px 0 0);
```
dracos: This is fine, just to note that if you did only want to do the one margin as originally, you could write it as:
`margin-#{$left}: 8px;`.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -869,7 +869,7 @@ $.extend(fixmystreet.set_up, {
         }
         $('#key-tools li:empty').remove();
         $('#report-updates-data').insertAfter($('#map_box'));
-        if (fixmystreet.page === 'report') {
+        if (['reports','report'].includes(fixmystreet.page)) {
```
dracos: `includes` is much too modern, I'm afraid.

web/cobrands/sass/_base.scss
```diff
@@ -1496,6 +1497,48 @@ html.js #map .noscript {
...
+      }
+    }
+    &.compress {
```
dracos: .expand and .compress still share most of their style, all of it apart from the image URLs could be on #toggle-fullscreen for example.

web/cobrands/sass/_base.scss
pezholio: Aaaah, shiny!

PR #1656, opened 2017-03-01
Make sure report Ajax call is not cached by IE11.

### LINKED ISSUE #1638: Update not always visible due to caching in IE11
### Description of problem

Updates might not be visible when visiting a cached report on IE11

### Version

2.0.1

### Steps to reproduce

1. Log in on FixMyStreet (I used https://beta.fixamingata.se/)
2. Visit an existing report through All reports
3. Make an update
4. Click on All reports and the body in question
5. Click on the report

### Actual results

After having performed Step 5, the update is *not* visible (Refreshing the page *will* display the update)

### Expected results

The update is visible after having performed Step 5, without having to refresh the page

### How reproducible

Seemingly always using IE11 (I could *not* reproduce the issue using Mozilla Firefox)

### Additional information

The time specified in the report list (after having performed Step 4) *does* reflect the update just made

dracos:
IE11 caches this Ajax call, not calling out to the server at all if it
is made again (e.g. after an update has been made). Fixes #1638.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1301,7 +1301,7 @@ fixmystreet.display = {
   },
 
   report: function(reportPageUrl, reportId, callback) {
-    $.ajax(reportPageUrl).done(function(html, textStatus, jqXHR) {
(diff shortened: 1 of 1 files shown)


pezholio:
This seems fine to me! 👍

PR #1658, opened 2017-03-03
Fix a script, and inspecting superuser, add a helper

### LINKED ISSUE #1616: generate_cobrand_po doesn't have correct setup
It needs to generate a virtualenv like gettext-extract does.
Current workaround:
1. `bin/gettext-extract`
2. `git checkout locale/FixMyStreet.po`
3. `PATH="local-gettext/bin:$PATH" PERL5LIB="local-gettext/lib/perl5:local/lib/perl5" bin/generate_cobrand_po`

### LINKED ISSUE #1640: Save with public update inspection fails
I've replicated this problem many times. I was logged in as superuser (not council user), and tried to inspect a report, I checked the "Save with public update" and received this error below:

If I uncheck the "Save with public update", everything works fine!

[Fri Feb 10 16:09:09.849212 2017] [fcgid:warn] [pid 9026:tid 139948707485440] [client 127.0.0.1:49322] mod_fcgid: stderr: [error] Caught exception in FixMyStreet::App::Controller::Report->inspect "Can't call method "name" on an undefined value at /var/www/mywebsite.org/fixmystreet/web/../perllib/FixMyStreet/App/Controller/Report.pm line 416.", referer: http://mywebsite.org/report/13

dracos:
There is no link between these three commits.
Fixes #1616. Fixes #1640.

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -549,6 +549,16 @@ sub is_open {
     return exists $self->open_states->{ $self->state } ? 1 : 0;
 }
 
+=head2 is_in_progress
+
+Sees if the problem is in an open, not 'confirmed' state.
+
+=cut
+
+sub is_in_progress {
+    my $self = shift;
+    return $self->is_open && $self->state ne 'confirmed' ? 1 : 0;
+}
 
 =head2 is_fixed
 
--- a/bin/generate_cobrand_po
+++ b/bin/generate_cobrand_po
@@ -13,14 +13,22 @@ the main FixMyStreet po files
 
 =cut
 
-my $cobrand = shift;
+use File::Basename qw(dirname);
+use File::Spec;
 
+my $cobrand = shift;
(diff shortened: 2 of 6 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -549,6 +549,16 @@ sub is_open {
     return exists $self->open_states->{ $self->state } ? 1 : 0;
 }
 
+=head2 is_in_progress
+
+Sees if the problem is in an open, not 'confirmed' state.
+
+=cut
+
+sub is_in_progress {
```
pezholio: I know it's only a small method, but it'd be good if we could have some tests for this 😄

dracos: Sorry, I rebased by mistake, but the tests I added can be seen at https://github.com/mysociety/fixmystreet/pull/1658/commits/a91d9562c6c0ec836f10290df779e2012f18ddec#diff-63d32f28ee766541c25f7354acd8680c

PR #1661, opened 2017-03-07
Use name at time of moderation, including superusers.

### LINKED ISSUE #1660: Moderation sometimes doesn't give the user enough information
When I moderate a report, the message displayed on the edited report is:

"Moderated by at [time]"

i.e. it doesn't say who moderated it. This isn't very transparent and as a result, I don't use the moderation features - instead I edit the report from the admin interface to be able to add a note like "this report has been edited by FixMyStreet admin to remove [whatever was problematic about it]".

I don't know why this happens but am hazarding a guess that it is because my user is not associated with any body? Displaying "Moderated by a FixMyStreet administrator" would be better in whatever the circumstances are, anyway!

dracos:
Fixes #1660.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Moderate.pm
+++ b/perllib/FixMyStreet/App/Controller/Moderate.pm
@@ -83,6 +83,11 @@ sub moderate_report : Chained('report') : PathPart('') : Args(0) {
     $c->detach( 'report_moderate_audit', \@types )
 }
 
+sub moderating_user_name {
+    my $user = shift;
+    return $user->from_body ? $user->from_body->name : 'a FixMyStreet administrator';
+}
+
 sub report_moderate_audit : Private {
     my ($self, $c, @types) = @_;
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Moderate.pm
```diff
@@ -83,6 +83,11 @@ sub moderate_report : Chained('report') : PathPart('') : Args(0) {
     $c->detach( 'report_moderate_audit', \@types )
 }
 
+sub moderating_user_name {
```
pezholio: Sorry to sound like a broken record, but tests here would be ace too 👍

dracos: No, quite right!

dracos:
Just to note should check what moderated names will show once this is deployed (as it starts to use name recorded at time of moderation, not current from_body name).

PR #1665, opened 2017-03-09
Update FAQ

JenMysoc:
A custom FAQ for the CuidoMiCiudad site

### DIFF
--- a/templates/web/base/about/faq-en-gb.html
+++ b/templates/web/base/about/faq-en-gb.html
@@ -6,128 +6,136 @@
 
 <dl>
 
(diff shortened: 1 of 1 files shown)


dracos:
Hi, you shouldn't change the base template as that's what every cobrand uses as its default. You should add this as a new file, probably at `templates/web/cuidomiciudad/about/faq.html` in the `cuidomiciudad` branch of the fixmystreet-international repo: https://github.com/mysociety/fixmystreet-international/tree/cuidomiciudad/templates/web/cuidomiciudad

PR #1666, opened 2017-03-09
Move staff-only JavaScript to separate file.

dracos:
This reduces the main JS size for non-staff users from 58.6Kb to 44.6Kb (compressed from 28.2Kb to 21.8Kb) saving just under a quarter.

Seemed easiest to make the couple of points that called staff-functions just check they existed first, rather than anything involving hooks or listeners.


templates/web/base/common_scripts.html
```diff
@@ -16,6 +16,12 @@
     version('/cobrands/fixmystreet/fixmystreet.js'),
 );
 
+IF c.user_exists AND (c.user.from_body OR c.user.is_superuser);
+    scripts.push(
+        version('/cobrands/fixmystreet/staff.js')
+    );
+END;
+
```
pezholio: This is a really nice approach :+1:

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -222,6 +222,12 @@ fixmystreet.update_list_item_buttons = function($list) {
   $list.children(':last-child').find('[name="shortlist-down"]').prop('disabled', true);
 };
 
+fixmystreet.run = function(fn) {
+    if (fn) {
+        fn.call(this);
+    }
+};
+
```
pezholio: I saw this being called below and didn't understand what it was doing. I think I know now (it only calls a function if it exists, so it won't error for non-staff users), but it could do with a comment for clarification 👍

PR #1669, opened 2017-03-16
Show loading indicator when loading pins

pezholio:
This shows a loading indicator whenever pins are being loaded onto the map / sidebar, adding some code to show a div with a loading indicator inside at the start of the pan data request handler, and then adding some hiding functionality at the end of the callback that places the pins.

This is how it looks:

hscv49vgv0


templates/web/base/maps/openlayers.html
```diff
@@ -33,4 +33,6 @@
     [% IF map.copyright %]
     <div class="olControlAttribution" style="position: absolute;">[% map.copyright %]</div>
     [% END %]
-
+    <div id="loading-indicator" class="hidden" aria-hidden="true">
+      <img src="/i/loading.svg" alt="Loading.." />
```
dracos: Three dots for an ellipsis ;) And if we're editing this line anyway, might as well have the normal four space indent too.

web/js/map-OpenLayers.js
```diff
@@ -801,6 +801,9 @@ OpenLayers.Strategy.FixMyStreetFixed = OpenLayers.Class(OpenLayers.Strategy.Fixe
 // params to /ajax if the user has filtered the map.
 OpenLayers.Protocol.FixMyStreet = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
     read: function(options) {
+        // Show the loading indicator over the map
+        $('#loading-indicator').removeClass('hidden');
+        $('#loading-indicator').attr('aria-hidden', false);
```
dracos: Does adding this mean the `Check we haven't received the data after the map has been clicked.` section below can be removed? As it's not possible to click the map whilst this indicator is visible and thus cause the race condition it's acting against.

dracos: Although no harm in leaving that code in, I guess, though if we do it will need the 'remove the indicator code' putting in there too.

web/js/map-OpenLayers.js
```diff
@@ -844,6 +847,9 @@ OpenLayers.Format.FixMyStreet = OpenLayers.Class(OpenLayers.Format.JSON, {
         if (typeof(obj.pagination) != 'undefined') {
             $('.js-pagination').html(obj.pagination);
         }
+        // Remove loading indicator
```
dracos: I wonder if this should be higher (just before the `if typeof json == 'string'` line) so that if something goes wrong with the HTML setting (don't think this has happened yet), the person isn't stuck with the spinner.

web/js/map-OpenLayers.js
pezholio: I think removing it might be best, as it's not ever going to be triggered

pezholio: (I'm a big advocate of deleting code 😄 )

web/js/map-OpenLayers.js
pezholio: Yeah, good plan 👍

dracos:
This was rebased/merged as 5370647

PR #1672, opened 2017-03-17
Fix Open311 JSON output, remove superfluous lists.

dracos:


perllib/FixMyStreet/App/Controller/Open311.pm
```diff
@@ -233,44 +233,42 @@ sub output_requests : Private {
...
-#            'zipcode' => [ {} ],
-            'interface_used' => [ $problem->service ], # Not in Open311 v2
+            'service_request_id' => $id,
+            'title' => $problem->title, # Not in Open311 v2
+            'detail'  => $problem->detail, # Not in Open311 v2
+            'description' => $problem->title .': ' . $problem->detail,
+            'lat' => $problem->latitude,
+            'long' => $problem->longitude,
+            'status' => $problem->state,
+#            'status_notes' => {},
```
pezholio: Are these commented lines likely to come back? If not, it's probably best to just delete them

perllib/FixMyStreet/App/Controller/Open311.pm
```diff
@@ -233,44 +233,42 @@ sub output_requests : Private {
...
-            'requested_datetime' => [ w3date($problem->confirmed) ],
-            'updated_datetime' => [ w3date($problem->lastupdate) ],
-#            'expected_datetime' => [ {} ],
-#            'address' => [ {} ],
-#            'address_id' => [ {} ],
-            'service_code' => [ $problem->category ],
-            'service_name' => [ $problem->category ],
-#            'service_notice' => [ {} ],
-#            'zipcode' => [ {} ],
-            'interface_used' => [ $problem->service ], # Not in Open311 v2
```
pezholio: Why are we removing the lists? A bit of context would be useful in the commit 👍

perllib/FixMyStreet/App/Controller/Open311.pm
dracos: Note I haven't commented anything out, I just fixed the arrayrefs in them too :) They are other fields that an Open311 service could/should return, so I think it's good to leave them in.

pezholio: Oh cool. I figured something like that might be the case

perllib/FixMyStreet/App/Controller/Open311.pm
dracos: Yeah, bit brief! It's to stop the JSON output being `{ "service_request_id": [ 123456 ], "title": [ "TITLE" ], … }` which seemed a bit wrong! Adding the `NoAttr => 1` to the XML output stops that being any different (which is why I assume whoever wrote this had put these lists in in the first place).

pezholio: Aaaaaaaaah, gotcha 👍

Issue #1335, opened 2016-02-17
allow send-reports to accept a single report id (use --debug option?)


davewhiteland:
One reason to run `bin/send-reports --debug` is to see the error message coming back (if it's failing to send, of course), but currently when we do this we're actually looping through all unsent reports. This is often OK but it's annoying when there's a single report under investigation.

So it would be handy to be able to pass in a single report ID and have the script not bother looping in such a case. Normally this is emphatically not what you want `send-reports` to do (it _should_ send all the reports!) so I propose it's only relevant in debug mode. In which case, how about allowing the `--debug` option to pass in an optional report id? Maybe one or other of these (the first way is probably better because it prevents the risk of its dangerous effect being overlooked).

```
bin/send-reports --debug --send-single-report=1234

bin/send-reports --debug=1234
```

dracos:
Dupe of #732.

Issue #938, opened 2014-11-14
document the debug (and maybe verbose) options on send-reports script


davewhiteland:
The `--debug` option on `bin/send-reports` is useful if you need it... but it's not documented.

Maybe add a technical note to http://fixmystreet.org/customising/send_reports/

PR #1673, opened 2017-03-17
Upgrade to using Email::Sender

### LINKED ISSUE #1639: Can't use SMTP due to deprecated parameter in SSL.pm
Whenever I try to run send reports or try to create a new account under /auth I get this in logs:

*******************************************************************
 Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
 is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
 together with SSL_ca_file|SSL_ca_path for verification.
 If you really don't want to verify the certificate and keep the
 connection open to Man-In-The-Middle attacks please set
 SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.
*******************************************************************
  at /var/www/platform/fixmystreet/web/../local/lib/perl5/Email/Send/SMTP.pm line 59.

Apache log:

[Fri Feb 10 15:08:36.932178 2017] [fcgid:warn] [pid 4249:tid 140227360274176] [client 127.0.0.1:48210] mod_fcgid: stderr: [error] Caught exception in FixMyStreet::App::Controller::Auth->email_sign_in "invalid SSL_version specified at /var/www/platform/fixmystreet/web/../local/lib/perl5/IO/Socket/SSL.pm line 389.", referer: http://mywebsite.org/auth?r=my

dracos:
Should fix #1639. Could do with some real-world testing on Ubuntu trusty or xenial, both of which have shown to exhibit the issue.


pauliuszaleckas:
I should be able to test this on Monday.

pauliuszaleckas:
I had switched git to this branch and run bin/install_perl_modules
And got this error when sending e-mail:
To use SSL please install IO::Socket::SSL with version>=2.007 at /var/www/fixmystreet/fixmystreet/bin/../local/lib/perl5/Net/SMTP.pm line 70.

dracos:
Thanks for trying it out – sorry, I did not realise that dependency existed. I've added it to this branch now, if you'd like to try again.

pauliuszaleckas:
Now there is no errors. However email sending in my case fails, but "bin/send-reports --debug" doesn't show failure reason. Maybe we can use Email::Sender::Failure to print reason?
Do you know how to force email sending with bin/send-reports, because now there is some timeout and mostly I see message "report 2: skipped by sender Email (might be due to previous failed attempts?)" ?

pauliuszaleckas:
Learned some Perl and finally managed to get the error:
error at after DATA: Relaying disallowed as "Paulius Zaleckas" <fms-report-7-3f854e71@zaleckas.lt>
So problem is that FMS uses some strange e-mail, but not the one from DO_NOT_REPLY_EMAIL or even CONTACT_EMAIL.

dracos:
[Reply to your first message before I saw second] The code does log the Email::Sender::Failure for front-end emails (e.g. confirmation emails). I'll look at getting better logging for cron email too. I'll do some testing here when I am able, hopefully very soon. Timeout - sorry no, you have to reset the db currently, #1335 #938 refers.

[Reply to your second message] Yes, FixMyStreet uses VERP addresses for its SMTP envelope so that it can potentially handle bounces and the like. They are of the form fms-[unique-string]@[email-domain].

pauliuszaleckas:
So SSL works OK now. I changed code to use no-reply instead of VERP.

pauliuszaleckas:
TLS seems to be working OK too.

dracos:
Glad to hear that. I'll do a bit more testing here and add the logging I mentioned and then get this merged. If you'd like to be able to upgrade easily when the codebase changes, please do consider making your VERP/not-reply change configurable and submitting it upstream as a PR so that others can benefit from it, thanks :)

perllib/FixMyStreet/Email/Sender.pm
```diff
@@ -0,0 +1,48 @@
...
+Email::Send::Test).
+
+=cut
+
+sub build_default_transport {
+    if ( FixMyStreet->test_mode ) {
+        Email::Sender::Util->easy_transport(Test => {});
+    } elsif ( my $smtp_host = FixMyStreet->config('SMTP_SMARTHOST') ) {
+        my $type = FixMyStreet->config('SMTP_TYPE') || '';
+        my $port = FixMyStreet->config('SMTP_PORT') || '';
```
davea: `$port` doesn't get used, did you mean to pass it into `easy_transport`?

PR #1674, opened 2017-03-20
Customisable defect types

pezholio:
For https://github.com/mysociety/fixmystreetforcouncils/issues/160

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/DefectType.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/DefectType.pm
@@ -1,0 +1,22 @@
+package FixMyStreet::DB::ResultSet::DefectType;
+use base 'DBIx::Class::ResultSet';
+
+use strict;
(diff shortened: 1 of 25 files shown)


perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -0,0 +1,22 @@
...
+    my ($rs, $bodies, $category) = @_;
+    my $attrs = {
+        'me.body_id' => $bodies,
+    };
+    if ($category) {
+        $attrs->{'contact.category'} = [ $category, undef ];
+    }
+    $rs->search($attrs, {
+        order_by => 'name',
+        join => { 'contact_defect_types' => 'contact' },
```
dracos: I think this function is always called with a category, so doesn't really matter, but we don't want to do this join if $category is not set, I imagine.

PR #1676, opened 2017-03-29
Small translation/map improvements

davea:
A few little bits and pieces that have fallen out of work done for the CuidoMiCiudad site:

 - Easily style Google Maps & use API key
 - Make more of the site (moderation/offline/dashboard) translatable
 - Don't try and use easting/northing on every cobrand's inspector UI
 - Make it easier for cobrands to override nav link styles

### DIFF
--- a/locale/FixMyStreet.po
+++ b/locale/FixMyStreet.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: matthew@mysociety.org\n"
-"POT-Creation-Date: 2017-02-02 10:31+0000\n"
+"POT-Creation-Date: 2017-03-30 17:33+0100\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <team@fixmystreet.com>\n"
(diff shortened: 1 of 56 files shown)


locale/FixMyStreet.po
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: matthew@mysociety.org\n"
-"POT-Creation-Date: 2017-02-02 10:31+0000\n"
+"POT-Creation-Date: 2017-03-17 11:29-0400\n"
```
dracos: This needs redoing due to changes on master since the rebase, e.g. I see some defect type strings.

conf/general.yml-example
```diff
@@ -114,6 +114,7 @@ GEOCODER: ''
 # relevant key here.
 GOOGLE_MAPS_API_KEY: ''
 BING_MAPS_API_KEY: ''
+GOOGLE_MAPS_JS_API_KEY: ''
```
davea: `GOOGLE_MAPS_API_KEY` is used in the Perl for the geocoder stuff, whereas this is solely for the maps JS. [The docs](https://developers.google.com/maps/documentation/javascript/get-api-key) made me think they'd need to be two separate keys but happy to be proven wrong if it removes the need for an extra config key!

dracos: Google API keys are independent of the APIs they're used for nowadays :)

web/cobrands/fixmystreet/offline.js
```diff
@@ -2,17 +2,27 @@ fixmystreet.offlineBanner = (function() {
     var toCache = 0;
     var cachedSoFar = 0;
 
+    // Extremely noddy function
```
davea: 💯

PR #1677, opened 2017-03-29
Suggest duplicate

pezholio:
Opening for discussion. Probably not ready for merge yet!

This attempts to reduce the amount of duplicate reports councils receive by searching within a 50 metre radius for possible duplicates. If a duplicate is found, a user can click a ⭐  to subscribe to a report and receive updates. This also adds a hidden comment. It currently doesn't do anything to notify the council, but this can be easily tweaked to offer this. Here's how it looks:

odslnr600l

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,16 +48,34 @@ web-based cross-browser testing tools for this project.
 * In development
     - Front end improvements:
         - Add ability to make map full screen on mobile report pages. #1655
+        - Move staff-only JavaScript to separate file. #1666
+        - Show loading indicator when loading pins. #1669
         - Allow users to reopen closed reports. #1607
     - Admin improvements:
         - Redirect to category-filtered /reports on login if present. #1622
         - Follow redirect to /admin after login if allowed. #1622
         - Include /admin link on top-level nav for admin users.
         - Add shortlist filters. #1629
+        - Add submit buttons to admin index search forms. #1551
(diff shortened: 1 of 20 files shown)


zarino:
At this point in the process…

still

…it’s kinda hard to tell whether the suggested duplicate is, in fact, the same as the issue you’re trying to report. All you have to go by is the report summary (often quite vague!) and the photo (too small). It's just not enough.

We could pull in more detail (eg: description, click-to-enlarge photo) for each possible duplicate, but whatever we do, we'll be trying to cram a lot of information (that already exists on the dedicated problem page) into a small space in the middle of the reporting form.

Perhaps we could break the process into two steps?

1. First, we show a list of possible duplicates—as in your mockup above—which the user can click to visit that report’s full detail page. 
2. Then, on the report detail page, there’s a prominent "I spotted this problem too" button, which adds the hidden comment, subscribes you to updates, and displays the "Thanks" message.

If we did this, the user would need to be able to _return_ to the "Report a Problem" page (ideally with the pin location, category, _and suggested duplicates_ already displayed) so they can either investigate other suggested duplicates, or carry on reporting a new problem.

Does that sound feasible, @pezholio?

An additional improvement would be to display pins for each of the suggested duplicates, on the map, and make them fade in/out when you hover over the suggestions, like on the `/around` page. That way, you can see the _location_ of the duplicates, as well as their summary and photo.

PR #1684, opened 2017-04-08
Update README.md

saharjs:
Hi
Required by the one of the assignments of my technical writing course I am supposed to edit other people's writing. I think the minor changes that I have proposed would increase the readability of your README file.
Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,12 @@ FixMyStreet Platform is an open source project to help people run websites for
 reporting common street problems such as potholes and broken street lights to
 the appropriate authority.
 
-Users locate problems using a combination of address and sticking a pin
-in a map without worrying about the correct authority to report it to.
+Users locate problems using a combination of address and droping a pin
+in a map without being worried about the correct authority to report it to.
 FixMyStreet then works out the correct authority using the problem location and
-type and sends a report, by email or using a web service such as Open311.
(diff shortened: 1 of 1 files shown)


dracos:
Thanks very much for finding our repository and making a contribution. I hope that you find the following review of your changes helpful and will continue to improve and make contributions to open source repositories in the future.

Firstly, I agree that "sticking" may not be the best verb to use – it is meant to have the connotation of sticking a physical pin in a physical map, but that might not work everywhere or if you're not a native English speaker. However, 'droping' is misspelled (it should be 'dropping'), and I'm not sure is a better word to use, as you are no more dropping a pin than you are sticking it. "Placing a pin on a map" would probably be what I would go for, though better suggestions are welcome.

`problem location and type and sends` is clumsy, but I don't think splitting it in two and using another "Then" makes it less so. Another comma before the final and would help.

Making long sentences smaller can help with readability, but in the other instances you have done this I think it actually makes it harder, as each time the second part of the sentence is in reference to the first. The "please see our site" is referring to seeing how we are "working hard"; and the "please do search through the other issues" is clarifying that the label previously referred to may not be what you're after.

Lastly, you removed "and leave updates" from one sentence, which was a wanted part of that sentence, informing people that FixMyStreet can be used to leave updates on existing reports, not just view them.

PR #1693, opened 2017-05-02
[Oxfordshire] Include nearest text in location field in RDI output

dracos:
This includes the output in the same style as currently sent to Exor (so without the more explanatory text the emails have). This PR also takes the opportunity to allow find_closest to be called more than once, and tidies up its signature and a few other related bits.


perllib/FixMyStreet/App/Controller/Admin/ExorDefects.pm
```diff
@@ -140,6 +140,18 @@ sub download : Path('download') : Args(0) {
 
         foreach my $report (@$inspections) {
             my ($eastings, $northings) = $report->local_coords;
+
+            my $location = "${eastings}E ${northings}N";
+            $location = "[DID NOT USE MAP] $location" unless $report->used_map;
+            my $closest_address = $c->cobrand->find_closest($report);
+            if ($closest_address) {
+                # Match code in bin/oxfordshire/open311_service_request.cgi
```
pezholio: As we're doing this twice, could we not pass a flag to the `find_closest` function to just return the address minus the other cruft that we're stripping out here?

dracos: Yeah, I wondered about that; we can't/won't change the bin/oxfordshire code, so thought it might be better to match it, but it would be nicer to get some data back, you're right.

PR #1694, opened 2017-05-04
Subscribe inspectors to alerts after updating an issue

### LINKED ISSUE #130: 'All reports' pages for councils that no longer exist
should make it clear that the council no longer exists.

e.g. http://www.fixmystreet.com/reports/Blyth+Valley

It would also be good to say which council(s) now cover the area in question.

pezholio:
Fixes https://github.com/mysociety/fixmystreetforcouncils/issues/130

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -124,7 +124,7 @@ sub load_problem_or_display_error : Private {
         $c->detach( '/page_error_404_not_found', [ _('Unknown problem ID') ] )
             unless $c->cobrand->show_unconfirmed_reports ;
     }
-    elsif ( $problem->hidden_states->{ $problem->state } or 
+    elsif ( $problem->hidden_states->{ $problem->state } or
             (($problem->get_extra_metadata('closure_status')||'') eq 'hidden')) {
         $c->detach(
             '/page_error_410_gone',
@@ -388,6 +388,15 @@ sub inspect : Private {
                     $reputation_change = 1 if $c->cobrand->reputation_increment_states->{$state};
                     $reputation_change = -1 if $c->cobrand->reputation_decrement_states->{$state};
                 }
+
+                # If an inspector has changed the state, subscribe them to
+                # updates
+                my $options = {
+                    cobrand      => $c->cobrand->moniker,
+                    cobrand_data => $problem->cobrand_data,
+                    lang         => $problem->lang,
+                };
+                $problem->user->create_alert($problem->id, $options);
             }
         }
 
--- a/perllib/FixMyStreet/App/Controller/Report/Update.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/Update.pm
@@ -548,24 +548,17 @@ sub signup_for_alerts : Private {
     my ( $self, $c ) = @_;
 
     my $update = $c->stash->{update};
+    my $user = $update->user;
+    my $problem_id = $update->problem_id;
+
     if ( $c->stash->{add_alert} ) {
         my $options = {
-            user => $update->user,
-            alert_type => 'new_updates',
-            parameter => $update->problem_id,
+            cobrand => $update->cobrand,
+            cobrand_data => $update->cobrand_data,
+            lang => $update->lang,
         };
-        my $alert = $c->model('DB::Alert')->find($options);
-        unless ($alert) {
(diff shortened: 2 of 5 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -454,6 +454,23 @@ sub inspect : Private {
                 $redirect_uri = $c->uri_for_action('my/planned');
             }
 
+            # Sign up to alerts by default
+            my $options = {
```
dracos: As this code very much overlaps with that in `Report/Update.pm`, could this be factored out? Presumably as a method on either Problem or User, can't decide which :)

pezholio: I hear ya. I'm very much in favour of taking as much logic out of the controller as possible, so I'll take a look at this 👍

pezholio: I'm thinking User would be best, as it's acting on the user, rather than the problem

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -204,6 +204,27 @@ sub alert_for_problem {
...
 }
 
+=head2 create_alert
+
+Sign a user up to receive alerts on a given problem
+
+=cut
+
+sub create_alert {
+    my ( $self, $id, %options ) = @_;
```
dracos: Given every call to this is derefencing a hashref, perhaps accept a hashref instead and deference it inside this function?

t/app/model/user.t
```diff
@@ -30,6 +30,32 @@ is $problem->user->latest_anonymity, 0, "User's last update was not anonyous";
...
 
+subtest "Sign user up for alerts" => sub {
+    my $user = $problem->user;
+    my $options = {
+      cobrand      => 'default',
+      lang         => 'en-gn',
+    };
+    $user->create_alert($problem->id, %$options);
+
+    my $alert = $user->alert_for_problem( $problem->id );
```
dracos: Just to be sure, you should check there isn't any alert at the start of this subtest (otherwise it would just be checking that create_alert doesn't create another one, which also could be something else to check).

t/app/controller/report_inspect.t
```diff
@@ -129,6 +129,21 @@ FixMyStreet::override_config {
         $report->update({ state => $old_state });
     };
 
+    subtest "changing state subscribes inspector to alerts" =>sub {
```
dracos: The alert was actually added in the test around line 60, the first time the form is submitted with a state change. You could check that the alert is added there as an addition to that test; this test is only checking that another state change submission doesn't create another alert, not what it says it is testing. I don't know what the `$old_state`, deleting comments, or `discard_changes` are doing, this feels like it's been copied-and-pasted.

perllib/FixMyStreet/App/Controller/Report/Update.pm
```diff
@@ -548,24 +548,17 @@ sub signup_for_alerts : Private {
...
     my $update = $c->stash->{update};
+    my $user = $update->user;
+    my $problem_id = $update->problem_id;
+
     if ( $c->stash->{add_alert} ) {
         my $options = {
-            user => $update->user,
-            alert_type => 'new_updates',
-            parameter => $update->problem_id,
+          cobrand      => $update->cobrand,
```
dracos: Style guide, as with the code this is replacing, is not to align, because otherwise there are needless diffs if the list of keys ever changes.

t/app/controller/report_inspect.t
pezholio: Ah, of course. I forgot that the test up there will affect state further down

perllib/FixMyStreet/App/Controller/Report/Update.pm
```diff
@@ -548,24 +548,17 @@ sub signup_for_alerts : Private {
...
+    my $problem_id = $update->problem_id;
+
     if ( $c->stash->{add_alert} ) {
         my $options = {
-            user => $update->user,
-            alert_type => 'new_updates',
-            parameter => $update->problem_id,
+        cobrand      => $update->cobrand,
+        cobrand_data => $update->cobrand_data,
+        lang         => $update->lang,
```
pezholio: Oh, that makes sense. That did seem like a weird request 😄

pezholio:
Gaaaah!

PR #1696, opened 2017-05-05
[Oxfordshire] Send RDI reports by email.

dracos:
This factors out the RDI reporting, and adds a script to send RDI files every day, for each inspector.
For https://github.com/mysociety/fixmystreetforcouncils/issues/183

### DIFF
--- a/bin/oxfordshire/send-rdi-emails
+++ b/bin/oxfordshire/send-rdi-emails
@@ -1,0 +1,69 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use DateTime;
+use Try::Tiny;
+use FixMyStreet;
+use FixMyStreet::Cobrand;
+use FixMyStreet::Email;
+use FixMyStreet::Integrations::ExorRDI;
+
+my $end_date = DateTime->now( time_zone => FixMyStreet->time_zone || FixMyStreet->local_time_zone )
+    ->truncate(to => 'hour')->set_hour(16);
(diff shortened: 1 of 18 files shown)


bin/oxfordshire/send-rdi-emails
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env perl
```
pezholio: Could this code be factored out into a package in perllib/FixMyStreet/scripts? Would (maybe) make it easier to test too? Certainly would make it easier to reuse if we need to further down the line

bin/oxfordshire/send-rdi-emails
```diff
@@ -0,0 +1,68 @@
...
+
+my $cobrand = FixMyStreet::Cobrand->get_class_for_moniker('oxfordshire')->new;
+my @inspectors = $cobrand->users->search({
+    'user_body_permissions.permission_type' => 'report_inspect'
+}, {
+    join => 'user_body_permissions',
+    distinct => 1,
+})->all;
+
+foreach my $inspector (@inspectors) {
```
dracos: That's the spec: https://github.com/mysociety/fixmystreetforcouncils/issues/183#issuecomment-297329233

bin/oxfordshire/send-rdi-emails
dracos: I think I'd prefer leaving that for such a time as that might (hopefully never) happen ;) It's not a very long script, the main bulk is already factored out generating the RDI, other than that it's only a small loop.

bin/oxfordshire/send-rdi-emails
pezholio: Oh yeah, `one email per inspector`. Carry on then 😄

bin/oxfordshire/send-rdi-emails
pezholio: Fair enough 👍

dracos:
Merged as 61eb613dd

PR #1698, opened 2017-05-08
Change defect type when category is changed

pezholio:
Fixes mysociety/fixmystreetforcouncils#185

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -312,27 +312,10 @@ sub inspect : Private {
     $c->forward('/admin/categories_for_point');
     $c->stash->{report_meta} = { map { $_->{name} => $_ } @{ $c->stash->{problem}->get_extra_fields() } };
 
-    my %category_body = map { $_->category => $_->body_id } map { $_->contacts->all } values %{$problem->bodies};
-
-    my @priorities = $c->model('DB::ResponsePriority')->for_bodies($problem->bodies_str_ids)->all;
-    my $priorities_by_category = {};
-    foreach my $pri (@priorities) {
-        my $any = 0;
-        foreach ($pri->contacts->all) {
-            $any = 1;
-            push @{$priorities_by_category->{$_->category}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
-        }
-        if (!$any) {
-            foreach (grep { $category_body{$_} == $pri->body_id } @{$c->stash->{categories}}) {
-                push @{$priorities_by_category->{$_}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
-            }
-        }
+    if ($c->cobrand->can('council_area_id')) {
+        my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+        $c->stash->{priorities_by_category} = $priorities_by_category;
     }
-    foreach (keys %{$priorities_by_category}) {
-        $priorities_by_category->{$_} = join('&', @{$priorities_by_category->{$_}});
-    }
-
-    $c->stash->{priorities_by_category} = $priorities_by_category;
 
     if ( $c->get_param('save') ) {
         $c->forward('/auth/check_csrf_token');
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -633,6 +633,12 @@ sub setup_categories_and_bodies : Private {
     # keysort does not appear to obey locale so use strcoll (see i18n.t)
     @contacts = sort { strcoll( $a->category, $b->category ) } @contacts;
 
+    # Get defect types for inspectors
+    if ($c->cobrand->can('council_area_id')) {
+        my $category_defect_types = FixMyStreet::App->model('DB::DefectType')->by_categories($c->cobrand->council_area_id, @contacts);
+        $c->stash->{category_defect_types} = $category_defect_types;
+    }
+
     my %seen;
     foreach my $contact (@contacts) {
 
--- a/perllib/FixMyStreet/DB/ResultSet/DefectType.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/DefectType.pm
@@ -3,20 +3,26 @@ use base 'DBIx::Class::ResultSet';
 
 use strict;
 use warnings;
+use Moo;
+use HTML::Entities;
 
-sub for_bodies {
-    my ($rs, $bodies, $category) = @_;
-    my $attrs = {
-        'me.body_id' => $bodies,
-    };
-    if ($category) {
-        $attrs->{'contact.category'} = [ $category, undef ];
-    }
-    $rs->search($attrs, {
-        order_by => 'name',
-        join => { 'contact_defect_types' => 'contact' },
-        distinct => 1,
-    });
+with('FixMyStreet::Roles::ContactExtra');
+
+sub join_table {
+    return 'contact_defect_types';
+}
+
+sub map_extras {
+    my ($rs, @ts) = @_;
+    return map {
+        my $meta = $_->get_extra_metadata();
+        my %extra = map { $_ => encode_entities($meta->{$_}) } keys %$meta;
+        {
+            id => $_->id,
+            name => encode_entities($_->name),
+            extra => \%extra
+        }
+    } @ts;
 }
 
 1;
--- a/perllib/FixMyStreet/DB/ResultSet/ResponsePriority.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/ResponsePriority.pm
@@ -3,20 +3,18 @@ use base 'DBIx::Class::ResultSet';
 
 use strict;
 use warnings;
+use Moo;
+use HTML::Entities;
 
-sub for_bodies {
-    my ($rs, $bodies, $category) = @_;
-    my $attrs = {
-        'me.body_id' => $bodies,
-    };
-    if ($category) {
-        $attrs->{'contact.category'} = [ $category, undef ];
-    }
-    $rs->search($attrs, {
-        order_by => 'name',
-        join => { 'contact_response_priorities' => 'contact' },
-        distinct => 1,
-    });
+with('FixMyStreet::Roles::ContactExtra');
+
+sub join_table {
+    return 'contact_response_priorities';
+}
+
+sub map_extras {
+    my ($rs, @ts) = @_;
+    return map { { id => $_->id, name => encode_entities($_->name) } } @ts;
 }
 
 1;
--- a/perllib/FixMyStreet/Roles/ContactExtra.pm
+++ b/perllib/FixMyStreet/Roles/ContactExtra.pm
@@ -1,0 +1,43 @@
+package FixMyStreet::Roles::ContactExtra;
+
+use Moo::Role;
+use JSON::MaybeXS;
+
+requires 'join_table', 'map_extras';
+
+sub for_bodies {
+    my ($rs, $bodies, $category) = @_;
+    my $join_table = $rs->join_table();
+    my $attrs = {
+        'me.body_id' => $bodies,
+    };
+    my $filters = {
+        order_by => 'name',
+        join => { $join_table => 'contact' },
+        distinct => 1,
+    };
+    if ($category) {
+        $attrs->{'contact.category'} = [ $category, undef ];
+    }
+    $rs->search($attrs, $filters);
+}
+
+sub by_categories {
+    my ($rs, $area_id, @contacts) = @_;
+    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my @body_ids = keys %body_ids;
+    my %extras = ();
+    my @results = $rs->for_bodies(\@body_ids, undef);
+    @contacts = grep { $body_ids{$_->body_id} } @contacts;
+
+    foreach my $contact (@contacts) {
+        my $join_table = $rs->join_table();
+        my @ts = grep { !defined($_->$join_table->first) || $_->$join_table->find({contact_id => $contact->get_column('id')}) } @results;
+        @ts = $rs->map_extras(@ts);
+        $extras{$contact->category} = encode_json(\@ts);
+    }
+
+    return \%extras;
+}
+
+1;
--- a/t/app/model/defecttype.t
+++ b/t/app/model/defecttype.t
@@ -1,21 +1,31 @@
 use FixMyStreet::App;
 use FixMyStreet::TestMech;
-my $mech = FixMyStreet::TestMech->new;
+use JSON::MaybeXS;
 
-my $oxfordshire = $mech->create_body_ok(2237, 'Oxfordshire County Council');
+my $mech = FixMyStreet::TestMech->new;
+my $area_id = 2237;
+
+my $oxfordshire = $mech->create_body_ok($area_id, 'Oxfordshire County Council');
+my $other_body = $mech->create_body_ok($area_id, 'Some Other Council');
 my $potholes_contact = $mech->create_contact_ok( body_id => $oxfordshire->id, category => 'Potholes', email => 'potholes@example.com' );
 my $traffic_lights_contact =$mech->create_contact_ok( body_id => $oxfordshire->id, category => 'Traffic lights', email => 'lights@example.com' );
+my $pavements_contact =$mech->create_contact_ok( body_id => $oxfordshire->id, category => 'Pavements', email => 'pavements@example.com' );
 
 my $potholes_defect_type = FixMyStreet::App->model('DB::DefectType')->find_or_create(
     {
         body_id => $oxfordshire->id,
-        name => 'Potholes',
-        description => 'This defect type is to do with potholes'
+        name => 'Potholes and Pavements',
+        description => 'This defect type is to do with potholes and Pavements'
     }
 );
+$potholes_defect_type->set_extra_metadata('defect_code' => 123);
 $potholes_defect_type->contact_defect_types->find_or_create({
     contact_id => $potholes_contact->id,
 });
+$potholes_defect_type->contact_defect_types->find_or_create({
+    contact_id => $pavements_contact->id,
+});
+$potholes_defect_type->update();
 
 my $general_defect_type = FixMyStreet::App->model('DB::DefectType')->find_or_create(
     {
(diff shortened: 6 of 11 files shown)


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -169,6 +169,8 @@ sub admin_pages {
         $pages->{defecttype_edit} = [ undef, undef ];
     };
 
+    $pages->{body} = [ undef, undef ];
```
pezholio: This is a bit of a hack to get the `defect_types_for_category` url to show up in `check_page_allowed`

templates/web/base/admin/report-category.html
```diff
@@ -7,7 +7,7 @@
   [% IF categories.size %]
     <optgroup label="[% loc('Available categories') %]">
       [% FOREACH cat IN categories %]
-        <option[% ' selected' IF problem.category == cat %]>[% cat | html %]</option>
+        <option[% ' selected' IF problem.category == cat %] data-defect-types='[% category_defect_types.$cat %]'>[% cat | html %]</option>
```
dracos: This doesn't feel like the right place for this, because this template is also included by the main admin's report edit page. Would it not fit better inside the loop at https://github.com/mysociety/fixmystreet/blob/a3c7641247b16fd4fd959eb95c54f1f81c6f4ba7/templates/web/base/report/_inspect.html#L42 to match the existing priorities code that appears to be doing the same functionality for the priority dropdown? (Would also therefore be good to match (share a function) how that does it for consistency, ie. it is deleting all but the first option, not deleting them all and then putting back a `-` option; could/should the priorities code use JSON too?)

web/cobrands/fixmystreet/staff.js
```diff
@@ -211,6 +211,19 @@ $.extend(fixmystreet.set_up, {
             $select.append($('<option>', { value: kv[0], text: decodeURIComponent(kv[1]) }));
         });
         $select.val(curr_pri);
+        // And the defect types
```
dracos: Just to reiterate, it would be good if this and the priorities code just above could be shared/act the same, given they are both recreating a select given some data.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -644,6 +645,14 @@ sub setup_categories_and_bodies : Private {
             my $metas = $contact->get_metadata_for_input;
             $category_extras{$contact->category} = $metas if @$metas;
 
+            if ($c->cobrand->can('council_id')) {
```
dracos: This makes one query per category; again, it would seem sensible to match how the priorities code handles it (which is around https://github.com/mysociety/fixmystreet/blob/afa66d3cd1b4d84e151bd8c438060fc8fc340bbe/perllib/FixMyStreet/App/Controller/Report.pm#L314 ) - fetching all the priorities and grouping them by category.

In fact, this code will not handle defect types that aren't associated with a category (which should therefore be associated with all categories), but the priority code does handle that.

pezholio: `for_bodies` handles your second point, see https://github.com/mysociety/fixmystreet/blob/515362a13bcaff003143df6d455251fb1c58f913/perllib/FixMyStreet/DB/ResultSet/DefectType.pm#L13. That said, it would make sense to cut down on the number of queries.

pezholio:
@dracos This should be ready for another code review pass now. I've factored out the response priority logic into the model too. Much more pleased with how this works now 😄

pezholio:
I've just noticed I need to apply `defect_type_format`s to the defect type dropdowns, so this may not be quite ready yet

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -311,25 +311,10 @@ sub inspect : Private {
...
-            $any = 1;
-            push @{$priorities_by_category->{$_->category}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
-        }
-        if (!$any) {
-            foreach (grep { $category_body{$_} == $pri->body_id } @{$c->stash->{categories}}) {
-                push @{$priorities_by_category->{$_}}, $pri->id . '=' . URI::Escape::uri_escape_utf8($pri->name);
-            }
-        }
+    if ($c->cobrand->can('council_id')) {
+      my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($c->cobrand->council_id, @{$c->stash->{contacts}});
```
dracos: Spacing! ;)

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -3,20 +3,38 @@ use base 'DBIx::Class::ResultSet';
...
+    };
+    if ($category) {
+        $attrs->{'contact.category'} = [ $category, undef ];
+    }
+    $rs->search($attrs, $filters);
+}
+
+sub by_categories {
+    my ($rs, $area_id, @contacts) = @_;
+    my $body_id = FixMyStreet::DB->resultset('BodyArea')->find({ area_id => $area_id })->body_id;
```
dracos: Theoretically, an area could cover multiple bodies so it would be good to cover this and have it fetch all the matching bodies.

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -3,20 +3,38 @@ use base 'DBIx::Class::ResultSet';
...
+    my %defect_types = ();
+    my @results = $rs->for_bodies([ $body_id ], undef);
+
+    foreach my $contact (grep { $_->body_id == $body_id } @contacts) {
+      my @ts = grep { !defined($_->contact_defect_types->first) || $_->contact_defect_types->find({contact_id => $contact->get_column('id')}) } @results;
+      use Data::Dumper;
```
dracos: Debug

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
pezholio: I've got the code to do this, but when writing tests and trying to create the extra body like so:

```
my $area_id = 123;
my $other_body = $mech->create_body_ok($area_id, 'Some Other Council');
FixMyStreet::App->model('DB::BodyArea')->create( { body => $other_body, area_id => $area_id } );
```

I get an error from Postgres:

```
DBIx::Class::ResultSet::create(): DBI Exception: DBD::Pg::st execute failed: ERROR:  duplicate key value violates unique constraint "body_areas_body_id_area_id_idx"
DETAIL:  Key (body_id, area_id)=(1, 123) already exists. [for Statement "INSERT INTO body_areas ( area_id, body_id) VALUES ( ?, ? )" with ParamValues: 1='123', 2='1'] at t/app/model/responsepriority.t line 73
    # Child (by_categories returns defect types if an area is covered by more than one body) exited without calling finalize()
```

This suggests that there's a constraint that stops bodies having multiple areas, so this will never happen. Am I missing something?

dracos: That error says you're trying to assign the *same* area to a body twice. If you assign a different area, it should be fine?

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -3,20 +3,39 @@ use base 'DBIx::Class::ResultSet';
...
+sub by_categories {
+    my ($rs, $area_id, @contacts) = @_;
+    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my @body_ids = keys %body_ids;
+    my %defect_types = ();
+    my @results = $rs->for_bodies(\@body_ids, undef);
+    @contacts = grep { $body_ids{$_->body_id} } @contacts;
+
+    foreach my $contact (@contacts) {
+      my @ts = grep { !defined($_->contact_defect_types->first) || $_->contact_defect_types->find({contact_id => $contact->get_column('id')}) } @results;
```
dracos: Spacing… (can your editor not have different settings by filetype or similar?)

perllib/FixMyStreet/DB/ResultSet/ResponsePriority.pm
```diff
@@ -3,20 +3,39 @@ use base 'DBIx::Class::ResultSet';
...
 use warnings;
+use JSON::MaybeXS;
 
 sub for_bodies {
     my ($rs, $bodies, $category) = @_;
     my $attrs = {
         'me.body_id' => $bodies,
     };
+    my $filters = {
+      order_by => 'name',
```
dracos: Spacing

perllib/FixMyStreet/DB/ResultSet/ResponsePriority.pm
```diff
@@ -3,20 +3,39 @@ use base 'DBIx::Class::ResultSet';
...
+sub by_categories {
+    my ($rs, $area_id, @contacts) = @_;
+    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my @body_ids = keys %body_ids;
+    my %priorities = ();
+    my @results = $rs->for_bodies(\@body_ids, undef);
+    @contacts = grep { $body_ids{$_->body_id} } @contacts;
+
+    foreach my $contact (@contacts) {
+      my @ts = grep { $_->contact_response_priorities->find({contact_id => $contact->get_column('id')}) } @results;
```
dracos: Spacing

templates/web/base/report/_inspect.html
```diff
@@ -43,11 +43,11 @@
           [% cat_prefix = category | lower | replace('[^a-z]', '') %]
           [% cat_prefix = "category_" _ cat_prefix _ "_" %]
           [% IF category == problem.category %]
-            <p data-category="[% category | html %]" data-priorities="[% priorities_by_category.$category %]">
+            <p data-category="[% category | html %]" data-priorities='[% priorities_by_category.$category %]' data-defect-types='[% category_defect_types.$category %]'>
```
dracos: What if one of the defect types of priorities contains an apostrophe? I don't think JSON encoded will escape apostrophes?

templates/web/base/report/_inspect.html
```diff
@@ -43,11 +43,11 @@
...
               [% INCLUDE 'report/new/category_extras_fields.html' %]
             </p>
           [% ELSE %]
-            <p data-category="[% category | html %]" class="hidden" data-priorities="[% priorities_by_category.$category %]">
+            <p data-category="[% category | html %]" class="hidden" data-priorities='[% priorities_by_category.$category %]' data-defect-types='[% category_defect_types.$category %]'>
```
dracos: Ditto here

perllib/FixMyStreet/DB/ResultSet/ResponsePriority.pm
dracos: You have not made the change here that we discussed at https://mysociety.slack.com/archives/D41T2K7T5/p1494601362486772 - this line should be the same as the defect types code.
In fact, these two functions are practically identical, can you factor them out to one function? The only difference looks to be the many-many lookup table used, and that defect types include `extra` in their output and priorities do not. You could use a Role to store the shared function (ie. put it under FixMyStreet/Roles and use `with` in the two ResultSet files). The same presumably applies to `for_bodies`.

t/app/model/defecttype.t
```diff
@@ -59,6 +68,44 @@ subtest 'Problem->defect_types behaves correctly' => sub {
     is $problem->defect_types->first->name, $general_defect_type->name, 'Correct defect type is returned for Traffic lights category';
 };
 
+subtest 'by_categories returns all defect types grouped by category' => sub {
+    FixMyStreet::App->model('DB::BodyArea')->create( { body => $oxfordshire, area_id => $area_id } );
```
dracos: I don't understand what this line is for. `create_body_ok` already puts an entry in BodyArea – if your aim is for body ID and area ID not to match, remove the `id => 2237` from the `create_body_ok` call above, no need for it to be in two areas here.

t/app/model/defecttype.t
```diff
@@ -59,6 +68,44 @@ subtest 'Problem->defect_types behaves correctly' => sub {
...
+
+    my @contacts = FixMyStreet::DB->resultset('Contact')->not_deleted->search( { body_id => [ $oxfordshire->id ] } )->all;
+    my $defect_types = FixMyStreet::App->model('DB::DefectType')->by_categories($oxfordshire->id, @contacts);
```
dracos: Is it not an area ID that is passed to this, not a body ID? The area ID is 2237, that is fine.

t/app/model/defecttype.t
```diff
@@ -59,6 +68,44 @@ subtest 'Problem->defect_types behaves correctly' => sub {
...
+    is scalar @$potholes, 2, 'Potholes have 2 defect types';
+    is scalar @$traffic_lights, 1, 'Traffic lights have 1 defect type';
+    is scalar @$pavements, 2, 'Pavements have 2 defect types';
+
+    is @$potholes[1]->{extra}->{defect_code}, 123, 'Defect code is present';
+};
+
+subtest 'by_categories returns defect types for an area with multiple bodies' => sub {
+    my $other_body = $mech->create_body_ok('345', 'Some Other Council');
+    FixMyStreet::App->model('DB::BodyArea')->create( { body => $other_body, area_id => $area_id } );
```
dracos: Area ID is the first argument to `create_body_ok`, so you could put $area_id there and not need this line.

t/app/model/defecttype.t
```diff
@@ -59,6 +68,44 @@ subtest 'Problem->defect_types behaves correctly' => sub {
...
+
+    is scalar @$potholes, 2, 'Potholes have 2 defect types';
+    is scalar @$traffic_lights, 1, 'Traffic lights have 1 defect type';
+    is scalar @$pavements, 2, 'Pavements have 2 defect types';
+
+    is @$potholes[1]->{extra}->{defect_code}, 123, 'Defect code is present';
+};
+
+subtest 'by_categories returns defect types for an area with multiple bodies' => sub {
+    my $other_body = $mech->create_body_ok('345', 'Some Other Council');
```
dracos: Could this body be deleted at the end.

t/app/model/responsepriority.t
```diff
@@ -0,0 +1,97 @@
...
+    }
+);
+$general_response_priority->contact_response_priorities->find_or_create({
+    contact_id => $potholes_contact->id,
+});
+$general_response_priority->contact_response_priorities->find_or_create({
+    contact_id => $traffic_lights_contact->id,
+});
+
+FixMyStreet::App->model('DB::BodyArea')->create( { body => $oxfordshire, area_id => $area_id } );
```
dracos: Same as with other tests, this line shouldn't be needed, I think, remove the ID from Oxfordshire instead to test different body/area ID.

t/app/model/responsepriority.t
```diff
@@ -0,0 +1,97 @@
...
+    my $priorities = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($area_id, @contacts);
+    my $potholes = decode_json($priorities->{Potholes});
+    my $traffic_lights = decode_json($priorities->{'Traffic lights'});
+
+    is scalar @$potholes, 2, 'Potholes have 2 defect types';
+    is scalar @$traffic_lights, 1, 'Traffic lights have 1 defect type';
+};
+
+subtest 'by_categories returns all response priorities for an area with multiple bodies' => sub {
+    my $other_body = $mech->create_body_ok('345', 'Some Other Council');
```
dracos: Again, ditto other test.

templates/web/base/report/_inspect.html
pezholio: Fixed this by using `encode_entities`, i.e https://github.com/mysociety/fixmystreet/pull/1698/files#diff-7e531cfc30ebb4348e8295a6a635ebbfR17

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -311,25 +311,10 @@ sub inspect : Private {
 
     my %category_body = map { $_->category => $_->body_id } map { $_->contacts->all } values %{$problem->bodies};
```
dracos: This does not appear to be used any more.

perllib/FixMyStreet/Roles/ContactExtra.pm
```diff
@@ -0,0 +1,49 @@
...
+
+sub for_bodies {
+    my ($rs, $bodies, $category) = @_;
+    my $join_table = $rs->join_table();
+    my $attrs = {
+        'me.body_id' => $bodies,
+    };
+    my $filters = {
+        order_by => 'name',
+        join => { "$join_table" => 'contact' },
```
dracos: No need for quotes around a string variable.

perllib/FixMyStreet/Roles/ContactExtra.pm
```diff
@@ -0,0 +1,49 @@
+package FixMyStreet::Roles::ContactExtra;
+
+use Moo::Role;
+use JSON::MaybeXS;
+
+requires 'join_table', 'map_extras';
+
+sub fetch_bodies {
```
dracos: Was the idea to use this function in by_categories? It's currently not being used.

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -3,20 +3,18 @@ use base 'DBIx::Class::ResultSet';
...
-    });
+with('FixMyStreet::Roles::ContactExtra');
+
+sub join_table {
+    return 'contact_defect_types';
+}
+
+sub map_extras {
+    my ($rs, @ts) = @_;
+    return map { { id => $_->id, name => encode_entities($_->name), extra => $_->get_extra_metadata() } } @ts;
```
dracos: What if some of the extra metadata contains apostrophes?

dracos:
Think DaveA is doing a fuller review, but just to quickly point out this has some bugs client side that I don't think used to be there. Picking defect type and changing categories couple of times leads to e.g.

screen shot 2017-05-30 at 09 17 00

Or if a category has a blank data-defect-types (e.g. if it's a different council's category in a two-tier area) then the `map` throws an error.

pezholio:
Ooooh, good spot. I'll have a squiz at that

perllib/FixMyStreet/DB/ResultSet/DefectType.pm
```diff
@@ -3,20 +3,31 @@ use base 'DBIx::Class::ResultSet';
...
-    });
+with('FixMyStreet::Roles::ContactExtra');
+
+sub join_table {
+    return 'contact_defect_types';
+}
+
+sub map_extras {
+    my ($rs, @ts) = @_;
+    my %meta;
```
davea: Don't think `%meta` is being used any more?

PR #1701, opened 2017-05-15
Larger map for inspectors using 1024px displays

zarino:
Fixes mysociety/fixmystreetforcouncils#190.

The green "inspector" column is narrower now for _all_ users. And for screens smaller than 1280px, the _main_ column also gets narrower, to make sure the map is still usable.

# 1024px, before

1024-before

# 1024px, after

1024-after

# 1260px, before

1260-before

# 1260px, after

1260-after

# 1280px, before

1280-before

# 1280px, after

1280-after

### DIFF
--- a/web/cobrands/oxfordshire/layout.scss
+++ b/web/cobrands/oxfordshire/layout.scss
@@ -59,31 +59,6 @@ body.twothirdswidthpage {
     }
 }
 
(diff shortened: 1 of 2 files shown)


tomoccscholes:
This looks good to me...

tomoccscholes:
This looks much better to me...

PR #1702, opened 2017-05-16
[Vagrant] Update submodule on any provisioning.

dracos:
On an initial clone, the Vagrantfile would check and make sure that the
submodule was present (and usable inside and outside the box), but on a
subsequent provisioning it wasn't making sure it was up to date.


dracos:
Merged as 16b4d05da

PR #1707, opened 2017-05-19
Tidy up state dropdowns, and use instruct permission in Oxfordshire

dracos:
Fixes https://github.com/mysociety/fixmystreetforcouncils/issues/191


dracos:
Whoops; this should be fixed now (plus test tweaked to check).

PR #1711, opened 2017-05-23
Fix linting errors

pezholio:
The Git JS linting hook didn't used to run on files with `OpenLayers` in the filename, but we've changed this now vendor scripts are in their own subdirectory (#1705). This means that our own map-OpenLayers.js file wasn't being linted before, but now it is. This PR just fixes the issues that the linter's picked up.

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -2,7 +2,7 @@ var fixmystreet = fixmystreet || {};
 
 (function() {
 
-    fixmystreet.maps = fixmystreet.maps || {}
+    fixmystreet.maps = fixmystreet.maps || {};
 
     $.extend(fixmystreet.maps, {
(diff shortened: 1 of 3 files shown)


web/js/map-OpenLayers.js
```diff
@@ -121,7 +113,8 @@ var fixmystreet = fixmystreet || {};
         return fixmystreet.markers.getFeaturesByAttribute('id', problem_id)[0];
       },
 
-      marker_size_for_zoom: function(zoom) {
+      marker_size_for_zoom: function() {
```
dracos: This function is overridden in the two wmts* files, so you'll need to change their behaviour too. Or have a shared function looking up the zoom calling different functions.

dracos: Could this function (and related) take an argument, and then default to what you have in none is provided? Then the function names still make sense :)

PR #1713, opened 2017-05-26
[Oxfordshire] Custom map pins with icons

zarino:
**TO-DO:**

- [x] Add problem status to map hover tooltip
- [x] Investigate whether correct icon is being shown next to reports on homepage (is it always just `pin-yellow-small.png`?)

Fixes mysociety/fixmystreetforcouncils#170

In Oxfordshire only:

* Map pins now have icons on them
* The "default" pin colour (for brand new reports and "confirmed", but not yet investigated, reports) is now yellow, rather than red
* The colour for "in progress" reports is now orange, rather than yelow
* The colour for "closed" (but not fixed) reports is now grey, rather than green

For all cobrands:

* Pins are now double resolution, so they'll look good on high-dpi screens
* It’s much easier to define a custom pin image location thanks to a `$pin_path` Sass variable
* It’s now possible to customise the "default" pin colour (for brand new reports and "confirmed", but not yet investigated, reports)
* New (unused) orange pin image and css class, in case any cobrands fancy adopting it in future


web/cobrands/sass/_report_list_pins.scss
```diff
@@ -1,24 +1,29 @@
...
   }
   &.red a {
-      background-image: url(/i/pin-red-small.png);
+      background-image: url('#{$pin_prefix}pin-red-small.png');
   }
   &.grey a {
-      background-image: url(/i/pin-grey-small.png);
+      background-image: url('#{$pin_prefix}pin-grey-small.png');
+  }
+  &.orange a {
```
davea: Should we have an orange pin in the default cobrand, or limit this to the Oxfordshire SCSS?

zarino:
@davea Ok! All pin and pin-shadow images are now double resolution. And there’s an (unused) set of orange pin icons in core, in case any cobrands fancy swapping in the extra colour at some point in the future.

zarino:
Oops, forgot to pngoptimise the images in core. Will do that now, and then merge and deploy.

dracos:
Force-pushed 'refactor pretty state display' because I had a panic this morning about moving states out of the template when I'd suggested that as a method of changing the displayed state names. So now they're back in the template, but just the one template, rendered from the template function.

PR #1721, opened 2017-06-14
Improve test suite (double speed, isolate with transactions) plus remove weird UK body ID 'feature'

dracos:
This PR improves the test suite by isolating each test file in a DB transaction, letting them run in parallel.
It also removes the uses whereby a UK body had to have the same ID as its MapIt area ID.


dracos:
Merged at 0252fc158

Issue #189, opened 2012-01-13
Rename "Include stale reports" link


Deborah-mySoc:
I don't think 'stale' makes much sense to most of our users.

Perhaps call them 'old' instead?

[Note: this change should be done in conjunction with putting this on map as a layer switcher]

Issue #188, opened 2012-01-12
Remember "Show my name publicly" state


dracos:
If the server rejects new report for some reason, the state of the public name checkbox appears to get reset.

PR #1724, opened 2017-06-16
WIP fix blank spaces being clickable.

dracos:
Float all labels left to fix 189 - this currently breaks e.g. Show X about Y. Is there a way to fix it without having to check all the labels.
Use a `<br>` to fix 188, but does this make the gap too big, is there a nicer way?

### DIFF
--- a/templates/web/base/reports/body.html
+++ b/templates/web/base/reports/body.html
@@ -37,13 +37,16 @@
 <div id="map_sidebar">
     <div id="side">
 
-<h1 id="reports_heading">
-    [% IF ward %]
-    [% ward.name %]<span>, </span><a href="[% body_url %]">[% body.name %]</a>
+  [% IF ward %]
+    <h1 id="reports_heading">
+        [% ward.name %]
+    </h1>
+    <a href="[% body_url %]">[% body.name %]</a>
     [% ELSE %]
-    [% body.name %]
-    [% END %]
-</h1>
+    <h1 id="reports_heading">
+        [% body.name %]
+    </h1>
+  [% END %]
 
 [% IF c.cobrand.moniker == 'hart' %]
     [% INCLUDE '_hart_hants_note.html' %]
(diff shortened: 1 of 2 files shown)


davea:
For #188, why not just move the `<a>` outside of the `<h1>`? Is there any reason it needs to be in there, a11y perhaps?

For #189 the simplest change I can see to target that label would be something like
```scss
.inspect-section {
  label {
    display: inline-block;
  }

  p {
    margin-bottom: 0;
  }
}
```

dracos:
Will do for 188. I guess the wider issue with 189 is the same issue affects other forms. I'll try inline-block everywhere. The `p` margin doesn't help there as far as I can see, as it's the label's bottom margin causing the issue, but you want that on non-checkbox inputs. Bit of a mess top/bottom margins on these forms, really.

dracos:
Have gone with the minimum for now, as you suggest. For the wider issue, you can start off by inline-blocking all labels (dunno why I didn't do that instead of floating), which fixes it, and then the main fall out is the extra vertical spacing caused by the removal of margin collapse. However, you can't currently just drop the label top-margin because it sometimes follows from e.g. divs with no bottom margin. There isn't much consistency with whether a label is 'top level' or inside a 'p' or what, or what margins things within a form have. It might be worth trying to fix this more generally, though that might be a bit yucky. I had something that fixed the major cases:

```
// Paragraphs have a 1em bottom margin by default, as do checkbox-group and plain-lists.
// Sometimes the labels are inside a p, sometimes they're directly after
p + label, p + p > label:first-child, .checkbox-group + label, .plain-list li + li > label:first-child {
    margin-top: 0.25em;
}
// Oh and form-box text/email inputs?
.form-box {
  & > input[type=text] + label, & > input[type=email] + label {
    margin-top: 0.25em;
  }
}
// Admin hints just confuse the issue
.admin-hint + p > label:first-child {
    margin-top: 0;
}
// Normal inputs (and the JS photo upload) have 0.5em bottom margin, not 1em
.form-control + label, .dropzone + label {
    margin-top: 0.75em;
}
```
(And having the labels display: block on admin-index-search).

Issue #1630, opened 2017-02-06
Reconsider reports page


MyfanwyNixon:
The marketing manager at one of our client councils has suggested that we either remove, or reword https://www.fixmystreet.com/reports.

One issue is that of course, client councils are going to show higher numbers of reports, giving an uneven view of the number of problems within their boundaries. This leads to bad press from journalists who don't stop to think about it.

One easyish fix might be to separate out client councils from the others, with a bit of explanatory text?

MyfanwyNixon:
Further nudge from said client - could the team have a think about this, so we can come back with  a definite yay or nay, even if the actual work doesn't get prioritised? 

cc @dracos @jukesie

jukesie:
We spoke to the bit of Oxfordshire we are dealing with re FMSfCouncils yesterday and they raised it as well. Going to add something to an upcoming sprint to investigate appropriate solutions.

PR #1726, opened 2017-06-19
All reports dashboard

zarino:
- [x] Static prototype of page styling and chart styling.
- [x] Pull in real data.
- [x] Commas in big numbers
- [x] Plurals
- [x] Translations? (Is this .com only? Opt-in somehow?)
- [x] 'problems updated' includes same problems if they get >1 update; change wording
- [x] Council names, not IDs! Plus sort order is wrong
- [x] Total average time to fix
- [x] Switch to most recent 100 per council, not last period, to cut out low-number report skews
- [x] ~~Make box "postcode or council name"?~~ Dropdown/autocomplete of body names?
- [x] Too slow to run 'live' as-is (c. 14s on live data), will need to move data generation offline or something.

dashboard

Fixes mysociety/fixmystreetforcouncils#173. Fixes mysociety/fixmystreet-commercial#806. Fixes https://github.com/mysociety/fixmystreet/issues/1630.


dracos:
Quick note - the JS should now be in web/vendor, presumably?

zarino:
@dracos – good point, yes, `chart.min.js` and `moment.min.js` should be in `vendor`. (I think this branch pre-dated our introduction of the `vendor` directory, and when I rebased it yesterday, I forgot to move the files.)

dracos:
Should this page have the Pro advert at the bottom like the current reports page does, presumably?

templates/web/fixmystreet.com/reports/index.html
```diff
@@ -0,0 +1,111 @@
...
+        <h2 class="dashboard-subheading">[% loc('Top 5 responsive councils') %]</h2>
+        <p>[% loc('Average time between a problem being reported and being fixed, last 100 reports.') %]</p>
+        <table class="dashboard-ranking-table">
+            <tbody>
+              [% FOR line IN top_five_bodies %]
+                <tr><td>[% line.name %]</td><td>[% tprintf(nget("%s day", "%s days", line.days), line.days) %]</td></tr>
+              [% END %]
+            </tbody>
+            <tfoot>
+                <tr><td>[% loc('UK average') %]</td><td>[% tprintf(nget("%s day", "%s days", average), average) %]</td></tr>
```
davea: Could this be reduced in precision a bit? 😄

screen shot 2017-07-06 at 17 05 14

templates/web/fixmystreet.com/reports/index.html
```diff
@@ -0,0 +1,111 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="dashboard-item dashboard-item--6">
+        <form class="dashboard-search" action="/reports">
+            <h2>[% loc('Show reports in your area') %]</h2>
+            <label for="body">[% loc('Pick your council') %]</label>
+            <div class="dashboard-search__input">
+                <select id="body" name="body" class="js-autocomplete">
```
davea: This dropdown is really keen for me to choose Aberdeen if I don't enter anything:

aberdeen

On iOS (this is iPhone 6 size), there's no way to dismiss the dropdown:

ios dropdown


Also, more generally, this is the only (I think?) place we show a location input field that doesn't accept a postcode. I guess that's because it's a dropdown really, not a text field, but might be a nice-to-have in the future.

templates/web/fixmystreet.com/reports/index.html
```diff
@@ -0,0 +1,111 @@
...
+[% INCLUDE 'header.html', title = loc('Dashboard'), bodyclass => 'dashboard fullwidthpage' %]
+
+<div class="dashboard-header">
+    <h1>[% loc('Dashboard') %]</h1>
+</div>
+
+<div class="dashboard-row">
+    <div class="dashboard-item dashboard-item--12">
+        <h2>UK: [% loc('All time') %]</h2>
+        <div class="labelled-line-chart">
```
davea: On narrow screens this looks a bit squished:

simulator screen shot 6 jul 2017 17 15 35

templates/web/fixmystreet.com/reports/index.html
```diff
@@ -0,0 +1,111 @@
...
+                [% line_count = line.count | format_number ~%]
+                <tr><td>[% line.category %]</td><td>[% tprintf(nget("%s report", "%s reports", line.count), decode(line_count)) %]</td></tr>
+              [% END %]
+            </tbody>
+            <tfoot>
+                <tr><td>[% loc('Other categories') %]</td><td>[% tprintf(nget("%s report", "%s reports", other_categories), decode(other_categories_formatted)) %]</td></tr>
+            </tfoot>
+        </table>
+    </div>
+</div>
```
davea: Are enough elements being closed before the footer? On the front page the FMS Pro banner is outside the white box of the main content, but here it looks a bit broken:

 
screen shot 2017-07-06 at 17 21 32

templates/web/fixmystreet.com/reports/index.html
dracos: Yes, it was going to be postcode originally (in fact, act like front page, plus geocoding + geolocation) but then I panicked about two-tier councils and having to add an interstitial (plus intermediary place selection (or factoring that out at least) like front page) and thought quicker to just go with a body selection for the moment.

templates/web/fixmystreet.com/reports/index.html
dracos: Haha! I must have had boring averages testing. Will do.

templates/web/fixmystreet.com/reports/index.html
dracos: I've turned off `showAllValues` for the moment to fix the iOS thing and the Aberdeen thing.

templates/web/fixmystreet.com/reports/index.html
dracos: Have made sure each Y is at least the previous Y.
WHICH ISN'T RIGHT, IS IT, MUPPET. Grrrrrr

dracos:
Since merge, after seeing on live site:
* Changed main chart labels to only be on right at 48em+, otherwise stay underneath (prevents horizontal scrolling due to numbers not fitting);
* Fixed setting of `lasty` for non-overlapping labels (doesn't really matter due to the above now, but hey);
* Added some top padding to main chart so final dot isn't cut off.

dracos:
e1ccf105f fixed issue where sparkline fixed was ignoring questionnaire responses.

Issue #525, opened 2013-08-20
Make the 'hide pins' and 'hide old' buttons more prominent.


MyfanwyNixon:
We get a couple of user support queries per month that make it clear the user hasn't seen these - those are obvs just the ones who bother to get in touch. Perhaps a different colour would help.

I also think the 'get updates' button is a little too much on the discreet side.

abimysoc:
User email today asked for this feature to be added, hadn't noticed it was there already!

zarino:
Some thoughts on this:

# Current behaviour

<details>
<summary>Click to expand</summary><p>

Problems are represented by a pin on the map and a corresponding list item in the sidebar. 6 months of problems are shown. Sidebar list items are sorted from newest to oldest, by default, although the user can change this.

A small "Show old" button in the bottom right corner of the map lets you reveal problems older than the 6 month cut-off period. And a similar "Hide pins" button lets you hide all pins on the map.

Clicking the map background begins the process of reporting a new problem. Clicking an existing map pin or sidebar list item brings up details of that report.

Benefits:

* There is a chance that users coming to report a problem at a particular location will notice an existing report for that problem – reducing duplicate reports for the councils to have to process.
* Users can get an impression of how prevalent reports are in an area by just eyeballing the map page.

Drawbacks:

* High concentrations of pins make it hard to click the map to report a new problem.
* "Hide pins" button is easily overlooked.

</p></details>

# Solution 1: Paginate

Paginate the problems so that only, say, 20 problems are shown at a time, sorted as they are now (newest-to-oldest, by default). Buttons in the sidebar will let users navigate back and forth through the pages – Google Maps does this when you search for amenities at a location, although the use-case there is obviously very different.

<details>
<summary>Show benefits and drawbacks</summary><p>

Benefits:

* Only 20 reports will be visible in the sidebar and on the map at any one time, meaning more space to click between pins.
* Rendering of pins on the map when you zoom/scroll will likely be faster, since the javascript is no longer having to request and draw hundreds of problems.

Drawbacks:

* Unlikely that people will click to show more than the first 20 reports – increasing the chances of duplicate reporting if their problem isn’t one of the 20 most recently reported in that area.
* Chance that people won’t notice the pagination links, and just assume they are viewing all reports in an area.

</p></details>

# Solution 2: More prominent "Hide pins" button

[...]

PR #1727, opened 2017-06-20
Move #hide_pins_link up near pan/zoom controls

zarino:
An attempt at helping people click the map in very busy areas – #525.

The "Hide pins" button is more prominent now.

hide-pins


templates/web/base/maps/openlayers.html
```diff
@@ -31,6 +31,11 @@
     <div id="map">
       [% INCLUDE 'maps/noscript_map.html' js = 1 %]
     </div>
+  [% IF c.req.params.no_pins %]
```
dracos: I guess it's possible you might want to turn off pins on a `/reports` or `/report` page, but I think we probably only want this on an `/around` page :)

web/cobrands/sass/_base.scss
```diff
@@ -1451,6 +1451,10 @@ html.js #map .noscript {
   top: 44px;
 }
 
+#hide_pins_link {
```
dracos: I think being able to hide pins on the mobile map (even as and when pagination is added) is quite important!

web/cobrands/sass/_layout.scss
```diff
@@ -614,6 +615,34 @@ body.authpage {
   bottom: 2em;
 }
 
+.big-hide-pins-link {
+  display: block;
+  position: absolute;
+  top: 248px; // top of #fms_pan_zoom_zoomout + height of #fms_pan_zoom_zoomout + 22px gap
```
dracos: 174+44+22 = 240, not 248?

web/cobrands/sass/_layout.scss
```diff
@@ -614,6 +615,34 @@ body.authpage {
...
+  right: 38px; // right of #fms_pan_zoom (0.5em) + right of #fms_pan_zoom_panup (30px)
```
dracos: I guess you could put this inside fms_pan_zoom to combat having to do this, but might not be worth it.

web/js/map-OpenLayers.js
```diff
@@ -614,16 +614,19 @@ $.extend(fixmystreet.utils, {
             zoomToBounds( fixmystreet.markers.getDataExtent() );
         }
 
-        $('#hide_pins_link').click(function(e) {
+        $('#hide_pins_link, .big-hide-pins-link').click(function(e) {
```
dracos: Is it worth having a `hide-pins-link` class on both links and then using that here/below? &lt;shrug>

web/cobrands/sass/_layout.scss
zarino: Haha! The extra `8px` is the `top` of `#fms_pan_zoom`, as you probably guessed. I left it out becuase the line was already getting pretty long, and I assumed nobody would care. Maybe I should add it in ;-)

web/cobrands/sass/_layout.scss
zarino: Don’t think it’s worth it. Also, it has nothing to do with panning or zooming :-(

abimysoc:
There's a FAQ entry that's now potentially confusing as a result of this:

>There are too many pins on the map; I can't make a report

> Use the buttons at the bottom right of the screen – 'hide pins', or 'hide old pins', to clear the map. 

Luckily we haven't actually removed the buttons at the bottom right, just added an extra one; but let this act as a reminder to check and if necessary adjust the FAQs when making a user-facing change...

PR #1728, opened 2017-06-21
Generate stats for body areas in all-reports.json

davea:
This is in support of mysociety/fixmystreet-international#50.

This was the simplest way I could see to add this feature, and given the incoming new dashboards I didn't want to go too deep into the problem as it'll likely be redundant soon.

### DIFF
--- a/perllib/FixMyStreet/Script/UpdateAllReports.pm
+++ b/perllib/FixMyStreet/Script/UpdateAllReports.pm
@@ -22,9 +22,12 @@ if ( FixMyStreet->config('BASE_URL') =~ /zurich|zueri/ ) {
 }
 
 sub generate {
+    my $include_areas = shift;
+
     my $problems = FixMyStreet::DB->resultset('Problem')->search(
         {
             state => [ FixMyStreet::DB::Result::Problem->visible_states() ],
+            bodies_str => \'is not null',
         },
         {
             columns => [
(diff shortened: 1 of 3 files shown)


davea:
Hmm, I've just noticed that this swells the `all-reports.json` from a few KB to several MB on fms.com - perhaps it should live behind a general.yml flag so it can be enabled on a per-vhost basis.

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -66,6 +67,21 @@ sub generate {
...
             }
         }
+
+        if ( FixMyStreet->config('AREAS_IN_ALLREPORTS_JSON') ) {
+            @areas = grep { $_ } split( /,/, $problem{areas} );
+            foreach my $area ( @areas ) {
+                my $duration_str = ( $problem{duration} > 2 * $fourweeks ) ? 'old' : 'new';
+                my $type = ( $problem{duration} > 2 * $fourweeks )
+                    ? 'unknown'
+                    : ($problem{age} > $fourweeks ? 'older' : 'new');
```
dracos: These don't change, so don't need to be in a loop (same applies to body loop!).

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -66,6 +67,21 @@ sub generate {
...
+                if (FixMyStreet::DB::Result::Problem->fixed_states()->{$problem{state}} || FixMyStreet::DB::Result::Problem->closed_states()->{$problem{state}}) {
```
dracos: As body loop has this conditional too, worth sticking it in a variable first and using?

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -66,6 +67,21 @@ sub generate {
                 $open{$cobrand}{$body}{$type}++;
             }
         }
+
+        if ( FixMyStreet->config('AREAS_IN_ALLREPORTS_JSON') ) {
+            @areas = grep { $_ } split( /,/, $problem{areas} );
```
dracos: As we're here, could we drop the broken lines 47-49ish which don't work when body ID != mapit area ID (and even if those do equal, the list pages don't include such reports which is a different issue I guess).

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -66,6 +67,21 @@ sub generate {
                 $open{$cobrand}{$body}{$type}++;
             }
         }
+
+        if ( FixMyStreet->config('AREAS_IN_ALLREPORTS_JSON') ) {
```
dracos: If this is temporary (perhaps even if not), rather than a config variable, perhaps a command line argument to the update-all-reports script, saying you wish to include area data?

davea:
Ta for the review, all good points! I've fixed things up in 45e6a0af7644900b62924dac271f4958c924d4ca, how's it look now?

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -40,23 +42,22 @@ sub generate {
...
+        my $duration_str = ( $problem{duration} > 2 * $fourweeks ) ? 'old' : 'new';
+        my $type = ( $problem{duration} > 2 * $fourweeks )
+            ? 'unknown'
+            : ($problem{age} > $fourweeks ? 'older' : 'new');
+        my $problem_fixed =
+               FixMyStreet::DB::Result::Problem->fixed_states()->{$problem{state}}
+            || FixMyStreet::DB::Result::Problem->closed_states()->{$problem{state}};
+
+        # Add to bodies it was sent to
+        @bodies = split( /,/, $problem{bodies_str} );
```
dracos: This will issue a warning on any problem not sent to any body, needs a `next unless $problem{bodies_str}` before it or similar? Or the DB query could exclude those with null entries.

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -40,23 +42,22 @@ sub generate {
...
+        $problem{bodies} = scalar @bodies;
```
dracos: I don't think this is used anywhere?

PR #1731, opened 2017-06-23
Allow $ENV overrides for hostname/port

davea:
Accessing the dev server via a service such as ngrok/vagrant share is made
difficult because some redirects (e.g. to /auth) and form POSTS (e.g. front
page) include the absolute URL including the BASE_URL hostname and port.
This commit allows 3 environment variables to be set which will be used in
absolute URL redirects.

### DIFF
--- a/perllib/Catalyst/Engine.pm
+++ b/perllib/Catalyst/Engine.pm
@@ -418,9 +418,9 @@ sub prepare_path {
 
     my $env = $ctx->request->env;
 
-    my $scheme    = $ctx->request->secure ? 'https' : 'http';
-    my $host      = $env->{HTTP_HOST} || $env->{SERVER_NAME};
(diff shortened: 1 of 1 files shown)


dracos:
One concern here is this is an unmodified copy of a Catalyst file, so might be lost in an upgrade/replacement. Can this be done at the FixMyStreet::App level, I assume it must be possible somehow (a similar way to how secure is set in the presence of a proxy?).

The host line doesn't look like it's needed, it defaults to whatever has been used in the URL? Or have I misunderstood this.

If I remember it right, this won't fix e.g. map page links because they use BASE_URL directly, nor links in emails. Is there a reason you can't alter BASE_URL? I can see that being more annoying if they change frequently, though. Perhaps tie in/add to Zarino's PR with a config switching script...

PR #1734, opened 2017-06-28
Resize photos client-side before uploading

### LINKED ISSUE #1411: Clear metadata from uploaded pictures
If metadata is not removed from uploaded pictures (and I'm not sure whether this is the case), we might accidentally reveal information about the uploading user (who may want to be anonymous on the web site).

davea:
The newest version of Dropzone has added support for client-side resizing of
images, so this commit upgrades to version 5.1.1 and takes advantage of that in
order to reduce the disc space consumed by uploaded images.

The target size & JPEG quality is quite aggressive, resulting in files in the
region of 200-300KB instead of several MB. I've tried to pick a value which
doesn't impact the usefulness of the photos in identifying problems.

Handily the new version also correctly rotates images according to the EXIF tag
so our patch has be reduced dramatically to deal with the sync nature of the
mini EXIF library in use.

As a consequence of the resizing, the EXIF metadata is mostly stripped out so
this should fix #1411.


web/js/dropzone.js.patch
```diff
@@ -1,36 +1,13 @@
...
- 
+--- dropzone.5.1.1.js	2017-06-28 17:33:36.000000000 +0100
++++ dropzone.exiffixes.js	2017-06-28 17:40:40.000000000 +0100
+@@ -1175,9 +1175,7 @@
+           };
+           if ((typeof EXIF !== "undefined" && EXIF !== null) && fixOrientation) {
+             loadExif = function(callback) {
+-              return EXIF.getData(img, function() {
+-                return callback(EXIF.getTag(this, 'Orientation'));
+-              });
```
dracos: If we altered src/exif.js to, I think, add `EXIF.getTag = function(img, tag) { return EXIF.result; };`, changed `EXIF.getData` to be a `function(img, callback)` and changed its return data to be `EXIF.result = readEXIFData(dataView, offset + 4); callback();` would that mean we could drop the dropzone patch entirely, apart from the Windows Phone thing. So meh, worth it if we could have dropped the whole thing.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -503,6 +503,8 @@ $.extend(fixmystreet.set_up, {
         addRemoveLinks: true,
         thumbnailHeight: 150,
         thumbnailWidth: 150,
+        resizeWidth: 1600,
```
dracos: I'm fine with this, but for some reason I have a niggling desire for it to be more like 2048 (3mp) than 1600. Can't really justify the feeling, how much difference do you think that would make to file size?

davea: Files still seem to come out around 400-500KB so 2048 it is 👍

web/js/dropzone.js.patch
davea: Yep, with the blacklist restored we still need a patch so might as well keep this bit in and not touch `exif.js`.

PR #1735, opened 2017-06-28
[FixaMinGata] Consolidate with upstream

jonkri:
This merge request aims to consolidate the FixaMinGata-specific code with the upstream project.

The changes are:

- FixaMinGata e-mail templates, including a logo image
- An updated problem form for FixaMinGata
- Updated Swedish translation strings
- Correction of a FixaMinGata-related map pin problem


templates/web/fixamingata/report/new/form_user.html
```diff
@@ -0,0 +1,15 @@
+<h2 class="form-section-heading form-section-heading--private">Andra uppgifter</h2>
```
dracos: Noting this template is (I believe) included because this string was missed in translation until 2.0.2, so could then be removed with an upgrade.

templates/web/fixamingata/report/new/form_report.html
```diff
@@ -0,0 +1,73 @@
+[% INCLUDE 'report/new/form_heading.html' %]
+
+[% IF field_errors.bodies %]
+    <p class='form-error'>[% field_errors.bodies %]</p>
+[% END %]
+
+[% PROCESS "report/new/category_wrapper.html" %]
+[% TRY %][% PROCESS 'report/new/after_category.html' %][% CATCH file %][% END %]
+
+    <h2 class="form-section-heading">Offentliga uppgifter</h2>
```
dracos: This template exists because of this missed string (also 2.0.2), plus extra tips underneath the summary. So introducing a template hook (after merging this, does not have to hold this up) for those tips would mean this template wasn't needed either.

templates/web/fixamingata/report/new/form_user.html
jonkri: Awesome! :)

dracos:
Do you know what "Correction of a FixaMinGata-related map pin problem" was referring to? I can't see anything about that in the changed files.

Anyway, I added the hook and moved the template, added the translations so the other template isn't needed. Hope that's okay, let me know if it doesn't work for you, but am merging now :)

jonkri:
Hi, @dracos!

Thanks a lot - I will! :)

I have now added the translated strings to the pull request.

By "Correction of a FixaMinGata-related map pin problem", I was referring to the "Hide pins" button not working on FixaMinGata &ndash; something which these two lines seem to have solved:

- https://github.com/mysociety/fixmystreet/pull/1735/commits/d653ca757489a453e343cdd8097dc2765f997f1e#diff-3e894f72d0411416e62f0e9ba9963182R574
- https://github.com/mysociety/fixmystreet/pull/1735/commits/d653ca757489a453e343cdd8097dc2765f997f1e#diff-920dce232ca1ae55f9028dd52d299e6bR222

Please let me know if there's anything else that I can do!

jonkri:
@dracos: I discovered that two strings in the questionnaire template were not translatable, so I added two <code>loc</code> calls to the file. Does it look OK?

jonkri:
By the way, I noticed that the "Fixed" and "Not fixed" strings in the email template are both pointing to the same URL. Also, I feel that it's a bit inconsistent that there is a "Don't know" alternative in the questionnaire web page, but not in the e-mail.

Should the template perhaps be updated to simply say something "neutral" instead, such as "Leave update"?

dracos:
Oh great, thanks for spotting those missing translations, sorry about that.

The reason the email template is like that, I think, is to perhaps encourage people to start the process, plus it should be relatively straightforward to have it preselect the right option once they click something, I guess that just hasn't been done as yet. There could perhaps be a "don't know" option in the email, yeah, I guess it was thought if you don't know you might not bother replying – @zarino any extra thoughts? Of course your template can differ if you like :)

zarino:
A "Don’t know" option could easily be added, but in FMS.com’s case, at least, I’m not sure it’d give us any useful information. We’re really only interested in hearing back from a user if they _know_ the problem is either fixed or unfixed. Or maybe I'm missing something?

jonkri:
I see.

It just seems a bit odd to me semantically somehow that there is a "Don't know" option on the page when there's no such option in the e-mail. How would a "Don't know" answering user even reach the questionnaire as things are now? Other than that, I don't think that we have a particular opinion on the "Don't know" option.

Perhaps we should create a separate issue for perhaps having one or three buttons in the e-mail, and if we have multiple actions, make them preselect the appropriate radio button on the page?

Thanks!

dracos:
@jonkri Two issues, that sounds like - certainly create an issue for preselecting, I thought there already was one, it definitely should do that. And yes, do do that, I've just thought, they need to opt in to receive another questionnaire so we do want don't know people to fill it in, not just ignore the email. So I think I wouldn't mind a Don't know being added to the email.

dracos:
I've rebased and squashed this (and done --no-wrap on the po file to match what we have), and added one more missing string spotted. Thanks! Will merge now.

dracos:
PS Have you seen https://www.fixmystreet.com/reports - planning to make this the default if that's okay!

jonkri:
@dracos:

Thanks for merging and helping us out with the changes!

About the new report page: Nice - I like it! I have informed the rest of the FixaMinGata team about the change. I'll let you know if they have any particular thoughts on it. :)

jonkri:
The new report page works great, and we will be upgrading to it on our production site soon. Thanks again!

PR #1736, opened 2017-06-28
Remove open311-update-reports

jonkri:
I'm pleased to announce that FixaMinGata has finally moved off open311-update-reports. I'm assuming that this means that we can finally remove the script from FixMyStreet. :)


dracos:
I guess we should probably leave these for any future uses that also cannot/do not use the update extension; I don't see any harm in doing so, anyway, I see I previously did add a comment to the top of the script, though this could be explained further/expanded in some documentation.

dracos:
Yeah, going to leave them :)

PR #1737, opened 2017-06-29
fix imbalanced paragraph tags in glossary

davewhiteland:
Spotted a rude `</dl>` showing up at the end of [the glossary](http://fixmystreet.org/glossary/) (and there was a `</p>` higher up too, which turned out to be the real culprit) — I believe* this fixes it.

fms-glossary-fail

 * Locally, Jekyll rendered the offending page cleanly, so I didn't manage to reproduce the error locally before modifying the markdown.


dracos:
Thank you, merged, seems to have worked :)

PR #1741, opened 2017-06-30
bin/switch-site script for easily switching configs

zarino:
Given a set of config files of the format `conf/general-{cobrand}.yml` this script lets you easily create and recreate a symlink at `conf/general.yml` that points to the cobrand config of your choice.

I’ve been using this for years; never thought to commit it! Thanks to @davea for the reminder.


zarino:
Given #1740 I wonder whether maybe we should move this new script into the `scripts` directory, for consistency?

dracos:
Don't really mind that, most scripts are in bin (e.g. fixture); does it want conf/general-*.yml adding to gitignore (and testing changed to use same format?)

PR #1746, opened 2017-07-04
Add stretch (etc) support

dracos:
This updates carton/cpanm to deal with e.g. `.` not being in `@INC` in Perl 5.24, so that modules will still install okay. It removes modules from the snapshot that were only there for squeeze support (dropped a long time ago), fixes some stretch installation issues, and tests perl 5.24 on Travis (which is still running precise).


cpanfile
```diff
@@ -11,6 +11,7 @@ requires 'ExtUtils::MakeMaker', '6.72'; # [1]
 requires 'Guard', '1.023';
 requires 'PadWalker', '2.2';
 requires 'aliased', '0.34';
+requires 'Net::SSLeay', '1.81';
```
davea: Woo, this seems to install cleanly on macOS! Now seems a good opportunity to upgrade `Net::Twitter::Lite::WithAPIv1_1` to 0.12008, which no longer depends on `Crypt::SSLeay` (the final macOS pain point) 😄

dracos: Your macOS, it's always installed fine on mine :P Sounds good, will stick that in.

dracos: I've added a commit to upgrade that and SOAP::Lite, which then drops Crypt::SSLeay entirely. Not sure the SOAP::Lite using things are well tested (Zurich geocoder, Angus/EastHants report submission) so might need careful deploy though nothing major in changelog I can see.

dracos:
@davea Do f89681f and 01b426f look okay to you?

davea:
Yep, they look fine to me! 👍

PR #1748, opened 2017-07-06
[Oxfordshire] Group RDI output by activity code.

dracos:


dracos:
Was just needing an `if $_[0]` adding :)

PR #1754, opened 2017-07-07
Allow static home page template override.

### LINKED ISSUE #1745: Have way of having an `/about` page used as front page
Cobrands would like the ability to have a static page as the front page, with the reporting page shown elsewhere. They don't want to have to override the index.html template and then recreate that somewhere else.

dracos:
If you have a templates/web/COBRAND/about/homepage.html file, then that page will be used as the front page of your site, and the normal front page will instead be available at `/report`. Fixes #1745 

- [x] Tests

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -28,14 +28,11 @@ Redirect to homepage unless C<id> parameter in query, in which case redirect to
 sub index : Path('') : Args(0) {
     my ( $self, $c ) = @_;
 
-    my $id = $c->get_param('id');
-
-    my $uri =
-        $id
-      ? $c->uri_for( '/report', $id )
-      : $c->uri_for('/');
-
-    $c->res->redirect($uri);
+    if ($c->stash->{homepage_template}) {
+        $c->stash->{template} = 'index.html';
+    } else {
+        $c->res->redirect('/');
+    }
 }
 
 =head2 report_display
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -28,14 +28,14 @@ Redirect to homepage unless C<id> parameter in query, in which case redirect to
...
 
-    my $id = $c->get_param('id');
-
-    my $uri =
-        $id
-      ? $c->uri_for( '/report', $id )
-      : $c->uri_for('/');
-
-    $c->res->redirect($uri);
+    if (my $id = $c->get_param('id')) {
```
davea: How old is this functionality? Is it still in use anywhere or could we ruthlessly chop it out?

templates/web/base/main_nav_items.html
```diff
@@ -1,4 +1,4 @@
-[%~ INCLUDE navitem uri='/' label=loc('Report a problem') attrs='class="report-a-problem-btn"' ~%]
+[%~ INCLUDE navitem uri=(homepage_uri OR '/') label=loc('Report a problem') attrs='class="report-a-problem-btn"' ~%]
```
davea: `homepage_uri` isn't always set, so e.g. on `/auth` the 'Report a problem' link is to `/`, not `/report`. This seems to be a bug, unless I'm misunderstanding something perhaps.

perllib/FixMyStreet/App/Controller/Report.pm
dracos: Looks like .com forces `/report` to redirect to `/reports` anyway, for typos, so we've never used it ourselves.
I don't see it being used anywhere, `/around` does it for e.g. front page form, so yeah, could strip it I assume.

templates/web/base/main_nav_items.html
dracos: No, quite correct! Hmm, guess it'll have to look it up on each request, then.

PR #1757, opened 2017-07-08
Add an 'inactive' state to categories

dracos:
This lets you mark a category so that it isn't used for new reports but can still be filtered/etc around the site (as opposed to deleted, which then isn't available for filtering etc either). Probably needs some tests, and not sure if there's some case I haven't thought of, but think it should work.

mysociety/fixmystreetforcouncils#192


templates/web/base/admin/contact-form.html
```diff
@@ -41,21 +41,19 @@
...
-  <div class="admin-hint">
     <p>
       [% loc("Check <strong>deleted</strong> to remove the category from use. 
         It will not appear as an available category in the drop-down menu on the report-a-problem page.") %]
     </p>
   </div>
   <p>
-    <input type="checkbox" name="deleted" value="1" id="deleted"[% ' checked' IF contact.deleted %]>
-    <label for="deleted" class="inline">[% loc('Deleted') %]</label>
+    <label for="state">[% loc('State') %]</label>
```
davea: The tooltip wording doesn't quite match the UI any more, could this be updated to describe the new 'inactive' behaviour?

PR #1760, opened 2017-07-14
[East Herts] Update template design.

dracos:


web/cobrands/eastherts/_colours.scss
```diff
@@ -29,6 +25,8 @@ $col_click_map_dark: darken($eh_green, 10%);
 $col_fixed_label: #00BD08;
 $col_fixed_label_dark: #4B8304;
 
-$body-font: Verdana,Arial,Helvetica,Geneva,sans-serif;
+$body-font: 'Open Sans', sans-serif;
```
davea: I can't see where this font is being loaded, and things are appearing as `sans-serif` here.

dracos: Whoops, forgot to commit template!

dracos:
Have given our menu their 'secondary' background colour, near enough.

PR #1767, opened 2017-07-20
Remove Bower and Grunt from fixmystreet.org build process

zarino:
I came to edit the site for the first time in a while (ever?) today, and noticed the build process could be significantly simpler for newcomers. So, here goes. We now:

* Let Jekyll compile the Sass files, rather than Grunt.
* Manage `github-pages` dependency with a Gemfile, rather than doing it manually.
* Remove Grunt entirely – The only downside is that our JavaScript files are no longer concatenated, which doesn’t feel like a big deal in this particular case.
* Remove Bower – With Grunt gone, we were installing npm and Bower just to manage 3 JavaScript libraries. We don't generally use Bower at mySociety, so removing it here brings this site more in line with our other products.

I also fixed 3 deprecation warnings we were getting from using an outdated Sass colour manipulation feature.


_config.yml
```diff
@@ -12,3 +12,8 @@ authors:
     display_name: Matthew
   zarino:
     display_name: Zarino
+
+# https://github.com/RobertDeRose/jekyll-livereload/
+# Enable livereload by default when running `jekyll serve` locally.
+# (Doesn't do anything on Github Pages.)
+livereload: true
```
dracos: jekyll's incremental is sadly also rubbish and buggy. To work around this, we need to add https://github.com/dracos/web-design-standards-docs/blob/cea68ab020748776030bec991386f4a8d9ac3e60/_plugins/incremental_regeneration_fixer.rb under `_plugins` and add to _config.yml:

```
incremental_regeneration_fixer:
  interdependent_files:
    - "{_posts,blog}/*"
```
I know.

PR #1768, opened 2017-07-20
Move dead sites into their own section

zarino:
Simple change. ZeroTB and Bellavista En Ación moved down to the bottom of the page, under an "Honourable mentions" heading. Broken links removed.

### DIFF
--- a/sites.md
+++ b/sites.md
@@ -24,18 +24,9 @@ Launched: 2014
 <br>
 Run by: <a href="http://sinarproject.org/">Sinar project</a>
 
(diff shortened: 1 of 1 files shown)


MyfanwyNixon:
Looks good - do we know the status of www.fixmystreet.org.au? That's not loading at all for me.

dracos:
@MyfanwyNixon .au just loaded for me okay here, if slowly.

MyfanwyNixon:
@dracos Yep, it's now fine for me.

PR #1772, opened 2017-07-25
Resend report if changing to category with different send_method

davea:
If a report’s category is changed to one with a different send_method to the original category, the report needs to be resent.

Fixes mysociety/fixmystreetforcouncils#209

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -10,6 +10,8 @@ use Digest::SHA qw(sha1_hex);
 use mySociety::EmailUtil qw(is_valid_email is_valid_email_list);
 use DateTime::Format::Strptime;
 use List::Util 'first';
+use List::MoreUtils 'uniq';
+use mySociety::ArrayUtils;
 
 use FixMyStreet::SendReport;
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -847,6 +849,21 @@ sub report_edit_category : Private {
         if (grep !$old_map{$_}, @new_body_ids) {
             $problem->whensent(undef);
         }
+        # If the send methods of the old/new contacts differ we need to resend the report
+        my @old_contacts = grep { $_->category eq $category_old } @{$c->stash->{contacts}};
+        my @new_send_methods = uniq map {
```
struan: Why is this uniq map and the old contacts plain map?

t/app/controller/report_inspect.t
```diff
@@ -247,6 +249,18 @@ FixMyStreet::override_config {
...
+        # Then change the category to the other category within the same council,
+        # which should cause it to be resent because it has a different send method
+        $mech->submit_form(button => 'save', with_fields => { category => 'Sheep', include_update => undef, });
+
+        $report3->discard_changes;
+        is $report3->category, "Sheep", "Report in correct category";
+        is $report3->whensent, undef, "Report marked as unsent";
+        is $report3->bodies_str, $oxon->id, "Reported to OCC";
+    };
+
```
struan: Unless it's elsewhere and I'm missing it it'd be nice to see a test that shows changing to a category with the same send method doesn't resend.

perllib/FixMyStreet/App/Controller/Admin.pm
struan: Ah no, I see, it's because there can only really be one old_contact in there.

davea:
Good shout on the extra test, I've added that in 4eda22305fec20749b540afa7f28174ffd46fad6.

PR #1779, opened 2017-08-01
Make map image in HTML emails clickable

### LINKED ISSUE #1596: Map within report to council should click through to report page
... well, that's how I felt, anyway! If I were trying to ascertain the precise position from a report like this:

fireshot screen capture 973 - fwd_ problem report_ street light out - fixmystreetsupport mysociety_org - mysociety mail - mail_google_com_mail_b_77

 I think I'd naturally click on the map.  And if I did that, the best result would be to be taken to the report page, where I can zoom in, pan around etc.

zarino:
Emails that contain a report’s title, description, and location in the ”sidebar” will now have a map that takes you to the report’s web page when clicked.

Fixes #1596. Thanks to @MyfanwyNixon for the suggestion!

### DIFF
--- a/templates/email/default/_email_sidebar.html
+++ b/templates/email/default/_email_sidebar.html
@@ -1,6 +1,19 @@
-[% DEFAULT report = object ~%]
+[%
+
+# There are cases (eg: when sending email about an update on a problem)
+# where `object` might not be a report. So in those cases, you can pass
+# in a `report` argument, which should be the report the email is about.
+# Otherwise, we assume the `object` you passed is a report.
+DEFAULT report = object;
+
+# There are cases (eg: when confirming a not-yet-published report) where
+# you will want the map image to link to somewhere other than the report
+# page. So in those cases, you can pass in a custom `url` argument.
+DEFAULT url = cobrand.base_url_for_report(report) _ report.url
+
+~%]
 <th style="[% td_style %][% secondary_column_style %]" id="secondary_column">
-  <img style="[% map_image_style %]" src="[% inline_image(report.static_map, 'map.jpeg') %]" width="310" height="200" alt="">
+  <a href="[% url %]"><img style="[% map_image_style %]" src="[% inline_image(report.static_map, 'map.jpeg') %]" width="310" height="200" alt=""></a>
   [% start_padded_box %]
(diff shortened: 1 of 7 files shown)


templates/email/default/_email_sidebar.html
```diff
@@ -1,6 +1,19 @@
-[% DEFAULT report = object ~%]
+[%
+
+# There are cases (eg: when sending email about an update on a problem)
+# where `object` might not be a report. So in those cases, you can pass
+# in a `report` argument, which should be the report the email is about.
+# Otherwise, we assume the `object` you passed is a report.
+DEFAULT report = object ~%]
+
+# There are cases (eg: when confirming a not-yet-published report) where
```
dracos: You have ended a `[% … %]` block just above, so all this is output in the email.

templates/email/default/problem-confirm-not-sending.html
```diff
@@ -22,7 +22,7 @@ <h1 style="[% h1_style %]">Please confirm your&nbsp;report</h1>
   <p style="[% p_style %]">If you no longer wish to publish this report, please take no further action.</p>
   [% end_padded_box %]
 </th>
-[% WRAPPER '_email_sidebar.html' object = report %]
+[% WRAPPER '_email_sidebar.html' object = report, url = token_url %]
```
dracos: Just to note you don't need a comma here for multiple variables, though it doesn't seem to mind them.

templates/email/default/_email_sidebar.html
zarino: Oops! Yes!

templates/email/default/problem-confirm-not-sending.html
zarino: Yeah, in some places we use commas, in others we use just spaces, and in others we use linebreaks. Shall I standardise on spaces then?

dracos: TBH the comma does make it clearer, doesn't it (unless it was `object=report url=token_url` I guess). So feel free to leave it, I don't mind.

zarino: That was my thinking. I'll keep the commas.

PR #1782, opened 2017-08-02
Site-wide extra fields on report form

### LINKED ISSUE #1743: Add admin interface to report_form_extras/template display
You can currently have a cobrand `report_form_extras` function to store extra fields to your reporting form (this only deals with accepting and storing them in the report metadata, you would still have to edit your HTML template). Add an admin interface to manage this feature so that you can add/remove extra fields and have them be displayed during creation automatically.

#1444 referred to documenting the coding way of doing this.

davea:
This PR adds a new feature that allows management via the admin of extra fields on the new report form. These fields can be managed on a per-category basis (i.e. the same as Open311 metadata fields), or site-wide (with restrictions via cobrand/language).

New site-wide fields can be added:
adding_extra_fields

Which then appear on the report form:
fields_on_form

Multiple sets of fields can be active, and limited to cobrand/language:
screen shot 2017-08-02 at 14 37 04

Metadata can be managed per-category, too:
open311_metadata

Fixes #1743

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
     - New features:
         - Body and category names can now be translated in the admin. #1244
         - Body users can now create reports as an anonymous user. #1796
+        - Extra fields can be added to report form site-wide. #1743
     - Front end improvements:
         - Always show pagination figures even if only one page.
     - Admin improvements:
         - Highlight current shortlisted user in list tooltip.
+        - Extra fields on contacts can be edited. #1743
     - Bugfixes:
         - Set up action scheduled field when report loaded. #1789
         - Stop errors from JS validator due to form in form.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
         - Improve performance of duplicate display.
     - Development improvements:
         - `switch-site` script to automate switching config.yml files. #1741
+        - `make_css --watch` can run custom script after each compilation.
 
 * v2.1.1 (3rd August 2017)
     - Email improvements:
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -375,6 +375,8 @@ sub update_contacts : Private {
             $contact->set_extra_metadata( reputation_threshold => int($c->get_param('reputation_threshold')) );
         }
 
+        $c->forward('update_extra_fields', [ $contact ]);
+
         if ( %errors ) {
             $c->stash->{updated} = _('Please correct the errors below');
             $c->stash->{contact} = $contact;
@@ -1971,6 +1973,46 @@ sub fetch_body_areas : Private {
     $c->stash->{areas} = [ sort { strcoll($a->{name}, $b->{name}) } values %$areas ];
     # Keep track of the areas we've fetched to prevent a duplicate fetch later on
     $c->stash->{fetched_areas_body_id} = $body->id;
+}
+
+sub update_extra_fields : Private {
+    my ($self, $c, $object) = @_;
+
+    my @indices = grep { /^metadata\[\d+\]\.code/ } keys %{ $c->req->params };
+    @indices = sort map { /(\d+)/ } @indices;
+
+    my @extra_fields;
+    foreach my $i (@indices) {
+        my $meta = {};
+        $meta->{code} = $c->get_param("metadata[$i].code");
+        next unless $meta->{code};
+        $meta->{order} = int $c->get_param("metadata[$i].order");
+        $meta->{datatype} = $c->get_param("metadata[$i].datatype");
+        my $required = $c->get_param("metadata[$i].required") && $c->get_param("metadata[$i].required") eq 'on';
+        $meta->{required} = $required ? 'true' : 'false';
+        my $notice = $c->get_param("metadata[$i].notice") && $c->get_param("metadata[$i].notice") eq 'on';
+        $meta->{variable} = $notice ? 'false' : 'true';
+        $meta->{description} = $c->get_param("metadata[$i].description");
+        $meta->{datatype_description} = $c->get_param("metadata[$i].datatype_description");
+
+        if ( $meta->{datatype} eq "singlevaluelist" ) {
+            $meta->{values} = [];
+            my $re = qr{^metadata\[$i\]\.values\[\d+\]\.key};
+            my @vindices = grep { /$re/ } keys %{ $c->req->params };
+            @vindices = sort map { /values\[(\d+)\]/ } @vindices;
+            foreach my $j (@vindices) {
+                my $name = $c->get_param("metadata[$i].values[$j].name");
+                my $key = $c->get_param("metadata[$i].values[$j].key");
+                push(@{$meta->{values}}, {
+                    name => $name,
+                    key => $key,
+                }) if $name;
+            }
+        }
+        push @extra_fields, $meta;
+    }
+    @extra_fields = sort { $a->{order} <=> $b->{order} } @extra_fields;
+    $object->set_extra_fields(@extra_fields);
 }
 
 sub trim {
--- a/perllib/FixMyStreet/App/Controller/Admin/ReportExtraFields.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/ReportExtraFields.pm
@@ -1,0 +1,61 @@
+package FixMyStreet::App::Controller::Admin::ReportExtraFields;
+use Moose;
+use namespace::autoclean;
+use List::MoreUtils qw(uniq);
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+
+sub begin : Private {
+    my ( $self, $c ) = @_;
+
+    $c->forward('/admin/begin');
+}
+
+sub index : Path : Args(0) {
+    my ( $self, $c ) = @_;
+
+    my @extras = $c->model('DB::ReportExtraFields')->search(
+        undef,
+        {
+            order_by => 'name'
+        }
+    );
+
+    $c->stash->{extra_fields} = \@extras;
+}
+
+sub edit : Path : Args(1) {
+    my ( $self, $c, $extra_id ) = @_;
+
+    my $extra;
+    if ( $extra_id eq 'new' ) {
+        $extra = $c->model('DB::ReportExtraFields')->new({});
+    } else {
+        $extra = $c->model('DB::ReportExtraFields')->find( $extra_id )
+            or $c->detach( '/page_error_404_not_found' );
+    }
+
+    if ($c->req->method eq 'POST') {
+        $c->forward('/auth/check_csrf_token');
+
+        foreach (qw/name cobrand language/) {
+            $extra->$_($c->get_param($_));
+        }
+        $c->forward('/admin/update_extra_fields', [ $extra ]);
+
+        $extra->update_or_insert;
+    }
+
+    $c->forward('/auth/get_csrf_token');
+    $c->forward('/admin/fetch_languages');
+
+    my @cobrands = uniq sort map { $_->{moniker} } FixMyStreet::Cobrand->available_cobrand_classes;
+    $c->stash->{cobrands} = \@cobrands;
+
+    $c->stash->{extra} = $extra;
+}
+
+__PACKAGE__->meta->make_immutable;
+
+1;
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -99,6 +99,7 @@ sub report_new : Path : Args(0) {
     # create a problem from the submitted details
     $c->stash->{template} = "report/new/fill_in_details.html";
     $c->forward('setup_categories_and_bodies');
+    $c->forward('setup_report_extra_fields');
     $c->forward('generate_map');
     $c->forward('check_for_category');
 
@@ -138,6 +139,7 @@ sub report_new_ajax : Path('mobile') : Args(0) {
     }
 
     $c->forward('setup_categories_and_bodies');
+    $c->forward('setup_report_extra_fields');
     $c->forward('process_user');
     $c->forward('process_report');
     $c->forward('/photo/process_photo');
@@ -185,6 +187,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
     }
 
     $c->forward('setup_categories_and_bodies');
+    $c->forward('setup_report_extra_fields');
 
     # render templates to get the html
     my $category = $c->render_fragment( 'report/new/category.html');
@@ -235,6 +238,7 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
         return 1;
     }
     $c->forward('setup_categories_and_bodies');
+    $c->forward('setup_report_extra_fields');
     $c->forward('check_for_category');
 
     my $category = $c->stash->{category} || "";
@@ -252,6 +256,9 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
         $generate = 1;
     }
     if ($c->stash->{unresponsive}->{$category}) {
+        $generate = 1;
+    }
+    if ($c->stash->{report_extra_fields}) {
         $generate = 1;
     }
     if ($generate) {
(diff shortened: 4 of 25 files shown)


web/cobrands/fixmystreet/admin.js
```diff
@@ -123,5 +123,93 @@ $(function(){
             }
         }
     });
+
+    // Bits for the report extra fields form builder:
+
+    // Some elements' visibility depends on JS being available
+    $(".js-metadata-items .js-hidden").hide();
```
dracos: We already have a `.hidden-js` class that will do this (and without a flash of content as it is in the CSS) :)

web/cobrands/fixmystreet/admin.js
```diff
@@ -123,5 +123,93 @@ $(function(){
...
+    $(".js-metadata-items .js-visible").show();
```
dracos: Similarly we have a `.hidden-nojs` class that will do this.

web/cobrands/fixmystreet/admin.js
```diff
@@ -123,5 +123,93 @@ $(function(){
...
+    // call immediately to perform page setup
+    $(".js-metadata-item-type").change();
+
+    // Options can be removed by clicking the 'remove' button
+    $(".js-metadata-items").on("click", ".js-metadata-option-remove", function(e) {
+        e.preventDefault();
+        var $this = $(this);
+        var $item = $this.closest(".js-metadata-item");
+        $this.closest('li').remove();
+        renumber_metadata_options($item);
```
dracos: Whilst the renumbering works okay, would it not be easier to do nothing upon removal, and on adding a new field set that to the highest index + 1? No need for the data-field-name stuff too then. The server only needs to read them in in order, they don't actually need to be consecutive, if I'm reading it right.

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -0,0 +1,81 @@
...
+                <option value="singlevaluelist" [% meta.datatype == 'singlevaluelist' ? 'selected' : '' %]>[% loc('List') %]</option>
+            </select>
+        </label>
+
+        <div class="js-metadata-options">
+            <div class="admin-hint"><p>[% loc('For each option, <strong>Key</strong> is the value which is stored in the database for that option and <strong>Name</strong> is the value displayed to the user.') %]</p></div>
+            [% loc('Options<span class="js-hidden"> (ignored if type is "String")</span>') %]
+            <ul>
+                [% outer_loop = loop %]
+                [% values = meta.exists('values') ? meta.values : [] %]
```
dracos: I wonder if `meta.values OR []` would be bette here, in case values exists but is an empty string (though I guess that couldn't/shouldn't happen?)

templates/web/base/admin/contact-form.html
```diff
@@ -135,4 +135,13 @@
...
     <input type="submit" class="btn" name="Create category" value="[% contact.in_storage ? loc('Save changes') : loc('Create category') %]" >
   </p>
+
+    <h2>[% loc('Extra data:') %] </h2>
+    <dl>
+      [% FOR pair IN contact.get_extra_metadata %]
+        <dt>[% pair.key %]</dt> <dd>[% pair.value %]</dd>
+      [% END %]
+    </dl>
+    [% INCLUDE 'admin/extra-metadata-form.html' metas=contact.get_metadata_for_input %]
```
dracos: This needs to be `(contact.get_metadata_for_input OR [])` here for on a body page contact is undefined, this is an empty string, and merge creates a list of two entries.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1856,6 +1858,43 @@ sub fetch_body_areas : Private {
...
+
+    my @indices = grep { /^metadata\[\d+\]\.code/ } keys %{ $c->req->params };
+    @indices = sort map { $_ =~ /(\d+)/ } @indices;
+
+    $object->set_extra_fields(());
+    foreach my $i (@indices) {
+        my $meta = {};
+        $meta->{code} = $c->get_param("metadata[$i].code");
+        next unless $meta->{code};
+        $meta->{order} = int $c->get_param("metadata[$i].order");
```
dracos: Order isn't actually used anywhere after this. It looks like PopulateSeviceList sorts by the order upon import and then stores them in this order. So I guess this should do the same.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1856,6 +1858,43 @@ sub fetch_body_areas : Private {
...
+                my $name = $c->get_param("metadata[$i].values[$j].name");
+                my $key = $c->get_param("metadata[$i].values[$j].key");
+                push(@{$meta->{values}}, {
+                    name => $name,
+                    key => $key,
+                }) if $name;
+            }
+        }
+
+        $object->push_extra_fields(($meta));
```
dracos: Don't need double brackets here (or above when emptying).

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -0,0 +1,81 @@
...
+            <input name="metadata[[% loop.index %]].variable" data-field-name="variable"  type=checkbox [% meta.variable == 'true' ? 'checked' : '' %]>
+        </label>
+
+        <div class="admin-hint"><p>[% loc('The field name as shown to the user on the report form.') %]</p></div>
+        <label>
+            [% loc('Description') %]
+            <input name="metadata[[% loop.index %]].description" data-field-name="description" type=text value="[% meta.description | html %]">
+        </label>
+
+        <div class="admin-hint"><p>[% loc('Some extra guidance text shown to the user alongside the field. The default template does not show this (<code>meta.datatype_description</code>), you must add it in <code>category_extras_fields.html</code>') %]</p></div>
```
dracos: It sounds like datatype_description is for describing the type of data - e.g. it's date input or something? Anyway, as you say, not used so guess people can use it whatever they want. Open311s we've seen do put things like "A string" in this field.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1856,6 +1858,43 @@ sub fetch_body_areas : Private {
...
+        $meta->{order} = int $c->get_param("metadata[$i].order");
+        $meta->{datatype} = $c->get_param("metadata[$i].datatype");
+        $meta->{required} = $c->get_param("metadata[$i].required") eq 'on' ? 'true' : 'false';
+        $meta->{variable} = $c->get_param("metadata[$i].variable") eq 'on' ? 'true' : 'false';
+        $meta->{description} = $c->get_param("metadata[$i].description");
+        $meta->{datatype_description} = $c->get_param("metadata[$i].datatype_description");
+
+        if ( $meta->{datatype} eq "singlevaluelist" ) {
+            $meta->{values} = [];
+            my $re = "^metadata\\[$i\\]\\.values\\[\\d+\\]\\.key";
```
dracos: If you wrap $re in `qr{}` so it's compiled immediately, then you can `s/\\\\/\\/` and make this look slightly nicer ;-)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1856,6 +1858,43 @@ sub fetch_body_areas : Private {
...
+            my @vindices = grep { /$re/ } keys %{ $c->req->params };
+            @vindices = sort map { $_ =~ /values\[(\d+)\]/ } @vindices;
```
dracos: $_ not needed

perllib/FixMyStreet/App/Controller/Admin/ReportExtraFields.pm
```diff
@@ -0,0 +1,67 @@
...
+
+    my @cobrands = uniq sort map { $_->{moniker} } FixMyStreet::Cobrand->available_cobrand_classes;
+    $c->stash->{cobrands} = \@cobrands;
+
+    my @languages;
+    foreach my $language (@{$c->cobrand->languages}) {
+        my ($prefix, $name, $code) = split(/,/, $language);
+        push @languages, [$prefix, $name];
+    }
+    $c->stash->{languages} = \@languages;
```
dracos: See Struan's PR for an admin function that does basically this if we want to make it easier to share the two PRs.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -940,6 +954,24 @@ sub set_report_extras : Private {
         }
     }
 
+    my @extra_fields = $c->model('DB::ReportExtraFields')->for_cobrand($c->cobrand)->for_language($c->stash->{lang_code})->all;
```
dracos: Can this call setup_report_extra_fields or has that already been done and the stash can be used?

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -0,0 +1,81 @@
...
+            [% loc('Type') %]
+            <select name="metadata[[% loop.index %]].datatype" data-field-name="datatype" class="js-metadata-item-type">
+                <option value="string" [% meta.datatype == 'string' ? 'selected' : '' %]>[% loc('String') %]</option>
+                <option value="singlevaluelist" [% meta.datatype == 'singlevaluelist' ? 'selected' : '' %]>[% loc('List') %]</option>
+            </select>
+        </label>
+
+        <div class="js-metadata-options">
+            <div class="admin-hint"><p>[% loc('For each option, <strong>Key</strong> is the value which is stored in the database for that option and <strong>Name</strong> is the value displayed to the user.') %]</p></div>
+            [% loc('Options<span class="js-hidden"> (ignored if type is "String")</span>') %]
```
dracos: This should be two strings (added bonus in that the HTML isn't then included).

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -0,0 +1,81 @@
...
+                <option value="singlevaluelist" [% meta.datatype == 'singlevaluelist' ? 'selected' : '' %]>[% loc('List') %]</option>
+            </select>
+        </label>
+
+        <div class="js-metadata-options">
+            <div class="admin-hint"><p>[% loc('For each option, <strong>Key</strong> is the value which is stored in the database for that option and <strong>Name</strong> is the value displayed to the user.') %]</p></div>
+            [% loc('Options') %]<span class="js-hidden"> [% loc('(ignored if type is "String")') %]</span>
+            <ul>
+                [% outer_loop = loop %]
+                [% values = meta.exists('values') ? meta.values : [] %]
```
dracos: I wonder if zmeta.values OR []` would be better here, in case values exists but is an empty string (though I guess that couldn't/shouldn't happen?)

templates/web/base/admin/reportextrafields/edit.html
```diff
@@ -0,0 +1,68 @@
...
+            [% loc('To limit this collection of fields to a single cobrand, select it here.') %]
+        </p>
+        </div>
+        <p>
+            <label>
+                [% loc('Cobrand') %]
+                <select name="cobrand">
+                    <option value="">[% loc('All cobrands') %]</option>
+                [% FOREACH cobrand IN cobrands %]
+                <option value="[% cobrand | html %]" [% IF cobrand == extra.cobrand %]selected[% END %]>[% cobrand %]</option>
```
dracos: Second cobrand missing `| html`

templates/web/base/admin/reportextrafields/edit.html
```diff
@@ -0,0 +1,68 @@
...
+            [% loc('To limit this collection of fields to a single language, select it here.') %]
+        </p>
+        </div>
+        <p>
+            <label>
+                [% loc('Language') %]
+                <select name="language">
+                    <option value="">[% loc('All languages') %]</option>
+                    [% FOREACH language IN languages %]
+                        <option value="[% language.0 | html %]" [% IF language.0 == extra.language %]selected[% END %]>[% language.1 %]</option>
```
dracos: language.1 missing `| html` (not sure if needed, but can't hurt).

perllib/FixMyStreet/App/Controller/Admin.pm
davea: Top tip, thanks!

web/cobrands/fixmystreet/admin.js
davea: Good point, you're right that it would be simpler to just increment the maximum. However the `data-field-name` stuff is still required cos you still need to update the names of fields in the newly added item.

dracos: Oh, I thought you could have the "template" name fields be e.g. `metadata[N].order` and then your loop would `find("input[name^=metadata]")` (or `.js-metadata-item input`) and do a `$input.attr("name", $input.attr("name").replace("N", n);` or similar.

dracos: But I'm not wedded on that! You can do the same for the js-metadata-option-add/remove too I guess :)

dracos:
Just to note I've done rereview of what's there and all looks good 🏅 Todo as far as I can see (you probably already have this list):
- [ ] Tests
- [ ] Should we also at this point change admin/report_edit.html to at least output any stored extra data rather than just yes/no?
- [ ] Increment max/do nothing on remove for js-metadata-option-add/remove (possibly more to remove data-extra-fields, not if not worth it)
- [ ] master (now this is rebased) has an admin function `fetch_languages` that could be used in ReportExtraFields.pm

web/cobrands/fixmystreet/admin.js
davea: I did it this way so non-JS editing still works, am happy with the current approach but will rework if it's a blocker on merging 😃

davea:
> Should we also at this point change admin/report_edit.html to at least output any stored extra data rather than just yes/no?

Sounds like a different ticket to me. Have opened #1809.

I think this is ready for re-review now.

templates/web/base/admin/reportextrafields/edit.html
```diff
@@ -0,0 +1,68 @@
...
+                <option value="[% cobrand | html %]" [% IF cobrand == extra.cobrand %]selected[% END %]>[% cobrand | html %]</option>
+                [% END %]
+                </select>
+            </label>
+        </p>
+    [% ELSE %]
+        <input type=hidden name=cobrand value="[% extra.cobrand | html %]" />
+    [% END %]
+
+    [% IF languages.values.size > 1 %]
```
dracos: No issue, just to note that `languages.size` would work on a hash and returns the same answer as `languages.values.size`.

CHANGELOG.md
```diff
@@ -4,17 +4,20 @@
     - New features:
         - Body and category names can now be translated in the admin. #1244
         - Body users can now create reports as an anonymous user. #1796
+        - Extra fields can be added to report form site-wide #1743
```
dracos: Full stops at the end, just to be consistent with the rest of the file? :)

t/app/controller/admin_reportextrafields.t
```diff
@@ -0,0 +1,311 @@
...
+}, sub {
+    subtest 'Language-specific fields are missing from /report/new for other language' => sub {
+        $mech->get_ok("/report/new?longitude=-1.351488&latitude=51.847235&category=" . $contact->category);
+        $mech->content_lacks("this is a test description");
+        $mech->content_lacks("this field is a list");
+    };
+};
+
+FixMyStreet::override_config {
+    ALLOWED_COBRANDS => [ { 'default' => '.' } ],
```
dracos: As the default is 0, this is testing the same thing as the next test, really. I guess there should be a cobrand that is set to 1 but with a different name to actually test this.

t/app/controller/admin_reportextrafields.t
```diff
@@ -0,0 +1,311 @@
...
+my $body = $mech->create_body_ok(2237, 'Oxfordshire County Council');
+my $contact = $mech->create_contact_ok( body_id => $body->id, category => 'Potholes', email => 'potholes@example.com' );
+
+my $body2 = $mech->create_body_ok(2651, 'Edinburgh City Council');
+my $contact2 = $mech->create_contact_ok( body_id => $body2->id, category => 'Potholes', email => 'potholes@example.com' );
+
+
+FixMyStreet::override_config {
+    ALLOWED_COBRANDS => [ { 'tester' => '.' } ],
+    MAPIT_TYPES => [ 'UTA' ],
```
dracos: No need for this override (and couple similar ones elsewhere) because your Tester cobrand is manually setting it (`MAPIT_TYPES` is only used by th default `area_types` function).

t/app/controller/admin_reportextrafields.t
```diff
@@ -0,0 +1,311 @@
...
+                title => "Test Report",
+                detail => "This is a test report",
+                category => $contact2->category,
+                "extra[$extra_id]string_test" => "Problem meta string",
+                "extra[$extra_id]list_test" => "key1",
+            }
+        } );
+        # my $content = $mech->content;
+        # $content =~ s/[^[:ascii:]]+//g;
+        # print $content;
```
dracos: Few debug lines left in. (PS Try `utf8::encode($content)` for the middle line there :) )

web/cobrands/fixmystreet/admin.js
dracos: No, it's fine, but not sure I was clear with what I said - does this `js-metadata-option-remove` option need to call renumber_metadata_options at all, or is calling remove enough like the `metadata-item-remove`? That's only thing (and presumably doesn't matter if it is left in).

PR #1783, opened 2017-08-02
[Ground Control] Platform improvements

davea:
This PR is a collection of small improvements to FMS that came about as part of the Ground Control site.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Root.pm
+++ b/perllib/FixMyStreet/App/Controller/Root.pm
@@ -16,6 +16,18 @@ FixMyStreet::App::Controller::Root - Root Controller for FixMyStreet::App
 
 =head1 METHODS
 
+=head2 begin
+
+Any pre-flight checking for all requests
+
+=cut
+sub begin : Private {
+    my ( $self, $c ) = @_;
+
+    $c->forward( 'check_login_required' );
+}
+
+
 =head2 auto
 
 Set up general things for this instance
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -130,6 +142,27 @@ sub page_error : Private {
...
+    return if $c->user_exists || !FixMyStreet->config('LOGIN_REQUIRED');
+
+    my $whitelist = [
+        '^auth$',
+        '^auth/',
+        '^js/translation_strings\.[a-zA-Z]{2}(-[a-zA-Z]{2})?\.js',
+        '^[PACQM]{1}/', # various tokens that log the user in
+    ];
+
+    foreach (@$whitelist) {
```
dracos: No reason for this to be an array reference, plus you could join the entries with `|` and do one regex? In fact, with `x` you could write it out as a multiline regex, think that's as clear, e.g.
```
my $whitelist = qr{
    ^auth(/|$)
  | ^js/translation_strings\.(.*?)\.js
  | ^[PACQM]/  # various tokens that log the user in
}x;
return if $c->request->path =~ $whitelist;
```

bin/make_css
```diff
@@ -35,6 +35,8 @@ GetOptions(
     'style|t=s' => sub { die "Unknown style\n" unless $_[1] =~ /nested|expanded|compact|compressed/; $style = $_[1]; },
     'verbose' => \my $verbose,
     'watch' => \my $watch,
+    'script=s'=> \my $watch_script,
+    'script-param=s'=> \my $watch_script_param,
```
dracos: If you put this in an array, you could pass multiple parameters (dunno if that's needed/wanted!)

perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -147,19 +147,33 @@ sub check_login_required : Private {
...
+        '^offline',
+    ];
+
     foreach (@$whitelist) {
         if ( $c->request->path =~ m#$_# ) {
             return;
         }
     }
 
+    foreach (@$blacklist) {
```
dracos: Same as with the whitelist.

perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -130,6 +142,27 @@ sub page_error : Private {
...
+sub check_login_required : Private {
+    my ($self, $c) = @_;
+
+    return if $c->user_exists || !FixMyStreet->config('LOGIN_REQUIRED');
+
+    my $whitelist = [
+        '^auth$',
+        '^auth/',
+        '^js/translation_strings\.[a-zA-Z]{2}(-[a-zA-Z]{2})?\.js',
+        '^[PACQM]{1}/', # various tokens that log the user in
```
davea: Duh :)

davea:
I think I've covered off all the points you raised, so this is ready for a re-review.

dracos:
Nice :)

PR #1784, opened 2017-08-02
Translatable body and category names

struan:
This adds a table to store translations for database strings such as category and body names, and then uses that to display translated body and category names. It also adds an interface in the admin to edit these.

screen shot 2017-08-04 at 21 21 01
screen shot 2017-08-04 at 21 21 42

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -244,6 +244,9 @@ sub bodies : Path('bodies') : Args(0) {
 
     $c->stash->{edit_activity} = $edit_activity;
 
+    $c->forward( 'fetch_languages' );
+    $c->forward( 'fetch_translations' );
+
     my $posted = $c->get_param('posted') || '';
     if ( $posted eq 'body' ) {
         $c->forward('check_for_super_user');
@@ -257,6 +260,9 @@ sub bodies : Path('bodies') : Args(0) {
                 $c->model('DB::BodyArea')->create( { body => $body, area_id => $_ } );
             }
 
+            $c->stash->{object} = $body;
+            $c->stash->{translation_col} = 'name';
+            $c->forward('update_translations');
             $c->stash->{updated} = _('New body added');
         }
     }
@@ -298,30 +304,6 @@ sub body_form_dropdowns : Private {
 
     my @methods = map { $_ =~ s/FixMyStreet::SendReport:://; $_ } keys %{ FixMyStreet::SendReport->get_senders };
     $c->stash->{send_methods} = \@methods;
-}
-
-sub body : Path('body') : Args(1) {
-    my ( $self, $c, $body_id ) = @_;
-
-    $c->stash->{body_id} = $body_id;
-
-    unless ($c->user->has_permission_to('category_edit', $body_id)) {
-        $c->forward('check_for_super_user');
-    }
-
-    $c->forward( '/auth/get_csrf_token' );
-    $c->forward( 'lookup_body' );
-    $c->forward( 'fetch_all_bodies' );
-    $c->forward( 'body_form_dropdowns' );
-
-    if ( $c->get_param('posted') ) {
-        $c->log->debug( 'posted' );
-        $c->forward('update_contacts');
-    }
-
-    $c->forward('fetch_contacts');
-
-    return 1;
 }
 
 sub check_for_super_user : Private {
@@ -407,6 +389,12 @@ sub update_contacts : Private {
             $contact->insert;
         }
 
+        unless ( %errors ) {
+            $c->stash->{translation_col} = 'category';
+            $c->stash->{object} = $contact;
+            $c->forward('update_translations');
+        }
+
     } elsif ( $posted eq 'update' ) {
         $c->forward('/auth/check_csrf_token');
 
@@ -446,9 +434,43 @@ sub update_contacts : Private {
             # Remove any others
             $c->stash->{body}->body_areas->search( { area_id => [ keys %current ] } )->delete;
 
+            $c->stash->{translation_col} = 'name';
+            $c->stash->{object} = $c->stash->{body};
+            $c->forward('update_translations');
+
             $c->stash->{updated} = _('Values updated');
         }
     }
+}
+
+sub update_translations : Private {
+    my ( $self, $c ) = @_;
+
+    foreach my $lang (keys(%{$c->stash->{languages}})) {
+        my $id = $c->get_param('translation_id_' . $lang);
+        my $text = $c->get_param('translation_' . $lang);
+        if ($id) {
+            my $translation = $c->model('DB::Translation')->find(
+                {
+                    id => $id,
+                }
+            );
+
+            if ($text) {
+                $translation->msgstr($text);
+                $translation->update;
+            } else {
+                $translation->delete;
+            }
+        } elsif ($text) {
+            my $col = $c->stash->{translation_col};
+            $c->stash->{object}->add_translation_for(
+                $col, $lang, $text
+            );
+        }
+    }
+
+    $c->stash->{updated} = _('Translations updated');
 }
 
 sub body_params : Private {
@@ -497,6 +519,44 @@ sub fetch_contacts : Private {
     return 1;
 }
 
+sub fetch_languages : Private {
+    my ( $self, $c ) = @_;
+
+    my $lang_map = {};
+    foreach my $lang (sort @{$c->cobrand->languages}) {
+        my ($id, $name, $code) = split(',', $lang);
+        $lang_map->{$id} = { name => $name, code => $code };
+    }
+
+    $c->stash->{languages} = $lang_map;
+
+    return 1;
+}
+
+sub fetch_translations : Private {
+    my ( $self, $c ) = @_;
+
+    my $translations = {};
+    if ($c->get_param('posted')) {
+        foreach my $lang (keys %{$c->stash->{languages}}) {
+            if (my $msgstr = $c->get_param('translation_' . $lang)) {
+                $translations->{$lang} = { msgstr => $msgstr };
+            }
+            if (my $id = $c->get_param('translation_id_' . $lang)) {
+                $translations->{$lang}->{id} = $id;
+            }
+        }
+    } elsif ($c->stash->{object}) {
+        my @translations = $c->stash->{object}->translation_for($c->stash->{translation_col})->all;
+
+        foreach my $tx (@translations) {
+            $translations->{$tx->lang} = { id => $tx->id, msgstr => $tx->msgstr };
+        }
+    }
+
+    $c->stash->{translations} = $translations;
+}
+
 sub lookup_body : Private {
     my ( $self, $c ) = @_;
 
@@ -516,35 +576,94 @@ sub lookup_body : Private {
     return 1;
 }
 
+sub body_base : Chained('/') : PathPart('admin/body') : CaptureArgs(0) { }
+
 # This is for if the category name contains a '/'
-sub category_edit_all : Path('body') {
+sub category_edit_all : Chained('body_base') : PathPart('') {
     my ( $self, $c, $body_id, @category ) = @_;
     my $category = join( '/', @category );
-    $c->go( 'category_edit', [ $body_id, $category ] );
-}
-
-sub category_edit : Path('body') : Args(2) {
-    my ( $self, $c, $body_id, $category ) = @_;
 
     $c->stash->{body_id} = $body_id;
+    $c->forward( 'lookup_body' );
+
+    my $contact = $c->stash->{body}->contacts->search( { category => $category } )->first;
+    $c->stash->{contact} = $contact;
+
+    $c->stash->{template} = 'admin/category_edit.html';
+    $c->forward( 'category_edit' );
+}
+
+sub body : Chained('body_base') : PathPart('') : CaptureArgs(1) {
+    my ( $self, $c, $body_id ) = @_;
+    $c->stash->{body_id} = $body_id;
+}
+
+sub edit_body : Chained('body') : PathPart('') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    unless ($c->user->has_permission_to('category_edit', $c->stash->{body_id})) {
+        $c->forward('check_for_super_user');
+    }
+
+    $c->forward( '/auth/get_csrf_token' );
+    $c->forward( 'lookup_body' );
+    $c->forward( 'fetch_all_bodies' );
+    $c->forward( 'body_form_dropdowns' );
+    $c->forward('fetch_languages');
+
+    if ( $c->get_param('posted') ) {
+        $c->forward('update_contacts');
+    }
+
+    $c->stash->{object} = $c->stash->{body};
+    $c->stash->{translation_col} = 'name';
+
+    # if there's a contact then it's because we're displaying error
+    # messages about adding a contact so grabbing translations will
+    # fetch the contact submitted translations. So grab them, stash
+    # them and then clear posted so we can fetch the body translations
+    if ($c->stash->{contact}) {
+        $c->forward('fetch_translations');
+        $c->stash->{contact_translations} = $c->stash->{translations};
+    }
+    $c->set_param('posted', '');
+
+    $c->forward('fetch_translations');
+    $c->forward('fetch_contacts');
+
+    $c->stash->{template} = 'admin/body.html';
+    return 1;
+}
+
+sub category : Chained('body') : PathPart('') : CaptureArgs(1) {
+    my ( $self, $c, $category ) = @_;
 
     $c->forward( '/auth/get_csrf_token' );
     $c->forward( 'lookup_body' );
 
     my $contact = $c->stash->{body}->contacts->search( { category => $category } )->first;
     $c->stash->{contact} = $contact;
+}
+
+sub category_edit : Chained('category') : PathPart('') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    $c->stash->{translation_col} = 'category';
+    $c->stash->{object} = $c->stash->{contact};
+
+    $c->forward('fetch_languages');
+    $c->forward('fetch_translations');
 
     my $history = $c->model('DB::ContactsHistory')->search(
         {
-            body_id => $body_id,
-            category => $category
+            body_id => $c->stash->{body_id},
+            category => $c->stash->{contact}->category
         },
         {
             order_by => ['contacts_history_id']
         },
     );
     $c->stash->{history} = $history;
-
     my @methods = map { $_ =~ s/FixMyStreet::SendReport:://; $_ } keys %{ FixMyStreet::SendReport->get_senders };
     $c->stash->{send_methods} = \@methods;
 
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -925,8 +925,8 @@ sub categories_for_point : Private {
     # Remove the "Pick a category" option
     shift @{$c->stash->{category_options}} if @{$c->stash->{category_options}};
 
-    $c->stash->{categories} = $c->stash->{category_options};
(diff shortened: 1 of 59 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -213,7 +213,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
             category        => $category,
             extra_name_info => $extra_name_info,
             titles_list     => $extra_titles_list,
-            categories      => $c->stash->{category_options},
+            categories      => [ map { $_->{name} } @{$c->stash->{category_options}} ], # XXX NEEDED??
```
dracos: This XXX, is this line needed/used by anything? Was added for mobile apps back in 8e8f83e22823e3cde2403589b5344911ee2a2633 but I can't see anything using it.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -306,6 +306,8 @@ sub body_check : Private {
     # We must now have a string to check
     my @bodies = $c->model('DB::Body')->search( { name => { -like => "$q_body%" } } )->all;
 
+    # XXX Needs to check translation also (or only have available at 'default' name)
```
dracos: This is still to be done, I think.

t/app/controller/admin_translations.t
```diff
@@ -0,0 +1,147 @@
...
+my $mech = FixMyStreet::TestMech->new;
+
+my $superuser = $mech->create_user_ok('superuser@example.com', name => 'Super User', is_superuser => 1);
+
+$mech->log_in_ok( $superuser->email );
+
+FixMyStreet::override_config {
+    MAPIT_URL => 'http://mapit.uk/',
+    MAPIT_TYPES => [ 'UTA' ],
+    BASE_URL => 'http://www.example.org',
```
dracos: Don't think BASE_URL (and MAPIT_TYPES?) needs to be overridden here.

t/roles/translatable.t
```diff
@@ -0,0 +1,64 @@
...
+is $contact->category_display, "Schlaglöcher";
+is $problem->category_display, "Schlaglöcher";
+
+FixMyStreet::override_config {
+    ALLOWED_COBRANDS => [ 'fiksgatami' ],
+}, sub {
+    $mech->get_ok($problem->url);
+	$mech->content_contains('Hull i veien');
+
+    # XXX TODO Any displayed problem category needs to use category_display
```
dracos: This comment, I think has been done?

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
+[% IF languages.size > 1 %]
+<h2>Translations</h2>
+<table>
+  <tr>
+    <th>Language</th>
+    <th>Translation</th>
+  </tr>
+  [% FOREACH translation IN translations %]
+  <tr>
+    <td><a href="[% action _ '/translation/' _ translation.lang %]">[% translation.lang %]</a></td>
```
dracos: Could this show the full language name, not just the code?

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
+[% IF languages.size > 1 %]
+<h2>Translations</h2>
```
dracos: All strings in templates need to be wrapped in `[% loc(…) %]` so they can be translated.

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
...
+<table>
+  <tr>
+    <th>Language</th>
+    <th>Translation</th>
+  </tr>
+  [% FOREACH translation IN translations %]
+  <tr>
+    <td><a href="[% action _ '/translation/' _ translation.lang %]">[% translation.lang %]</a></td>
+    <td>[% translation.msgstr %]</td>
```
dracos: I realise this probably won't be done often, but it would be nice if translations could be edited from this table, rather than having to go to a separate page?

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
...
+  [% END %]
+  [% IF translations.size == 0 %]
+  <tr>
+      <td colspan="2">No translations</td>
+  </tr>
+  [% END %]
+</table>
+
+<h2>Add Translation</h2>
+<form method="post" action="[% action %]" enctype="application/x-www-form-urlencoded" accept-charset="utf-8" id="category_edit">
```
dracos: Is the ID or enctype here needed?

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
...
+    <p>
+    <label for="language">[% loc('Language') %]</label>
+    <select id="language" name="language" required>
+      [% FOREACH lang IN languages %]
+        <option value="[% lang.id %]">[% lang.name %]</option>
+      [% END %]
+    </select>
+    </p>
+    <p>
+      <strong>[% loc('Translation') %] </strong><input type="text" class="form-control" name="translation" size="30" required>
```
dracos: This should be a label.

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,39 @@
...
+    </p>
+
+    <input type="submit" class="btn" name="Add translation" value="Add translation" >
```
dracos: I don't think this needs a name.

templates/web/base/admin/body.html
```diff
@@ -142,4 +142,6 @@
 </div>
 [% END %][%# Only show all the above if no errors with category form %]
 
+[% INCLUDE 'admin/_translations.html' action=c.uri_for('body', body_id) %]
```
dracos: Should this be contained within the edit body details box? It sort of feels like it's part of editing the body details after all. But I realise that it's a separate form. Should existing translations be editable in the one form along with the other body details? And put up with the main "name" field?

templates/web/base/admin/body_translation_edit.html
```diff
@@ -0,0 +1,31 @@
+[% INCLUDE 'admin/header.html' title=tprintf(loc('Edit %s Translation for %s'), language_name, body.name ) -%]
+
+<form method="post" action="[% c.uri_for('body', body_id ) %]" enctype="application/x-www-form-urlencoded" accept-charset="utf-8" id="category_edit">
+    <input type="hidden" name="translation_id" value="[% translation.id| html %]" >
+    <input type="hidden" name="token" value="[% csrf_token %]" >
+    <input type="hidden" name="posted" value="edit_translation" >
+    <input type="hidden" name="language" value="[% translation.lang| html %]" >
+    <p>
+    <strong>Original</strong>: [% body.name | html %]
```
dracos: I don't think this line is needed as the body name is given in the title just above.

Although I have just realised that if you add a translation in the same language you are viewing the admin site in, then the main body.name in "Edit body" will of course change to be that (which I guess doesn't matter, but is a bit confusing). Hmm.

templates/web/base/admin/body_translation_edit.html
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    <p>
+    <strong>Language</strong>: [% language_name | html %] ([% translation.lang | html %])
+    <p>
+      <strong>[% loc('Translation') %] </strong><input type="text" class="form-control" name="translation" size="30" value="[% translation.msgstr| html %]" required>
```
dracos: Should be a label.

templates/web/base/admin/body_translation_edit.html
```diff
@@ -0,0 +1,31 @@
...
+    </p>
+
+    <input type="submit" class="btn" name="Create category" value="Update translation">
```
dracos: Doesn't need a name.

templates/web/base/admin/body_translation_edit.html
```diff
@@ -0,0 +1,31 @@
...
+</form>
+
+<p></p>
```
dracos: This doesn't do anything.

templates/web/base/admin/category_translation_edit.html
```diff
@@ -0,0 +1,31 @@
+[% INCLUDE 'admin/header.html' title=tprintf(loc('Edit %s Translation for %s'), language_name, contact.category) -%]
+
+<form method="post" action="[% c.uri_for('body', body_id, contact.category ) %]" enctype="application/x-www-form-urlencoded" accept-charset="utf-8" id="category_edit">
+    <input type="hidden" name="translation_id" value="[% translation.id| html %]" >
+    <input type="hidden" name="token" value="[% csrf_token %]" >
+    <input type="hidden" name="posted" value="edit_translation" >
+    <input type="hidden" name="language" value="[% translation.lang| html %]" >
+    <p>
+    <strong>Original</strong>: [% contact.category | html %]
```
dracos: Same comments for here (is it worth sharing these two templates, they seem very similar).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -449,6 +425,138 @@ sub update_contacts : Private {
...
+                object_id => $body_id,
+                col => 'name',
+                lang => $lang,
+            }
+        );
+
+        $translation->msgstr($text);
+
+        if ( $translation->in_storage ) {
+            $c->stash->{updated} = _('Translation added');
```
dracos: Should this be "Translation updated"?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -449,6 +425,138 @@ sub update_contacts : Private {
...
+            $translation->update;
+        } else {
+            $c->stash->{updated} = _('New body translation added');
```
dracos: Just "New translation added" so less strings to translate?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -449,6 +425,138 @@ sub update_contacts : Private {
...
+}
+
+sub update_category_translations : Private {
+    my ( $self, $c ) = @_;
+
+    my $posted = $c->get_param('posted');
+    my $editor = $c->forward('get_user');
+
+    if ( $posted eq 'translation' ) {
+        $c->forward('check_for_super_user');
```
dracos: All this code seems shared with the body update.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,41 +636,165 @@ sub lookup_body : Private {
...
+        my ($id, $name, $code) = split(',', $supported_lang);
+        $language_name = $name if $id eq $lang;
+    }
+
+    $c->stash->{language_name} = $language_name;
+
+    return 1;
+}
+
+sub category : Chained('body_base') : PathPart('') : CaptureArgs(2) {
```
dracos: I like all the chained stuff. Should this be Chained off body rather than body_base?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,41 +636,165 @@ sub lookup_body : Private {
...
+
     my $contact = $c->stash->{body}->contacts->search( { category => $category } )->first;
     $c->stash->{contact} = $contact;
+}
+
+sub category_edit : Chained('category') : PathPart('') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    if ( $c->get_param('posted') ) {
+        $c->log->debug( 'posted' );
```
dracos: Debug line

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,41 +636,165 @@ sub lookup_body : Private {
...
         {
             order_by => ['contacts_history_id']
         },
     );
     $c->stash->{history} = $history;
 
+    my @translations = $c->model('DB::Translation')->search(
+        {
+            tbl => 'contact',
+            object_id => $c->stash->{contact}->id,
```
dracos: Do we want a helper method in Result/Contact.pm that does this lookup? And same for Body.pm (line 679)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,41 +636,165 @@ sub lookup_body : Private {
...
     my @methods = map { $_ =~ s/FixMyStreet::SendReport:://; $_ } keys %{ FixMyStreet::SendReport->get_senders };
     $c->stash->{send_methods} = \@methods;
 
     return 1;
 }
 
+sub category_translation_edit : Chained('category') : PathPart('translation') : Args(1) {
+    my ( $self, $c, $lang ) = @_;
+
+    my $translation = $c->model('DB::Translation')->find(
```
dracos: Ditto here, could actually be same helper method in fact (if lang/no lang provided).
Perhaps auto-provided by the Role?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,41 +636,165 @@ sub lookup_body : Private {
...
+            object_id => $c->stash->{contact}->id,
+            col => 'category',
+            lang => $lang
+        }
+    );
+
+    $c->stash->{translation} = $translation;
+
+    my $language_name = '';
+    foreach my $supported_lang (@{$c->cobrand->languages}) {
```
dracos: This again seems to overlap with the body code doing the same.

templates/web/base/admin/_translations.html
struan: done.

templates/web/base/admin/_translations.html
struan: done.

templates/web/base/admin/_translations.html
struan: done

templates/web/base/admin/_translations.html
struan: done

templates/web/base/admin/_translations.html
struan: done

templates/web/base/admin/body_translation_edit.html
struan: I wanted to make it totally obvious what you were translating. It's a bit redundant but having the thing you are translating next to the translation and also not embedded in a title seemed a bit clearer.

perllib/FixMyStreet/Roles/Translatable.pm
```diff
@@ -0,0 +1,101 @@
...
+        object_id => $self->id,
+        col => $col,
+        lang => $lang,
+    };
+
+    if ($lang) {
+        $props->{lang} = $lang;
+    }
+
+    my $translation = $schema->resultset('Translation')->find_or_new($props);
```
dracos: I think here down to the return can be replaced with:
```
$props->{msgstr} = $msgstr; # Actually, put this in the setup above
my $translation = $schema->resultset('Translation')->update_or_create($props, { key => 'translation_tbl_object_id_col_lang_key' });
```
Think that does the same, anyway, worth a check :)

templates/web/base/admin/body.html
dracos: Now they both submit to the same place, I think this is definitely worth doing – combine them as one form (so they appear on the new body form too), update the translations if posted is `new` or `update`, and then you can have the translations appear alongside the column they're translating.

templates/web/base/admin/_translations.html
```diff
@@ -0,0 +1,28 @@
...
+    <td>
+        [% languages.$language.name %] ([% language %])
+        <input type="hidden" name="translation_id_[% language %]" value="[% translations.$language.id %]">
+    </td>
+    <td><input type="text" name="translation_[% language %]" value="[% translations.$language.msgstr %]"></td>
+  </tr>
+  [% END %]
+  [% IF translations.size == 0 %]
+  <tr>
+      <td colspan="2">[% loc('No translations') %]</td>
```
dracos: Is this not implicit because all the text boxes will be blank.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -495,6 +510,33 @@ sub fetch_contacts : Private {
     return 1;
 }
 
+sub fetch_languages : Private {
+    my ( $self, $c ) = @_;
+
+    my $lang_map = {};
+    foreach my $lang (@{$c->cobrand->languages}) {
```
dracos: Could we have a sort of some sort, each refresh of admin page is changing my order :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -444,11 +432,52 @@ sub update_contacts : Private {
...
         }
     }
 }
 
+sub update_translations : Private {
+    my ( $self, $c ) = @_;
+
+    my $posted = $c->get_param('posted');
+
+    if ( $posted eq 'new' or $posted eq 'body' or $posted eq 'update') {
```
dracos: I think this has to be true by the time you get here now :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -495,6 +524,44 @@ sub fetch_contacts : Private {
...
+        foreach my $lang (keys %{$c->stash->{languages}}) {
+            if (my $msgstr = $c->get_param('translation_' . $lang)) {
+                $translations->{$lang} = { msgstr => $msgstr };
+            }
+            if (my $id = $c->get_param('translation_id_' . $lang)) {
+                $translations->{$lang}->{id} = $id;
+            }
+        }
+    } elsif ($c->stash->{object}) {
+        my @translations = $c->stash->{object}->translation_for($c->stash->{translation_col})->all;
```
dracos: object won't be defined here on a GET to /bodies (ie. for the new body form).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,35 +581,84 @@ sub lookup_body : Private {
...
+    $c->forward('fetch_languages');
+
+    if ( $c->get_param('posted') ) {
+        $c->forward('update_contacts');
+    }
+
+    $c->stash->{object} = $c->stash->{body};
+    $c->stash->{translation_col} = 'name';
+
+    $c->forward('fetch_translations');
```
dracos: The order/flow here doesn't work – e.g. if you've edited a category, then when this fetch_translations is called it will return the category data rather than the body data it wants (because `posted` is true). I think you need a way to call fetch_translations so it always does the GET case (or split it into two fns for POST/GET) for here (and perhaps could be used for new body then too).

perllib/FixMyStreet/App/Controller/Admin.pm
struan: that's what the check for `$c->stash->{object}` is for.

dracos: Apologies, I had a page open at 6c75d0f9f8b11a5b723f0f1b765f3f1dd223b752 before your force push.

struan: phew, I was wondering what I'd missed :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -514,35 +574,94 @@ sub lookup_body : Private {
...
+    }
+
+    $c->stash->{object} = $c->stash->{body};
+    $c->stash->{translation_col} = 'name';
+
+    # if there's a contact then it's becuase we're displaying error
+    # messages about adding a contact so grabbing translations will
+    # fetch the contact submitted translations. So grab them, stash
+    # them and then clear posted so we can fetch the body translations
+    if ($c->stash->{contact}) {
```
dracos: Sorry, this still doesn't work. `$c->stash->{contact}` is not set when a category is successfully created or edited, because that submission comes to this function, `edit_body`, so the category translations are still then shown in the edit body form you get. I see what this is doing if there is an error, but I think you need to set posted to false regardless. (At least I think that would work - I think this also needs a new test that upon submitting a new category/editing a category, the body translations are still shown in the right place :) )

PR #1786, opened 2017-08-02
[fixmystreet.com] More realistic asphalt tile image

### LINKED ISSUE #1715: [UK] @2x version of FMS.com homepage asphalt background image
The yellow asphalt, especially, looks a bit nasty on retina screens.

zarino:
The black and yellow background image tiles now look more like asphalt
and less like leather.

Also, high-dpi screens now get a double resolution tile, which avoids
noticeable jpeg artifacting on the yellow #front-main in particular.

Fixes #1715.


web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -83,7 +95,7 @@ h3, h4,
 
 #site-header {
     background: none;
-    border-image-source: url(images/tile-y-border.jpg);
+    border-image-source: url(images/tile-y.jpg);
```
davea: Should this have a retina version too?

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -67,9 +67,21 @@ h3, h4,
...
+        background-size: 500px;
+    }
+}
+
 // Tiled background stripe, not plain colour
 #front-main {
     background: $primary url(images/tile-y.jpg);
+
+    @media ($high-dpi-screen) {
+        background-image: url(images/tile-y@2x.jpg);
```
davea: Something is causing both `tile-y.jpg` and `tile-y@2x.jpg` to be requested, maybe because the non-retina version uses `background:` and the retina uses `background-image:`? (Or maybe because of the `border-image-source`, see other comment)

web/cobrands/fixmystreet.com/layout.scss
zarino: I couldn’t wrap my head around how to make double dpi images work inside of `border-image`, and when I tried it out, I realised, in a 4px high border, you really can’t notice when an image has been scaled up anyway. Do you disagree?

web/cobrands/fixmystreet.com/layout.scss
zarino: I think it’s because of the `border-image-source` on `#header`.

web/cobrands/fixmystreet.com/layout.scss
davea: Nope, you're absolutely right. At that size it doesn't really matter if it's a retina asset or not. I guess I was thinking mostly of the unnecessary extra request for the `@1x` image, but not worth bothering with really.

zarino: Mmhmm. We _could_ use a special, smaller, image (like the old tile-y-border.jpg) to reduce the overall filesize … but we’d only want to do that for high-dpi screens, since standard screens will just reuse the existing tile-y.jpg image. Maybe the extra file and a few more lines in `layout.scss` are worth it, to save retina devices a few kilobytes of download cost?

davea: If there's a simple way to save a few KB, it gets my vote 👍

zarino: Worked out how to do it without loading any additional images _at all_. Thanks for the poke ;-)

wrightmartin:
👍  from me, looks lovely on retina™™™

PR #1787, opened 2017-08-04
Always show pagination figures.

dracos:
Even if no next/previous page, the number of items can be useful.

Fixes https://github.com/mysociety/fixmystreetforcouncils/issues/202


dracos:
Yep, should have hidden 0/1 cases now.

PR #1795, opened 2017-08-09
[FixaMinGata] Add updated translation

jonkri:
This pull request adds the lastest (2.1) translation, as "downloaded for use" from Transifex.

Thank you.


dracos:
Thanks!

PR #1796, opened 2017-08-11
Add report as anonymous option for body users

struan:
Adds a permission and an option to allow body users to make an anonymous report/update.

This uses the existing body user as the report user but marks the report/update as anonymous.

Fixes mysociety/fixmystreetforcouncils#207


web/cobrands/fixmystreet/staff.js
```diff
@@ -173,21 +173,31 @@ $.extend(fixmystreet.set_up, {
...
+        } else if (val === 'anonymous_user') {
+            $emailInput.val('-').prop('disabled', true);
+            $nameInput.val('-').prop('disabled', true);
+            $phoneInput.val('-').prop('disabled', true);
+            $showNameCheckbox.prop('checked', false).prop('disabled', true);
+            $addAlertCheckbox.prop('checked', false).prop('disabled', true);
         } else if (val === 'body') {
             $emailInput.val('-').prop('disabled', true);
             $nameInput.val(txt).prop('disabled', true);
+            $phoneInput.val($phoneInput.prop('defaultValue')).prop('disabled', false);
```
dracos: Not sure we should let them change the phone number if they're reporting as body?

struan: That's the current behaviour. I was a bit puzzled by it so yes, I'll change that.

PR #1797, opened 2017-08-11
[WIP] Simplified email & RSS alert signup

zarino:
Partway through implementing the new signup flow in mysociety/fixmystreet-commercial#866.

To do:

- [x] Remove photos of recent reports
- [x] Two submit buttons on first page
   - Work out what to do about geolocation link (if the delivery method is decided by which submit button the user uses, we can’t just forward them to the list page when they click the geolocation link).
- [x] Remove RSS links from list page
- [x] Remove special cases around “Get updates” slide-in drawer on /around pages
- [x] If RSS feed requested, redirect to that when list page is submitted
   - Because `alert_page_buttons` no longer intercepts the form submission, the user is given "ugly" latlng-based feed URLs, rather than postcode-based URLs. Can we do anything about this?
- [x] Progressively enhance results page to ajax submit for real-time RSS feed URL preview
- [ ] Tabs on list page, when in two-tier area
- [ ] Live map preview on list page
- [ ] Customisable radius as `<input type="text">` in first option (RSS feeds only)
- [ ] Google Analytics tracking for location search, list page submission, and RSS feed URL preview mousedowns
- [x] Fix tests
- [ ] Changelog entry

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Alert.pm
+++ b/perllib/FixMyStreet/App/Controller/Alert.pm
@@ -464,8 +464,6 @@ sub setup_request : Private {
         $c->stash->{rznvy} ||= $c->user->email;
     }
 
-    $c->stash->{template} = 'alert/list-ajax.html' if $c->get_param('ajax');
-
     return 1;
 }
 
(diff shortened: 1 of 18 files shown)


zarino:
Closing in favour of https://github.com/mysociety/fixmystreet/pull/1829 which addresses mysociety/fixmystreet-commercial#866 in a more minimal way.

PR #1800, opened 2017-08-15
allow body users to filter on all problem categories

### LINKED ISSUE #1790: Allow body users to filter by report status under "All reports" and "Your reports"
The new report filters (including the possibility to show unfixed, closed and fixed reports) are much appreciated by FixaMinGata municipality users. However, we have received a feature request for extending the granularity of this filter with the report states (such as Not responsible). Thank you.

struan:
Not just New/Fixed/Closed but the Investigating etc

Fixes mysociety/fixmystreetforcouncils#187
Fixes #1790

### DIFF
--- a/templates/web/base/reports/_list-filters.html
+++ b/templates/web/base/reports/_list-filters.html
@@ -9,9 +9,21 @@
         <option value="shortlisted"[% ' selected' IF filter_status.shortlisted %]>[% loc('Shortlisted') %]</option>
         <option value="unshortlisted"[% ' selected' IF filter_status.unshortlisted %]>[% loc('Unshortlisted') %]</option>
       [% END %]
+      [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+        <option value="confirmed"[% ' selected' IF filter_status.confirmed %]>[% loc('Open') %]</option>
+        <option value="investigating"[% ' selected' IF filter_status.investigating %]>[% loc('Investigating') %]</option>
+        <option value="in progress"[% ' selected' IF filter_status.in_progress %]>[% loc('In progress') %]</option>
+        <option value="action scheduled"[% ' selected' IF filter_status.action_scheduled %]>[% loc('Action scheduled') %]</option>
+        <option value="fixed"[% ' selected' IF filter_status.fixed %]>[% loc('Fixed reports') %]</option>
+        <option value="unable to fix"[% ' selected' IF filter_status.unable_to_fix %]>[% loc('No further action') %]</option>
+        <option value="not responsible"[% ' selected' IF filter_status.not_responsible %]>[% loc('Not responsible') %]</option>
+        <option value="internal referral"[% ' selected' IF filter_status.internal_referral %]>[% loc('Internal referral') %]</option>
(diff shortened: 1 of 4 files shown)


templates/web/base/reports/_list-filters.html
```diff
@@ -9,9 +9,21 @@
         <option value="shortlisted"[% ' selected' IF filter_status.shortlisted %]>[% loc('Shortlisted') %]</option>
         <option value="unshortlisted"[% ' selected' IF filter_status.unshortlisted %]>[% loc('Unshortlisted') %]</option>
       [% END %]
+      [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+        <option value="open"[% ' selected' IF filter_status.open %]>[% loc('New reports') %]</option>
+        <option value="investigating"[% ' selected' IF filter_status.investigating %]>[% loc('Investigating') %]</option>
+        <option value="in progress"[% ' selected' IF filter_status.in_progress %]>[% loc('In progress') %]</option>
+        <option value="action scheduled"[% ' selected' IF filter_status.action_scheduled %]>[% loc('Scheduled reports') %]</option>
```
dracos: Just to keep translation easy (and I don't think it really changes how this is used), could these be the states alone, e.g. "Action Scheduled" or "No further action", without 'reports' on the end. I don't think a staff user will mind it like that.
(Also, thinking ahead, if these states are going into the database.)

templates/web/base/reports/_list-filters.html
```diff
@@ -9,9 +9,21 @@
         <option value="shortlisted"[% ' selected' IF filter_status.shortlisted %]>[% loc('Shortlisted') %]</option>
         <option value="unshortlisted"[% ' selected' IF filter_status.unshortlisted %]>[% loc('Unshortlisted') %]</option>
       [% END %]
+      [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+        <option value="open"[% ' selected' IF filter_status.open %]>[% loc('New reports') %]</option>
```
dracos: I guess they're hoping they're new, but if they get reopened by a user etc, they might not be? Looks like Oxon have "New" group covering confirmed and investigating. Perhaps "Open" to match that?
screen shot 2017-08-15 at 14 21 18

templates/web/base/reports/_list-filters.html
```diff
@@ -9,9 +9,21 @@
...
+        <option value="investigating"[% ' selected' IF filter_status.investigating %]>[% loc('Investigating') %]</option>
+        <option value="in progress"[% ' selected' IF filter_status.in_progress %]>[% loc('In progress') %]</option>
+        <option value="action scheduled"[% ' selected' IF filter_status.action_scheduled %]>[% loc('Scheduled reports') %]</option>
+        <option value="fixed"[% ' selected' IF filter_status.fixed %]>[% loc('Fixed reports') %]</option>
+        <option value="no further action"[% ' selected' IF filter_status.no_further_action %]>[% loc('No further action reports') %]</option>
```
dracos: Annoyingly, the state phrase for this is `unable to fix`.

templates/web/base/reports/_list-filters.html
dracos: The value of this should be `confirmed` if you don't want all open states.

PR #1801, opened 2017-08-15
Improve new translation lookup performance

dracos:
This should remember a translation looked up for the remainder of the request, plus add a way to look up all translations when fetching all bodies at once.

### DIFF
--- a/perllib/FixMyStreet/Roles/Translatable.pm
+++ b/perllib/FixMyStreet/Roles/Translatable.pm
@@ -2,6 +2,13 @@ package FixMyStreet::Roles::Translatable;
 
 use Moo::Role;
 
+has _translated  => (is => 'rw');
+
+sub translated {
+    my $self = shift;
+    $self->_translated or $self->_translated({});
+}
+
 sub translate_around {
     my ($orig, $self) = (shift, shift);
     my $fallback = $self->$orig(@_);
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Roles/Translatable.pm
```diff
@@ -18,9 +25,16 @@ sub translate_column {
...
 
+    my %cols = $self->get_columns;
+    return $cols{msgstr} if $cols{msgstr};
+
     my $schema = $self->result_source->schema;
     my $table = lc $self->result_source->source_name;
     my $id = $self->id;
+    my $lang = $schema->lang || '';
+
+    my $translated = $self->translated->{$table}{$id}{$col}{$lang};
```
dracos: `_translated` is attached to the model, so probably don't need $table or $id here, do I.

perllib/FixMyStreet/Roles/Translatable.pm
```diff
@@ -18,9 +25,16 @@ sub translate_column {
...
+    return $translated if $translated;
```
davea: Could this come after the next `if` block in order to speed up problem category translation too?

dracos: Doesn't that slow it down? I guess it could store the problem lookup against the original table/id as well, if that's what you mean.

PR #1804, opened 2017-08-15
Store duplicate reverse lookups upon dupe setting.

dracos:
This avoids the need to search the whole table to find a report's
duplicates. And is easier than constructing an index or upgrading
PostgreSQL.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -325,7 +325,7 @@ sub inspect : Private {
         my %update_params = ();
 
         if ($permissions->{report_inspect}) {
-            foreach (qw/detailed_information traffic_information duplicate_of/) {
+            foreach (qw/detailed_information traffic_information/) {
                 $problem->set_extra_metadata( $_ => $c->get_param($_) );
             }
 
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -360,7 +360,17 @@ sub inspect : Private {
             }
             if ( $problem->state ne 'duplicate' ) {
                 $problem->unset_extra_metadata('duplicate_of');
+            } else {
+                my $duplicate_of = $c->get_param('duplicate_of');
+                $problem->set_extra_metadata( duplicate_of => $duplicate_of );
+                my $dupe = $c->model('DB::Problem')->find($duplicate_of);
```
struan: If you mark a report as a duplicate but don't select a report then this causes a 500 error so there should probably be something to catch that before we get to here.

dracos:
Hopefully works same as before now, plus I factored the common code together.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1115,6 +1115,17 @@ has shortlisted_user => (
     },
 );
 
+sub set_duplicate_of {
+    my ($self, $other_id) = @_;
+    $self->set_extra_metadata( duplicate_of => $other_id );
+    my $dupe = $self->result_source->schema->resultset("Problem")->find($other_id);
+    my $dupes_duplicates = $dupe->duplicates;
```
davea: Rather than hitting the DB here, could you just push `$self->id` onto the existing `duplicates` list in`$dupe->extra`?

Issue #1577, opened 2016-11-17
If no reports on map, "sort by" doesn't work


abimysoc:
https://www.fixmystreet.com/around?pc=PO17+6AR has no reports on the map (at time of writing!).

The list of reports shown in the sidebar is the fallback: closest 20 to that postcode. This means the "sort by" dropdown has no effect - even though the list of reports looks on first glance to be the same as any other list of reports - and thus appears to be broken.

Should perhaps say something like "No reports to show on map, here are some nearby ones" as a clue that this list of reports won't behave like other ones. Or make the sort work on those 20 reports regardless of whether they're on the map or not.

dracos:
It does say the text mentioned above now, added as part of pagination. Think that should do; there's not much point in making the sort work because it is the 20 closest to the map, not necessarily the 20 newest/oldest etc, which it might make you think it was otherwise iyswim.

PR #1805, opened 2017-08-15
Add pagination to around page

dracos:
This factors this code to use the same code `/reports` and `/my` are using, and removes the show old pins code. For #525.
Also adds heading for #1577 (as those nearby ones are excluded from the pagination, which is a bit odd I grant you, but didn't seem odd enough to try and work something out).

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -33,16 +33,20 @@ If no search redirect back to the homepage.
 sub index : Path : Args(0) {
     my ( $self, $c ) = @_;
 
-    # handle old coord systems
-    $c->forward('redirect_en_or_xy_to_latlon');
-
-    # Check if we have a partial report
-    my $partial_report = $c->forward('load_partial');
+    if ($c->get_param('ajax')) {
+        $c->detach('ajax');
+    }
 
     # Check if the user is searching for a report by ID
     if ( $c->get_param('pc') && $c->get_param('pc') =~ $c->cobrand->lookup_by_ref_regex ) {
         $c->go('lookup_by_ref', [ $1 ]);
     }
+
+    # handle old coord systems
+    $c->forward('redirect_en_or_xy_to_latlon');
+
+    # Check if we have a partial report
+    my $partial_report = $c->forward('load_partial');
 
     # Try to create a location for whatever we have
     my $ret = $c->forward('/location/determine_location_from_coords')
@@ -275,7 +279,7 @@ sub map_features : Private {
     $c->stash->{distance} = $distance;
 }
 
-=head2 /ajax
+=head2 ajax
 
 Handle the ajax calls that the map makes when it is dragged. The info returned
 is used to update the pins on the map and the text descriptions on the side of
@@ -283,7 +287,7 @@ the map.
 
 =cut
 
-sub ajax : Path('/ajax') {
+sub ajax : Private {
     my ( $self, $c ) = @_;
 
     my $bbox = $c->get_param('bbox');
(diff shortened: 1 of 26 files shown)


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -140,11 +140,11 @@ sub _recent {
 # Problems around a location
 
 sub around_map {
-    my ( $rs, $limit, %p) = @_;
+    my ( $rs, %p) = @_;
     my $attr = {
         order_by => $p{order},
     };
-    $attr->{rows} = $limit if $limit;
+    $attr->{rows} = FixMyStreet->config('ALL_REPORTS_PER_PAGE');
```
struan: Should this not use `cobrand->reports_per_page`?

templates/web/base/around/on_map_list_items.html
```diff
@@ -1,14 +1,32 @@
...
+[% IF around_map.size %]
+<ul class="item-list item-list--reports">
+    [% FOREACH problem IN around_map %]
+        [% dist = tprintf("%.1f", (problem.distance || 0) );
+           problem = problem.problem;
+        %]
+        [% INCLUDE 'reports/_list-entry.html' %]
+    [% END %]
+</ul>
+[% END %]
```
struan: This change seems to have broken the highlight pin on rollover behaviour.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -167,10 +167,7 @@ sub display_location : Private {
     my $latitude  = $c->stash->{latitude};
     my $longitude = $c->stash->{longitude};
 
-    # Deal with pin hiding/age
-    my $all_pins = $c->get_param('all_pins') ? 1 : undef;
-    $c->stash->{all_pins} = $all_pins;
-    my $interval = $all_pins ? undef : $c->cobrand->on_map_default_max_pin_age;
+    my $interval = $c->cobrand->on_map_default_max_pin_age;
```
struan: So how do I see really old reports now?

dracos: Whoops, no need for that interval at all any more, is there.

templates/web/base/around/on_map_list_items.html
dracos: Turns out this is also already broken on e.g. a `/reports` page after you pan the map and the pins have updated. Will fix for everywhere hopefully.

perllib/FixMyStreet/DB/ResultSet/Problem.pm
dracos: Annoying have to get the cobrand down in that fn now though, hmm.

dracos:
Tested on live today for a while:
- Pagination expects ajax calls to be same as normal calls with ajax=1 query parameter. `/around` currently uses `/ajax`. This also means /around has to work with a bbox parameter, like the ajax version does.
- Pagination reload with bbox in URL does not remember zoom (it does get centre from the bbox parameter). This currently works on `/reports` because that has zoom to extent on load and no reload of pins on load. I think having zoom to extent on other pages would be okay, too, maybe.
- ajax currently ignores pagination (/reports etc pagination reloads full page); this means when the pins load on page load on /around they reset to page 1.

The pagination links should perform an ajax call rather than a full page load, and that would hopefully fix most/all the above.

davea:
There's an odd bug which is stopping all pins being shown on the map when you zoom in.

If you go to a location and zoom in a couple of times, no more pins are added to the map:
zoom1

However if you change the filters once you're zoomed in to force a reload of all pins in the viewport, the 'hidden' pins are revealed:
zoom2

Here's a comparison of the two end results highlighting the difference:
comparison


I'm guessing this is due to some odd interaction between the pagination and the 'don't reload everything when zooming in' behaviours?

dracos:
Ah yes, I need to behave as on `/reports` and refresh on zoom (as default strategy assumes you're showing everything within viewport so need do nothing on zoom in). Have tweaked default Strategy (that `/reports` strategy inherits from) to refresh on zoom, hopefully that fixes it.

PR #1806, opened 2017-08-16
show all report updates, even if no public text

struan:
Create a comment entry for all updates to reports regardless of whether
there is any text part. Also adds an entry to extra if a defect was
raised and includes that in the update.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -320,7 +320,7 @@ sub inspect : Private {
         $c->forward('/auth/check_csrf_token');
 
         my $valid = 1;
-        my $update_text;
+        my $update_text = '';
         my $reputation_change = 0;
         my %update_params = ();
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -290,6 +300,10 @@ sub meta_line {
 
     }
 
+    if ($self->get_extra_metadata('defect_raised')) {
+        $update_state .= ( $update_state ? ', ' : '' ) . _( 'and a defect raised' );
```
dracos: I think putting this in `$update_state` will mean the next update, even if the state stays the same, will lead to a "marked as action scheduled" message because `$update_state` itself will have changed.

dracos: Is there a situation where update_state can't be set here? It doesn't feel like there should be.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -364,6 +364,8 @@ sub inspect : Private {
             if ( $problem->state ne $old_state ) {
                 $c->forward( '/admin/log_edit', [ $problem->id, 'problem', 'state_change' ] );
 
+                $update_params{problem_state} = $problem->state;
```
dracos: Oh dear that this wasn't in there before!

PR #1811, opened 2017-08-17
Don't fetch pins serverside with JS on

### LINKED ISSUE #1808: Sidebar hover behaviour lost after map pan
The JS handler is being placed on an element that is replaced when new data is pulled in via JS.

dracos:
The page currently fetches the pins again client-side (as we don't know the extent of the map at the server point). I realise this isn't ideal, but should speed up the normal case quite a bit.
Also fixes #1808.


web/cobrands/fixmystreet/staff.js
```diff
@@ -93,7 +93,7 @@ $.extend(fixmystreet.set_up, {
   },
 
   list_item_actions: function() {
-    $('.item-list--reports').on('click', ':submit', function(e) {
+    $('#js-reports-list').on('click', ':submit', function(e) {
```
dracos: This breaks the actions on the front page which is not within this ID.

PR #1814, opened 2017-08-18
Improve /report loading time for inspectors

davea:
On `/report` pages several queries were missing a `JOIN`, meaning hundreds of excess queries and a very slow page load time when logged in as an inspector.

This PR improves the queries by adding the `JOIN` which makes things nice and snappy again.

[skip changelog]


davea:
Merged in 6a2734fb46c9edebe6d455c4ab1984065cc64485, will open another PR for speed ups on /reports.

PR #1816, opened 2017-08-18
Speed up report lists for inspectors

davea:


dracos:
Just to note on 20f9342, this was to reduce a set of queries from 500 to 100 where `owns_problem` looks up bodies and body_areas, and is called five times per report on `/reports/...` - pin_colour twice (for colour, and Oxfordshire extra class name), base_url_for_report twice (pin and sidebar), and once directly for some sidebar text. Should be possible to get down to one per body if we can cache that lookup somewhere.

PR #1817, opened 2017-08-21
Another couple of /reports speedups

dracos:
[skip changelog]


dracos:
Yeah, those would probably work in Catalyst context, but I thought this may as well do the same e.g. in a cron script too (though probably less important). Perhaps we should add InstancePerContext anyway for the couple of times we are passing it down through the stack… There is a DB subclass of that at https://metacpan.org/pod/Catalyst::TraitFor::Model::DBIC::Schema::PerRequestSchema which should let us do that, though ISTR having an issue when I tried this before, I'll have a quick go again now.

PR #1820, opened 2017-08-22
Redirect to original URL after inspecting

davea:
Fixes mysociety/fixmystreetforcouncils#204

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -316,6 +316,10 @@ sub inspect : Private {
         $c->stash->{templates_by_category} = $templates_by_category;
     }
 
+    if ($c->user->has_body_permission_to('planned_reports')) {
+        $c->stash->{post_inspect_url} = $c->req->referer;
+    }
+
     if ( $c->get_param('save') ) {
         $c->forward('/auth/check_csrf_token');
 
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -449,20 +453,23 @@ sub inspect : Private {
...
             if ($c->user->has_body_permission_to('planned_reports')) {
-                my $categories = join(',', @{ $c->user->categories });
-                my $params = {
-                    lat => $problem->latitude,
-                    lon => $problem->longitude,
-                };
-                $params->{filter_category} = $categories if $categories;
-                $params->{js} = 1 if $c->get_param('js');
-                $redirect_uri = $c->uri_for( "/around", $params );
+                unless ($redirect_uri = $c->get_param("post_inspect_url")) {
```
dracos: Looks like this not being present never happens in test now.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -449,20 +453,23 @@ sub inspect : Private {
...
-            # shortlist because it makes it much simpler to inspect many reports
-            # in the same location. The shortlist is always a single click away,
-            # being on the main nav.
+            # Or if inspector, redirect back to the map view they came from
+            # with the right filters. If that wasn't set, go to /around at this
+            # report's location.
+            # We go here rather than the shortlist because it makes it much
+            # simpler to inspect many reports in the same location. The
+            # shortlist is always a single click away, being on the main nav.
             if ($c->user->has_body_permission_to('planned_reports')) {
```
dracos: As this overrides the work done before, should it perhaps come first in the previous if statement (I say first so that it doesn't have to bother doing the `owns_problem` check if this one passes).

davea:
Ta for the review. Good catch on the opportunity to remove a DB query! I've fixed that up and added a test for when `post_inspect_url` is empty.

Away until Tuesday so please do merge if you're happy with this!

PR #1823, opened 2017-08-29
Add debug toolbar middleware

dracos:
And a fix for make_css new perl warnings.
screen shot 2017-08-29 at 09 41 09


script/server
```diff
@@ -3,4 +3,5 @@
 set -e
 cd "$(dirname "$0")/.."
 
-script/fixmystreet_app_server.pl --fork --restart --debug $@
+export FIXMYSTREET_APP_DEBUG=1
```
davea: Could this be conditional on `FIXMYSTREET_APP_DEBUG` not already being defined? It'd be handy to be able to disable the toolbar by running `FIXMYSTREET_APP_DEBUG=0 ./script/server`.

PR #1827, opened 2017-08-29
Move report states to database

dracos:
This moves almost all the report states to the database where they can be edited/removed/added, and translated. It also makes 'fixed' optional so if it's not present certain things will not function (e.g. questionnaires, list filter). The internal states remain hardcoded as do the core 'open' and 'closed' (which are also used for their 'group' names).

One outstanding thing is I've asked @zarino to make the state change in an update look nicer now that it is stilted.


zarino:
@dracos I’ve been thinking about an icon to represent the concept of "this is the report’s state". How about a `#` hashtag? Social media users will be familiar with the idea of hashtags as declaring meta data about a message, and you sometimes also see them used to denote categories on blogs (like https://www.mysociety.org/blog).

We could make a super minimal change in as part of this PR, that just adds the hash symbol before the updated status. And then, separately, I can do another PR that introduces the hash symbol near the top of every report (jnstead of the current "hanging" status tab/bookmark) to reinforce the meaning.

PR #1828, opened 2017-09-01
Fix a crash if changing report category without state field present

davea:
Spotted on the Ground Control cobrand, where the inspector form only shows the category field and not the state field.

(NB: it looks like this functionality is [tested in `report_inspect.t`](https://github.com/mysociety/fixmystreet/blob/0c4b5041d807899ab708db4c5dd15203d23df257/t/app/controller/report_inspect.t#L231) so I'm not sure why this wasn't noticed sooner, unless I'm reading the test wrong.)

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -295,6 +295,16 @@ sub permissions {
 sub has_permission_to {
     my ($self, $permission_type, $body_ids) = @_;
 
+    # Nobody, including superusers, can have a permission which isn't available
+    # in the current cobrand.
+    my $cobrand = $self->result_source->schema->cobrand;
+    my $cobrand_perms = $cobrand->available_permissions;
+    my %available = map { %$_ } values %$cobrand_perms;
+    # The 'trusted' permission is never set in the cobrand's
+    # available_permissions (see note there in Default.pm) so include it here.
+    $available{trusted} = 1;
+    return 0 unless $available{$permission_type};
+
     return 1 if $self->is_superuser;
     return 0 if !$body_ids || (ref $body_ids && !@$body_ids);
     $body_ids = [ $body_ids ] unless ref $body_ids;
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -295,6 +295,15 @@ sub permissions {
 sub has_permission_to {
     my ($self, $permission_type, $body_ids) = @_;
 
+    # Nobody, including superusers, can have a permission which isn't available
+    # in the current cobrand.
+    if (my $cobrand = $self->result_source->schema->cobrand) {
```
dracos: Shouldn't cobrand always be set if we've got here? It looks like only test failure was in `t/app/sendreport/inspection_required.t` in  https://travis-ci.org/mysociety/fixmystreet/jobs/270780037 so is the issue with the test? It trying to use a cobrand without overriding ALLOWED_COBRANDS, maybe?

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -295,6 +295,15 @@ sub permissions {
...
+        my $cobrand_perms = $cobrand->available_permissions;
+        my $flat_perms = { map { %$_ } values %$cobrand_perms };
+        my %available = map { $_ => 1 } keys %$flat_perms;
```
dracos: These two lines could be `my %available = map { %$_ } values %$cobrand_perms;`

No need to put in a hashref and then read it back out again (the issue before your last commit was you just needed `keys map`, no ref in between), but also no need to  rewrite the values to be `1` if you're just checking a value exists :)

perllib/FixMyStreet/DB/Result/User.pm
davea: I saw a failure in `t/app/view/email.t`locally and so added that check, but yes I'll look into whether the test itself is wrong.

davea: Sorted, one of the tests was calling`has_permission_to` directly without a cobrand ever having been set up.

PR #1829, opened 2017-09-01
Simpler alerts signup page

zarino:
Fixes mysociety/fixmystreet-commercial#866.

Alert options on the /alert/list page are now in an order that is easier to parse, with a slightly clearer message for users in two-tier locations, and a more prominent email subscription CTA.

Minimal changes to styling. No changes to the form behaviour.

# Standard location

alerts

# Two-tier location

alerts two tier

@jukesie, do you want some Google Analytics event tracking so we know how many people actually sign up for email alerts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     - Front end improvements:
         - Always show pagination figures even if only one page. #1787
         - Report pages list every update to a report. #1806
+        - Clearer wording and more prominent email input on alert page.
         - Cobrands can implement `hide_areas_on_reports` to hide outline on map.
     - Admin improvements:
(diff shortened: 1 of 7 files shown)


dracos:
screen shot 2017-09-05 at 15 49 48

zarino:
Yeah, I think the way forward with the two-tier display is listing them vertically and then progressively enhancing to tabs with JavaScript. But I figured we could deploy what I've got here, then add the tabs in later.

dracos:
I like it better in two columns, so as I've done it, I'll include it if that's okay with you? Unless you're saying you'd prefer it vertically in a fallback situation.

PR #1833, opened 2017-09-05
[Oxfordshire] Log when problems are sent in RDI

davea:
The timestamp of when a problem was included in an RDI is shown on the problem’s
report_edit page in the Oxfordshire cobrand admin.

If a report hasn't been inspected yet:
screen shot 2017-09-05 at 12 00 53

If it's been inspected and instructed but not yet sent:
screen shot 2017-09-05 at 12 00 45

And once it's been sent in an RDI:
screen shot 2017-09-05 at 09 41 15

Fixes mysociety/fixmystreetforcouncils#223.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Integrations/ExorRDI.pm
+++ b/perllib/FixMyStreet/Integrations/ExorRDI.pm
@@ -23,7 +23,7 @@ use Scalar::Util 'blessed';
 use FixMyStreet::DB;
 use namespace::clean;
 
-has [qw(start_date end_date inspection_date)] => (
+has [qw(start_date end_date inspection_date mark_as_processed)] => (
     is => 'ro',
     required => 1,
 );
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Integrations/ExorRDI.pm
```diff
@@ -198,6 +208,14 @@ sub construct {
         0, 0, 0 # error counts, always zero
     );
 
+    # Mark all these problems are having been included in an RDI
+    my $now = DateTime->now->strftime( '%Y-%m-%d %H:%M' );
+    $problems->reset;
+    while ( my $report = $problems->next ) {
+        $report->set_extra_metadata('rdi_processed' => $now);
```
dracos: Do we want to do this if the RDI file has been generated via the admin page, rather than by email?

perllib/FixMyStreet/Integrations/ExorRDI.pm
```diff
@@ -43,13 +43,23 @@ sub construct {
...
-            'admin_log_entries.whenedited' => { '>=', $dtf->format_datetime($self->start_date) },
-            'admin_log_entries.whenedited' => { '<=', $dtf->format_datetime($self->end_date) },
+            -or => [
+                -and => [
+                    'admin_log_entries.action' => 'inspected',
+                    'admin_log_entries.whenedited' => { '>=', $dtf->format_datetime($self->start_date) },
+                    'admin_log_entries.whenedited' => { '<=', $dtf->format_datetime($self->end_date) },
+                ],
+                -and => [
+                    extra => { -not_like => '%rdi_processed%' },
```
dracos: As well as the "probably don't want to do this when admin page", is there a performance concern here? I guess it is only the one cobrand + state and date, so probably not?

templates/web/oxfordshire/admin/report_edit_extra_info.html
```diff
@@ -0,0 +1,12 @@
+<li class="sm">
+    [% loc('RDI sent:') %]
+    [% IF problem.inspection_log_entry %]
+        [% IF problem.get_extra_metadata('rdi_processed') %]
```
dracos: If this is true, the former must be true, so can swap and not have to do that check?
It also looks like we set 'inspected' on problem metadata (used elsewhere in template), which presumably saves an admin log search if we used that.

PR #1834, opened 2017-09-05
basic area stats pages

struan:
Admin page to show some simple summary stats for an area.

If the user has been assigned to an area then they will see the stats
for that area. Superusers can pick which are they want to view.

For mysociety/fixmystreetforcouncils#2

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
@@ -1,0 +1,218 @@
+package FixMyStreet::App::Controller::Admin::AreaStats;
+use Moose;
+use namespace::autoclean;
+use List::Util qw(sum);
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+sub index : Path : Args(0) {
+    my ( $self, $c ) = @_;
+
+    my $user = $c->user;
+
+    if ($user->is_superuser) {
+        $c->forward('/admin/fetch_all_bodies');
+    } elsif ( $user->from_body ) {
+        $c->forward('load_user_body', [ $user->from_body->id ]);
+        $c->stash->{body_id} = $user->from_body->id;
+        if ($user->area_id) {
+            $c->stash->{area_id} = $user->area_id;
+            $c->forward('setup_area');
+            $c->visit( 'stats' );
+        } else {
+            # visit body_stats so we load the list of child areas
+            $c->visit( 'body_stats', [ $user->from_body->id], [] );
+        }
+    } else {
+        $c->detach( '/page_error_404_not_found' );
+    }
+}
+
+sub check_user : Private {
+    my ( $self, $c, $body_id, $area_id ) = @_;
+
+    my $user = $c->user;
+
+    return if $user->is_superuser;
+
+    if ($body_id and $user->from_body->id eq $body_id) {
+        if (not $user->area_id) {
+            return;
+        } elsif ($area_id and $user->area_id eq $area_id) {
+            return;
+        }
+    }
+
+    $c->detach( '/page_error_404_not_found' );
+}
+
+sub setup_area : Private {
+    my ($self, $c) = @_;
+
+    my $area = mySociety::MaPit::call('area', $c->stash->{area_id} );
+    $c->detach( '/page_error_404_not_found' ) if $area->{error};
+    $c->stash->{area} = $area;
+}
+
+sub body_base : Chained('/') : PathPart('admin/areastats') : CaptureArgs(1) {
+    my ($self, $c, $body_id) = @_;
+
+    $c->forward('/admin/lookup_body', $body_id);
+    $c->stash->{areas} = mySociety::MaPit::call('area/children', [ $c->stash->{body_id} ] );
+}
+
+sub body_stats : Chained('body_base') : PathPart('') : Args(0) {
+    my ($self, $c) = @_;
+
+    if ($c->get_param('area')) {
+        $c->forward('check_user', [$c->stash->{body_id}, $c->get_param('area')]);
+        $c->stash->{area_id} = $c->get_param('area');
+        $c->forward('setup_area');
+    } else {
+        $c->forward('check_user', [$c->stash->{body_id}]);
+    }
+    $c->forward('stats');
+}
+
+sub stats : Private {
+    my ($self, $c) = @_;
+
+    my $date = DateTime->now->subtract(days => 30);
+
+    $c->forward('/admin/fetch_contacts');
+
+    $c->stash->{template} = 'admin/areastats/area.html';
+
+    my $dtf = $c->model('DB')->storage->datetime_parser;
+    my $time = $dtf->format_datetime($date);
+
+    my $params = {
+        'me.confirmed' => { '>=', $time },
+    };
+
+    my %area_param = ();
+    if ($c->stash->{area}) {
+        my $area_id = $c->stash->{area_id};
+        $c->stash->{area_name} = $c->stash->{area}->{name};
+        $params->{'problem.areas'} = { like => "%,$area_id,%" };
+        %area_param = (
+            areas => { like => "%,$area_id,%" },
+        );
+    } else {
+        $c->stash->{area_name} = $c->stash->{body}->name;
+    }
+
+    my %by_category = map { $_->category => {} } $c->stash->{contacts}->all;
+    my %recent_by_category = map { $_->category => 0 } $c->stash->{contacts}->all;
+
+    my $state_map = {};
+
+    $state_map->{$_} = 'open' foreach FixMyStreet::DB::Result::Problem->open_states;
+    $state_map->{$_} = 'closed' foreach FixMyStreet::DB::Result::Problem->closed_states;
+    $state_map->{$_} = 'fixed' foreach FixMyStreet::DB::Result::Problem->fixed_states;
+    $state_map->{$_} = 'scheduled' foreach ('planned', 'action scheduled');
+
+    # current problems by category and state
+    my $problems = $c->model('DB::Problem')->to_body(
+        $c->stash->{body}
+    )->search(
+        \%area_param,
+        {
+            group_by => [ 'category', 'state' ],
+            select   => [ 'category', 'state', { count => 'me.id' } ],
+            as       => [ qw/category state state_count/ ],
+        }
+    );
+
+    while (my $p = $problems->next) {
+        my $meta_state = $state_map->{$p->state};
+        $by_category{$p->category}->{$meta_state} += $p->get_column('state_count');
+    }
+    $c->stash->{by_category} = \%by_category;
+
+    # problems this month by state
+    $c->stash->{$_} = 0 for values %$state_map;
+
+    $c->stash->{open} = $c->model('DB::Problem')->to_body(
+        $c->stash->{body}
+    )->search(
+        {
+            %area_param,
+            confirmed => { '>=' => $time },
+        }
+    )->count;
+
+    my $comments = $c->model('DB::Comment')->to_body(
+        $c->stash->{body}
+    )->search(
+        $params,
+        {
+            join     => 'problem'
+        }
+    );
+
+    # you can have multiple comments with the same problem state so need to only count
+    # one instance.
+    my %state_seen = ();
+    while (my $comment = $comments->next) {
+        my $meta_state = $state_map->{$comment->problem_state};
+        my $key = $comment->problem->id . "-$meta_state";
+        next if $state_seen{$key};
+        $c->stash->{$meta_state} += 1;
+        $state_seen{$key} = 1;
+    }
+
+    $params = {
+        %area_param,
+        'me.confirmed' => { '>=', $time },
+    };
+
+    # problems this month by category
+    my $recent_problems = $c->model('DB::Problem')->to_body(
+        $c->stash->{body}
+    )->search(
+        $params,
+        {
+            group_by => [ 'category' ],
+            select   => [ 'category', { count => 'me.id' } ],
+            as       => [ qw/category category_count/ ],
+        }
+    );
+
+    while (my $p = $recent_problems->next) {
+        $recent_by_category{$p->category} += $p->get_column('category_count');
+    }
+    $c->stash->{recent_by_category} = \%recent_by_category;
+
+    # average time to state change in last month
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+
+sub body_stats : Chained('body_base') : PathPart('') : Args(0) {
+    my ($self, $c) = @_;
+    my $user = $c->user;
+
+    $c->forward('check_user', [$c->stash->{body_id}]);
+    if ($c->get_param('area')) {
+        $c->stash->{area_id} = $c->get_param('area');
+    } else {
+        $c->stash->{area_id} = $c->stash->{body}->id;
```
dracos: Area IDs don't have to match Body IDs (and definitely don't on my dev site). You can't do a MapIt lookup with a body ID, I'm afraid. Bodies have their own name, you only need to use MapIt for ward names (if wards are a thing).

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+    $c->stash->{area_id} = $area_id;
+    $c->forward('stats');
+}
+
+sub stats : Private {
+    my ($self, $c) = @_;
+
+    my $date = DateTime->now->subtract(days => 30);
+    my $area_id = $c->stash->{area_id};
+    my $area = mySociety::MaPit::call('area', $area_id );
```
dracos: (This is the line that won't work for bodies, only wards.)

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+    my $user = $c->user;
+
+    $c->forward('/admin/fetch_contacts');
+
+    $c->stash->{template} = 'admin/areastats/area.html';
+    if ($area->{name}) {
```
dracos: You could detach the 404 here and not indent the rest of the function :) (But I think this will be changing anyway.)

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+        $c->stash->{area_id} = $c->get_param('area');
+    } else {
+        $c->stash->{area_id} = $c->stash->{body}->id;
+    }
+    $c->forward('stats');
+}
+
+sub ward_stats : Chained('body_base') :PathPart('') : Args(1) {
+    my ($self, $c, $area_id) = @_;
+    my $user = $c->user;
```
dracos: $user not used.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+    $c->forward('stats');
+}
+
+sub stats : Private {
+    my ($self, $c) = @_;
+
+    my $date = DateTime->now->subtract(days => 30);
+    my $area_id = $c->stash->{area_id};
+    my $area = mySociety::MaPit::call('area', $area_id );
+    my $user = $c->user;
```
dracos: Not used.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+        $c->detach( '/page_error_404_not_found' );
+    }
+}
+
+sub check_user : Private {
+    my ( $self, $c, $body_id, $area_id ) = @_;
+
+    my $user = $c->user;
+
+    if ($user->is_superuser) {
```
dracos: Might just be me, but I prefer a simple condition like this to be written `return if $user->is_superuser;` as it's clear it's a "get out" clause.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+        my %recent_by_category = map { $_->category => 0 } $c->stash->{contacts}->all;
+
+        my $state_map = {};
+
+        $state_map->{$_} = 'open' foreach FixMyStreet::DB::Result::Problem->open_states;
+        $state_map->{$_} = 'closed' foreach FixMyStreet::DB::Result::Problem->closed_states;
+        $state_map->{$_} = 'fixed' foreach FixMyStreet::DB::Result::Problem->fixed_states;
+        $state_map->{$_} = 'scheduled' foreach ('planned', 'action scheduled');
+
+        for my $p ($problems->all) {
```
dracos: A while loop with `->next` would use less memory (I'd hope).

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+                as       => [ qw/category state state_count/ ],
+            }
+        );
+
+        my %by_category = map { $_->category => {} } $c->stash->{contacts}->all;
+        my %recent_by_category = map { $_->category => 0 } $c->stash->{contacts}->all;
+
+        my $state_map = {};
+
+        $state_map->{$_} = 'open' foreach FixMyStreet::DB::Result::Problem->open_states;
```
dracos: This will include planned/action scheduled which are also under scheduled, is that as expected?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+        $state_map->{$_} = 'scheduled' foreach ('planned', 'action scheduled');
+
+        for my $p ($problems->all) {
+            my $meta_state = $state_map->{$p->state};
+            $by_category{$p->category}->{$meta_state} += $p->get_column('state_count');
+        }
+
+        $params->{'me.confirmed'} = { '>=', $time };
+
+        my $recent_problems = $c->model('DB::Problem')->search(
```
dracos: Isn't the result of this query covered by a subset of the results of the last query? As in, with a conditional in the prior loop, you could construct this without going back to the database?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+
+        $c->stash->{$_} = 0 for values %$state_map;
+        $c->stash->{open} = $c->model('DB::Problem')->search(
+            {
+                areas => { like => "%,$area_id,%" },
+                confirmed => { '>=' => DateTime::Format::W3CDTF->format_datetime($date) },
+            }
+        )->count;
+
+        for my $comment ($comments->all) {
```
dracos: A comment `problem_state` is set (most of) the time with an update, regardless of if there's been a state change or not; so I don't think it can be used as is as a statistic for the number of issues scheduled/closed/fixed in the past month :-/ You could perhaps go on staff users only? But not even then I imagine. Not sure what to suggest here, something to discuss.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
+package FixMyStreet::App::Controller::Admin::AreaStats;
+use Moose;
+use namespace::autoclean;
+use List::Util qw(sum);
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+sub begin : Private {
```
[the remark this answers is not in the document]
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    $c->detach( '/page_error_404_not_found' );
+}
+
+sub body_base : Chained('/') : PathPart('admin/areastats') : CaptureArgs(1) {
+    my ($self, $c, $body_id) = @_;
+    $c->stash->{body_id} = $body_id;
+    $c->forward('/admin/lookup_body');
```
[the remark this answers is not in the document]
struan: That runs into problems with `check_page_allowed`.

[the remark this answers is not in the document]
dracos: Hmm, I see; I guess putting a `lookup_body` function back (and having admin's `sub body` just call that) is probably easiest then?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: It means doing all the summing in code rather than in the DB because it isn't selecting `confirmed` and grouping by that doesn't make sense. And I'm not sure that one loop over every record returned is better than two DB queries, especially doing date maths on every record. :/

dracos: Fair enough. You could group on the boolean "me.confirmed >= $time" and then you'd have either true or false, but that's probably too confusing. As long as the query speed on the whole dataset is okay, it's fine.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: ✅

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,217 @@
...
+      unless $body;
+    $c->stash->{body} = $body;
+    $c->stash->{areas} = mySociety::MaPit::call('area/children', [ $c->stash->{body_id} ] );
+}
+
+sub body_stats : Chained('body_base') : PathPart('') : Args(0) {
+    my ($self, $c) = @_;
+
+    $c->forward('check_user', [$c->stash->{body_id}]);
+    if ($c->get_param('area')) {
```
dracos: Not sure what this is doing? I can't see `area_id` being used after this point, I might have missed it!

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,217 @@
...
+            as       => [ qw/category state state_count/ ],
+        }
+    );
+
+    my %by_category = map { $_->category => {} } $c->stash->{contacts}->all;
+    my %recent_by_category = map { $_->category => 0 } $c->stash->{contacts}->all;
+
+    my $state_map = {};
+
+    $state_map->{$_} = 'open' foreach FixMyStreet::DB::Result::Problem->open_states;
```
dracos: Going by how this is displayed, I think they expect a report to appear only in one of open/scheduled/closed/fixed. So perhaps open should be all open_states that aren't manually being used for scheduled?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+
+    $c->stash->{$_} = 0 for values %$state_map;
+    $c->stash->{open} = $c->model('DB::Problem')->to_body(
+        $c->stash->{body}
+    )->search(
+        {
+            %area_param,
+            confirmed => { '>=' => DateTime::Format::W3CDTF->format_datetime($date) },
```
dracos: I think you can use `$time` again here?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -0,0 +1,217 @@
...
+        }
+    )->count;
+
+    while (my $comment = $comments->next) {
+        my $meta_state = $state_map->{$comment->get_column('state')};
+        $c->stash->{$meta_state} += $comment->get_column('state_count');
```
dracos: Not sure if this was missed: A comment problem_state is set (most of :-/ ) the time with an update, regardless of if there's been a state change or not; so I don't think it can be used as is as a statistic for the number of issues scheduled/closed/fixed in the past month :-/ You could perhaps go on staff users only? But not even then I imagine.
Inspect form now leaves an update; if the admin editing is similarly updated, then at least we'd have an update for every state change (after a month). But even then you'd have to go through the updates to know when one was a 'change' and when one wasn't changing it.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
struan: Isn't that taken care of by:

`$state_map->{$_} = 'scheduled' foreach ('planned', 'action scheduled');`

And for the "problems this month bit" I read that as "n problems opened of which x were scheduled, y were fixed and z closed" so that people can check that problems are flowing through.

dracos: Oh yes, silly me.

struan:
Is this now ok to go?

dracos:
Looks like it; could you check #1849, then we merge that, and rebase/tidy this on top of that just to check the test failures were caused by that issue, thanks.

PR #1835, opened 2017-09-06
sort body dashboard CSV export by report confirmed time

struan:
And update the tests to avoid a random failure caused by inconsistent
ordering.

### DIFF
--- a/t/app/controller/dashboard.t
+++ b/t/app/controller/dashboard.t
@@ -531,12 +531,12 @@ FixMyStreet::override_config {
             desc => 'Selecting no state does nothing',
             p1 => {
                     state   => 'fixed - user',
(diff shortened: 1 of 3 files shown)


t/app/controller/dashboard.t
```diff
@@ -531,12 +531,12 @@ FixMyStreet::override_config {
             desc => 'Selecting no state does nothing',
             p1 => {
                     state   => 'fixed - user',
-                    conf_dt => DateTime->now(),
+                    conf_dt => DateTime->now()->subtract( minutes => 1 ),
```
dracos: If they're now sorted by time, is there a need to have this as well? Guess it doesn't matter, so might as well leave, but surely the two can't be in the same microsecond? :)

PR #1837, opened 2017-09-08
Enable including extra text at all problem confirmation stages

struan:
This adds some extra templates to enable adding cobrand messages to

* Confirmation email sent page.
* The confirm problem email.
* Problem confirmed page.

It also display the expected response time message in those locations for the Oxfordshire cobrand.

Fixes mysociety/fixmystreetforcouncils#219

### DIFF
--- a/templates/email/default/problem-confirm.html
+++ b/templates/email/default/problem-confirm.html
@@ -17,6 +17,8 @@ INCLUDE '_email_top.html';
 Please note that [% c.cobrand.council_name %] is not responsible for this type
 of problem, so it will instead be sent to [% report.body %].
 [% END %]
+
+[% INCLUDE '_problem-confirm_extra.html' %]
   </p>
   <p style="[% p_style %]">Your report will also appear on the [% site_name %] website.</p>
   <p style="margin: 20px auto; text-align: center">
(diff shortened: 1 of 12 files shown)


templates/email/default/problem-confirm.html
```diff
@@ -17,6 +17,8 @@ <h1 style="[% h1_style %]">Please confirm your&nbsp;report</h1>
 Please note that [% c.cobrand.council_name %] is not responsible for this type
 of problem, so it will instead be sent to [% report.body %].
 [% END %]
+
+[% INCLUDE '_problem-confirm_extra.html' %]
```
dracos: If you use TRY / INCLUDE / CATCH file / END here (see examples elsewhere) you won't need to have the blank files under `default/`.

templates/web/base/email_sent.html
```diff
@@ -14,6 +14,8 @@
     [% END %]
     </p>
 
+    [% INCLUDE '_email_sent_extra.html' %]
```
dracos: Ditto

templates/email/default/problem-confirm.html
struan: ah, for some reason I had it in my head that empty files were better than TRY / INCLUDE / CATCH. I have no idea why :|

PR #1838, opened 2017-09-08
default response priorities

struan:
Add and use default response priorities.

Fixes mysociety/fixmystreetforcouncils#224

### DIFF
--- a/perllib/FixMyStreet/DB/Result/ResponsePriority.pm
+++ b/perllib/FixMyStreet/DB/Result/ResponsePriority.pm
@@ -28,6 +28,8 @@ __PACKAGE__->add_columns(
   { data_type => "text", is_nullable => 1 },
   "external_id",
   { data_type => "text", is_nullable => 1 },
+  "is_default",
+  { data_type => "boolean", default_value => \"false", is_nullable => 0 },
 );
 __PACKAGE__->set_primary_key("id");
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/DB/Result/ResponsePriority.pm
```diff
@@ -28,6 +28,8 @@ __PACKAGE__->add_columns(
   { data_type => "text", is_nullable => 1 },
   "external_id",
   { data_type => "text", is_nullable => 1 },
+  "is_default",
+  { data_type => "boolean", default_value => \"false", is_nullable => 0 },
```
dracos: i'm guessing as the MD5 hash below didn't change, this wasn't generated by running `db/rerun_dbic_loader.pl` :) Which doesn't really matter, as it seems like column order in different dev databases causes changes there anyway, but if running it adds this and doesn't change anything else but update the hash, might be worth doing.

t/app/controller/admin.t
```diff
@@ -1470,6 +1470,7 @@ subtest "response priorities can be added" => sub {
         name => "Cat 1A",
         description => "Fixed within 24 hours",
         deleted => undef,
+        is_default => undef,
```
dracos: Could we have a test setting it as well? I realise it's straightforward, but might as well to make sure it doesn't break in future :)

PR #1841, opened 2017-09-13
[Oxfordshire] Use local time for rdi_processed, not UTC

davea:
[skip changelog]


perllib/FixMyStreet/Integrations/ExorRDI.pm
```diff
@@ -37,13 +37,18 @@ has user => (
     },
 );
 
+my $now = DateTime->now(
```
dracos: Has to be in the construct function, I'm afraid, otherwise the web interface will only get it sent once on server startup :)

PR #1844, opened 2017-09-14
Fix issues with Open/Closed translated too early.

dracos:
It was possible for `_hardcoded_states` to be reached with no locale
selected (e.g. `bin/update-all-reports`), causing an error trying to
translate Open/Closed. But translating those there wasn't correct,
because that would then be cached and returned even if a different
language was being used. Caching was ignored in testing, which did
not help.

We no longer translate the state names in their objects, only upon
display at the point that we will know the locale. We can't simply
return the translation of Open/Closed because there may be entries
in the translation table as well.

[Had 1 test failure originally on last test showing "Open" instead; can't see how that could happen?]
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/State.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/State.pm
@@ -7,8 +7,9 @@ use Memcached;
 
 sub _hardcoded_states {
     my $rs = shift;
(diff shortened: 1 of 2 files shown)


dracos:
Merged as dc7613329 (removed bit of diff leaked into commit message).

PR #1845, opened 2017-09-15
improve report status filter default message

struan:
Partly to restore a more meaningful default message to the Oxfordshire cobrand for the default status of the report status filter, also to make the all option work better for body users.

https://github.com/mysociety/fixmystreetforcouncils/issues/227

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -539,16 +539,19 @@ sub stash_report_filter_status : Private {
         my $s = FixMyStreet::DB::Result::Problem->open_states();
         %filter_problem_states = (%filter_problem_states, %$s);
         $filter_status{open} = 1;
+        $filter_status{$_} = 1 for keys %$s;
     }
     if ($status{closed}) {
         my $s = FixMyStreet::DB::Result::Problem->closed_states();
         %filter_problem_states = (%filter_problem_states, %$s);
         $filter_status{closed} = 1;
+        $filter_status{$_} = 1 for keys %$s;
     }
     if ($status{fixed}) {
         my $s = FixMyStreet::DB::Result::Problem->fixed_states();
         %filter_problem_states = (%filter_problem_states, %$s);
         $filter_status{fixed} = 1;
+        $filter_status{$_} = 1 for keys %$s;
     }
 
     if ($status{all}) {
--- a/templates/web/base/reports/_list-filters.html
+++ b/templates/web/base/reports/_list-filters.html
@@ -1,10 +1,18 @@
 [% select_status = BLOCK %]
     <select class="form-control js-multiple" name="status" id="statuses" multiple
-      data-all="[% loc('All') %]" data-all-options='["open","closed","fixed"]'
-    [%~ IF c.cobrand.on_map_default_status == 'open' %]
-      data-none="[% loc('Unfixed reports') %]"
+    data-all="[% loc('All') %]"
+    [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+    [% options = []; FOR group IN filter_states; FOR state IN group.1; NEXT IF state == 'hidden'; options.push(state); END; END %]
+    data-all-options='["[% options.join('", "') %]"]'
+    [%~ ELSE ~%]
+      [%~ IF has_fixed_state ~%]
+        data-all-options='["open","closed","fixed"]'
+      [%~ ELSE ~%]
+        data-all-options='["open","closed"]'
+      [%~ END ~%]
     [%~ END ~%]
-      >
+    [% INCLUDE 'reports/_status_filter_options.html' %]
+    >
       [% IF c.user_exists AND c.user.has_body_permission_to('planned_reports') AND !shortlist %]
         <option value="shortlisted"[% ' selected' IF filter_status.shortlisted %]>[% loc('Shortlisted') %]</option>
         <option value="unshortlisted"[% ' selected' IF filter_status.unshortlisted %]>[% loc('Unshortlisted') %]</option>
--- a/templates/web/base/reports/_status_filter_options.html
+++ b/templates/web/base/reports/_status_filter_options.html
@@ -1,0 +1,3 @@
+    [%~ IF c.cobrand.on_map_default_status == 'open' %]
+      data-none="[% prettify_state('confirmed') %]"
+    [%~ END ~%]
--- a/templates/web/oxfordshire/reports/_status_filter_options.html
+++ b/templates/web/oxfordshire/reports/_status_filter_options.html
@@ -1,0 +1,7 @@
+    [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+      data-none="All unresolved"
+      data-extra="All unresolved"
+      data-extra-options='["confirmed","investigating","action scheduled"]'
+    [% ELSE %]
+      data-none="Open"
+    [%~ END ~%]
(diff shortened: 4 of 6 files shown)


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -160,6 +160,8 @@ sub open311_pre_send {
 
 sub on_map_default_status { return 'open'; }
 
+sub on_map_default_body_user_status { return ('confirmed', 'investigating', 'action scheduled'); }
```
dracos: Rather than having a separate `on_map_default_body_user_status` function, would it be easier to in `Reports.pm` to set `$filter_status{$state} = 1` for each open state if $status{open} is set (and similar)? Then all the relevant options would be selected in the form, whether it was body user or not.

t/app/controller/reports.t
```diff
@@ -318,6 +319,40 @@ subtest "it allows body users to filter by subtypes" => sub {
...
+        my ($investigating_problem) = $mech->create_problems_for_body(1, $body_edin_id, 'Investigating report');
+        my ($scheduled_problem) = $mech->create_problems_for_body(1, $body_edin_id, 'A Scheduled report');
+        my ($in_progress_problem) = $mech->create_problems_for_body(1, $body_edin_id, 'In progress report');
+
+        $investigating_problem->update({ state => 'investigating' });
+        $scheduled_problem->update({ state => 'action scheduled' });
+        $in_progress_problem->update({ state => 'in progress' });
+
+        $mech->get_ok('/reports/City+of+Edinburgh+Council');
+        $mech->content_contains('<option value="investigating" selected>Investigating</option>');
```
dracos: Should this also check confirmed is selected?

templates/web/base/reports/_status_filter_options.html
```diff
@@ -0,0 +1,9 @@
+    [%~ IF c.cobrand.on_map_default_status == 'open' %]
+      data-none="[% loc('Open') %]"
```
dracos: I guess this should be `prettify_state('confirmed')` to match the `<option>`.

templates/web/base/reports/_status_filter_options.html
```diff
@@ -0,0 +1,9 @@
...
+    [%~ END ~%]
+    data-all="[% loc('All') %]"
+    [%~ IF c.user.is_superuser OR c.user.from_body == body ~%]
```
dracos: The conditional used to print out these options is `[% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]` so it'd be good to match.

templates/web/base/reports/_status_filter_options.html
```diff
@@ -0,0 +1,9 @@
...
+    data-all-options='["confirmed","investigating","action scheduled","closed","fixed - council","in progress", "internal referral", "not responsible","duplicate","unable to fix"]'
```
dracos: These states shouldn't be hard-coded any more. It needs to be the states in filter_states, easiest with a loop similar to that which prints out the `<option>`s.

templates/web/base/reports/_status_filter_options.html
```diff
@@ -0,0 +1,9 @@
...
+    [%~ ELSE ~%]
+    data-all-options='["open","closed","fixed"]'
```
dracos: This should only have fixed if `has_fixed_state`, same as printing out the options.

templates/web/oxfordshire/reports/_status_filter_options.html
```diff
@@ -0,0 +1,10 @@
+    [% IF c.user.is_superuser OR c.user.from_body %]
+      data-none="[% loc('All unresolved') %]"
```
dracos: No need to translate these strings.

templates/web/oxfordshire/reports/_status_filter_options.html
```diff
@@ -0,0 +1,10 @@
...
+      data-extra="[% loc('All unresolved') %]"
+      data-all-options='["confirmed","investigating","action scheduled","fixed - council","not responsible","duplicate","unable to fix"]'
```
dracos: Feels like with the above changes to not hardcode it in base, this wouldn't be necessary as it would be the right list already. So this template would only need to add the extra option, plus change data-none?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -113,15 +113,25 @@ function isR2L() {
...
+                options: allOpts
+            }];
+
+            if (extra) {
+                presets.push({name: extra, options: extraOpts});
+            }
         $this.multiSelect({
             allText: all,
-            noneText: all,
+            noneText: none,
```
dracos: Good to have this back, not sure what happened there!

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
struan: Oxfordshire though, AIUI, don't want to include 'in progress' as part of the default but that is an open state hence this.

dracos: But it won't appear, will it? The states that appear come from their `state_groups_inspect` function. So sure, `$filter_status{in progress}` will be 1, but there won't ever be a checkbox to select using that.

t/app/controller/reports.t
struan: I think this whole test goes actually if `on_map_default_body_user_status` is being removed

templates/web/base/reports/_list-filters.html
```diff
@@ -1,10 +1,14 @@
...
     <select class="form-control js-multiple" name="status" id="statuses" multiple
-      data-all="[% loc('All') %]" data-all-options='["open","closed","fixed"]'
-    [%~ IF c.cobrand.on_map_default_status == 'open' %]
-      data-none="[% loc('Unfixed reports') %]"
+    data-all="[% loc('All') %]"
+    [% IF c.user_exists AND c.user.is_superuser OR c.user.belongs_to_body(body.id) %]
+    [% options = []; FOR group IN filter_states; FOR state IN group.1; NEXT IF state == 'hidden'; options.push(state); END; END %]
+    data-all-options='["[% options.join('", "') %]"]'
+    [%~ ELSE ~%]
+    data-all-options='["open","closed","fixed"]'
```
dracos: This comment is still outstanding - fixed should only be included if `has_fixed_state`, same as printing out the options.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -529,7 +529,9 @@ sub stash_report_filter_status : Private {
     my ( $self, $c ) = @_;
 
     my @status = $c->get_param_list('status', 1);
-    @status = ($c->cobrand->on_map_default_status) unless @status;
+    unless (@status) {
```
dracos: This change is unnecessary :)

templates/web/oxfordshire/reports/_status_filter_options.html
```diff
@@ -0,0 +1,8 @@
+    [% IF c.user.is_superuser OR c.user.from_body %]
+      data-none="All unresolved"
+      data-extra="All unresolved"
+      data-extra-options='["confirmed","investigating","action scheduled"]'
+    [% ELSE %]
+      data-none="Open"
+      data-all-options='["open","closed","fixed"]'
```
dracos: data-all-options is already being set by the base template.

templates/web/oxfordshire/reports/_status_filter_options.html
```diff
@@ -0,0 +1,8 @@
+    [% IF c.user.is_superuser OR c.user.from_body %]
```
dracos: This IF statement doesn't match the others.

struan:
This is ready for another look.

PR #1847, opened 2017-09-18
Create a comment with blank text when the report state is updated in admin

### LINKED ISSUE #1846: Report status changed in admin should create comment entry
As with status changes made in the inspector panel that do not have a public update a change to a report's status made in the admin should create an entry in the comment table.

This is partly so we can track things for stats purposes and also so they show up on the site.

struan:
Fixes #1846

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -913,7 +913,7 @@ sub report_edit : Path('report_edit') : Args(1) {
         }
         $problem->set_inflated_columns(\%columns);
 
-        $c->forward( '/admin/report_edit_category', [ $problem ] );
+        $c->forward( '/admin/report_edit_category', [ $problem, $problem->state ne $old_state ] );
         $c->forward('update_user', [ $problem ]);
 
         # Deal with photos
@@ -935,6 +935,27 @@ sub report_edit : Path('report_edit') : Args(1) {
 
         if ( $problem->state ne $old_state ) {
             $c->forward( 'log_edit', [ $id, 'problem', 'state_change' ] );
+
+            my $name = _('an adminstrator');
+            my $extra = { is_superuser => 1 };
+            if ($c->user->from_body) {
+                $name = $c->user->from_body->name;
+                delete $extra->{is_superuser};
+                $extra->{is_body_user} = $c->user->from_body->id;
+            }
+            my $timestamp = \'current_timestamp';
+            $problem->add_to_comments( {
+                text => $c->stash->{update_text} || '',
+                created => $timestamp,
+                confirmed => $timestamp,
+                user_id => $c->user->id,
+                name => $name,
+                mark_fixed => 0,
+                anonymous => 0,
+                state => 'confirmed',
+                problem_state => $problem->state,
+                extra => $extra
+            } );
         }
         $c->forward( 'log_edit', [ $id, 'problem', 'edit' ] );
 
--- a/perllib/FixMyStreet/DB/Result/Comment.pm
+++ b/perllib/FixMyStreet/DB/Result/Comment.pm
@@ -229,13 +229,24 @@ sub meta_line {
 
     if ($self->anonymous or !$self->name) {
         $meta = sprintf( _( 'Posted anonymously at %s' ), Utils::prettify_dt( $self->confirmed ) )
-    } elsif ($self->user->from_body) {
+    } elsif ($self->user->from_body || $self->get_extra_metadata('is_body_user') || $self->get_extra_metadata('is_superuser') ) {
         my $user_name = FixMyStreet::Template::html_filter($self->user->name);
-        my $body = $self->user->body;
-        if ($body eq 'Bromley Council') {
-            $body = "$body <img src='/cobrands/bromley/favicon.png' alt=''>";
-        } elsif ($body eq 'Royal Borough of Greenwich') {
-            $body = "$body <img src='/cobrands/greenwich/favicon.png' alt=''>";
+        my $body;
+        if ($self->get_extra_metadata('is_superuser')) {
+            $body = _('an administrator');
+        } else {
+            # use this meta data in preference to the user's from_body setting
+            # in case they are no longer with the body, or have changed body.
+            if (my $body_id = $self->get_extra_metadata('is_body_user')) {
+                $body = FixMyStreet::App->model('DB::Body')->find({id => $body_id})->name;
+            } else {
+                $body = $self->user->body;
+            }
+            if ($body eq 'Bromley Council') {
+                $body = "$body <img src='/cobrands/bromley/favicon.png' alt=''>";
+            } elsif ($body eq 'Royal Borough of Greenwich') {
+                $body = "$body <img src='/cobrands/greenwich/favicon.png' alt=''>";
+            }
         }
         my $can_view_contribute = $c->user_exists && $c->user->has_permission_to('view_body_contribute_details', $self->problem->bodies_str_ids);
         if ($self->text) {
--- a/t/app/controller/report_updates.t
+++ b/t/app/controller/report_updates.t
@@ -918,16 +918,96 @@ subtest 'check meta correct for second comment marking as reopened' => sub {
     like $update_meta->[2], qr/Open/, 'update meta says reopened';
 };
 
-subtest "check first comment with status change but no text is displayed" => sub {
-    $user->from_body( $body->id );
-    $user->update;
+subtest 'check meta correct for comment after mark_fixed with not problem_state' => sub {
+    $report->comments->delete;
+    my $comment = FixMyStreet::App->model('DB::Comment')->create(
+        {
+            user          => $user,
+            problem_id    => $report->id,
+            text          => 'update text',
+            confirmed     => DateTime->now( time_zone => 'local'),
+            problem_state => '',
+            anonymous     => 0,
+            mark_open     => 0,
+            mark_fixed    => 1,
+            state         => 'confirmed',
+        }
+    );
+
+    $mech->get_ok( "/report/" . $report->id );
+    my $update_meta = $mech->extract_update_metas;
+    like $update_meta->[0], qr/fixed/i, 'update meta says fixed';
+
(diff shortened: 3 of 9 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -903,6 +903,28 @@ sub report_edit : Path('report_edit') : Args(1) {
 
         if ( $problem->state ne $old_state ) {
             $c->forward( 'log_edit', [ $id, 'problem', 'state_change' ] );
+
+            unless (
+              $problem->state eq 'unconfirmed' || $old_state eq 'unconfirmed' ||
+              $problem->state eq 'hidden' || $old_state eq 'hidden'
+            ) {
+                my $name = _('admin');
```
dracos: I think I'd go for "an administrator" here (and also change the untranslated similar string in Moderate.pm to match this).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -903,6 +903,28 @@ sub report_edit : Path('report_edit') : Args(1) {
 
         if ( $problem->state ne $old_state ) {
             $c->forward( 'log_edit', [ $id, 'problem', 'state_change' ] );
+
+            unless (
+              $problem->state eq 'unconfirmed' || $old_state eq 'unconfirmed' ||
+              $problem->state eq 'hidden' || $old_state eq 'hidden'
```
dracos: If the old state was hidden, do we want an update to record it's no longer hidden?
Perhaps even if new state is hidden too, just so it's all recorded (and if it's unhidden later, you have a log of when it was hidden/unhidden then).
Not sure about unconfirmed, but I guess it even could do then as well (doubtful that's ever used anyway).

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -229,14 +229,18 @@ sub meta_line {
 
     if ($self->anonymous or !$self->name) {
         $meta = sprintf( _( 'Posted anonymously at %s' ), Utils::prettify_dt( $self->confirmed ) )
-    } elsif ($self->user->from_body) {
+    } elsif ($self->user->from_body || ( $self->name eq _('admin') && $self->user->is_superuser)) {
```
dracos: Hmm, comparing the name against a string feels a bit wrong.
Perhaps we should record (in the update's extra) if someone is a staff user or superuser at the time they make the update - this would also mean if they stop being a superuser/staff user, that information is not lost. At present I would suggest they got a different account rather than switch the status of the user, iyswim. But that's probably quite a bigger change.
So perhaps this will do for now, what do you think?

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -229,14 +229,18 @@ sub meta_line {
...
         my $user_name = FixMyStreet::Template::html_filter($self->user->name);
         my $body = $self->user->body;
         if ($body eq 'Bromley Council') {
             $body = "$body <img src='/cobrands/bromley/favicon.png' alt=''>";
         } elsif ($body eq 'Royal Borough of Greenwich') {
             $body = "$body <img src='/cobrands/greenwich/favicon.png' alt=''>";
         }
+        if ($self->name eq _('admin') && !$self->user->from_body) {
+            $c->log->debug('admin');
```
dracos: Debug line

perllib/FixMyStreet/DB/Result/Comment.pm
struan: Yeah, I was a bit unsure about this too.

perllib/FixMyStreet/App/Controller/Admin.pm
struan: I wasn't sure about exposing hiding, unhiding and confirming reports on the report page which is why this isn't here. I suppose I could record this here and change the logic on the report page, although that might be a bit fiddly. Possibly a non-public flag in the extra information is the way to go.

dracos: I think I lean towards recording all state changes for consistency and so we can always recreate the full history, and if we don't want to display ones, do that on display as you say.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -903,6 +903,32 @@ sub report_edit : Path('report_edit') : Args(1) {
...
+            ) {
+                my $name = _('an adminstrator');
+                my $extra = { is_superuser => 1 };
+                if ($c->user->from_body) {
+                    $name = $c->user->from_body->name;
+                    delete $extra->{is_superuser};
+                    $extra->{is_body_user} = $c->user->from_body->id;
```
dracos: Nice :)

templates/web/base/report/updates.html
```diff
@@ -1,5 +1,11 @@
 [% global.last_state = '' %]
+[% prev_state = '' %]
 [% FOREACH update IN updates %]
+[%- IF prev_state == 'hidden' OR prev_state == 'unconfirmed' OR update.problem_state == 'hidden' OR update.problem_state == 'unconfirmed' %]
```
dracos: If (somehow, I guess it shouldn't but I wouldn't like to promise) a report has three updates, first has problem_state `hidden`, second has blank problem_state, third has problem_state `confirmed`, then this loop will show the third update that unhid the report, when that doesn't appear to be its intention.

It feels like you should be able to use `global.last_state` here, which should be doing exactly the same job as this `prev_state`. If the logic setting global.last_state was moved up here, then that would simplify it all I think?

templates/web/base/report/updates.html
```diff
@@ -1,11 +1,11 @@
 [% global.last_state = '' %]
-[% prev_state = '' %]
 [% FOREACH update IN updates %]
-[%- IF prev_state == 'hidden' OR prev_state == 'unconfirmed' OR update.problem_state == 'hidden' OR update.problem_state == 'unconfirmed' %]
-  [%- prev_state = update.problem_state %]
+[%- IF global.last_state == 'Hidden' OR global.last_state == 'Unconfirmed' OR update.problem_state == 'hidden' OR update.problem_state == 'unconfirmed' %]
+  [%- IF update.problem_state != '' %]
```
dracos: It'd be more consistent if this did the same as `update.html` and checked the result of `problem_state_display` (not sure it makes any difference, it shouldn't, but might as well?).

[edit: Actually, then given my next comment about translating strings, perhaps `global.last_state` should be changed to use `problem_state` rather than problem state display, so that translation is not necessary. But meh]

templates/web/base/report/updates.html
```diff
@@ -1,11 +1,11 @@
 [% global.last_state = '' %]
-[% prev_state = '' %]
 [% FOREACH update IN updates %]
-[%- IF prev_state == 'hidden' OR prev_state == 'unconfirmed' OR update.problem_state == 'hidden' OR update.problem_state == 'unconfirmed' %]
-  [%- prev_state = update.problem_state %]
+[%- IF global.last_state == 'Hidden' OR global.last_state == 'Unconfirmed' OR update.problem_state == 'hidden' OR update.problem_state == 'unconfirmed' %]
```
dracos: These will need to be translated strings.

struan: 👍

templates/web/base/report/updates.html
struan: 👍

templates/web/base/report/update.html
```diff
@@ -43,9 +43,11 @@ <h4 class="static-with-rule">[% loc('Updates') %]</h4>
                     </div>
                     [% END %]
 
-                  [% SET update_state = update.problem_state_display(c) %]
-                  [% IF update_state AND update_state != global.last_state AND NOT (global.last_state == "" AND update.problem_state == 'confirmed') %]
-                    <p class="meta-2">[% loc('State changed to:') %] [% update_state %]</p>
+                  [% SET update_state = update.problem_state %]
+                  [% IF ( update_state AND update_state != global.last_state AND NOT (global.last_state == "" AND update.problem_state == 'confirmed') ) OR
+                     update.mark_fixed OR update.mark_open
```
dracos: Hmm, so if there's an update that marks fixed but doesn't set problem state (227,000 of those), that'll set last_state to the empty string. I guess easiest thing to do is change the `SET update_state` so that it sets `update_state` to 'fixed' if mark_fixed, 'confirmed' if mark_open, problem_state otherwise. Think that should do it?

struan:
Squashed and rebased.

templates/web/base/report/update.html
```diff
@@ -43,10 +43,14 @@ <h4 class="static-with-rule">[% loc('Updates') %]</h4>
...
+                  %]
+                    <p class="meta-2">[% loc('State changed to:') %] [% update.problem_state_display(c) %]</p>
                     [%- global.last_state = update_state %]
+                    [%- IF update_state == "" AND update.mark_fixed %][% global.last_state = 'user - fixed' %][% END %]
```
dracos: This should be `fixed - user`

t/app/controller/report_updates.t
```diff
@@ -918,16 +918,96 @@ subtest 'check meta correct for second comment marking as reopened' => sub {
...
+    my $update_meta = $mech->extract_update_metas;
+    like $update_meta->[0], qr/fixed/i, 'update meta says fixed';
+
+    $comment = FixMyStreet::App->model('DB::Comment')->create(
+        {
+            user          => $user,
+            problem_id    => $report->id,
+            text          => 'update text',
+            confirmed     => DateTime->now( time_zone => 'local' ) + DateTime::Duration->new( minutes => 1 ),
+            problem_state => 'user - fixed',
```
dracos: fixed - user.

templates/web/base/report/update.html
dracos: Just to note in case this ever bites us - it's theoretically possible for one update to have `fixed - user`, the next to have `fixed - council`, and then it would say a state change. There are a very small number of such cases, looks to be due to rare condition such as user and council leaving updates at the same time, so I don't think this is worth doing anything about, but perhaps that is why it previously used the `problem_state_display` for comparison and not `problem_state`.

PR #1848, opened 2017-09-19
Maximum length for inspector detailed information 

struan:
Add an optional maximum length, configured in the cobrand, for the detailed information field in the inspector panel.

Fixes mysociety/fixmystreetforcouncils#228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
         - Area summary statistics page in admin #1834
     - Bugfixes
         - Shortlist menu item always remains a link #1855
+    - Admin improvements:
+      - Character length limit can be placed on report detailed information #1848
 
 * v2.2 (13th September 2017)
     - New features:
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -326,6 +326,8 @@ sub inspect : Private {
         $c->stash->{has_default_priority} = scalar( grep { $_->is_default } $problem->response_priorities );
     }
 
+    $c->stash->{max_detailed_info_length} = $c->cobrand->max_detailed_info_length;
+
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Releases
 
 * Unreleased
+    - Admin improvements:
+      - Character length limit can be placed on report detailed
```
dracos: Think this'd be okay on one line :)

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -335,8 +337,21 @@ sub inspect : Private {
...
-                $problem->set_extra_metadata( $_ => $c->get_param($_) );
+            $problem->set_extra_metadata( traffic_information => $c->get_param('traffic_information') );
+
+            if ( my $info = $c->get_param('detailed_information') ) {
+                $problem->set_extra_metadata( detailed_information => $info );
+                if ($c->cobrand->max_detailed_info_length &&
+                    length($info) > $c->cobrand->max_detailed_info_length
+                ) {
+                    $valid = 0;
+                    $c->stash->{errors} ||= [];
```
dracos: Doesn't matte, and I see this is same as elsewhere, but just to note this line isn't needed as it will be auto-vivified by the next line.

web/cobrands/fixmystreet/staff.js
```diff
@@ -299,6 +299,21 @@ $.extend(fixmystreet.set_up, {
         toggle_public_update();
     });
 
+    $('#detailed_information').on('keyup', function() {
+        var $this = $(this),
+        counter = $('#detailed_information_length');
+        if (!$this.data('max-length')) {
```
dracos: This could be done outside the function, so the keyup is only set up if the data is present.

PR #1850, opened 2017-09-20
show nearest address for report in inspector panel

struan:
If we've stored the nearest address then display that at the top of the
inspector panel.

Fixes mysociety/fixmystreetforcouncils#230

### DIFF
--- a/templates/web/base/report/_inspect.html
+++ b/templates/web/base/report/_inspect.html
@@ -34,6 +34,12 @@
           <input type="hidden" name="longitude" value="[% problem.longitude %]">
           <input type="hidden" name="latitude" value="[% problem.latitude %]">
         </p>
+        [% IF problem.nearest_address() %]
+        <p>
+            <strong>[% loc('Nearest calculated address:') %]</strong>
+            [% problem.nearest_address() %]
+        </p>
+        [% END %]
         <p>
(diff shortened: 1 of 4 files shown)


templates/web/base/report/_inspect.html
```diff
@@ -34,6 +34,12 @@ <h2 class="inspect-form-heading">[% loc('Inspect report') %]</h2>
           <input type="hidden" name="longitude" value="[% problem.longitude %]">
           <input type="hidden" name="latitude" value="[% problem.latitude %]">
         </p>
+        [% IF problem.nearest_address(c) %]
```
dracos: Don't think you need to pass `c` into this, doesn't look to be used?

dracos:
(I think the patch will pass if you rebase)

templates/web/base/report/_inspect.html
```diff
@@ -34,6 +34,12 @@ <h2 class="inspect-form-heading">[% loc('Inspect report') %]</h2>
           <input type="hidden" name="longitude" value="[% problem.longitude %]">
           <input type="hidden" name="latitude" value="[% problem.latitude %]">
         </p>
+        [% IF problem.nearest_address() %]
```
dracos: Just a note you don't need the brackets in template calls, it's automatic.

PR #1852, opened 2017-09-21
Reduce filesize of sprite.png

zarino:
`web/cobrands/fixmystreet/images/sprite.png` is loaded by most cobrands. Optimising it reduced the filesize from 97 KB to 36 KB.


zarino:
@dracos Changelog entry added!

PR #1853, opened 2017-09-22
move area stats page problems by status calc into database

struan:
Doing this calculation in code turns out to be much too slow.

As part of this also fix an issue where if a report changed state last
month but had a further comment this month that was counted as being
a state change this month.

Also tweak the 'last month' start date to be midnight so the stats don't
change throughout the day.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
@@ -207,7 +207,12 @@ sub stats : Private {
             join     => 'problem'
         }
     )->first;
-    $c->stash->{average} = int( ($comments->get_column('time')||0)/ 60 / 60 / 24 + 0.5 );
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -145,21 +147,30 @@ sub stats : Private {
...
+            group_by => [ 'problem_state' ],
+            select   => [ 'problem_state', { count => 'me.id' } ],
+            as       => [ qw/problem_state state_count/ ],
         }
     );
 
-    # you can have multiple comments with the same problem state so need to only count
-    # one instance.
-    my %state_seen = ();
+    my $comments2 = $c->model('DB::Comment')->to_body(
```
dracos: What is $comments2 for?

PR #1855, opened 2017-09-25
make sure shortlist menu item is link on report pages

struan:
Because we move to report pages using ajax and pushstate rather than links the shortlist menu item remains inactive because the page doesn't reload. Fix this by making the Shortlist menu item a link all the time.

Fixes mysociety/fixmystreetforcouncils#233


dracos:
As an easier method that avoids dealing with pushState, how about just making Shortlist always a link? :)

Issue #1857, opened 2017-09-27
Two update-leaving bugs in the category change feature, plus blank updates in email


jonkri:
I have a question about the (terrific) new feature letting us body users change the category of a problem.

When trying out the feature, I posted an empty update like this ("Save with a public update"):

![change category 1](https://user-images.githubusercontent.com/777056/30926866-d7b93564-a3b6-11e7-93c3-1de484f94059.png)

When I submitted the form, I got an error message letting me know that I should provide a public update, which I guess is appropriate. :)

![change category 2](https://user-images.githubusercontent.com/777056/30926873-e3349b7c-a3b6-11e7-9333-b6817dfc4c55.png)

However, I could, when I refreshed the page, see an update signaling that the category had changed, even though the category change had not actually been carried out (the category remained "Park").

![change category 3](https://user-images.githubusercontent.com/777056/30926886-e97bca64-a3b6-11e7-81f2-dff6745142c7.png)

Have I made a mistake setting this up, or am I misunderstanding something? Or is this perhaps a bug?

When I unticked the "Save with a public update" checkbox, the textarea for the update disappeared, and I could make the category change by submitting the form. However, this time around, I noticed two updates: one regarding the category update, and one empty update.

![change category 4](https://user-images.githubusercontent.com/777056/30926889-ef17d210-a3b6-11e7-8e6b-00f72eaf82f9.png)

Could it be that the last (empty) update is redundant, and could have been left out?

Thanks! :)

dracos:
"Have I made a mistake setting this up, or am I misunderstanding something? Or is this perhaps a bug?" - certainly looks to be a bug, it will make an update for a category change even if the category change then doesn't go ahead due to other errors.

And yes to your other comment as well, it looks like it leaves two updates if you submit that form *and* change the category, which seems a bit superfluous!

Thanks for pointing these out.

jonkri:
Oh, I just noticed one more thing!

In the notification e-mail that goes out, the empty updates are included, like so:

![change category 5](https://user-images.githubusercontent.com/777056/30927187-f9a66196-a3b7-11e7-8080-12680df59484.png)

dracos:
Oh dear! I swear I remember that being spotted and dealt with, but I can't find anything doing that now.

jonkri:
Oh. :) We're using the commit of the 2.2 tag, by the way!

struan:
Both these are now fixed as of 31b8008

PR #1864, opened 2017-10-02
Fix category updating issues with inspect form

struan:
This changes the admin code that updates problem categories to do two things:
 * Only update the problem category if the updates are valid
 * Only create a single comment for the changes rather than one for the category change and another for other changes.

For #1857

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -424,7 +424,11 @@ sub inspect : Private {
         }
 
         if ($permissions->{report_inspect} || $permissions->{report_edit_category}) {
-            $c->forward( '/admin/report_edit_category', [ $problem ] );
+            $c->forward( '/admin/report_edit_category', [ $problem, 1 ] );
+
+            if ($c->stash->{update_text}) {
+                $update_text .= "\n\n" . $c->stash->{update_text};
+            }
 
             # The new category might require extra metadata (e.g. pothole size), so
             # we need to update the problem with the new values.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
     - Bugfixes
         - Shortlist menu item always remains a link #1855
         - Fix encoded entities in RSS output. #1859
+        - Only save category changes if staff user update valid #1857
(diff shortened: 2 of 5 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -423,24 +423,33 @@ sub inspect : Private {
...
-            # The new category might require extra metadata (e.g. pothole size), so
-            # we need to update the problem with the new values.
-            my $param_prefix = lc $problem->category;
-            $param_prefix =~ s/[^a-z]//g;
-            $param_prefix = "category_" . $param_prefix . "_";
-            my @contacts = grep { $_->category eq $problem->category } @{$c->stash->{contacts}};
-            $c->forward('/report/new/set_report_extras', [ \@contacts, $param_prefix ]);
-        }
 
         # Updating priority must come after category, in case category has changed (and so might have priorities)
```
dracos: This comment seems quite relevant, as you've moved the the category change after it :) I guess this should move down under the bit you've moved.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -423,24 +423,33 @@ sub inspect : Private {
...
         if ($c->get_param('priority') && ($permissions->{report_inspect} || $permissions->{report_edit_priority})) {
             $problem->response_priority( $problem->response_priorities->find({ id => $c->get_param('priority') }) );
         }
 
+        if ($permissions->{report_inspect} || $permissions->{report_edit_category}) {
+            $c->stash->{updated_category} = $c->get_param('category');
```
dracos: To check, this is for redisplaying in the form if there's an error? Could this just do `$problem->category($c->get_param('category'))` so no template change needed? It won't get committed if there's a problem.

perllib/FixMyStreet/App/Controller/Report.pm
struan: I confess I don't really understand what this comment means as nothing in here seems dependant on the report category.

dracos: `$problem->response_priorities` uses `$self->category` to locate the appropriate priorities.

PR #1865, opened 2017-10-02
In RSS, only escape the necessary five/ctrl chars.

### LINKED ISSUE #1859: Error in RSS feed - malformed XML
Bug report via user support email: 

there’s a new parsing error for my own local postcode. See here:
https://www.fixmystreet.com/rss/pc/SG36JW
 
Putting the results through an XML validator says:
Line 105: The entity "rsquo" was referenced, but not declared.
 
The offending line: <title>Door dumped in bush near footpath. Council cut bush around it 2 week&rsquo;
 
XML doesn’t like HTML entity codes, they need to be escaped or converted.
 
I’m wondering if your RSS generator has always had this issue or if it was introduced in your Sep update of your code?
 
I would be grateful if it can be fixed as I’m sure it will impact feeds for other postcodes where similar text is encountered.

dracos:
By default encode_entities uses named entities, which don't exist
by default in XML. It seemed nicer to switch to only escaping the
minimum rather than switching to numeric entities.

Fixes #1859

### DIFF
--- a/t/app/controller/rss.t
+++ b/t/app/controller/rss.t
@@ -1,3 +1,4 @@
+use open ':std', ':locale';
 use FixMyStreet::TestMech;
 use FixMyStreet::App;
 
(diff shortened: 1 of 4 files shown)


t/app/controller/rss.t
```diff
@@ -44,7 +45,7 @@ FixMyStreet::override_config {
 }, sub {
     $mech->get_ok("/rss/pc/EH11BB/2");
 };
-$mech->content_contains( "Testing, 10th October" );
+$mech->content_contains( "Test’ing ☃, 10th October" );
```
struan: Should possibly check that it encodes <> etc here too for completeness sake

dracos:
Will have to ignore codecov as I'm not doing EastHants coverage for this.

PR #1866, opened 2017-10-02
Avoid blank updates in alert emails

struan:
This both ignores blank updates for email alerts and also includes useful text in the update if the status of the problem was changed.

For #1857

### DIFF
--- a/perllib/FixMyStreet/Script/Alerts.pm
+++ b/perllib/FixMyStreet/Script/Alerts.pm
@@ -39,6 +39,7 @@ sub send() {
                    $item_table.name as item_name, $item_table.anonymous as item_anonymous,
                    $item_table.confirmed as item_confirmed,
                    $item_table.photo as item_photo,
+                   $item_table.problem_state as item_problem_state,
                    $head_table.*
             from alert, $item_table, $head_table
                 where alert.parameter::integer = $head_table.id
@@ -65,6 +66,7 @@ sub send() {
         $query = FixMyStreet::DB->schema->storage->dbh->prepare($query);
         $query->execute();
         my $last_alert_id;
+        my $last_problem_state = '';
         my %data = ( template => $alert_type->template, data => [], schema => $schema );
         while (my $row = $query->fetchrow_hashref) {
 
@@ -87,6 +89,7 @@ sub send() {
                 parameter => $row->{item_id},
             } );
             if ($last_alert_id && $last_alert_id != $row->{alert_id}) {
+                $last_problem_state = '';
                 _send_aggregated_alert_email(%data);
                 %data = ( template => $alert_type->template, data => [], schema => $schema );
             }
@@ -103,6 +106,18 @@ sub send() {
             my $url = $cobrand->base_url_for_report($row);
             # this is currently only for new_updates
             if (defined($row->{item_text})) {
+                # this might throw up the odd false positive but only in cases where the
+                # state has changed and there was already update text
+                if ($row->{item_problem_state} &&
+                    !( $last_problem_state eq '' && $row->{item_problem_state} eq 'confirmed' ) &&
+                    $last_problem_state ne $row->{item_problem_state}
+                ) {
+                    my $state = FixMyStreet::DB->resultset("State")->display($row->{item_problem_state}, 1, $cobrand);
+
+                    my $update = _('State changed to:') . ' ' . $state;
+                    $row->{item_text} = $row->{item_text} ? $row->{item_text} . "\n\n" . $update :
+                                                            $update;
+                }
                 next unless $row->{item_text};
                 if ( $cobrand->moniker ne 'zurich' && $row->{alert_user_id} == $row->{user_id} ) {
                     # This is an alert to the same user who made the report - make this a login link
--- a/perllib/FixMyStreet/Script/Alerts.pm
+++ b/perllib/FixMyStreet/Script/Alerts.pm
@@ -103,6 +103,7 @@ sub send() {
             my $url = $cobrand->base_url_for_report($row);
             # this is currently only for new_updates
             if (defined($row->{item_text})) {
+                next unless $row->{item_text};
                 if ( $cobrand->moniker ne 'zurich' && $row->{alert_user_id} == $row->{user_id} ) {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -39,6 +39,7 @@ sub send() {
                    $item_table.name as item_name, $item_table.anonymous as item_anonymous,
                    $item_table.confirmed as item_confirmed,
                    $item_table.photo as item_photo,
+                   $item_table.problem_state as item_problem_state,
```
dracos: Should update the comment at line 22 (which is already out of date). Basically this assumes it's the comment table, doesn't it, while looking generic.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -103,6 +106,14 @@ sub send() {
             my $url = $cobrand->base_url_for_report($row);
             # this is currently only for new_updates
             if (defined($row->{item_text})) {
+                # this might throw up the odd false positive but only in cases where the
+                # state has changed and there was already update text
+                if ($row->{item_problem_state} && $last_problem_state ne $row->{item_problem_state}) {
+                    my $update = sprintf(_('Problem marked as %s'), $row->{item_problem_state});
```
dracos: This will need to use problem_state_display (somehow... :-/ )

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -103,6 +106,14 @@ sub send() {
             my $url = $cobrand->base_url_for_report($row);
             # this is currently only for new_updates
             if (defined($row->{item_text})) {
+                # this might throw up the odd false positive but only in cases where the
```
dracos: Won't it also output on the first update of every report? The logic should probably match how the templates do it to try and cover this.

struan: It will yes, but if it's a blank update then that means the state has changed so we should output it and if not then we'd output it anyway. The false positive is it might mention the status change when there hasn't been one which I'm not really sure how to avoid.

perllib/FixMyStreet/Script/Alerts.pm
struan: Yes, it does.

perllib/FixMyStreet/Script/Alerts.pm
dracos: Sorry I wasn't clear, I meant that yes, for every normal first update with text and no state change, it will also always say "Problem marked as confirmed" which doesn't feel right. The template has the additional conditional of ignoring a problem_state of confirmed while last_problem_state is the empty string, which would also work here unless I'm missing something?

t/app/controller/alert_new.t
```diff
@@ -547,6 +547,84 @@ subtest "Test alerts are not sent for no-text updates" => sub {
...
+    }, sub {
+        FixMyStreet::Script::Alerts::send();
+    };
+
+    $mech->email_count_is(1);
+    my $email = $mech->get_email;
+    my $body = $mech->get_text_body_from_email($email);
+    like $body, qr/The following updates have been left on this report:/, 'email is about updates to existing report';
+    like $body, qr/Staff User/, 'Update comes from correct user';
+    unlike $body, qr/marked as confired/s, 'no marked as confirmed text';
```
dracos: Typo in test is why it didn't fail :)

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -108,7 +108,10 @@ sub send() {
             if (defined($row->{item_text})) {
                 # this might throw up the odd false positive but only in cases where the
                 # state has changed and there was already update text
-                if ($row->{item_problem_state} && $last_problem_state ne $row->{item_problem_state}) {
+                if ($row->{item_problem_state} && (
+                    ( $last_problem_state eq '' && $row->{item_problem_state} eq 'confirmed' ) ||
```
dracos: I think you want to `!` this part of the `if` statement - you /don't/ want to add the line if last is blank and state is confirmed.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -112,7 +112,10 @@ sub send() {
                     ( $last_problem_state eq '' && $row->{item_problem_state} eq 'confirmed' ) ||
                     $last_problem_state ne $row->{item_problem_state}
                 ) ) {
-                    my $update = sprintf(_('Problem marked as %s'), $row->{item_problem_state});
+                    my $update = sprintf(
+                        _('Problem marked as %s'),
```
dracos: For translation ease, probably best if we go with _('State changed to:') as the template.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -112,10 +112,17 @@ sub send() {
                     ( $last_problem_state eq '' && $row->{item_problem_state} eq 'confirmed' ) ||
                     $last_problem_state ne $row->{item_problem_state}
                 ) ) {
-                    my $update = sprintf(
-                        _('Problem marked as %s'),
-                        FixMyStreet::DB->resultset("State")->display($row->{item_problem_state}, 1)
-                    );
+                    my $state = '';
+                    if ($row->{update_state} eq 'not responsible') {
```
dracos: Should we move this into resultset("State")->display so it doesn't have to be duplicated?

t/app/controller/alert_new.t
struan: 👍

perllib/FixMyStreet/Script/Alerts.pm
struan: 👍

perllib/FixMyStreet/Script/Alerts.pm
struan: 👍

perllib/FixMyStreet/Script/Alerts.pm
struan: 👍

PR #1868, opened 2017-10-05
Accept +1 geographic numbers as possibly mobile.

dracos:
We have no way of knowing whether a +1 geographic number is fixed
line or mobile, so accept them all as potentially mobile.
[skip changelog]


dracos:
Yes, there's no way of telling whether a +1 number is a landline or a mobile without actually trying to send the text to it. I don't think at this stage we need to bother trying to deal with that.

PR #1873, opened 2017-10-06
make sure duplicate id provided when marking as duplicate

struan:
It was possible to mark a report as a duplicate in the inspector panel
without providing an id of the duplicate. This prevents that.

Fixes mysociety/fixmystreetforcouncils#236

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -377,15 +377,15 @@ sub inspect : Private {
             if ( $problem->state eq 'hidden' ) {
                 $problem->get_photoset->delete_cached;
             }
-            if ( $problem->state eq 'duplicate' && $old_state ne 'duplicate' ) {
-                # If the report is being closed as duplicate, make sure the
-                # update records this.
-                $update_params{problem_state} = "duplicate";
-            }
-            if ( $problem->state ne 'duplicate' ) {
+            if ( $problem->state eq 'duplicate') {
+                if (my $duplicate_of = $c->get_param('duplicate_of')) {
+                    $problem->set_duplicate_of($duplicate_of);
+                } elsif (not $c->get_param('public_update')) {
+                    $valid = 0;
+                    push @{ $c->stash->{errors} }, _('Please provide a duplicate ID or public update for this report.');
+                }
+            } else {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -377,15 +377,20 @@ sub inspect : Private {
...
+                if (my $duplicate_of = $c->get_param('duplicate_of')) {
+                    $problem->set_duplicate_of($duplicate_of);
+                    if ( $old_state ne 'duplicate' ) {
+                        # If the report is being closed as duplicate, make sure the
+                        # update records this.
+                        $update_params{problem_state} = "duplicate";
+                    }
+                } else {
+                    $valid = 0;
+                    push @{ $c->stash->{errors} }, _('Please provide a duplicate id for this report.');
```
dracos: ID should be in capitals, I think.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -377,15 +377,20 @@ sub inspect : Private {
...
-                # If the report is being closed as duplicate, make sure the
-                # update records this.
-                $update_params{problem_state} = "duplicate";
-            }
-            if ( $problem->state ne 'duplicate' ) {
+            if ( $problem->state eq 'duplicate') {
+                if (my $duplicate_of = $c->get_param('duplicate_of')) {
+                    $problem->set_duplicate_of($duplicate_of);
+                    if ( $old_state ne 'duplicate' ) {
+                        # If the report is being closed as duplicate, make sure the
```
dracos: Is this now not covered by line 399, which always sets it.

struan:
Yeah, the "do we want to enforce it" question occurred to me after I wrote this. However I think if you mark it as a duplicate there needs to be something that indicates what it's a duplicate of otherwise it's not very useful information. How does someone else later find the other record.

Could enforce either an ID or a note?

dracos:
Yeah, mandating a public update would make sense to me.

perllib/FixMyStreet/App/Controller/Report.pm
struan: 👍

struan:
Oops, now capitalised.

PR #1876, opened 2017-10-10
Some text messaging tweaks

dracos:
Spot any sync errors when sending the text and inform the user, and add the option of using a Twilio Messaging Service. [skip changelog]

### DIFF
--- a/templates/web/base/auth/change_phone.html
+++ b/templates/web/base/auth/change_phone.html
@@ -25,14 +25,8 @@ END
     <input type="hidden" name="token" value="[% csrf_token %]">
 
     <fieldset>
-    [% IF username_error;
-        errors = {
-            nonmobile = loc('Please enter a mobile number'),
-            missing_phone = loc('Please enter your phone number'),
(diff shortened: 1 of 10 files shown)


templates/web/base/auth/change_phone.html
```diff
@@ -28,6 +28,7 @@
     [% IF username_error;
         errors = {
             nonmobile = loc('Please enter a mobile number'),
+            sms_failed = tprintf('Sending a confirmation text failed: "%s"', sms_error),
```
struan: Should this, and the one below not be translated?

templates/web/base/auth/general.html
```diff
@@ -42,6 +42,7 @@
 
                 errors = {
                     nonmobile = loc('Please enter a mobile number'),
+                    sms_failed = tprintf('Sending a confirmation text failed: "%s"', sms_error),
```
struan: And thinking a bit more perhaps extract this and the identical one above into a partial so we're only defining these in one place?

PR #1879, opened 2017-10-18
Some small permission issues and fixups

### LINKED ISSUE #1878: Should staff users be able to hide reports?
In v1.8.2, staff users can't seem to hide reports. However, when trying out v2.2, we noticed that the "hidden" state is now seems to be available in the dropdown menu for body users.

Is this intentional? If so, is there a way to deactivate it, in case we don't want this change?

I think we want only admin users to be able to hide reports on the site.

Thank you!

dracos:
Fix #1878 
[Oxfordshire] Fix https://github.com/mysociety/fixmystreetforcouncils/issues/244
[Bromley] Removes unnecessary override
Plus bit of tidy and a dashboard fix.


perllib/FixMyStreet/Integrations/ExorRDI.pm
```diff
@@ -116,13 +119,15 @@ sub construct {
                 $location .= " Nearest postcode: $closest_address->{postcode}{postcode}." if $closest_address->{postcode};
             }
 
-            my $description = sprintf("%s %s", $report->external_id || "", $report->get_extra_metadata('detailed_information') || "");
+            my $traffic_information = $report->get_extra_metadata('traffic_information') || 'none';
+            my $description = sprintf("%s %s %s %s",
+                $report->external_id || "",
+                $initials,
+                'TM' . ($tm_lookup{$traffic_information} || '0'),
+                $report->get_extra_metadata('detailed_information') || "");
```
davea: There might be a few inspections done before this is deployed that have a 172-length `detailed_information`, which should probably be handled here.

PR #1880, opened 2017-10-18
Add per-body reports summary pages

dracos:
Firstly this updates `update-all-reports` to allow it to output the same data used for `/reports` restricted to a particular body, or output for each body. Then it updates the front end to display this data (with ward selection rather than body selection; whoops, guess it shouldn't show it if the area has no children). Then lastly, for .com it restricts those pages to logged in users whose domain matches a council one, and provides a customised login page at `/about/council-dashboard` which if an email is entered for a relevant body, acts like normal and sends confirmation email, or if not, sends us an email for manual checking (probably by adding the domain if possible and sending them a confirmation email? Perhaps list shouldn't be in version control.)

This final form could be duplicated on `/pro` somewhere and post directly to here, to allow easier updating of any text around it etc.

I thought adding a hook to about pages was a nice way of doing this for a particular cobrand for a quasi-static page, but other opinions welcome. It needs some tests.

### DIFF
--- a/bin/update-all-reports
+++ b/bin/update-all-reports
@@ -16,16 +16,18 @@ BEGIN {
     require "$d/../setenv.pl";
 }
 
+use FixMyStreet::DB;
 use FixMyStreet::Script::UpdateAllReports;
-use File::Path ();
-use File::Slurp;
+use Path::Tiny;
 use Getopt::Long::Descriptive;
 use JSON::MaybeXS;
 
 my ($opt, $usage) = describe_options(
     '%c %o',
     [ 'table', "Output JSON for old table-style page." ],
-    [ 'areas', "Include area IDs in output JSON." ],
+    [ 'body=i', "Restrict results to a particular body (dashboard-style)." ],
+    [ 'all-bodies', "Generate set of results for all bodies." ],
+    [ 'areas', "Include area IDs in output JSON (table-style)." ],
     [ 'help', "print usage message and exit", { shortcircuit => 1 } ],
 );
 print($usage->text), exit if $opt->help;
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -263,4 +249,67 @@ sub stuff_by_day_or_year {
     return %out;
 }
 
+sub calculate_top_five_bodies {
+    my ($data, $rs_c) = @_;
+
+    my(@top_five_bodies);
+    $data->{top_five_bodies} = \@top_five_bodies;
```
struan: Why do this here and not just assign this at the end of the function?

bin/update-all-reports
```diff
@@ -16,29 +16,42 @@ BEGIN {
...
+        output("all-reports-dashboard-" . $body->id, $data);
+    }
 } else {
-    $data = FixMyStreet::Script::UpdateAllReports::generate_dashboard();
-    $filename = 'all-reports-dashboard.json';
+    my $body = $opt->body || "";
+    $data = FixMyStreet::Script::UpdateAllReports::generate_dashboard($body);
+    $body = "-$body" if $body;
+    output("all-reports-dashboard$body", $data);
 }
```
struan: There's a chance that this will end up with weird data if the body doesn't exist but then again if the body doesn't exist the weird data shouldn't be used so I guess that's fine. Might be nice to issue a warning though?

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -62,5 +65,96 @@ sub extra_contact_validation {
...
+        return if $body->id eq $c->stash->{body}->id;
+
+        # Matched /a/ body, redirect to its summary page
+        $c->stash->{body} = $body;
+        $c->stash->{wards} = [ { name => 'summary' } ];
+        $c->detach('/reports/redirect_body');
+    }
+
+    $c->res->redirect('/about/council-dashboard');
+}
```
struan: Should this not also check if they are flagged as belonging to the council in the database?

dracos: No, as I understand it we're letting anyone with a council email address see these pages, whether they're a staff user or not.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -155,6 +155,9 @@ sub ward : Path : Args(2) {
             $c->res->redirect($ward);
             $c->detach;
         }
+        if ($c->cobrand->moniker eq 'fixmystreet') {
+            $c->cobrand->only_council_user();
+        }
```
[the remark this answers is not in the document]
dracos: Could do, but It didn't seem like something of much use! I will rename it to be more specific, and if anyone notices this and wants to use it, can adjust then.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
struan: Sorry, wasn't very clear. I meant as an extra check in the same way we check if they are a superuser. Just on the off chance that they are a staff user but don't have a matching email for whatever reason.

dracos: Aha, right, yep, will do.

dracos:
All niggles hopefully fixuped, so can be rereviewed, will hopefully add some more tests tomorrow.

PR #1882, opened 2017-10-26
always allow problems to be removed from shortlist

struan:
If a user is trying to remove a problem from their shortlist we should
always allow it regardless of the state of the problem. Previously if a
problem wasn't displayed on the site then it could not be removed from
the shortlist which was an issue with council cobrands and reports that
had changed body.

Fixes mysociety/fixmystreetforcouncils#243

### DIFF
--- a/templates/web/base/report/_item.html
+++ b/templates/web/base/report/_item.html
@@ -2,7 +2,7 @@
 [% PROCESS 'admin/report_blocks.html' ~%]
 [% END ~%]
 
-[% IF c.user.has_permission_to('planned_reports', problem.bodies_str_ids) ~%]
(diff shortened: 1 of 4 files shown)


templates/web/base/report/_item.html
```diff
@@ -2,7 +2,7 @@
 [% PROCESS 'admin/report_blocks.html' ~%]
 [% END ~%]
 
-[% IF c.user.has_permission_to('planned_reports', problem.bodies_str_ids) ~%]
+[% IF c.user.has_permission_to('planned_reports', problem.bodies_str_ids) OR c.user.is_planned_report(problem) ~%]
```
dracos: There is also the remove button on a report's page (in `report/_main.html`).

struan: I think that's ok as it is?

dracos: Oh yes, ignore me.

PR #1883, opened 2017-10-26
cobrand option to control prefilling of report fields for inspectors

### LINKED ISSUE #1854: Appended text in problem form
When trying out FixMyStreet 2.2, we experienced some text being appended to the form fields in the problem reporting form. Here is a picture that illustrates the problem:

update_problem_fields

The problem seem to occur when logged in, from the `update_problem_fields` method in staff.js.

Is this the intended behavior (sorry - *behaviour* ;))? Have I missed a step or two somewhere? :)

Thanks!

struan:
Add a `prefill_report_fields_for_inspectors` config to cobrands,
defaulted to 0, to control the automatic pre-population of report
fields.

Also add default of 1 for UKCouncil cobrand.

Fixed #1854

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
       - Remove hidden from default staff state dropdown. #1878
       - Marking an item as a duplicate enforces providing duplicate id or
         a public update #1873
+      - Report field pre-filling for inspectors configurable #1854
 
 * v2.2 (13th September 2017)
     - New features:
--- a/templates/web/base/report/new/category.html
+++ b/templates/web/base/report/new/category.html
@@ -5,7 +5,11 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
+    <select class="form-control" name="category" id="form_category"
+    [%~ IF c.user.from_body =%]
+      data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]"
+    [%~ END ~%]
+    >
         [%~ FOREACH cat_op IN category_options ~%]
         [% cat_op_lc = cat_op.name | lower =%]
         <option value='[% cat_op.name | html %]'[% ' selected' IF report.category == cat_op.name || category_lc == cat_op_lc || (category_options.size == 2 AND loop.last) ~%]
--- a/templates/web/base/report/new/category_wrapper.html
+++ b/templates/web/base/report/new/category_wrapper.html
@@ -1,7 +1,11 @@
 <div id="form_category_row">
 [% IF js %]
     <label for="form_category">[% loc('Category') %]</label>
-    <select class="form-control" name="category" id="form_category" data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" required><option>[% loc('Loading...') %]</option></select>
+    <select class="form-control" name="category" id="form_category"
+    [%~ IF c.user.from_body =%]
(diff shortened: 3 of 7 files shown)


templates/web/base/report/new/category_wrapper.html
```diff
@@ -1,7 +1,7 @@
 <div id="form_category_row">
 [% IF js %]
     <label for="form_category">[% loc('Category') %]</label>
-    <select class="form-control" name="category" id="form_category" data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" required><option>[% loc('Loading...') %]</option></select>
+    <select class="form-control" name="category" id="form_category" data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]" required><option>[% loc('Loading...') %]</option></select>
```
dracos: Should we only output the data-role/body/prefill if they have the permission? Just so it's not bothered with for normal users?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
         - Don't cover whole map with pin loading indicator.
         - Add Expand map toggle to more mobile maps.
         - Add functionality to have per-body /reports page.
+    - Front end improvements:
```
dracos: I'd count this as an admin improvement :)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -441,6 +441,7 @@ $.extend(fixmystreet.set_up, {
 
         args.latitude = $('input[name="latitude"]').val();
         args.longitude = $('input[name="longitude"]').val();
+        args.prefill_reports = $(this).data('prefill');
```
dracos: I wonder why these aren't all just set in the args dict above?

dracos: Ah, and this should be set after the getJSON because otherwise it's passed to the server in the next line.

dracos: Lastly could put role/data in args, not sure why they're separate arguments to the hook.

struan:
Ready for a re-review.

templates/web/base/report/new/category.html
```diff
@@ -5,7 +5,11 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]' data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]">
+    <select class="form-control" name="category" id="form_category"
+    [%~ IF c.user.from_body ~%]
+      data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]"
+    [%~ END ~%]
+    required>
```
dracos: This wasn't `required` before - not sure if it should be or not, but probably shouldn't be done as an aside to this change.

templates/web/base/report/new/category.html
```diff
@@ -5,7 +5,11 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]' data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]">
+    <select class="form-control" name="category" id="form_category"
+    [%~ IF c.user.from_body ~%]
```
dracos: This will mean the output is `id="form_category"data-role=` (no space), and similar at end.

templates/web/base/report/new/category.html
```diff
@@ -5,7 +5,11 @@
     <label for='form_category' id="form_category_label">
         [%~ loc('Category') ~%]
     </label>[% =%]
-    <select class="form-control" name='category' id='form_category' data-role='[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]' data-body='[% c.user.from_body.name %]'>
+    <select class="form-control" name="category" id="form_category"
+    [%~ IF c.user.from_body ~%]
```
dracos: This will mean the output is `id="form_category"data-role=` (no space), and similar at end, and in other template.

PR #1884, opened 2017-10-27
allow admin to unban a user

### LINKED ISSUE #1881: Ability for site admins to un-ban users as well as ban them
Supposing an admin (cough, me) accidentally clicked the "Ban user" button, there's no way for them to reverse that in the admin interface, it needs developer intervention.

struan:
Add an unban button to the user edit page when a user is in the abuse
table.

Fixes #1881

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1415,6 +1415,7 @@ sub user_edit : Path('user_edit') : Args(1) {
     }
 
     $c->stash->{user} = $user;
+    $c->forward( 'check_username_for_abuse', [ $user ] );
 
     if ( $user->from_body && $c->user->has_permission_to('user_manage_permissions', $user->from_body->id) ) {
         $c->stash->{available_permissions} = $c->cobrand->available_permissions;
@@ -1428,7 +1429,10 @@ sub user_edit : Path('user_edit') : Args(1) {
             '<p><em>' . $c->flash->{status_message} . '</em></p>';
     }
 
-    if ( $c->get_param('submit') ) {
+    if ( $c->get_param('submit') and $c->get_param('unban') ) {
+        $c->forward('/auth/check_csrf_token');
+        $c->forward('unban_user', [ $user ]);
+    } elsif ( $c->get_param('submit') ) {
         $c->forward('/auth/check_csrf_token');
 
         my $edited = 0;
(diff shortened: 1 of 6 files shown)


templates/web/base/admin/user-form.html
```diff
@@ -33,6 +33,12 @@
         <input type="hidden" name="phone_verified" value="0">
       [% END %]
 
+      [% IF username_in_abuse %]
+      <li>
+            <p class="error">User is in Abuse table <input name="unban" type="submit" value="Unban"></p>
```
dracos: Needs translation of both strings here (and if you used "User in abuse table" and tweaked that string to not have the brackets, it could share that translation).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1870,6 +1874,32 @@ sub ban_user : Private {
...
+        my $abuse = $c->model('DB::Abuse')->find({ email => $user->phone });
+        if ( $abuse ) {
+            $abuse->delete;
+            $c->stash->{status_message} = _('user removed from abuse list');
+        } else {
+            $c->stash->{status_message} = _('user not in abuse list');
+        }
+        $c->stash->{username_in_abuse} = 0;
+    }
+    return 1;
```
dracos: Given you removed an unused return 1 further down, don't think this one is used.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1870,6 +1874,32 @@ sub ban_user : Private {
...
+        }
+        $c->stash->{username_in_abuse} = 0;
+    }
+    if ($user->phone_verified && $user->phone) {
+        my $abuse = $c->model('DB::Abuse')->find({ email => $user->phone });
+        if ( $abuse ) {
+            $abuse->delete;
+            $c->stash->{status_message} = _('user removed from abuse list');
+        } else {
+            $c->stash->{status_message} = _('user not in abuse list');
```
dracos: If they have both verified email and phone, and email was in abuse and phone wasn't, the message output will be "user not in abuse list" instead of "user removed". Given the button is only shown if they're in the list, probably can just set status_message if something is removed, and not have a message in the case it isn't.

dracos: You could also probably combine these two parts with a search on `email` being either email or phone (if verified), then calling delete on that resultset?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1870,6 +1874,28 @@ sub ban_user : Private {
...
+
+    my @username;
+    if ($user->email_verified && $user->email) {
+        push @username, $user->email;
+    }
+    if ($user->phone_verified && $user->phone) {
+        push @username, $user->phone;
+    }
+    if (@username) {
+        my $abuse = $c->model('DB::Abuse')->find({ email => \@username });
```
dracos: Think you want `search` rather than `find` here, in case there are two results.

PR #1885, opened 2017-10-27
Add date range options to council dashboard

struan:


templates/web/base/dashboard/index.html
```diff
@@ -143,6 +143,14 @@
       [% END %]
     [% END %]
 </select>
+</p>
+<p>
+<label for="start_date">Start Date</label><input name="start_date" type="date" value="[% start_date %]">
```
dracos: This and end_date need escaping, otherwise it's just reflecting user input.

dracos: While I'm here, there's also a superfluous `</select>` above the Report state field.

zarino:
Ok @dracos @struan, it looks like this now:

screen shot 2017-11-13 at 16 48 00

It saves a bit of space, and looks more consistent with the filters further up the page.

Issue #1890, opened 2017-11-01
[fixmystreet.com] Enable Facebook login


davea:
We have the code, we have a FB app, we should join the dots and allow login via Facebook on fixmystreet.com.

It'd be a good idea for this to be off by default for council cobrands, but make it easy for individual cobrands to opt-in if they want.

PR #1891, opened 2017-11-01
[fixmystreet.com] Disable social login on council cobrands

davea:
We want to enable FB login on fixmystreet.com, but not the various UK council cobrands also active on that instance. This PR allows individual cobrands to opt-out of social login, and does so for our council cobrands.

For #1890.

### DIFF
--- a/t/cobrand/councils.t
+++ b/t/cobrand/councils.t
@@ -1,6 +1,15 @@
 use FixMyStreet::TestMech;
 my $mech = FixMyStreet::TestMech->new;
 
+my $oxon = $mech->create_body_ok(2237, 'Oxfordshire County Council' );
+my $contact = $mech->create_contact_ok( body_id => $oxon->id, category => 'Cows', email => 'cows@example.net' );
+
+my ($report) = $mech->create_problems_for_body(1, $oxon->id, 'Test', {
+    category => 'Cows', cobrand => 'fixmystreet',
+});
+my $report_id = $report->id;
+
+
 foreach my $council (qw/oxfordshire bromley/) {
     FixMyStreet::override_config {
         ALLOWED_COBRANDS => [ $council ],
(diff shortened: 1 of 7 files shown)


t/cobrand/councils.t
```diff
@@ -11,4 +22,31 @@ foreach my $council (qw/oxfordshire bromley/) {
...
 }
 
+
+foreach my $test (
+    { cobrand => 'fixmystreet', social => 1 },
+    { cobrand => 'bromley', social => 0 },
+) {
+
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => [ { $test->{cobrand} => '.' } ],
```
dracos: This is fine, but note that it could be `ALLOWED_COBRANDS => [ $test->{cobrand} ]` :)

t/cobrand/councils.t
```diff
@@ -1,6 +1,17 @@
 use FixMyStreet::TestMech;
 my $mech = FixMyStreet::TestMech->new;
 
+my $oxon = $mech->create_body_ok(2237, 'Oxfordshire County Council' );
+my $contact = $mech->create_contact_ok( body_id => $oxon->id, category => 'Cows', email => 'cows@example.net' );
+
+my ($report) = $mech->create_problems_for_body(1, $oxon->id, 'Test', {
+    category => 'Cows', cobrand => 'fixmystreet', areas => ',2237,2420',
```
dracos: Technically areas needs a comma at the end, but actually I don't think you need areas at all, or a particular lat/lon?

templates/web/base/auth/general.html
```diff
@@ -16,7 +16,8 @@
 
         <input type="hidden" name="r" value="[% c.req.params.r | html %]">
 
-[% IF NOT oauth_need_email AND (c.config.FACEBOOK_APP_ID OR c.config.TWITTER_KEY) %]
+[% SET social_auth_enabled = (c.config.FACEBOOK_APP_ID OR c.config.TWITTER_KEY) AND NOT c.cobrand.call_hook("social_auth_disabled") %]
```
dracos: Again, fine, but as it's used three times in three templates, should we factor them as one function (not sure best place, Default.pm and call it as c.cobrand.social_auth_enabled, and have it check the config variables and call call_hook...? Other suggestions welcome).

PR #1892, opened 2017-11-01
Only allow one multiselect to be open at a time, per form.

davea:
dropdowns

Fixes mysociety/fixmystreetforcouncils#247.

[skip changelog]


zarino:
@davea – I’ve made the changes I discussed, and in lieu of a better requirement solution for the plugin, I‘ve at least moved it to `/web/vendor` so it’s a bit more obvious that it’s not technically part of fixmystreet.

Seems I can’t request a review off you (because you’re the PR author?) but if you could check this and let me know if you’re happy, then I'll get this merged.

davea:
It’s a rare treat to be able to approve my own PR, so 👍🏼🎉💯🚢

PR #1894, opened 2017-11-02
[UK] Site logo as inline SVG

### LINKED ISSUE #1887: [UK] Use SVG (with png fallback) for FMS.com logo
Considerations:

* Inline the SVG into the HTML page, to save on bandwidth
* PNG fallback for browsers that don’t support SVG
* Current logo has a black "halo" to increase contrast with asphalt background – do we want to replicate that?
* Small (base64’ed) logo on mobile and on most desktop pages; Huge logo (from sprite) on desktop homepage.

zarino:
Not only does this improve the sharpness of the logo on high-dpi screens, but it should also make the logo quicker to load on slow connections. When combined with the Critical CSS work in #1893, effectively the entire top half of the homepage will appear to load instantly.

Fixes #1887.


templates/web/fixmystreet.com/header_logo.html
```diff
@@ -0,0 +1,19 @@
+<a href="[% c.cobrand.base_url IF admin %]/" id="site-logo" title="FixMyStreet">
+<svg width="175" height="35" viewBox="0 0 175 35" xmlns="http://www.w3.org/2000/svg">
```
dracos: Presumably this `<svg>` could be `site-logo-default.svg` (which isn't used?) and then included with an `INSERT`? Then it's only in one place.

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -4,16 +4,48 @@
...
+    width: 175px;
+    height: 0;
+    padding-top: 35px;
+    overflow: hidden;
+
+    // We don't use high-dpi-background-image() here because we don't want an SVG background.
+    background: transparent url("/cobrands/fixmystreet.com/images/site-logo-default.png") 0 0 no-repeat;
+    background-size: 175px 35px;
+    @media ($high-dpi-screen) {
+        background-image: url("/cobrands/fixmystreet.com/images/site-logo-default@2x.png");
```
dracos: Is it worth having the 2x sprite fallbacks? Isn't this only for IE8-, do we care?

web/cobrands/fixmystreet.com/images/site-logo-homepage.svg
```diff
@@ -0,0 +1 @@
... .434-.134.823-.309 1.173-.652 1.254-1.528 1.254-3.756 1.254v10.793c0 1.907.699 2.126 2.273 2.126 1.792 0 2.14 1.342 2.14 2.25 0 2.647-3.014 2.821-4.72 2.821-5.896 0-6.028-3.427-6.028-6.025V26.999c-.742 0-1.443 0-2.055-.303-.218-.13-1.178-.65-1.178-2.124 0-.39.044-.778.218-1.087.61-1.298 1.922-1.298 3.015-1.298v-3.074c.045-1 .045-2.17.83-3.034.175-.178.875-.999 2.357-.999.44 0 .876.086 1.224.218 1.883.693 1.883 2.556 1.924 3.858v3.031zm-11.01-1.82c0 2.082-1.884 3.25-3.497 3.25-.918 0-1.705-.346-2.624-1.558-.217-.306-.393-.606-.655-.911-.964-1.258-2.62-2.08-5.11-2.08-3.1 0-5.248 1.128-5.248 3.119 0 .695.22 1.344.874 1.866.658.52 1.748.822 3.893 1.385 6.769 1.822 7.472 1.996 9.043 2.991 1.88 1.217 3.888 3.512 3.888 7.066 0 3.165-1.53 5.675-3.583 7.282-2.534 1.95-5.813 2.6-9.044 2.6-3.19 0-5.985-.52-8.304-1.646-4.631-2.298-4.938-5.243-4.938-6.026 0-2.124 1.75-3.423 3.412-3.423 1.442 0 2.358.909 3.364 2.385.96 1.429 2.097 3.033 6.512 3.033 1.002 0 2.094-.087 3.054-.476.352-.131 2.318-.91 2.318-3.166 0-1.212-.524-1.907-1.266-2.424-.879-.608-1.527-.784-5.64-1.822-5.065-1.344-10.745-2.863-10.745-9.146 0-5.55 4.37-9.275 12.322-9.275 7.645 0 11.973 3.9 11.973 6.975z" fill="#FFF"/></g></svg>
```
dracos: Where is this used? Also, isn't it just the same as the other SVG? ;)

dracos:
Some stats, front page, my dev site:
* master branch: 14K/4.2K. base.css 86K/23K. layout.css 22K/5K.
* this branch: 19.5K/6.8K. base.css 78K/15K. layout.css 23K/5K.
So base.css goes down 8K (yay!), but inlining the logo adds 2.6K to the gzipped HTML of every page. Is it worth considering inlining only on the front page (for first page load speed), and having an external SVG file by default for the logo, as that can be cached for all later pages?

zarino:
I have a feeling we can reduce the size of the inlined SVG dramatically by rounding to fewer decimal places, or maybe even simplifying the paths to contain fewer points. When the logo is only displayed at 175x35 pixel size, it seems crazy to have such a geometrically accurate SVG. Maybe I'll try that first, and if I can't get it smaller, we can investigate inlining _only_ on the homepage.

web/cobrands/fixmystreet.com/base.scss
zarino: I think only IE8 and some _very_ old versions of Android will see this fallback, so yep, we can drop the `@2x` here.

templates/web/fixmystreet.com/header_logo.html
zarino: As you spotted, the `web/cobrands/fixmystreet.com/site-logo-*.svg` files aren’t actually linked to from anywhere. I checked them in just to match the `{.png,@2x.png,.svg}` pattern for all the other images, and in case someone, in future, wanted to use the SVGs for something ;-) I guess we could remove them if you feel they’re gonna confuse people?

I'd only want to `INSERT` an external `.svg` file into the HTML page if we could make it clear that it isn’t just a straight SVG, but one that includes the `<switch>`, `<foreignObject>`, and IE conditional comment hacks that [make SVG inlining work](https://codepen.io/tigt/post/inline-svg-fallback-without-javascript-take-2). But I agree, it’d be nice not to just have this huge chunk of SVG in the middle of the HTML template. Maybe I should name it `site-logo-with-fallback.svg` or something.

zarino:
@dracos Rounding numbers in the SVG to a single decimal place (`svgo --precision 1`) reduced the file size of the SVG by about 40%. Unfortunately, there were no savings to be made by "simplifying" the paths (`Object > Path > Simplify` in Illustrator) without ruining the shapes.

By my reckoning (running `FIXMYSTREET_APP_DEBUG=0 script/server` and then manually gzipping the source code, since it looks like `script/server` doesn’t gzip, even when not in debug mode) the homepage now weighs in at ~5.75 KB once gzipped. Compared to ~5.0 KB for the live homepage.

Not sure whether that 0.75 KB figure will carry through to other types of pages, but if we assume it stays somewhere around 0.75-1 KB, do you feel that added to every page would be acceptable, or would you still want to inline the SVG _only_ on the homepage?

dracos:
The live front page will have different (probably more, unless your dev site has enough photo reports in it) stuff on it; you need to compare against your own dev master branch to compare like with like. With your new change, I agree this branch stats for homepage HTML are now c. 17.2K/5.6K, big improvement if it looks the same :)

So I'd say the inline SVG logo adds 1.4K to the page size. Fetching another page bears this out:
```
(1887-uk-svg-logo $%=) § curl -s http://fixmystreet.127.0.0.1.nip.io:3000/report/5|gzip -c|wc -c
    8434
(1887-uk-svg-logo $%=) § git checkout master
(master $%=) § bin/make_css 
(master $%=) § curl -s http://fixmystreet.127.0.0.1.nip.io:3000/report/5|gzip -c|wc -c
    6974
```

As to whether that's acceptable or not... dunno. 1.4K isn't much in absolute terms, but the pages are already quite small so relatively it's a 20-40% increase! Perhaps the inline logo could only be the spanner, replaced by the words when they download? ;)

PR #1896, opened 2017-11-06
do not override template when displaying body contacts as text

### LINKED ISSUE #1895: "Text only version" of body's email addresses is not being rendered as text only
screenshot from 2017-11-02 17-31-39

...etc.

struan:
move setting the body template to before `fetch_contacts` so it doesn't
override setting the text only email address template.

Fixes #1895


struan:
Merged

PR #1898, opened 2017-11-08
[Oxfordshire] Clearly mark RDI emails if sent from staging

davea:
Fixes mysociety/fixmystreetforcouncils#251

[skip changelog]

### DIFF
--- a/bin/oxfordshire/send-rdi-emails
+++ b/bin/oxfordshire/send-rdi-emails
@@ -40,6 +40,12 @@ foreach my $inspector (@inspectors) {
         user => $inspector,
         mark_as_processed => 1,
     };
+    my $email_params = {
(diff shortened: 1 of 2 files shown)


bin/oxfordshire/send-rdi-emails
```diff
@@ -40,6 +40,10 @@ foreach my $inspector (@inspectors) {
         user => $inspector,
         mark_as_processed => 1,
     };
+    my $email_params = {
+        %$params,
+        staging_site => FixMyStreet->config('STAGING_SITE'),
```
dracos: If we're having separate email params, only the dates and user are shared, don't really need mark_as_processed or inspection_date here, so could be split more. But no biggie.

PR #1902, opened 2017-11-13
Better image sprites

### LINKED ISSUE #1888: [UK] Investigate whether sprite.png is still worth including
Over time we’ve used more and more image assets on FMS.com that aren’t part of `web/cobrands/fixmystreet/images/sprite.png`.

We should investigate how much of `sprite.png` we’re actually still using, and either: ditch `sprite.png` and just use individual image files; or make an effort to include all the image assets into the sprite, perhaps with it being auto-generated and auto-referenced in the CSS, to reduce the chance of us getting into this same position in a few years’ time.

zarino:
In this PR:

- Proper handling of site logo (SVG with PNG fallback, like the fixmystreet.com cobrand has had for ages) in the default cobrand, and also FixaMinGata and FiksGataMi. … Which meant the latter two cobrands no longer needed their own `sprite.png` files.
- Collected the pan/zoom controls, `sub_map_links` icons, `#key-tools` icons, and general button icons into their own, tiny little SVG sprites, with PNG fallbacks.
- Removed a few obsolete images.

Fixes #1888.

A few stats:

* Default homepage:
   * Before - 47.38 kB of images
   * After – 14.60 kB of images
* Default map page
   * Before – 44.69 kB of images (ignoring map tiles and OSM bing logo)
   * After – 15.29 kB of images (ignoring map tiles and OSM bing logo)


zarino:
When we merge this, we’ll need to remember to also update any cobrands in mysociety/fixmystreet-commercial and mysociety/fixmystreet-international that override any of the code – such as swapping the `<input>` in `templates/web/acv/report/display_tools.html` (fixmystreet-commercial) for a `<button>`.

web/cobrands/sass/_base.scss
```diff
@@ -1449,62 +1461,57 @@ html.js #map .noscript {
...
+
+  .feed:after {
+    background-position: -12px 0;
+  }
+
+  #hide_pins_link:after {
+    background-position: -24px 0;
+  }
+
+  // TODO: Is there still a "show/hide old reports" button?
```
dracos: No, there is not, pagination got rid of it.

web/cobrands/sass/_base.scss
```diff
@@ -1451,63 +1451,58 @@ html.js #map .noscript {
...
-      background-position: flip(right, -341px) -4070px;
-    }
-    &#fms_shortlist_all {
-      padding: 0.6em 1em 0.5em 1em;
-      &:after {
-         content: "";
-         display: inline-block;
-         width: 16px;
-         height: 16px;
-         @include svg-background-image('/cobrands/fixmystreet/images/map-marker');
```
dracos: map-marker image can be removed then? Was it ever used?

web/cobrands/sass/_layout.scss
```diff
@@ -613,28 +613,58 @@ body.authpage {
...
-    background:#000 inline-image("../fixmystreet/images/triangle-grey-#{$right}.svg") 50% 50% no-repeat;
-    @include border-radius(flip(0.25em 0 0 0.25em, 0 0.25em 0.25em 0));
-    &:hover {
-      #{$left}: -1.5em;
-      //use border so we don't have to redefine the background-position
-      border-#{$right}: 0.5em solid #000;
+}
+
+#map_links_toggle {
+ display: block;
```
dracos: Spacing bit funny here.

web/cobrands/sass/_layout.scss
```diff
@@ -613,28 +613,58 @@ body.authpage {
...
+    margin-#{$left}: ((16px - 6px) / 2); // horizontally centre in 16px wide parent
+    background-size: 96px 12px;
+    @include svg-background-image('/cobrands/fixmystreet/images/map-tools');
+  }
+
+  // Expanded arrow points towards edge of window.
+  // Closed arrow points away from edge of window.
+  @if ($right == 'right') {
+
+    &:after {
```
dracos: Any reason this isn't inside the &:after above?

web/cobrands/sass/_layout.scss
```diff
@@ -678,65 +678,36 @@ body.authpage {
...
+
+    // Cancel centre alignment, if the *only child* in list.
+    &:first-child:last-child {
+      text-align: $right;
+    }
+  }
+
+  a, button {
+    font-size: 0.75em;
+    line-height: 18px; // match `body`
```
dracos: I don't understand the comment here. body's line height is (at normal font) 24px.

web/cobrands/sass/_base.scss
```diff
@@ -496,7 +496,7 @@ ul.error {
   display: block;
   width: 175px;
   height: 60px;
-  background: url($image-sprite) -8px 5px no-repeat;
+  background: transparent url('images/site-logo.png') 50% 50% no-repeat;
```
dracos: Is this the same as site-logo-default in fixmystreet.com cobrand? Should we move that to the base now?

web/cobrands/sass/_base.scss
zarino: Done!

web/cobrands/sass/_layout.scss
dracos: Dur, font-size is 0.75em.

web/cobrands/sass/_layout.scss
zarino: Good spot!

web/cobrands/sass/_base.scss
zarino: `cobrands/fixmystreet/images/site-logo.png` and `cobrands/fixmystreet.com/images/site-logo-default.png` are identical, yes.

Are you suggesting we change `.site-logo__fallback` in `cobrands/fixmystreet.com/base.scss` to use `svg-background-image("/cobrands/fixmystreet/images/site-logo");` (and then move `cobrands/fixmystreet.com/images/site-logo-default.svg` to `cobrands/fixmystreet/images/site-logo.svg` to match)? And then, since the base cobrand would have an svg version of the logo at its disposal, use that with `svg-background-image()` in `sass/_base.scss` too?

_Or_ are you suggesting we do all the fancy inline SVG stuff in the base cobrand?

dracos: I was suggesting using SVG in the base, ignoring the front page stuff (which would cover most of the .com-only stuff, I think?).
We could also do the inline stuff in base and wrap the insert in a TRY (same a I did the critical CSS) so it works if it's there, not if it isn't.

PR #1903, opened 2017-11-14
Performance improvements with JS rearranging

dracos:
This PR reduces the JavaScript loaded by the front page of fixmystreet.com (*excluding* any prefetching) from 78.6K to 4.9K (1.4K translation strings, 613b JS flag to pass to around page form, 925b geolocation, 857b prefetch polyfill, 1K 'other country' banner). All JS used there is now vanilla. The total fetched by the page will be similar if prefetching occurs, but much more now in the background:

The front page prefetches all the base JavaScript (jQuery, validation, multi-select, and the actual application JS) plus the map page JavaScript (the map itself, plus Dropzone and Fancybox). Other non-map pages load the base JavaScript but also prefetch the map JavaScript (I imagine as with the front page much of this could be dropped, I doubt many of those pages need much/any of the JavaScript). Map pages load all the JavaScript (though even there I have e.g. factored out geolocation as only front/alert/inspect pages need it).


dracos:
(Had to do some rearranging so jQuery loaded on front page if logged in staff user, as staff functions/offline use it.)

PR #1910, opened 2017-11-22
Allow cobrands to disable moderation emails

davea:
This adds a new `send_moderation_notifications` method which can be overridden
by cobrands to prevent “your report has been moderated” emails from being sent.

Fixes mysociety/fixmystreet-commercial#936


davea:
(sorry for the diff noise, the inconsistent whitespace in the changelog got the better of me)

PR #1911, opened 2017-11-22
Enable authentication by token

struan:
Adds token based authentication via either an `Authorization` header or an `access_token` parameter. Currently this is only enabled for `check_auth` and dashboard CSV export.

Also allows superusers to generate a token from `/my`.


templates/web/base/my/my.html
```diff
@@ -61,6 +61,7 @@ <h1>[% loc('Your account') %]</h1>
 
 <p class="my-account-buttons">
   <a href="/auth/change_password">[% loc('Change password') %]</a>
+  <a href="/auth/generate_token">[% loc('Generate token') %]</a>
```
dracos: We probably only want to show this if they're a superuser.

templates/web/base/auth/generate_token.html
```diff
@@ -0,0 +1,50 @@
+[%
+SET bclass = 'authpage';
+SET bclass = 'fullwidthpage' IF password_changed;
+INCLUDE 'header.html', title = loc('Change password'), bodyclass = bclass
```
dracos: This title and bits before it don't seem right :)
I'd go with fullwidthpage always I think.

templates/web/base/auth/generate_token.html
```diff
@@ -0,0 +1,50 @@
...
+
+    [% IF existing_token %]
+        <div class="form-field">
+            <strong>[% loc('Token:') %]</strong>
+            <span>[% existing_token | html %]</span>
+        </div>
+
+        <div>
+          <p>
+          [% loc('If you generate a new token the existing token will no longer work.') %]
```
dracos: I'd put this message underneath the button (as it is, it looks like it's part of the token display).

templates/web/base/auth/generate_token.html
```diff
@@ -0,0 +1,50 @@
...
+          </p>
+        </div>
+    [% END %]
+    <div class="final-submit">
+        <input name="generate_token" type="submit" class="btn" value="[% loc('Generate token') %]">
```
dracos: Now that I've said make it fullwidthpage, I woudn't bother right aligning this.

perllib/FixMyStreet/App.pm
```diff
@@ -81,6 +81,18 @@ __PACKAGE__->config(
                 user_model => 'DB::User',
             },
         },
+        access_token => {
+            use_session => 0,
+            credential => {
+                class => 'AccessToken',
+                token_field => 'extra',
+                token_lookup => { like => "%access_token,I20:TOKEN,%" },
+            },
```
struan: Might be worth a comment here that this means the token has to be 20 characters long, as much to document it somewhere.

dracos: Will do, but random_token generates 18 char long token, which is fine, updating the count now.

dracos:
Just thought - the user we are writing this for is not a superuser. Either need to extend to staff users (easiest), add a way for superusers to add tokens to other accounts (through admin?), or add the token manually.

PR #1912, opened 2017-11-23
display council contacts on council dashboard

struan:
Display a list of the email addresses used to contact the council on the
council dashboard. Does not display emails if the council has an
integration, rejects FMS reports or is currently turned off.

Fixes mysociety/fixmystreet-commercial#941

[skip changelog]

### DIFF
--- a/templates/web/base/reports/index.html
+++ b/templates/web/base/reports/index.html
@@ -139,4 +139,58 @@
     </div>
 </div>
 
+[% IF display_contacts %]
+<div class="dashboard-row">
+    <div class="dashboard-item dashboard-item--12">
+        <h2 class="dashboard-subheading">[% tprintf( loc('Where we send %s reports'), body.name ) %]</h2>
+        [% IF body.send_method == 'Refused' %]
+          <p>
+            [% tprintf( loc('%s currently does not accept reports from FixMyStreet.'), body.name) %]
+          </p>
+
+          <p>
+            [% loc('If you&rsquo;d like to discuss this then <a href="/contact">get in touch</a>.') %]
+          </p>
+        [% ELSIF body.send_method == 'Noop' %]
+          <p>
+            [% tprintf( loc('Reports are currently not being sent to %s.'), body.name ) %]
+          </p>
+        [% ELSIF body.send_method != 'Email' AND body.send_method != ''  %]
+          <p>
+              [% tprintf( loc('Reports to %s are currently sent directly into backend services.'), body.name) %]
+          </p>
+        [% ELSE %]
+        <p>
+          [% loc('We currently send all reports to the email addresses below.') %]
+        </p>
+
+        <p>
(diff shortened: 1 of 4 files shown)


templates/web/base/reports/index.html
```diff
@@ -139,4 +139,58 @@ <h2 class="dashboard-subheading">[% loc('Top 5 most used categories') %]</h2>
...
+        <p>
+          [% loc('If you would like to change either the categories or the contact emails below then <a href="/contact">get in touch</a>.') %]
+        <p>
+        <table class="dashboard-ranking-table">
+          <tr>
+            <th>[% loc('Category') %]</th>
+            <th>[% loc('Contact') %]</th>
+          </tr>
+          [% WHILE ( cat = contacts.next ) %]
+            <tr [% IF cat.state == 'deleted' %]class="is-deleted"[% END %]>
```
dracos: This doesn't do anything here. Probably shouldn't show deleted contacts, loop though `live_contacts` rather than `contacts`?

templates/web/base/reports/index.html
```diff
@@ -139,4 +139,58 @@ <h2 class="dashboard-subheading">[% loc('Top 5 most used categories') %]</h2>
...
+          [% loc('If that&rsquo;s new to you, <a href="https://www.mysociety.org/2013/01/10/open311-introduced/">take a look at our simple Open311 primer</a> to see what you need to do to get up and running in a few days.') %]
+        </p>
+
+        <p>
+          [% loc('If you would like to change either the categories or the contact emails below then <a href="/contact">get in touch</a>.') %]
+        <p>
+        <table class="dashboard-ranking-table">
+          <tr>
+            <th>[% loc('Category') %]</th>
+            <th>[% loc('Contact') %]</th>
```
dracos: It looks a bit odd with a centred heading and right-aligned email addresses. Not sure whether right-aligned heading, or left-aligning everything would be nicer, @zarino any thoughts?
image

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -157,6 +157,8 @@ sub ward : Path : Args(2) {
             $c->detach;
         }
         $c->cobrand->call_hook('council_dashboard_hook');
+        $c->forward( 'load_contacts' );
```
dracos: If this isn't .com (which is mandating login in the hook), anyone can then view these details at this page. I think this should only be a .com feature at present, so these two lines could be moved to the council_dashboard_hook and I think that would do it?

templates/web/base/reports/index.html
zarino: You can get away with centred column headings when the columns are fairly narrow. If the table here is always going to be this wide, then yep, I’d align the column headings to match the alignment of the cells below them. Which makes me ask – why are the email addresses right-aligned?

dracos: It's inheriting the same class as the other tables on that page. I guess as they're probably all at the same domain it makes scanning them a bit easier.

PR #1914, opened 2017-11-23
Fix UI bug stopping superusers making reports

### LINKED ISSUE #1913: 'Report as' UI blocks superusers from making reports
If you're logged in as a superuser with no `from_body`, then the default state of the 'Report as' dropdown stops you from entering your name when clicking the map on `/around`:

no_name

If you change the dropdown to another value and back, it does the right thing:

select_change

If you load `/report/new` directly, it works:

report_new

davea:
This checks that the user has a from_body *and* the planned_reports
permission when considering whether to default to 'yourself' or the
council in the 'report as' field.

Fixes #1913.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
         - Prevent text overflow bug on homepage stats #1722
         - Stop page jumping too far down on inspect form. #1863
         - Prevent multiple 'Expand map' links appearing. #1909
+        - Superusers without a from_body can make reports again. #1913
     - Admin improvements:
         - Character length limit can be placed on report detailed information #1848
         - Inspector panel shows nearest address if available #1850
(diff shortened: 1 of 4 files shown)


davea:
I've redone this in a way that fixes the problem with the JavaScript, instead of papering over it with logic changes in the template.

As you suggested, @dracos, ideally the template shouldn't ever output the 'as body' option for users without a body. That is what `can_contribute_as_body` in `form_user_loggedin.html` does, except then it's being overridden by the `js` in `[% IF js || can_contribute_as_body %]`. I couldn't quite determine the purpose of that `js` variable - it looks as though it's hard-coded in a template further up the stack.

dracos:
`js` is so that it is always output on an `/around` page and then it is up to the JS that's pulled in when you start reporting to decide whether to hide it or not.

dracos:
So I think it sounds like you've fixed this correctly, I mean :)

PR #1921, opened 2017-11-28
Fix back bug, from report after using list filters

### LINKED ISSUE #1920: Back button resets report filters
### Description of problem

Pressing the browser's Back button resets report filters.

### Version

Presumably at least 2.0, 2.1, 2.2; confirmed on both FixMyStreet and FixaMinGata.

### Steps to reproduce

1. Search for any location with pins with more than one state (like [here](https://www.fixmystreet.com/around?latitude=51.50642;longitude=-0.12721&js=1))
2. Use the first dropdown to select a filter, for instance, changing from *All* to *Closed*
3. Click on a report
4. Press the browsers "native" Back button (not *Back to all reports*)

### Actual results

The filters are reset. The first dropdown shows *All*.

### Expected results

The same filter configuration as appeared after Step 2 above - that is, the first dropdown showing the value *Closed*.

### How reproducible

Seemingly always, at least using Mozilla Firefox and Internet Explorer 11.

Related to #1649.

dracos:
The popstate event for a filter change assumed that the list was already
being displayed, but it was not if you were going back from a report.

The reports_list function to display the list page is idempotent (though
this commit makes it more so), so can happily be called whether the list
page is currently displayed or not.

Fixes #1920

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1073,8 +1073,12 @@ fixmystreet.display = {
   reports_list: function(reportListUrl, callback) {
     // If the report list is already in the DOM,
     // just reveal it, rather than loading new page.
-    if ($('#side').length) {
-        $('#side').show();
+    var side = document.getElementById('side');
+    if (side) {
+        if (side.style.display !== 'none') {
+            return;
+        }
+        side.style.display = '';
         $('#side-form').hide();
         // Report page may have been one or two columns, remove either
         $('#side-report').remove();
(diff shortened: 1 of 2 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1073,8 +1073,12 @@ fixmystreet.display = {
   reports_list: function(reportListUrl, callback) {
     // If the report list is already in the DOM,
     // just reveal it, rather than loading new page.
-    if ($('#side').length) {
-        $('#side').show();
+    var side = $('#side');
```
davea: Could this be `$side` to make it more obvious it's a jQuery object?

dracos: I have used `getElementById` instead :)

PR #1922, opened 2017-11-28
Nicer multi-select elements in Admin pages

### LINKED ISSUE #1589: Area select boxes in admin hard to use
The work to make multi-select boxes nice in the front end http://fixmystreet.org/2016/12/06/v2.0-multi-select/ has annoyingly made them less nice in the admin; the CSS is reducing them to one item’s height, as it is assuming all multiple selects are using this new code.
Options are:
* stop the admin multiples from having that CSS (makes them a number of lines high, better than one, so slightly better)
* turn on the tick-box plugin for them – this has issues in that the list is very long, very wide, appears underneath the footer, and does a weird page jump when you select/deselect your first item.

zarino:
Fixes #1589.

screen shot 2017-11-28 at 11 49 48


CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
         - Report field pre-filling for inspectors configurable #1854
         - Admins can now unban users #1881
         - Council dashboard has date range for report generation #1885
+        - JavaScript-enhanced `<select multiple>` elements #1589
```
dracos: "More" at the start, quite a few already were :)

templates/web/base/admin/body-form.html
```diff
@@ -75,8 +75,7 @@
     </div>
     <p>
         <label for="area_ids">[% loc('Area covered') %]</label>
-        <select class="form-control" name="area_ids" id="area_ids" multiple>
-            <option value=""> -- [% loc('Select an area') %] -- </option>
+        <select class="form-control js-multiple" name="area_ids" id="area_ids" multiple data-none="-- [% loc('Select an area') %] --">
```
dracos: Do we want to keep the '--' here, do you think?

web/cobrands/sass/_multiselect.scss
```diff
@@ -1,14 +1,15 @@
...
 }
 
 .multi-select-menu {
     position: absolute;
     left: 0;
     top: 0.8em;
-    z-index: 1;
-    float: left;
-    min-width: 100%;
+    z-index: 2;
```
dracos: Why did this go up?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -112,27 +112,46 @@ function isR2L() {
...
+                settings.noneText = settings.allText;
             }
-        $this.multiSelect({
-            allText: all,
-            noneText: none,
-            positionMenuWithin: $('#side'),
-            presets: presets
-        });
+
+            if ( $select.data('all-options') ) {
```
dracos: I think this changes the behaviour of the JS? Before there was always an `all` preset (even if there were no `all-options`, so it defaulted to `[]`) and now there isn't?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -112,27 +112,46 @@ function isR2L() {
...
+                settings.presets = [];
+                settings.presets.push({
+                    name: settings.allText,
+                    options: $select.data('all-options')
+                });
+            }
+        }
+
+        if ( $select.data('extra') && $select.data('extra-options') ) {
+            if ( typeof settings.presets === 'undefined' ) {
```
dracos: This can just be `settings.presets = settings.presets || [];` I think. Or at least `if (!settings.presets)`, no need for typeof here.

templates/web/base/admin/body-form.html
zarino: I left it in to match the other `<select>` elements in the same form.

web/cobrands/fixmystreet/fixmystreet.js
zarino: Yes, `settings.presets = settings.presets || [];` does the job. Not sure what I was thinking there!

web/cobrands/sass/_multiselect.scss
zarino: To stack above the footer. I will leave a comment in the sass file.

web/cobrands/fixmystreet/fixmystreet.js
zarino: Ah, interesting … what about if you don’t want an "all" option? (Like, I figured the `#area_ids` select in the body form didn’t need an "all" preset – maybe I'm wrong?)

dracos: I agree, it doesn't. But that doesn't have a `data-all`? I mean if something has set a data-all but no all-options, you still want a preset then, as that's what it does currently. Wouldn't affect your new ones.

zarino: Yeah, good point. Will add it back in now.

zarino:
@dracos – I’ve pushed a few fixes to those issues. Can you check they were what you were expecting?

PR #1924, opened 2017-11-28
Use auto-response ResponseTemplates for text when fetching Open311 updates

davea:
If an Open311 service request update with an empty description field is received and there is a matching ResponseTemplate for the problem's category & state (and it's marked as auto-response) then the text from that ResponseTemplate will be used for the created update.

This PR adds that functionality, as well as improving the admin UI for managing response templates and marking them as auto-responses.

Fixes mysociety/fixmystreetforcouncils#263

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1102,19 +1102,52 @@ sub template_edit : Path('templates') : Args(2) {
     } } @live_contacts;
     $c->stash->{contacts} = \@all_contacts;
 
-    if ($c->req->method eq 'POST') {
+    # bare block to use 'last' if form is invalid.
+    if ($c->req->method eq 'POST') { {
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1135,19 +1135,52 @@ sub template_edit : Path('templates') : Args(2) {
...
+                # If the new template has not specific categories (i.e. it
+                # applies to all categories) then we need to check each of those
+                # category ids for existing auto-response templates.
+                if (!scalar @check_contact_ids) {
+                    @check_contact_ids = @live_contact_ids;
+                }
+                my $query = {
+                    'auto_response' => 1,
+                    'contact.id' => [ @check_contact_ids, undef ],
+                    'me.state' => $template->state,
```
dracos: Does this code need skipping/ignoring if the new template's state isn't being set?

PR #1925, opened 2017-11-28
Consolidate summary statistics

### LINKED ISSUE #1919: Consolidate summary statistics
There is /admin/stats, /admin/areastats and /dashboard, all performing the same sort of functionality, and could usefully be consolidated in one. Proposal:
* /admin/stats has three 'special pages' it currently has, and otherwise links to /dashboard;
* /admin/areastats etc redirect to /dashboard;
* /dashboard shows data for body immediately if staff user, or summary for all if superuser (and dropdown to pick body)
* /dashboard filters: date range, category, body/ward (if on body), state (much as filters presently are there), potentially add (cobrand/service) on body page;
* Allow grouping of results by category/state/(cobrand/service)/month/(category+stagegroup);
* Shows summary table, ditch list of reports.
* Download as CSV button should work immediately on what is picked, not after you've loaded the HTML equivalent.

dracos:
This joins together admin/stats admin/areastats and dashboard, all at /dashboard.
Contains most of all of them, with unused/non-working bits removed.
Fixes #1919.


templates/web/base/dashboard/index.html
```diff
@@ -152,31 +74,85 @@ <h2>[% loc('Reports') %]</h2>
...
+| <a href="[% c.uri_with({ group_by => 'state' }) %]">state</a>
+| <a href="[% c.uri_with({ group_by => 'month' }) %]">month</a>
+| <a href="[% c.uri_with({ group_by => 'category+state' }) %]">category+state</a>
+| <a href="[% c.uri_with({ group_by => 'device/site' }) %]">device/site</a>
+</p>
+
+Current state of reports that match filters:
+
+<table width="100%" id="overview">
+  [% FOR k IN grouped.keys.sort %]
```
struan: This sort order doesn't really make sense for anything other than category names. It's particularly unhelpful for months.

templates/web/base/dashboard/index.html
```diff
@@ -152,31 +74,85 @@ <h2>[% loc('Reports') %]</h2>
...
-        <a class="btn export_as_csv" href="[% c.req.uri_with({ export => 1 }) %]">[% loc('Export as CSV') %]</a>
-    </p>
 </div>
 
-<table width="100%" id="reports">
+<p align="center">
+    <input type="hidden" name="body" value="[% body.id | html %]">
+    <input type="submit" class="btn" value="[% loc('Look up') %]">
+    <input type="submit" class="btn" name="export" value="[% loc('Export as CSV') %]">
+</p>
```
struan: Might be worth making the grouping a hidden field here otherwise if you drill down it resets the grouping to the default.

templates/web/base/dashboard/index.html
```diff
@@ -24,114 +30,30 @@
...
-
-    <tr class='subtotal' id="not_marked">
-        <th scope="row">[% loc('Total not marked') %]</th>
-        <td>[% problems.wtd.not_marked %]</td>
-        <td>[% problems.week.not_marked %]</td>
-        <td>[% problems.weeks.not_marked %]</td>
-        <td>[% problems.ytd.not_marked %]</td>
-    </tr>
-
-</table>
```
struan: Are we removing all these stats on time to state change?

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
```diff
@@ -1,225 +1,18 @@
...
-    }
-}
-
-sub load_user_body : Private {
-    my ($self, $c, $body_id) = @_;
-
-    $c->stash->{body} = $c->model('DB::Body')->find($body_id)
-        or $c->detach( '/page_error_404_not_found' );
+    $c->res->redirect('/dashboard?body=' . $c->stash->{body_id});
 }
```
struan: Given all this does is redirect is there any point in keeping this, or the link in the admin navigation?

perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -77,141 +69,219 @@ sub check_page_allowed : Private {
...
-    $counts{wtd} = $c->forward( 'updates_search',
-        [ $dtf->format_datetime( $t->clone->subtract( days => $t->dow - 1 ) ) ] );
-    $counts{week} = $c->forward( 'updates_search',
-        [ $dtf->format_datetime( $now->clone->subtract( weeks => 1 ) ) ] );
-    $counts{weeks} = $c->forward( 'updates_search',
-        [ $dtf->format_datetime( $now->clone->subtract( weeks => 4 ) ) ] );
-    $counts{ytd} = $c->forward( 'updates_search',
-        [ $dtf->format_datetime( $t->clone->set( day => 1, month => 1 ) ) ] );
+    my $state = $c->stash->{q_state};
+    if ( $state eq 'fixed - council' ) {
```
struan: I'm a bit wary about this one fixed state having special status in here biting us in the future. Would it not make more sense for all fixed states to behave like this?

perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -77,141 +69,219 @@ sub check_page_allowed : Private {
...
-    }
-    my $params = {
-        %$prob_where,
-        'me.confirmed' => { '>=', $dtf->format_datetime( $now->clone->subtract( days => 30 ) ) },
-    };
-
-    if ( $c->get_param('start_date') or $c->get_param('end_date') ) {
+    my $sd = $c->stash->{start_date};
+    my $ed = $c->stash->{end_date};
+    if ($sd or $ed) {
```
struan: These seem a bit cryptic as variable names :)

templates/web/base/dashboard/index.html
dracos: Gah, doh. Sorting states is also what current admin/stats does, so I'm not too bothered about keeping that, though I agree bunching open/fixed/closed together or similar would be better. Have to do something about months though, yes! So might try and sort both at once.

perllib/FixMyStreet/App/Controller/Dashboard.pm
dracos: Yes, that is its purpose - the drop down in the front end always has 'fixed - council' because it's the staff user dropdown, and if they pick that option, this is saying map it to all fixed states in the lookup.

perllib/FixMyStreet/App/Controller/Admin/AreaStats.pm
dracos: No not really. Wondered if anyone might have bookmarked it, but meh, they can update.

templates/web/base/dashboard/index.html
dracos: Yes. Not sure if anyone using them, plus they haven't been correct for ages (ignoring most closed states, for example). The marketing page still has an average, and it can always be put back somewhere if missed, but I thought it best to cut back as much as possible first.

templates/web/base/dashboard/index.html
dracos: Ooh, yes, will do.

perllib/FixMyStreet/App/Controller/Dashboard.pm
struan: The `fixed - council` thing in the dropdown is becuase that's whats listed in `state_groups_inspect` by default. I was just thinking about if someone changed that in a cobrand. Possibly I am thinking too hard about this :)

dracos: Ah, if they did that, they'd be very silly :) Whilst you can rename the fixed state in the admin, you can't have multiple ones, it's only a tickbox that creates a 'fixed' state, and if that's present, the code sticks in 'fixed - user' and 'fixed - council' too. So all you could possibly change it to is 'fixed' or 'fixed - user', which wouldn't make much sense. But yes, it could check fixed_states, why not.

dracos:
Have found one issue, for the key1+key2 group options where if one key1 misses out an option that is present in another, the rows are missed. Storing columns to print out instead.

PR #1937, opened 2017-12-19
[Oxfordshire] Allow lat/lon to be sent as integers to Exor Open311 CGI

davea:
It took a few years but somebody finally made a report at a whole-number
lat/lon (52°N in this case) which was duly rejected by the OCC Exor
script as not being a valid coordinate due to the assumption that all
lat/lon values will contain a `.` character.

This commit makes the decimal part of the value optional, allowing
integer values to be sent.

[skip changelog]


bin/oxfordshire/open311_service_request.cgi
```diff
@@ -114,7 +114,7 @@ XML
 # returns true if this looks like a long/lat value
 #------------------------------------------------------------------
 sub is_longlat {
-    return $_[0] =~ /^-?\d+\.\d+$/o? 1 : 0;
+    return $_[0] =~ /^-?\d+(\.\d+)?$/o? 1 : 0;
```
dracos: While you're here, stick a space before that `?`?

PR #1947, opened 2018-01-03
Lazy load footer images, other image tidying

### LINKED ISSUE #1907: [UK] Do we *need* to load tile.jpg, the footer app store logos, and social media icons on map pages?
The footer and body background are all hidden on the map page but, because they’re not explicitly `display: none`, we’re still loading all the image files.

Visitors who come via the homepage will hopefully have all of those assets cached, so it won’t make much difference. But anyone who comes directly to a map page (eg: they’ve just followed a link to a report from social media or an email) will be waiting a few extra milliseconds for these useless images to load.

dracos:
Don't load tile.jpg on the map page (never visible), remove some 2x PNGs that aren't needed, and biggest, lazyload the footer images.

It does this using an IntersectionObserver that fires when something is about to come into view, no need for scroll monitoring. At that point it removes the class setting `background-image: none` which in turn causes the image to load. No need for attribute on the image or any extra information than the fallback. I haven't seen this straightforward technique used anywhere.
https://github.com/ApoorvSaxena/lozad.js is the nicest I found and that still needs you to set `data-background-image` (well, you can provide your own `load` but might as well roll your own then)

Have to decide whether to serve a polyfill to Safari or not...

Fixes #1907.


templates/web/fixmystreet.com/footer_extra.html
```diff
@@ -9,11 +9,11 @@ <h2 class="mysoc-footer__site-name">FixMyStreet</h2>
                 </div>
 
                 <div class="fms-app-badges">
-                    <a href="https://itunes.apple.com/gb/app/fixmystreet/id297456545">
-                        <img alt="FixMyStreet app on the App Store" src="[% base %]/cobrands/fixmystreet/images/itunes_store_logo.png" srcset="[% base %]/cobrands/fixmystreet/images/itunes_store_logo.png 1x, [% base %]/cobrands/fixmystreet/images/itunes_store_logo@2.png 2x" />
+                    <a class="js-lazyload fms-app-badge fms-app-badge--ios" href="https://itunes.apple.com/gb/app/fixmystreet/id297456545">
+			FixMyStreet app on the App Store
```
zarino: Eeek! Tabs!

PR #1949, opened 2018-01-04
Include "SameSite=Lax" with all set cookies.

dracos:
This prevents FixMyStreet cookies from being sent from third-party ``s and the like, in supporting browsers.

### DIFF
--- a/perllib/FixMyStreet/App/Engine.pm
+++ b/perllib/FixMyStreet/App/Engine.pm
@@ -1,0 +1,68 @@
+package FixMyStreet::App::Engine;
+
+use Moose;
+extends 'Catalyst::Engine';
+
+use CGI::Cookie;
+use utf8;
+
+use namespace::clean -except => 'meta';
+
+=head1 NAME
+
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Engine.pm
```diff
@@ -0,0 +1,68 @@
+package FixMyStreet::App::Engine;
+
+use Moose;
+extends 'Catalyst::Engine';
```
struan: It looks like we already have our own version of `Catalyst::Engine` so why not just change `finalize_cookies` in that?

dracos:
Our offline iframe HTML (for appcache) would have to always return, rather than a 404, as it will no longer be able to check. I don't think this matters.

perllib/FixMyStreet/App/Engine.pm
dracos: The Catalyst/Engine.pm is a copy of Catalyst's without changes (just a more recent version), so I thought it was clearer to keep that and subclass it properly.

PR #1968, opened 2018-01-18
Hide 'updates not sent' message on reports sent via Open311

davea:
The 'updates are not sent to the council' message was incorrectly being
shown on reports where updates would in fact be sent. This was affecting
non-Bromley/Stevenage reports on fms.com and all cobrands using Open311.

This PR removes the logic from the templates and adds the
Problem->updates_sent_to_body method which inspects the receiving body's
Open311 configuration in order to determine whether updates will be sent to
the receiving body or not.

[skip changelog]


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -867,6 +867,26 @@ sub update_send_failed {
...
+sub updates_sent_to_body {
+    my $self = shift;
+    return unless $self->send_method_used && $self->send_method_used eq 'Open311';
+
+    # Some bodies only send updates *to* FMS, they don't receive updates.
+    # NB See also the list in bin/send-comments
+    my $excluded = qr{Lewisham|Oxfordshire};
+
+    my @bodies = values %{ $self->bodies };
+    my @updates_sent = grep { $_ } map { $_->send_comments && !($_->name =~ /$excluded/) } @bodies;
```
dracos: `s'grep { $_ } map'grep'` :)

dracos: Does this need to check $_->send_method is still Open311 as well? If the body had potentially switched back to email, I mean. OTOH, if they're temporarily on Noop due to brokenness their end...

davea: Good idea, I've updated it to hide the message if the send method is `Open311` or `Noop`.

PR #1969, opened 2018-01-18
Allow admin to anonymize user

### LINKED ISSUE #1942: Allow admin to publicly anonymize all user's reports/updates
A user can already publicly anonymize all of their reports/updates at once; an admin should be able to do it on behalf of a user.

### LINKED ISSUE #1943: Way for admin to hide all a user's reports/updates/etc
I include this step rather than only a full account deletion because that is probably better done as a script that periodically purges hidden reports (so catches any that have been hidden through any method) and any users with no associated data etc. May want to retain hidden reports for a while for auditing purposes too.

dracos:
Fix #1942. Fix #1943.


t/app/controller/admin.t
```diff
@@ -1556,6 +1556,20 @@ FixMyStreet::override_config {
 
 };
 
+subtest "Anonymizing user from admin" => sub {
+    my $user = $mech->create_user_ok('existing@example.com', name => 'Existing User');
+    my $count_p = FixMyStreet::DB->resultset('Problem')->search({ user_id => $user->id })->count;
+    my $count_u = FixMyStreet::DB->resultset('Comment')->search({ user_id => $user->id })->count;
+    # FixMyStreet::DB->resultset('Problem')->search({ user_id => $user->id })->update({ anonymous => 'f' });
+    # FixMyStreet::DB->resultset('Comment')->search({ user_id => $user->id })->update({ anonymous => 'f' });
```
davea: Can these comments be removed?

dracos:
Sorry, I've stuck another commit in here for #1943 :)

PR #1976, opened 2018-01-29
Show loading message on /around page load

davea:
The map and sidebar are empty on /around when the page loads, and JS is used
to pull in the relevant reports and populate the page. The previous message
indicated that there were no reports at all, but in fact they were being loaded
asynchronously. This commit shows a more helpful message whilst the pins are
being loaded.

### DIFF
--- a/templates/web/base/around/on_map_list_items.html
+++ b/templates/web/base/around/on_map_list_items.html
@@ -14,7 +14,11 @@
     </li>
 [% ELSE %]
     <li class="item-list__item item-list__item--empty">
-        <p>[% loc('There are no reports to show.') %]</p>
+        [% IF c.get_param('js') %]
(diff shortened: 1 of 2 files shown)


templates/web/base/around/on_map_list_items.html
```diff
@@ -14,7 +14,11 @@
     </li>
 [% ELSE %]
     <li class="item-list__item item-list__item--empty">
-        <p>[% loc('There are no reports to show.') %]</p>
+        [% IF js %]
```
dracos: This is always true (because of the `PROCESS` sidebar.html which doesn't appear to use the variable), whether JS is being used or not, so will always be shown. The `js` variable is for displaying the new report form (hidden) on the around page; the thing you're wanting to test here is whether the URI has js=1 in it or not, so `c.get_param('js')`.

PR #1978, opened 2018-01-30
Allow 'report as another user' to provide just a phone number

davea:
This PR removes the requirement for an email address when using the 'report as another user' feature. Now it requires an email *or* a phone number.

Fixes mysociety/fixmystreet-commercial#836


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1369,12 +1382,14 @@ sub redirect_or_confirm_creation : Private {
...
     if ( $report->confirmed ) {
-        # Subscribe problem reporter to email updates
-        $c->forward( 'create_reporter_alert' );
-        if ($c->stash->{contributing_as_another_user}) {
-            $c->send_email( 'other-reported.txt', {
-                to => [ [ $report->user->email, $report->name ] ],
-            } );
+        if ($report->user->email) {
+            # Subscribe problem reporter to email updates
+            $c->forward( 'create_reporter_alert' );
```
dracos: This signed you up for an alert even if phone-only, so that if email later got added to the account, you'd then start getting the email alerts. Dunno if we want to keep that?

perllib/FixMyStreet/App/Controller/Report/Update.pm
```diff
@@ -468,11 +475,13 @@ sub redirect_or_confirm_creation : Private {
     # If confirmed send the user straight there.
     if ( $update->confirmed ) {
         $c->forward( 'update_problem' );
-        $c->forward( 'signup_for_alerts' );
-        if ($c->stash->{contributing_as_another_user}) {
-            $c->send_email( 'other-updated.txt', {
-                to => [ [ $update->user->email, $update->name ] ],
-            } );
+        if ($update->user->email) {
+            $c->forward( 'signup_for_alerts' );
```
dracos: Ditto here I guess.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1063,6 +1071,11 @@ sub check_for_errors : Private {
         delete $field_errors{username};
     }
 
+    # if we're contributing as someone else then allow landline numbers
+    if ( $field_errors{username} && $c->stash->{contributing_as_another_user} && !$c->stash->{phone_may_be_mobile}) {
```
dracos: If they've entered an invalid email address, this will let it through, because phone_may_be_mobile will be false.
(same check is in Update.pm too)

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -791,12 +791,20 @@ sub process_user : Private {
...
+    if ( $c->stash->{contributing_as_another_user} && !$params{username} ) {
+        # The 'report as another user' form doesn't have a 'username' field, so
+        # get a value for that param from either the email or phone.
+        $params{username} = $params{email} || $params{phone};
+    }
+
     my $parsed = FixMyStreet::SMS->parse_username($params{username});
     my $type = $parsed->{type} || 'email';
-    $type = 'email' unless FixMyStreet->config('SMS_AUTHENTICATION');
+    $type = 'email' unless FixMyStreet->config('SMS_AUTHENTICATION') || $c->stash->{contributing_as_another_user};
```
dracos: Glad it was this straightforward! :)

t/app/controller/report_as_other.t
```diff
@@ -155,6 +263,7 @@ sub start_report {
     FixMyStreet::override_config {
         ALLOWED_COBRANDS => [ 'fixmystreet' ],
         MAPIT_URL => 'http://mapit.uk/',
+        PHONE_COUNTRY => 'GB',
```
dracos: Hmm, guess we should document that you'll have to set `PHONE_COUNTRY` if you want to use report as another user with phone numbers.

dracos:
There's an error sending reports for an account with no email; however,`user_details` doesn't appear to be used anywhere (think it was SSSS?).

davea:
Thanks for the review - I've addressed all those things now I think.

PR #1980, opened 2018-02-01
Admin ability to remove user account details.

### LINKED ISSUE #1944: Admin way of removing account
This would replace name/email/phone with anonymous versions (so as to maintain associated reports/updates), delete all alerts, anything else it would need to do?

dracos:
Fixes #1944

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1423,6 +1423,8 @@ sub user_edit : Path('user_edit') : Args(1) {
 
     if ( $c->get_param('submit') and $c->get_param('unban') ) {
         $c->forward('unban_user', [ $user ]);
+    } elsif ( $c->get_param('submit') and $c->get_param('logout_everywhere') ) {
+        $c->forward('user_logout_everywhere', [ $user ]);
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1777,6 +1779,27 @@ sub user_hide_everywhere : Private {
     $c->stash->{status_message} = _('That user’s reports and updates have been hidden.');
 }
 
+# Anonymize and remove name from all problems/updates, disable all alerts.
+# Remove their account's email address, phone number, password, etc.
+sub user_remove_account : Private {
+    my ( $self, $c, $user ) = @_;
+    $user->problems->update({ anonymous => 1, name => '' });
```
struan: Should this not also make sure that questionnaire sending is turned off for the problems? More to stop annoying bounces than anything else.

dracos: Ooh, good point. Also moderation emails, thinking about it, hmm (probably rare enough not to bother there).

dracos:
"I'm not completely sure why these changes cause that though." - my new logout code tests `user_exists` (which passes as the user is still present in the session data) and then uses `user` (which doesn't exist as it's been deleted). I should test `user` directly (just in the weird case a DB entry is deleted, even if it shouldn't happen).

PR #1981, opened 2018-02-02
Add password checking on setting.

dracos:
Minimal length and comparison against common password list.

### DIFF
--- a/web/js/validation_rules.js
+++ b/web/js/validation_rules.js
@@ -1,5 +1,11 @@
     validation_rules = {
         title: { required: true },
         detail: { required: true },
-        update: { required: true }
+        update: { required: true },
+        password_register: {
+          remote: {
+            url: '/auth/common_password',
+            type: 'post'
+          }
+        }
     };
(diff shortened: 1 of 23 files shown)


web/js/validation_rules.js
```diff
@@ -1,5 +1,11 @@
     validation_rules = {
         title: { required: true },
         detail: { required: true },
-        update: { required: true }
+        update: { required: true },
+        password_register: {
+          remote: {
+            url: '/auth/common_password',
+            type: 'post'
```
struan: Why do only the common password bit here but not the length?

dracos: I was going to say the length is handled by fixmystreet.js in the addMethod, but that's then fixed whereas I made server-side cobrand specific didn't I, hmm.

dracos:
"Should this not also handle updating passwords too?" oh yes!

web/js/validation_rules.js
dracos: Will still do it client side as it's quicker, but will pass a variable from server to client.

PR #1983, opened 2018-02-05
Add two-factor authentication for superusers.

dracos:
Todo:
- [x] test of turning on/off 2FA
- [x] test superuser being denied report making by code, and go through 2FA login flow

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -232,6 +232,9 @@ sub process_login : Private {
     $c->detach( '/page_error_403_access_denied', [] )
         if FixMyStreet->config('SIGNUPS_DISABLED') && !$user->in_storage && !$data->{old_user_id};
 
+    # Superusers using 2FA can not log in by code
+    $c->detach( '/page_error_403_access_denied', [] ) if $user->has_2fa;
+
     if ($data->{old_user_id}) {
         # Were logged in as old_user_id, want to switch to $user
         if ($user->in_storage) {
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -273,6 +273,11 @@ Used after signing in to take the person back to where they were.
 
 sub redirect_on_signin : Private {
     my ( $self, $c, $redirect, $params ) = @_;
+
+    if ($c->stash->{detach_to}) {
+        $c->detach($c->stash->{detach_to}, $c->stash->{detach_args});
+    }
+
     unless ( $redirect ) {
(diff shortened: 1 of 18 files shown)


templates/web/base/auth/2faform.html
```diff
@@ -0,0 +1,25 @@
+[% INCLUDE 'header.html', bodyclass = 'fullwidthpage', title = loc('Confirm account') %]
+
+  <div class="confirmation-header confirmation-header--phone">
+    [% IF incorrect_code %]
+        <h1>[% loc('Sorry, that wasn&rsquo;t the correct code') %]</h1>
+        <p>[% loc('Try again') %]:</p>
+    [% ELSE %]
+        <h1>[% loc("Nearly done! Now check your phone&hellip;") %]</h1>
+        <p>[% loc("Please generate a 2FA code and enter it below:") %]</p>
```
struan: Not sure that's the correct title to use :)

templates/web/base/auth/generate_token.html
```diff
@@ -31,6 +52,7 @@ <h1>[% loc('Generate token') %]</h1>
 
     <p>
     <input name="generate_token" type="submit" class="btn" value="[% existing_token ? loc('Replace token') : loc('Generate token') %]">
+    <input name="toggle_2fa" type="submit" class="btn" value="[% has_2fa ? loc('Deactivate 2FA') : loc('Activate 2FA') %]">
```
struan: Should we not only show this button to superusers?

templates/web/base/auth/generate_token.html
```diff
@@ -12,12 +12,33 @@ <h1>[% loc('Your token has been generated') %]</h1>
         <span>[% existing_token | html %]</span>
     </p>
 
+<img src="[% qr_code %]">
```
struan: Will this ever appear here?

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -232,6 +232,10 @@ sub process_login : Private {
     $c->detach( '/page_error_403_access_denied', [] )
         if FixMyStreet->config('SIGNUPS_DISABLED') && !$user->in_storage && !$data->{old_user_id};
 
+    # Superusers using 2FA can not log in by code
```
struan: There seems to be some inconsistency between this and creating a new report which doesn't allow logging in by code at all for super users. I think I'd be fine with preventing it everywhere.

templates/web/base/auth/2faform.html
dracos: I assumed most people will have the Authenticator app on their phone, so that is enough right for me :)

dracos:
I've changed it so superuser-with-2FA-on can not log in by email/text, but to make up for that they *can* now log in by password/2FA whilst making a report/update, with a variant on the FB/Twitter way of remembering where they are in the flow.

PR #1985, opened 2018-02-06
prevent timezone related util test failure

### LINKED ISSUE #1984: utils tests of prettify_dt can fail if local timezone is different from tests timezone
e.g. local timezone is UTC+13 and test timezone is UTC then tests fail in the morning as:

```
#   Failed test at t/utils.t line 71.
#          got: '21:37, Tuesday'
#     expected: '21:37 today'
```

Arguably this is actually a problem with `prettify_dt`.

struan:
Make sure that the prettify_dt tests use the same timezone as FMS
otherwise the test can fail if now in the local timezone is actually
yesterday/tomorrow in the FMS timezone.

Fixes #1984

This does point to underlying issues with `prettify_dt` but given that FMS doesn't really handle multiple timezones I suspect that `prettify_dt` getting passed a date in a different timezone is probably a sign of something else wrong.

### DIFF
--- a/t/utils.t
+++ b/t/utils.t
@@ -67,7 +67,9 @@ is Utils::cleanup_text( "This has new\n\n\nlines in it", { allow_multiline => 1 
 
 
 is Utils::prettify_dt(), "[unknown time]";
-my $dt = DateTime->now;
+# Make sure we create the date using the FMS timezone that prettify_dt uses
+# otherwise this can fail if the local timezone is not the same as the FMS one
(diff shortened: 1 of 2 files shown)


t/utils.t
```diff
@@ -67,7 +67,9 @@ is Utils::cleanup_text( "This has new\n\n\nlines in it", { allow_multiline => 1
 
 
 is Utils::prettify_dt(), "[unknown time]";
-my $dt = DateTime->now;
+# Make sure we create the date using the FMS timezone that prettify_date uses
```
dracos: It's `prettify_dt` :)

PR #1987, opened 2018-02-07
Script to remove expired sessions.

dracos:
This also can set up users so that the admin
"Log user out" function works correctly.

### DIFF
--- a/bin/expire-sessions
+++ b/bin/expire-sessions
@@ -1,0 +1,79 @@
+#!/usr/bin/env perl
+
+# expire-sessions: Run regularly to remove old sessions (plus
+# can set up data for 'log user out' admin functionality).
+
+use strict;
+use warnings;
+require 5.8.0;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
+use FixMyStreet::DB;
+use Getopt::Long;
+use List::Util qw(uniq);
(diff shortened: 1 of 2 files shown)


bin/expire-sessions
```diff
@@ -0,0 +1,79 @@
...
+use Storable;
+
+GetOptions(
+    'init' => \my $init,
+);
+
+my $rs = FixMyStreet::DB->resultset("Session");
+
+# Delete expired sessions (including from in User object)
+while (my $session = $rs->search({ expires => { '<', time() - 86400*7 } })->next) {
```
struan: might be nice to include a comment here saying why we are using expired a week ago as the criteria.

I assume just to give a bit of leeway?

bin/expire-sessions
```diff
@@ -0,0 +1,79 @@
...
+    while (my $session = $rs->next) {
+        my $user = get_user($session) or next;
+        my $id = get_id($session);
+        my $sessions = $user->get_extra_metadata('sessions');
+        my @new_sessions = uniq @$sessions, $id;
+        update_user_sessions($user, \@new_sessions) if @new_sessions != @$sessions;
+    }
+}
+
+# ---
```
struan: ?

bin/expire-sessions
dracos: Yeah, dunno why TBH (db should ignore any past expiry), will change to just be now.

dracos:
Merged as 5260280b60a1192f87987c34a78fce63fc5c1d26 (added line to crontab).

PR #1988, opened 2018-02-08
Buckinghamshire Cobrand

davea:
- Adds Buckinghamshire cobrand
 - Displays Bucks gritbins/street lights on map

Deployed at http://buckinghamshire.commercial.fixmystreet.staging.mysociety.org.
Originally for mysociety/fixmystreet-commercial#793, see also mysociety/fixmystreetforcouncils#278.

[skip changelog]


davea:
This is now ready for review. Another couple of non-Bucks-specific commits made their way in, but they're small. Hopefully nothing major wrong here as they want to launch on the 3rd 😄

templates/web/buckinghamshire/report/update-form.html
```diff
@@ -0,0 +1,43 @@
+[% allow_creation = !c.cobrand.only_authed_can_create || (c.user && c.user.from_body) %]
+[% RETURN IF NOT allow_creation %]
+
+<div id="update_form">
+  [% IF NOT login_success AND NOT oauth_need_email %]
+    <h2[% IF two_column_sidebar %] class="hidden-js"[% END %]>[% loc( 'Provide additional information') %]</h2>
```
struan: This seems like a lot of template duplication just to change the heading.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -51,11 +51,23 @@ sub open311_config {
...
 
 sub map_type { 'Buckinghamshire' }
 
+sub default_map_zoom { 3 }
+
+sub category_extra_hidden {
+    my ($self, $meta) = @_;
+    return 1 if $meta eq 'asset_details';
+    return 0;
+}
```
struan: This could also be done using the automated attributes in open311-adapter. Don't suppose it matters where we do it though.

templates/web/buckinghamshire/report/new/form_report.html
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+[% END %]
+
+[% TRY %][% PROCESS 'report/new/after_photo.html' %][% CATCH file %][% END %]
+
+    <label for="form_detail">[% loc('Explain what’s wrong') %]</label>
+[% IF field_errors.detail %]
+    <p class='form-error'>[% field_errors.detail %]</p>
+[% END %]
+    <textarea class="form-control" rows="7" cols="26" name="detail" id="form_detail" placeholder="Dimensions, landmarks, direction of travel etc." required>[% report.detail | html %]</textarea>
```
struan: There's again a lot of duplication here to change a few bits of text. Not sure if there's a better solution as having lots of tiny templates with odd bits of text also seems unweildy.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -72,4 +72,206 @@ sub category_extra_hidden {
...
+        "63068",
+        "63470",
+        "63476",
+        "63501",
+        "63507",
+        "63517",
+        "63554",
+        "63715",
+        "63723"
+    ];
```
struan: Do we really need to hardcode this list?

templates/web/buckinghamshire/report/new/fill_in_details_form.html
```diff
@@ -0,0 +1,37 @@
...
+
+<div id="js-bucks-responsibility" class="box-warning hidden">
+    <strong>Invalid location</strong>
+    <p id="js-not-bucks-road" class="hidden js-responsibility-message">
+        The selected road is not maintained by Buckinghamshire County Council.
+    </p>
+    <p id="js-not-a-road" class="hidden js-responsibility-message">
+        Please select a road on which to make a report.
+    </p>
+</div>
```
struan: This should probably be an optional partial for the cobrand.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
davea: Initially this was because adding `CPC` to the `/areas` call on MapIt was timing out. That's now been fixed, so we could remove this whitelist, but that would mean the "areas" multiselect in the admin would contain all parishes in the UK which isn't ideal.

struan: Oh, yes. That seems like a good reason :)

PR #1990, opened 2018-02-08
[Rutland] Branded email templates

zarino:
Fixes mysociety/fixmystreet-commercial#972.

I’ve used the (crest-less) logo that they’re currently using on their website, rather than the (crest-ful) logo currently being applied in the cobrand web header.

screen shot 2018-02-08 at 12 01 40
[skip changelog]


davea:
That big yellow box should probably be omitted given we do integrate with Rutland.

zarino:
@davea @dracos Ah, do we want the "Where did this report come from?" and "Is there a better contact address?" gone too?

dracos:
TBH, I don't think we'll be sending reports via email, so this'll never be seen :)

zarino:
> TBH, I don't think we'll be sending reports via email, so this'll never be seen :)

@dracos I assume this is why `_submit_footer.html` hasn’t been overridden in the Oxfordshire cobrand either – because `submit.html` isn’t actually used any more in cobrands with full integration?

PR #1991, opened 2018-02-08
Display map loading spinner for asset layers

davea:
- Displays the loading spinner whilst asset layers are being loaded
 - Slight refactor of spinner code to keep track of multiple things loading simultaneously


davea:
Good idea, I'll add the event handler to that layer and merge. Thanks for the review 🍰

PR #1993, opened 2018-02-08
Make asset layers more awesome

davea:
This PR makes the asset layers code easier to work with if you want to load features from a quirky WFS endpoint, or show a layer independent of the category field, or provide your own styles for features.

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -113,7 +113,7 @@ function check_zoom_message_visibility() {
 
 function layer_visibilitychanged() {
     check_zoom_message_visibility.call(this);
-    var layers = fixmystreet.map.getLayersByName('WFS');
+    var layers = fixmystreet.map.getLayersBy('assets', true);
     var visible = 0;
     for (var i = 0; i<layers.length; i++) {
         if (layers[i].getVisibility()) {
@@ -304,7 +304,8 @@ fixmystreet.add_assets = function(options) {
             visibility: false,
             maxResolution: options.max_resolution,
             minResolution: options.min_resolution,
-            styleMap: options.stylemap || get_asset_stylemap()
+            styleMap: options.stylemap || get_asset_stylemap(),
+            assets: true
         };
         if (options.srsName !== undefined) {
             layer_options.projection = new OpenLayers.Projection(options.srsName);
@@ -319,7 +320,7 @@ fixmystreet.add_assets = function(options) {
                 value: options.filter_value
             });
         }
-        asset_layer = new OpenLayers.Layer.Vector("WFS", layer_options);
+        asset_layer = new OpenLayers.Layer.Vector(options.name || "WFS", layer_options);
 
         // A non-interactive layer to display existing street light faults
         if (options.wfs_fault_feature) {
@@ -331,7 +332,8 @@ fixmystreet.add_assets = function(options) {
             var lo = {
                 strategies: [new OpenLayers.Strategy.BBOX()],
                 protocol: fault_protocol,
-                styleMap: get_fault_stylemap()
+                styleMap: get_fault_stylemap(),
+                assets: true
             };
             OpenLayers.Util.applyDefaults(lo, layer_options);
             asset_fault_layer = new OpenLayers.Layer.Vector("WFS", lo);
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -17,7 +17,7 @@ function asset_selected(e) {
     close_fault_popup();
     var lonlat = e.feature.geometry.getBounds().getCenterLonLat();
 
-    // Check if there is a known fault with the light that's been clicked,
+    // Check if there is a known fault with the asset that's been clicked,
     // and disallow selection if so.
     var fault_feature = find_matching_feature(e.feature, this.fixmystreet.fault_layer, this.fixmystreet.asset_id_field);
     if (!!fault_feature) {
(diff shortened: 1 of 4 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -311,43 +341,63 @@ fixmystreet.add_assets = function(options) {
...
+                // as it causes some mad flickering effect.
+                if (e.feature.renderIntent == 'select') {
+                    return false;
+                }
+            });
+        }
+        if (options.asset_category) {
+            fixmystreet.map.events.register( 'zoomend', asset_layer, check_zoom_message_visibility);
+        }
+        if (!options.always_visible) {
```
dracos: If you're on an around page (shows always_visible assets), click map, click back, they disappear and never return.

web/cobrands/fixmystreet/assets.js
```diff
@@ -258,29 +258,52 @@ fixmystreet.add_assets = function(options) {
...
             visibility: false,
             maxResolution: options.max_resolution,
             minResolution: options.min_resolution,
             styleMap: get_asset_stylemap()
         };
-        if (fixmystreet.wmts_config) {
+        if (options.srsName !== undefined) {
+            layer_options.projection = new OpenLayers.Projection(options.srsName);
+        }
+        if (fixmystreet.wmts_config && layer_options.projection === undefined) {
```
dracos: This could be an `} else if (fixmystreet.wmts_config)`.

web/cobrands/fixmystreet/assets.js
```diff
@@ -258,29 +258,52 @@ fixmystreet.add_assets = function(options) {
...
+            protocol_options = {
+                version: "1.1.0",
+                url: options.wfs_url,
+                featureType: options.wfs_feature,
+                geometryName: options.geometryName
+            };
+            if (options.srsName !== undefined) {
+                protocol_options.srsName = options.srsName;
+            }
+            if (fixmystreet.wmts_config && protocol_options.srsName === undefined) {
```
dracos: This could be an `} else if (fixmystreet.wmts_config)`

web/cobrands/fixmystreet/assets.js
```diff
@@ -258,29 +258,52 @@ fixmystreet.add_assets = function(options) {
...
-            protocol_options.propertyNames = options.propertyNames;
+        var protocol_options;
+        var protocol;
+        if (options.http_options !== undefined) {
+            protocol_options = OpenLayers.Util.extend(options.http_options, {});
+            if (protocol_options.format_class) {
+                protocol_options.format = new protocol_options.format_class(protocol_options.format_options);
+            } else {
+                protocol_options.format = new OpenLayers.Format.GML({
+                    featureNS: "http://mapserver.gis.umn.edu/mapserver",
```
dracos: I think this is the default anyway?

web/cobrands/fixmystreet/assets.js
```diff
@@ -258,29 +258,52 @@ fixmystreet.add_assets = function(options) {
...
+                    geometryName: options.geometryName
+                });
+            }
+            protocol = new OpenLayers.Protocol.HTTP(protocol_options);
+            protocol.srsInBBOX = true;
```
dracos: Is there a reason this is set here rather than in the options? Not sue I follow the code flow entirely, but I think this means it's actually false for the purposes of the BBOX in the URL requested (as it's only read during object setup).

davea: You're right, this didn't actually do anything. I'd left it in accidentally.

web/cobrands/fixmystreet/assets.js
```diff
@@ -416,28 +393,62 @@ fixmystreet.add_assets = function(options) {
                 }
                 zoom_to_assets(asset_layer);
             } else {
-                asset_layer.setVisibility(false);
+                if (!asset_layer.fixmystreet.always_visible) {
```
dracos: Should you wrap the whole `on()` change function in this check - presumably you don't want any of it to run when there's a category change?

web/cobrands/fixmystreet/assets.js
```diff
@@ -416,28 +393,62 @@ fixmystreet.add_assets = function(options) {
...
 
-    // Make sure the assets get hidden if the back button is pressed
-    fixmystreet.maps.display_around = (function(original) {
-        function hide_assets() {
-            asset_layer.setVisibility(false);
-            if (asset_fault_layer) {
-                asset_fault_layer.setVisibility(false);
+
+        // Make sure the assets get hidden if the back button is pressed
+        fixmystreet.maps.display_around = (function(original) {
```
dracos: You only want to do this once now, not on every call to `add()`.

web/cobrands/fixmystreet/assets.js
```diff
@@ -416,28 +393,62 @@ fixmystreet.add_assets = function(options) {
...
+                }
+                fixmystreet.markers.setVisibility(true);
+                original.apply(fixmystreet.maps);
+            }
+            return hide_assets;
+        })(fixmystreet.maps.display_around);
+    },
+
+    init: function() {
+        if (fixmystreet.map === undefined) {
```
dracos: Do we want the check-fixmystreet-is-defined safety check to do nothing at all? Or do we think that's okay now.

web/cobrands/fixmystreet/assets.js
```diff
@@ -416,28 +393,62 @@ fixmystreet.add_assets = function(options) {
...
+            // Map's not loaded yet, let's try again soon...
+            setTimeout(fixmystreet.assets.init, 250);
+            return;
+        }
+        if (fixmystreet.page != 'new' && fixmystreet.page != 'around') {
+            // We only want to show light markers when making a new report
```
dracos: Might as well fix this comment if it's being moved.

web/cobrands/fixmystreet/assets.js
davea: 🤦‍♂️

web/cobrands/fixmystreet/assets.js
davea: Looks like this is still required, due to the order that things are loaded in. Ideally there'd be a post-setup hook for `fixmystreet.map`, I'll see if that's a simple thing to add.

PR #1998, opened 2018-02-13
Improvements to questionnaire process

dracos:
Some definite bits, some bits we might not want:
* Adds don't know to questionnaire email;
* Prefills the first question when one of those links is clicked;
* Rearranges the form slightly so (esp. on mobile) more obvious it is a form at the top by putting report under first question, and numbering the questions;
* Stores the questionnaire data as soon as link in email clicked – that includes updating the report, so someone clicking Yes to is it fixed will instantly have their report updated even if they don't fill in the form. If they do fill in the form, it updates the same questionnaire row (there might be edge bugs I haven't got yet if they change their answer from what they clicked on in link, but only in terms of the questionnaire's "old" state iyswim [1]).
Comments welcome, I think all the above is good, but might have missed something.

[1] ie. if problem was open, they click Yes it's fixed in email, problem updated to be fixed, they change question 1 to be No it's not, then submit, questionnaire would get state change of "fixed" to "open" rather than correct "open" to "open". I guess it would have to check/use the 'unfinished' questionnaire's stored result as the origin rather than the report's current state.


dracos:
Sigh, looks like update questionnaire showing not been tested before.

PR #1999, opened 2018-02-15
More asset/inspector form improvements

davea:
- Display asset layers beneath existing markers on the map
 - Fill report `usrn` field from highways asset layer independently of e.g. street lights/grit bins
 - Fix bug that would delete values of hidden report extra fields when inspector form saved
 - Show hidden report extra fields on inspector form

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -523,4 +523,25 @@ OpenLayers.Layer.Vector.prototype.getNearestFeature = function(point, threshold)
     return nearest_feature;
 };
 
+
+/*
+ * MapServer 6 (the version available on Debian Wheezy) outputs incorrect
+ * GML for MultiCurve geometries - see https://github.com/mapserver/mapserver/issues/4924
+ * The end result is that features with 'curveMembers' elements in their
+ * geometries will be missing from the map as the default GML parser doesn't
+ * know how to handle these elements.
+ * This subclass works around the problem by parsing 'curveMembers' elements.
+ */
+OpenLayers.Format.GML.v3.MultiCurveFix = OpenLayers.Class(OpenLayers.Format.GML.v3, {
+    readers: $.extend(true, {}, OpenLayers.Format.GML.v3.prototype.readers, {
+        "gml": {
+            "curveMembers": function(node, obj) {
+                this.readChildNodes(node, obj);
+            }
+        }
+    }),
+
+    CLASS_NAME: "OpenLayers.Format.GML.v3.MultiCurveFix"
+});
+
 })();
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -4,6 +4,8 @@ var fixmystreet = fixmystreet || {};
 
 var selected_feature = null;
 var fault_popup = null;
+var selected_usrn = null;
+var usrn_field = null;
 
 function close_fault_popup() {
     if (!!fault_popup) {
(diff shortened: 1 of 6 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -426,12 +421,24 @@ fixmystreet.assets = {
...
             }
+
+            // Don't cover the existing pins layer
+            if (pins_layer) {
+                layer.setZIndex(pins_layer.getZIndex()-1);
+            }
+
+            // Make sure the fault markers always appear beneath the linked assets
+            if (layer.fixmystreet.fault_layer) {
+                layer.fixmystreet.fault_layer.setZIndex(layer.getZIndex()-1);
```
dracos: We're in a loop here through all the layers, both assets and faults. So tis won't work (I think) because it'll do the asset layer first (set z to pins-1, set faults to assets-1), then it'll do the fault layer and set z to pins-1 again?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -417,6 +417,9 @@ $.extend(fixmystreet.set_up, {
             } else {
                 $category_meta.empty();
             }
+            if (fixmystreet.assets) {
+                fixmystreet.assets.update_usrn_field();
```
dracos: I'm not sure what this is for/ how the USRN could change at this point?

web/cobrands/fixmystreet/assets.js
```diff
@@ -445,11 +442,79 @@ fixmystreet.assets = {
...
+            // whatever the map projection uses. Convert to metres in order to
+            // draw a meaningful comparison to the threshold value.
+            var p1 = new OpenLayers.Geometry.Point(details.x0, details.y0);
+            var p2 = new OpenLayers.Geometry.Point(details.x1, details.y1);
+            var line = new OpenLayers.Geometry.LineString([p1, p2]);
+            var distance_m = line.getGeodesicLength(this.map.getProjectionObject());
+
+            if (distance_m <= threshold) {
+                nearest_feature = candidate;
+                nearest_distance = details.distance;
```
dracos: To save (potentially?) on some processing, do you want to always set nearest_distance so that it can skip any further away, even if outside the threshold?

web/cobrands/fixmystreet/assets.js
```diff
@@ -445,11 +442,79 @@ fixmystreet.assets = {
...
 };
 
 $(function() {
     fixmystreet.assets.init();
 });
 
+OpenLayers.Layer.Vector.prototype.getFeatureAtPoint = function(point) {
+    for (var i = 0; i < this.features.length; i++) {
+        var feature = this.features[i];
+        if (feature.geometry.containsPoint && feature.geometry.containsPoint(point)) {
```
dracos: Do we know all features will have geometries?

web/cobrands/fixmystreet/assets.js
davea: D'oh. Good idea!

web/cobrands/fixmystreet/assets.js
davea: Hah, no, they won't. I just discovered this the hard way. Will fix!

web/cobrands/fixmystreet/fixmystreet.js
davea: This is for the situation where you've changed the category and so the report extra fields have been altered in the DOM.

web/cobrands/fixmystreet/assets.js
davea: Yep, you're right. Will rejig this to handle that.

PR #2000, opened 2018-02-19
[Bristol] Always include an email address for Open311 reports

davea:
Bristol's Open311 endpoint isn't fully compliant with [the spec](http://wiki.open311.org/GeoReport_v2/#post-service-request) as it rejects service requests that have no email parameter. This is causing stuck reports since [email addresses aren't always required on fixmystreet.com](https://github.com/mysociety/fixmystreet/pull/1978) any more.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bristol.pm
+++ b/perllib/FixMyStreet/Cobrand/Bristol.pm
@@ -71,4 +71,10 @@ sub categories_restriction {
     return $rs->search( { 'me.send_method' => undef } );
 }
 
+sub open311_config {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bristol.pm
```diff
@@ -71,4 +71,9 @@ sub categories_restriction {
     return $rs->search( { 'me.send_method' => undef } );
 }
 
+sub open311_pre_send {
```
dracos: I think the `open311_config` hook would be a better place to set this?

Issue #1462, opened 2016-07-15
Problem creating a new report in busy areas


BenJam:
Where we have particularly busy areas residents are frequently (one authority handling 5 per day) leaving a comment on an existing report that should be a new issue.

dracos:
Possibilities:
* #536 (note East Sussex has something on these lines already, making someone choose first if they're reporting or leaving an update);
* #393 (now #1351 is done, the converse could now be done, whereby viewing a report page directly could act as it does when you click a report from a list page (ie. other reports are shown very small, you can start a new report from it more easily, etc)
* #525 (making hide pins more prominent)
* Reduce time before pins stop appearing on the map by default.

abimysoc:
I'd be interested to see what effect #1351 has had on this. The support mail (or lack of) for the main site is already indicating to me that it's changed user behaviour significantly: much, much fewer people now seem to be mistaking the 'report abuse' link on an existing report for the way to make a new report. Hopefully this means it is more obvious how to make a new report instead of updating/flagging an existing one.

abimysoc:
User support emails about the crowdedness of the map / the difficulty of placing new pins definitely seem to be on the increase as the number of reports on the site goes up, particularly in places like Oxfordshire.

dracos:
Here are my current thoughts (could also see @zarino's previous suggestions at https://github.com/mysociety/fixmystreet/issues/525#issuecomment-309734261 – 1 and 2 were done in that list - and my previous plus others at https://github.com/mysociety/fixmystreet/issues/1519#issuecomment-246993721).

I think the following should be implemented:
* (Solution 3 in Zarino's comment, dealing with mis-click on report, or just being on a repot page in general) On a report page (directly or via around), have a way of starting a new report directly at that location: "Report a different problem at this location" [easier to implement if come via around, in JS]
* (asked for couple of months ago, https://github.com/mysociety/fixmystreet-commercial/issues/963) Tracking any clicks on green banner, just to see
* Make hide pins bar/stuff much bigger on mobile, easier to see and to hit.

I think the following should be done, tested ("please try and report a pothole at X" where X is quite busy):
* Moving the green banner on top of the map (comment in #1519);
* Have clicking the green banner, where it is or moved from above, hide/show the pins (or even just on hover hide them, have the bar be some form of toggling interface);
* Replacing the green banner with a draggable pin and instructions to drag it on to the map to report a problem - *starting* a drag immediately hides all map pins. Clicking would still work too;
[...]

PR #2003, opened 2018-02-23
Increase size of sub map links

zarino:
The tiny buttons could prove hard to trigger, especially on touchscreen devices, and with large screens now more commonplace, there’s no reason not to give them a few more pixels.

One of a number of enhancements suggested in #1462.

# Mobile: Before / After

mobile-before mobile-after

# Desktop: Before / After

desktop-before desktop-after

### DIFF
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -671,11 +671,13 @@ body.authpage {
 
   &:after {
     content: "";
-    display: inline-block;
-    vertical-align: middle;
+    display: block;
+    position: absolute;
     width: 6px;
     height: 12px;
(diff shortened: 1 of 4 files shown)


web/cobrands/sass/_layout.scss
```diff
@@ -671,11 +671,13 @@ body.authpage {
 
   &:after {
     content: "";
-    display: inline-block;
-    vertical-align: middle;
```
dracos: Just curious as to why top 50% -1/2 height is better than this (I'm sure it is, just for my own benefit).

zarino:
Changelog added! Merging…

PR #2006, opened 2018-02-27
Development email previewer

dracos:
This adds `/_dev/email` when run in development, that lets you easily preview any HTML email as it would look, including inline images.

TODO: note in docs

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Develop.pm
+++ b/perllib/FixMyStreet/App/Controller/Develop.pm
@@ -1,0 +1,136 @@
+package FixMyStreet::App::Controller::Develop;
+use Moose;
+use namespace::autoclean;
+
+use File::Basename;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+=head1 NAME
+
+FixMyStreet::App::Controller::Develop - Catalyst Controller
+
+=head1 DESCRIPTION
+
+Developer-helping Catalyst Controller.
+
+=head1 METHODS
+
+=over 4
+
+=item auto
+
+Makes sure this controller is only available when run in development.
+
+=cut
+
+sub auto : Private {
+    my ($self, $c) = @_;
+    $c->detach( '/page_error_404_not_found' ) unless $c->config->{STAGING_SITE};
+    return 1;
+}
+
+=item email_list
+
+Shows a list of links to preview HTML emails.
+
+=cut
+
+sub email_list : Path('/_dev/email') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    my @include_path = @{ $c->cobrand->path_to_email_templates($c->stash->{lang_code}) };
+    push @include_path, $c->view('Email')->config->{INCLUDE_PATH}->[0];
+    my %templates;
+    foreach (@include_path) {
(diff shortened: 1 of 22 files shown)


templates/web/base/develop/email_list.html
```diff
@@ -0,0 +1,11 @@
+[% INCLUDE 'header.html' %]
+
+<h1>[% loc('Templates') %]</h1>
+
+<ul>
+[% FOR template IN templates %]
+<li><a href="[% template.url %]">[% template.name %]</a></li>
```
zarino: Relative `href` here (eg: `href="alert-confirm"`) works if you access the email dev page via the URL `http://fixmystreet.local:3000/_dev/email/` (with a trailing slash) but _not_ if you access it via `http://fixmystreet.local:3000/_dev/email` (without the trailing slash).

Just mention it because, at first, none of the links worked for me, and I realised it was because I was missing a trailing slash, so all the relative `href`s were leading to places like `http://fixmystreet.local:3000/_dev/alert-confirm`. I wonder whether there’s a simple way to avoid that?

perllib/FixMyStreet/App/Controller/Develop.pm
```diff
@@ -0,0 +1,136 @@
...
+        } elsif ($part->content_type =~ m[text/html]i) {
+            $html = $part->body_str;
+        }
+    });
+
+    foreach (keys %images) {
+        $html =~ s/cid:([^"]*)/data:$images{$1}/g;
+    }
+
+    $c->response->body($html);
```
zarino: Something is injecting jQuery and the plDebug toolbar into the response that my browser is receiving from the server. I guess it _shouldn’t_ matter, because neither of these things should affect the layout of the emails, but I just wondered whether it would be more _theoretically pure_ if the HTML we received in this preview was _literally_ just the HTML of the HTML part of the email, and nothing else?

dracos: You can run `FIXMYSTREET_APP_DEBUG=0 script/server` to run the development server with debug switched off (including the toolbar). As you say, it shouldn't affect it (and might be useful to see what queries are being made just in the render, perhap). The toolbar is looking for `</body>` and attaching itself, as it operates 'above' the main code I don't think there's a way for the code to disable it for a particular route (there probably is somehow, but can't see one obviously).

templates/web/base/develop/email_list.html
dracos: Hmm, yes, will look at that.

dracos: Okay, it should work at either now.

perllib/FixMyStreet/App/Controller/Develop.pm
zarino: Yes, you’re right, there might be tools in the debug toolbar that you want to use to interrogate the email. (Even if not right now, maybe in the future!). Good tip about disabling `FIXMYSTREET_APP_DEBUG` at runtime. :+1:

templates/web/base/develop/email_list.html
zarino: :+1:

PR #2007, opened 2018-02-27
Consolidate with upstream

jonkri:
[skip changelog]


dracos:
@jonkri Hi! I've reworked this a bit and rebased on top of current master. A few things I've done, please do check them out and see if it's all okay and working for you before I merge:
* I've added a `bounded=1` to the OSM geocoder as that actually restricts results to the bounding box, not sure how that was missed by us before! For one or two of your special cases, it works better;
* I added a `geocoder_munge_results` hook to OSM.pm rather than have cobrand specific stuff in the file.
* I tidied up the special `bounds` cases to make that code smaller.

jonkri:
I think the changes look good. :) Thank you!

PR #2010, opened 2018-03-01
Edge-to-edge email content on narrow screens

### LINKED ISSUE #1928: Useless padding either side of emails on narrow screens
Emails from fixmystreet.com include 20px of grey padding either side, which just takes up space.

screen shot 2017-11-29 at 15 19 58

Could we remove it?

zarino:
Fixes #1928.

# Narrow screen, before and after:

before after

(Might look a bit weird on the white Github ticket background, but looks much better on a phone!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
     - Front end improvements:
         - Improve questionnaire process. #1939 #1998
         - Increase size of "sub map links" (hide pins, permalink, etc) #2003
+        - Edge-to-edge email layout on narrow screens #2010
     - Bugfixes:
         - Stop asset layers obscuring marker layer. #1999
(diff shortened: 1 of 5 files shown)


zarino:
> What was the [%~ # 550px = 5+5+5+520+5+5+5 %] about?

`520` was the `min-width:` from the inline styles applied to `#main`. And the six `5`s were the padding-left and padding-right on the three `<th>` elements (created via the `cellpadding="5"` attribute on the main wrapper table).

The aim was to apply the given media query _all the way up until_ the window was large enough to house the 520px-wide container and all the horizontal padding outside it (6 × 5px). So, we set a `max-width` of 1px less than 520+(6×5), or 549px.

Anyway, now that the wrapper table doesn’t have any `cellpadding`, all the logic could be dispensed with :-)

PR #2011, opened 2018-03-02
[Zurich] Fix couple of bugs + general admin edit bug

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1007
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/975
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/976
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/981

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -780,6 +780,28 @@ sub update_user : Private {
     return 0;
 }
 
+sub report_edit_display : Private {
+    my ( $self, $c ) = @_;
+
+    my $problem = $c->stash->{problem};
+
+    $c->stash->{page} = 'admin';
+    FixMyStreet::Map::display_map(
+        $c,
+        latitude  => $problem->latitude,
+        longitude => $problem->longitude,
+        pins      => $problem->used_map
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -814,34 +800,46 @@ sub report_edit : Path('report_edit') : Args(1) {
...
-            return;
-        } else {
-            return 1;
-        }
-    }
+sub report_edit : Path('report_edit') : Args(1) {
+    my ( $self, $c, $id ) = @_;
 
-    $c->forward('categories_for_point');
+    my $problem = $c->cobrand->problems->search( { id => $id } )->first;
```
struan: I assume this has to be `search` and not a `find` as the Resultset returned by `$c->cobrand->problems` prevents that, and `id` has to be unique so this will never return more than one row?

dracos: `find` would be fine here I think (though yes, I think search on primary key is basically equivalent), that's just the existing code (says it's part of the diff as it got moved around, `git diff --histogram` is a bit nicer on this commit for some reason).

PR #2019, opened 2018-03-12
Fix check for visible sub map links.

dracos:
I don't think this check could ever have worked, because items are invisible
if their parent is set to `display: none`, so it would hide the links bar if
it was already hidden, and show it if it was already shown.

Use a js- class for the movement of the feed item into the sub map links.


dracos:
This is a prime candidate for a Cypress regression test now we have that.

PR #2020, opened 2018-03-13
Add Cypress testing

dracos:
This adds some in-browser headless testing to the codebase.
The only test so far makes a new report (logging in while doing so), nothing complicated, but that will be testing map click with JavaScript.

I upgraded Test::PostgreSQL to one that didn't break the exit code of any script using it (so we can tell Travis whether cypress passes/fails), but that has its own bug whereby if postgres quits too quickly at the end it thinks it hasn't and needlessly waits 10 seconds (I've made a PR at https://github.com/TJC/Test-postgresql/pull/22 which I think fixes that), but that's just an annoyance rather than it always saying the tests passed.

To use, install cypress with `npm install cypress` (or `npm install -g cypress` if you're happy with where global installs go, I am on my machine, as it's in `~/.npm-global` ;) ) and then you can run `bin/run-cypress open` to open cypress interactively or `bin/run-cypress run` to run headlessly. All cypress commands work, they're just passed through after: running a new dev server on port 3001 with mocked MapIt, fixmystreet cobrand; compiling CSS; running fixture. 

Note OSM is blacklisted to prevent much running fetching many tiles, so that's why the tiles look broken when not run headlessly.


dracos:
"I noticed that it generates a .cypress directory in my project directory – should the whole of .cypress be in .gitignore?" - it's included as part of this PR - it contains the actual tests! (under `.cypress/cypress/integration/` :) )

Issue #2021, opened 2018-03-13
Find out ratio of people entering their postcode to view existing problems vs to report a new problem


zarino:
A core assumption that has appeared at the centre of recent tickets (#2016, #1462) is that people are viewing the `/around` map to _view_ reports just as much as they are to _create_ new reports. But @wrightmartin and I just realised we have _no data_ to back this argument up either way.

Let’s do some science!!

Quick amendment to the FMS.com homepage that replaces the single "GO" button with two buttons: "Report a new problem" and "Show problems". And then track clicks on those buttons in Google Analytics. Both buttons will just submit the form as normal, taking people to the `/around` page.

wrightmartin:
Ok, interesting science ahead. I might need to consult a member of the research team to verify the findings and my analytics are a _little_ rusty, but it looks like **more users are _reporters_ than _browsers_**.

ajparsons:
Storing here, there are still conversions coming from the browsing path, but less:

![image](https://user-images.githubusercontent.com/8157058/39368267-0c40cb50-4a31-11e8-9233-9bd4b3e5bb5d.png)

ajparsons:
Constructing another segment that gets people who have clicked BOTH buttons, this accounts for 784 reports - so the browse is still leading to conversions, but less. 

But also, there is then another behaviour that leads to people clicking both.

zarino:
Okay, so the answer to this ticket was: Roughly **2 "browsers"** for every **3 "reporters"** … although, some people who intially click "browse" go on to actually make a report.

That’s good enough for me!

dracos:
Some reporters may then not report anything, too, of course.

wrightmartin:
> Some reporters may then not report anything, too, of course.

About 8% according to the exit numbers

![image](https://user-images.githubusercontent.com/2292925/39534682-1221e0b6-4e2a-11e8-9a46-f0d630493e5a.png)
8%

PR #2031, opened 2018-03-16
[UK] Dual submit buttons on homepage postcode form

zarino:
Present "report" vs "browse" choice at time of postcode entry, so we can track intent, as part of a temporary experiment into what people expect to see on the fixmystreet.com/around page.

Google Analytics will be able detect which button was pressed by looking for the `intent=report` or `intent=report` query string param in the resulting pageview location.

We supply a *third* submit button, immediately before the other two, and hidden from visual users and assistive devices, to prevent either of the "intent" submit values being sent when the user submits the form via the Return/Enter key.

Part of #2021. [skip changelog]

# desktop

screen shot 2018-03-20 at 11 44 15

# mobile 

screen shot 2018-03-20 at 11 45 34

### DIFF
--- a/templates/web/fixmystreet.com/around/postcode_form.html
+++ b/templates/web/fixmystreet.com/around/postcode_form.html
@@ -1,0 +1,34 @@
+<div id="front-main">
+    <div id="front-main-container">
+        [% UNLESS possible_location_matches %]
+            [% INCLUDE 'around/intro.html' %]
+        [% END %]
+
+        [%
+            question = c.cobrand.enter_postcode_text || loc('Enter a nearby street name and area');
+        %]
+
+        <form action="[% c.uri_for('/around') %]" method="get" name="postcodeForm" class="js-geolocate postcode-form-test">
+            <label for="pc">[% question %]:</label>
+            <input type="text" name="pc" value="[% pc | html %]" id="pc" size="10" maxlength="200" placeholder="[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]" class="postcode-form-test__postcode">
+            <a href="[% c.uri_for('/around') %]" id="geolocate_link" class="btn btn--geolocate">[% loc('or locate me automatically') %]</a>
+
+            [%# Form will be submitted via this button when user presses Enter key %]
+            <button type="submit" class="visuallyhidden" aria-hidden="true" tabindex="-1">Go</button>
+
+            <div class="postcode-form-test__buttons">
+                [%# We can now segment users in Google Analytics by looking at "intent" query string param %]
+                <button type="submit" name="intent" value="report" class="btn">[% loc('Report a new problem') %]</button>
+                <button type="submit" name="intent" value="browse" class="btn">[% loc('Show problems') %]</button>
+            </div>
+
+            [% IF partial_token %]
+                <input type="hidden" name="partial" value="[% partial_token.token %]">
+            [% END %]
+
+            [% IF c.user_exists AND c.user.categories.size %]
+              <input type="hidden" name="filter_category" value="[% c.user.categories.join(",") | html %]">
+            [% END %]
+        </form>
+    </div>
+</div>
--- a/web/cobrands/fixmystreet.com/base.scss
+++ b/web/cobrands/fixmystreet.com/base.scss
@@ -299,3 +299,67 @@ html.lazyload .js-lazyload {
     margin-bottom: 0;
   }
 }
+
+.postcode-form-test {
+    margin: 0 -1em;
+    padding: 1em;
+    background-color: $primary;
+    color: $primary_text;
+
+    label {
+        margin: 0;
+    }
+
+    // Lots of !important flags here to overpower
+    // selectors like `#front-main a#geolocate_link` and
+    // `body.fullwidthpage #front-main a#geolocate_link`
+    // in the core and fixmystreet.com styles.
+    #geolocate_link {
+        background: transparent !important;
+        border: none;
+        height: auto !important;
+        margin-top: 0 !important;
+        padding: 1em !important;
+
+        &:before {
+            vertical-align: -0.3em;
+        }
+    }
+}
+
+// Overloaded selector to beat input[type="text"]
+input.postcode-form-test__postcode {
+    max-width: 100%;
+    width: 30em;
+    margin: 0.5em auto 0 auto;
+    padding: 0.75em;
+    border: 1px solid darken($primary, 10%);
+    box-shadow: 0 1px 1px rgba(#000, 0.2);
+}
+
+.postcode-form-test__buttons {
+    margin: 1em -10px 0 -10px;
+
+    // Browsers without flex support will just lay out the buttons
+    // side-by-side, breaking onto two lines at narrow widths.
+    @include flex-container();
+    @include justify-content(center);
+
+    $btn-blue: #469bfc;
+
+    .btn {
+        color: #fff !important;
+        background: $btn-blue;
+        background-image: linear-gradient($btn-blue, darken($btn-blue, 5%));
+        border: none;
+        border-radius: 2px;
+        text-shadow: 0 1px 1px rgba(0,0,0,0.2);
+        box-shadow: 0 1px 1px rgba(0,0,0,0.2);
+        margin: 0 10px 10px 10px;
+
+        &:hover, &:focus {
+            background-color: darken($btn-blue, 5%);
+            background-image: linear-gradient(darken($btn-blue, 5%), darken($btn-blue, 10%));
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


templates/web/fixmystreet.com/around/postcode_form.html
```diff
@@ -0,0 +1,34 @@
...
+            question = c.cobrand.enter_postcode_text || loc('Enter a nearby street name and area');
+        %]
+
+        <form action="[% c.uri_for('/around') %]" method="get" name="postcodeForm" class="js-geolocate postcode-form-test">
+            <label for="pc">[% question %]:</label>
+            <input type="text" name="pc" value="[% pc | html %]" id="pc" size="10" maxlength="200" placeholder="[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]" class="postcode-form-test__postcode">
+            <a href="[% c.uri_for('/around') %]" id="geolocate_link" class="btn btn--geolocate">[% loc('or locate me automatically') %]</a>
+
+            [%# Form will be submitted via this button when user presses Enter key %]
+            <button type="submit" class="visuallyhidden" aria-hidden="true" tabindex="-1">Go</button>
```
dracos: Why `button` and not `input`?

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -299,3 +299,57 @@ html.lazyload .js-lazyload {
     margin-bottom: 0;
   }
 }
+
+.postcode-form-test {
+    margin: 0 -1em;
+    padding: 1em;
+    background-color: $primary;
```
dracos: This is missing a colour.

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -299,3 +299,57 @@ html.lazyload .js-lazyload {
...
+        padding: 1em !important;
+
+        &:before {
+            vertical-align: -0.2em;
+        }
+    }
+}
+
+// Overloaded selector to beat input[type="text"]
+input.postcode-form-test__postcode {
```
dracos: The border of this now doesn't look like other text inputs (or itself before), is that deliberate?

templates/web/fixmystreet.com/around/postcode_form.html
```diff
@@ -0,0 +1,34 @@
...
+
+            <div class="postcode-form-test__buttons">
+                [%# We can now segment users in Google Analytics by looking at "intent" query string param %]
+                <button type="submit" name="intent" value="report" class="btn">[% loc('Report a new problem') %]</button>
+                <button type="submit" name="intent" value="browse" class="btn">[% loc('Show problems') %]</button>
```
dracos: At narrow width these go onto two lines and mean the instructions are pushed out of sight - can we keep them on one line perhap?

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -299,3 +299,57 @@ html.lazyload .js-lazyload {
...
+    margin: 0 -1em;
+    padding: 1em;
+    background-color: $primary;
+
+    label {
+        margin: 0;
+    }
+
+    #geolocate_link {
+        background: transparent !important;
```
dracos: Using `!important`, should be a comment explaining why.

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -169,6 +169,23 @@ body.fullwidthpage {
     }
 }
 
+.postcode-form-test {
+    background: transparent;
+    padding-top: 1.5em;
+}
+
+input.postcode-form-test__postcode {
+    padding: 0.5em 1em;
```
dracos: This is the same.

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -169,6 +169,23 @@ body.fullwidthpage {
...
+    margin-top: 0.5em;
```
dracos: I think this value would be fine for the base too? 1em feels a bit big there.

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -169,6 +169,23 @@ body.fullwidthpage {
...
+}
+
+.postcode-form-test__buttons {
+    .btn {
+        font-size: 1.2em;
+        font-family: MuseoSans,Helmet,Freesans,sans-serif;
```
dracos: `$body-font`?

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -299,3 +299,57 @@ html.lazyload .js-lazyload {
...
+}
+
+.postcode-form-test__buttons {
+    margin: 1em -10px 0 -10px;
+
+    $btn-blue: #469bfc;
+
+    .btn {
+        color: #fff !important;
+        background: $btn-blue linear-gradient($btn-blue, darken($btn-blue, 5%));
```
dracos: Does this need a fallback? We have a LG mixin. And a btn mixin.

templates/web/fixmystreet.com/around/postcode_form.html
zarino: No reason! Just because I created this element by copying one of the `button`s from a few lines further below – and _they_ were `<button>` elements rather than `<input>`s because then I could have their form value (eg: `browse`) be different from their visual text content (eg: "Show problems").

Can you think of a reason we’d prefer inputs, over buttons, in one or both of these places?

dracos: Nah, I'm just very old and remember browsers that support input and not button, but they'll all be useless if anyone's still using them.

zarino: Mmmhmmm. For the last few months I’ve been trying to use `<button>`s wherever I can in forms. Mainly because, unlike `<input>`s, they can have `:before` and `:after` pseudo-children, which is handy for adding icons to buttons without messing around with background positioning and i18n.

And the nice thing about `<button>`s is that they don’t _have_ to live in forms. So they’re a nice, more natively accessible drop-in replacement for the `<a href="#">` that we’d typically attach JavaScript listeners to (although not very often in FMS!).

So then you’ve basically got one element you use _anywhere_ you want some discrete, button-like clickable element. 🎉

web/cobrands/fixmystreet.com/layout.scss
zarino: Hilariously, `$body-font` is defined in `sass/_base.scss`, which means it’s not available to `fixmystreet.com/layout.scss`. I guess this is why `MuseoSans,Helmet,Freesans,sans-serif` is hardcoded elsewhere in this file.

This feels like a bit of a thorny problem to solve as part of this PR. (We might want to add a `$body-font` declaration to `sass/_layout.scss`, or we might want to spin the font variables out into a `_fonts.scss` file like some of our other cobrands do.) Shall I create a new ticket for it?

dracos: Oh, sorry!

zarino: Actually, you know what, I’ll just add `$body-font: MuseoSans,Helmet,Freesans,sans-serif` to the top of `fixmystreet.com/layout.scss` – at least then we cut down on the duplication, and leave the door open for someone to come along and refactor the variables out in the future.

web/cobrands/fixmystreet.com/base.scss
zarino: Yes, intentional. It looked a bit weird before. This brings it closer to our normal input styling, but I’ve swapped the inner shadow for a drop shadow, to make it stand out from the yellow background.

web/cobrands/fixmystreet.com/base.scss
zarino: Done!

web/cobrands/fixmystreet.com/base.scss
zarino: Done!

templates/web/fixmystreet.com/around/postcode_form.html
zarino: Good suggestion :+1: Turns out flexbox makes this one a doozy. Done in `fixmystreet.com/base.scss` -> `.postcode-form-test__buttons`

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -299,3 +299,67 @@ html.lazyload .js-lazyload {
...
+    }
+}
+
+// Overloaded selector to beat input[type="text"]
+input.postcode-form-test__postcode {
+    max-width: 100%;
+    width: 30em;
+    margin: 0.5em auto 0 auto;
+    padding: 0.75em;
+    border: 1px solid darken($primary, 10%);
```
dracos: Poor `$primary_b`, its only use, supplanted ;)

web/cobrands/fixmystreet.com/base.scss
zarino: Let the record show – it turns out I’d missed a `border:` rule, so browsers were doing inconsistent things with the input. Turns out Chrome’s idea of a default input border is a little… garish 😉

Issue #2016, opened 2018-03-12
Reduce confusion of 'click on map' banner


crowbot:
Possible solutions from [summary in #1462](https://github.com/mysociety/fixmystreet/issues/1462#issuecomment-364899132):

* Moving the green banner on top of the map (idea from comment in #1519, screenshot below);
* Have clicking the green banner, where it is or moved from above, hide/show the pins (or even just on hover hide them, have the bar be some form of toggling interface) (and perhaps alert user further they should then click on map);
* Replacing the green banner with a draggable pin and instructions to drag it on to the map to report a problem - starting a drag immediately hides all map pins. Clicking would still work too;
* Having an always visible clear toggle between "viewing" mode and "reporting" mode that does the sae as hide/show pins but both options always visible - like a Map/Satellite layer picker, maybe? Anyway, as long as it's clear there are two modes and you can switch between them. Clicking in view mode would still work too.

![c4b4b632-7fdb-11e6-957a-7109251e1d59](https://user-images.githubusercontent.com/2912/37280470-c8661b02-25e5-11e8-925f-db164311af2a.png)

zarino:
Another, slightly different, suggestion:

* Embrace the fact that people will click the green banner. And if they click it, just act as if they’d clicked the centre of the map. Create a new report at the centre of the viewport. They can always move the pin by clicking the map again (there’s an instruction about that right near the top of the reporting form).

wrightmartin:
After 39 days of running the experiment, there's little difference between the two, in fact a hair's breadth in favour of keeping it as is

![screen shot 2018-04-27 at 15 00 18](https://user-images.githubusercontent.com/2292925/39366438-d44cc7bc-4a2b-11e8-9777-d32bb9a96ce0.png)

wrightmartin:
This test did come immediately after we added the 'click the green box to start the report' functionality - which may have helped with some of the issues we've seen with the reporting process

abimysoc:
So placing the banner over the map doesn't make it any more or less noticeable? Bah.

Maybe #2012 is the answer (accept that people are going to click things other than the map, and provide a simple way for them to get back on the reporting path).

zarino:
Ok, so moving the banner over the map made no big difference. I’ve opened up a ticket for reverting that change in #2115.

[...]

Issue #1519, opened 2016-09-07
Make "click map to report a problem" instruction much more obvious


abimysoc:
It is time I opened a ticket for this: often, when I look at FixMyStreet my eyes _completely skim over_ the big white-on-green banner with the essential instruction "click map to report a problem".

It's clear to me that this is the cause of the problems outlined in #536 & #1462 (people updating existing reports instead of making new ones), and #1072 (people mistaking the 'report abuse' mechanism for how to report a problem). We also see user support emails along the lines of "I can see all the pins/reports on the map, but how do I make a new one?".

The appearance and placement of the banner need revisiting. It'd be an excellent candidate for some A/B testing. Would it help to make it look more like a button? Change the colour scheme?

davewhiteland:
Maybe in these modern times it could do something dynamic (e.g. enlarge a weeny bit and bounce back to size half a second after page load). Honesty don't know, but if the problem is related to there being so much going on the screen (my guess) this might be worth considering as a way to elevate the primary call to action. Dunno. Just a suggestion.

wrightmartin:
Here's the result of 5 minutes chat, we'll follow this up with a better mockup and some details about the UX 

![group](https://cloud.githubusercontent.com/assets/2292925/18441942/c05c63cc-7906-11e6-9973-31c8c63ca179.png)

wrightmartin:
Here's solution – part one to this problem, we make the 'location' a more present part of the form as below, when you click the map it changes the location. This way it's harder to miss the 'select a location' step of the process.

**No location selected**
![1 no location](https://cloud.githubusercontent.com/assets/2292925/18475501/8cf9fc34-79be-11e6-8718-139724c6c1fa.png)

**Click map to select location**
![untitled1](https://cloud.githubusercontent.com/assets/2292925/18475515/99819d0e-79be-11e6-8160-2bd3dffead0b.gif)

**Location set**
![7](https://cloud.githubusercontent.com/assets/2292925/18475516/998b29a0-79be-11e6-96bc-08b6515facae.png)

wrightmartin:
See also https://github.com/mysociety/fixmystreetforcouncils/issues/83, it seems @zarino has already been down this path

abimysoc:
:warning: You're one step too far into the process. It's possible to fail at the first hurdle without ever seeing the problem reporting form.

Start at the homepage and enter a postcode as instructed, and you'll end up at a page that looks like:

[...]

PR #2032, opened 2018-03-16
[UK] A/B test "big green banner" position

zarino:
Part of #2016. [skip changelog]

# medium screens

screenshot-2018-3-16 viewing a location 1

# wider screens

screenshot-2018-3-16 viewing a location

# very wide screens

screenshot-2018-3-16 viewing a location 2

### DIFF
--- a/templates/web/base/around/display_location.html
+++ b/templates/web/base/around/display_location.html
@@ -73,6 +73,12 @@
             <a href="/">[% loc('Home') %]</a>
             <span>[% loc('Place pin on map') %]</span>
         </div>
+
+      [% IF c.config.BASE_URL == "https://www.fixmystreet.com" %]
+        <h1 class="big-green-banner banner-position-variant-1">
+            [% loc( 'Click map to report a problem' ) %]
+        </h1>
+      [% END %]
     </div>
 
 
--- a/templates/web/fixmystreet.com/header_extra.html
+++ b/templates/web/fixmystreet.com/header_extra.html
@@ -2,6 +2,22 @@
 <link rel="prefetch" href="[% version('/vendor/OpenLayers.Projection.OrdnanceSurvey.js') %]">
 [% END %]
 
+[% IF bodyclass.match('mappage') AND c.config.BASE_URL == "https://www.fixmystreet.com" %]
+<script src="//www.google-analytics.com/cx/api.js?experiment=jydOGgMrQRaWedSNoiW-SQ"></script>
+<script nonce="[% csp_nonce %]">
+    var variation = typeof cxApi !== "undefined" ? cxApi.chooseVariation() : 0,
+        docElement = document.documentElement,
+        className = docElement.className;
+    docElement.className = className + ' banner-position-variant-' + variation;
+</script>
+<style>
+    html .banner-position-variant-1 { display: none !important; }
+    html.banner-position-variant-1 .banner-position-variant-0 { display: none !important; }
+    html.banner-position-variant-1 .banner-position-variant-1 { display: block !important; }
+    html.banner-position-variant-1 #side .big-green-banner ~ section.full-width { margin-top: -1em; }
+</style>
+[% END %]
+
 [% IF bodyclass.match('frontpage') %]
 <link rel="prefetch" href="[% version('/cobrands/fixmystreet.com/js.js') %]">
 [% END %]
--- a/templates/web/fixmystreet.com/around/_report_banner.html
+++ b/templates/web/fixmystreet.com/around/_report_banner.html
@@ -1,0 +1,6 @@
+<h1 class="big-green-banner banner-position-variant-0">
+    [% loc( 'Click map to report a problem' ) %]
+</h1>
+<a id="skip-this-step" href="[% url_skip %]" rel="nofollow">
+    [% loc("Can't see the map? <em>Skip this step</em>") %]
+</a>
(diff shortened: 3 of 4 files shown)


templates/web/fixmystreet.com/header_extra.html
```diff
@@ -2,6 +2,10 @@
 <link rel="prefetch" href="[% version('/vendor/OpenLayers.Projection.OrdnanceSurvey.js') %]">
 [% END %]
 
+[% IF bodyclass.match('mappage') %]
```
dracos: I'd have a `AND c.config.BASE_URL == "https://www.fixmystreet.com"` here too so it only happens on live, not development.

web/cobrands/fixmystreet.com/big-green-banner-ab-test.js
```diff
@@ -0,0 +1,52 @@
+(function($){
+    var variation = 0;
+    if ( typeof cxApi !== "undefined" ) {
+        variation = cxApi.chooseVariation();
+    }
+
+    if (variation === 1) {
+        var $original = $('#side .big-green-banner');
+        var $clone = $original.clone();
```
dracos: This seems like a very complicated way of going about this :) Would it not be far less code to behave as 4afd0026262dd74425d40ac78649db3d8148c883 did – have another `big-green-banner` in the HTML inside `map_box` with class variant1, add class variant0 to the original banner, have this CSS:
```
-html .variant1 { display: none !important; }
-html.variant1 .variant0 { display: none !important; }
-html.variant1 .variant1 { display: block !important; }
```
and set variant1 on `html` if it is that variant. Everything appears/disappears as it should, no more JS.

zarino: Interesting! I'll give it a try!

templates/web/base/around/display_location.html
```diff
@@ -73,6 +73,10 @@
             <a href="/">[% loc('Home') %]</a>
             <span>[% loc('Place pin on map') %]</span>
         </div>
+
+        <h1 class="big-green-banner banner-position-variant-1">
```
dracos: This needs a .com only IF around it.

templates/web/fixmystreet.com/header_extra.html
```diff
@@ -2,6 +2,22 @@
...
+        className = docElement.className;
+    docElement.className = className + ' banner-position-variant-' + variation;
+</script>
+<style>
+    html .banner-position-variant-1 { display: none !important; }
+    html.banner-position-variant-1 .banner-position-variant-0 { display: none !important; }
+    html.banner-position-variant-1 .banner-position-variant-1 { display: block !important; }
+    html.banner-position-variant-1 #side .big-green-banner ~ section.full-width { margin-top: -1em; }
+</style>
+[% END %]
```
dracos: The style needs to be outside of (at least some of) the `IF`, otherwise non-live sites will have both buttons.

dracos:
So we're just testing whether clicks are more/less depending upon the position; not attached to a goal of creating a report and seeing if that's more/less?

zarino:
> So we're just testing whether clicks are more/less depending upon the position; not attached to a goal of creating a report and seeing if that's more/less?

If we deploy #2033 at the same time as this A/B test (don’t _have_ to, I guess, but seems a shame to hold off on a UX improvement) then clicks on the banner(s) _will_ count towards new reports being created, which isn’t really what we want to measure. We want to know whether moving the banner makes people less likely to click the banner. 

Originally, I thought maybe we’d measure this by looking for the variant that resulted in _the least_ `big-green-banner` events (#2001) – but the way GA Content Experiments are presented, it seemed clearer to me to test for the _presence_ of a goal rather than the _absence_ of one. So I went for the other side of the coin – testing which variant results in _the most_ clicks on the map (since that avoids the "report created" goal that could be influenced by #2033, as above).

Do you think that makes any sense?

templates/web/fixmystreet.com/header_extra.html
zarino: Except, if I add the `fixmystreet.com` check around `<h1 class="big-green-banner banner-position-variant-1">` in `display_location.html`, as you suggested above, then I assume I won’t need to worry about this, since the `.banner-position-variant-1` banner will only appear on fixmystreet.com?

dracos: Oh, sorry, for the `IF` I mentioned I assumed it'd be `IF c.cobrand.moniker == 'fixmystreet'` - otherwise you've made it impossible to do any development on it ;) But yes, if it's a "live-only" `IF` then you're right, it just makes further development a bit harder (though I assume there won't be any/much).

dracos:
"clicks on the banner(s) will count towards new reports being created" - I don't think they will, the creation event only fires right at the end of the process if a report is actually created, as I understand it.

"We want to know whether moving the banner makes people less likely to click the banner." – well, we want to know whether moving the banner makes people more likely to use the site successfully for their goal; I'm not sure measuring whether the banner gets clicked or not is a valid proxy for that (say there are fewer clicks over the map, is that because people are now clicking the map okay or not clicking it at all?). But we can still look at it if you think it would be useful :)

templates/web/fixmystreet.com/header_extra.html
zarino: Haha! Yeah, I just disabled that part of the check when I needed to try it out locally. Figured we probably don’t want development installs sending hits back to our GA experiment. (Although now I think about it, maybe this is _already_ mitigated by whatever we already have in place to prevent GA pageviews/events being recorded in development. Hmmm…)

zarino:
You’re right, now I‘ve dug back, I notice the _original_ original ticket that inspired this experiment (#1519) is much less "click-the-map" focussed than #2016, and more just "make the reporting flow more obvious" – so using "reports created" as our experiment success criteria makes sense. I’ll switch up the Content Experiment settings now.

PR #2033, opened 2018-03-19
Clicking "big green banner" on /around begins a report

zarino:
Data collected by #2001 has shown that some people are clicking the big green "Click map to report a problem" banner on fixmystreet.com/around. We assume the same is true for other cobrands.

As suggested in #2016, this commit adds a click handler to that banner element, which will begin a new report at the centre of the screen. It might not be the exact right location, but the pin can be repositioned from the `/report/new` form, and beginning a report is better than just soaking up the click and doing nothing.

### DIFF
--- a/.cypress/cypress/integration/simple_spec.js
+++ b/.cypress/cypress/integration/simple_spec.js
@@ -14,3 +14,17 @@ describe('My First Test', function() {
         cy.get('form').submit();
     });
 });
+
+describe('Clicking the "big green banner" on a map page', function() {
+    before(function() {
+        cy.visit('/');
+        cy.get('[name=pc]').type('BS10 5EE');
+        cy.get('#postcodeForm').submit();
+        cy.get('.big-green-banner').click();
+    });
+
+    it('begins a new report', function() {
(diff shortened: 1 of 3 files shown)


zarino:
> only question is if the pointer should now change on hovering on the green banner; guessing not, but could do?

I'm not sure what the advantage is to adding a visual affordance for the clickability of the banner. We don’t really want to _encourage_ people to click it. We just want to help them get to the right place if they make the mistake of clicking it ;-)

dracos:
Test failure is because the test fixture sets up area 2608 (South Gloucestershire) to have categories etc, so you'd have to go there for form_category to exist. Or check for something else in the sidebar :)

dracos:
Though the merge passes, perhaps it was some bug in the fixture I fixed on master; see if a rebase fixes it?

dracos:
Or perhaps it's just a race issue, whether the server response (that removes the category dropdown) replies before the test it's there fires. Anyway, I'd change it to check for something other than `form_category`.

zarino:
Yep – will change it to a different form element 👍

PR #2036, opened 2018-03-20
[fixmystreet.com] Add admin page to list refused.

dracos:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Stats.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Stats.pm
@@ -72,4 +72,27 @@ sub questionnaire : Local : Args(0) {
     return 1;
 }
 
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Admin/Stats.pm
```diff
@@ -72,4 +72,24 @@ sub questionnaire : Local : Args(0) {
     return 1;
 }
 
+sub refused : Local : Args(0) {
+    my ($self, $c) = @_;
+
+    my $contacts = $c->model('DB::Contact')->not_deleted->search({ email => 'REFUSED' }, { prefetch => 'body' });
```
struan: Should this also not cover devolved contact with the send_method set to Refused?

PR #2037, opened 2018-03-21
Make homepage #pc input required

zarino:
Will give some feedback to people who click one of the submit buttons without first entering a location to search for.

screen shot 2018-03-21 at 08 54 12


zarino:
@dracos – in Base now, as well as a handful of other cobrands (eg: demo.fixmystreet.com) that had their own `postcode_form.html` overrides. :+1:

PR #2044, opened 2018-03-26
[Bromley] Asset layers and related

struan:
This PR does asset-related things for Bromley Council, see https://github.com/mysociety/fixmystreet-commercial/issues/1005

It adds bins, street lights, trees, TfL red routes, and rights of way layers. The first three are point layers that can be selected, the red routes are used in certain categories to explain/divert to TfL (unless asset selected), and rights of way are displayed and ID taken through. In order to have multiple 'road' type layers, the USRN/road code is refactored into its own Layer class, which shouldn't affect anything but needs some testing.

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -301,6 +301,9 @@ function find_matching_feature(feature, layer, asset_id_field) {
 }
 
 function check_zoom_message_visibility() {
+    if (this.fixmystreet.non_interactive) {
+        return;
+    }
     var category = $("#problem_form select#form_category").val(),
         prefix = category.replace(/[^a-z]/gi, ''),
         id = "category_meta_message_" + prefix,
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -238,6 +238,12 @@ function asset_selected(e) {
         return;
     }
 
+    var layer = e.feature.layer;
+    var feature = e.feature;
+
+    // Keep track of selection in case layer is reloaded or hidden etc.
+    selected_feature = feature.clone();
+
     // Pick up the USRN for the location of this asset. NB we do this *before*
     // handling the attributes on the selected feature in case the feature has
     // its own USRN which should take precedence.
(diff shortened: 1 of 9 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -16,139 +16,91 @@ var fixmystreet = fixmystreet || {};
...
-        if (layer && layer.fixmystreet.actions) {
-            layer.fixmystreet.actions.not_found(layer);
+    road_not_found: function() {
+        if (this.fixmystreet.actions) {
+            this.fixmystreet.actions.not_found(this);
         } else {
-            if ( fixmystreet.roads.last_road && fixmystreet.roads.last_road.fixmystreet.actions.unselected ) {
-                fixmystreet.roads.last_road.fixmystreet.actions.unselected();
-                fixmystreet.roads.last_road = null;
+            if ( this.fixmystreet.actions.unselected ) {
```
struan: If there is no `this.fixmystreet.actions` then surely this will never be true?

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,139 +16,91 @@ var fixmystreet = fixmystreet || {};
...
-            if (feature == null) {
-                // The click wasn't directly over a road, try and find one
-                // nearby
-                feature = usrn_layer.getNearestFeature(point, 10);
-            }
-            if (feature !== null) {
-                selected_usrn = feature.attributes[usrn_layer.fixmystreet.usrn.attribute];
+    checkFeature: function(evt, lonlat) {
+        this.getNearest(lonlat);
+        this.updateUSRNField();
```
struan: This working relies, I think, on us only ever having one layer for each USRN, which I expect we would, but it might be worth making that explicit somewhere.

web/cobrands/fixmystreet/assets.js
dracos: Hmm, yes. I admit I didn't really understand this. `not_found` is either called right after `last_road` was set (and so no need to look it up), or if the road layer had disappeared, and I wasn't sure how that could happen. It doesn't look like `unselected` would ever be called?

web/cobrands/fixmystreet/assets.js
dracos: I'm not sure what you mean, sorry. You could have two layers with `usrn: { attribute: "...", field: "..." }`  and they would both work independently?

web/cobrands/fixmystreet/assets.js
struan: The purpose is that if you need to do cleanup when a feature is unselected (e.g. hiding some additional text displayed) it is run. Previously this worked by storing the last selected layer and running unselected on that in not_found. In this case we shouldn't need to do that as each layer recieves the event so can call it's own unselected function.

web/cobrands/fixmystreet/assets.js
struan: But if both of those had the same `field` set then then they might do conflicting things when updating the field.

dracos:
Only changes from Bromley were to use FEATURE_ID not CENTRAL_AS, and to add another TfL category.

PR #2047, opened 2018-03-27
Gitignore cypress.env.json

zarino:
The Cypress docs suggest putting machine-specific config into a
`cypress.env.json` file at the same location as `cypress.json`.
(For example, if you want to disable video recording of test sessions
on your machine, but not everyone else’s.) Since this file will be
machine-specific, we don’t want it tracked in Git.

[skip changelog]


zarino:
Mmm, good point. More explicit is good! Will change it.

PR #2048, opened 2018-03-28
Allow response templates to be triggered by Confirm status code.

davea:
Fixes mysociety/fixmystreetforcouncils#297. As mentioned in that ticket, Confirm has a more fine-grained status code system than FMS which we want to use to trigger response templates/emails to users.

This PR:

 - Fetches the `external_status_code` attribute from Open311 service request updates
 - Allows response templates to be associated with an external status code instead of an FMS state
 - Triggers response templates by `external_status_code` when fetching updates.

### DIFF
--- a/perllib/Open311/GetServiceRequestUpdates.pm
+++ b/perllib/Open311/GetServiceRequestUpdates.pm
@@ -110,6 +110,7 @@ sub update_comments {
 
             if ( !$c->first ) {
                 my $state = $open311->map_state( $request->{status} );
+                my $external_status_code = $request->{external_status_code};
                 my $comment = $self->schema->resultset('Comment')->new(
                     {
                         problem => $p,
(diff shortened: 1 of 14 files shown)


perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -110,12 +110,13 @@ sub update_comments {
...
             if ( !$c->first ) {
                 my $state = $open311->map_state( $request->{status} );
+                my $external_status_code = $request->{external_status_code};
                 my $comment = $self->schema->resultset('Comment')->new(
                     {
                         problem => $p,
                         user => $self->system_user,
                         external_id => $request->{update_id},
-                        text => $self->comment_text_for_request($request, $p, $state),
+                        text => $self->comment_text_for_request($request, $p, $state, $external_status_code),
```
dracos: This is added here, but the fn isn't updated until a different commit (so I guess this line change should wait until then, or you might be squashing them together in which case doesn't matter!).

davea: Oops, yes, that should be in the later commit.

PR #2051, opened 2018-03-28
[UK] Improving access to council summary dashboards

zarino:
# Login page

screenshot-2018-4-4 dashboard request access

# "Submit" email footer

screen shot 2018-04-04 at 12 48 57

# FMS Pro promo at bottom of homepage and `/reports` page

screen shot 2018-04-04 at 12 50 47

# Top corner of `/reports/<area>` page

screen shot 2018-04-04 at 12 51 41

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1016.

[skip changelog]


MyfanwyNixon:
@zarino Just a reminder (as discussed) to look at campaign tracking for the `submit.html` email link.

zarino:
> I would wonder if any tracking in email link would put people off clicking (it would me), compared to what value it may give. We can see how many visited the page without a referrer already, presumably.

The aim is to get a handle on whether the `submit.html` email is sending us any visitors. We can obviously track pageviews on `/about/council-dashboard`, and we can see which _other_ pages brought people there, but right now, incoming visitors from email is a black hole. 😞 

Tracking whether the link in `submit.html` generates any clicks will help us measure the effectiveness of our messaging there (like: do the people receiving these emails actually care about their organisation’s responsiveness and how it compares with other councils?) – and if we see that generating very few clicks, we can try tweaking the value proposition.

I agree tracked links in emails can look a bit scummy, but _at least_ these are just `utm` params added onto the real URL, rather than those horrible obfuscated / shortened links that services like Mailchimp send out.

dracos:
"incoming visitors from email is a black hole. 😞" surely any visit there without a referrer is from this email. It's unlikely to be from anywhere else ;)

PR #2054, opened 2018-03-29
Issues/2005 ie chart fixes

### LINKED ISSUE #2005: [fixmystreet.com] Check chart.js dashboard charts in IE8–11
I don’t think chart.js works at all in IE8, so we should check that the `/reports` and `/reports/<area>/summary` pages work fine without it.

There also look to be some chart.js rendering issues in IE9, and in IE10–11 the big "pull out" numbers on the chart at the top of `/reports` are way too far right, suggesting either a CSS issue, or a chart.js issue.

wrightmartin:
Fixes #2005 

All charts render in IE10+ and look fine, IE9 and below get the same content just not presented in quite the same way, but it doesn't look broken

### DIFF
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -98,6 +98,8 @@
         position: absolute;
         margin-top: -1em;
         margin-#{$right}: 0;
+        right: 0;
+        width: 18%; // 20% padding from .labelled-line-chart, 2% gutter
     }
 }
 
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -63,7 +63,6 @@
 
     canvas {
         width: 100% !important;
-        height: auto !important;
     }
 }
 
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -329,3 +329,23 @@
         margin-top: -1px;
     }
 }
+
+.ie9 {
+    .labelled-line-chart,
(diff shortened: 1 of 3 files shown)


web/js/dashboard.js
```diff
@@ -1,276 +1,276 @@
 $(function(){
+    if (!$('html').is('.ie9')) {
```
zarino: Rather than indenting the entire contents of this `$(function(){` block, how about just bailing out early? Like…

```javascript
$(function(){
    if ( $('html').is('.ie9') ) {
        return;
    }

    Chart.defaults.global.defaultFontSize = 16;
    …
```

web/cobrands/sass/_dashboard.scss
```diff
@@ -60,7 +60,6 @@
 
     canvas {
         width: 100% !important;
-        height: auto !important;
```
zarino: I think you missed the word "height" out of your commit message for this one!

web/cobrands/sass/_dashboard.scss
```diff
@@ -95,6 +95,8 @@
         position: absolute;
         margin-top: -1em;
         margin-#{$right}: 0;
+        right: 0;
+        width: 18%;
```
zarino: Could we maybe get a comment here to explain the `18%` magic number? I assume it’s the `padding-right: 20%` from `.labelled-line-chart` minus an extra `2%` gutter?

web/cobrands/sass/_dashboard.scss
wrightmartin: Fixed!

PR #2055, opened 2018-04-02
display cobrand assets on the national map

struan:
Mechanism to add assets to the national map and some extra config to only display assets that are relevant to the bodies that the map covers.

[skip changelog]

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1086


t/cobrand/national_assets.t
```diff
@@ -0,0 +1,41 @@
...
+        ALLOWED_COBRANDS => [ 'fixmystreet' ],
+        MAPIT_URL => 'http://mapit.uk/',
+        COBRAND_ASSETS => [ 'bromley' ]
+    }, sub {
+        $mech->get_ok("/report/new?latitude=51.494885&longitude=-2.602237");
+        $mech->content_lacks('buckinghamshire/assets.js');
+    };
+};
+
+subtest 'cobrand assets includes not applied on cobrand sites', sub {
```
davea: Should there be a test that ensures that a cobrand's `assets.js` is included in the cobrand even if `COBRAND_ASSETS` is/isn't set?

web/cobrands/buckinghamshire/assets.js
```diff
@@ -0,0 +1,193 @@
...
+    },
+    stylemap: new OpenLayers.StyleMap({
+        'default': highways_style
+    }),
+    always_visible: true,
+    non_interactive: true,
+    road: true,
+    asset_item: 'road',
+    all_categories: true,
+    actions: {
```
davea: These functions seem to be getting called everywhere you click on the map, which means trying to make a report anywhere that's not a Bucks road results in the `not_found` function disabling the report form. e.g. trying to make a report in BANES:

bucks_a_bit_keen

web/cobrands/buckinghamshire/assets.js
```diff
@@ -0,0 +1,193 @@
...
+    symbolizer: {
+        strokeColor: "#555555"
+    }
+});
+highways_style.addRules([rule_owned, rule_not_owned]);
+
+function show_responsibility_error(id) {
+    hide_responsibility_errors();
+    $("#js-bucks-responsibility").removeClass("hidden");
+    $("#js-bucks-responsibility .js-responsibility-message").addClass("hidden");
```
davea: These elements aren't present in the fms.com template, so no message is being shown.

web/cobrands/buckinghamshire/assets.js
```diff
@@ -0,0 +1,193 @@
...
+                hide_responsibility_errors();
+                enable_report_form();
+            } else {
+                // User has clicked a road that Bucks don't maintain.
+                show_responsibility_error("#js-not-bucks-road");
+                disable_report_form();
+            }
+        },
+
+        not_found: function(layer) {
```
davea: Should this function (and `found` above) change its behaviour according to which cobrand is in effect? I think it might be out of the scope of this PR to implement the "your report isn't on a Bucks road, so we won't send it to them, but it will be displayed on FMS.com" feature, but perhaps it should only be active if the current cobrand is Buckinghamshire.

struan: Yeah, I was wondering if it makes sense to work out how to have national asset behaviour and cobrand asset behaviour, although I need to work out a clean way to do that.

struan:
In the first animation I think a lot of the issue (and I assume the issue with that is the which council text changing) that's an existing issue to do with the update_pin then calling the category change code, both of which update that text. I don't intend to fix that as part of this.

dracos:
"I assume the issue with that is the which council text changing" isn't it that the LHS is totally blank when you click on the map?

dracos:
Superseded.

PR #2056, opened 2018-04-03
Standardise styling for #sub_map_links and #mob_sub_map_links

### LINKED ISSUE #2018: Make hide pins bar/stuff much bigger on mobile, easier to see and to hit
Spun out from https://github.com/mysociety/fixmystreet/issues/1462#issuecomment-364899132 - previous work in #2003 but @dracos thinks we could go even bigger

zarino:
**To do**

* Check tests

Fixes #2018 by making the `#sub_map_links` bigger.

But the greater win here is that we’re using fewer ID selectors in our CSS, and both `#sub_map_links` and `#mob_sub_map_links` now share the same styling – both of which are good for maintainability.

<table>
<thead>
<tr><th>Before</th><th>After</th></tr>
</thead>
<tr><td></td><td>
</td></tr>
<tr><td></td><td>
</td></tr>
<tr><td></td><td>
</td></tr>
</table>

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -1468,121 +1468,138 @@ html.js #map .noscript {
   top: 44px;
 }
 
-//hide pins, show old reports etc
-#sub_map_links {
-  position: absolute;
-  #{$left}: 0;
-  #{$right}: 0;
-  bottom: 0;
-  background: #333;
-  background: rgba(0, 0, 0, 0.7);
-  margin: 0;
-  font-size: 1em;
-  line-height: 1em;
-
-  a {
+%sub-map-link-icon {
+    content: "";
     display: inline-block;
-    font-size: 0.8em;
-    color: #fff;
-    padding: 0.7em 1em;
-
-    &:hover {
-      background-color: #000;
-      text-decoration: none;
-    }
-
-    &:after {
-      content: "";
-      display: inline-block;
-      width: 12px;
-      height: 12px;
-      margin-#{left}: 8px;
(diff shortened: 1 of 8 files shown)


zarino:
@dracos – Can you think of any situations where there are more than two links in `#sub_map_links` on mobile?

@wrightmartin – Do you think I should increase the dimensions of the little white icons in the links, to match the new, larger text? I can’t tell whether I like them small or not.

dracos:
No, but only through luck. If a site had a Satellite link (e.g. Zurich) and also had alerts or shortlist-makers (not Zurich), then it's possible there could be three - e.g. on around page satellite/hide-pins/get updates, or satellite/shortlist-visible/expand-close. But nothing currently is like that, I think. If we moved All Reports 'feed' link up to that bar as we discussed recently, then that could have three at zoomed in with updates/expand-close/shortlist-visible. Not easy to work this out!

wrightmartin:
> Do you think I should increase the dimensions of the little white icons in the links, to match the new, larger text? I can’t tell whether I like them small or not.

Unsure, they could go a bit bigger but without looking they might appear a bit clumsy.

web/cobrands/sass/_base.scss
```diff
@@ -1468,121 +1468,138 @@ html.js #map .noscript {
...
-      bottom: 0;
-      width: 4em;
-      background: #fff;
-      color: #000;
 
-      &:before {
-        content: "";
+    // "Fade out" the map.
+    height: 100%;
+    background-color: rgba(#000, 0.3);
```
dracos: Nice :)

PR #2057, opened 2018-04-03
User import admin page

davea:
This is intended to be an internal tool for quickly creating staff
users in bulk. As such, it's hidden at `/admin/user_import` and is only
available to superusers.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -12,6 +12,7 @@ use DateTime::Format::Strptime;
 use List::Util 'first';
 use List::MoreUtils 'uniq';
 use mySociety::ArrayUtils;
+use Text::CSV;
 
 use FixMyStreet::SendReport;
 use FixMyStreet::SMS;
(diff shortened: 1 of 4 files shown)


templates/web/base/admin/user_import.html
```diff
@@ -0,0 +1,76 @@
...
+    <table>
+        <tr>
+            <th>[% loc('Name') %]</th>
+            <th>[% loc('Email') %]</th>
+            <th>[% loc('Body') %]</th>
+        </tr>
+        [% FOREACH user IN new_users %]
+            <tr>
+                <td>
+                    <a href="[% c.uri_for( c.controller('Admin').action_for('user_add'), user.id ) %]">
```
dracos: Shouldn't this be `user_edit`, not add? :) In fact, `[% c.uri_for_action( 'admin/user_edit', user.id %]` is all you need, I think.

templates/web/base/admin/user_import.html
```diff
@@ -0,0 +1,76 @@
...
+    <table>
+        <tr>
+            <th>[% loc('Name') %]</th>
+            <th>[% loc('Email') %]</th>
+            <th>[% loc('Body') %]</th>
+        </tr>
+        [% FOREACH user IN existing_users %]
+            <tr>
+                <td>
+                    <a href="[% c.uri_for( c.controller('Admin').action_for('user_add'), user.id ) %]">
```
dracos: Ditto

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1636,6 +1637,56 @@ sub user_edit : Path('user_edit') : Args(1) {
...
 }
 
+sub user_import : Path('user_import') {
+    my ( $self, $c, $id ) = @_;
+
+    $c->forward('/auth/get_csrf_token');
+    return unless $c->user_exists && $c->user->is_superuser;
+
+    if ($c->req->method eq 'POST') {
+        $c->forward('check_for_super_user');
```
dracos: Don't need this given line above.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1636,6 +1637,56 @@ sub user_edit : Path('user_edit') : Args(1) {
...
+            my @permissions = split(/:/, $permissions);
+
+            my $user = FixMyStreet::DB->resultset("User")->find_or_new({ email => $email, email_verified => 1 });
+            if ($user->in_storage) {
+                push @{$c->stash->{existing_users}}, $user;
+                next;
+            }
+
+            $user->name($name);
+            $user->from_body($from_body);
```
dracos: I imagine this will be the empty string if not given - should it be undef?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1636,6 +1637,56 @@ sub user_edit : Path('user_edit') : Args(1) {
...
+            $user->update_or_insert;
+
+            my @user_permissions = grep { $available_permissions{$_} } @permissions;
+            $user->admin_user_body_permissions->search({
+                body_id => $user->from_body->id,
+                permission_type => { '!=' => \@user_permissions },
+            })->delete;
```
dracos: Assume this is for later change to deal with updating users, but as we're only creating new users, this delete will never delete anything, perhaps a comment?

PR #2062, opened 2018-04-04
[BANES] remove fill from LLPG layer labels

struan:
Fixes mysociety/fixmystreetforcouncils#305


web/cobrands/bathnes/staff.js
```diff
@@ -6,8 +6,7 @@ if (!fixmystreet.maps || !fixmystreet.maps.banes_defaults) {
 
 var llpg_stylemap = new OpenLayers.StyleMap({
     'default': new OpenLayers.Style({
-        fillColor: "#FFFF00",
-        fillOpacity: 0.6,
+        fillOpacity: 0,
```
davea: This could be `fill: false`, but meh, makes no difference to the visible end result 😃

PR #2064, opened 2018-04-04
Check recent reports for any hidden since cached.

dracos:
Also stop caching lat/lon specific alert photo lists, and improve
Memcached code.

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/Problem.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Problem.pm
@@ -132,32 +132,23 @@ sub _recent {
     };
 
     my $probs;
-    my $new = 0;
-    if (defined $lat) {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -132,34 +132,25 @@ sub _recent {
...
 
-    if ( $new ) {
-        Memcached::set($key, $probs, 3600);
-    } else {
-        # Need to reattach schema so that confirmed column gets reinflated.
-        $probs->[0]->result_source->schema( $rs->result_source->schema ) if $probs->[0];
-    }
-
+    # Catch any cached ones since hidden
+    $probs = [ grep { ! $_->is_hidden } @$probs ];
```
struan: surely we only need to do this with the cached problems so could move this into the `if ($prob)` bit above?

PR #2067, opened 2018-04-06
Open311 fetch all reports

struan:
This adds an option to not set start and end dates when fetching problems over Open311. This is either if you always want to fetch all of them or if the API doesn't have a date options and you want to suppress the warnings about problems outside the default date range.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -218,9 +218,14 @@ sub bodies : Path('bodies') : Args(0) {
         $c->forward('check_for_super_user');
         $c->forward('/auth/check_csrf_token');
 
-        my $params = $c->forward('body_params');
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -464,7 +474,10 @@ sub body_params : Private {
     );
     my %params = map { $_ => $c->get_param($_) || $defaults{$_} } keys %defaults;
     $c->forward('check_body_params', [ \%params ]);
-    return \%params;
+    my @extras = qw/fetch_all_problems/;
+    %defaults = map { $_ => '' } @extras;
+    my %extras = map { $_ => $c->get_param($_) || $defaults{$_} } @extras;
+    return { params => \%params, extras => \%extras };
```
davea: This is nice.

PR #2082, opened 2018-04-13
[Open311] Pretty print all request parameters.

### LINKED ISSUE #2078: More helpful error messages for open311 sending errors
At the moment the errors from problems sent via open311 include the raw dump of the POST and details of the error condition. As such it's a bit hard to pick out things like where the report was being sent, which report it was as these are embedded in the POST dump.

We should update this to include some basic information about the body, report id and probably the category in an obvious way to make initial triage a bit easier.

dracos:
Fixes #2078

### DIFF
--- a/perllib/Open311.pm
+++ b/perllib/Open311.pm
@@ -257,7 +257,7 @@ sub get_service_request_updates {
     my $end_date = shift;
 
     my $params = {
-        api_key => $self->api_key,
+        api_key => $self->api_key || '',
     };
 
     if ( $start_date || $end_date ) {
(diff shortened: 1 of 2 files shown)


perllib/Open311.pm
```diff
@@ -420,9 +420,12 @@ sub _get {
     $params->{ jurisdiction_id } = $self->jurisdiction
         if $self->jurisdiction;
     $uri->path( $uri->path . $path );
+    my $base_uri = $uri->clone;
     $uri->query_form( $params );
 
-    $self->debug_details( $self->debug_details . "\nrequest:" . $uri->as_string );
+    my $debug_request = "GET " . $base_uri->as_string . "\n";
+    $debug_request .= join("\n", map { "$_: $params->{$_}" } keys %$params);
```
struan: Maybe an extra \n in here just to split up the params from the request details for readability?

PR #2083, opened 2018-04-13
Abort if a script is run in an old deploy.

### LINKED ISSUE #2079: Scripts should check they are running in latest deployment
It's possibly when FixMyStreet has been deployed using timestampted deployments to find yourself running a script in the previous deployment. E.g. if you are on a server, have changed to the FixMyStreet directory and then re-deployed you will be in the previous deployment.

If a script is running inside a symlinked directory it should do some checks to make sure it's running under the latest deployment.

dracos:
Fixes #2079 
[skip changelog]

### DIFF
--- a/setenv.pl
+++ b/setenv.pl
@@ -6,9 +6,20 @@ use warnings;
 my $root;
 
 BEGIN {
-    use File::Basename qw(dirname);
-    use File::Spec;
-    $root = dirname(File::Spec->rel2abs(__FILE__));
+    use Cwd qw(abs_path);
+    use File::Basename qw(dirname basename);
+    $root = dirname(abs_path(__FILE__));
+}
+
+# Check there is not a later timestamped deploy, if running in such a system
+
+if ($root =~ /fixmystreet-\d\d\d\d-\d\d-\d\dT\d\d-\d\d-\d\d$/) {
+    my @deploys = sort map { basename $_ } glob("$root/../fixmystreet-*");
+    if (basename($root) ne $deploys[-1] && !$ENV{OLD_DEPLOY_ACKNOWLEDGED}) {
+        require Term::ANSIColor;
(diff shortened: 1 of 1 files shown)


setenv.pl
```diff
@@ -6,9 +6,20 @@
...
-    $root = dirname(File::Spec->rel2abs(__FILE__));
+    use Cwd qw(abs_path);
+    use File::Basename qw(dirname basename);
+    $root = dirname(abs_path(__FILE__));
+}
+
+# Check there is not a later timestamped deploy, if running in such a system
+
+if ($root =~ /fixmystreet-\d\d\d\d-\d\d-\d\dT\d\d-\d\d-\d\d$/) {
+    my @deploys = sort map { basename $_ } glob("$root/../fixmystreet-*");
```
struan: I am not sure what is going on but this use of glob upsets at least the test running scripts:

```
photek:fixmystreet struan$ (2079-check-latest-deploy $%=) ./script/test
Spinning up a Pg cluster/database...
Use of uninitialized value in glob at /Users/struan/src/mysociety/FMS-vagrant/fixmystreet/local/lib/perl5/Test/PostgreSQL.pm line 72.
Use of uninitialized value in glob at /Users/struan/src/mysociety/FMS-vagrant/fixmystreet/local/lib/perl5/Test/PostgreSQL.pm line 72.
Use of uninitialized value in glob at /Users/struan/src/mysociety/FMS-vagrant/fixmystreet/local/lib/perl5/Test/PostgreSQL.pm line 72.
```
which seems to be because caused by this changing the `glob` behaviour when called on a non-existant path.

I don't totally understand what is going on though.

struan: If you remove the glob here then it all works fine.

dracos: Yeah, using `glob` autoloads File::Glob but then the unload bit was removing it (it's for removing File::Spec etc) which would confuse poor `glob`! Fixed to skip glob.

PR #2088, opened 2018-04-17
[Open311] avoid spurious warning when params are undef

struan:
If there was an undefined parameter passed to `_post` then generating
the debug string would generate a warning about an undefined value in a
concatenation. So, check for undefined params and substitute empty
string to avoid.

[skip changelog]


dracos:
I see how this fixes the warning, but isn't the actual issue that an undef should never get to this point? So the warning when it happen is a good thing... :)

perllib/Open311.pm
```diff
@@ -418,6 +418,21 @@ sub split_name {
...
 }
 
+sub _params_to_string {
+    my( $self, $params, $request_string ) = @_;
+
+    my $undefined;
+
+    my $string = join("\n", map {
+        $undefined .= "$_ undefined\n" unless defined $params->{$_};
+        "$_: " . ( defined $params->{$_} ? $params->{$_} : '' );
```
dracos: This is fine, but just to note `defined $params->{$_} ? $params->{$_} : ''` could be `$params->{$_} // ''` (`//` is like `||` but tests definedness rather than truth, so e.g. would let through `0`.

PR #2092, opened 2018-04-17
cobrand hook to deny access to pages

struan:
Add a check in the root controller `auto` to a cobrand hook that denies
access if it returns true. This goes here so that cobrands and users are
set up, which is not the case for `check_login_required`. Used to do
things like deny site access unless the user is a superuser.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Root.pm
+++ b/perllib/FixMyStreet/App/Controller/Root.pm
@@ -39,6 +39,7 @@ sub auto : Private {
 
     # decide which cobrand this request should use
     $c->setup_request();
+    $c->detach('/auth/redirect') if $c->cobrand->call_hook('check_login_disallowed');
 
     return 1;
 }
--- a/t/app/controller/root.t
+++ b/t/app/controller/root.t
@@ -1,4 +1,5 @@
 use FixMyStreet::TestMech;
+use Test::MockModule;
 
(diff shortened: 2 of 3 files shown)


perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -39,6 +39,7 @@ sub auto : Private {
 
     # decide which cobrand this request should use
     $c->setup_request();
+    $c->detach('/auth/redirect', []) if $c->cobrand->call_hook(check_login_disallowed => $c);
```
dracos: You don't need to pass in `$c` here, it's available as `$self->{c}` in a cobrand during a web request.

t/app/controller/root.t
```diff
@@ -73,4 +74,17 @@ FixMyStreet::override_config {
...
+subtest "check_login_disallowed cobrand hook" => sub {
+    my $cobrand = Test::MockModule->new('FixMyStreet::Cobrand::Default');
+    $cobrand->mock('check_login_disallowed', sub {
+            return 1;
+        }
+    );
+
+    $mech->get('/');
+    is $mech->status, 302, 'disallowed page issues a redirect';
+    is $mech->res->headers->header('location'), 'http://localhost/auth?r=auth', 'redirects to auth page';
```
dracos: Redirecting to `/auth?r=auth` doesn't feel right; if you fetched the front page, shouldn't it have been `/auth?r=`?

PR #2096, opened 2018-04-19
add a send login email button to user edit page

### LINKED ISSUE #2041: Add admin button to send user a login email

struan:
Add a button to the user edit page that sends a login token email to the
user. Helpful for user support situations where someone is having
trouble logging in. Also for situations where you have added a user and
want to get them logged in.

Fixes #2041

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1417,17 +1417,18 @@ sub user_add : Path('user_edit') : Args(0) {
     my $email_v = $c->get_param('email_verified');
     my $phone_v = $c->get_param('phone_verified');
 
-    unless ($email || $phone) {
-        $c->stash->{field_errors}->{username} = _('Please enter a valid email or phone number');
-    }
-    if (!$email_v && !$phone_v) {
-        $c->stash->{field_errors}->{username} = _('Please verify at least one of email/phone');
-    }
     if ($email && !is_valid_email($email)) {
         $c->stash->{field_errors}->{email} = _('Please enter a valid email');
     }
     unless ($c->get_param('name')) {
         $c->stash->{field_errors}->{name} = _('Please enter a name');
+    }
+
+    unless ($email || $phone) {
+        $c->stash->{field_errors}->{username} = _('Please enter a valid email or phone number');
+    }
+    if (!$email_v && !$phone_v) {
+        $c->stash->{field_errors}->{username} = _('Please verify at least one of email/phone');
     }
 
     if ($phone_v) {
@@ -1503,6 +1504,11 @@ sub user_edit : Path('user_edit') : Args(1) {
         $c->forward('user_hide_everywhere', [ $user ]);
     } elsif ( $c->get_param('submit') and $c->get_param('remove_account') ) {
         $c->forward('user_remove_account', [ $user ]);
+    } elsif ( $c->get_param('submit') and $c->get_param('send_login_email') ) {
+        my $email = lc $c->get_param('email');
+        my %args = ( email => $email );
+        $args{user_id} = $id if $user->email ne $email || !$user->email_verified;
+        $c->forward('send_login_email', [ \%args ]);
     } elsif ( $c->get_param('submit') ) {
 
         my $edited = 0;
@@ -1915,6 +1921,31 @@ sub user_hide_everywhere : Private {
         $update->hide;
     }
     $c->stash->{status_message} = _('That user’s reports and updates have been hidden.');
+}
+
+sub send_login_email : Private {
+    my ( $self, $c, $args ) = @_;
+
+    my $token_data = {
+        email => $args->{email},
+    };
+
+    $token_data->{old_user_id} = $args->{user_id} if $args->{user_id};
+    $token_data->{name} = $args->{name} if $args->{name};
+
+    my $token_obj = $c->model('DB::Token')->create({
+        scope => 'email_sign_in',
+        data  => $token_data,
+    });
+
+    $c->stash->{token} = $token_obj->token;
+    my $template = 'login.txt';
+
+    # do not use relative URIs in the email, obvs.
+    $c->uri_disposition('absolute');
+    $c->send_email( $template, { to => $args->{email} } );
+
+    $c->stash->{status_message} = _('The user has been sent a login email');
 }
 
 # Anonymize and remove name from all problems/updates, disable all alerts.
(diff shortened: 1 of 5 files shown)


struan:
Not very sure about the button placing FWIW.

dracos:
Oh yes, and for button placement, I'd move it to be under/next to the email address box :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1384,19 +1384,28 @@ sub user_add : Path('user_edit') : Args(0) {
...
-        $c->stash->{field_errors}->{username} = _('Please verify at least one of email/phone');
-    }
     if ($email && !is_valid_email($email)) {
         $c->stash->{field_errors}->{email} = _('Please enter a valid email');
     }
     unless ($c->get_param('name')) {
         $c->stash->{field_errors}->{name} = _('Please enter a name');
     }
 
+    if ( $c->get_param('send_login_email') ) {
```
dracos: So this is for if they fill in a bit of the Add User form in the admin, but then use it to send a login email, same as if they filled in `/auth` on behalf of someone else, I guess? And the rest of the form (apart from name) is ignored. Is it worth having this, or is it clearer if it's only on the user edit page, do you think?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1470,6 +1479,11 @@ sub user_edit : Path('user_edit') : Args(1) {
         $c->forward('user_hide_everywhere', [ $user ]);
     } elsif ( $c->get_param('submit') and $c->get_param('remove_account') ) {
         $c->forward('user_remove_account', [ $user ]);
+    } elsif ( $c->get_param('submit') and $c->get_param('send_login_email') ) {
+        my $email = lc $c->get_param('email');
+        my @args = ( $email );
+        push @args, ( '', $id ) if $user->email ne $email || !$user->email_verified;
```
dracos: Not really a big issue, but the empty string here makes me thing you should be passing a hash of email/id here (and email/name in the other place), rather than a list.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1884,6 +1898,35 @@ sub user_hide_everywhere : Private {
...
+    my $token_data;
+    $token_data = {
+        email => $email,
+    };
+
+    if ( $user_id ) {
+        $token_data->{old_user_id} = $user_id;
+    }
+
+    $token_data->{name} = $name if $name;
```
dracos: If I'm being really pernickety (3yo was annoying earlier), odd to have this and above be two different ways of doing the same thing :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1884,6 +1898,35 @@ sub user_hide_everywhere : Private {
...
+
+    my $token_obj = $c->model('DB::Token')->create({
```
dracos: Could this and much of Auth.pm's `email_sign_in` be factored together, looks to basically be the same code, or not worth it?

perllib/FixMyStreet/App/Controller/Admin.pm
struan: I just thought it was a bit more obvious to use this form if they want to add a user rather than add user and then send them a log in email. Although I can see arguments the other way too. I don't really have strong feelings about it.

perllib/FixMyStreet/App/Controller/Admin.pm
struan: I'm not really sure it's worth it as the factoring out is going to add about as much extra code in to each method for the setup of the shared code as it saves I expect.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1884,6 +1898,32 @@ sub user_hide_everywhere : Private {
     $c->stash->{status_message} = _('That user’s reports and updates have been hidden.');
 }
 
+sub send_login_email : Private {
+    my ( $self, $c, $args ) email, $name, $user_id ) = @_;
```
dracos: Typo here.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1884,6 +1898,32 @@ sub user_hide_everywhere : Private {
...
+
+    my $token_data;
+    $token_data = {
```
dracos: Can join this line with line above.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1384,19 +1384,28 @@ sub user_add : Path('user_edit') : Args(0) {
...
     unless ($c->get_param('name')) {
         $c->stash->{field_errors}->{name} = _('Please enter a name');
     }
 
+    if ( $c->get_param('send_login_email') ) {
+        if (!$email) {
+            $c->stash->{field_errors}->{email} = _('Please enter a valid email');
+        }
+        return if %{$c->stash->{field_errors}};
+        return $c->forward('send_login_email', { email => $email, name => $c->get_param('name') });
```
dracos: Not sure what it does, but on the safe side, this should be `'send_login_email', [ { email => $email, name => $c->get_param('name') ] })` - second arg is an arrayref of arguments.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1470,6 +1479,11 @@ sub user_edit : Path('user_edit') : Args(1) {
         $c->forward('user_hide_everywhere', [ $user ]);
     } elsif ( $c->get_param('submit') and $c->get_param('remove_account') ) {
         $c->forward('user_remove_account', [ $user ]);
+    } elsif ( $c->get_param('submit') and $c->get_param('send_login_email') ) {
+        my $email = lc $c->get_param('email');
+        my %args = ( email => $email );
+        $args{user_id} = $id if $user->email ne $email || !$user->email_verified;
+        $c->forward('send_login_email', \%args);
```
dracos: Similarly

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1884,6 +1898,31 @@ sub user_hide_everywhere : Private {
...
+        scope => 'email_sign_in',
+        data  => $token_data,
+    });
+
+    $c->stash->{token} = $token_obj->token;
+    my $template = 'login.txt';
+
+    # do not use relative URIs in the email, obvs.
+    $c->uri_disposition('absolute');
+    $c->send_email( $template, { to => $email } );
```
dracos: Not defined here.

struan:
FWIW, I'll squash all this down into a single commit once the reviewing is done.

PR #2102, opened 2018-04-23
Don’t trigger response templates unless problem state changes

### LINKED ISSUE #2075: Should response templates only be triggered if state changes?
Reports having their updates fetched from Confirm are ending up with duplicate responses, e.g.: https://fix.bathnes.gov.uk/report/1278104

In this case there are multiple `EnquiryStatusLog` entries within Confirm that haven't changed `EnquiryStatusCode`, and `open311-adapter` is simply converting these into service request updates.

The response template handling code in `GetServiceRequests.pm` should probably only trigger the template if the problem state has *changed*, and if it hasn't and the update text is blank then the update should be marked as hidden.

davea:
Response templates won't be triggered unless the problem state or
external status code is changed.

Fixes #2075 
[skip changelog]

### DIFF
--- a/perllib/Open311/GetServiceRequestUpdates.pm
+++ b/perllib/Open311/GetServiceRequestUpdates.pm
@@ -110,13 +110,18 @@ sub update_comments {
 
             if ( !$c->first ) {
                 my $state = $open311->map_state( $request->{status} );
-                my $external_status_code = $request->{external_status_code};
+                my $old_state = $p->state;
+                my $external_status_code = $request->{external_status_code} || '';
+                my $old_external_status_code = $p->get_extra_metadata('external_status_code') || '';
                 my $comment = $self->schema->resultset('Comment')->new(
                     {
                         problem => $p,
                         user => $self->system_user,
                         external_id => $request->{update_id},
-                        text => $self->comment_text_for_request($request, $p, $state, $external_status_code),
+                        text => $self->comment_text_for_request(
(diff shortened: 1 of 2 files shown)


perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -184,22 +189,28 @@ sub update_comments {
...
-        $state_params->{'me.external_status_code'} = $external_status_code;
-    };
-
-    if (my $template = $problem->response_templates->search({
-        auto_response => 1,
-        -or => $state_params,
-    })->first) {
-        return $template->text;
+    # Response templates are only triggered if the state/external status has changed
+    my $state_changed = defined $state && defined $old_state && $state ne $old_state;
```
dracos: Is there any way `$state` or `$old_state` could not be defined here? `map_state` presumably has already warned if `request->{status}` wasn't set so we can assume it was.

perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -184,22 +189,28 @@ sub update_comments {
...
+    my $ext_code_changed = defined $ext_code && defined $old_ext_code && $ext_code ne $old_ext_code;
```
dracos: OTOH I can see how these could not be defined; would it be a bit nicer to have  `|| ''` when fetching the old/new external status codes, so we can assume they are strings from then on.

PR #2105, opened 2018-04-24
[Open311] warn about send-comments errors if verbose flag set

### LINKED ISSUE #2091: Issue more errors when sending Open311 updates fail
If sending an update over Open311 fails an error message is emitted on the first failure and then never again. `send_comments` continues to try and send the update but just silently fails.

As with sending reports over Open311 regular reminders of the error should probably happen until it's fixed.

struan:
Previously send-comments errors would only be printed the first time
they occurred so there was no visibility of ongoing errors. This brings
send-comments in line with send-reports by emitting errors if the script
is called with --verbose.

Fixes #2091

### DIFF
--- a/bin/send-comments
+++ b/bin/send-comments
@@ -153,8 +153,12 @@ while ( my $body = $bodies->next ) {
             $comment->update( {
                 send_fail_count => $comment->send_fail_count + 1,
                 send_fail_timestamp => \'current_timestamp',
-                send_fail_reason => 'Failed to post over Open311',
+                send_fail_reason => "Failed to post over Open311\n\n" . $o->error,
             } );
+
+            if ( $verbose && $o->error ) {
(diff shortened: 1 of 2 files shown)


bin/send-comments
```diff
@@ -155,6 +155,10 @@ while ( my $body = $bodies->next ) {
                 send_fail_timestamp => \'current_timestamp',
                 send_fail_reason => 'Failed to post over Open311',
```
dracos: Might be good to copy SendReport/Open311.pm here and put the Open311 error in `send_fail_reason` (it puts this string plus the open311->error). Then we can always see the error in the DB too as well as in email.

dracos:
Oh, will need to update crontab to sometimes use `--verbose` as currently it does not.

PR #2108, opened 2018-04-25
Keep It In The Community core changes

dracos:
One hook also in the Collidescope PR, and some more hook/sub-templates.


perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -117,9 +117,9 @@ sub short_name {
...
 
 sub find_closest {
-    my ( $self, $problem, $as_data ) = @_;
+    my ($self, $problem) = @_;
 
-    my $data = $self->SUPER::find_closest($problem, $as_data);
+    my $closest = $self->SUPER::find_closest($problem);
 
     my $mapit_url = FixMyStreet->config('MAPIT_URL');
     my ($lat, $lon) = map { Utils::truncate_coordinate($_) } $problem->latitude, $problem->longitude;
```
davea: This line is now crashing with `Can't call method "latitude" on unblessed reference` because `$problem` is a hashref when called from `Around->location_closest_address`

t/cobrand/closest.t
```diff
@@ -72,8 +72,9 @@ FixMyStreet::override_config {
     $near = $c->find_closest_address_for_rss($report);
 
     ok !$near, 'no closest address for RSS if not cached';
+
+    my $json = $mech->get_ok_json('/ajax/closest?lat=55&lon=-1');
```
davea: This will use the `Default` cobrand, not `UK` like the rest of the tests, as `FixMyStreet::override_config` doesn't set `ALLOWED_COBRANDS`.

dracos: Added another UK-test.

PR #2112, opened 2018-04-27
Allow cobrands to supply arbitrary extra query params for map pins

dracos:
So that you can build functionality to add extra limitations on to the map pins displayed. Useful for Collideoscope initially to filter out (or in) reports from the Department of Transport's Stats19 Data.
[skip changelog]


dracos:
"Is there a reason this isn't used in the nearby call in the Report controller?" - not really, just this was for 'normal' use and that is the duplicate staff code. But no reason it couldn't handle just in case.
Will add a test, yep.

PR #2113, opened 2018-04-27
[Open311] Hide fetched update if no data to show.

dracos:
If no text, photo, or state change, hide the update from display.

Question: hide, or not insert them? Not inserting means it'll process them again next time, but presumably with same outcome? So might be better.
Todo: test.
[skip changelog]


davea:
I suppose ignoring rather than hiding would mean you could add a response template after the update is made in Confirm/whatever and it would eventually get triggered & sent. But don't think it really matters either way.

PR #2116, opened 2018-05-02
Extra help text for categories

dracos:
This displays a category's extra `help_text` if present in list filters and when reporting.
It also takes the opportunity to just pass around Contact objects rather than map them to smaller hashes.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -233,7 +233,7 @@ sub check_and_stash_category : Private {
             body_id => [ keys %bodies ],
         },
         {
-            columns => [ 'category' ],
+            columns => [ 'category', 'extra' ],
             order_by => [ 'category' ],
             distinct => 1
         }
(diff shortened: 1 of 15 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -679,15 +679,15 @@ sub setup_categories_and_bodies : Private {
...
         @category_options = (
-            { name => _('-- Pick a category --'), value => _('-- Pick a category --'), group => '' },
-            grep { $_->{name} ne _('Other') } @category_options );
-        push @category_options, { name => _('Other'), value => $seen{_('Other')}, group => _('Other') } if $seen{_('Other')};
+            { category => _('-- Pick a category --'), category_display => _('-- Pick a category --'), group => '' },
+            grep { $_->category ne _('Other') } @category_options );
+        push @category_options, $seen{_('Other')} if $seen{_('Other')};
     }
 
     $c->cobrand->call_hook(munge_category_list => \@category_options, \@contacts, \%category_extras);
```
struan: Just a note that anywhere that uses this hook will need updated although I can't find anywhere that does.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -238,9 +238,8 @@ sub check_and_stash_category : Private {
             distinct => 1
         }
     )->all;
-    my @categories = map { { name => $_->category, value => $_->category_display } } @contacts;
     $c->stash->{filter_categories} = \@categories;
-    my %categories_mapped = map { $_->{name} => 1 } @categories;
+    my %categories_mapped = map { $_->category => 1 } @categories;
```
struan: This is much simpler. Yay!

PR #2117, opened 2018-05-03
[Bromley] fix inability to click hide update checkbox

struan:
The internal link for updates to enable #update_n anchor links to work
has some padding/margin CSS to make sure the place scrolled to doesn't
obscure any of the update. This is partly based on the page header and
does not interact well with the header for the Bromley cobrand
preventing the user from clicking on the update moderation checkboxes.
By overriding the margin/padding for Bromley the anchor link still works
and the checkboxes also work

Fixes mysociety/fixmystreet-commercial#1027

[skip changelog]

### DIFF
--- a/templates/web/base/report/update.html
+++ b/templates/web/base/report/update.html
@@ -4,6 +4,7 @@
     <ul class="item-list item-list--updates">
 [% END %]
         <li class="item-list__item item-list__item--updates">
+            <a name="update_[% update.id %]" class="internal-link-fixed-header"></a>
           [% IF permissions.moderate; original_update = update.moderation_original_data %]
             <form method="post" action="/moderate/report/[% problem.id %]/update/[% update.id %]">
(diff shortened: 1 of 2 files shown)


dracos:
This reduces it, but e.g. "Show name publicly?" is still only half-clickable even with this change. The padding/margin 'issue' doesn't always come into effect, haven't worked out why. Modern browsers seem to do better at not hiding internal links under fixed headers, too. Think this needs a designer to look at to pick a different fix, I think @zarino might have written the current one?

zarino:
@dracos I’ve reverted @struan’s commit and gone about it a different way. I'll check it in a few different browsers and cobrands after lunch.

I guess we’ll want to squash the commits and add a changelog entry (since my new changes affect the core theme, not just Bromley) before we merge.

zarino:
@dracos Ace, I'll squash and merge now then.

PR #2120, opened 2018-05-11
Add UI for selecting multiple wards from /reports/ body page

davea:
multi_wards

Fixes mysociety/fixmystreetforcouncils#231

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -126,6 +126,17 @@ sub ward : Path : Args(2) {
     my @wards = split /\|/, $ward || "";
     $c->forward( 'body_check', [ $body ] );
 
+    # If viewing multiple wards, rewrite the url from
+    # /reports/Borsetshire?ward=North&ward=East
+    # to
+    # /reports/Borsetshire/North|East
+    my @ward_params = $c->get_param_list('ward');
+    if ( @ward_params ) {
+        $c->stash->{wards} = [ map { { name => $_ } } (@wards, @ward_params) ];
+        delete $c->req->params->{ward};
+        $c->detach("redirect_body");
+    }
+
     my $body_short = $c->cobrand->short_name( $c->stash->{body} );
     $c->stash->{body_url} = '/reports/' . $body_short;
 
--- a/templates/web/base/reports/_ward-list.html
+++ b/templates/web/base/reports/_ward-list.html
@@ -1,16 +1,29 @@
 [% IF children.size %]
-<section id="council_wards" class="hidden-js">
+[% SET show_wards_hidden_class = c.req.params.show_wards ? 'hidden' : 'js-ward-single' %]
+[% SET show_wards_shown_class = c.req.params.show_wards ? '' : 'js-ward-multi hidden' %]
+<section id="council_wards"[% UNLESS c.req.params.show_wards %] class="hidden-js"[% END %]>
 <h2>[% ward_text %]</h2>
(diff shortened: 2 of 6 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1118,6 +1126,7 @@ fixmystreet.display = {
                 $sideReport.find('#key-tool-problems-nearby').addClass('js-back-to-report-list');
             }
             fixmystreet.set_up.map_sidebar_key_tools();
+            fixmystreet.set_up.ward_select_multiple();
```
dracos: This won't do anything, you're switching to a individual report page that lacks this link.
The main page is then hidden, so you don't need to reinstate it when you go back either, it looks like.

templates/web/base/reports/_ward-list.html
```diff
@@ -1,16 +1,27 @@
...
+          </label>
+          <a class="js-ward-single" href="[% child.url %]">[% child.name %]</a>
         </li>
       [% END %]
     </ul>
     </section>
+    <p>
+      <input class="btn js-ward-multi hidden" type="submit" value="[% loc('View wards') %]" />
+    </p>
+    <p class="js-ward-single"><a href="#" class="js-ward-select-multiple">[% loc('Select multiple wards?') %]</a></p>
```
dracos: Can we have this at the top? Bit hard to find at the bottom.

Also the button shouldn't be viewable unless JS is enabled – or actually *ideally*, the button would submit to the server and reshow the page with the checkboxes shown, please :)

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -126,6 +126,17 @@ sub ward : Path : Args(2) {
     my @wards = split /\|/, $ward || "";
     $c->forward( 'body_check', [ $body ] );
 
+    # If viewing multiple wards, rewrite the url from
+    # /reports/Borsetshire?ward=North&ward=East
+    # to
+    # /reports/Borsetshire/North|East
+    my @ward_params = $c->get_param_list('ward');
+    if ( @ward_params ) {
+        $c->stash->{ward_names} = [ @wards, @ward_params ];
```
dracos: Any reason you didn't use `wards` here - it looks like you do want to pass them through short_name for escaping purposes, given it's the names coming through?

templates/web/base/reports/_ward-list.html
```diff
@@ -1,16 +1,27 @@
...
+    <p class="js-ward-single">[% loc('Follow a ward link to view only reports within that ward.') %]</p>
+    <p class="js-ward-multi hidden">[% loc('Select multiple wards to view only reports within those wards.') %]</p>
+    <form method="get" action="">
     <section class="full-width">
     <ul class="item-list item-list--wards">
       [% FOR child IN children.values.sort('name') %]
         <li class="item-list__item item-list--wards__item">
-            <a href="[% child.url %]">[% child.name %]</a>
+          <label class="js-ward-multi hidden">
+            <input type="checkbox" name="ward" value="[% child.name | html %]">
```
dracos: Alternatively to my previous comment, if you stored the short name (last bit of the URL) here you could just pass them through.

templates/web/base/reports/_ward-list.html
dracos: How about f26feef26d8a1961b124869d47aa4c749f25c5af? :)

davea: That'll do nicely!

PR #2125, opened 2018-05-15
User/report inactivity processing

dracos:
This PR adds support for anonymizing/emailing inactive users (activity based on session expiry or log in/out) at custom timescales, and anonymizing/closing-to-updates inactive reports (and their associated updates/alerts). And admin interface for closing-to-updates any particular report.
- [x] Tests on closing-to-updates admin feature/display of update form gone when on.

### DIFF
--- a/bin/expire-sessions
+++ b/bin/expire-sessions
@@ -17,8 +17,6 @@ BEGIN {
 use FixMyStreet::DB;
 use Getopt::Long;
 use List::Util qw(uniq);
-use MIME::Base64;
-use Storable;
 
 GetOptions(
     'init' => \my $init,
@@ -28,8 +26,8 @@ my $rs = FixMyStreet::DB->resultset("Session");
 
 # Delete expired sessions (including from in User object)
 while (my $session = $rs->search({ expires => { '<', time() } })->next) {
-    if (my $user = get_user($session)) {
-        my $id = get_id($session);
+    if (my $user = $session->user) {
+        my $id = $session->id_code;
         my $sessions = $user->get_extra_metadata('sessions');
         my @new_sessions = grep { $_ ne $id } @$sessions;
         update_user_sessions($user, \@new_sessions) if @new_sessions != @$sessions;
@@ -41,8 +39,8 @@ if ($init) {
     # Update sessions to make sure all present in User objects
     print "Setting up sessions in user objects\n";
     while (my $session = $rs->next) {
-        my $user = get_user($session) or next;
-        my $id = get_id($session);
+        my $user = $session->user or next;
+        my $id = $session->id_code;
         my $sessions = $user->get_extra_metadata('sessions');
         my @new_sessions = uniq @$sessions, $id;
         update_user_sessions($user, \@new_sessions) if @new_sessions != @$sessions;
@@ -50,23 +48,6 @@ if ($init) {
 }
 
 # ---
-
-sub get_user {
-    my $session = shift;
-    return unless $session->session_data;
-    my $data = Storable::thaw(MIME::Base64::decode($session->session_data));
-    return unless $data->{__user};
-    my $user = FixMyStreet::DB->resultset("User")->find($data->{__user}{id});
-    return $user;
-}
-
-sub get_id {
-    my $session = shift;
-    my $id = $session->id;
-    $id =~ s/^session://;
-    $id =~ s/\s+$//;
-    return $id;
-}
 
 sub update_user_sessions {
     my ($user, $sessions) = @_;
--- a/bin/expire-sessions
+++ b/bin/expire-sessions
@@ -1,11 +1,11 @@
 #!/usr/bin/env perl
 
 # expire-sessions: Run regularly to remove old sessions (plus
-# can set up data for 'log user out' admin functionality).
+# can set up data for 'log user out' admin functionality, and
+# inactive user processing).
 
-use strict;
+use v5.14;
 use warnings;
-require 5.8.0;
 
 BEGIN {
     use File::Basename qw(dirname);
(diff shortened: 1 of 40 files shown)


bin/expire-sessions
```diff
@@ -19,41 +19,37 @@ use Getopt::Long;
...
 GetOptions(
+    # Update sessions to make sure all present in User objects
     'init' => \my $init,
 );
 
 my $rs = FixMyStreet::DB->resultset("Session");
+my $now = time();
 
 # Delete expired sessions (including from in User object)
-while (my $session = $rs->search({ expires => { '<', time() } })->next) {
```
struan: Could we not leave this search in if init is missing to avoid looping over every session in the DB? I guess it's probably not a big issue given this is happening offline.

perllib/FixMyStreet/Script/Inactive.pm
```diff
@@ -0,0 +1,79 @@
+package FixMyStreet::Script::Inactive;
+
+use v5.14;
+use warnings;
```
struan: If this is just to turn on say then `use feature qw/say/` might be more explicit about it.

bin/process-inactive-accounts
```diff
@@ -0,0 +1,42 @@
...
+
+use Getopt::Long;
+use FixMyStreet::Script::Inactive;
+use Pod::Usage;
+
+my %h;
+GetOptions(\%h, 'anonymize=i', 'email=i', 'verbose|v', 'help|h');
+pod2usage(0) if $h{help};
+pod2usage(1) unless $h{anonymize} && $h{email};
+pod2usage("Anonymize time must be greater than email time")
```
struan: I wonder if being able to only anonymize accounts without email some would be useful?

perllib/FixMyStreet/Script/Inactive.pm
```diff
@@ -0,0 +1,79 @@
...
+        my $site = CronFns::site($base_url);
+        FixMyStreet::Cobrand->get_class_for_moniker($site)->new;
+    },
+);
+
+sub users {
+    my $self = shift;
+    $self->anonymize_users;
+    $self->email_inactive_users;
+}
```
struan: I wonder about this having some sort of summary mode in between "say nothing" and "print out everything I am doing" might be an idea. Mostly as a reassurance that something happened but without screeds of output. Probably not as I guess this will mostly be run by cron and no one wants emails about things that just work.

perllib/FixMyStreet/Script/Inactive.pm
```diff
@@ -23,12 +23,63 @@ has cobrand => (
...
+        say "Anonymizing problem #" . $problem->id if $self->verbose;
+
+        # Remove personal data from the report
+        $problem->update({
+            user => $self->anonymous_user,
+            name => '',
+            anonymous => 1,
+            send_questionnaire => 0,
+        })
+            if $problem->user != $self->anonymous_user;
```
struan: This formatting is weird and a bit confusing in that the if does not look like it's associated with the update.

t/script/inactive.t
```diff
@@ -15,6 +15,20 @@ $t = DateTime->now->subtract(months => 4);
...
+        dt => $t,
+        lastupdate => "$t",
+        state => $m % 2 ? 'fixed - user' : 'confirmed',
+    });
+}
+
+$in->reports;
+
+my $count = FixMyStreet::DB->resultset("Problem")->search({ user_id => $user->id })->count;
+is $count, 6, 'Six non-anonymised';
```
struan: This should probably test anonymising updates and new updates on anonymised reports.

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -451,6 +451,25 @@ sub adopt {
...
+    $self->password('', 1);
+    $self->update({
+        email => 'removed-' . $self->id . '@' . FixMyStreet->config('EMAIL_DOMAIN'),
+        email_verified => 0,
+        name => '',
+        phone => '',
+        phone_verified => 0,
+        title => undef,
+        twitter_id => undef,
+        facebook_id => undef,
```
struan: I wonder if this should unset superuser and body things just to be on the safe side. Or in some other way special case those?

bin/expire-sessions
dracos: It has to loop over all sessions to set the last active time, not just when init is set (the `if $user` bit always runs).

perllib/FixMyStreet/Script/Inactive.pm
dracos: Not just that! Also means I don't have to say `use strict`, has better (correct) Unicode handling on string operations. Could also use `state`. Probably should have it everywhere :)

bin/process-inactive-accounts
dracos: Yes, good idea.

perllib/FixMyStreet/DB/Result/User.pm
dracos: Would you ever want to anonymize a staff/superuser account, even if inactive, good question. Note there should be no way of logging in with these accounts as email and phone verified are both 0, no password, and email won't be delivered (and even if it were, it would create a new account with the same email, verified). So it might be useful to know if an account was superuser/body without knowing who it was?

jonkri:
Hi!

Regarding...

<blockquote>This PR adds support for anonymizing/emailing inactive users (activity based on session expiry or log in/out)</blockquote>

... I was wondering when sessions expire, or how long after the last "true" activity of a user will the user be considered to have been inactive.

For instance, if a user visits the site and is already logged in, we would like for that to "update" the last active time of that account. And if the user closes the browser and never returns to the site, we would ideally like the exact time of that last site visit to be the time of the last activity for the user. Is that how it works now? (If not, how does it work?)

Thanks a lot for all the great work!

All the best,
Jon

dracos:
"For instance, if a user visits the site and is already logged in, we would like for that to "update" the last active time of that account. And if the user closes the browser and never returns to the site, we would ideally like the exact time of that last site visit to be the time of the last activity for the user. Is that how it works now?" - yes, that is how it works, session expiry is updated on each page view.

templates/email/default/inactive-account.txt
```diff
@@ -0,0 +1,16 @@
+Subject: Your inactive account on [% site_name %]
+
+Hello [% user.name %],
+
+Your account on [% site_name %] has been inactive for [% email_from %]
+[% nget('month', 'months', email_from) %], and we automatically remove
+accounts that have been inactive after [% anonymize_from %]
+[% nget('month', 'months', anonymize_from) %]. If you wish to keep your
+account, please log in to the site and that will keep it active.
```
davea: It'd be handy to have a login link here. Also, no HTML version?

templates/web/base/report/update-form.html
```diff
@@ -1,5 +1,5 @@
 [% allow_creation = !c.cobrand.only_authed_can_create || (c.user && c.user.from_body) %]
-[% RETURN IF NOT allow_creation %]
+[% RETURN IF NOT allow_creation OR problem.extra.closed_updates %]
```
davea: Does this check need to go into the Bromley `update-form.html` too?

dracos: Oh yes. Will steal refactor commit from other open PR that refactors new and add update to that.

templates/email/default/inactive-account.txt
dracos: Will do HTML version. A login link as in just a link to `/my` or one that logs you in? I think might not want the latter (risk of clicking by mistake, but former sure).

PR #2129, opened 2018-05-16
[Bromley] Email text and default to unticking show name box

dracos:
[skip changelog]
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1037
- [x] new para
- [x] Fix wrong site-name being used

### DIFF
--- a/templates/email/bromley/_problem-confirm_extra.html
+++ b/templates/email/bromley/_problem-confirm_extra.html
@@ -1,0 +1,6 @@
+</p>
+<p style="[% p_style %]">
+By clicking validate you consent that for the purposes of investigating and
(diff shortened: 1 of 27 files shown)


templates/email/bromley/_problem-confirm_extra.html
```diff
@@ -0,0 +1,4 @@
+By clicking validate you consent that for the purposes of investigating and
```
zarino: Maybe we should prefix this with a `</p><p style="[% p_style %]">` so that is displays as a new paragraph in the HTML email? Looks a bit intimidating when it’s just appended to the existing paragraph!

screen shot 2018-05-17 at 12 46 22

PR #2132, opened 2018-05-18
Avoid empty space either side of "key tools" drawer when open

### LINKED ISSUE #2124: Weird design quirk opening ward menu
On e.g.https://fix.bathnes.gov.uk/reports/Bath+and+North+East+Somerset clicking to view wards jumps in a bit for the drawer to go up. This has happened for a while, but did not used to happen.

zarino:
Fixes #2124.

# Before

screen shot 2018-05-18 at 16 00 15

# After

screen shot 2018-05-18 at 16 00 44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
         - Make OpenGraph description translatable.
         - Stop double-escaping title in alert-update email.
         - Use inspection states in response template admin.
+        - Fixed CSS padding/overflow bug during sidebar "drawer" animations. #2132
     - Admin improvements:
         - Inspectors can set non_public status of reports. #1992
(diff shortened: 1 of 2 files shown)


zarino:
> I thought the arrow changed 90 degrees when open, but perhaps I imagined that?!

The last big change to the chevron icons was https://github.com/mysociety/fixmystreet/commit/9d3248ee49 – I can’t see any evidence there that the chevron used to point downwards when the drawer was open, before that. But maybe it did even further back in time.

> Changelog entry needed, assuming this bug has been here for a while and isn't new.

Added!

dracos:
https://github.com/mysociety/fixmystreet/pull/1902 removed the `.hover` class that showed a different arrow when the drawer was open.

PR #2133, opened 2018-05-21
[fixmystreet.com] Add info about right to object

crowbot:
[skip changelog]


dracos:
Thanks! Merged as a33aa0720 (commit message edited to have `[fixmystreet.com]` at the start).

PR #2137, opened 2018-05-22
Reintroduce “nicetable” class for simple table styling

zarino:
While doing https://github.com/mysociety/collideoscope/issues/3 I noticed that the cookie tables we put into our Privacy Policy pages weren’t getting styled up. The narrow, 10px gutters, and lack of gridlines, made it hard to see where cells begin and end.

Meanwhile, it rang a bell that, recently, the old `.nicetable` style got [reintroduced in KIITC](https://github.com/mysociety/fixmystreet-commercial/commit/1eed07d2392b37b17201c748a8207757347582a6) to display the list of authorities and their assets.

I figured, why not reintroduce a simpler version of the old `.nicetable`, for those times you need to make tables legible; and then use it in all the Privacy Policies, and also on the KIITC `/reports` page.

This PR does the first two of those three things, and https://github.com/mysociety/fixmystreet-commercial/pull/1050 does the third.

Here’s how it looks on a very simple, single-row table, in `/about/privacy` under the `fixmystreet.com` theme:

screen shot 2018-05-22 at 15 27 01

[skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -2392,6 +2392,58 @@ a#geolocate_link.loading, .btn--geolocate.loading {
     vertical-align: -2px;
 }
 
+$nicetable-cell-padding: 0.6em 1.5em !default;
+$nicetable-border-color: #ccc !default;
+$nicetable-stripe-background: rgba($primary, 0.05) !default;
+$nicetable-hover-background: rgba($primary, 0.15) !default;
+
+.nicetable {
+    width: 100%;
+    margin-bottom: 1.5em;
+
+    td, th {
+        padding: $nicetable-cell-padding;
+        padding-#{$left}: 0;
+        border-bottom: 1px solid $nicetable-border-color;
+        text-align: $left;
+
+        &:last-child {
+            padding-#{$right}: 0;
+        }
+    }
+
+    // Give your .nicetable an extra .stripe class to stripe alternate rows.
+    &.stripe {
+        td, th {
+            padding: $nicetable-cell-padding;
+        }
+
+        tr:nth-child(even) {
+            background: $nicetable-stripe-background;
+        }
+    }
+
+    // Give your .nicetable an extra .hover class to enable row hover effect.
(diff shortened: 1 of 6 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -2392,6 +2392,39 @@ a#geolocate_link.loading, .btn--geolocate.loading {
...
 }
 
+$nicetable-hover-background: rgba($primary, 0.5);
+
+.nicetable {
+    width: 100%;
+    margin-bottom: 1.5em;
+
+    td, th {
+        padding: 0.6em 1.5em 0.6em 0;
```
dracos: Should this be `flip`ed for RTL? And the last-child below.

web/cobrands/sass/_base.scss
```diff
@@ -2392,6 +2392,39 @@ a#geolocate_link.loading, .btn--geolocate.loading {
     vertical-align: -2px;
 }
 
+$nicetable-hover-background: rgba($primary, 0.5);
+
+.nicetable {
+    width: 100%;
+    margin-bottom: 1.5em;
+
+    td, th {
```
dracos: I miss the alternate colouring :)

zarino: 😢 🌹 👋

web/cobrands/sass/_base.scss
zarino: Yes, it should, because `:last-child` will match the _left_-most child in RTL languages. Good catch!

web/cobrands/sass/_base.scss
zarino: Now includes a `.stripe` class you can add to the table to enable alternate row striping.

We used to do it via a `class="a"` on alternate rows, but that makes it tricky to add padding to the non-`a` siblings. And striping seems like the sort of enhancement that we can happily add for modern browsers (IE9 and up) through `:nth-child()` and leave out for IE8 and lower.

PR #2138, opened 2018-05-22
[FixaMinGata] Update privacy policy

jonkri:


dracos:
This has been merged into the user-inactivity branch, so will be dealt with when #2125 is merged.

PR #2147, opened 2018-05-31
Improve handling of loading spinner display.

dracos:
Going back to /around from /report/new can trigger loadend (and the
spinner hiding function) twice (if the strategy activation starts a
read), with triggerRead aborting that read and triggering the event
itself. This means the spinner is not displayed, as the count falls
below 0. We could pin the count above 0, but instead let’s log each
layer ID while ‘active’ and switch off when they’re all gone.

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -1,3 +1,15 @@
+if (!Object.keys) {
+  Object.keys = function(obj) {
+    var result = [];
+    for (var prop in obj) {
+      if (Object.prototype.hasOwnProperty.call(obj, prop)) {
+        result.push(prop);
+      }
+    }
+    return result;
+  };
+}
+
 var fixmystreet = fixmystreet || {};
 
 fixmystreet.utils = fixmystreet.utils || {};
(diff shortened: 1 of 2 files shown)


web/js/map-OpenLayers.js
```diff
@@ -247,18 +259,18 @@ $.extend(fixmystreet.utils, {
        * the spinner in the DOM.
        */
       loading_spinner: {
-          count: 0,
+          count: {},
           show: function() {
-              fixmystreet.maps.loading_spinner.count++;
-              if (fixmystreet.maps.loading_spinner.count > 0) {
+              fixmystreet.maps.loading_spinner.count[this.id] = 1;
```
davea: Previously `fixmystreet.maps.loading_spinner.show` could be called from anywhere, but this change requires it to be called in the context of an OpenLayers event due to `this.id`. Not necessarily a problem, as it is currently only called when an event has fired, but worth noting this implicit coupling.

This could be worked around by using `fixmystreet.maps.loading_spinner.show.call({id: 'test'})`, or something similar.

PR #2148, opened 2018-05-31
[Zurich] Don't delete extra fields when editing contact

davea:
Editing a contact with extra fields would cause those fields to be
deleted because they weren't included in the POST request when saving.
This change ensures they're present in the DOM, but hidden by default.

[skip changelog]

### DIFF
--- a/templates/web/zurich/admin/contact-form.html
+++ b/templates/web/zurich/admin/contact-form.html
@@ -19,6 +19,16 @@
     <input type="text" class="form-control" name="email" value="[% contact.email | html %]" size="30">
 
     <p>
+      <strong>[% loc('Extra fields:') %]</strong>
+      [% IF contact.in_storage %]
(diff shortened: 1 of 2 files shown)


templates/web/zurich/admin/contact-form.html
```diff
@@ -18,6 +18,16 @@ <h1>[% contact.category_display | html %]</h1>
     <p><strong>[% loc('Email:') %] </strong>
     <input type="text" class="form-control" name="email" value="[% contact.email | html %]" size="30">
 
+    <p>
+      <strong>[% loc('Extra fields:') %]</strong>
+      [% IF contact.in_storage %]
+        <a href="#" class="js-show-extra-fields">([% loc('show') %])</a>
```
dracos: Wan a `hidden-nojs` class here too?

davea: Good thinking!

PR #2149, opened 2018-06-05
add reminder to include link to report on contact form

struan:
We get quite a few emails to support that are about a specific problem
but fail to include a link to that problem. This adds some text to the
contact form to ask people to do so.

### DIFF
--- a/templates/web/base/contact/index.html
+++ b/templates/web/base/contact/index.html
@@ -103,6 +103,10 @@
         [% END %]
         <textarea class="form-control required" name="message" id="form_message" rows="7" cols="50">[% message | html %]</textarea>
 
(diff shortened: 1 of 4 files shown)


templates/web/base/contact/index.html
```diff
@@ -103,6 +103,10 @@ <h2>[% problem.title_safe | html %]</h2>
         [% END %]
         <textarea class="form-control required" name="message" id="form_message" rows="7" cols="50">[% message | html %]</textarea>
 
+        [% IF NOT problem AND NOT update %]
+        <p>[% loc('If your are contacting us about a specific report or update please include a link to the report in the message.') %]</p>
```
abimysoc: Typo: "your are" should be "you are".

PR #2152, opened 2018-06-06
Don’t delete email contacts on devolved bodies.

dracos:
If we assume that any body with can_be_devolved set is probably a mix of
Open311/email, we can ignore any email address contacts for such a body.
[skip changelog]

### DIFF
--- a/perllib/Open311/PopulateServiceList.pm
+++ b/perllib/Open311/PopulateServiceList.pm
@@ -320,6 +320,14 @@ sub _delete_contacts_not_in_service_list {
         }
     );
 
+    if ($self->_current_body->can_be_devolved) {
+        # If the body has can_be_devolved switched on, it's most likely a
+        # combination of Open311/email, so ignore any email addresses.
+        $found_contacts = $found_contacts->search(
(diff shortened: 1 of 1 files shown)


perllib/Open311/PopulateServiceList.pm
```diff
@@ -335,27 +343,6 @@ sub _delete_contacts_not_in_service_list {
...
-    # the email addresses
-    if ($self->_current_body->name eq 'Warwickshire County Council' ||
-        $self->_current_body->name eq 'Bristol City Council' ||
-        $self->_current_body->name eq 'Bath and North East Somerset Council') {
-        $found_contacts = $found_contacts->search(
-            {
-                email => { -not_like => '%@%' }
-            }
-        );
-    } elsif ($self->_current_body->name eq 'East Hertfordshire District Council' ||
```
struan: Is it ok to remove this?

dracos: Yeah, the Stevenage/east Herts "Other" category is also an email, so should be caught by the same thing.

PR #2154, opened 2018-06-07
[Oxfordshire] Rebrand to match new oxfordshire.gov.uk website

zarino:
# Home page

screenshot-2018-6-19 fixmystreet

# Map page

screenshot-2018-6-19 viewing a location fixmystreet

Fixes mysociety/fixmystreet-commercial#1046. Nav menu links use flexbox to allow for extra items to be added (eg: admin links for staff users). Floated fallback in place for IE9.

[skip changelog]


web/cobrands/oxfordshire/base.scss
```diff
@@ -5,6 +5,17 @@
 @import "../sass/base";
 @import "../sass/report_list_pins";
 
+body {
+    background-color: $base_bg;
```
dracos: Should we make this a variable for all cobrands? I see it could default to white, but a few would change it.

web/cobrands/oxfordshire/base.scss
```diff
@@ -28,6 +28,33 @@ a {
...
+    height: 0;
+    padding-top: 52px;
+    overflow: hidden;
+    text-indent: 0;
+    margin: 0 -5px;
+}
+
+#nav-link {
+    height: 0;
+    padding-top: 3em;
```
dracos: Should this change to zero height + padding be made to `sass` for everywhere? If not, is it easier to keep it consistent with base?

web/cobrands/oxfordshire/layout.scss
```diff
@@ -42,6 +42,7 @@ $mappage-header-height: 72px;
 }
 
 #site-header {
+    @include box-sizing(border-box); // so that $mappage-header-height doesn't have to account for internal padding
```
dracos: Sure this is okay, but the comment on header-height is including some padding?

dracos:
Not outdated, github! Grr!

dracos:
Merged as part of occ-integration

PR #2155, opened 2018-06-08
include automated field when editing contacts

### LINKED ISSUE #2136: Editing contact removes server_set from extra data
Unlikely occurrence, but just happened to me :)

struan:
Display fields that are automatically filled in the contact editing form
otherwise the data is lost. Also include the automated field in the
form.

Fixes #2136

### DIFF
--- a/templates/web/base/admin/extra-metadata-form.html
+++ b/templates/web/base/admin/extra-metadata-form.html
@@ -6,7 +6,17 @@
         <div class="admin-hint"><p>[% loc('The ordering of this field on the report page. Fields are shown in ascending order according to this value.') %]</p></div>
         <label>
             [% loc('Order') %]
-            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order | html %]">
+            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order OR 0 | html %]">
+        </label>
+
+        <div class="admin-hint"><p>[% loc('Whether the field is displayed to the user, included as a hidden field and automatically populated, or set by the server upon Open311 submission. This field is usually set automatically.') %]</p></div>
+        <label>
+            [% loc('Automated') %]
+            <select name="metadata[[% loop.index %]].automated" data-field-name="automated" class="js-metadata-item-type">
+                <option value="" [% meta.automated == '' ? 'selected' : '' %]></option>
+                <option value="server_set" [% meta.automated == 'server_set' ? 'selected' : '' %]>[% loc('Server Set') %]</option>
+                <option value="hidden_field" [% meta.automated == 'hidden_field' ? 'selected' : '' %]>[% loc('Hidden Field') %]</option>
+            </select>
         </label>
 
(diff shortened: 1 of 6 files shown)


templates/web/base/admin/extra-metadata-form.html
```diff
@@ -6,7 +6,13 @@
...
         <label>
             [% loc('Order') %]
-            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order | html %]">
+            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order OR 0 | html %]">
+        </label>
+
+        <div class="admin-hint"><p>[% loc('Whether the field is displayed to the user, in which case it will be blank, otherwise it will be automatically populated. This field is set usually set automatically.') %]</p></div>
+        <label>
+            [% loc('Automated') %]
+            <input name="metadata[[% loop.index %]].automated" data-field-name="automated" type=text value="[% meta.automated | html %]">
```
davea: Would this be better as a `<select>` given the limited number of options?

struan: Yeah, good point.

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -6,7 +6,18 @@
         <div class="admin-hint"><p>[% loc('The ordering of this field on the report page. Fields are shown in ascending order according to this value.') %]</p></div>
         <label>
             [% loc('Order') %]
-            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order | html %]">
+            <input name="metadata[[% loop.index %]].order" data-field-name="order" type=text value="[% meta.order OR 0 | html %]">
+        </label>
+
+        <div class="admin-hint"><p>[% loc('Whether the field is displayed to the user, in which case it will be blank, otherwise it will be automatically populated. This field is set usually set automatically.') %]</p></div>
```
dracos: Typo here: "set usually set". This doesn't explain the hidden field/server set difference, perhaps something like "Whether the field is displayed to the user, output in the form as a hidden field to be populated automatically, or only generated by the server upon Open311 submission" or similar?

templates/web/base/admin/extra-metadata-form.html
dracos: I see the select has been added, but the `<input>` is still here unless I'm being dense :)

templates/web/base/admin/extra-metadata-form.html
```diff
@@ -6,7 +6,18 @@
...
+        <label>
+            [% loc('Automated') %]
+            <input name="metadata[[% loop.index %]].automated" data-field-name="automated" type=text value="[% meta.automated | html %]">
+            <select name="metadata[[% loop.index %]].automated" data-field-name="automated" class="js-metadata-item-type">
+                <option value="" [% meta.automated == '' ? 'selected' : '' %]></option>
+                <option value="server_set" [% meta.automated == 'string' ? 'selected' : '' %]>[% loc('Server Set') %]</option>
```
dracos: should be `server_set`, not `string`

PR #2158, opened 2018-06-13
Hide old reports by default

### LINKED ISSUE #2098: Hide fixed reports by default
> Can you remove pins from the map a time after they've been fixed? I'm seeing some that are years old 

> The location of post code N10 2HR is so over populated with problem flags on the map that it is very difficult to add new problem flag locations in that area. Can you please clear the very old report flags so that new ones can be added?

>  Just wondering if you have a policy for removing older posts, particularly those that are reported as fixed? The map is getting crowded with fault reports, and it can be difficult to find an appropriate blank space in which to pin a new problem, or to see if it has already been reported.

> Is there a way of removing the location symbols relating to old and fixed reports from the map. It is getting very cluttered in places. Many of these relate to problems that are several years old. 

> Would it be possible to make all old & resolved issues/ content invisible ? The site looks a bit too busy and not user-friendly anymore now, I think

> Can you please remove/delet fixed problems after say 6 months because there are locations where it is difficult to report a new problem.

> When using the site it is getting more difficult to place mark an incident as there are so many old issues still on the maps. Surely it would possible to remove fixed incident place markers when entering a new incident. There is no need to show old things that are going back months

> Hi, please could you consider remiving the yellow report icons frim the maps which are older than say 6 months. This is because there are some locations such as the Golders Green war memorial area. This has so many yellow icons that it isnpractically impossible to create a new repor.

>  Can we remove old reports to make space for new

> on some roads , old reports are still on your site , which stops new reports being posted . for example simonsway , near manchester airport , there are so many old reports that it is virtually impossible to report a new one

> When trying to place a new problem on the map very often it is a busy place-ie High St so A) There a a jumble of 'pins' B) a lot of the problems have been cleared and are old. Can the map be cleared of old issues sometime.

> E.G. Faulty streetlight deported 2014 is still showing on the map. The map is very cluttered. Should there be a timeout to remove such flags? The site risks using credibility. It is also difficult to read all flags when there are multiple in close proximity.

> When are the reports removed. Now more people are using the app the there are too many point. Either they need to be removed after a certain amount of time or if they’ve been reported fixed. There are reports from 2014 still showing.

>  Can you take down the jobs that have been fixed it’s very confusing 

> Can issues that have been fixed be removed from the map, it will save a lot of time and encourage more use as the map, especially on the app can become very cluttered. 

> it is impossible to see the map due to historic reports that have been fixed, is it possible to remove them?

> I think it could be a good idea to have a box to tick/button to click, to be able to only see recent reports. Historic reports are interesting to read, and as you have stated in the help section, they are of research value etc. However, when trying to look at an area with lots of pins all in the same place, it becomes hard to see and read them. To have a way to hide reports that are 5 - 10 years old would make it much easier to use.

struan:
Hides reports older than 6 months on the around page by default.

Fixes #2098

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -253,6 +253,7 @@ sub map_features : Private {
 
     $c->forward( '/reports/stash_report_filter_status' );
     $c->forward( '/reports/stash_report_sort', [ 'created-desc' ]);
+    $c->stash->{show_old_reports} = $c->get_param('show_old_reports');
 
     return if $c->get_param('js'); # JS will request the same (or more) data client side
 
--- a/perllib/FixMyStreet/DB/ResultSet/Nearby.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Nearby.pm
@@ -10,7 +10,7 @@ sub to_body {
 }
 
 sub nearby {
-    my ( $rs, $c, $dist, $ids, $limit, $mid_lat, $mid_lon, $categories, $states, $extra_params ) = @_;
+    my ( $rs, $c, $dist, $ids, $limit, $mid_lat, $mid_lon, $categories, $states, $extra_params, $report_age ) = @_;
 
     unless ( $states ) {
         $states = FixMyStreet::DB::Result::Problem->visible_states();
@@ -22,6 +22,9 @@ sub nearby {
     $params->{id} = { -not_in => $ids }
         if $ids;
     $params->{category} = $categories if $categories && @$categories;
+
+    $params->{$c->stash->{report_age_field}} = { '>=', \"current_timestamp-'$report_age'::interval" }
+        if $report_age;
 
     FixMyStreet::DB::ResultSet::Problem->non_public_if_possible($params, $c);
 
--- a/perllib/FixMyStreet/DB/ResultSet/Problem.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Problem.pm
@@ -172,6 +172,9 @@ sub around_map {
             latitude => { '>=', $p{min_lat}, '<', $p{max_lat} },
             longitude => { '>=', $p{min_lon}, '<', $p{max_lon} },
     };
+
+    $q->{$c->stash->{report_age_field}} = { '>=', \"current_timestamp-'$p{report_age}'::interval" } if
+        $p{report_age};
     $q->{category} = $p{categories} if $p{categories} && @{$p{categories}};
 
     $rs->non_public_if_possible($q, $c);
--- a/templates/web/base/reports/_list-filters.html
+++ b/templates/web/base/reports/_list-filters.html
@@ -80,6 +80,13 @@
             </select>
             <input type="submit" name="filter_update" value="[% loc('Go') %]">
         </p>
+        [% IF page == 'around' %]
+        <p id="show_old_reports_wrapper" class="report-list-filters[% ' hidden' UNLESS num_old_reports > 0 %]">
+            <label for="show_old_reports">[% loc('Show older reports') %]</label>
+            <input type="checkbox" name="show_old_reports" id="show_old_reports" value="1"[% ' checked' IF show_old_reports %]>
+            <input type="submit" name="filter_update" value="[% loc('Go') %]">
+        </p>
+        [% END %]
 
 [% IF use_form_wrapper %]
     </form>
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1291,6 +1291,9 @@ $(function() {
                 } else if ('page_change' in e.state) {
                     $('.pagination').data('page', e.state.page_change.page)
                         .trigger('change.filters');
+                    if ( fixmystreet.page != 'reports' ) {
+                        fixmystreet.display.reports_list(location.href);
+                    }
                 } else if ('filter_change' in e.state) {
                     $('#filter_categories').val(e.state.filter_change.filter_categories);
                     $('#statuses').val(e.state.filter_change.statuses);
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1280,7 +1280,7 @@ $(function() {
                     if (fixmystreet.utils && fixmystreet.utils.parse_query_string) {
                         var qs = fixmystreet.utils.parse_query_string();
                         var page = qs.p || 1;
-                        $('.pagination').data('page', page)
+                        $('.pagination:first').data('page', page)
                             .trigger('change.filters');
                     }
                     fixmystreet.display.reports_list(location.href);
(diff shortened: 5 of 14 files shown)


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -172,6 +172,10 @@ sub around_map {
             latitude => { '>=', $p{min_lat}, '<', $p{max_lat} },
             longitude => { '>=', $p{min_lon}, '<', $p{max_lon} },
     };
+
+    my $interval = $c->cobrand->report_age;
+    $q->{lastupdate} = { '>=', \"current_timestamp-'$interval'::interval" } unless
```
dracos: Using `lastupdate` means very old reports could be shown if updated recently. Which is fine, but just to note as "Show older problems" could be taken to just mean the problems are old, not their latest activity.
And then, should this vary depending on what sort order `$p{order}` is set to? If you're sorting by newest, perhaps it should also be newest - last updated, lastupdate, and then oldest it probably doesn't make much sense to do it at all? Dunno.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1280,12 +1280,18 @@ $(function() {
                 } else if ('newReportAtLonlat' in e.state) {
                     fixmystreet.display.begin_report(e.state.newReportAtLonlat, false);
                 } else if ('page_change' in e.state) {
-                    $('.pagination').data('page', e.state.page_change.page)
+                    fixmystreet.markers.protocol.use_page = true;
+                    $('#show_old_reports').prop('checked', e.state.page_change.show_old_reports == 'on');
+                    $('.pagination:first').data('page', e.state.page_change.page) //;
                         .trigger('change.filters');
+                    if ( fixmystreet.page != 'reports' ) {
```
dracos: I *think* it is safe to call this regardless of the page already being this (see how filter_change just below always calls it), maybe?

perllib/FixMyStreet/DB/ResultSet/Problem.pm
struan: I see your point on `lastupdate` so will change that to `confirmed`. It's also probably more in line with what councils expect. 
Not sure that changing which reports show dependant on sort order is a very obvious behaviour.

web/js/map-OpenLayers.js
```diff
@@ -349,7 +349,9 @@ $.extend(fixmystreet.utils, {
 
     function replace_query_parameter(qs, id, key) {
         var value = $('#' + id).val();
-        if (value) {
+        if ( $('#' + id).prop('type') == 'checkbox' ) {
+            value = $('#' + id).prop('checked') ? 'on' : '';
```
dracos: This no longer updates `qs` to put the value in it. I think you want one if/else statement first to set `value`, then another to set/delete `qs[key]`?
Also there seems to be confusion as to whether the value is `1` or `on`. Could you put a `value=1` in the checkbox and then use `1` everywhere without worrying?

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -253,6 +253,7 @@ sub map_features : Private {
 
     $c->forward( '/reports/stash_report_filter_status' );
     $c->forward( '/reports/stash_report_sort', [ 'created-desc' ]);
+    $c->stash->{show_old_reports} = $c->get_param('show_old_reports') && $c->get_param('show_old_reports') eq '1' ? 'on' : 'off';
```
dracos: Can this not be `$c->get_param('show_old_reports') ? 'on' : 'off';`, if it's either set or not set?

dracos: And then perhaps reduce it to only `$c->get_param('show_old_reports')` and then you can do `IF show_old_reports` rather than have to compare to a string.

perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -172,6 +172,10 @@ sub around_map {
             latitude => { '>=', $p{min_lat}, '<', $p{max_lat} },
             longitude => { '>=', $p{min_lon}, '<', $p{max_lon} },
     };
+
+    my $interval = $p{report_age};
```
dracos: Not sure `$interval` var needed here, isn't one in Nearby.pm (or if you do want to use it both times in the line underneath rather than only once :) )

perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -172,6 +172,10 @@ sub around_map {
...
+    $q->{confirmed} = { '>=', \"current_timestamp-'$interval'::interval" } if
```
dracos: Been thinking about this some more, and I do think it needs to adjust to sort order. For example, if I select sort by recently updated, I expect a report updated yesterday to appear, even if it was created 2 years ago.
I think it should work as follows:
* sort order newest/oldest/most commented, use confirmed;
* sort order recently updated/least recently updated, use lastupdate.

I think that makes sense?

templates/web/base/reports/_list-filters.html
```diff
@@ -73,6 +73,13 @@
             </select>
             <input type="submit" name="filter_update" value="[% loc('Go') %]">
         </p>
+        [% IF page != 'reports' %]
+        <p id="show_old_reports_wrapper" class="report-list-filters[% ' hidden' UNLESS num_old_reports > 0 %]">
+            <label for="show_old_reports">[% loc('Show older reports') %]</label>
+            <input type="checkbox" name="show_old_reports" id="show_old_reports"[% ' checked' IF show_old_reports == 'on' %]>
```
dracos: I suggested a `value=1` here to simplify all the on/off stuff to 1/blank.

dracos:
Oh, also wonder why a Zurich test fails with this PR, some bad interaction going on somehow?

perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -173,7 +173,11 @@ sub around_map {
             longitude => { '>=', $p{min_lon}, '<', $p{max_lon} },
     };
 
-    my $age_field = $c->stash->{sort_key} =~ /updated/ ? 'lastupdate' : 'confirmed';
+    # this is a { -desc => order } hash so flatten to array
```
dracos: It's only that if it's `desc`, if it was ascending then it's only a string: lastupdate/confirmed/comment_count.
(and if it's comment_count you won't want to be using that for a time comparison :)

perllib/FixMyStreet/DB/ResultSet/Nearby.pm
```diff
@@ -23,6 +23,9 @@ sub nearby {
         if $ids;
     $params->{category} = $categories if $categories && @$categories;
 
+    $params->{confirmed} = { '>=', \"current_timestamp-'$report_age'::interval" }
```
dracos: Oh, this presumably needs the same change as Problem.pm?

dracos:
Pushed a couple of commits to fix back button issues due to on/off changes, and going back to the initial page needing to look at the URL for the tickbox. Hopefully that's it, apart from the Nearby.pm thing above, unless there was a reason that was different.

dracos:
Spotted I had used `show_all_reports` in one place instead of `show_old_reports`, plus changed uses of `page != reports` to `page == around` because I don't think we want this behaviour on `/my` either.

PR #2162, opened 2018-06-18
[Bromley] Improve geocoding

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1064


davea:
`_error_multiple.html` and `choose.html` say "We show up to ten matches" which presumably will be inaccurate if the OSM geocoder is in use?

PR #2164, opened 2018-06-18
send unresponsive details back with category extras

struan:
This allows the mobile app to explicitly detect when a body is
unresponsive and display the category_extra text.

For mysociety/fixmystreet-mobile#266


struan:
I've added in some extra bits to a) test this and b) send back an unresponsive key when fetching category data so we can use that in the mobile app in future.

PR #2168, opened 2018-06-21
Use CSV escaping for categories in URLs.

### LINKED ISSUE #2166: filter_category query parameter does not return all reports
### Description of problem

`filter_category` query parameter does not return all reports.

### Version

2.3.4.

### Steps to reproduce

Visit https://fixamingata.se/reports/Klippans+kommun?ajax=1&filter_category=Gator,%20skyltar%20och%20parkering&sort=updated-desc&p=1&bbox=12.751934920312,55.996038045544,13.751690779688,56.236442780696.

### Actual results

No reports are included in the response.

### Expected results

The report https://fixamingata.se/report/8946 is included in the response.

### How reproducible

Seemingly always.

### Additional information

The report is visible in the list here: https://fixamingata.se/reports/Klippans+kommun. The category was renamed yesterday.

dracos:
Categories could contain commas, so splitting on comma is not good enough.
Let’s escape the fields as if it’s a line in CSV. Fixes #2166.

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -15,6 +15,18 @@ var fixmystreet = fixmystreet || {};
 fixmystreet.utils = fixmystreet.utils || {};
 
 $.extend(fixmystreet.utils, {
+    array_to_csv_line: function(arr) {
+        var out = [], s;
+        for (var i=0; i<arr.length; i++) {
+            s = arr[i];
+            if (/[",]/.test(s)) {
+                s = '"' + s.replace('"', '""') + '"';
+            }
+            out.push(s);
+        }
+        return out.join(',');
+    },
+
     parse_query_string: function() {
         var qs = {};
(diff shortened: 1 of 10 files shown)


web/js/map-OpenLayers.js
```diff
@@ -898,8 +910,8 @@ OpenLayers.Protocol.FixMyStreet = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
         options.params = options.params || {};
         $.each({ filter_category: 'filter_categories', status: 'statuses', sort: 'sort' }, function(key, id) {
             var val = $('#' + id).val();
-            if (val !== undefined) {
-                options.params[key] = val;
+            if (val && val.length) {
+                options.params[key] = val.join ? fixmystreet.utils.array_to_csv_line(val) : val;
```
davea: Nice test to distinguish strings from arrays 💯

PR #2174, opened 2018-06-25
BANES parks & grounds asset layer

davea:
- Adds a new parks & grounds WFS layer
 - Allows contact group to be edited in admin

Fixes mysociety/fixmystreet-commercial#1029.


web/cobrands/fixmystreet/assets.js
```diff
@@ -227,10 +241,14 @@ function asset_selected(e) {
 
 function asset_unselected(e) {
     fixmystreet.markers.setVisibility(true);
+    selected_feature = null;
+    if (e.feature.attributes._preserve_form_values_on_unselect) {
```
dracos: Looks like this works fine, and thanks for the big comment further up :) Just wondering why you need to set this value at all, though, can't you do `if (this.fixmystreet.disable_pin_snapping)` in `asset_unselected` as well?

templates/web/bathnes/around/intro.html
```diff
@@ -1 +1,2 @@
-<h1>Report problems on the road</h1>
+<h1>Report a problem</h1>
+<h1>Roads, pavements, and parks</h1>
```
dracos: Should this be one `<h1>` with a `<br>` or something?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -356,6 +356,10 @@ sub update_contacts : Private {
         if ( $c->get_param('reputation_threshold') ) {
             $contact->set_extra_metadata( reputation_threshold => int($c->get_param('reputation_threshold')) );
         }
+        if ( my $group = $c->get_param('group') ) {
```
dracos: No way to unset the group.

web/cobrands/fixmystreet/assets.js
```diff
@@ -546,6 +538,18 @@ fixmystreet.assets = {
             select_feature_control = new OpenLayers.Control.SelectFeature( asset_layer );
             asset_layer.events.register( 'featureselected', asset_layer, asset_selected);
             asset_layer.events.register( 'featureunselected', asset_layer, asset_unselected);
+            if (options.disable_pin_snapping) {
+                // The pin is snapped to the centre of a feature by the select
+                // handler. We can stop this handler from running, and the pin
+                // being snapped, by returning false from a beforefeatureselected
+                // event handler. This handler does need to make sure the
+                // attributes of the clicked feature are applied to the extra
+                // details form fields first though.
```
dracos: Do we want to trigger an `assets:selected` event, in case there is a roads layer that might be interacting here as well? Or just note this in case that happens in future.

web/cobrands/fixmystreet/assets.js
```diff
@@ -360,6 +349,9 @@ function select_nearest_asset() {
     var nearest_feature = this.getNearestFeature(marker.geometry, threshold);
     if (nearest_feature) {
         get_select_control(this).select(nearest_feature);
+    } else if (this.fixmystreet.disable_pin_snapping) {
```
dracos: Not sure this fixes the issue, as this is only run on layer load isn't it? So what clears the fields if you click elsewhere on the map?

PR #2175, opened 2018-06-26
[UK] Fix issue when body ID not equal to MapIt ID.

dracos:
Hitherto when creating a body or ward alert on a UK site, the MapIt area ID has been stored instead of the body ID. This is okay for www.fixmystreet.com which for historical reasons does have body IDs that match MapIt area IDs, but other UK-based sites may well not. The alert lookup looks for body ID, meaning those alerts will not work. Save the body ID instead, plus fix some tests that were making the same assumption.

### DIFF
--- a/perllib/FixMyStreet/Script/Alerts.pm
+++ b/perllib/FixMyStreet/Script/Alerts.pm
@@ -185,9 +185,12 @@ sub send() {
                     # Get a report object for its photo and static map
                     $data{report} = $schema->resultset('Problem')->find({ id => $row->{id} });
                 }
-                if ($ref eq 'area_problems' || $ref eq 'council_problems' || $ref eq 'ward_problems') {
+                if ($ref eq 'area_problems') {
                     my $va_info = mySociety::MaPit::call('area', $row->{alert_parameter});
                     $data{area_name} = $va_info->{name};
+                } elsif ($ref eq 'council_problems' || $ref eq 'ward_problems') {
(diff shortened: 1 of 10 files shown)


dracos:
"Should this not also apply to the Default cobrand?" that doesn't use the `council:` or `ward:` type alerts, only `area:` which does use MapIt ID.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -185,9 +185,12 @@ sub send() {
                     # Get a report object for its photo and static map
                     $data{report} = $schema->resultset('Problem')->find({ id => $row->{id} });
                 }
-                if ($ref eq 'area_problems' || $ref eq 'council_problems' || $ref eq 'ward_problems') {
+                if ($ref eq 'area_problems') {
                     my $va_info = mySociety::MaPit::call('area', $row->{alert_parameter});
                     $data{area_name} = $va_info->{name};
+                } elsif ($ref eq 'council_problems' || $ref eq 'ward_problems') {
+                    my $body = FixMyStreet::DB->resultset('Body')->find({ id => $row->{alert_parameter} });
+                    $data{area_name} = $body->name;
```
struan: Is there not the potential that this name is different from the one in MapIt and hence different from the one shown on the sign up screen?

dracos: Possibly, but not in any current usage.

PR #2178, opened 2018-06-29
Body specific asset layers

struan:
Associate layers with a body and only show them if the body covers the area displayed on the map. This is to allow us to use asset layers on the national site and only display the appropriate layers for an area.

This adds a body property to layers which specifies the body associated with the layer, a bodies property to the fixmystreet object to specify which bodies are associated with the area shown on the map, and the associated code to only display that layer if the two match.

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1086
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -9,6 +9,7 @@ use Encode;
 use JSON::MaybeXS;
 use Utils;
 use Try::Tiny;
+use Text::CSV;
 
 =head1 NAME
 
@@ -230,6 +231,10 @@ sub check_and_stash_category : Private {
     my $all_areas = $c->stash->{all_areas};
     my @bodies = $c->model('DB::Body')->active->for_areas(keys %$all_areas)->all;
     my %bodies = map { $_->id => $_ } @bodies;
+    my @list_of_names = map { $_->name } values %bodies;
+    my $csv = Text::CSV->new();
+    $csv->combine(@list_of_names);
+    $c->{stash}->{list_of_names_as_string} = $csv->string;
 
     my @categories = $c->model('DB::Contact')->not_deleted->search(
         {
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -208,6 +208,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
 
     my $extra_titles_list = $c->cobrand->title_list($c->stash->{all_areas});
 
+    my @list_of_names = map { $_->name } values %{$c->stash->{bodies}};
     my $contribute_as = {};
     if ($c->user_exists) {
         my @bodies = keys %{$c->stash->{bodies}};
@@ -221,6 +222,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
 
     my $body = encode_json(
         {
+            bodies          => \@list_of_names,
             councils_text   => $councils_text,
             councils_text_private => $councils_text_private,
             category        => $category,
@@ -254,8 +256,9 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
     $category = '' if $category eq _('-- Pick a category --');
 
     my $bodies = $c->forward('contacts_to_bodies', [ $category ]);
+    my $list_of_names = [ map { $_->name } @$bodies ];
     my $vars = {
-        $category ? (list_of_names => [ map { $_->name } @$bodies ]) : (),
+        $category ? (list_of_names => $list_of_names) : (),
     };
 
     my $category_extra = '';
@@ -281,12 +284,14 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
     my $councils_text_private = $c->render_fragment( 'report/new/councils_text_private.html');
 
     $unresponsive = $c->stash->{unresponsive}->{$category} || $c->stash->{unresponsive}->{ALL} || '';
+
     my $body = encode_json({
         category_extra => $category_extra,
         councils_text => $councils_text,
         councils_text_private => $councils_text_private,
         category_extra_json => $category_extra_json,
         unresponsive => $unresponsive,
+        bodies => $list_of_names,
     });
 
     $c->res->content_type('application/json; charset=utf-8');
@@ -706,6 +711,12 @@ sub setup_categories_and_bodies : Private {
     $c->stash->{category_extras_hidden}  = \%category_extras_hidden;
     $c->stash->{non_public_categories}  = \%non_public_categories;
     $c->stash->{extra_name_info} = $first_area->{id} == COUNCIL_ID_BROMLEY ? 1 : 0;
+
+    # escape these so we can then split on , cleanly in the template.
+    my @list_of_names = map { $_->name } values %bodies_to_list;
+    my $csv = Text::CSV->new();
+    $csv->combine(@list_of_names);
+    $c->stash->{list_of_names_as_string} = $csv->string;
 
     my @missing_details_bodies = grep { !$bodies_to_list{$_->id} } values %bodies;
     my @missing_details_body_names = map { $_->name } @missing_details_bodies;
--- a/templates/web/base/maps/openlayers.html
+++ b/templates/web/base/maps/openlayers.html
@@ -25,6 +25,9 @@
 [% IF include_key -%]
     data-key='[% c.config.BING_MAPS_API_KEY %]'
 [%- END -%]
+[% IF list_of_names_as_string -%]
+    data-bodies='[% list_of_names_as_string | html %]'
+[%- END -%]
 >
 </div>
 <div id="map_box" aria-hidden="true">
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -16,12 +16,59 @@ var fixmystreet = fixmystreet || {};
     };
 })();
 
+OpenLayers.Layer.VectorAsset = OpenLayers.Class(OpenLayers.Layer.Vector, {
+    initialize: function(name, options) {
+        OpenLayers.Layer.Vector.prototype.initialize.apply(this, arguments);
+        $(fixmystreet).on('report_new:category_change', this.update_layer_visibility.bind(this));
+    },
+
+    update_layer_visibility: function() {
+        if (!fixmystreet.map) {
+          return;
+        }
+
+        if (!this.fixmystreet.always_visible) {
+            // Show/hide the asset layer when the category is chosen
+            var category = $('#problem_form select#form_category').val();
+            if (fixmystreet.assets.check_layer_relevant(this.fixmystreet, category)) {
+                this.setVisibility(true);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(true);
+                }
+                this.zoom_to_assets();
+            } else {
+                this.setVisibility(false);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(false);
+                }
+            }
+        } else {
+            if (this.fixmystreet.body) {
+                this.setVisibility(OpenLayers.Util.indexOf(fixmystreet.bodies, this.fixmystreet.body) != -1 );
+            }
+        }
+    },
+
+    zoom_to_assets: function() {
+        // This function is called when the asset category is
+        // selected, and will zoom the map in to the first level that
+        // makes the asset layer visible if it's not already shown.
+        if (!this.inRange) {
+            var firstVisibleResolution = this.resolutions[0];
+            var zoomLevel = fixmystreet.map.getZoomForResolution(firstVisibleResolution);
+            fixmystreet.map.zoomTo(zoomLevel);
+        }
+    },
+
+    CLASS_NAME: 'OpenLayers.Layer.VectorAsset'
+});
+
 // Handles layers such as USRN, TfL roads, and the like
-OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.Vector, {
+OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorAsset, {
     selected_feature: null,
 
     initialize: function(name, options) {
-        OpenLayers.Layer.Vector.prototype.initialize.apply(this, arguments);
+        OpenLayers.Layer.VectorAsset.prototype.initialize.apply(this, arguments);
         $(fixmystreet).on('maps:update_pin', this.checkFeature.bind(this));
         $(fixmystreet).on('assets:selected', this.checkFeature.bind(this));
         // Might only be able to fill in fields once they've been returned from the server
@@ -31,10 +78,13 @@ OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.Vector, {
     },
 
     checkFeature: function(evt, lonlat) {
+        if (!this.getVisibility()) {
+          return;
+        }
         this.getNearest(lonlat);
         this.updateUSRNField();
         if (this.fixmystreet.road) {
-            var valid_category = this.fixmystreet.all_categories || (this.fixmystreet.asset_category && this.fixmystreet.asset_category.indexOf($('select#form_category').val()) != -1);
+            var valid_category = this.fixmystreet.all_categories || (this.fixmystreet.asset_category && fixmystreet.assets.check_layer_relevant( this.fixmystreet, $('select#form_category').val() ) );
             if (!valid_category || !this.selected_feature) {
                 this.road_not_found();
             } else {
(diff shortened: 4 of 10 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -16,8 +16,58 @@ var fixmystreet = fixmystreet || {};
...
+                }
+                this.zoom_to_assets();
+            } else {
+                this.setVisibility(false);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(false);
+                }
+            }
+        } else {
+            if (fixmystreet.bodies && this.fixmystreet.body) {
```
dracos: My worry here is if the centre of the map is in one council, but part of a council which there are assets is being shown and presumably any road-based layers should be shown there. Otherwise you could have a slight case (perhaps a race condition) of clicking on a Bromley part of the map that is centred on Bexley, and it not finding the TfL road asset (though I guess when the map panned and it all loaded in, it might then kick off another 'find' of the asset then? [1]) But I do see we don't want to bother HTTP-requsting checking a Bromley layer for a map in Scotland...

[1] Hmm, I guess if the parent/grandparent change I suggest is made, then the visibility event will fire before the check the features event, so it might be okay, though what if the assets need to be loaded from the server? Would be good to confirm this more robustly somehow!

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,8 +16,58 @@ var fixmystreet = fixmystreet || {};
...
+
+        if (!this.fixmystreet.always_visible) {
+            // Show/hide the asset layer when the category is chosen
+            var category = $('#problem_form select#form_category').val();
+            if (fixmystreet.assets.check_layer_relevant(this.fixmystreet, category)) {
+                this.setVisibility(true);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(true);
+                }
+                this.zoom_to_assets();
```
dracos: As this is now firing on both category change and when extras are received, could this cause a confusion at some point? (But see other comment made later.)

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,8 +16,58 @@ var fixmystreet = fixmystreet || {};
     };
 })();
 
+OpenLayers.Layer.VectorAsset = OpenLayers.Class(OpenLayers.Layer.Vector, {
+    initialize: function(name, options) {
+        OpenLayers.Layer.Vector.prototype.initialize.apply(this, arguments);
+        // Might only be able to fill in fields once they've been returned from the server
+        $(fixmystreet).on('report_new:category_change:extras_received', this.update_layer_visibility.bind(this));
```
dracos: Do we think this is necessary? As I imagine the extras are only ever going to cut the number of bodies, will it matter if we're still loading both two-tier assets if present? I can think of a scenario where we would want to do so (a county asking for info on district related things, type thing).

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -281,12 +282,15 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
     my $councils_text_private = $c->render_fragment( 'report/new/councils_text_private.html');
 
     $unresponsive = $c->stash->{unresponsive}->{$category} || $c->stash->{unresponsive}->{ALL} || '';
+
+    my @bodies = map { "" . $_->id } @$bodies;
```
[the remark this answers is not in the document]
struan: I want to make sure it's always a string in the JSON and this does that.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -883,6 +888,15 @@ $.extend(fixmystreet.set_up, {
             }
         });
     });
+  },
+
+  bodies: function() {
+     if (fixmystreet && !fixmystreet.bodies) {
+         if ($('#js-map-data') && $('#js-map-data').data('bodies') ) {
+             var bodies = '' + $('#js-map-data').data('bodies');
```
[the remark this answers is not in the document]
struan: Again to make sure that it's a string so the comparison works.

web/cobrands/fixmystreet/assets.js
struan: Given we don't show most assets until they click the pin and select a category I think this problem is largely moot. Yes, we might then need to load a road layer but with the parent/grandparent change that problem is fixed I think.

web/cobrands/fixmystreet/map.js
```diff
@@ -18,6 +18,12 @@ var fixmystreet = fixmystreet || {};
         }
     });
 
+    // bodies data needs slightly different handling
+    if (map_data.getAttribute('data-bodies')) {
+       var bodies = '' + map_data.getAttribute('data-bodies');
```
dracos: The attribute will always be a string, so no need for `'' + ` here.

dracos: In fact, looks like you could add `bodies` to the `map_key` array and not have this bit at all, replacing it with a line like the `fixmystreet.areas` one underneath.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -940,6 +943,12 @@ fixmystreet.update_pin = function(lonlat, savePushState) {
             lb.before(data.extra_name_info);
         }
 
+        if (data.bodies) {
+            fixmystreet.bodies = data.bodies;
+        } else {
+            fixmystreet.bodies = [];
+        }
```
dracos: `fixmystreet.bodies = data.bodies || [];`? :)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -883,7 +886,7 @@ $.extend(fixmystreet.set_up, {
             }
         });
     });
-  }
+  },
```
dracos: Bad comma.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -409,6 +409,9 @@ $.extend(fixmystreet.set_up, {
             } else {
                 $category_meta.empty();
             }
+            if (data.bodies) {
+                fixmystreet.bodies = data.bodies;
```
dracos: SHould this be like the other case and default to empty array?

web/cobrands/fixmystreet/assets.js
```diff
@@ -631,7 +649,12 @@ fixmystreet.assets = {
             fixmystreet.map.addControl(fixmystreet.assets.controls[i]);
             fixmystreet.assets.controls[i].activate();
         }
-    }
+    },
+
+    check_layer_relevant: function(layer, category) {
+      return layer.asset_category.indexOf(category) != -1 &&
+        ( !fixmystreet.bodies.length || !layer.body || fixmystreet.bodies.indexOf(layer.body + '') != -1 );
```
dracos: Ditto

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,12 +16,60 @@ var fixmystreet = fixmystreet || {};
...
+                this.zoom_to_assets();
+            } else {
+                this.setVisibility(false);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(false);
+                }
+            }
+        } else {
+            if (fixmystreet.bodies.length && this.fixmystreet.body) {
+                this.setVisibility(fixmystreet.bodies.indexOf('' + this.fixmystreet.body) != -1 );
```
dracos: Ditto

web/cobrands/fixmystreet/assets.js
```diff
@@ -631,7 +649,12 @@ fixmystreet.assets = {
             fixmystreet.map.addControl(fixmystreet.assets.controls[i]);
             fixmystreet.assets.controls[i].activate();
         }
-    }
+    },
+
+    check_layer_relevant: function(layer, category) {
+      return layer.asset_category.indexOf(category) != -1 &&
```
dracos: `indexOf` is only in IE9+. Site seems to work in IE8 fine apart from assets so would be good to not have this if possible. There’s an `OpenLayers.Util.indexOf` function could use.

dracos:
I've just realised that `master` already uses `body` on layer, in Bromley, with it set to "TfL", to override the body used there. That conflicts with the use here (which now that I think of it is using MapIt IDs assuming they're the same as body IDs) so I think that needs to be done differently, either using the body names everywhere (to match current usage) or something else.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -230,6 +230,9 @@ sub check_and_stash_category : Private {
     my $all_areas = $c->stash->{all_areas};
     my @bodies = $c->model('DB::Body')->active->for_areas(keys %$all_areas)->all;
     my %bodies = map { $_->id => $_ } @bodies;
+    my @body_names_to_list = map { $_->name } values %bodies;
+    $c->{stash}->{bodies_to_list} = \%bodies;
```
dracos: This line no longer needed

templates/web/base/maps/openlayers.html
```diff
@@ -25,6 +25,9 @@
 [% IF include_key -%]
     data-key='[% c.config.BING_MAPS_API_KEY %]'
 [%- END -%]
+[% IF bodies_to_list -%]
```
dracos: ...if this checks body_names_to_list

templates/web/base/maps/openlayers.html
```diff
@@ -25,6 +25,9 @@
...
+    data-bodies='[% body_names_to_list.join(',') %]'
```
dracos: This will break if any body name includes either a comma or a single quote. Copy what the category listing does with its CSV thing, the same should work here I think?

web/cobrands/fixmystreet/assets.js
```diff
@@ -631,7 +649,12 @@ fixmystreet.assets = {
             fixmystreet.map.addControl(fixmystreet.assets.controls[i]);
             fixmystreet.assets.controls[i].activate();
         }
-    }
+    },
+
+    check_layer_relevant: function(layer, category) {
+      return OpenLayers.Util.indexOf(layer.asset_category, category) != -1 &&
+        ( !fixmystreet.bodies.length || !layer.body || OpenLayers.Util.indexOf(fixmystreet.bodies, layer.body + '') != -1 );
```
dracos: This no longer needs the `+ ''` either.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -230,6 +230,9 @@ sub check_and_stash_category : Private {
...
+    $c->{stash}->{body_names_to_list} = \@body_names_to_list;
```
dracos: The ajax call that returns the same data calls list `list_of_names`, might be good to match that here? (Not that it's a great name.)

dracos:
Sorry, while testing #2179 I've come across another bug which I think is here, not there, to do with always visible layers (and is why the JS test is failing there, good catch by it!). Once the Bucks always_visible layer is being loaded, it starts off visible by default, and nothing changes that until a category change calling update_layer_visibility. But this means clicking the map anywhere else in the country gives a "no result" output because they're being checked against those roads and failing, and so hiding the report form.

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,12 +16,61 @@ var fixmystreet = fixmystreet || {};
     };
 })();
 
+OpenLayers.Layer.VectorAsset = OpenLayers.Class(OpenLayers.Layer.Vector, {
+    initialize: function(name, options) {
+        OpenLayers.Layer.Vector.prototype.initialize.apply(this, arguments);
+        // But also want to do it immediately in case it's hiding the form or something
+        $(fixmystreet).on('report_new:category_change', this.update_layer_visibility.bind(this));
+        $(window).on('load', this.update_layer_visibility.bind(this));
```
dracos: Could you call this in the `init_asset_layer` function, keep that all together?

dracos: Just the load bit, I was meaning as init_asset_layer is already being set up to run in the same event.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -695,12 +700,15 @@ sub setup_categories_and_bodies : Private {
         push @category_options, $seen{_('Other')} if $seen{_('Other')};
     }
 
+    @body_names_to_list = map { $_->name } values %bodies_to_list;
```
dracos: This presumably needs to be renamed to list_of_names and dealt with the same way as Around.pm, for when you've visited /report/new directly

web/cobrands/fixmystreet/assets.js
```diff
@@ -631,7 +656,12 @@ fixmystreet.assets = {
             fixmystreet.map.addControl(fixmystreet.assets.controls[i]);
             fixmystreet.assets.controls[i].activate();
         }
-    }
+    },
+
+    check_layer_relevant: function(layer, category) {
+      return OpenLayers.Util.indexOf(layer.asset_category, category) != -1 &&
+        ( !fixmystreet.bodies.length || !layer.body || OpenLayers.Util.indexOf(fixmystreet.bodies, layer.body ) != -1 );
+    },
```
dracos: Hmm, bit annoying if our automated JS linter stops you doing an okay loop thing, but misses this, not sure what happened there.

web/cobrands/fixmystreet/assets.js
```diff
@@ -116,6 +167,10 @@ function init_asset_layer(layer, pins_layer) {
         fixmystreet.map.events.register( 'zoomend', layer, check_zoom_message_visibility);
     }
 
+    if (layer.CLASS_NAME == 'OpenLayers.Layer.VectorAsset' || layer.CLASS_NAME == 'OpenLayers.Layer.VectorNearest') {
```
dracos: This if will always be true, only layers of those two classes are created by this file.

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: This presumably needs to be renamed to list_of_names and dealt with the same way as Around.pm, for when you've visited /report/new directly

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -281,12 +282,15 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
     my $councils_text_private = $c->render_fragment( 'report/new/councils_text_private.html');
 
     $unresponsive = $c->stash->{unresponsive}->{$category} || $c->stash->{unresponsive}->{ALL} || '';
+
+    my @bodies = map { $_->name } @$bodies;
```
dracos: And here also?

web/cobrands/fixmystreet/assets.js
dracos: This if will always be true, only layers of those two classes are created by this file.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -221,6 +221,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
 
     my $body = encode_json(
         {
+            bodies          => \@bodies,
```
dracos: And here is still returning IDs, not names.

web/cobrands/fixmystreet/assets.js
dracos: And this must happen first, before addLayer, or all the layers trigger HTTP requests before it gets here.

web/cobrands/fixmystreet/assets.js
```diff
@@ -16,12 +16,60 @@ var fixmystreet = fixmystreet || {};
...
+                }
+                this.zoom_to_assets();
+            } else {
+                this.setVisibility(false);
+                if (this.fixmystreet.fault_layer) {
+                    this.fixmystreet.fault_layer.setVisibility(false);
+                }
+            }
+        } else {
+            if (fixmystreet.bodies.length && this.fixmystreet.body) {
```
dracos: I think we should remove the length requirement - dev is much harder when most areas are returning nothing but then all the national layers are loading everywhere. I don't think it should affect anything if the layers only show if we know the body is relevant.

web/cobrands/fixmystreet/assets.js
```diff
@@ -631,6 +656,11 @@ fixmystreet.assets = {
             fixmystreet.map.addControl(fixmystreet.assets.controls[i]);
             fixmystreet.assets.controls[i].activate();
         }
+    },
+
+    check_layer_relevant: function(layer, category) {
+      return OpenLayers.Util.indexOf(layer.asset_category, category) != -1 &&
+        ( !fixmystreet.bodies.length || !layer.body || OpenLayers.Util.indexOf(fixmystreet.bodies, layer.body) != -1 );
```
dracos: length requirement here too.

perllib/FixMyStreet/App/Controller/Report/New.pm
struan: This is only ever used in a json object so I don't think it needs the CSV handling because were not splitting it.

dracos: Good point!

PR #2179, opened 2018-06-29
WIP Include asset files on national site

struan:
Enable asset files to be included on the national site by adding the body id to a `COBRAND_ASSETS` setting.

Body specific asset js is included in `web/$cobrand/assets.js` and should handle all the layer related code. Other cobrand specific code can remain in `js.js`. This also removes cobrand specific openlayers code in favours of a single WFS enabled openlayers build.

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1086

Code to cope with layers that display a "not a council road" style message on the national site. This also adds a do_not_send parameter to enable the front end to specify bodies reports should not be sent to because of rules in the asset JS.


templates/web/base/common_scripts.html
```diff
@@ -57,7 +57,12 @@
         version('/cobrands/fixmystreet/map.js'),
         version('/vendor/dropzone.min.js'),
         version('/vendor/fancybox/jquery.fancybox-1.3.4.pack.js'),
+        version('/cobrands/fixmystreet/assets.js'),
```
dracos: Do we always want to load this, not all cobrands use assets?

conf/general.yml-example
```diff
@@ -168,6 +168,12 @@ ALLOWED_COBRANDS:
   - cobrand2: 'hostname_substring2'
   - cobrand3
 
+# Control the loading of cobrand specific map asset layers.
+# If a cobrand is named here then `/web/cobrands/$cobrand/assets.js` will be
+# included in the JavaScript. You probably want to leave the blank.
+COBRAND_ASSETS:
```
dracos: What's the preference for this over listing them in a `templates/web/fixmystreet.com/` specific template?

web/cobrands/bathnes/assets.js
```diff
@@ -0,0 +1,242 @@
...
+    strategy_class: OpenLayers.Strategy.FixMyStreet,
+    asset_category: "",
+    asset_item: "asset",
+    asset_type: 'spot',
+    max_resolution: 2.388657133579254,
+    min_resolution: 0.5971642833948135,
+    asset_id_field: 'feature_no',
+    attributes: null,
+    geometryName: 'msGeometry',
+    body: 2551,
```
dracos: See comment on #2178.

perllib/FixMyStreet/Map/FMS.pm
```diff
@@ -12,7 +12,7 @@ use strict;
 sub map_template { 'fms' }
 
 sub map_javascript { [
-    '/vendor/OpenLayers/OpenLayers.fixmystreet.js',
+    '/vendor/OpenLayers/OpenLayers.wfs.js',
```
dracos: Not sure there's any way round this, but this is a 40% increase in gzipped filesize :-/

perllib/FixMyStreet/Map/BathNES.pm
```diff
@@ -7,12 +7,13 @@ use base 'FixMyStreet::Map::OSM';
 use strict;
 
 sub map_javascript { [
-    '/vendor/OpenLayers/OpenLayers.bathnes.js',
+    '/vendor/OpenLayers/OpenLayers.wfs.js',
```
dracos: This loses `OpenLayers/Protocol/WFS/v2_0_0.js` - is this needed?

conf/general.yml-example
struan: None really, other than it seemed like a config thing so it should go in the config file. I guess there's an argument for putting it in a template that's updated as part of the adding the cobrand.

perllib/FixMyStreet/Map/BathNES.pm
struan: oops, yes it is.

perllib/FixMyStreet/Map/FMS.pm
struan: well, some faff means this is reduced to a 30 ish % increase.

conf/general.yml-example
dracos: I think I prefer a template, given it's not exactly something being used by other installations at present.

perllib/FixMyStreet/Map/FMS.pm
dracos: Ah, sadly FramedCloud is used by the fault assets stuff.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -954,7 +955,9 @@ sub process_report : Private {
                 my $body = $c->model('DB::Body')->search({ name => $single_body_only })->first;
                 $body ? $body->id : '-1';
             } else {
-                my $bodies = $c->forward('contacts_to_bodies', [ $report->category ]);
+                my $contact_options = {};
+                $contact_options->{do_not_send} = [ split(',', $params{do_not_send}) ] if $params{do_not_send};
```
dracos: Needs to be CSV handler.

web/cobrands/fixmystreet.com/national.js
```diff
@@ -0,0 +1,40 @@
...
+
+fixmystreet.enable_report_form = function() {
+    $("#problem_form").children().each( function (i, c) {
+        if (c.id != 'js-layer-message') {
+            $(c).show();
+        }
+    } );
+};
+
+$(fixmystreet).on('assets:check_do_not_send', fixmystreet.check_do_not_send.check);
```
dracos: This doesn't feel like a good event to me, as you're saying "do this thing when I tell you to do this thing", rather than "this thing has happened, listen for this thing happening".

web/cobrands/buckinghamshire/assets.js
```diff
@@ -0,0 +1,208 @@
...
+                disable_report_form();
+            }
+        },
+
+        not_found: function(layer) {
+            // If a feature wasn't found at the location they've clicked, it's
+            // probably a field or something. Show an error to that effect,
+            // unless an asset is selected.
+            if (fixmystreet.do_not_send) {
+                fixmystreet.do_not_send[layer.fixmystreet.body] = 1;
```
dracos: Presumably `found()` needs to remove this if present.

web/cobrands/fixmystreet.com/national.js
```diff
@@ -0,0 +1,40 @@
...
+    check: function() {
+        var do_not_send = [];
+        if (fixmystreet.do_not_send && fixmystreet.bodies) {
+            for ( var i = 0; i < fixmystreet.bodies.length; i++ ) {
+                if ( fixmystreet.do_not_send['' + fixmystreet.bodies[i]] == 1 ) {
+                    do_not_send.push(fixmystreet.bodies[i]);
+                }
+            }
+        }
+        $('#do_not_send').val(do_not_send.join(','));
```
dracos: Needs to be CSV

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1013,10 +1015,19 @@ sub process_report : Private {
...
+    my ($self, $c, $category, $options) = @_;
 
     my @contacts = grep { $_->category eq $category } @{$c->stash->{contacts}};
 
+    # check that the front end has not indicated that we should not send to a
+    # body. This is usually because the asset code thinks it's not near enough
+    # to a road.
+    if ($options->{do_not_send}) {
+        my %do_not_send_check = map { $_ => 1 } @{$options->{do_not_send}};
+        my @contacts_filtered = grep { !$do_not_send_check{$_->body->id} } @contacts;
```
struan: This should be checking `$_->body->name` here?

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1013,10 +1015,19 @@ sub process_report : Private {
...
+        @contacts = @contacts_filtered if scalar @contacts_filtered;
+    }
```
struan: if we've filtered out all the contacts why would want to leave the list of contacts as it was before?

t/app/controller/report_new.t
```diff
@@ -967,6 +967,13 @@ foreach my $test (
         extra_fields => { single_body_only => 'Invalid council' },
         email_count => 1,
     },
+    {
+        desc => "test do_not_send means body is ignored",
+        category => 'Street lighting',
+        councils => [ 2326 ],
+        extra_fields => { do_not_send => $body_ids{2226} },
```
struan: Should this not be passing in body names now?

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: You wrote this bit ;) I guess didn't want a report to not get sent anywhere that hadn't told the user this?

struan: That was weeks ago, I've changed ;)

I see the point, yes. In theory if it's not going to be sent at all the front end shouldn't let them submit and if it has then it's probably a sign of something up in which case sending things is probably a better option.

PR #2186, opened 2018-07-09
WIP Show not road message on national site

struan:
Code to cope with layers that display a "not a council road" style message on the national site. This also adds a `do_not_send` parameter to enable the front end to specify bodies reports should not be sent to because of rules in the asset JS.


dracos:
Included in #2179 now.

PR #2187, opened 2018-07-09
Some small fixes

dracos:
* Use DNR reply for sent report email
* Stop duplicate category listing on /my
* Hide password help field to match others
* Show all reports by default on /my


dracos:
Pushed as ef5182539 with changelogs.

PR #2192, opened 2018-07-16
[WIP] Ground Control pothole changes

davea:
For mysociety/fixmystreet-commercial#997
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1312,6 +1312,8 @@ sub users: Path('users') : Args(0) {
     } else {
         $c->forward('/auth/get_csrf_token');
         $c->forward('fetch_all_bodies');
+        $c->cobrand->call_hook('admin_user_edit_extra_data');
+
 
         # Admin users by default
         my $users = $c->cobrand->users->search(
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -2108,6 +2109,9 @@ sub fetch_all_bodies : Private {
     my ($self, $c ) = @_;
 
     my @bodies = $c->model('DB::Body')->translated->all_sorted;
+    if ( my @cobrand_bodies = $c->cobrand->call_hook('admin_fetch_all_bodies') ) {
```
dracos: This needs `@bodies` passing to it, looks like, and then the `zurich` `if` below can presumably be removed.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1473,6 +1473,7 @@ sub user_edit : Path('user_edit') : Args(1) {
 
     $c->forward('fetch_all_bodies');
     $c->forward('fetch_body_areas', [ $user->from_body ]) if $user->from_body;
+    $c->cobrand->call_hook('admin_user_edit_extra_data');
```
dracos: Does this also need to be in the user add controller, so it's on the new user form?

Issue #666, opened 2014-02-19
throw up a modal dialogue inviting alert subscription on a newly created problem report


davewhiteland:
related to #665 -- so sometimes there's no confirmation page (e.g. you're logged in already) -- be bold and throw a modal up on page load when a page of a just-created-by-previous-page problem report is displayed.

If so, this presumably would be instead of #665 because that would be a bit much.

Note this is quite "aggressive" behaviour but let's see how it goes.

dracos:
There is a confirmation page always now.

PR #2194, opened 2018-07-20
Avoid input placeholders

### LINKED ISSUE #2180: Consider ditching placeholders for a more accessible approach
For discussion, probably

https://www.smashingmagazine.com/2018/06/placeholder-attribute/

zarino:
Fixes #2180.

Now most of the places that we had placeholder text inside of inputs, it’s been removed – sometimes replaced with help text, but most often, just dropped entirely, because the existing label was descriptive enough.

### DIFF
--- a/templates/web/base/alert/index.html
+++ b/templates/web/base/alert/index.html
@@ -30,8 +30,10 @@ within a certain distance of a particular location.', "%s is the site name"), si
 </p>
 <form method="get" action="/alert/list" class="form-box js-geolocate">
   <fieldset>
+    <label for="pc">[% loc('Postcode or street name and area') %]</label>
+    <p class="form-hint" id="pc-hint">[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]</p>
     <div class="form-txt-submit-box">
-      <input class="form-control" type="text" name="pc" value="[% pc | html %]" placeholder="[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]">
+      <input class="form-control" type="text" name="pc" value="[% pc | html %]" aria-describedby="pc-hint">
       <input class="green-btn" type="submit" value="[% loc('Go') %]">
     </div>
   </fieldset>
--- a/templates/web/base/around/postcode_form.html
+++ b/templates/web/base/around/postcode_form.html
@@ -10,8 +10,9 @@
 
         <form action="[% c.uri_for('/around') %]" method="get" name="postcodeForm" id="postcodeForm" class="js-geolocate">
             <label for="pc">[% question %]:</label>
+            <p class="form-hint" id="pc-hint">[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]</p>
             <div>
-                <input type="text" name="pc" value="[% pc | html %]" id="pc" size="10" maxlength="200" placeholder="[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]" required>
+                <input type="text" name="pc" value="[% pc | html %]" id="pc" size="10" maxlength="200" required aria-describedby="pc-hint">
                 <input type="submit" value="[% loc('Go') %]" id="sub">
             </div>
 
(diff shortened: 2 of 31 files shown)


templates/web/base/report/display_tools.html
```diff
@@ -43,10 +43,10 @@
         </a>
         [% loc('Receive email when updates are left on this problem.' ) %]</p>
         <fieldset>
-        <label class="hidden n" for="alert_rznvy">[% loc('Your email') %]</label>
+        <label for="alert_rznvy">[% loc('Your email') %]</label>
```
dracos: Giant gap above this now.

templates/web/base/alert/index.html
```diff
@@ -30,8 +30,10 @@ <h1>[% loc('Local RSS feeds and email alerts') %]</h1>
 </p>
 <form method="get" action="/alert/list" class="form-box js-geolocate">
   <fieldset>
+    <label for="pc">[% loc('Postcode or street name and area') %]</label>
+    <p class="form-hint" id="pc-hint">[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]</p>
```
dracos: Could form-hint be a bit grey?

templates/web/base/report/new/form_report.html
```diff
@@ -49,12 +49,13 @@ <h2 class="form-section-heading">[% loc( 'Public details' ) %]</h2>
 [% TRY %][% PROCESS 'report/new/after_photo.html' %][% CATCH file %][% END %]
 
     [% DEFAULT form_detail_label = loc('Explain what’s wrong') %]
+    [% DEFAULT form_detail_placeholder = loc('e.g. ‘This pothole has been here for two months and…’') %]
     <label for="form_detail">[% form_detail_label %]</label>
+    <p class="form-hint" id="detail-hint">[% form_detail_placeholder %]</p>
```
dracos: Also spacing feels a bit wide around the new examples, as now it's equal between label, examples, text input and next label. Feels like should be bit smaller in the same-thing bits.

templates/web/base/around/postcode_form.html
```diff
@@ -10,8 +10,9 @@
 
         <form action="[% c.uri_for('/around') %]" method="get" name="postcodeForm" id="postcodeForm" class="js-geolocate">
             <label for="pc">[% question %]:</label>
+            <p class="form-hint" id="pc-hint">[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]</p>
```
dracos: I wonder if this might be better under the form? Perhaps to the left, with locate automatically to the right? Feels like part of heading where it is now, but might just be used to it.

templates/web/base/alert/index.html
dracos: Stop that form being grey :)

web/cobrands/sass/_base.scss
```diff
@@ -1775,8 +1786,15 @@ label .muted {
...
 
-#problem_form > label:first-child {
-  margin-top: 0;
+#problem_form {
+  & > label:first-child {
+    margin-top: 0;
+  }
+
+  // Reduce space between labels and their form-hints
+  label + .form-hint {
```
dracos: This is nice, should it be more general? e.g. for the one on `/alert`, or if you added similar to the update form, an admin form, or wherever.

zarino: Good idea. Looks like we already have something similar (`margin-top: -0.25em`)on `#front-main #postcodeForm label + .form-hint` so I see whether I can get all these standardised.

templates/web/base/alert/index.html
zarino: `.form-hint`s are now `#666` grey!

screen shot 2018-08-01 at 10 24 32

templates/web/base/around/postcode_form.html
zarino: I think we leave it between the label and the input. That’s what we do everywhere else.

templates/web/base/report/new/form_report.html
zarino: Fixed!

screen shot 2018-08-01 at 10 24 32

Issue #2012, opened 2018-03-02
When viewing a report, make it easier to create a new report nearby


zarino:
Spun out from https://github.com/mysociety/fixmystreet/issues/1462#issuecomment-364899132.

Users trying to report an issue in a congested area will often accidentally click an existing pin, because there isn’t enough empty space _around_ the existing pins.

Right now, such users would have to go back to the main reports view (clicking the "&lt; Back to all reports" link in the top left, or using their browser’s Back button) and then re-attempt to click on the empty map background, to report their problem, avoiding any nearby pins.

There’s a chance that, if we gave such users a quicker way to create a new report _near_ the one they are currently (accidentally) viewing, it might save them going through the tricky "click the map" process a second time in that congested area, or just giving up in exasperation.

**"Report a new problem near here"** and **"Report a different problem at this location"** have been suggested as a clear wordings for a call-to-action on this.

The CTA would need to be **very prominent** in order to catch the user’s attention as soon as possible after they’ve realised that they’re looking at an existing report, and certainly before they automatically go back to re-attempt the click on the map.

My initial reaction was to put it in the sidebar, either above or below the `.problem-header` (which houses the report’s title, photo, and description). But I worry that _still_ wouldn’t be obvious enough. There’s a lot to parse in that sidebar.

![screenshot-2018-3-2 viewing a location](https://user-images.githubusercontent.com/739624/36898579-5fb7a460-1e13-11e8-86bb-c03c35d82293.png)

Perhaps it might work in the sidebar, if we reduced competing content? For example, we could hide all of the "Provide an update" stuff by default, and instead, present two fairly prominent buttons below the problem description (but above any comments that have been left) saying – "Report a new problem near here" and "Provide an update on this problem" – clicking the latter would unhide the "Provide an update" form and the login fields. With the "Provide an update" form hidden, there’d be much less visual competition for attention, and people might spot the "Report a new problem near here" button more easily.

![screenshot-2018-3-2 viewing a location 1](https://user-images.githubusercontent.com/739624/36898587-658ee27c-1e13-11e8-8bb7-997b03a1a172.png)

Maybe a totally alternative position might be a pop-up balloon on the map itself, very near the pin you’ve just clicked? If the pop-up appears quick enough, the user’s eye will likely still be focussed on the (now enlarged) pin, and once they’ve darted across to the sidebar and noticed it’s not a "new report" form, they might remember where they saw a button that could help them out.

[...]

PR #2195, opened 2018-07-27
"New report near here" button in nav bar

### LINKED ISSUE #2016: Reduce confusion of 'click on map' banner

zarino:
Fixes #2016 and is a partway solution to #2012.

Rather than always saying "Report a problem" and always just linking back to the homepage postcode search form, the `.report-a-problem-btn` in the nav bar now behaves differently, based on what else is on the page.

The general idea is that, if you’re looking at an individual problem already, then the button will say "Report _another_ problem here" and will automatically begin creating a new report at the same coordinates. Otherwise, if you’re zoomed a sensible distance into a map, then the button will say "Report a problem _here_" and will begin a report at the centre of the visible map. And finally, if neither of those cases are true, it’ll just do what it always did, and take you to the homepage.

When clicking the button, if it’s possible for the current page to be seamlessly transformed into the reporting form, then we do that. Otherwise, we fall back to a server-side page load.

# /around page, zoomed far out

screenshot_2018-08-02 viewing a location

# /around page, zoomed closer

screenshot_2018-08-02 viewing a location 1

# /report/<id> page

screenshot_2018-08-02 full litter bins - viewing a problem

# /reports/<area> page, zoomed fairly close

screenshot_2018-08-02 borsetshire county council - summary reports

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -671,6 +671,27 @@ $.extend(fixmystreet.set_up, {
     });
   },
 
+  report_a_problem_btn: function() {
+    $(fixmystreet).on('maps:update_view', fixmystreet.update_report_a_problem_btn);
+
+    // Hide button on new report page.
+    if ( fixmystreet.page === 'new' ) {
+      $('.report-a-problem-btn').hide();
+    }
+
+    $('.report-a-problem-btn').on('click', function(e){
+      var url = this.href;
+      if ( url.indexOf('report/new') > -1 ) {
+        try {
+          e.preventDefault();
+          fixmystreet.display.begin_report( fixmystreet.map.getCenter() );
+        } catch (error) {
+          window.location = url;
+        }
+      }
+    });
+  },
+
   map_controls: function() {
     //add permalink on desktop, force hide on mobile
     //add links container (if its not there)
@@ -891,6 +912,34 @@ $.extend(fixmystreet.set_up, {
 
 });
 
+fixmystreet.update_report_a_problem_btn = function() {
+    var zoom = fixmystreet.map.getZoom();
+    var center = fixmystreet.map.getCenterWGS84();
+    var new_report_url = '/report/new?longitude=' + center.lon.toFixed(6) + '&latitude=' + center.lat.toFixed(6);
+
+    var href = '/';
+    var visible = true;
+    var text = translation_strings.report_a_problem_btn.default;
+
+    if (fixmystreet.page === 'new') {
+        visible = false;
+
+    } else if (fixmystreet.page === 'report') {
+        text = translation_strings.report_a_problem_btn.another;
+        href = new_report_url;
+
+    } else if (fixmystreet.page === 'around' && zoom > 1) {
+        text = translation_strings.report_a_problem_btn.here;
+        href = new_report_url;
+
+    } else if (fixmystreet.page === 'reports' && zoom > 12) {
+        text = translation_strings.report_a_problem_btn.here;
+        href = new_report_url;
+    }
+
+    $('.report-a-problem-btn').attr('href', href).text(text).toggle(visible);
+};
+
 fixmystreet.update_councils_text = function(data) {
     $('#js-councils_text').html(data.councils_text);
     $('#js-councils_text_private').html(data.councils_text_private);
@@ -1084,6 +1133,8 @@ fixmystreet.display = {
     }
 
     fixmystreet.page = 'new';
+
+    fixmystreet.update_report_a_problem_btn();
   },
 
   report: function(reportPageUrl, reportId, callback) {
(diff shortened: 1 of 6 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -667,6 +667,27 @@ $.extend(fixmystreet.set_up, {
...
+  report_a_problem_btn: function() {
+    $(fixmystreet).on('maps:update_view', fixmystreet.update_report_a_problem_btn);
+
+    // Hide button on new report page.
+    if ( fixmystreet.page === 'new' ) {
+      $('.report-a-problem-btn').toggle(false);
+    }
+
+    $('.report-a-problem-btn').on('click', function(e){
+      var url = $(this).attr('href');
```
dracos: Or... `this.href` :)

web/js/map-OpenLayers.js
```diff
@@ -749,6 +749,16 @@ $.extend(fixmystreet.utils, {
             fixmystreet.map.setCenter(centre, fixmystreet.zoom || 3);
         }
 
+        // map.getCenter() returns a position in "map units", but sometimes you
+        // want the center in GPS-style latitude/longitude coordinates (WGS84)
+        // for example, to pass as GET params to fixmystreet.com/report/new.
+        fixmystreet.map.getCenterWGS84 = function() {
+            return fixmystreet.map.getCenter().transform(
```
dracos: Could this cut decimals to 6dp, pointless having more.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -887,6 +908,33 @@ $.extend(fixmystreet.set_up, {
...
+    var href = '/';
+    var visible = true;
+    var text = translation_strings.report_a_problem_btn.default;
+
+    if (fixmystreet.page === 'new') {
+        visible = false;
+
+    } else if (fixmystreet.page === 'report') {
+        text = translation_strings.report_a_problem_btn.another;
+        href = '/report/new?longitude=' + center.lon + '&latitude=' + center.lat;
```
dracos: You do this href three times, so perhaps set it once somewhere first?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -667,6 +667,27 @@ $.extend(fixmystreet.set_up, {
     });
   },
 
+  report_a_problem_btn: function() {
+    $(fixmystreet).on('maps:update_view', fixmystreet.update_report_a_problem_btn);
+
+    // Hide button on new report page.
+    if ( fixmystreet.page === 'new' ) {
+      $('.report-a-problem-btn').toggle(false);
```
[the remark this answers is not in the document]
zarino: Oops! In a previous version of the code, this line was a lot closer to the later:

```
$('.report-a-problem-btn').attr('href', href).text(text).toggle(visible);
```

So it made a bit more sense to use the same `.toggle()` method. But yeah, now they’re so far apart, `.hide()` is probably a better fit here.

web/js/map-OpenLayers.js
zarino: We use `.transform()` a few times in `map-OpenLayers.js`, but I don’t think we actively round to 6 decimal places in any of those situations – why start now?

dracos: I was thinking about it only recently, in fact; those other places should presumably do it too. Just because they don't yet doesn't mean this one shouldn't, though :)

zarino: Yes, comment poorly worded on my part! I meant more "why change it here and not everywhere else" ;-)

If it’s a think we’re going to keep doing everywhere, do you think we should just create our own wrapper around `.transform()` that does the rounding, and use that everywhere instead?

dracos: There's only one other obvious use I can see that would benefit from it, in `update_pin` (and the admin drag callback is already doing it), so probably not worth it.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -916,20 +916,24 @@ fixmystreet.update_report_a_problem_btn = function() {
     var visible = true;
     var text = translation_strings.report_a_problem_btn.default;
 
+    var new_report_url = function(lonlat) {
```
dracos: Why isn't this `var new_report_url = '/report/new?longitude=' + center.lon + '&latitude=' + center.lat;`? :)

PR #2198, opened 2018-08-02
"Report another problem here" button on confirmation page

zarino:
Part of #2012.

Should make it much quicker to report multiple problems in the same location, one after the other.

screenshot_2018-08-02 confirmation

Uses the same new `loc('Report another problem here')` string that was introduced in #2195.

### DIFF
--- a/templates/web/base/tokens/confirm_problem.html
+++ b/templates/web/base/tokens/confirm_problem.html
@@ -31,6 +31,12 @@
 
 [% TRY %][% INCLUDE 'tokens/_extras_confirm.html' %][% CATCH file %][% END %]
 
+  <p class="confirmation-again">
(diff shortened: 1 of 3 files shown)


templates/web/base/tokens/confirm_problem.html
```diff
@@ -31,6 +31,12 @@ <h2>[% loc('Thank you for reporting this issue!') %]</h2>
 
 [% TRY %][% INCLUDE 'tokens/_extras_confirm.html' %][% CATCH file %][% END %]
 
+  <p class="confirmation-again">
+    <a href="/report/new?latitude=[% report.latitude %]&longitude=[% report.longitude %]" class="btn btn-primary">
```
dracos: Needs to be `&amp;`.

Issue #2202, opened 2018-08-08
Can't fetch submodules


chellem:
Hello,

I can't clone the submodules, it seems that git.mysociety.org is down.

Is there any possibility of resolving that issue?

dracos:
Hi, thanks for your message. I've just done a fresh clone and it worked fine, do you have any more details than "is down"?
```
§ git clone https://github.com/mysociety/fixmystreet
Cloning into 'fixmystreet'...
remote: Counting objects: 106910, done.
remote: Compressing objects: 100% (222/222), done.
remote: Total 106910 (delta 340), reused 384 (delta 277), pack-reused 106411
Receiving objects: 100% (106910/106910), 120.82 MiB | 2.66 MiB/s, done.
Resolving deltas: 100% (74373/74373), done.)   
§ cd fixmystreet/
§ git submodule update --init
Submodule 'commonlib' (git://git.mysociety.org/commonlib) registered for path 'commonlib'
Cloning into '/Users/matthew/temp/fixmystreet/commonlib'...
Submodule path 'commonlib': checked out '20f0ad17b4d0d1a0772690ed5fbfea1581a909cc'
§ 
```

chellem:
Well, from my workplace, am behind a proxy, I thought that was the problem but it's not the case.  I can't clone from git.mysociety.org/commonlib.  Host not found and I think there's a ssl problem on it, right?

dracos:
What output do you get if you do a `git clone git://git.mysociety.org/commonlib` from your command line? As that works fine for me, from a server with no special permissions/access. "Host not found" sounds like an issue with your setup I'm afraid, rather than anything at our end, the host definitely does exist. SSL isn't involved (though as far as I know our setup is fine, our git isn't available over HTTPS if that's what you tried). What does `dig git.mysociety.org` show? We haven't had any other reports. The reason we don't use github for the submodule is to prevent a potential external dependency when deploying but perhaps we need to look at that again.

dracos:
@chellem Did you try any of my debugging suggestions, or have any luck in tracking it down?

chellem:
Output of dig:

`
 ; <<>> DiG 9.10.3-P4-Debian <<>> git.mysociety.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45240
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

[...]

PR #2203, opened 2018-08-08
Update .gitmodules

chellem:
#2202 

[skip changelog]


dracos:
Thanks very much for this, but for the reason given in #2202 (not wanting an external dependency when deploying) I think we want to leave this as is. Happy to look at what could be causing the issue with your setup, or of course you can change this in your fork if it really can't be sorted out.

PR #2204, opened 2018-08-09
Few small breakages to fix

dracos:
* New OS map doesn't work at zoom level 11 any more.
* 27700 library not loaded for Oxfordshire, so inspectors not getting easting/northing when pin dragged.
* Set up fully whether pins are draggable or not, which should fix inspector issue dragging/unable to click pins that shouldn't be draggable. Theoetically this means the DragFeatureFMS control could be loaded at all times, and it shouldn't affect any non-draggable pins, but I have not done that here.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -783,7 +783,8 @@ sub ajax : Private {
 
     my @pins = map {
         my $p = $_;
-        [ $p->{latitude}, $p->{longitude}, $p->{colour}, $p->{id}, $p->{title} ]
+        # lat, lon, 'colour', ID, title, type/size, draggable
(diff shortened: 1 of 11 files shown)


perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -783,7 +783,7 @@ sub ajax : Private {
 
     my @pins = map {
         my $p = $_;
-        [ $p->{latitude}, $p->{longitude}, $p->{colour}, $p->{id}, $p->{title} ]
+        [ $p->{latitude}, $p->{longitude}, $p->{colour}, $p->{id}, $p->{title}, '', JSON->false ]
     } @{$c->stash->{pins}};
```
struan: I'd maybe add a comment here to say last entry is pin dragging?

PR #2212, opened 2018-08-17
Make it easier to prevent a form_detail_placeholder being printed

zarino:
Previously, if you didn’t want a hint to appear under the main "details" textarea on the new report form, you had to override the entire `form_report.html` template, or leave the hint element in the markup but hide it with CSS.

Now, you can set `form_detail_placeholder` to a falsey value, and the template won’t output the hint element at all. It also amends the `aria-describedby` attribute on the textarea so it doesn’t end up referencing a hint element that doesn’t exist.

I needed this for https://github.com/mysociety/collideoscope/pull/28, because the existing `[% DEFAULT form_detail_placeholder = … %]` block gave me no easy way to prevent the hint being shown.

# How it looks:

When no custom `form_detail_placeholder` is defined in the cobrand, eg:

```
# /templates/web/<cobrand>/report/new/_form_labels.html
[% form_detail_label = loc('Can you describe what happened?') %]
```

undefined

When a custom `form_detail_placeholder` is defined in the cobrand, eg:

```
# /templates/web/<cobrand>/report/new/_form_labels.html
[% form_detail_placeholder = "Helpful cobrand-specific hint here" %]
[% form_detail_label = loc('Can you describe what happened?') %]
```

truthy

**NEW:** When `form_detail_placeholder` is set to a falsey value, eg:

```
# /templates/web/<cobrand>/report/new/_form_labels.html
[% form_detail_placeholder = 0 %]
[% form_detail_label = loc('Can you describe what happened?') %]
```

falsey

### DIFF
--- a/templates/web/base/report/new/form_report.html
+++ b/templates/web/base/report/new/form_report.html
@@ -49,13 +49,22 @@
 [% TRY %][% PROCESS 'report/new/after_photo.html' %][% CATCH file %][% END %]
 
     [% DEFAULT form_detail_label = loc('Explain what’s wrong') %]
-    [% DEFAULT form_detail_placeholder = loc('e.g. ‘This pothole has been here for two months and…’') %]
     <label for="form_detail">[% form_detail_label %]</label>
+
+[%# You can prevent a hint being printed by setting form_detail_placeholder to a Falsey value %]
+[% IF NOT form_detail_placeholder.defined %]
+    [% SET form_detail_placeholder = loc('e.g. ‘This pothole has been here for two months and…’') %]
+[% END %]
+
+[% IF form_detail_placeholder %]
     <p class="form-hint" id="detail-hint">[% form_detail_placeholder %]</p>
+[% END %]
+
 [% IF field_errors.detail %]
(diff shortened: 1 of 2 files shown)


templates/web/base/report/new/form_report.html
```diff
@@ -49,13 +49,22 @@ <h2 class="form-section-heading">[% loc( 'Public details' ) %]</h2>
 [% TRY %][% PROCESS 'report/new/after_photo.html' %][% CATCH file %][% END %]
 
     [% DEFAULT form_detail_label = loc('Explain what’s wrong') %]
-    [% DEFAULT form_detail_placeholder = loc('e.g. ‘This pothole has been here for two months and…’') %]
```
dracos: What you have works fine, but I wonder if it would be clearer if this line became `[% IF NOT  form_detail_placeholder.defined %][% SET form_detail_placeholder = ... %][% END %]` and then below would be `[% IF form_detail_placeholder %]...` without needing to duplicate e.g. the aria_describedby line. What do you think? Not much in it.

PR #2214, opened 2018-08-20
Add cobrand hook for disallowing updates on problems

davea:
For mysociety/collideoscope#20.

### DIFF
--- a/templates/web/base/report/display.html
+++ b/templates/web/base/report/display.html
@@ -61,14 +61,18 @@
               c.uri_for( '/report/new', { longitude => longitude, latitude => latitude } )
           ) %]
     </p>
-[% ELSIF NOT shown_form %]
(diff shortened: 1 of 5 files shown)


templates/web/base/report/display.html
```diff
@@ -61,7 +61,7 @@
               c.uri_for( '/report/new', { longitude => longitude, latitude => latitude } )
           ) %]
     </p>
-[% ELSIF NOT shown_form %]
+[% ELSIF NOT shown_form AND NOT c.cobrand.call_hook('updates_disallowed', c, problem) %]
```
dracos: No need to pass in `c` as mentioned elsewhere.
I wonder (untested) if `c.cobrand.updates_disallowed(problem)` works? As template just fails silently if methods don't exist, unlike code.

PR #2216, opened 2018-08-22
 Allow cobrand to add extra ability to moderate

dracos:
This does some tidying of the moderation code, and adds a hook so that a cobrand can e.g. let users edit their own reports on the site.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Moderate.pm
+++ b/perllib/FixMyStreet/App/Controller/Moderate.pm
@@ -42,6 +42,7 @@ sub moderate : Chained('/') : PathPart('moderate') : CaptureArgs(0) { }
 sub report : Chained('moderate') : PathPart('report') : CaptureArgs(1) {
     my ($self, $c, $id) = @_;
     my $problem = $c->model('DB::Problem')->find($id);
+    $c->detach unless $problem;
 
     my $cobrand_base = $c->cobrand->base_url_for_report( $problem );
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -331,9 +331,23 @@ sub split_name {
...
+        $ids = $object->bodies_str_ids;
+    }
 
-    return 1 if $self->has_permission_to(moderate => $problem->bodies_str_ids);
+    my $staff_perm = exists($perms{staff}) ? $perms{staff} : $self->has_permission_to(moderate => $ids);
+    return 1 if $staff_perm;
+
+    # See if the cobrand wants to allow it in some circumstance
+    my $cobrand = $self->result_source->schema->cobrand;
+    return $cobrand->call_hook('moderate_permission', $self, $type => $problem);
```
struan: `$object` at the end surely?

perllib/FixMyStreet/App/Controller/Moderate.pm
```diff
@@ -234,9 +223,9 @@ sub moderate_update : Chained('update') : PathPart('') : Args(0) {
     $c->forward('update_moderate_hide');
 
     my @types = grep $_,
```
struan: I feel like there ought to be a check in here that we've found the comment given it falls back to moderating the problem now.

PR #2229, opened 2018-09-11
[Bromley] Update asset for category changes.

dracos:
[skip changelog]

This displays street lights and bins for too many things, but there's no way to filter spot assets apart from category at present.


web/cobrands/bromley/assets.js
```diff
@@ -67,6 +67,17 @@ var highways_stylemap = new OpenLayers.StyleMap({
     })
 });
 
+var bromley_to_tfl = {
```
struan: Just to be clear, these are Bromley categories where we want to send the report only to TfL? (assuming it's on a red route)

dracos: Correct.

PR #2230, opened 2018-09-11
[Bromley] Asset display fixes

dracos:
This moves `check_layer_relevant` from a global function to on the layer class, which means then it is easily subclassable to provide more nuanced behaviour. I think. Will test in the morning! [skip changelog]


struan:
yay! this is nice.

PR #2232, opened 2018-09-12
[Oxfordshire] display message if user selects PROW category

struan:
Adds some div wrappers round the non category bit of the form so we can
then hide easily.

[skip changelog]


templates/web/base/report/new/form_user.html
```diff
@@ -1,3 +1,4 @@
+<div id="private_form">
```
dracos: Wonder if better to put this wrap in the parent fill_in_details_form.html, less likely to be overridden?

struan: equally though, it's harder to override, which was sort of my reasoning.

PR #2237, opened 2018-09-14
Issues/collideoscope/42 report button mobile

zarino:
Fixes mysociety/collideoscope/issues/42.

On the homepage:

* Pressing "Report a problem" (in the nav menu on wide screens, or at the bottom of the page on small screens) now scrolls you to the search input and focusses it, so you can begin typing.
* Pressing the "Report" convenience button in the header on narrow screens does the same as above.

~~I also tried to refactor code in `front.js` a bit, since I’m guessing we want this file to be as short as possible, given it’s on the homepage, and we want that to get delivered as quickly as possible? (Hence why, for example, we don’t load jQuery on the homepage?)~~

### DIFF
--- a/web/js/front.js
+++ b/web/js/front.js
@@ -17,7 +17,7 @@ document.getElementById('pc').focus();
     }
 
     var lk = document.querySelector('span.report-a-problem-btn');
-    if (lk.addEventListener) {
+    if (lk && lk.addEventListener) {
         lk.addEventListener('click', function(){
             scrollTo(0,0);
         });
--- a/web/js/front.js
+++ b/web/js/front.js
@@ -10,6 +10,7 @@ document.getElementById('pc').focus();
         el.value = 1;
         form.insertBefore(el, form.firstChild);
     }
+
     var around_links = document.querySelectorAll('a[href*="around"]');
     for (i=0; i<around_links.length; i++) {
         var link = around_links[i];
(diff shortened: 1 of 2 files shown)


web/js/front.js
```diff
@@ -1,25 +1,39 @@
...
+        for (var i=0; i<elements.length; i++) {
+            callback(elements[i]);
+        }
+    };
+
+    var goToSearch = function(e){
+        if (e) { e.preventDefault(); }
+        var form = document.querySelectorAll('#postcodeForm')[0];
+        if (form) {
+            window.scroll(0, form.getBoundingClientRect().top + window.scrollY);
```
dracos: I don't understand what the scrollY is doing here. Would `0` be good enough, even on small screens?

web/js/front.js
```diff
@@ -1,25 +1,39 @@
...
-    for (i=0; i<around_links.length; i++) {
-        var link = around_links[i];
+    });
+
+    each('a[href*="around"]', function(link){
         link.href = link.href + (link.href.indexOf('?') > -1 ? '&js=1' : '?js=1');
-    }
+    });
+
+    each('span.report-a-problem-btn', function(el){
```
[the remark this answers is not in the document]
dracos: Note this only saves 1 byte.

web/js/front.js
```diff
@@ -1,25 +1,39 @@
...
+        el.addEventListener('click', goToSearch);
```
dracos: Need to check addEventListener exists somewhere.

zarino:
@dracos – here it is, without the big refactor. As you say, gzipping saves about as much as the refactoring ever would, and for now the code’s probably just about simple enough to stay as-is, even if it does repeat itself a bit ;-)

What do you think?

PR #2239, opened 2018-09-18
Clearer relocation options while you’re reporting a problem

### LINKED ISSUE #2238: Hard to search for a new location, now, on UK council cobrands
Since #2195 removed the hard-coded nar bar link to the homepage, it’s become difficult for users of council cobrands to search for new locations, once they get trapped on an "inner" page.

The issue is particularly pronounced for customer services staff, who spend a lot of time searching for new locations, to file new reports from telephone users. Now, with the homepage link removed from the nav bar, and the main header logo linking to _the council’s homepage_ rather than the FMS instance homepage, they have no easy way to search for a new location, other than manually editing the URL or navigating back in their browser history.

screen shot 2018-09-18 at 12 11 48

A quick fix would be to make it easier to return to the homepage. Perhaps giving council cobrands a "Home" nav link (in addition to the new "Report a problem [near here]" link). BUT we’d need to be careful about the wording of such a link, so as not to compete with the council’s _actual_ homepage.

Another solution would be to make it easier to start at a new location, from _inside_ the reporting flow.

Source: https://github.com/mysociety/fixmystreet-freshdesk/issues/13

zarino:
Now that the "Report a problem" link in the nav bar links, more often than not, to the reporting form rather than the homepage, it could be challenging to actually start a new report in a location *other* than the one currently on screen.

Rather than adding a link to the homepage, this change hopefully adds a link right at the moment the user expects it – right on the "Report a problem" form.

screen shot 2018-09-18 at 16 40 05

It also gave us an opportunity to reword the "Wrong location" message and give it an icon more suitable for high-dpi displays.

@LouiseMySociety @wrightmartin @dracos – thoughts?
Fixes #2238.


wrightmartin:
Is this removing the big green not-a-button?

dracos:
@wrightmartin No, this is when the form is being shown, after that. And it is a 'button' now (for a while now), in that clicking it does do something.

wrightmartin:
Damn, got my hopes up then. From a UX perspective this makes perfect sense, code looks fine too 👍 from me

zarino:
@dracos I‘ve checked the cobrands that that automatic checker flagged up – Yo Denuncio and Zurich are fine to remain unchanged because they both still have links to the homepage in the headers. The same was true for KIITC, but it seemed a particular shame not to include the new `.change_location` element there, given it’s such a new cobrand, so I’ve opened a PR to include it: https://github.com/mysociety/keepitinthecommunity/pull/93

Good to merge?

PR #2240, opened 2018-09-19
cypress tests for the filters on the around page

struan:
Plus a mock mapit handler for returning a geojson outline to make the
page load.

FIxes mysociety/fixmystreet-commercial#1186


davea:
These tests fail locally - do I need to configure something first?

struan:
No, this is part of the me not realising the fixture code generated randomish data so the tests are broken :(

PR #2243, opened 2018-09-20
Add configuration option for memcached host

sagepe:
Previously we assumed that any memcache instance would be running on the local loopback interface. This commit makes this configurable with a `FMS_MEMCACHED_HOST` option. If left unset, this will default to `127.0.0.1`.

### DIFF
--- a/conf/general.yml-example
+++ b/conf/general.yml-example
@@ -197,6 +197,10 @@ SMTP_PASSWORD: ''
 # this as is.
 GAZE_URL: 'https://gaze.mysociety.org/gaze'
 
+# Memcached host
+# This can be safely left out and will default to '127.0.0.1' even if not present.
+MEMCACHED_HOST: '127.0.0.1'
+
 # Should problem reports link to the council summary pages?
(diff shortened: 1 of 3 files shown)


conf/general.yml-example
```diff
@@ -223,3 +223,7 @@ LOGIN_REQUIRED: 0
 # If you want to stop new users from registering, set this to 1.
 # NB: This also disables all Facebook/Twitter logins.
 SIGNUPS_DISABLED: 0
+
+# Memcached host
+# This can be safely left out and will default to '127.0.0.1' even if not present.
+FMS_MEMCACHED_HOST: '127.0.0.1'
```
dracos: The `FMS_` prefix is a side-effect of our internal DB variable setup - it's not needed for any other variables (it's all FMS :) ). So I'd go with `MEMCACHED_HOST`.

dracos: Also, I'd perhaps put it next to the SMTP/Gaze config - it sort of fits in there in terms of placement of a perhaps-external service.

PR #2245, opened 2018-09-21
Switch to direct app server, not FastCGI.

dracos:
This allows `script/server` to take arguments, switches the example sysvinit script to use it rather than the FastCGI daemon, and the example nginx config to proxy rather than use FastCGI.


sagepe:
Looks good to me.

dracos:
Have added access-log and error-log entries.

PR #2248, opened 2018-09-24
Speed up fetching lists of bodies.

dracos:
Use a HashRefInflator wherever `all_sorted` is used, with consequential changes to deal with it now not being an object (e.g. add some `with_*` functions for manual fetching of extra data).


dracos:
Interesting thing thrown up by the new templates checker - I can fix the mentions of `b.get_column('area_count')` to be `b.area_count`, but any cobrand I don't know about, this will in a small way break it. The new script of #2244 will help by showing the user this change between releases, but instead I could put more code in so that `get_column('area_count')` still works and returns `area_count`; but it would be extra code.

PR #2251, opened 2018-09-25
[BANES] Branding for HTML emails

zarino:
[skip changelog] Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/19

~James is currently reviewing this on the staging site, and will let us know when he’s happy for it to be merged and deployed to the live B&NES site.~ – James has given us the thumbs up! 👍


zarino:
@davea `templates/email/banes/_email_top.html` has an extra media query block in it:

```
  @media only screen and (max-width: [% wrapper_max_width - 1 %]px) {
    #header {
      padding-left: 20px !important;
    }
  }
```

I needs the extra padding at narrow widths because otherwise the logo would touch the edge of the window. At wider widths, it doesn’t need the padding, because that way the logo lines up with the text container below it. It’s all caused by the BANES logo no longer being inside a coloured, padded box, basically :-/

PR #2252, opened 2018-09-26
WIP options for cobrand front end testing

struan:
[skip changelog]


.cypress/cypress/integration/regressions.js
```diff
@@ -5,22 +5,24 @@ Cypress.Commands.add('cleanUpXHR', function() {
     cy.visit('/404', { failOnStatusCode: false });
 });
 
+var pc = Cypress.env('postcode') || 'BS10+5EE';
```
dracos: You could put this in the `/.cypress/cypress.json` file:
```
{
  "env": {
    "postcode": "BS10 5EE"
  }
```
then that would be tidier code and only need to specify it in one place.

PR #2254, opened 2018-09-27
Add Docker + Compose support

dracos:


bin/docker.preinit
```diff
@@ -0,0 +1,42 @@
...
+# Make sure that the Postgres environment is up and running.
+echo "Testing connection to ${FMS_DB_HOST}."
+while ! pg_isready -h $FMS_DB_HOST >/dev/null 2>&1 ; do
+  echo "Still waiting for ${FMS_DB_HOST}..."
+  sleep 1
+done
+echo "Done."
+
+# If there's a password for the postgres user, set it up for root and see if we need
+# to create an FMS user. This is intended for use when using a deciated local postgres
```
dracos: dedicated

bin/install-as-user
```diff
@@ -92,19 +94,23 @@ then
...
-bin/update-schema --commit
-echo $DONE_MSG
 
 # Generate po and mo files (these invocations taken from Kagee's script):
 echo "Creating locale .mo files"
 commonlib/bin/gettext-makemo FixMyStreet
 echo $DONE_MSG
 
-bin/update-all-reports
+if [ $INSTALL_DB = true ]; then
```
dracos: This could be moved up to be with the other db stuff, no need for it to be here.

bin/install-as-user
```diff
@@ -92,19 +94,23 @@ then
...
+    # Create the database if it doesn't exist:
+    echo -n "Setting up database... "
+    if ! psql -l | egrep "^ *$DB_NAME *\|" > /dev/null
+    then
+        createdb --owner "$UNIX_USER" "$DB_NAME"
+        echo 'CREATE LANGUAGE plpgsql;' | psql -U "$UNIX_USER" "$DB_NAME" || true
+    fi
+    bin/update-schema --commit
+    echo $DONE_MSG
```
dracos: It does seem (once update-all-reports moved up here) that this overlaps with the middle bit of docker.preinit but not sure if worth factoring together somehow or not.

bin/site-specific-install.sh
```diff
@@ -26,27 +26,39 @@ misuse() {
...
+if [ $INSTALL_POSTFIX = true ]; then
+    install_postfix
+fi
 
-if [ ! "$DEVELOPMENT_INSTALL" = true ]; then
+if [ ! "$DEVELOPMENT_INSTALL" = true ] && [ ! "$DOCKER" = true ]; then
     install_nginx
     add_website_to_nginx
     # Check out the current released version
     su -l -c "cd '$REPOSITORY' && git checkout '$VERSION' && git submodule update" "$UNIX_USER"
```
dracos: I realise the checked out branch can now be set with BRANCH_OVERRIDE in the parent script and then again here with VERSION_OVERRIDE, but it seems a bit odd for the docker install not to run this line (which feels like a line you'd only want to bypass on a DEVELOPMENT_INSTALL).

bin/site-specific-install.sh
```diff
@@ -26,27 +26,39 @@ misuse() {
 [ -z "$DISTRIBUTION" ] && misuse DISTRIBUTION
 [ -z "$VERSION" ] && misuse VERSION
 [ -z "$DEVELOPMENT_INSTALL" ] && misuse DEVELOPMENT_INSTALL
+[ -z "$DOCKER" ] && misuse DOCKER
+[ -z "$INSTALL_DB" ] && misue INSTALL_DB
```
davea: Typo on `misuse` here.

conf/general.yml-docker
```diff
@@ -0,0 +1,233 @@
...
+# If you're behind a proxy, set this to a two-element list containing the
+# trusted HTTP header and the required value. For example:
+#   SECURE_PROXY_SSL_HEADER: [ 'X-Forwarded-Proto', 'https' ]
+
+# This is so the fixmystreet container knows it is running
+# behind an nginx container (not on the same host)
+using_frontend_proxy: 1
+
+# Email domain used for emails, and contact name/email for admin use.
+EMAIL_DOMAIN: '127.0.0.1.xip.io'
```
dracos: I've found nip.io more stable than xip.io, shrug

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM jgoerzen/debian-base-standard:stretch
+MAINTAINER bettercities@mysociety.org
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG REPOSITORY_URL_OVERRIDE=git://github.com/mysociety/fixmystreet.git
```
dracos: This line, and line below, won't be needed in the end.

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+        aliases:
+          - nginx.svc
+  fixmystreet:
+    image: sagepe/fixmystreet:0.5
+    tty: true
+    depends_on:
+      - postgres
+      - memcached
+    volumes:
+      - ./conf/general.yml-docker:/var/www/fixmystreet/fixmystreet/conf/general.yml
```
dracos: Not for now, perhaps, but it would be good to be able to generate this from the example so we don't have to remember to keep it up to date.

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    ports:
+      - "8000:80"
+    volumes:
+      - ./conf/nginx.conf-docker:/etc/nginx/conf.d/default.conf
+    networks:
+      default:
+        aliases:
+          - nginx.svc
+  fixmystreet:
+    image: sagepe/fixmystreet:0.5
```
sagepe: Before we merge, we should build and push an "official" version into the fixmystreet namespace.

Dockerfile
sagepe: Should we change the MAINTAINER line, too?

docker-compose.yml
sagepe: Yes, good idea.

Another possibility to consider is allowing the environment variables to override the defaults at start-up perhaps, so we don't need to maintain a separate file at all.

bin/install-as-user
sagepe: Not sure I completely understand - they run at such different stages in the lifecycle? Or do you mean calling out from the preinit script in some way to avoid the repetition?

It might be worth checking in some way for `plpgsql`, this just generates an error on more recent versions and it feels a bit inelegant (I left it out of the preinit script altogether).

bin/site-specific-install.sh
sagepe: Yes - good point, we want the public image to be the latest tagged release.

bin/install-as-user
dracos: Yeah, just the avoiding repetition was all, no biggie.

docker-compose.yml
sagepe: The current file is essentially an example of how you provide your own `general.yml` file - that's partly what I intended it to be, anyway.

bin/site-specific-install.sh
sagepe: Hmm, just thinking this might not work without an override at the moment as-is until at least `docker.preinit` is in the default `$VERSION` and that gets copied over after the scripts are run.

sagepe: It's only an issue before the initial merge, after that we'd want a copy consistent with the tagged release I think unless we were building test images in which case we can add VERSION_OVERRIDE while building as you suggest.

I'll build the first version of the container user `git show`, but omit this from the final version.

sagepe: Hmm, or perhaps include the `git show` until we have a tagged release that contains these commits.

dracos: Yeah, don't really think it matters - could fetch from github like the install-site even ;)

conf/general.yml-docker
sagepe: If we're going to change this, let's do it as a separate exercise - there are a few references in various places and repos that would need updating for consistency.

sagepe:
I think we're pretty much ready to merge this, then I can build a matching container and push to Docker Hub.

PR #2257, opened 2018-09-28
Vagrant: Use the mysociety/fixmystreet box by default

sagepe:
This sets the default box to be `mysociety/fixmystreet` and provides
an optional argument, `--base-box`, that can be used to override this.

When using the `mysociety/fixmystreet` box, the Perl modules are
prebuilt in `/usr/share/fixmystreet/local` and are bind-mounted into
`/home/vagrant/fixmystreet/local`. This ensures that these are
compatible with the guest machine and speeds up an initial launch.

The `mysociety/fixmystreet` box is also based on Debian Stretch so is
a closer match to our current production environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
         - Add script to list/diff template changes in core that
           might need applying to a cobrand.
         - Move away from FastCGI in sample conf/sysvinit config.
+        - Customised Vagrant box available, with an override option.
 
 * v2.4 (6th September 2018)
     - Security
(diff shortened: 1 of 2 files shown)


sagepe:
Not the whole script; for now it's a minimal Debian install with Virtualbox Additions, the dependencies for FixMyStreet and a pre-compiled set of Perl modules (FMS specific bits are in [here](https://github.com/mysociety/public-builds/blob/master/packer/scripts/04_site.sh) for now at least). The current pattern for the development install of using the user's local repository is still used.

This seemed like a good start; I think we can add more of the build steps in due course (for example all the locale stuff, etc) but I wanted to get something up and running quickly without making the box file too big and unwieldy or changing the workflow dramatically.

PR #2250, opened 2018-09-25
WIP area page front end tests

struan:
Fixes mysociety/fixmystreet-commercial#1188
Fixes mysociety/fixmystreet-commercial#1186

PR #2262, opened 2018-10-01
Version 2.4.1

dracos:
Need to:
- [x] Reply reverted commit in commit after release commit.
- [x] Get #2250 done/merged with history bugfix

Please check the following:


cpanfile
```diff
@@ -82,7 +82,7 @@ requires 'namespace::autoclean';
 requires 'Net::Amazon::S3';
 requires 'Net::DNS::Resolver';
 requires 'Net::Domain::TLD', '1.75';
-requires 'Net::Facebook::Oauth2', '0.10';
+requires 'Net::Facebook::Oauth2', '0.11';
```
struan: Might be useful to mention this in the release notes somewhere in case people don't always run the module updates?

dracos: They'll have had issues many times :) https://fixmystreet.org/updating/ tells you you need to run `script/update` every time you update the code; I'd actually like that first thing top of the page, will think of some way to reword it for those who already know about git.

PR #2268, opened 2018-10-02
Example systemd service unit file

sagepe:
This adds a `systemd` service unit file for managing the FixMyStreet Catalyst process manager.

Please check the following:


sagepe:
I've tested it and made a very small change, but it seems to work fine in an AMI build.

sagepe:
Just noting that I've rebased in an update to `commonlib` to include the updated install scripts that support using this file.

PR #2269, opened 2018-10-02
[FixaMinGata] Incremental updates

jonkri:


dracos:
Hi @jonkri - what's your aim with this PR? It contains two commits already present in v2.4/master, a merge of v2.3.4 and also two FixaMinGata specific commits - do you want those two applying on top of current upstream master which seems most likely? Or tidied up on top of v2.3.4, or something else? Let me know :)

PR #2271, opened 2018-10-03
OCC reference search

struan:
Please check the following:

Allow cobrands to do their own search from the front page if the string matches `lookup_by_ref_regexp'.
Add in a custom search for OCC if the search matches a customer_reference.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1200

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -399,10 +399,13 @@ sub _geocode : Private {
 sub lookup_by_ref : Private {
     my ( $self, $c, $ref ) = @_;
 
-    my $problems = $c->cobrand->problems->search([
-        id => $ref,
-        external_id => $ref
-    ]);
+    my $criteria = $c->cobrand->call_hook("lookup_by_ref", $ref) ||
+        [
+            id => $ref,
+            external_id => $ref
+        ];
+
+    my $problems = $c->cobrand->problems->search( $criteria );
 
     my $count = try {
         $problems->count;
--- a/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
@@ -53,6 +53,22 @@ sub users_can_hide { return 1; }
 
 sub default_show_name { 0 }
 
+sub lookup_by_ref_regex {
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -53,6 +53,27 @@ sub users_can_hide { return 1; }
...
+    my ($self, $ref) = @_;
+
+    my $criteria = [
+        id => $ref,
+        external_id => $ref
+    ];
+
+    if ( $ref =~ /^ENQ/ ) {
+        my $filter = "%T18:customer_reference,T\\d+:$ref,%";
+        $criteria = { 'extra' => \[ 'SIMILAR to ?', [ {} => $filter ] ] };
```
dracos: This could be `{ extra => [ '~', "T18:customer_reference,T\\d+:$ref," ] }`? All `SIMILAR TO` does is rewrite to a regex afaik.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -398,10 +398,13 @@ sub _geocode : Private {
 sub lookup_by_ref : Private {
     my ( $self, $c, $ref ) = @_;
 
-    my $problems = $c->cobrand->problems->search([
-        id => $ref,
-        external_id => $ref
-    ]);
+    my $problems;
+    unless ( $problems = $c->cobrand->call_hook("lookup_by_ref", $ref) ) {
```
dracos: Unless we foresee some weird things being done by this hook, should it perhaps return the parameter for `search()` but not do the search itself? Then it could return nothing to stick with the default also.

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -53,6 +53,27 @@ sub users_can_hide { return 1; }
...
 sub default_show_name { 0 }
 
+sub lookup_by_ref_regex {
+    return qr/^\s*((?:ENQ)?\d+)\s*$/;
+}
+
+sub lookup_by_ref {
+    my ($self, $ref) = @_;
+
+    my $criteria = [
```
dracos: Don't need this now due to your other change; you can return the criteria if `$ref =~ /^ENQ/` and nothing if not and it'll fall back.

PR #2272, opened 2018-10-03
Deal with a variety of low-level test warnings.

dracos:
* Stop red warning being output by S3 bucket creation failure test
* Make sure 'num_old_reports' is always a number on /around pages
* Stop 'category_extra' being generated when unnecessary
* Prevent warning printed in getservicerequestupdates.t
* Prevent info lines from being output in develop.t

[skip changelog]

### DIFF
--- a/t/app/controller/root.t
+++ b/t/app/controller/root.t
@@ -75,7 +75,6 @@ FixMyStreet::override_config {
 };
 
 subtest "check_login_disallowed cobrand hook" => sub {
-    warn '#' x 50 . "\n";
     my $cobrand = Test::MockModule->new('FixMyStreet::Cobrand::Default');
     $cobrand->mock('check_login_disallowed', sub {
             my $self = shift;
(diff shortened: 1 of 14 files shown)


t/app/controller/root.t
```diff
@@ -75,7 +75,6 @@ FixMyStreet::override_config {
 };
 
 subtest "check_login_disallowed cobrand hook" => sub {
-    warn '#' x 50 . "\n";
```
struan: oops, that is totally me leaving debug in :(

t/photostorage/s3.t
```diff
@@ -112,7 +113,9 @@ FixMyStreet::override_config {
             $create_bucket_called = 1;
         });
 
-        ok !$s3->init(), "PhotoStorage::S3::init failed";
+        warning_like {
+            $s3->init();
```
davea: Does this no longer test the return value of `init()`, or does `warning_like` handle that?

dracos: Hmm, you're probably right that it no longer does, not sure if I had an issue combining with `warning_like` or just missed it quite possibly.

PR #2274, opened 2018-10-04
update Params::Classify to avoid build problems on later Perls

struan:
e.g. module installation fails on perl 5.28 due to changes to
op codes.

Please check the following:


dracos:
Ah, looks like we worked on this at the same time! https://github.com/mysociety/fixmystreet/compare/master...test-5.26

dracos:
Have added second commit to https://github.com/mysociety/fixmystreet/pull/2272 with this included and other bits if you want to take a look.

PR #2277, opened 2018-10-05
Move send-comments to package and related

### LINKED ISSUE #2109: Move send-comments code into a module
Most of the `send-comments` code is in the script itself which means it's not being tested. As with `send-reports` we should move this into a Script module so we can write tests for it.

### LINKED ISSUE #2170: Improve Open311 send/fetch-comments tests
`send-comments` doesn't seem to have any test coverage, so should probably be refactored into `Open311/PostServiceRequestUpdates.pm` and tested.
Similarly, the tests for `Open311::GetServiceRequestUpdates` should be improved to test the `fetch` function which is used by `fetch-comments`.

dracos:
Fixes #2109. Fixes #2170.

### DIFF
--- a/bin/send-comments
+++ b/bin/send-comments
@@ -23,13 +23,7 @@ use Open311::PostServiceRequestUpdates;
 
 my ($verbose, $nomail) = CronFns::options();
 
-# Set up site, language etc.
-my $base_url = FixMyStreet->config('BASE_URL');
-my $site = '';
-$site = 'fixmystreet.com' if $base_url eq "https://www.fixmystreet.com";
-
 my $updates = Open311::PostServiceRequestUpdates->new(
-    site => $site,
     verbose => $verbose,
 );
 $updates->send;
--- a/bin/send-comments
+++ b/bin/send-comments
@@ -18,167 +18,18 @@ BEGIN {
 }
 
 use CronFns;
-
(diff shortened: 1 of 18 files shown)


bin/send-comments
```diff
@@ -18,162 +18,18 @@ BEGIN {
...
-use DateTime;
 use FixMyStreet;
-use FixMyStreet::Cobrand;
-use FixMyStreet::DB;
-use Open311;
-
-# send_method config values found in by-area config data, for selecting to appropriate method
-use constant SEND_METHOD_EMAIL      => 'email';
-use constant SEND_METHOD_OPEN311    => 'Open311';
+use Open311::PostServiceRequestUpdates;
```
struan: yay!

perllib/Open311/PostServiceRequestUpdates.pm
```diff
@@ -58,6 +59,10 @@ sub open311_params {
         $open311_conf{use_customer_reference} = 1;
     }
 
+    if ( $body->areas->{+COUNCIL_ID_BUCKS} ) {
+        $open311_conf{mark_reopen} = 1;
+    }
```
struan: I would prefer this if it was in the DB, especially as there's a chance of using this with OCC.

PR #2279, opened 2018-10-08
Speed up dashboard CSV export.

### LINKED ISSUE #2278: Dashboard update CSV export is very slow
Looks like it's missing a prefetch for problem (and user for Bath's extra columns) so it's doing an awful lot of DB queries it doesn't need to.

dracos:

Fixes #2278


dracos:
Changed after to split user db query into two, as two separate parts 80ms each but together 20s. Had tested separate, assumed putting together would make no difference, fool.

PR #2281, opened 2018-10-09
[Vagrant] Improve provisioning for mySociety base box

sagepe:
- When using the official box, just run `script/update` at startup as
   the latest release (0.0.3) has everything fully pre-installed;
 - When using the official box, always run the provisioner that checks
   and bind mounts the local pre-built Perl modules when starting as
   these would previously have been missing after a `halt` and `up`;
 - When using any other box, just run the full install process.

[skip ci]

Please check the following:

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,10 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = "2"
+
+# Provide for a `--base-box` option to permit overriding the box used
 require 'getoptlong'
 opts = GetoptLong.new(
   [ '--base-box', GetoptLong::OPTIONAL_ARGUMENT ]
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -80,11 +102,18 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
   config.vm.synced_folder ".", "/home/vagrant/fixmystreet", :owner => "vagrant", :group => "vagrant"
 
-  config.vm.provision "shell" do |s|
-    s.inline = $setup
-    s.args   = "#{baseBox}"
+  # When using the mySociety box, just mount the local perl modules and run `script/update`
+  # For any other box, just run the full setup process.
+  if "#{baseBox}" == "mysociety/fixmystreet"
+    config.vm.provision "shell", run: "always", inline: $mount_modules
+    config.vm.provision "shell", inline: $update
```
dracos: I'm wondering if this should be always run too; would save you having to do it/remember to do it after logging in.

sagepe: I did wonder about that, in the end I didn't because I thought users might want to decide for themselves whether to run it or not. I don't have a strong opinion, so would be happy to make this change.

sagepe:
In the new box, `/home/vagrant/.cpanm` should be owned by `vagrant` as I've changed the way it gets built. I don't think that these changes are fully compatible with the older box, perhaps I should have bumped the version to `0.1.0` and made this clearer.

sagepe:
I've made those changes. The corresponding Vagrant box was built with the updated builder here in PR [5](https://github.com/mysociety/public-builds/pull/5) in the public-builds repo.

PR #2285, opened 2018-10-09
Add cobrand hook for dashboard viewing.

dracos:
This allows a cobrand to e.g., as the test does, allow public access to dashboard CSV export for a body, but not otherwise.

### DIFF
--- a/t/app/controller/dashboard.t
+++ b/t/app/controller/dashboard.t
@@ -236,15 +236,12 @@ FixMyStreet::override_config {
 sub test_table {
     my ($content, @expected) = @_;
     my $res = $categories->scrape( $mech->content );
-    my $i = 0;
+    my @actual;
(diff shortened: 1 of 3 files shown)


t/app/controller/dashboard.t
```diff
@@ -230,21 +243,37 @@ FixMyStreet::override_config {
...
     my $res = $categories->scrape( $mech->content );
-    my $i = 0;
+    my @actual;
     foreach my $row ( @{ $res->{rows} }[1 .. 11] ) {
-        foreach my $col ( @{ $row->{cols} } ) {
-            is $col, $expected[$i++];
-        }
+        push @actual, @{$row->{cols}} if $row->{cols};
     }
+    is_deeply \@actual, \@expected;
```
struan: I am being picky here but it might be nice if this unrelated change was in it's own commit, or explained in the commit message.

PR #2289, opened 2018-10-12
Bathnes roadworks

struan:
Add roadworks.org layer to Bathnes cobrand.

Also introduces ability to filter on which roadworks messages are shown.

Fixes mysociety/fixmystreet-freshdesk#17

### DIFF
--- a/web/cobrands/bathnes/js.js
+++ b/web/cobrands/bathnes/js.js
@@ -37,6 +37,20 @@ fixmystreet.roadworks.display_message = function(feature) {
         $('.change_location').after($msg);
 };
 
+fixmystreet.roadworks.filter = function() {
+  var category = $('select#form_category').val(),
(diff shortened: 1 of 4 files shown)


web/cobrands/bathnes/js.js
```diff
@@ -0,0 +1,72 @@
...
+        $('.change_location').after($msg);
+};
+
+fixmystreet.roadworks.filter = function(feature) {
+  var category = $('select#form_category').val(),
+      parts = feature.attributes.symbol.split(''),
+      valid_types = ['h', 'n', 'l', 'w'],
+      valid_subtypes = ['15', '25'],
+      type = parts[2],
+      sub_type = parts[4] + parts[5],
```
davea: This is `subtype` below, typo?

PR #2291, opened 2018-10-15
Allow moderation to potentially change photos.

dracos:
Lets a cobrand provide a standard photo upload form (as on new report/update) in the report display so moderation can add/remove photos to a report, not just show/hide.


dracos:
Some fixups for the error, test, and then to prevent clashes with the update photo upload (both JS only looking for one form, and server not putting the problem photos in the update photo upload preview).

PR #2292, opened 2018-10-15
list user's alerts on user edit page in admin

### LINKED ISSUE #676: show list of alerts currently subscribed-to when looking at a user in the admin

struan:
Include a list of alerts the user is subscribed to at the bottom of the
user_edit page in the admin. Also allows disabling, enabling and deleting of alerts.

Fixes #676 

Please check the following:

Thanks for contributing!

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -1512,10 +1512,6 @@ sub user_edit : Path('user_edit') : Args(1) {
     $c->forward('fetch_all_bodies');
     $c->forward('fetch_body_areas', [ $user->from_body ]) if $user->from_body;
 
-    unless ( $c->cobrand->moniker eq 'zurich' ) {
-        $c->forward('user_alert_details');
-    }
-
     if ( defined $c->flash->{status_message} ) {
         $c->stash->{status_message} =
             '<p><em>' . $c->flash->{status_message} . '</em></p>';
@@ -1538,6 +1534,8 @@ sub user_edit : Path('user_edit') : Args(1) {
         my %args = ( email => $email );
         $args{user_id} = $id if $user->email ne $email || !$user->email_verified;
         $c->forward('send_login_email', [ \%args ]);
+    } elsif ( $c->get_param('update_alerts') ) {
+        $c->forward('update_alerts');
     } elsif ( $c->get_param('submit') ) {
 
         my $edited = 0;
@@ -1715,6 +1713,11 @@ sub user_edit : Path('user_edit') : Args(1) {
             active => $active_contacts{$_->id},
         } } @live_contacts;
         $c->stash->{contacts} = \@all_contacts;
+    }
+
+    # this goes after in case we've delete any alerts
+    unless ( $c->cobrand->moniker eq 'zurich' ) {
+        $c->forward('user_alert_details');
     }
 
     return 1;
@@ -1943,6 +1946,25 @@ sub ban_user : Private {
         $c->stash->{username_in_abuse} = 1;
     }
     return 1;
+}
+
+sub update_alerts : Private {
+    my ($self, $c) = @_;
+
+    my $changes;
+    for my $alert ( $c->stash->{user}->alerts ) {
+        my $edit_option = $c->get_param('edit_alert[' . $alert->id . ']');
+        next unless $edit_option;
+        $changes = 1;
+        if ( $edit_option eq 'delete' ) {
+            $alert->delete;
+        } elsif ( $edit_option eq 'disable' ) {
+            $alert->disable;
+        } elsif ( $edit_option eq 'enable' ) {
+            $alert->confirm;
+        }
+    }
+    $c->flash->{status_message} = _("Updated!") if $changes;
 }
 
 sub user_logout_everywhere : Private {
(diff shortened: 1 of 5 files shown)


templates/web/base/admin/user-alerts.html
```diff
@@ -0,0 +1,52 @@
+[% IF user.alerts %]
+<h3>[% loc("User's alerts") %]</h3>
+<form method="POST">
+<input type="hidden" name="token" value="[% csrf_token %]" >
+<input type="hidden" name="update_alerts" value="1" >
+<table>
+<tr>
+    <th>ID</th>
```
dracos: This all needs translations.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1844,6 +1850,27 @@ sub get_user : Private {
     return $user;
 }
 
+sub user_alert_details : Private {
+    my ( $self, $c ) = @_;
+
+    my $alerts = $c->stash->{user}->alerts;
```
dracos: Probably want to `prefetch` with alert_type to save one query per row in the next bit. Plus stick this on the stash and use in the template because otherwise `user.alerts` will do another query both times in the template.

templates/web/base/admin/user-alerts.html
```diff
@@ -0,0 +1,52 @@
...
+    <th>Details</th>
+    <th>Enable</th>
+    <th>Disable</th>
+    <th>Delete</th>
+</tr>
+[% FOREACH alert IN user.alerts %]
+<tr>
+    <td>[% alert.id %]</td>
+    <td>[% alert.alert_type.ref %]</td>
+    <td>[% alert.parameter %] [% alert.parameter2 %]</td>
```
dracos: No need for this if the later column lists them too. (Or that could be fallback for that column if not all cases covered?)

templates/web/base/admin/user-alerts.html
```diff
@@ -0,0 +1,52 @@
...
+            New problems near <a href="[% c.uri_for( '/around', { lon => alert.parameter, lat => alert.parameter2 } ) %]">[% alert.parameter %],[% alert.parameter2 %]</a>
+        [% CASE 'ward_problems' %]
+            [% body = alert.parameter %]
+            [% ward = alert.parameter2 %]
+            New problems for [% IF alert_areas.$ward %]<a href="[% c.uri_for('/reports', body_names.$body, alert_areas.$ward.name ) %]">[% alert_areas.$ward.name %]</a> ward[% ELSE %]ward id [% alert.parameter2 %][% END %] in [% IF body_names.$body %]<a href="[% c.uri_for('/reports', body_names.$body) %]">[% body_names.$body %]</a>[% ELSE %] body id [% body %][% END %]
+        [% CASE 'council_problems' %]
+            [% body = alert.parameter %]
+            New problems for [% IF body_names.$body %]<a href="[% c.uri_for('/reports', body_names.$body) %]">[% body_names.$body %]</a>[% ELSE %][% body %][% END %]
+        [% CASE 'area_problems' %]
+            New problems for area id [% alert.parameter %]
```
dracos: Could be a link to MapIt?

templates/web/base/admin/user-alerts.html
```diff
@@ -0,0 +1,52 @@
...
+    [% END %]
+    </td>
+    <td><input type="checkbox" name="enable_alert[[% alert.id %]]" value="[% alert.id %]"[% ' disabled' UNLESS alert.whendisabled %]></td>
```
dracos: I'm not sure checkboxes are the right UI here, given you won't want to pick more than one - radio buttons, perhaps?

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1916,6 +1943,25 @@ sub ban_user : Private {
...
+sub update_alerts : Private {
+    my ($self, $c) = @_;
+
+    my $changes;
+    for my $alert ( $c->stash->{user}->alerts ) {
+        if ( $c->get_param('delete_alert[' . $alert->id . ']') ) {
+            $alert->delete;
+            $changes = 1;
+        } elsif ( $c->get_param('disable_alert[' . $alert->id . ']') ) {
+            $alert->update({ whendisabled => \"now()" });
```
dracos: alert has a `disable` function. Going to try this new GitHub thing...
```suggestion
            $alert->disable;
```

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1916,6 +1943,25 @@ sub ban_user : Private {
...
+            $changes = 1;
+        } elsif ( $c->get_param('enable_alert[' . $alert->id . ']') ) {
+            $alert->update({ whendisabled => undef });
```
dracos: Also has a `confirm` function which you might want to use here (I guess you'll want to confirm the alert if you're reenabling it and it isn't, otherwise there's no way to do so).

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1916,6 +1943,25 @@ sub ban_user : Private {
     return 1;
 }
 
+sub update_alerts : Private {
+    my ($self, $c) = @_;
+
+    my $changes;
+    for my $alert ( $c->stash->{user}->alerts ) {
```
dracos: This could use stashed alerts if they were put on stash on lookup. But then I guess you need to update that stash for any deleted ones...

perllib/FixMyStreet/App/Controller/Admin.pm
dracos: I think so, I imagine someone not getting confirmation email and contacting support is a thing that could happen.

dracos:
Having a ward alert would test the remaining bit, icing on cake 🎂

PR #2295, opened 2018-10-16
Simplify /auth page

### LINKED ISSUE #2208: Make the 'Sign in' interface easier to understand
The `/auth` page, labelled in the menu as 'Sign in', currently covers multiple tasks and I think it may have become a bit overloaded over time. On it, you can:

* Sign in with a password 
* Sign in without a password
* Change your password 
* Register for an account

What I’m presuming is the default case here (that if you click a ‘Sign in’ link, you have an email address and password ready) is split up by needing to answer the question of whether you have a password or not.

The ‘not’ case is made more complicated by the fact that although it’s signposted as not having a password, it gives you an optional password field.

It also has a field for your name, which presumably doesn't do anything when signing in, only when 
this form is being used to register for an account.

The password change element is a bit inconsistent with elsewhere - you can set a password here without knowing your old password, but if you try to do it from within your account, it looks like you need to give your old password.

The 'How to register' instructions on the help page are not very straightforward and if you follow them, you can get an account without a name, which doesn't seem ideal. Presumably this is a low volume path, but it would be nice to give it first order status. 

An initial proposal:

**Sign in**

The page that you get to from 'Sign in' is simple - it has fields for your email, password (required) and the 'keep me signed in' checkbox.

It also has a link that says something like "No password? No problem. If you forgot your password or never set one we can confirm who you are by email". This takes them to another page that just asks for their email address and says we’ll send them an email so that they can confirm who they are and tells them that once they're into your account they can change or set a password.

Maybe a separate registration page that has an email address field, a required password field and a required name field. 

**Account page**

You should be able to set a name on your account page as at the moment (I think) have an account with no name attached to it. When does your name get attached to your account? Is it automatically transferred from a problem report? 

if you don't have a password, it should say "set a password" not "change your password"

dracos:
See sketch on ticket for what it should be. Fixes #2208. Splits up sign in/create/forgotten password to separate views (note we think this flow is less used than e.g. new report flow), with email/password question first.
- [ ] Needs some design work to make sign in button look like email, tidy up heading run-on, probably something else I've missed.


davea:
A few bits I've spotted:

Would be nice to focus the password field if the user clicks the "sign in with password" button or hits enter (or tab!) in the email field:
signin

Should we ask for the new password twice when creating an account?
image
(nb we already do this when changing your password:
image
)

This has sadly turned FMS into one of those annoying sites whose login forms break 1Password's form submission:
1password

dracos:
"Should we ask for the new password twice when creating an account?" - I dunno, I don't think anyone does this, so I'd probably prefer stopping doing it on change password (perhaps paired with a 'show password' various sites have e.g. BBC).

"This has sadly turned FMS into one of those annoying sites whose login forms break 1Password's form submission" - gah, I specifically tried this (I use it too) and it works for me - but I don't have auto-submit turned on. So it filled in email, and when I clicked "Sign in with password" it had pre-filled password. So I guess I just need to figure out how it can find the submit...

dracos:
Hmm, even clicking 'sign in with password' first and then autofill 1password with submit on doesn't appear to submit. The form at that point should be standard. So not sure what I've done/broken/should change.

dracos:
It should now work on enter in email box, focus on password box, and work with (at least 1password) autosubmit (it shows the form quickly on password change, which appears to be enough here anyway).

templates/web/base/auth/general.html
```diff
@@ -77,61 +74,27 @@ <h3>[% tprintf(loc("Do you have a %s password?", "%s is the site name"), site_na
...
 
         <div class="form-txt-submit-box">
-            <input type="password" name="password_sign_in" class="form-control" id="password_sign_in" value="">
+            <input type="password" name="password_sign_in" class="form-control" id="password_sign_in" value="" autocomplete="current-password">
             <input class="green-btn" type="submit" name="sign_in_by_password" value="[% loc('Sign in') %]">
         </div>
 
-        <div class="checkbox-group">
-            <input type="checkbox" id="remember_me" name="remember_me" value='1'[% ' checked' IF remember_me %]>
-            <label class="inline n" for="remember_me">[% loc('Keep me signed in on this computer') %]</label>
```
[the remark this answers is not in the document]
dracos: It's tricky. It might mean they have to sign in less often. On a phone (assuming browser just left alone, never dying except on version upgrade), or on a desktop with modern 'sleep' rather than being shut down (or indeed, Chrome/Firefox "Continue Where I Left Off" restores session cookies, or at least certainly used to), a session cookie could easily end up lasting longer than a timed out cookie.

OTOH, 'keep me signed in' would set a timed cookie and the cookie expiry would move into the future as you used the site, so would be 4 weeks from last use. So perhaps it should always use a timed cookie rather than always use a session cookie; but either way we should pick one I think.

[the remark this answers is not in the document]
davea: Perhaps I've just become sensitised because I get irritated when I have to sign back in to Fantasy Bake Off *every week* (I mean really, set the cookie length to 8 days or something) or perhaps I'm just irritable 😃.

Do we have any clue how many people tick that box?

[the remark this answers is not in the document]
dracos: In past four weeks, looks like we have 29k sessions in the db, of which 24.5k are session, 4.5k.
I think I'd be happy with a timed, extending, cookie. BBC doesn't mention anything on sign in, and says elsewhere in help their cookie keeps you signed in for two years (it is indeed a two year cookie).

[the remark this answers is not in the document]
davea: 👍 to that idea.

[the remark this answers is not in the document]
dracos: On checking, that is what this code is already doing, I think, so yay past me :)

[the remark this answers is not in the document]
dracos: Oh, except if they come from a token, when it explicitly makes a session cookie in that case. Dunno if we should change that to match/be consistent.

web/cobrands/sass/_base.scss
```diff
@@ -849,8 +870,14 @@ input.final-submit {
 }
 
 .btn--block {
-    display: block;
-    text-align: center;
+    // !importants to override any .btn / .btn-primary styles defined in cobrands
```
dracos: Is this needed for a cobrand you're aware of? I can't see any that it would affect from a quick look; presumably they've all been okay without the `!important` so far. Just wondering if it'd be nicer to go without them if possible.

zarino: I needed it in fixmystreet.com cobrand because `.btn-primary` sets its own `width: auto`, and is defined _after_ `.btn--block`, so takes precedence. Didn’t check any of the other cobrands.

dracos: Ah I didn't find that one because I searched for `btn--`, odd that one (and btn-danger) is named differently from the rest.
Shall we fix that, then? My idea would be button-reset should itself include two mixins, one for layout, one for colours, and then .com (and lincs and oxon) could include the colour mixin and not affect the layout, and then it'd all be flat?

zarino: Yes, we started moving towards a bootstrap-style `class="btn btn-modifier"` (or, given our leaning towards BEM, `btn--modifier`) pattern a little while ago, but never completed it. I agree it’d be good to set layout-related stuff in a `.btn` class, and then presentation/colour-related stuff in `.btn-primary`, `.btn-default` etc.

Do you think that should be done as part of this PR though? It could be quite far-reaching. Ideally we’d use it as an opportunity to clean up the classes applied to button elements in every cobrand’s templates…

dracos: I would say don't let the perfect be the enemy of the good, and do what is needed to not have `!important`s on all the settings (looks like btn-primary is only thing affected, and I think .com/Lincs/Oxfordshire are all that is affected) and note other thing you'd like to do so they're not forgotten.

web/cobrands/oxfordshire/base.scss
```diff
@@ -213,6 +213,24 @@ dd, p {
...
+// We want OCCLSS buttons to look like FMS buttons.
+.occlss-button--primary {
+    @include button-size();
+}
+
+// No border and slightly different padding for *all* types of button.
+.btn,
+#front-main #postcodeForm div input#sub,
+.green-btn,
+button.green-btn,
```
dracos: There aren't any button with class `green-btn`. Also I don't think there are any input.green-btn (there are inputs, but nothing that would need the specificity?)

web/cobrands/sass/_base.scss
```diff
@@ -754,7 +754,16 @@ footer {
 
 /*BUTTONS*/
 
-.btn {
+.btn  {
```
dracos: Double space

web/cobrands/sass/_base.scss
```diff
@@ -754,7 +754,16 @@ footer {
...
+  @include button-reset;
+}
+
+// Deprecated
+// Try to use `btn btn--primary` etc instead of `btn-primary` where possible.
+.btn-primary,
+.green-btn,
+.btn-danger,
+.red-btn {
   @include button-reset;
```
dracos: Should this be button-size? Given you then call button-variant on all options of this.

web/cobrands/oxfordshire/base.scss
zarino: Yup. I have a feeling these selectors have all just been cargo culted (in the words of hart.scss) from older cobrands. Maybe the extra specificity was required a while back, but I can’t spot anywhere that needs it now. 👍

web/cobrands/oxfordshire/base.scss
```diff
@@ -213,35 +213,49 @@ dd, p {
...
+// We want OCCLSS buttons to look like FMS buttons.
+.occlss-button--primary {
+    @include button-size();
+}
+
+// No border and slightly different padding for *all* types of button.
+.btn,
+#front-main #postcodeForm div input#sub,
+.green-btn,
+input.green-btn,
```
dracos: Another to be removed.

web/cobrands/oxfordshire/base.scss
```diff
@@ -213,35 +213,49 @@ dd, p {
...
     );
-    border: none;
-    padding: 0.7em 1em 0.85em 1em;
 }
 
+// Override all the "primary" buttons across the site!
 #front-main #postcodeForm div input#sub,
 .green-btn,
-button.green-btn,
 input.green-btn,
```
dracos: Another to be removed.

PR #2296, opened 2018-10-16
Update areas when moving report pin

### LINKED ISSUE #2181: Moving existing report pin does not reset areas
> Not an urgent one this.. but a bug nonetheless.
> If a report is moved across a divisional boundary it is still associated with the original division.
> See: https://fixmystreet.oxfordshire.gov.uk/reports/Oxfordshire/North+Hinksey
screen shot 2018-07-03 at 10 07 05

> and note the pin way out to the right.
> It’s this report: https://fixmystreet.oxfordshire.gov.uk/report/1329287
> Which was moved from within North Hinksey into Oxford.

davea:
The change in report_inspect.t from WODC to Oxford City was necessary
because moving reports to a location not in Mock::MapIt was resulting in
an empties bodies_str in the "test category/body changes" test.

Fixes #2181.

Please check the following:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
         - Add perl 5.26/5.28 support.
         - Fix subcategory issues when visiting /report/new directly #2276
         - Give superusers access to update staff dropdowns. #2286
+        - Update report areas when moving its location. #2181
     - Development improvements:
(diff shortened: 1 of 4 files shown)


davea:
Merged in 08e335ce05c81db337f3f7623e1e3fe734546ce9.

PR #2297, opened 2018-10-18
WIP cobrand validation 

struan:
Add some validation restrictions for specific cobrands

Fixes mysociety/fixmystreet-commercial#1144

Please check the following:

### DIFF
--- a/web/cobrands/rutland/validation_rules.js
+++ b/web/cobrands/rutland/validation_rules.js
@@ -1,0 +1,4 @@
+validation_rules.name = {
+  required: true,
+  maxlength: 40
+};
--- a/templates/web/buckinghamshire/footer_extra_js.html
+++ b/templates/web/buckinghamshire/footer_extra_js.html
@@ -1,0 +1,3 @@
+[% scripts.push(
+    version('/cobrands/buckinghamshire/validation_rules.js'),
+) %]
--- a/templates/web/lincolnshire/footer_extra_js.html
+++ b/templates/web/lincolnshire/footer_extra_js.html
@@ -1,0 +1,3 @@
+[% scripts.push(
+    version('/cobrands/lincolnshire/validation_rules.js'),
+) %]
--- a/templates/web/rutland/footer_extra_js.html
+++ b/templates/web/rutland/footer_extra_js.html
@@ -1,0 +1,3 @@
+[% scripts.push(
+    version('/cobrands/rutland/validation_rules.js'),
+) %]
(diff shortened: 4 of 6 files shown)


perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -328,6 +328,21 @@ sub report_check_for_errors {
         );
     }
 
+    # can't put this in the relevant cobrand as it won't then apply on national site
+    unless ( $errors{name} ) {
+        if ( $report->to_body_named('Buckinghamshire') && length($report->name) > 50 ) {
+            $errors{name} = sprintf( _('Names are limited to %s characters in length.'), 50 );
```
davea: Don't think we need translations here, unless these counties are planning their own localisations :) I guess it doesn't matter from a site performance POV but would save putting extra strings in `FixMyStreet.po`.

struan:
The problem with that is that the validation needs to apply on the national site as well so you can't really put it in the cobrand. Unless we did something similar to the open311-adapter and had the validation do something like

```
foreach ($self->plugins) {
     if ($report->to_body_names($_->body)) {
          $_->validate($report);
    }
}
```

And then have some sort of `FixMyStreet::Validation::` namespace. Which I guess might work and then could have roles.

davea:
Hmm, yes, that does make it a bit more complicated. There is `Body::get_cobrand_handler` which returns the cobrand module that the report was sent to, that might be helpful?

struan:
So, I've got: https://github.com/mysociety/fixmystreet/pull/2309 which sort of addresses this.

I've had a play with using this:

```perl
    my $cobrand = $self->get_body_handler_for_problem($report);
    if ( $cobrand->can('report_validation') ) {
        $cobrand->report_validation( $report, \%errors );
    }
```

Which works but that has the problem that you need a cobrand and it needs to be in `ALLOWED_COBRANDS` which in practice always seems to be the case but I guess might not be?

web/cobrands/fixmystreet-uk-councils/confirm_validation_rules.js
```diff
@@ -0,0 +1,12 @@
+validation_rules.name = {
+  required: true,
+  maxlength: 50
+};
+validation_rules.phone = {
+  required: true,
```
davea: Don't think this `required` needs to be here.

struan:
I've added an extra commit on to this that runs the JS rules on the national site as well. I am not totally convinced it's a good idea.

web/cobrands/rutland/validation_rules.js
```diff
@@ -0,0 +1,4 @@
+validation_rules.name = {
```
davea: This file doesn't seem to be required any more, I think?

PR #2298, opened 2018-10-18
Restore zoom level when navigating back to /around from /report

### LINKED ISSUE #1649: Automatically zoom in on report pin
A FixaMinGata user has made a suggestion that could be a UX improvement for the FixMyStreet platform.

When a user clicks on a report, the map pin becomes larger. The request is to also have the map zoom in on the report. The benefit of this would be to more clearly see where on the map the report is.

If this behaviour was realized, then the map could also be made to automatically zoom back out when the navigating back to the report list.

I'm not sure what should happen if the user has panned or zoomed the map while viewing the report, but my spontanous idea is that the map should be reset to the way it was prior to clicking on the report initially, since any zooming or panning seems unlikely to be relevant at the point at which the user hits the back button anyway.

davea:
Clicking a pin twice on /around zooms in, and this PR ensures the back button and 'back to all reports' link zooms the map back to where you were initially.

Fixes #1649.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -994,30 +994,35 @@ $.extend(fixmystreet.set_up, {
     });
 
     $('#map_sidebar').on('click', '.js-back-to-report-list', function(e) {
-        if (e.metaKey || e.ctrlKey) {
-            return;
-        }
-
-        e.preventDefault();
-        var reportListUrl = $(this).attr('href');
-        fixmystreet.display.reports_list(reportListUrl, function() {
-            // Since this navigation was the result of a user action,
-            // we want to record the navigation as a state, so the user
-            // can return to it later using their Back button.
-            if ('pushState' in history) {
-                history.pushState({
-                    initial: true,
-                    mapState: around_map_state
-                }, null, reportListUrl);
-            }
-            if (around_map_state) {
-                fixmystreet.maps.set_map_state(around_map_state);
-            }
-        });
+        var report_list_url = $(this).attr('href');
+        var map_state = around_map_state;
+        var set_map_state = true;
+        fixmystreet.back_to_reports_list(e, report_list_url, map_state, set_map_state);
     });
   }
 
 });
+
+fixmystreet.back_to_reports_list = function(e, report_list_url, map_state, set_map_state) {
+    if (e.metaKey || e.ctrlKey) {
+        return;
+    }
+    e.preventDefault();
+    fixmystreet.display.reports_list(report_list_url, function() {
+        // Since this navigation was the result of a user action,
+        // we want to record the navigation as a state, so the user
+        // can return to it later using their Back button.
+        if ('pushState' in history) {
+            history.pushState({
+                initial: true,
+                mapState: map_state
+            }, null, report_list_url);
+        }
+        if (set_map_state && map_state) {
+            fixmystreet.maps.set_map_state(map_state);
+        }
+    });
+};
 
 fixmystreet.update_report_a_problem_btn = function() {
     var zoom = fixmystreet.map.getZoom();
@@ -1313,10 +1318,14 @@ fixmystreet.display = {
             // wrong (but what is shown is correct).
             $('.js-back-to-report-list').attr('href', fixmystreet.original.href);
 
-            // Problems nearby should act the same as 'Back to all reports' on around,
-            // but on /my and /reports should go to that around page.
-            if (fixmystreet.original.page == 'around') {
-                $sideReport.find('#key-tool-problems-nearby').addClass('js-back-to-report-list');
+            // Problems nearby on /my should go to the around page,
+            // otherwise show reports within the current map view.
+            if (fixmystreet.original.page === 'around' || fixmystreet.original.page === 'reports') {
+                $sideReport.find('#key-tool-problems-nearby').click(function(e) {
+                    var report_list_url = fixmystreet.original.href;
+                    var map_state = fixmystreet.maps.get_map_state();
+                    fixmystreet.back_to_reports_list(e, report_list_url, map_state);
+                });
             }
             fixmystreet.set_up.map_sidebar_key_tools();
             fixmystreet.set_up.form_validation();
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -935,6 +935,8 @@ $.extend(fixmystreet.set_up, {
   },
 
   ajax_history: function() {
+    var around_map_state = null;
+
     $('#map_sidebar').on('click', '.item-list--reports a', function(e) {
         if (e.metaKey || e.ctrlKey) {
             return;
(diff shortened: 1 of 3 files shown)


dracos:
Haven't tested this yet, but cursory reading looks good - one thing that might be a bug is `/around` -> click report -> click report again -> click different nearby report -> click 'back to all reports', I think `around_map_state` might not have the right data in then?

davea:
Good catch! I think ba5c5de152180aaac74a79c1a30993aa962e17b6 sorts that out.

dracos:
Afraid not, I think that's broken it entirely on an `/around` page, sorry.

dracos:
And saying something more useful, if you look at `fixmystreet.original` that is used for similar situation, setting something at the start to be remembered when you finally get back there (title, sub map links) which I think might work here too?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -890,18 +892,27 @@ $.extend(fixmystreet.set_up, {
...
+                history.pushState({
+                    reportId: reportId,
+                    reportPageUrl: reportPageUrl,
+                    mapState: fixmystreet.maps.get_map_state()
+                }, null);
             }
             return;
         }
 
+        if (fixmystreet.page === 'reports' || fixmystreet.page === 'around') {
```
dracos: Any reason not to have `my` here too? I imagine the same behaviour works the same there too.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -890,18 +892,27 @@ $.extend(fixmystreet.set_up, {
...
         fixmystreet.display.report(reportPageUrl, reportId, function() {
             // Since this navigation was the result of a user action,
             // we want to record the navigation as a state, so the user
             // can return to it later using their Back button.
             if ('pushState' in history) {
                 history.pushState({
                     reportId: reportId,
-                    reportPageUrl: reportPageUrl
+                    reportPageUrl: reportPageUrl,
+                    mapState: around_map_state
```
dracos: I think this should be fixmystreet.maps.get_map_state() so that further pin clicks don't go back to the first map state.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -919,7 +930,13 @@ $.extend(fixmystreet.set_up, {
...
             // can return to it later using their Back button.
             if ('pushState' in history) {
-                history.pushState({ initial: true }, null, reportListUrl);
+                history.pushState({
+                    initial: true,
+                    mapState: around_map_state
+                }, null, reportListUrl);
+            }
+            if (around_map_state) {
+                fixmystreet.maps.set_map_state(around_map_state);
```
dracos: Do we want to do this? This makes "Back to all reports" different to clicking Back (link resets the view, clicking Back does not) - and I wonder when you click "Back to all reports" do you mean "Back to the list of reports I was looking at" or "Back to the view of all reports". I think I lean towards the latter, so if you've scrolled the map when looking at individual reports, clicking "Back to all reports" would show me all the reports around where I now am on the map. Ooh and especially as "Problems nearby" does the same thing as "Back to all reports" and that definitely carries weight of it being map-view-currently-seen.

web/cobrands/fixmystreet/fixmystreet.js
davea: Yep, good catch.

web/cobrands/fixmystreet/fixmystreet.js
davea: Oops, yes :)

web/cobrands/fixmystreet/fixmystreet.js
davea: Hmm, tricky. I can see advantages to both. I think I'd expect it to zoom back out and give an overview of all reports, based on the link text. Maybe it should do something different on `/reports` vs `/around`? On `/reports` it could restore the original map view which shows all reports, and `/around` would leave the current zoom level intact and show a list of nearby reports. (Let's ignore the third behaviour of "zoom in and show nearby reports" when clicked from a directly-loaded `/report` page!)

dracos: I should note Google Maps invariably doesn't do what I want in this area, always seems to be messing up going back stuff :) I think I'd rather avoid different behaviour between pages, unless we're pretty sure that's better for some reason, I can see both being wanted on `/reports` too.

I think I'm happy for it to do what you suggest for `/reports` everywhere, but only if the Back button does the same thing (you are going Back to the previous view, as it were). (I assume your other PR on using replaceState might affect this too, will clicking Back go to the first/last replacedState thing...), so that would mean having the initial state save also save the map position so it can be used when going Back there. And if "problems nearby" then always did the `/around` behaviour you suggest, as that seems clear, and lets people (should they manage to work it out :) ) choose what they want.

davea: From our conversation on Slack:

> * Clicking "Back to all reports" or clicking Back (at the point that would be going back to all reports IYSWIM, not to another report): should zoom out and reset location to as it was. This means having the initial state save also save the map position so it can be used when clicking Back;
> * "Problems nearby" should NOT move map, but stay where you are, presumably also keeping the same zoom.

I've implemented these changes.

davea:
This is ready for another review at last!

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1218,10 +1247,19 @@ fixmystreet.display = {
...
             $('.js-back-to-report-list').attr('href', fixmystreet.original.href);
 
-            // Problems nearby should act the same as 'Back to all reports' on around,
-            // but on /my and /reports should go to that around page.
-            if (fixmystreet.original.page == 'around') {
-                $sideReport.find('#key-tool-problems-nearby').addClass('js-back-to-report-list');
+            // Problems nearby on /my should go to the around page,
+            // otherwise show reports within the current map view.
+            if (fixmystreet.original.page === 'around' || fixmystreet.original.page === 'reports') {
+                $sideReport.find('#key-tool-problems-nearby').click(function(e) {
```
dracos: Need a meta/ctrl key check here.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1218,10 +1247,19 @@ fixmystreet.display = {
...
+                    fixmystreet.display.reports_list(fixmystreet.original.href, function() {
```
dracos: Interestingly, going direct to a report page, the "Problems nearby" link becomes 'live' and updates as you pan the map, then taking you to where you are. This doesn't otherwise happen, but perhaps it should, I imagine that would be nice to have here too (but later).

web/cobrands/fixmystreet/fixmystreet.js
dracos: In fact, this function is still the same as the click handler for `'.js-back-to-report-list`, just presumably without the map state setting (if I've read it right). So might be better to factor those together?
(Also, do you still need to set `initial:true` so that the relevant popstate code is run if this page is later gone Back to; would also be fixed by factoring the two together.)

davea: Well spotted, that's made things simpler.

davea:
@dracos Think we're there now...

PR #2301, opened 2018-10-19
Track map state in URL for easier link sharing

### LINKED ISSUE #2242: Update the map URL parameters as you pan and zoom
Is there a reason we don’t do this?

With the History API, we could fairly easily update the `/around` and `/area/<slug>` URL parameters when you pan and zoom around the map, without creating new browser History states.

This would let people copy the currently visible URL to share it, rather than them having to spot the expertly hidden "Permalink" button in the little black toolbar in the bottom right corner of the map.

This feature would be particularly useful on collideoscope.org.uk where one of the main use-cases of the area maps is campaigners sharing / linking to maps of incidents in a particular area.

davea:
- [x] /around pages
 - [x] /reports pages

Fixes #2242.


web/js/map-OpenLayers.js
```diff
@@ -925,7 +935,21 @@ OpenLayers.Control.PermalinkFMS = OpenLayers.Class(OpenLayers.Control.Permalink,
...
+        // with commas here, which will then be urlencoded correctly by
+        // getParameterString.
+        if (OpenLayers.Util.isArray(params.status)) {
+            params.status = params.status.join(",");
+        }
+        if (OpenLayers.Util.isArray(params.filter_category)) {
+            params.filter_category = params.filter_category.join(",");
+        }
+
+        href += separator + OpenLayers.Util.getParameterString(params);
```
dracos: Now I notice this puts an awful lot of layers in the URL (due to all the hidden road ones) - should we strip that out?

web/js/map-OpenLayers.js
```diff
@@ -925,7 +935,21 @@ OpenLayers.Control.PermalinkFMS = OpenLayers.Class(OpenLayers.Control.Permalink,
         if ( alter_zoom ) {
             zoom += fixmystreet.zoomOffset;
         }
-        href += separator + OpenLayers.Util.getParameterString(this.createParams(center, zoom));
+
+        var params = this.createParams(center, zoom);
+        // Roundtripping a param containing urlencoded commas through
+        // createParams and getParameterString results in the commas
+        // becoming un-urlencoded, so let's avoid this by joining the arrays
```
dracos: What's the problem being solved here, why does it matter if they're un-urlencoded? I'm presumably missing something.

web/js/map-OpenLayers.js
```diff
@@ -680,6 +680,12 @@ $.extend(fixmystreet.utils, {
             if (fixmystreet.map.popups.length) {
                 fixmystreet.map.removePopup(fixmystreet.map.popups[0]);
             }
+
+            // Ensures the permalink control is updated when the filters change
+            var permalink_controls = fixmystreet.map.getControlsByClass(/Permalink/);
```
dracos: This only happens when the pins have loaded from the server, though - could you do this at the point the filters are changed/pin server request starts?

web/js/map-OpenLayers.js
```diff
@@ -567,8 +567,8 @@ $.extend(fixmystreet.utils, {
                     f.geometry = new_geometry;
                     this.removeAllFeatures();
                     this.addFeatures([f]);
-                    var qs = fixmystreet.utils.parse_query_string();
-                    if (!qs.bbox) {
+                    var qs = OpenLayers.Util.getParameters(fixmystreet.original.href);
+                    if (!qs.bbox && !qs.lat && !qs.lon) {
```
dracos: I understand the lat/lon addition, but why check the original href and not what it is now?

perllib/FixMyStreet/Map/OSM.pm
```diff
@@ -57,6 +57,7 @@ sub display_map {
         if defined $c->get_param('lat');
     $params{longitude} = Utils::truncate_coordinate($c->get_param('lon') + 0)
         if defined $c->get_param('lon');
+    $params{zoomToBounds} = !defined $c->get_param('zoom');
```
dracos: This can be a bit odd on an around page - whilst if `js=1` the pins are presumably within the viewport and so zooming to their extent doesn't do anything, it's not needed, and if `js=0` pins might be out of the viewport so your map ends up centred somewhere else (otoh, better for the pins loaded, but bit odd if you're expecting a particular centre). Perhaps this should be if `any_zoom` *and* not defined zoom parameter? Or was there a reason you removed `any_zoom` here.

(Aside: I think there's some long-standing interaction with OL's ArgParser that is looking for lat/lon in the URL and updating map centre if it sees them (defaulting to zoom 0), so e.g. `/around?pc=foo` will use default calculated zoom level, but `/around?lat=same&lon=location-as-foo` will go to zoom 0. Hmm.)

web/js/map-OpenLayers.js
dracos: Should it remove the `pc` parameter if there's a lat/lon, assuming those take precedence.

davea: Good idea - although will removing `pc` prevent the "user originally entered" text being captured if they make a report from that page?

web/js/map-OpenLayers.js
davea: It's not solving a problem as such, it was to get the frontend tests to pass. Specifically [`cy.url().should('include', 'status=closed%2Cfixed');`](https://github.com/mysociety/fixmystreet/blob/6e2c77b2e1f3f693d806a9c0237403e3273378af/.cypress/cypress/integration/around_filters.js#L22). Not sure if there's a better way, so I just went with consistency.

web/js/map-OpenLayers.js
davea: 👍

web/js/map-OpenLayers.js
davea: At this point I have totally forgotten my reasoning 😞 I think it was to deal with odd behaviour on `/reports`, but can't replicate it so will go back to `fixmystreet.utils.parse_query_string`.

web/js/map-OpenLayers.js
dracos: It's also in a hidden field on the page, so I think it'll still get remembered through the form submission.

davea:
Thanks for the feedback - I think I've addressed everything!

web/js/map-OpenLayers.js
dracos: The Cypress failure now is because the map is no longer zooming out on a `/reports` page once the shapefile loads, because the URL has already updated by then to include location parameters - presumably that was the reason for this! It feels like there should be a better way of doing this, but I'm not sure what.

web/js/map-OpenLayers.js
dracos: I think we can update the test :) Guess slight concern as to what 'runs' first, perhaps, though? Or can we know that for sure.

dracos: Ah I think the issue might be the call to updateLink being in `categories_or_status_changed` (called on change and on history navigation) rather than `categories_or_status_changed_history` (only called on actual change) - we don't need to update the URL when clicking back/fwd as it's already done for us. I think?

web/js/map-OpenLayers.js
```diff
@@ -913,6 +919,14 @@ OpenLayers.Control.PanZoomFMS = OpenLayers.Class(OpenLayers.Control.PanZoom, {
 /* Overriding Permalink so that it can pass the correct zoom to OSM */
 OpenLayers.Control.PermalinkFMS = OpenLayers.Class(OpenLayers.Control.Permalink, {
     _updateLink: function(alter_zoom) {
+        // this.base was originally set in initialize(), but the window's href
+        // may have changed since then if e.g. the map filters have been updated.
+        // NB this won't change the base of the 'problems nearby' permalink on
+        // /report, as this would result in it pointing at the wrong page.
+        if (this.base !== '/around' && fixmystreet.page !== 'report') {
```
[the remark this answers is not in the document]
struan: So, thinking a bit about this, should we even be in this function if we are on a report page? All it's going to do is add cruft to the permalink url.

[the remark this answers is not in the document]
dracos: You *might* want to permalink an offset pin? e.g. https://www.fixmystreet.com/report/1234567?zoom=2&lat=51.5588&lon=-1.14649&layers=BT But I agree this seems pretty unlikely.

PR #2305, opened 2018-10-23
Fix "two taps to show report" behaviour on iOS

### LINKED ISSUE #2294: Two touches necessary to open a report
### Description of problem

Perhaps this is intentional, but it takes two touches (referred to as "clicks" below) in order to open a report from the reports page on iPhone 6 (iOS 12) (but it may be the case on any kind of phone).

### Version

2.3.5.

### Steps to reproduce

1. Visit https://fixamingata.se
2. Click on Alla rapporter
3. Search for Ale kommun under Lokala rapporter and press Gå
4. Click on a report in the list
5. Click on the same report in the list

### Actual results

Step 4 above changes the background on the list item, but does not open the report.

### Expected results

Step 4 would open the report (making Step 5 irrelevant).

### How reproducible

Always.

davea:
iOS will activate the hover state, if present, the first time an element is tapped, meaning two taps are required to perform an action like following a link.

This PR moves the hover styles to the desktop-only stylesheet so iOS never even knows they’re there, and disables the JS that highlights pins when hovering over the report list on mobile.

Some concerns:

 - Still requires two taps on iPads (and iPhones in landscape)
 - This lift-and-shift of the styles doesn’t seem like a great solution but my Sassiness isn’t sassy enough to know if there’s a better way.
 - There are probably cobrand `base.scss` files with `:hover` styles too

Fixes #2294.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
         - Fix subcategory issues when visiting /report/new directly #2276
         - Give superusers access to update staff dropdowns. #2286
         - Update report areas when moving its location. #2181
+        - Don't require two taps on reports in list on iOS. #2294
     - Development improvements:
         - Add cobrand hook for dashboard viewing permission. #2285
         - Have body.url work in hashref lookup. #2284
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1201,6 +1201,9 @@ fixmystreet.display = {
                 $sideReport.appendTo('#map_sidebar');
             }
             $('#map_sidebar').scrollTop(0);
+            if ($("html").hasClass("mobile")) {
+                $(document).scrollTop(0);
+            }
 
             var found = html.match(/<title>([\s\S]*?)<\/title>/);
             var page_title = found[1];
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -180,7 +180,6 @@ a {
     color: #0BA7D1;
   }
 
-  &:hover,
   &:active {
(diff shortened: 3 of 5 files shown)


davea:
You're right, it's not a very neat solution is it. A fresh look at this reveals that actually none of the CSS changes are required to solve the original bug after all - the JS change in 639b8a3a38484226e729b49db98f547d2546e048 is sufficient!

I've reverted the CSS changes which makes for a much simpler diff - what do you think?

zarino:
@davea It’s good to see the CSS changes weren’t necessary.

But this branch still conflates "touch" with "small screen". It bypasses the problem for narrow-windowed devices (`(min-width: 48em)` according to the Modernizr line in `web/cobrands/fixmystreet/header.js` rather than for touch devices. Browsing FMS on my iPad, I still have to double-tap reports in the [sidebar of a page like this](https://www.fixmystreet.com/reports/Oxfordshire/), because my iPad is wide enough to no longer have a class of `mobile` on the `html` element.

It feels like we need to be avoiding "am I mobile?"-type questions, and checking, instead, "am I a touch event?". I wonder whether we could do something like bind to the `touchstart` event, and whether devices would then execute _that_ callback, rather than the `mouseenter` callback, on first tap?

dracos:
Don't know if this is helpful, but in case it is, my commit de36c49d1ff2bf7858c8cc8762c4c7b9d05c9cb3 had some info in the commit message about touch events on the map. If what I say there is right, pins do operate on touchstart.

davea:
You're right, that solution isn't ideal for a number of reasons. 762df434a76c1f95ccbdfde8d945db93530bfe39 does things properly, with touchevent detection for that `mouseenter` handler. Works as expected on iPad!

PR #2306, opened 2018-10-24
add report_mark_private permission

struan:
Allows user's to see the inspector panel to mark reports as Private, and
also to view those non-public reports. Useful for call centre staff who
want to record private reports but don't need to other permissions.

Fixes mysociety/fixmystreet-commercial#1213

Please check the following:

### DIFF
--- a/docs/_includes/admin-tasks-content.md
+++ b/docs/_includes/admin-tasks-content.md
@@ -250,6 +250,29 @@ were made during the period when the user was banned — these remain unsent.
 
 </div>
 
+<div class="admin-task" markdown="1" id="create-reports-private">
+
+### Creating/Viewing private reports
+
+<span class="admin-task__permissions">Permissions required: User must be marked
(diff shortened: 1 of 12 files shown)


docs/_includes/admin-tasks-content.md
```diff
@@ -250,6 +250,26 @@ were made during the period when the user was banned — these remain unsent.
...
+<div class="admin-task" markdown="1" id="create-reports-private">
+
+### Creating/Viewing private reports
+
+<span class="admin-task__permissions">Permissions required: User must be marked
+as staff; one or more of ‘View/Mark private reports’ and ‘Markup problem
+details’ must be ticked.</span>
+
+If a resident makes a report that has to contain information that should
+not be make public, e.g. Names and addresses, then you can create a
```
dracos: This feels like it should say you can convert the report to a private one (it reads like the resident makes the report on the site, I mean). Then next paragraph could talk about a new report as it does.

PR #2310, opened 2018-10-25
Disable questionnaires when closing/fixing reports via Open311

davea:
Stops questionnaires being sent for reports closed/fixed via Open311.

Not sure if we want this behaviour for situations where a report has been closed/fixed by an admin/council staff user too?

Connects to https://github.com/mysociety/fixmystreet-freshdesk/issues/24.

Please check the following:


davea:
All good points! I must have imagined the fixed requirement, sorry about that. Will force-push a much simpler change shortly.

PR #2311, opened 2018-10-26
send reports on highways agency roads to highways agency

### LINKED ISSUE #736: Report road-related issues on Highways Agency managed roads to the Highways Agency
<a href="http://www.dracos.co.uk/temp/scrshot-2014.03.28-22:12:10.png">Bit of the M61</a>

(This is one particular body that is mentioned in #16)

A few people have written over the years asking about FixMyStreet sending reports about Highways Agency managed roads to the Highways Agency rather than the council covering the point. 

[This WhatDoTheyKnow request](https://www.whatdotheyknow.com/request/highways_agency_managed_roads?unfold=1) contains a shapefile of Highways Agency roads.

The dataset is all line strings (so no width), so I guess you might have to do something when the map is clicked and then perhaps add a road choice to the category (or separately) if there's one nearby. Or just make an assumption if it's a road category  and close enough to a line.

struan:
Includes an option to send to the council instead for e.g. reports on
underpasses or bridges.

Fixes #736

Also demonstrates how mysociety/fixmystreet-commercial#1191 can be done

Please check the following:

screen recording 2018-10-30 at 14 02 21

### DIFF
--- a/templates/web/fixmystreet.com/footer_extra_js.html
+++ b/templates/web/fixmystreet.com/footer_extra_js.html
@@ -10,6 +10,7 @@ IF bodyclass.match('mappage');
     scripts.push( version('/cobrands/buckinghamshire/assets.js') );
     scripts.push( version('/cobrands/lincolnshire/assets.js') );
     scripts.push( version('/cobrands/oxfordshire/assets.js') );
+    scripts.push( version('/cobrands/highways/assets.js') );
     scripts.push(
         version('/vendor/OpenLayers.Projection.OrdnanceSurvey.js'),
     );
--- a/web/cobrands/buckinghamshire/assets.js
+++ b/web/cobrands/buckinghamshire/assets.js
@@ -156,6 +156,18 @@ function is_only_body(body) {
     return false;
 }
 
+$(fixmystreet).on('report_new:highways_change', function() {
+    if (fixmystreet.body_overrides.get_only_send() === 'Highways England') {
+        hide_responsibility_errors();
+        enable_report_form();
+        $('#bucks_dangerous_msg').hide();
+    } else {
+        $('#bucks_dangerous_msg').show();
+        $(fixmystreet).trigger('report_new:category_change', [ $('#form_category') ]);
+    }
+});
+
+
 fixmystreet.assets.add($.extend(true, {}, defaults, {
     http_options: {
         params: {
--- a/web/cobrands/highways/assets.js
+++ b/web/cobrands/highways/assets.js
@@ -68,7 +68,9 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
         },
         not_found: function(layer) {
             fixmystreet.body_overrides.location = null;
-            fixmystreet.body_overrides.remove_only_send();
+            if (fixmystreet.body_overrides.get_only_send() === 'Highways England') {
+                fixmystreet.body_overrides.remove_only_send();
+            }
             $('#highways').remove();
         }
     }
@@ -87,6 +89,7 @@ function add_highways_warning(road_name) {
         .on('click', function() {
             fixmystreet.body_overrides.location = null;
             fixmystreet.body_overrides.only_send('Highways England');
+            $(fixmystreet).trigger('report_new:highways_change');
         })
         .appendTo($radios);
     $('<label>')
@@ -104,6 +107,7 @@ function add_highways_warning(road_name) {
                 longitude: $('#fixmystreet\\.longitude').val()
             };
             fixmystreet.body_overrides.remove_only_send();
+            $(fixmystreet).trigger('report_new:highways_change');
         })
         .appendTo($radios);
     $('<label>')
@@ -113,6 +117,9 @@ function add_highways_warning(road_name) {
         .appendTo($radios);
     $radios.appendTo($warning);
     $('.change_location').after($warning);
+    fixmystreet.body_overrides.location = null;
+    fixmystreet.body_overrides.only_send('Highways England');
+    $(fixmystreet).trigger('report_new:highways_change');
 }
 
 })();
--- a/web/cobrands/highways/assets.js
+++ b/web/cobrands/highways/assets.js
@@ -20,9 +20,7 @@ var defaults = {
     // the M6
     max_resolution: 20,
     min_resolution: 0.5971642833948135,
-    asset_id_field: 'CENTRAL_AS',
-    geometryName: 'msGeometry',
-    srsName: "EPSG:3857",
+    srsName: "EPSG:900913",
     strategy_class: OpenLayers.Strategy.FixMyStreet
 };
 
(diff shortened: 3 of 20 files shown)


web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,112 @@
...
+        params: {
+            SERVICE: "WFS",
+            VERSION: "1.1.0",
+            REQUEST: "GetFeature",
+            SRSNAME: "urn:ogc:def:crs:EPSG::3857"
+        }
+    },
+    format_class: OpenLayers.Format.GML.v3.MultiCurveFix,
+    asset_type: 'area',
+    max_resolution: 4.777314267158508,
```
dracos: I assume this should be higher, it's not activating at very zoomed out levels.

web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+}));
+
+function add_highways_warning(road_name) {
+  var $warning = $('<div class="box-warning" id="highways"><p>It looks like you clicked on the ' + road_name + ' which is managed by <strong>Highways England</strong>.<p></div>');
+    $('<a>')
+        .attr('href', '#')
+        .attr('id', 'js-not-highways')
+        .html('Send to <strong>' + fixmystreet.bodies.join('</strong> ' + translation_strings.or + ' <strong>') + '</strong> instead.')
```
dracos: I think the wording here (and related) needs to help the user know whether they should click it or not. I think making it a radio question would make it clearer perhaps? It could say something like:
"You’ve clicked on the A46, managed by Highways England.
Does your report actually concern something on this road, or somewhere else (e.g. a road crossing it)?
(x) On the A46 ( ) Somewhere else"

web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,112 @@
...
+    always_visible: true,
+
+    non_interactive: true,
+    asset_type: 'area',
+    road: true,
+    all_categories: true,
+    nearest_radius: 15,
+    actions: {
+        found: function(layer, feature) {
+            // this is to stop this firing when we change category
```
dracos: I guess his wouldn't be needed if you had a radio question always shown.

web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+    },
+    stylemap: highways_stylemap,
+    always_visible: true,
+
+    non_interactive: true,
+    asset_type: 'area',
+    road: true,
+    all_categories: true,
+    nearest_radius: 15,
```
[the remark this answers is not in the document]
struan: Right, I based it on whether clicking in the middle of a couple of Motorways worked. I did wonder if trying to see if the click was within an area would be the best way to deal with it but that seems like a lot of complexity to add if we can avoid it.

templates/web/fixmystreet.com/footer_extra_js.html
```diff
@@ -10,6 +10,7 @@
     scripts.push( version('/cobrands/buckinghamshire/assets.js') );
     scripts.push( version('/cobrands/lincolnshire/assets.js') );
     scripts.push( version('/cobrands/oxfordshire/assets.js') );
+    scripts.push( version('/cobrands/highways/assets.js') );
```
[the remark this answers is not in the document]
struan: I tested it on a few and it worked. I need to check them all to be sure but certainly it worked on Bathnes, Bristol and OCC.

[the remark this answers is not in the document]
dracos: Should the cobrands be including this JS file then, I mean?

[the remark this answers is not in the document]
struan: ah yes, I had done that locally but not pushed the change :(

web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,118 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+var defaults = {
+    http_options: {
+        url: "https://struan.tilma.dev.mysociety.org/mapserver/highways",
```
dracos: Will need putting on live presumably :)

web/cobrands/buckinghamshire/assets.js
```diff
@@ -200,7 +215,10 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
             // unless an asset is selected.
             fixmystreet.body_overrides.do_not_send(layer.fixmystreet.body);
             fixmystreet.body_overrides.remove_only_send();
-            if (fixmystreet.assets.selectedFeature()) {
+            if (fixmystreet.body_overrides.get_only_send() === 'Highways England') {
```
dracos: This never gets called because the only send was just removed the line above. I think it's then covered by the new event, anyway. Should it just return at the start, like the `found` function? I /think/ they can be consistent.

web/cobrands/highways/assets.js
```diff
@@ -0,0 +1,121 @@
...
+            }
+            $('#highways').remove();
+            if ( !fixmystreet.assets.selectedFeature() ) {
+                fixmystreet.body_overrides.only_send('Highways England');
+                add_highways_warning(feature.attributes.ROA_NUMBER);
+            }
+        },
+        not_found: function(layer) {
+            fixmystreet.body_overrides.location = null;
+            fixmystreet.body_overrides.remove_only_send();
```
dracos: When this runs after a Bucks road has been selected with Flytipping already picked, the only-send-to-Bucks is lost because it is always removed by this function. I think this needs to not do this line if only_send is already set to something other than Highways England, does that work?

web/cobrands/buckinghamshire/assets.js
```diff
@@ -172,6 +184,9 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
     all_categories: true,
     actions: {
         found: function(layer, feature) {
+            if (fixmystreet.body_overrides.get_only_send() === 'Highways England') {
```
dracos: This causes a delay upon e.g. clicking from a motorway to a non-Bucks maintained road because this section fires first time (so nothing changes), highways layer removes the only_send, then when the server responds with updated point information it then skips this bit and disables the form. Perhaps this if isn't needed?

dracos: And similarly if clicking from motorway to a Bucks road - only after the server response is it set up for e.g. Bucks-only for flytipping.

web/cobrands/buckinghamshire/assets.js
dracos: Or just removed, if I'm saying it can be consistent with `found` and my later comment there.

struan:
Sometimes cypress seems to get confused and think that the button is behind another element and the `force` gets round that.

PR #2312, opened 2018-10-26
[FixMyStreet.com] Enable category groups

davea:
- Fixes a bug that meant items in top-level category dropdown were out of order if there was a mix of grouped and ungrouped categories
 - There are existing Open311 contacts (Bristol, Greenwich) in the DB which have unhelpful groups applied, presumably because they weren't intended to be seen publicly.
 - This PR inspects the cobrand's `enable_category_groups` value when populating Open311 categories and only sets the contact's `group` metadata if the handling cobrand has groups enabled.

Connects to mysociety/fixmystreet-commercial#1148.

### DIFF
--- a/perllib/Open311/PopulateServiceList.pm
+++ b/perllib/Open311/PopulateServiceList.pm
@@ -11,6 +11,9 @@ has schema => ( is => 'ro', lazy => 1, default => sub { FixMyStreet::DB->schema-
 has _current_body => ( is => 'rw' );
 has _current_open311 => ( is => 'rw' );
 has _current_service => ( is => 'rw' );
+has _current_body_cobrand => ( is => 'ro', lazy => 1, default => sub {
+    return shift->_current_body->get_cobrand_handler;
+} );
 
 sub process_bodies {
     my $self = shift;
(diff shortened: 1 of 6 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -506,6 +506,13 @@ $.extend(fixmystreet.set_up, {
     });
     // Sort elements in place, but leave the first 'pick a category' option alone
     $group_select.find("option").slice(1).sort(function(a, b) {
+        // 'Other' should always be at the end.
+        if (a.label === 'Other') {
```
dracos: Just to note translation issue here (probably could stick a class on the Other `<option>` when output, or add the string to translations, hmm), but think also probably okay to leave for the moment.

perllib/Open311/PopulateServiceList.pm
```diff
@@ -19,6 +22,7 @@ sub process_bodies {
         next unless $body->endpoint;
         next unless lc($body->send_method) eq 'open311';
         $self->_current_body( $body );
+        $self->_current_body_cobrand( $body->get_cobrand_handler );
```
dracos: Do you need this line any more given the change to default in the class?

davea: Nope, good spot!

PR #2313, opened 2018-10-30
allow comment fetching to be limited by body name

struan:
Allows creating scripts that fetch comments for a single body, e.g for
batch updating or because they require special setup.

Please check the following:

### DIFF
--- a/bin/fixmystreet.com/backfill-open311-comments
+++ b/bin/fixmystreet.com/backfill-open311-comments
@@ -1,0 +1,64 @@
+#!/usr/bin/env perl
+#
+# This script utilises the Open311 extension explained at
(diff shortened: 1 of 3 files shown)


bin/fixmystreet.com/backfill-open311-comments
```diff
@@ -0,0 +1,62 @@
...
+use Open311::GetServiceRequestUpdates;
+
+my $start_date = DateTime::Format::W3CDTF->parse_datetime( $opt->start_date );
+my $end_date = DateTime::Format::W3CDTF->parse_datetime( $opt->end_date );
+
+die "start and end date required\n" unless $start_date and $end_date;
+
+my $current_date = $start_date->clone;
+
+while ( $current_date < $end_date ) {
```
dracos: Could do with a check that start is less than end, otherwise this will never end.

PR #2316, opened 2018-11-01
more granular staff permissions

struan:
This adds `report_prefill` and `default_to_body` permissions for staff which split out some of the report_inspect behaviours into separate permissions.

Staff with the `report_prefill`/ permission will have the report title and detail auto populated with some default text.

Staff with the `default_to_body` permission will, assuming they also have the `can_contribute_as_body` permission, default to reporting new reports and updates as the body rather than themselves.

Please check the following:

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1218
Connects https://github.com/mysociety/fixmystreet-freshdesk/issues/23

### DIFF
--- a/t/app/controller/report_new.t
+++ b/t/app/controller/report_new.t
@@ -2048,8 +2048,31 @@ subtest "check map click ajax response for inspector" => sub {
     }, sub {
         $extra_details = $mech->get_ok_json( '/report/new/ajax?latitude=55.952055&longitude=-3.189579' );
     };
-    like $extra_details->{category}, qr/data-role="inspector/, 'category has correct data-role';
+    like $extra_details->{category}, qr/data-prefill="0/, 'inspector prefill not set';
     ok !$extra_details->{contribute_as}, 'no contribute as section';
+};
+
+subtest "check map click ajax response for inspector and uk cobrand" => sub {
+    $mech->log_out_ok;
+
+    my $extra_details;
+    $inspector->user_body_permissions->find_or_create({
+        body => $bodies[4],
+        permission_type => 'planned_reports',
+    });
+    $inspector->user_body_permissions->find_or_create({
+        body => $bodies[4],
+        permission_type => 'report_inspect',
+    });
+
+    $mech->log_in_ok('inspector@example.org');
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => [ { bromley => '.' } ],
+        MAPIT_URL => 'http://mapit.uk/',
+    }, sub {
+        $extra_details = $mech->get_ok_json( '/report/new/ajax?latitude=51.402096&longitude=0.015784' );
+    };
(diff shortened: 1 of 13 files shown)


dracos:
Also, PR description says "This also adds a nearest location lookup to the report/new/ajax call which is used to add address data to the auto populated text." but I can't see that in the code changes at all. (I'd be worried about speed of ajax call if that was always done, should it be a separate parallel ajax call?)

web/cobrands/fixmystreet/staff.js
```diff
@@ -456,7 +456,7 @@ $(fixmystreet).on('report_new:category_change', function(evt, $this) {
     var role = $this.data('role');
```
dracos: Can remove this now (and the setting of data-role in the HTML, given this was only thing using it).

templates/web/base/report/new/category.html
```diff
@@ -15,7 +15,8 @@
     </label>[% =%]
     <select class="form-control[% IF category_groups.size %] js-grouped-select[% END %]" name="category" id="form_category"
     [%~ IF c.user.from_body =%]
-      data-role="[% c.user.has_body_permission_to('planned_reports') ? 'inspector' : 'user'  %]" data-body="[% c.user.from_body.name %]" data-prefill="[% c.cobrand.prefill_report_fields_for_inspector %]"
+      [%~ prefill_report = c.cobrand.prefill_report_fields_for_inspector || c.user.has_body_permission_to('report_prefill') %]
```
dracos: Current behaviour is it prefilled if `prefill_report_fields_for_inspector` (all UK council cobrands) and user was inspector. This behaviour is now either `prefill_report_fields_for_inspector` or permission, which doesn't match, it'll turn it on for everyone on the council cobrands.
I think it (and same in wrapper) needs to be `(c.cobrand.prefill_report_fields_for_inspector && inspector) || report_prefill` to match.

Ideally, perhaps not this PR, but it should give everyone with inspector permission the prefill permission (in a schema update) and then this line would only be the prefill permission, as you'd expect.

t/app/controller/report_new.t
```diff
@@ -1992,6 +1992,61 @@ subtest "check map click ajax response for inspector" => sub {
         $extra_details = $mech->get_ok_json( '/report/new/ajax?latitude=55.952055&longitude=-3.189579' );
     };
     like $extra_details->{category}, qr/data-role="inspector/, 'category has correct data-role';
+    like $extra_details->{category}, qr/data-prefill="0"/, 'inspector prefill not set';
```
dracos: Just in the interest of this already being a long test suite, I think these tests are covered by the JS test and so aren't needed.

templates/web/base/report/new/category.html
struan: Yup. I have put in a test to cover this as automated cobrand testing in the JS isn't really a thing.

PR #2319, opened 2018-11-05
Deal with possible multiple send_method_useds.

### LINKED ISSUE #2121: send_method_used method can be missed
Open311 send method calls `discard_changes` to get rid of local issues, but that also loses any change to send_method_used by a previous send method. Could either add all the send methods after the loop, only reset extra (that's all Open311 changes), or other. Also a couple of places still don't realise `send_method_used` could have multiple methods in it.

dracos:
Fixes #2121 

### DIFF
--- a/perllib/FixMyStreet/Script/Reports.pm
+++ b/perllib/FixMyStreet/Script/Reports.pm
@@ -211,12 +211,13 @@ sub send(;$) {
         # Multiply results together, so one success counts as a success.
         my $result = -1;
 
+        my @methods;
         for my $sender ( keys %reporters ) {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Script/Reports.pm
```diff
@@ -229,6 +230,10 @@ sub send(;$) {
                 if FixMyStreet->test_mode && $sender->can('open311_test_req_used');
         }
 
+        foreach (@methods) {
```
struan: Maybe worth a comment on why this is happening so it doesn't get optimised away again in the future.

PR #2325, opened 2018-11-12
Store remaining original moderation data

dracos:
Some new things are allowed to be moderated (category, location, extra) but these weren't being stored in original data.

Question: should this be bigger and store a history of changes rather than just the original?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
     - Admin improvements:
         - Allow moderation to potentially change category. #2320
         - Add Mark/View private reports permission #2306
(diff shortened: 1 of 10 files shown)


dracos:
Have added a commit to scroll up to top as it was annoying me this didn't happen, especially if Discard ticked - @zarino, look okay?

zarino:
@dracos I agree, it feels like it makes more sense when you scroll to top on entering/leaving the moderation mode. Otherwise you end up stranded halfway down the form, just because that’s where the buttons happen to be. :+1:

PR #2326, opened 2018-11-13
Photo storage updates

### LINKED ISSUE #2134: Strip EXIF data from report photos
An app user has emailed to suggest FixMyStreet removes EXIF tags from photos before they're made available for download from the site.

> Using exiftool (for example), I could see the make/model of my phone and various other settings. [...] I can also see the potential for finger-pointing, finger-printing a device or identifying a user posting an issue anonymously. [...] Best to strip (eg mogrify --strip) before publicly displaying.

We already run `jhead` on uploaded photos to correct the orientation, perhaps `mogrify` could be run subsequently. Alternatively, if we wanted to preserve the EXIF tags in the on-disk file, they could be stripped at the point `PhotoSet::get_image_data` generates the cached files.

dracos:
This enforces server-side the current client-side restriction on maximum photo size, plus also allows an instance to state that we can symlink full size photos from the photo cache rather than copy them there. Plus makes the ImageMagick wrapper a lot neater :) Also fixes #2134

### DIFF
--- a/perllib/FixMyStreet/App/Model/PhotoSet.pm
+++ b/perllib/FixMyStreet/App/Model/PhotoSet.pm
@@ -4,12 +4,6 @@ package FixMyStreet::App::Model::PhotoSet;
 
 use Moose;
 
-my $IM = eval {
-    require Image::Magick;
-    Image::Magick->import;
-    1;
-};
-
 use Scalar::Util 'openhandle', 'blessed';
 use Image::Size;
 use IPC::Cmd qw(can_run);
@@ -17,6 +11,7 @@ use IPC::Open3;
 use MIME::Base64;
 
 use FixMyStreet;
+use FixMyStreet::ImageMagick;
 use FixMyStreet::PhotoStorage;
 
 # Attached Catalyst app, if present, for feeding back errors during photo upload
@@ -221,14 +216,15 @@ sub get_image_data {
         return $image;
     }
 
+    my $im = FixMyStreet::ImageMagick->new(blob => $photo);
     if ( $size eq 'tn' ) {
-        $photo = _shrink( $photo, 'x100' );
+        $photo = $im->shrink('x100')->as_blob;
     } elsif ( $size eq 'fp' ) {
-        $photo = _crop( $photo );
+        $photo = $im->crop->as_blob;
     } elsif ( $size eq 'full' ) {
         # do nothing
     } else {
-        $photo = _shrink( $photo, $args{default} || '250x250' );
+        $photo = $im->shrink($args{default} || '250x250')->as_blob;
     }
 
     return {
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Model/PhotoSet.pm
```diff
@@ -71,6 +73,15 @@ has storage => (
     }
 );
 
+has symlinkable => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my $cfg = FixMyStreet->config('PHOTO_STORAGE_OPTIONS');
+        return $cfg ? $cfg->{SYMLINK_FULL_SIZE} : 0;
```
davea: This should be documented in `docs/customising/config.md`.

Should there be some way for the storage backend to indicate whether it supports symlinks? S3 doesn't, for example, so there might need to be some sanity checking of the config too.

perllib/FixMyStreet/PhotoStorage/FileSystem.pm
```diff
@@ -81,7 +81,7 @@ sub retrieve_photo {
     my $file = $self->get_file($fileid, $type);
     if ($file->exists) {
         my $photo = $file->slurp_raw;
-        return ($photo, $type);
+        return ($photo, $type, $file);
```
davea: Comment above needs updating to explain why this is being returned.

perllib/FixMyStreet/App/Model/PhotoSet.pm
```diff
@@ -203,6 +215,12 @@ sub get_image_data {
     my $photo = $image->{data};
 
     my $size = $args{size};
+
+    if ($self->symlinkable && $size eq 'full') {
```
davea: If this also checked if `$image->{object}` was defined then we could remove the `$object` from the return of `S3::retrieve_photo`. (Otherwise, if I'm understanding it correctly, the `symlink` call in `Photo.pm` would crash/fail?)

dracos: Yeah, and then that would also deal with your "storage backend to indicate whether it supports symlinks" too, good idea.

perllib/FixMyStreet/ImageMagick.pm
```diff
@@ -0,0 +1,68 @@
...
+    my ($self, $direction) = @_;
+    return $self unless $self->image;
+    my $err = $self->image->Rotate($direction);
+    return 0 if $err;
+    return $self;
+}
+
+# Shrinks a picture to the specified size, but keeping in proportion.
+sub shrink {
+    my ($self, $size) = shift;
```
davea: Typo:
```suggestion
    my ($self, $size) = @_;
```

This was a fun one to track down - nothing was breaking! The result was full-size images being output as e.g. `web/photo/1432427.0.jpeg` because no new size was being specified.

PR #2327, opened 2018-11-16
[Open311] Improve handling of created/updated datetimes of fetched reports

davea:
This should reduce the incidence of the ‘Problem id X for Y has an
invalid time, not creating’ cron errors we’ve been seeing.

See also mysociety/open311-adapter#26.

### DIFF
--- a/perllib/Open311/GetServiceRequests.pm
+++ b/perllib/Open311/GetServiceRequests.pm
@@ -122,16 +122,9 @@ sub create_problems {
                 if $self->verbose;
             next;
         }
-
         my $updated = DateTime::Format::W3CDTF->format_datetime(
             $updated_time->clone->set_time_zone('UTC')
         );
-        if ($args->{start_date} && $args->{end_date} && ($updated lt $args->{start_date} || $updated gt $args->{end_date}) ) {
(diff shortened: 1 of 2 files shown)


perllib/Open311/GetServiceRequests.pm
```diff
@@ -141,49 +134,64 @@ sub create_problems {
...
         my $problems;
         my $criteria = {
             external_id => $request_id,
         };
-        $problems = $self->schema->resultset('Problem')->to_body($body)->search( $criteria );
+
+        # Skip if this problem already exists (e.g. it may have originated from FMS and is being mirrored back!)
+        next if $self->schema->resultset('Problem')->to_body($body)->search( $criteria )->count;
+
+        if ($args->{start_date} && $args->{end_date} && ($created lt $args->{start_date} || $created gt $args->{end_date}) ) {
```
dracos: The dates given to this script are used, if I understand it right, to get all requests that have status changes (by LoggedTime in Confirm integration) between them. So should this be using `$updated` as it was previously, not `$created`? As I can see old reports being returned from this.

davea: Good point!

davea:
Thanks, think this is ready for re-review!

PR #2330, opened 2018-11-22
[Bristol] Use new WFS server for street lighting categories

davea:
BCC have moved their lighting assets to a new WFS URL.

[skip changelog]

Fixes mysociety/fixmystreet-freshdesk#18.

### DIFF
--- a/web/cobrands/bristol/assets.js
+++ b/web/cobrands/bristol/assets.js
@@ -4,25 +4,31 @@ if (!fixmystreet.maps) {
     return;
 }
 
-var options = {
-    wfs_url: "https://maps.bristol.gov.uk/arcgis/services/ext/FixMyStreetSupportData/MapServer/WFSServer",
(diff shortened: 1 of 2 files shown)


web/cobrands/bristol/assets.js
```diff
@@ -45,11 +45,42 @@ fixmystreet.assets.add($.extend({}, options, {
     filter_value: 'GRITBIN'
 }));
 
-fixmystreet.assets.add($.extend({}, options, {
-    asset_category: "Street Lighting",
+var lighting_options = $.extend({}, options, {
+    wfs_url: "https://maps.bristol.gov.uk/arcgis/services/ext/datagov/MapServer/WFSServer",
```
dracos: Given it's overriding most of the attributes of `options`, perhaps having a base options that both options and lighting_options can derive from might be a little clearer, but not much in it.

PR #2333, opened 2018-11-23
Use standard email sending for handlemail.

### LINKED ISSUE #2332: handlemail doesn't handle utf-8 correctly
If the reply-autoresponse template is a UTF-8 encoded text file with non-ASCII characters it doesn't seem to send it correctly.

I'm not sure what the issue is, I can see that it has correctly set the Content-Type header:

```Content-Type: text/plain; charset="utf-8"```

Still the text shows up as ASCII (in for instance gmail).

Sorry for not investigating further before creating an issue, I can perhaps find time to look more at it eventually.

We're using Postfix as MTA. See attached sample headers.

[headers.txt](https://github.com/mysociety/fixmystreet/files/2609345/headers.txt)

dracos:
The EmailUtil library expects "\n" line endings, otherwise it gets confused. It also doesn't handle a different SMTP port, and so on. Fixes #2332.

### DIFF
--- a/bin/handlemail
+++ b/bin/handlemail
@@ -22,7 +22,6 @@ BEGIN {
 use FixMyStreet;
 use FixMyStreet::DB;
 use FixMyStreet::Email;
-use mySociety::EmailUtil;
 use mySociety::HandleMail;
 use mySociety::SystemMisc qw(print_log);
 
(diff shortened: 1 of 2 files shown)


dracos:
Fixed post merge  to make sure return path had space/`<`/`>` stripped.

PR #2334, opened 2018-11-27
Adds cobrand support for handlemail.

### LINKED ISSUE #2331: handlemail always uses the auto-reply template from the default cobrand
When the handlemail script replies with the reply-autoresponse template, it always uses the one from the default cobrand.

**To Reproduce**
1. Configure the .forward file for the fms user to execute the handlemail script
2. Set up an alias or similar for the DO_NOT_REPLY address to the fms user
3. Send a regular email to the DO_NOT_REPLY address.

Note that the DO_NOT_REPLY address needs to start with the "fms-" prefix, otherwise it seems the handlemail script doesn't handle the mail (even if it matches the DO-NOT-REPLY regex in the script, not sure if this was intended).

I can supply a pull request with a fix for this if it's ok to specify cobrand with an argument to handlemail. As in:

```handlemail --cobrand=fixamingata```

### LINKED ISSUE #2332: handlemail doesn't handle utf-8 correctly

### LINKED ISSUE #2336: handlemail should use the Auto-Submitted header
handlemail currently detects out-of-office and other automatic replies by matching the subject line and body against patterns commonly found in such replies.

In RFC 3834, it is recommended that automatic replies SHOULD include the Auto-Submitted header (set to auto-replied). handlemail could check this as a first step, if no such header is included it could continue using the pattern matching.

When handlemail replies to an email sent to the DO_NOT_REPLY_EMAIL address, it should also set this header.

There may also be other, non-standard, headers to consider, see for instance:

https://stackoverflow.com/questions/9426801/detect-auto-reply-emails-programatically

### PR DESCRIPTION
**Pull request by joesiltberg**
Fixes #2331. Fixes #2332. Fixes #2336.

I also added an option for specifying a separate destination for the bounces we'd like to have a look at. The default is still that the CONTACT_EMAIL address handles these, but it can be overridden (we want it to go to a separate mailbox for FixaMinGata). It could have been a new config parameter I suppose but I didn't want to clutter the config file for this. Let me know if you don't like it and I'll remove it or change it.

### DIFF
--- a/bin/handlemail
+++ b/bin/handlemail
@@ -19,6 +19,7 @@ BEGIN {
     require "$d/../setenv.pl";
 }
 
+use Getopt::Long;
 use FixMyStreet;
 use FixMyStreet::DB;
 use FixMyStreet::Email;
@@ -28,6 +29,15 @@ use mySociety::SystemMisc qw(print_log);
 # Don't print diagnostics to standard error, as this can result in bounce
 # messages being generated (only in response to non-bounce input, obviously).
 mySociety::SystemMisc::log_to_stderr(0);
+
+my $cobrand = "default";
+
+# Where to forward mail that should be looked at by a person,
+# such as a permanent bounce for a report.
+my $bouncemgr = FixMyStreet->config('CONTACT_EMAIL');
+
+GetOptions ("cobrand=s" => \$cobrand,
+            "bouncemgr=s" => \$bouncemgr);
 
 my %data = mySociety::HandleMail::get_message();
 my @lines = @{$data{lines}};
@@ -103,7 +113,7 @@ sub handle_permanent_bounce {
         $object->disable();
     } elsif ($type eq 'report') {
         print_log('info', "Received bounce for report " . $object->id . ", forwarding to support");
-        forward_on_to(FixMyStreet->config('CONTACT_EMAIL'));
+        forward_on_to($bouncemgr);
     }
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: bin/handlemail [hunk 1]
```diff
@@ -161,14 +171,15 @@ sub handle_non_bounce_to_null_address {
...
+    my $fp = FixMyStreet->path_to("templates", "email", $cobrand, $template)->open or exit 75;
     $template = join('', <$fp>);
     $fp->close;
 
     # We generate this as a bounce.
     my ($rp) = $data{return_path} =~ /^\s*<(.*)>\s*$/;
     my $mail = FixMyStreet::Email::construct_email({
-        From => [ FixMyStreet->config('CONTACT_EMAIL'), 'FixMyStreet' ],
+        From => [ FixMyStreet->config('CONTACT_EMAIL'),
+                  FixMyStreet->config('CONTACT_NAME') ],
```
joesiltberg: Note that this is potentially a change in behaviour even if the command line options aren't used.

But I guess this is how we want it? The UK site probably has CONTACT_NAME = FixMyStreet anyway?

dracos: Yes, that looks fine to me!

joesiltberg:
Tested the UTF-8 issue, works fine now, thanks!

dracos:
I've added another commit for #2336, if you'd like to take a look, then I think we can get this merged!

joesiltberg:
Looks good, thanks @dracos !

PR #2347, opened 2018-12-10
Don’t escape HTML entities in reports got by ajax.

### LINKED ISSUE #2346: Titles of reports pulled in via JS have escaped entities.
e.g. https://www.fixmystreet.com/report/1431259 fine when visited directly, but has `&amp;` shown when pulled in via JS.

dracos:
Fixes #2346.


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1354,7 +1354,17 @@ fixmystreet.display = {
             }
 
             var found = html.match(/<title>([\s\S]*?)<\/title>/);
-            var page_title = found[1];
+            var map = {
```
zarino: I feel like maybe this could do with a comment, explaining:

1. We’re unencoding HTML entities in the title string, to make it suitable for use with`document.title` later
2. We’re only worried about these 5 particular entities because they’re the only 5 that the `html_filter` method of `FixMyStreet/Template.pm` encodes in the first place

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
         - Keep all moderation history, and show in report/update admin. #2329
     - Bugfixes:
         - Restore map zoom out when navigating to /around from /report. #1649
+        - Don’t escape HTML entities in reports pulled in by ajax. #2346
```
zarino: In report _titles_?

PR #2353, opened 2018-12-13
[Open311] warn on staging if fail to match updates

struan:
If we fail to match an update to a problem on the staging server then
print a warning so we can diagnose issues.

Please check the following:

- [X] All cobrand-specific commits start their commit message with the cobrand in square brackets;

### DIFF
--- a/perllib/Open311/GetServiceRequestUpdates.pm
+++ b/perllib/Open311/GetServiceRequestUpdates.pm
@@ -195,6 +195,10 @@ sub update_comments {
                     }
                 }
             }
+        # we get lots of comments that are not related to FMS issues from Lewisham so ignore those otherwise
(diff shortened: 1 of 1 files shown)


perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -195,6 +195,8 @@ sub update_comments {
                     }
                 }
             }
+        } elsif (FixMyStreet->config('STAGING_SITE')) {
+            warn "Failed to match comment $request_id to problem for " . $body->name . "\n";
```
davea: This means we couldn't find a problem whose `external_id` is `$request_id`, could the warning clarify that? The current wording makes it seem like we're searching for a comment with `$request_id`.

PR #2356, opened 2018-12-14
[Buckinghamshire] set incoming Bucks reports to non public

struan:
Also adds some basic filtering of incoming report descriptions to remove emails and phone numbers.

Please check the following:

Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/42

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -109,6 +109,23 @@ sub open311_config_updates {
     $params->{mark_reopen} = 1;
 }
 
+sub filter_report_description {
+    my ($self, $description) = @_;
+
+    # this allows _ in the domain name but I figure it's unlikely to
+    # generate false positives so lets go with that for the same of
+    # a simpler regex
+    $description =~ s/\b[\w.!#$%&'*+\-\/=?^_{|}~]+\@[\w\-]+\.[^ ]+\b//g;
+    $description =~ s/ (?: \+ \d{2} \s? | \b 0 ) (?:
+        \d{2} \s? \d{4} \s? \d{4}   # 0xx( )xxxx( )xxxx
+      | \d{3} \s \d{3} \s? \d{4}    # 0xxx xxx( )xxxx
+      | \d{3} \s? \d{2} \s \d{4,5}  # 0xxx( )xx xxxx(x)
+      | \d{4} \s \d{5,6}            # 0xxxx xxxxx(x)
+    ) \b //gx;
+
+    return $description;
+}
+
 sub map_type { 'Buckinghamshire' }
 
 sub default_map_zoom { 3 }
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -109,6 +109,15 @@ sub open311_config_updates {
     $params->{mark_reopen} = 1;
 }
 
+sub filter_report_description {
+    my ($self, $description) = @_;
+
+    $description =~ s/\b[\w]+\@[\w]+\.[^ ]+\b//gms;
```
dracos: The bit before an `@` could also have a variety of characters most notably `.`, but also any of `!#$%&'*+-/=?^_`{|}~`. The bit after can also have hyphen and cannot have underscore (included in `\w`).

dracos: Not sure what the `m` and `s` modifiers are for here.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -109,6 +109,15 @@ sub open311_config_updates {
...
+    $description =~ s/\b0[\d]{3}\s?\d\s?[\d]{2}\s?[\d]{4}\b//gms;
```
dracos: Given we have Number::Phone installed, feels a shame not to use it, but realise probably bit complex to setup (e.g. assumes only passed a number). OTOH, it does cover all possible valid displays of numbers, if you can get the validators out of it somehow; this regex does not match e.g. `020` style numbers or `013873` ones, written normally. Looks like possible normal formatting chunks include `07{3}{6}`, `0{2}{4}{4}` (e.g. 020), `0{3}{3}{4}`, `0{5}{4,5}`, `01{3}{5,6}`.  Bucks numbers look to be of the last, mostly.

Ditto on the `m`/`s` question.

dracos: This could be changed a little - currently it will match, if I've read it right: `0\d{10}`, `0\d{6} \d{4}`, `0\d{4} \d{6}`, `0\d{4} \d{2} \d{4}`, `0\d{3} \d{7}`, `0\d{3} \d{3} \d{4}`, `0\d{3} \d \d{6}`, and `0\d{3} \d \d{2} \d{4}` some of which seem a bit implausible :) Perhaps something like:
```
s/ \b 0 (?:
    \d{2} \s? \d{4} \s? \d{4}   # 0xx( )xxxx( )xxxx
  | \d{3} \s \d{3} \s? \d{4}    # 0xxx xxx( )xxxx
  | \d{3} \s? \d{2} \s \d{4,5}  # 0xxx( )xx xxxx(x)
  | \d{4} \s \d{5,6}            # 0xxxx xxxxx(x)
) \b //gx
```
which covers some more bases (e.g. London-based numbers) and seems a bit more explicit as to what it's looking for. What do you think?

PR #2361, opened 2018-12-20
Show reopening/fixed questionnaire responses lacking updates.

### LINKED ISSUE #2357: Report re-opened through questionnaire without comment
**Describe the bug**
We've noticed a few issues recently where a body has marked a report as Fixed, yet it shows up as still Open. 

It looks like this when viewing the report:

solvedornot

The only update in the report is from the body, which has marked the report as Fixed (circled in the picture - "Löst" in Swedish). The map pin shows it as open, and there's no fixed banner at the top right of the report frame.

Checking the problem state in the database:
```
fixmystreet=> SELECT state FROM problem WHERE id = 10766;
   state
-----------
 confirmed
(1 row)
```
Checking the updates:

```
fixmystreet=> SELECT id, created, text, problem_state FROM comment WHERE problem_id = 10766;
  id   |          created           |                      text                      |  problem_state
-------+----------------------------+------------------------------------------------+-----------------
 17435 | 2018-12-03 08:58:29.936004 | Uppsatt på asfalteringslista. /Forshaga kommun | fixed - council
(1 row)
```

Checking the questionnaires:

```
fixmystreet=> SELECT * FROM questionnaire WHERE problem_id = 10766;
  id   | problem_id |          whensent          |       whenanswered        | ever_reported |    old_state    | new_state
-------+------------+----------------------------+---------------------------+---------------+-----------------+-----------
 11758 |      10766 | 2018-12-05 08:52:04.496757 | 2018-12-06 19:54:27.02879 |               | fixed - council | confirmed
(1 row)
```

So it seems the reporter has re-opened the report through the questionnaire. This doesn't show up in the updates however.

According to https://fixmystreet.org/running/surveys/ the reporter needs to leave a comment when re-opening an issue:

"If the user changes the state of a problem that is currently fixed to something else, that is, they effectively re-open the problem, then the update comment is not optional."

So I suppose there should have been another row in the comment table for this report?

**To Reproduce**
We have been unable to reproduce. When we re-open through the questionnaire a comment is required (and that comment correctly shows up in the report)

**Expected behavior**
When a report is re-opened through the questionnaire, it should be visible among the updates for the report.

**Additional context**
FixaMinGata is currently running v2.3.4 of FixMyStreet.

Any tips on how I can investigate this further, or something I can try in order to reproduce, would be much appreciated!

dracos:

Should fix #2357 (cc @joesiltberg).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
     - Bugfixes:
         - Restore map zoom out when navigating to /around from /report. #1649
         - Don’t escape HTML entities in report titles pulled in by ajax. #2346
+        - Show reopening/fixed questionnaire responses lacking updates. #2357
     - Open311 improvements:
         - Fix bug in contact group handling. #2323
         - Improve validation of fetched reports timestamps. #2327
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -186,14 +186,30 @@ sub load_updates : Private {
         { order_by => [ 'confirmed', 'id' ] }
     );
 
-    my $questionnaires = $c->model('DB::Questionnaire')->search(
+    my $questionnaires_still_open = $c->model('DB::Questionnaire')->search(
         {
             problem_id => $c->stash->{problem}->id,
             whenanswered => { '!=', undef },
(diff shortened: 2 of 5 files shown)


joesiltberg:
Looks good to me also!

PR #2363, opened 2018-12-21
add private only report filter

struan:
This adds an option on the reports screen to display only private reports.

Please check the following:

Thanks for contributing!

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/Problem.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Problem.pm
@@ -28,14 +28,23 @@ sub body_query {
 sub non_public_if_possible {
     my ($rs, $params, $c) = @_;
     if ($c->user_exists) {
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -28,14 +28,27 @@ sub body_query {
...
-                non_public => 0,
-                $rs->body_query($c->user->from_body->id),
-            ];
+            if ($only_non_public) {
+                $params->{'-and'} = [
+                        non_public => 1,
+                        $rs->body_query($c->user->from_body->id),
+                ];
+            } else {
+                $params->{'-or'} = [
```
dracos: I think this else condition can be the same as it was before this PR - you either want non_public = 0 or to be in the right body (in which case it doesn't matter what non_public is).

PR #2370, opened 2019-01-08
Example cobrand for Westminster

zarino:
Fixes mysociety/community#57

[skip changelog]


zarino:
@dracos Just thinking about how we’d run this on the commercial staging server – we probably wouldn’t want to use `--empty` because that’d lose all the existing bodies/reports/users for that instance. But we couldn’t just run the script as-is either, because it’d create a new Westminster body.

Maybe I should add a `--body-id` command line argument that lets you bypass the `find_or_create` and just use an existing body? I guess it would also need to delete/recreate that body’s categories too.

Or another alternative is that we leave the script as it is, but for this special case, just drop the Westminster body manually, in the database, so the script can create a new one.

Thoughts?

dracos:
If it's a one-off script, your last option sounds easiest. Depends if then changes will be made in admin to the body, presumably not?

If it's needed to be able to run without disturbing existing contents, then it's not really a fixture script any more. I would change the code to look for the existing body first (somehow), and if not there, create it (so not use `find_or_create`, but separate `search` and then `create` if needed). Then as you say, it would also need to do the same for each category, checking for presence and creating if not there. (For the latter, you could improve find_or_create to work with multiple id_field (body+category in that case), but for body bit harder given you could e.g. have two bodies covering the same area - Westminster + TfL for example, so would need to decide what you're basing the creation on).

zarino:
No need to merge this into master yet. Will revisit and amend `issues/community/57-westminster-cobrand` as needed, next month.

PR #2372, opened 2019-01-10
Add start and end date extra fields to licensing categories for demo

struan:
This also sneaks in the ability to set the type of extra field inputs so we can use the nice build in date picker for these fields.

Please check the following:

### DIFF
--- a/bin/fixmystreet.com/fixture
+++ b/bin/fixmystreet.com/fixture
@@ -82,6 +82,10 @@ for my $cat ('Dropped Kerbs', 'Skips') {
         category => $cat
     });
     $child_cat->set_extra_metadata( group => 'Licensing' );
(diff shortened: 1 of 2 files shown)


struan:
I guess in theory but in this case where we just want a date and not a date and time it would be good to be able to differentiate. And of course unhelpfully the html5 input type for a datetime isn't datetime but datetime-local :/

I think it needs a bit more thought, I was just hacking this in here to look nice for the demo :)

dracos:
"datetime: The input generated must be able to transform into a valid ISO 8601 date. Represented in an HTML form using <input> tags " - implies it could just be date... I just meant it seemed to make sense to use datatype (even if you just put 'date' in it) rather than a new variable.

PR #2381, opened 2019-01-15
Allow state editing via moderation.

dracos:
If a state is passed in (on a cobrand), update problem state and add a comment to record this (as state changes are updates not moderations).

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Moderate.pm
+++ b/perllib/FixMyStreet/App/Controller/Moderate.pm
@@ -81,6 +81,7 @@ sub moderate_report : Chained('report') : PathPart('') : Args(0) {
     $c->forward('report_moderate_hide');
 
     my @types = grep $_,
+        $c->forward('moderate_state'),
         ($c->user->can_moderate_title($problem, 1)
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Moderate.pm
```diff
@@ -299,6 +306,31 @@ sub moderate_category : Private {
...
+
+    my $new_state = $c->get_param('state');
+    return unless $new_state;
+
+    my $problem = $c->stash->{problem};
+    if ($problem->state ne $new_state) {
+        $problem->state($new_state);
+        $problem->update;
+        $problem->add_to_comments( {
+            text => $c->stash->{moderation_reason},
```
struan: Normally this doesn't get made public so is that not going to be a bit unexpected?

dracos: Hmm, so this won't be on .com, only on a cobrand that's deliberately added state to the moderation form (like kiitc) where it's clearer it's a public editing, but yes, I guess it still might be odd. Guess I'll add a comment here at least about thinking about that for any other cobrand that might use it.

PR #2383, opened 2019-01-17
[Buckinghamshire] Question for on-road flytipping.

dracos:

Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/45

* Tidies up test file to not hard-code IDs, new test is at bottom.
* JS for showing/hiding extra question.
* Perl to get extra question into DB, not being pulled from Open311.
* Server side code to work out what to do if both bodies still given (question made mandatory in most cases, but not if district only).
* Script to auto close >3 weeks old district flytipping

### DIFF
--- a/bin/fixmystreet.com/buckinghamshire-flytipping
+++ b/bin/fixmystreet.com/buckinghamshire-flytipping
@@ -1,0 +1,80 @@
+#!/usr/bin/env perl
+#
+# If a district flytipping report within Buckinghamshire has not been closed
+# after three weeks, close it with a message. If it's older than six weeks,
+# use a different message and suppress any alerts.
+
+use v5.14;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use constant BUCKS_AREA_ID => 2217;
+use constant DISTRICT_IDS => (2255, 2256, 2257, 2258);
+use constant TIME_OPEN => '3 weeks';
+use constant TIME_OPEN_ALERT => '6 weeks';
+
+use FixMyStreet::DB;
+use FixMyStreet::Script::ArchiveOldEnquiries;
+use Getopt::Long::Descriptive;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['commit|c', "actually close reports and send emails. Omitting this flag will do a dry-run"],
+    ['help|h', "print usage message and exit" ],
+);
+print($usage->text), exit if $opts->help;
+
+my $body = FixMyStreet::DB->resultset("Body")->for_areas(BUCKS_AREA_ID)->first;
+die "Could not find Bucks body" unless $body;
+
+my @districts = FixMyStreet::DB->resultset("Body")->for_areas(DISTRICT_IDS)->all;
+my @district_ids = map { $_->id } @districts;
+die "Did not find all districts" unless @district_ids == 4;
+
+find_problems(TIME_OPEN_ALERT, TIME_OPEN, 'Auto-closure', 1);
+find_problems(undef, TIME_OPEN_ALERT, 'Auto-closure (old)', 0);
+
+sub find_problems {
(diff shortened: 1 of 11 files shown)


perllib/Open311/PopulateServiceList.pm
```diff
@@ -266,6 +266,19 @@ sub _add_meta_to_contact_cobrand_overrides {
...
+            datatype => 'singlevaluelist',
+            description => 'Is the fly-tip located on',
+            order => 100,
+            required => 'false',
+            variable => 'true',
+            values => [
+                { key => 'road', name => 'The road' },
+                { key => 'off-road', name => 'Off the road/on a verge' },
+            ],
+        } if $self->_current_service->{service_name} eq 'Flytipping';
```
struan: This really feels like something we should push into the cobrand modules at some point.

dracos: Have pushed extra commit that should do this.

struan: yay!

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -126,6 +126,30 @@ sub open311_contact_meta_override {
     } if $service->{service_name} eq 'Flytipping';
 }
 
+sub process_open311_extras {
+    my ($self, $c, $body, $extra) = @_;
+
+    my $report = $c->stash->{report};
+    if ($report->bodies_str =~ /,/) {
```
struan: Is this relying on only the Flytipping category begin sent to both District and County so it'll be the only one with a comma in the `bodies_str`? That sort of feels like we should have a `return unless $report->category eq "Flytipping"`; check first.

dracos: Looks like Trees are also sent to both, so yes, thanks!

bin/fixmystreet.com/buckinghamshire-flytipping
```diff
@@ -0,0 +1,51 @@
...
+die "Did not find all districts" unless @district_ids == 4;
+
+my $template = FixMyStreet::DB->resultset("ResponseTemplate")->search({
+    body => $body, title => 'Auto-closure',
+});
+die "Could not find Bucks Flytipping template" unless $template;
+
+# Fetch all Flytipping problems sent only to districts, older than TIME_OPEN
+my $q = FixMyStreet::DB->resultset("Problem")->to_body(\@district_ids)->search(
+    \[ "NOT regexp_split_to_array(bodies_str, ',') @> ?", [ {} => [ $body->id ] ] ]
```
struan: I did not know about @> as an operator.

dracos: Me neither! `to_body` uses `&&` so returns any that overlap, guess a one-query alternative would be using: `\[ "regexp_split_to_array(bodies_str, ',') = ?", [ {} => [ $body->id ] ] ]` which might be clearer too.

struan: would that not get reports only sent to the body?

perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -135,22 +138,27 @@ sub close_problems {
     return unless $opts->{commit};
 
     my $problems = shift;
+    my $extra = { auto_closed_by_script => 1 };
+    $extra->{is_superuser} = 1 if !$opts->{user_name};
```
struan: Does the `auto_closed_by_script` not belong with the report rather than the comment? I guess it doesn't make much difference as we can work back from the comments if we need to.

bin/fixmystreet.com/buckinghamshire-flytipping
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env perl
+#
+# If a district flytipping report within Buckinghamshire has not been closed
+# after three weeks, close it with a message.
+
```
struan: might want to update this comment to explain the extra bits for older reports.

PR #2386, opened 2019-01-29
Duplicate report suggestion UI

zarino:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1229.

See also https://github.com/mysociety/fixmystreet-commercial/pull/1274 for related change in Ground Control cobrand.


templates/web/base/report/_item_expandable.html
```diff
@@ -0,0 +1,81 @@
...
+    [% PROCESS 'report/_item_heading.html' %]
+  [% CATCH file %]
+    <h3>
+        <a href="[% c.cobrand.base_url_for_report( problem ) %][% problem.url %]">
+            [% problem.title | html %]
+        </a>
+    </h3>
+  [% END %]
+
+    <small>
```
dracos: All the `small` is the same in `_item.html` and should be factored out to a new shared template - makes it much easier to then compare to the main one.

templates/web/base/report/_item_expandable.html
```diff
@@ -0,0 +1,81 @@
+[%#
+  This list item contains a full photo set and description, which is hidden
+  until the `.expanded` class is added. It is useful for times when you want
+  to surface the full information about a report, without reqiring the user
```
dracos: Typo

templates/web/base/report/_item_expandable.html
```diff
@@ -0,0 +1,81 @@
...
+      [% problem.detail | truncate(75, '…') | html %]
+    </p>
+
+    <div class="item-list__item--expandable__hide-until-expanded">
+      [% problem.detail | add_links | html_para %]
+      [% INCLUDE 'report/photo.html' object=problem %]
+    </div>
+
+    <div class="item-list__item--expandable__actions">
+      <button class="btn btn--small js-toggle-expansion">Read more</button>
```
dracos: Missing translation

dracos: And this shouldn't show if there's no truncation/extra photos to show, I guess?

templates/web/base/report/new/duplicate_suggestions.html
```diff
@@ -0,0 +1,11 @@
+[% IF c.cobrand.suggest_duplicates %]
+<div id="js-duplicate-reports" class="duplicate-report-suggestions hidden">
+    <button class="duplicate-report-suggestions__close js-hide-duplicate-suggestions">[% loc('Close') %]</button>
+    <h2 class="form-section-heading">[% loc('Already been reported?') %]</h2>
+    <div class="form-section-description">
+        <p>[% loc('We’ve already sent these similar nearby reports to the council. You can track them to receive updates.') %]</p>
+    </div>
+    <ul class="item-list"></ul>
+    <button class="btn btn--block js-hide-duplicate-suggestions">[% loc('No, continue with my new report') %]</button>
```
dracos: On mobile, this is immediately followed by the main navigation which looks a bit odd.

templates/web/base/report/_item_expandable.html
```diff
@@ -0,0 +1,81 @@
...
+  to visit the dedicated report page.
+
+  Since the photo set includes `<a>` elements of its own, we drop the usual
+  "wrapper" `<a>` and the associated `.item-list--reports__item` class, to
+  avoid anchor nesting.
+%]
+
+<li class="item-list__item item-list__item--expandable [% item_extra_class %]"
+    data-report-id="[% problem.id | html %]"
+    data-lastupdate="[% problem.lastupdate %]"
```
dracos: Think this is only for offline reports, so isn't needed here.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1049,6 +1049,239 @@ $.extend(fixmystreet.set_up, {
...
+        refresh_duplicate_list().always(function(){
+            // Wait an extra second until we allow another reload.
+            setTimeout(function(){
+                category_changing = false;
+            }, 1000);
+        });
+    }
+
+    // Want to show potential duplicates when a regular user starts a new report.
+    // TODO: This also fires after you move the pin while making a report. Do we want that?
```
dracos: I guess we do, but perhaps if they haven't started filling in the textboxes/uploaded an image?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1049,6 +1049,239 @@ $.extend(fixmystreet.set_up, {
...
+        }
+
+        $.when.apply(this, animations).then(cb);
+    }
+
+    function remove_duplicate_pins() {
+        fixmystreet.markers.removeFeatures( current_duplicate_markers );
+    }
+
+    function inspect_form_state_change() {
```
dracos: This function (and any related staff-only bits) could remain in staff.js, if it's only applicable to staff...

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1049,6 +1049,239 @@ $.extend(fixmystreet.set_up, {
         var set_map_state = true;
         fixmystreet.back_to_reports_list(e, report_list_url, map_state, set_map_state);
     });
+  },
+
+  expandable_list_items: function(){
+      // All of the .off/.on event binding here is intended to allow us
+      // to call fixmystreet.set_up.expandable_list_items() multiple times
+      // (eg: after report list items are loaded by ajax) without the
+      // effects stacking up on top of each other.
```
dracos: But you're setting the event on document; so why do you ever need to set it up again once it's been set up the first time?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1049,6 +1049,239 @@ $.extend(fixmystreet.set_up, {
...
+                  e.preventDefault();
+                  $li.addClass('expanded');
+                  $li.find('.js-toggle-expansion').text(translation_strings.read_less);
+              }
+          }
+      });
+  },
+
+  manage_duplicates: function() {
+    // Can skip all this if there’s no duplicates container on the page.
```
dracos: Lot of extra JS though, in that case. Can we perhaps have this in a separate JS file, that is only included then if that template also is? Might be easier/better than separating out staff/non-staff, dunno.

web/cobrands/fixmystreet/staff.js
```diff
@@ -1,84 +1,4 @@
...
-      // Deal with changes to report state by inspector/other staff, specifically
-      // displaying nearby reports if it's changed to 'duplicate'.
-      function refresh_duplicate_list() {
-          var report_id = $("#report_inspect_form .js-report-id").text();
-          var args = {
-              filter_category: $("#report_inspect_form select#category").val(),
-              latitude: $('input[name="latitude"]').val(),
-              longitude: $('input[name="longitude"]').val()
-          };
-          $("#js-duplicate-reports ul").html('<li class="item-list__item">Loading...</li>');
```
dracos: Do we mind the loading indicator has now gone?

web/cobrands/fixmystreet/fixmystreet.js
zarino: Because I need to bind click handlers to the list items (which are grandparents of the `.js-toggle-expansion` elements), so I select them with `$('.js-toggle-expansion').closest('li')` on line 1071, and that means, if this gets run at pageload, I might need to run it again, later, if new `.js-toggle-expansion` elements are added to the DOM.

I guess we could avoid the `$('.js-toggle-expansion').closest('li');` line by binding to `$(document).on('click', '.item-list__item--expandable')` instead?

dracos: Yes, something like that would be easier, I think (perhaps another js class for the item list to show it's a js binding class).

web/cobrands/sass/_base.scss
```diff
@@ -1194,12 +1194,10 @@ input.final-submit {
...
     border-bottom: none;
+    background-color: rgba(255, 255, 255, 0.5);
 
-    .item-list__item {
-        background-color: rgba(255, 255, 255, 0.5);
-    }
-
-    .item-list--reports__item--selected {
+    .item-list__item--selected,
+    .js & .item-list__item--selected.expanded:hover {
```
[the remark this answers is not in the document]
zarino: Hmm yes, this should probably be in “Expandable list item component” instead 👍

templates/web/base/report/new/duplicate_suggestions.html
zarino: Turns out this is also a problem _before_ the suggestion UI is even shown:

screen shot 2019-02-20 at 11 19 20

zarino: I’ve added a commit that fixes this for all map pages at mobile widths.

templates/web/base/report/_item_expandable.html
```diff
@@ -0,0 +1,55 @@
...
+        <a href="[% c.cobrand.base_url_for_report( problem ) %][% problem.url %]">
+            [% problem.title | html %]
+        </a>
+    </h3>
+  [% END %]
+
+    <small>[% PROCESS 'report/_item_small.html' %]</small>
+
+    <p class="item-list__item--expandable__hide-when-expanded">
+      [% problem.detail | truncate(75, '…') | html %]
```
zarino: @dracos Shame we can’t use `[% truncated_detail %]` here for neatness. Maybe we could, if we wrapped it in a `<div class="item-lis…panded">` instead of a `<p>`?

dracos: Yeah, if that works for you, I wasn't sure if that could happen or not, go for it if it's okay.

zarino: Yup, turns out it works fine, so I’ve made that change :+1:

templates/web/base/report/new/duplicate_suggestions.html
```diff
@@ -6,9 +6,33 @@
...
+        <p>[% loc('We’ve already reported these similar problems nearby reports to the council. Is one of them yours?') %]</p>
     </div>
     <ul class="item-list"></ul>
-    <button class="btn btn--block js-hide-duplicate-suggestions">[% loc('No, continue with my new report') %]</button>
+    <button class="btn btn--block js-hide-duplicate-suggestions">[% loc('Continue – report a new problem') %]</button>
+</div>
+<div class="js-template-get-updates hidden">
+    <div id="alerts" class="get-updates">
+        <p id="rznvy_hint">
+            <strong>[% loc('Great!') %]</strong>
```
dracos: Is it great? They've got an open problem, after all. Perhaps the last updated date is months ago and this makes it worse? :) I think we should drop this, anyway.

templates/web/base/report/new/duplicate_suggestions.html
```diff
@@ -6,9 +6,33 @@
...
+          [% IF c.user_exists %]
+            [% loc('Would you like us to notify you when this problem is updated or fixed?') %]
+          [% ELSE %]
+            [% loc('If you let us know your email address, we’ll notify you when this problem is updated or fixed.') %]
+          [% END %]
```
dracos: This is fine, but I think it could say the logged in text either way and it'd be clear enough?

templates/web/base/report/new/duplicate_suggestions.html
```diff
@@ -6,9 +6,33 @@
     <button class="duplicate-report-suggestions__close js-hide-duplicate-suggestions">[% loc('Close') %]</button>
     <h2 class="form-section-heading">[% loc('Already been reported?') %]</h2>
     <div class="form-section-description">
-        <p>[% loc('We’ve already sent these similar nearby reports to the council. You can track them to receive updates.') %]</p>
+        <p>[% loc('We’ve already reported these similar problems nearby reports to the council. Is one of them yours?') %]</p>
```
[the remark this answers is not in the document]
zarino: The reason I mentioned "reports" and having been sent/reported "to the council" was to make it clearer to the user that they _don’t need to report these issues again_. I’m a bit worried, if we just say, "We’ve found these problems nearby", people won’t understand that reporting them again won’t help?

templates/web/base/report/new/duplicate_suggestions.html
zarino: Fair enough! Drop it.

templates/web/base/report/new/duplicate_suggestions.html
zarino: Do you think it’ll be obvious enough, for a first-time user, without an account, that when they see `Would you like us to notify you when this problem is updated or fixed?` and a `Your email` input box below it, that they should enter their email address in the box to get updates?

I mean, when I explain it like that, of course it feels obvious enough. But I’m a bit worried that the answer to "Would you like us to notify/email you?" is a yes/no, which is fine when it’s followed by just a button, but a bit weird when it’s followed by a text input. What do you think?

dracos: I dunno. I'm happy either way I think.

davea:
The sprite for the cross icon isn't aligned quite right (this is on macOS 10.14 Safari 12 on a retina Mac)  image

Wording suggestion from @LouiseMySociety:
> Since it’s on B&NES co brand _We’ve already reported these nearby problems to the council. Is one of them yours_ doesn’t really work, because We is the council already.
> What about _There are similar problems nearby that we’re already aware of, is one of them yours?_

PR #2393, opened 2019-02-08
[fixmystreet.com] Improve two-tier unresponsive.

dracos:
Similarly to 1f69e28c, we were previously only checking the first matching entry, which led to confusing behaviour in places. Include consequential amendments for e.g. one body being unresponsive, the other not.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -286,7 +286,9 @@ sub by_category_ajax_data : Private {
     # unresponsive must return empty string if okay, as that's what mobile app checks
     if ($type eq 'one' || ($type eq 'all' && $unresponsive)) {
         $body->{unresponsive} = $unresponsive;
-        if ($type eq 'all' && $unresponsive) {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -286,7 +286,9 @@ sub by_category_ajax_data : Private {
     # unresponsive must return empty string if okay, as that's what mobile app checks
     if ($type eq 'one' || ($type eq 'all' && $unresponsive)) {
         $body->{unresponsive} = $unresponsive;
-        if ($type eq 'all' && $unresponsive) {
+        # Check for no bodies here, because if there are any (say one
+        # unresponsive, one not), can use default display code for that.
+        if ($type eq 'all' && $unresponsive && !@$bodies) {
```
struan: if `$type` is `'all'` will `$unresponsive` not always be true here?

dracos: Yes; I assume that was to just be clear it was matching the second condition of the outer if statement. But now I've added a new bit to it, not much point.

PR #2395, opened 2019-02-13
Allow user to be associated with multiple areas.

dracos:
Update database to store an array of IDs rather than only one; consequential changes to the admin and the dashboard to allow selection. Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1270.

### DIFF
--- a/t/app/controller/dashboard.t
+++ b/t/app/controller/dashboard.t
@@ -113,7 +113,7 @@ FixMyStreet::override_config {
     };
 
     subtest 'area user can only see their area' => sub {
-        $counciluser->update({area_id => $area_id});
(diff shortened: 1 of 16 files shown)


t/app/controller/dashboard.t
```diff
@@ -122,7 +122,7 @@ FixMyStreet::override_config {
         $mech->get_ok("/dashboard?ward=$alt_area_id");
         $mech->content_contains('<h1>Trowbridge</h1>');
 
-        $counciluser->update({area_id => undef});
+        $counciluser->update({area_ids => undef});
     };
```
struan: We probably should have a test that checks assigning multiple area ids just for completeness

PR #2404, opened 2019-02-21
[FixaMinGata] Adds a cobrand hook (threshold for responsiveness top 5 list)

### LINKED ISSUE #1957: Changed "Top 5 responsive councils" criteria
The "Top 5 responsive councils" list on FixaMinGata shows a municipality that has only solved one problem (back in 2013) at the top, since that action took only about four minutes, and that beats any other average response time on the site.

stats1

This does not seem like a fair representation of responsiveness, although I'm not sure what a better criteria would be. Perhaps there should be a minimum number of reports to be able to make the list. Perhaps reports of a certain age should not be taken into account.

joesiltberg:
This adds an optional threshold so that a body with very few fixed reports don't end up in the top 5 list under /reports.

I haven't tested the change in App/Controller/Dashboard.pm, am I supposed to be able to see the responsiveness somewhere under /dashboard? Or should I look somewhere else?

Fixes #1957

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -292,7 +292,7 @@ sub generate_summary_figures {
 sub generate_body_response_time : Private {
     my ( $self, $c ) = @_;
 
-    my $avg = $c->stash->{body}->calculate_average;
+    my $avg = $c->stash->{body}->calculate_average($c->cobrand->call_hook("body_responsiveness_threshold"));
     $c->stash->{body_average} = $avg ? int($avg / 60 / 60 / 24 + 0.5) : 0;
 }
 
--- a/perllib/FixMyStreet/DB/Result/Body.pm
+++ b/perllib/FixMyStreet/DB/Result/Body.pm
@@ -204,7 +204,8 @@ sub get_cobrand_handler {
 }
 
 sub calculate_average {
-    my ($self) = @_;
+    my ($self, $threshold) = @_;
+    $threshold ||= 0;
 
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -292,7 +292,7 @@ sub generate_summary_figures {
 sub generate_body_response_time : Private {
     my ( $self, $c ) = @_;
 
-    my $avg = $c->stash->{body}->calculate_average;
+    my $avg = $c->stash->{body}->calculate_average($c->$cobrand->call_hook("body_responsiveness_threshold") || 0);
```
dracos: `s/$cobrand/cobrand` here should fix the tests.

perllib/FixMyStreet/DB/Result/Body.pm
```diff
@@ -204,7 +204,7 @@ sub get_cobrand_handler {
 }
 
 sub calculate_average {
-    my ($self) = @_;
+    my ($self, $threshold) = @_;
```
dracos: Perhaps here have a safety `$threshold ||= 0;` line, and then you don't need to make sure it's 0 when calling the function in the two places.

joesiltberg:
Thanks for the feedback @dracos ! Should I make the change for filtering out hidden reports in the regular code (i.e. not a cobrand hook)?

dracos:
Yes, I can't see why it shouldn't be in the main code, thanks!

dracos:
I've reworded one commit (wasn't FixaMinGata specific :) ), and put the fixes from the second commit in the first one (in general, if you say "also" in a commit message, it should probably be a separate commit ;) Makes it easier to squash/fixup). Thanks, will merge when tests pass.

joesiltberg:
Great, thanks!

PR #2405, opened 2019-02-22
[Buckinghamshire] Add contributed_by to CSV export

dracos:
Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/49


t/cobrand/bucks.t
```diff
@@ -172,6 +177,42 @@ for my $test (
...
+
+    @rows = $mech->content_as_csv;
+    is scalar @rows, 1, '1 (header) + 0 (updates)';
+    is scalar @{$rows[0]}, 9, '9 columns present';
+    is_deeply $rows[0],
+        [
+            'Report ID', 'Update ID', 'Date', 'Status', 'Problem state',
+            'Text', 'User Name', 'Reported As', 'Staff User',
+        ],
+        'Column headers look correct';
```
davea: This doesn't seem to test the contents of the CSV as there are no updates - worth adding some?

PR #2414, opened 2019-02-28
Switch default default_show_name to off.

dracos:


perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -1000,9 +1000,7 @@ Returns true if the show name checkbox should be ticked by default.
 
 =cut
 
-sub default_show_name {
-    1;
-}
```
struan: We should probably made this super clear in the CHANGELOG in case people do still want this as the default.

PR #2415, opened 2019-03-01
Add DateRange class.

dracos:
This simplifies the places where a date range is used; it also makes Dashboard date parsing more flexible. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DateRange.pm
+++ b/perllib/FixMyStreet/DateRange.pm
@@ -1,0 +1,72 @@
+package FixMyStreet::DateRange;
+
+use DateTime;
+use DateTime::Format::Flexible;
+use Moo;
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DateRange.pm
```diff
@@ -0,0 +1,69 @@
...
+    return unless $dt;
+    $self->formatter->format_datetime($dt);
+}
+
+sub start_formatted { $_[0]->_formatted($_[0]->start) }
+sub end_formatted { $_[0]->_formatted($_[0]->end) }
+
+sub sql {
+    my ($self, $default) = @_;
+    my $sql = { '>=', $self->start_formatted };
```
struan: Does this need to do something to check that there is a start date?

PR #2418, opened 2019-03-04
Improve inline checkbox spacing.

### LINKED ISSUE #2411: Reduce gap between labels and checkboxes in inline forms
> there is a gap between the tickbox and the ‘Show my name publicly’ field which means the person testing with a zoomed-in screen didn’t see the tickbox

image

dracos:
Reduce the space between checkbox and label, so the text isn't lost if the page is zoomed in. Fixes #2411.

Replace the creator_fixed question with segmented control to match the full questionnaire, and the admin verified checkboxes with inline-text as they need no padding. Having done that, all uses of .inline are now on their own lines, so no post-padding is needed any longer, but we’ll leave some in case of future use.


wrightmartin:
Looks good from here 👍

PR #2420, opened 2019-03-04
Stream dashboard CSV output.

dracos:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -3,6 +3,7 @@ use Moose;
 use namespace::autoclean;
 
 use DateTime;
+use Encode;
 use JSON::MaybeXS;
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -408,12 +409,17 @@ hashref of extra data to include that can be used by 'columns'.
 sub generate_csv : Private {
     my ($self, $c) = @_;
 
+    my $filename = $c->stash->{csv}->{filename};
+    $c->res->content_type('text/csv; charset=utf-8');
+    $c->res->header('content-disposition' => "attachment; filename=\"${filename}.csv\"");
+    $c->res->header('Surrogate-Control' => 'content="BigPipe/1.0"');
```
struan: Might be worth a comment on what this is for. And also a comment to explain anything that might need to happen on the webserver end?

PR #2425, opened 2019-03-07
[Buckinghamshire] Send drainage reports via email as well as Open311

davea:
Bucks want drainage reports sent via email as well as Confirm.

Includes a slight refactor of `Buckinghamshire::open311_post_send` to allow multiple categories to be sent via email once they've been sent to Confirm.

Fixes mysociety/fixmystreet-commercial#1306

Please check the following:

### DIFF
--- a/t/cobrand/bucks.t
+++ b/t/cobrand/bucks.t
@@ -10,6 +10,7 @@ my $counciluser = $mech->create_user_ok('counciluser@example.com', name => 'Coun
 
 $mech->create_contact_ok(body_id => $body->id, category => 'Flytipping', email => "FLY");
 $mech->create_contact_ok(body_id => $body->id, category => 'Potholes', email => "POT");
+$mech->create_contact_ok(body_id => $body->id, category => 'Blocked drain', email => "DRA");
 
 my $district = $mech->create_body_ok(2257, 'Chiltern');
 $mech->create_contact_ok(body_id => $district->id, category => 'Flytipping', email => "flytipping\@chiltern");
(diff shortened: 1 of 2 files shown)


t/cobrand/bucks.t
```diff
@@ -177,13 +179,27 @@ for my $test (
...
+my ($report2) = $mech->create_problems_for_body(1, $body->id, 'Drainage problem', {
+    category => 'Blocked drain', cobrand => 'fixmystreet',
+    latitude => 51.812244, longitude => -0.827363,
+});
+
+subtest 'blocked drain sent to extra email' => sub {
+    $mech->clear_emails_ok;
+    FixMyStreet::Script::Reports::send();
+    my $email = $mech->get_email;
+    my $e = join('', 'floodmanagement', '@', 'buckscc.gov.uk');
```
dracos: `join('@', 'floodmanagement', 'buckscc.gov.uk')` perhaps? Looks less like an email to any spam github scrapers. (And similar elsewhere, not sure why I did a join with blank?)

t/cobrand/bucks.t
```diff
@@ -177,13 +179,27 @@ for my $test (
     };
 }
 
+my ($report2) = $mech->create_problems_for_body(1, $body->id, 'Drainage problem', {
```
dracos: This test block needs to be higher up (after `flytipping off road sent to extra email` I guess?), otherwise `$cobrand` has been redefined and it makes an external call to Tilma and doesn't work offline. (This is why the coverage jumped, sadly :) ).

PR #2429, opened 2019-03-12
Bexley cobrand [go-live branch]

dracos:
[skip changelog] This should be all reviewed, ready for master-merge code.


web/cobrands/sass/_base.scss
```diff
@@ -127,11 +127,9 @@ ol.big-numbers {
       top:-0.2em;
       color:$col_big_numbers;
       line-height:1;
-      font: {
```
davea: Can these Sass changes be in their own commit explaining why they're necessary?

PR #2430, opened 2019-03-13
[BANES,Bucks,Lincs] Include reference in .com confirm sent email

dracos:
And fix bug where two-tier message not appearing in confirm sent email.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1273 [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/Reports.pm
+++ b/perllib/FixMyStreet/Script/Reports.pm
@@ -127,10 +127,19 @@ sub send(;$) {
             $missing = join(' / ', @missing) if @missing;
         }
 
+        my $send_confirmation_email = $cobrand->report_sent_confirmation_email;
+
         my @dear;
         my %reporters = ();
         my $skip = 0;
         while (my $body = $bodies->next) {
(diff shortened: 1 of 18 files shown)


perllib/FixMyStreet/Script/Reports.pm
```diff
@@ -127,10 +127,19 @@ sub send(;$) {
...
 
+        my $send_confirmation_email = $cobrand->report_sent_confirmation_email;
+
         my @dear;
         my %reporters = ();
         my $skip = 0;
         while (my $body = $bodies->next) {
+            # See if this body wants confirmation email (in case report made on national site, for example)
+            if (my $cobrand_body = $body->get_cobrand_handler) {
+                if (my $id_ref = $cobrand_body->report_sent_confirmation_email) {
```
davea: Any chance either of these conditions would fail, and `$send_confirmation_email` would have the value set in the previous loop iteration?

dracos: I think that's okay? We want it to be true if the logged-cobrand has it set, or if any body to which it's being sent (that has a cobrand) has it set.

PR #2433, opened 2019-03-18
Duplicate report feedback

dracos:
[skip changelog]

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1304

### DIFF
--- a/web/js/duplicates.js
+++ b/web/js/duplicates.js
@@ -1,4 +1,4 @@
-$(function() {
+(function() {
 
     // Store a reference to the "duplicate" report pins so we can
     // quickly remove them when we’re finished showing duplicates.
@@ -9,13 +9,14 @@ $(function() {
     var report_id = $("#report_inspect_form .js-report-id").text() || undefined;
 
     function refresh_duplicate_list() {
-        // This function will return a jQuery Promise, so callbacks can be
-        // hooked onto it, once the ajax request as completed.
-        var dfd = $.Deferred();
+        var category = $('select[name="category"]').val();
+        if (category === '-- Pick a category --') {
+            return;
+        }
 
         var nearby_url;
         var url_params = {
-            filter_category: $('select[name="category"]').val(),
+            filter_category: category,
             latitude: $('input[name="latitude"]').val(),
             longitude: $('input[name="longitude"]').val()
         };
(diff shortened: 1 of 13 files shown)


web/js/duplicates.js
```diff
@@ -188,6 +194,7 @@ $(function() {
     // Want to show potential duplicates when a regular user starts a new
     // report, or changes the category/location of a partial report.
     $("#problem_form").on("change.category", "select#form_category", problem_form_category_change);
+    $("select#form_category").change(); // In case on page load something is selected
```
zarino: I wonder whether this line should be in a separate commit to the other changes?

This line does the actual " Look up duplicate in case onload."

Whereas the other lines—I think?—fix a different bug, which was that the `nearby_url` API was getting queried even if `$('select[name="category"]').val()` was `-- Pick a category --` (which was obviously inefficient, so makes sense to resolve the defer immediately, as you’ve now done).

Or have I missed something?

web/js/duplicates.js
```diff
@@ -13,9 +13,15 @@ $(function() {
         // hooked onto it, once the ajax request as completed.
         var dfd = $.Deferred();
 
+        var category = $('select[name="category"]').val();
+        if (category === '-- Pick a category --') {
```
zarino: My Perl is a little rusty, but I’m assuming the `_('-- Pick a category --')` in `DB/Result/Problem.pl` and `Controller/Report/New.pl` mean this string can be translated?

If so, should we somehow check for the (potentially translated) string here, in the JavaScript, rather than just the English? Could do it with a `data-` attribute containing the "default" value to check against, perhaps?

dracos: Pick a category is already buggy (see it's not translated anywhere in fixmystreet.js) so can be fixed separately (especially as this isn't generically used code at present).

web/js/duplicates.js
dracos: Yeah, could be; I only noticed because of it happening onload after this change, but it could happen if someone changed back to Pick as well, yes.

dracos:
@zarino Have pushed a commit to bring the roadworks/ "Pick an asset"/ duplicates/ extra questions all to the same place.

dracos:
Slight issue because Oxfordshire puts its ROW message in category_meta, then switches off hide-if-invalid-category and this PR puts one inside the other. It shouldn't really be using category_meta, I did an emergency commit on commercial-staging that prepends it to #js-post-category-messages instead and I think that would be okay, just to remind me to add that here in the morning.

PR #2434, opened 2019-03-19
[Oxfordshire] correctly display PRoW message

struan:
This makes sure that we re-enable the report form if you select a non
OCC category after selecting PRoW, and also enables correct hiding and
displaying of the message.

Please check the following:


dracos:
This doesn't work on /around, only on /report/new. There's no category_meta to attach to coming from /around.

PR #2437, opened 2019-03-25
Northants cobrand

struan:


struan:
I've merged this in to the alloy-vector-layers branch for ease of merging later.

PR #2442, opened 2019-03-26
[Buckinghamshire] Message on Rights of Way category.

dracos:
[skip changelog]
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1326


dracos:
Ah! Yes, actually I did make something similar in 84e9f39efcf0bbd3c8d71e7fde4678f7356b7159 when merging the staging branches (overlaps with duplicate reporting feedback PR too), so I'll make sure that all tallies.

dracos:
Okay, should be similar to Oxfordshire now; and then can tidy them both up in #2433

PR #2445, opened 2019-03-27
make sure the archive report script sets up lang etc

struan:
Previously if no reports were closed with an email this meant that the
lang setup never ran and you got an error when _ was called in
close_problems. So, we call the lang setup in close_problems too.

Please check the following:

Thanks for contributing!

### DIFF
--- a/perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
+++ b/perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
@@ -138,8 +138,16 @@ sub close_problems {
     return unless $opts->{commit};
 
     my $problems = shift;
+
+    return unless $problems->count;
+
     my $extra = { auto_closed_by_script => 1 };
     $extra->{is_superuser} = 1 if !$opts->{user_name};
+
+    # need to do this again in case no reports were closed with an
+    # email in which case we won't have set the lang and domain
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -141,6 +141,11 @@ sub close_problems {
     my $extra = { auto_closed_by_script => 1 };
     $extra->{is_superuser} = 1 if !$opts->{user_name};
 
+    # need to do this again in case no reports were closed with an
+    # email in which case we won't have set the lang and domain
+    my $cobrand = FixMyStreet::Cobrand->get_class_for_moniker($opts->{cobrand})->new();
+    $cobrand->set_lang_and_domain($problems->first->lang, 1);
```
davea: This will fail in the situation that `$problems->first` doesn't exist. Better to reuse the `my @problems = $problems->all; return if scalar(@problems) == 0;` pattern from `send_email_and_close` above?

davea: (or more simply `return unless $problems->count;` perhaps?)

struan: Ah yes, good point.

PR #2446, opened 2019-03-28
[Northamptonshire] Hide form on emergency categories

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1296
[skip changelog]

### DIFF
--- a/web/cobrands/northamptonshire/assets.js
+++ b/web/cobrands/northamptonshire/assets.js
@@ -483,4 +483,30 @@ fixmystreet.assets.add($.extend(true, {}, northants_road_defaults, {
     ]
 }));
 
+// Hide form when emergency category used
+function check_emergency() {
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, northants_defaults.body) > -1;
+    var relevant_cat = !!$('label[for=form_emergency]').length;
+    var relevant = relevant_body && relevant_cat;
+    var currently_shown = !!$('#northants-emergency-message').length;
+    var body = $('#form_category').data('body');
+
+    if (relevant === currently_shown || body) {
+        // Either should be shown and already is, or shouldn't be shown and isn't
+        return;
+    }
+
+    if (!relevant) {
+        $('#northants-emergency-message').remove();
+        $('.js-hide-if-invalid-category').show();
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


web/cobrands/northamptonshire/assets.js
```diff
@@ -483,4 +483,30 @@ fixmystreet.assets.add($.extend(true, {}, northants_road_defaults, {
     ]
 }));
 
+// Hide form when emergency category used
+function check_emergency() {
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, northants_defaults.body) > -1;
+    var relevant_cat = $('label[for=form_emergency]').length;
+    var relevant = relevant_body && relevant_cat;
+    var currently_shown = !!$('#northants-emergency-message').length;
+    var body = $('#form_category').data('body');
```
struan: This does mean if someone who has a council login for another council uses this they'll still be able to report. I'm not sure that's much of an issue in practice though.

web/cobrands/northamptonshire/assets.js
```diff
@@ -483,4 +483,30 @@ fixmystreet.assets.add($.extend(true, {}, northants_road_defaults, {
...
+    }
+
+    if (!relevant) {
+        $('#northants-emergency-message').remove();
+        $('.js-hide-if-invalid-category').show();
+        return;
+    }
+
+    var $msg = $('<div class="box-warning" id="northants-emergency-message"></div>');
+    $msg.html($('label[for=form_emergency]').html());
```
struan: I still seem to get both this added message and the standard extra information displayed.

Screenshot 2019-03-29 at 12 27 31

dracos: Oh right, yes, this will only be hidden once #2433 is merged (which I should do now) because that then puts the extra questions inside `js-hide-if-invalid-category`.

dracos: That's now merged, I've rebased this on top so this should be okay now.

struan: One other thing, if I select repeated emergency categories I see the message multiple times, so e.g. if I select Drains -> Blocked flooding private property and then Blocked flooding road/path I get the message twice.

dracos: Grr, stupid booleans. Should be okay now!

PR #2447, opened 2019-03-29
Fix superusers creating anonymous reports.

### LINKED ISSUE #2435: Admin user not associated with body cannot create anonymous report
If a superuser that is not associated with a body tries to create an anonymous report it generates an error:

    Caught exception in FixMyStreet::App::Controller::Report::New->process_user "Can't call method "name" on an undefined value at /data/vhost/staging.fixmystreet.com/fixmystreet/perllib/FixMyStreet/App/Controller/Report/New.pm line 821

To recreate:
1: Log in as an super user
2: Remove any body association
3: Create a report and select "Anonymous user" from the "Report as" dropdown
4: Submit report

dracos:
Fixes #2435.

### DIFF
--- a/t/app/controller/report_as_other.t
+++ b/t/app/controller/report_as_other.t
@@ -16,6 +16,8 @@ my $test_email = 'body-user@example.net';
 my $user = $mech->log_in_ok($test_email);
 $user->update({ from_body => $body->id, name => 'Body User' });
 
+my $superuser = $mech->create_user_ok('superuser@example.net', name => "Super", is_superuser => 1);
+
 my ($report_to_update) = $mech->create_problems_for_body(1, $body->id, 'Title', { category => 'Potholes' });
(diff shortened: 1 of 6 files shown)


t/app/controller/report_as_other.t
```diff
@@ -150,19 +152,20 @@ subtest "Body user, has permission to add report as another (existing) user with
     push @users, $report->user;
 };
 
-subtest "Body user, has permission to add report as anonymous user" => sub {
+subtest "Superuser, can add report as anonymous user" => sub {
```
struan: Does this not mean we are losing the test coverage for body users reporting as anonymous users?

dracos: Sort of, explicitly, but I figured that the functionality was covered by this test in combination with others.

PR #2448, opened 2019-04-02
[UK] 2019 structural changes

dracos:
Keep SW&T sending to separate addresses when we create shared body. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/MapIt.pm
+++ b/perllib/FixMyStreet/MapIt.pm
@@ -1,0 +1,18 @@
+package FixMyStreet::MapIt;
+
+use FixMyStreet;
+use mySociety::MaPit;
+
+sub call {
+    my ($url, $params, %opts) = @_;
+
+    # 'area' always returns the ID you provide, no matter its generation, so no
+    # point in specifying it for that. 'areas' similarly if given IDs, but we
+    # might be looking up types or names, so might as well specify it then.
+    $opts{generation} = FixMyStreet->config('MAPIT_GENERATION')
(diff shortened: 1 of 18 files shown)


perllib/FixMyStreet/MapIt.pm
```diff
@@ -0,0 +1,15 @@
+package FixMyStreet::MapIt;
+
+use FixMyStreet;
+use mySociety::MaPit;
+
+sub call {
+    my ($url, $params, %opts) = @_;
+
+    $opts{generation} = FixMyStreet->config('MAPIT_GENERATION')
+        if $url ne 'area' && FixMyStreet->config('MAPIT_GENERATION');
```
struan: A comment to explain why we don't use generation for area would be helpful.

perllib/FixMyStreet/SendReport/Email.pm
```diff
@@ -111,4 +111,12 @@ sub send {
...
 }
 
+sub _get_district_for_contact {
+    my ( $lat, $lon ) = @_;
+    my $district =
+      FixMyStreet::MapIt::call( 'point', "4326/$lon,$lat", type => 'DIS' );
+    ($district) = keys %$district;
+    return $district;
+}
+
```
struan: Should this not be in the following commit?

dracos: Would make more sense, yeah (though I shouldn't have removed it in a978974b5770ebc173d302034cbe179a45cb40d5 and ideally would have force pushed this but bit late now :) ).

PR #2449, opened 2019-04-02
[Northamptonshire] prevent report submission without asset selection

struan:
Prevent users from making report unless they have selected an asset from the map.

Please check the following:

Thanks for contributing!

### DIFF
--- a/templates/web/northamptonshire/report/new/roads_message.html
+++ b/templates/web/northamptonshire/report/new/roads_message.html
@@ -1,0 +1,19 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <div id="js-not-an-asset" class="hidden js-responsibility-message">
+        <p>Please select <span id="js-roads-asset">an item</span> from the map on which to make a report.</p>
+    </div>
+    <div id="js-not-a-speedhump" class="hidden js-responsibility-message">
+        <p>Please select <b>a speed hump</b> from the map on which to make a report.</p>
+    </div>
+    <div id="js-not-a-ped-barrier" class="hidden js-responsibility-message">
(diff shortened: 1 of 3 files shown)


web/cobrands/northamptonshire/assets.js
```diff
@@ -473,47 +548,56 @@ fixmystreet.assets.add($.extend(true, {}, fixmystreet.assets.alloy_defaults, {
...
       layerVersion: '173.1-',
     },
     stylemap: new OpenLayers.StyleMap({
         'default': prow_style
     }),
-    body: "Northamptonshire County Council",
-    road: true,
-    always_visible: false,
-    non_interactive: true,
+    no_asset_msg_id: "#js-not-a-prow",
```
dracos: This doesn't exist in the HTML.

templates/web/northamptonshire/report/new/roads_message.html
```diff
@@ -0,0 +1,15 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
```
dracos: Equivalent messages on .com need to be made presumably so that they display there, currently just getting "not maintained". Check doesn't interfere with Bucks similar.

templates/web/northamptonshire/report/new/roads_message.html
```diff
@@ -8,6 +8,10 @@
     <div id="js-not-a-ped-barrier" class="hidden js-responsibility-message">
         <p>Please select <b>a pedestrian barrier</b> from the map on which to make a report.</p>
     </div>
+    <div id="js-not-a-prow" class="hidden js-responsibility-message">
+        <p>The location you have selected doesn't appear to be on a right of way.</p>
```
dracos: Can we use `’` everywhere?

PR #2452, opened 2019-04-03
[Northamptonshire] add streetlight category message

struan:
Fixes mysociety/fixmystreet-commercial#1343

Please check the following:

Thanks for contributing!

### DIFF
--- a/web/cobrands/northamptonshire/assets.js
+++ b/web/cobrands/northamptonshire/assets.js
@@ -343,6 +343,7 @@ $(fixmystreet).on('report_new:highways_change', function() {
     if (fixmystreet.body_overrides.get_only_send() === 'Highways England') {
         hide_responsibility_errors();
         enable_report_form();
+        $('#ncc_streetlights').remove();
     } else {
         $(fixmystreet).trigger('report_new:category_change', [ $('#form_category') ]);
     }
@@ -422,7 +423,7 @@ var northants_defaults = $.extend(true, {}, fixmystreet.assets.alloy_defaults, {
       var emergency_state = ncc_is_emergency_category();
 
       disable_report_form();
-      if (!emergency_state.relevant || emergency_state.body) {
+      if ((!emergency_state.relevant || emergency_state.body) && this.visibility) {
           show_responsibility_error('#js-not-an-asset', this.fixmystreet.asset_item, this.fixmystreet.asset_type);
       } else {
           hide_responsibility_errors();
@@ -627,5 +628,26 @@ function check_emergency() {
     $('.js-hide-if-invalid-category').hide();
 }
 $(fixmystreet).on('report_new:category_change', check_emergency);
+
+function ncc_check_streetlights() {
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, northants_defaults.body) > -1;
+    var relevant_cat = $('#form_category').val() == 'Street lighting';
+    var relevant = relevant_body && relevant_cat;
+    var currently_shown = !!$('#ncc_streetlights').length;
+
+    if (relevant === currently_shown || fixmystreet.body_overrides.get_only_send() == 'Highways England') {
+        return;
+    }
+
+    if (!relevant) {
+        $('#ncc_streetlights').remove();
+        return;
+    }
+
+    var $msg = $('<p id="ncc_streetlights" class="box-warning">Street lighting in Northamptonshire is maintained by Balfour Beatty on behalf of the County Council under a Street Lighting Private Finance Initiative (PFI) contract. Please view our <b><a href="https://www3.northamptonshire.gov.uk/councilservices/northamptonshire-highways/roads-and-streets/Pages/street-lighting.aspx">Street Lighting</a></b> page to report any issues.</p>');
+    $msg.insertBefore('#js-post-category-messages');
+    disable_report_form();
+}
+$(fixmystreet).on('report_new:category_change', ncc_check_streetlights);
 
 })();
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -405,6 +405,8 @@ function layer_visibilitychanged() {
             this.road_not_found();
         }
         return;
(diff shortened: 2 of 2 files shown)


web/cobrands/northamptonshire/assets.js
```diff
@@ -516,4 +516,25 @@ fixmystreet.assets.add($.extend(true, {}, fixmystreet.assets.alloy_defaults, {
     }
 }));
 
+function ncc_check_streetlights() {
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, 'Northamptonshire County Council') > -1;
```
dracos: If you put `body: "Northamptonshire County Council"`  in `northants_defaults`, then you could use `northants_defaults.body` here.

web/cobrands/northamptonshire/assets.js
```diff
@@ -516,4 +516,25 @@ fixmystreet.assets.add($.extend(true, {}, fixmystreet.assets.alloy_defaults, {
...
+        // Either should be shown and already is, or shouldn't be shown and isn't
+        return;
+    }
+
+    if (!relevant) {
+        $('#ncc_streetlights').remove();
+        return;
+    }
+
+    var $msg = $('<p id="ncc_streetlights" class="box-warning">Street lighting in Northamptonshire is maintained by Balfour Beatty on behalf of the County Council under a Street Lighting Private Finance Initiative (PFI) contract. Please view our <a href="https://www3.northamptonshire.gov.uk/councilservices/northamptonshire-highways/roads-and-streets/Pages/street-lighting.aspx">Street Lighting page</a> to report any issues');
```
dracos: Missing an end `</p>`?

web/cobrands/northamptonshire/assets.js
```diff
@@ -618,4 +619,33 @@ function check_emergency() {
...
+        $('#ncc_streetlights').remove();
+        return;
+    }
+
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, northants_defaults.body) > -1;
+    var relevant_cat = $('#form_category').val() == 'Street lighting';
+    var relevant = relevant_body && relevant_cat;
+    var currently_shown = !!$('#ncc_streetlights').length;
+
+  console.log(relevant, currently_shown);
```
dracos: Debug left in

web/cobrands/northamptonshire/assets.js
```diff
@@ -618,4 +619,33 @@ function check_emergency() {
...
+        $('#ncc_streetlights').remove();
+        enable_report_form();
+        return;
+    }
+
+    if (!currently_shown) {
+        var $msg = $('<p id="ncc_streetlights" class="box-warning">Street lighting in Northamptonshire is maintained by Balfour Beatty on behalf of the County Council under a Street Lighting Private Finance Initiative (PFI) contract. Please view our <b><a href="https://www3.northamptonshire.gov.uk/councilservices/northamptonshire-highways/roads-and-streets/Pages/street-lighting.aspx">Street Lighting</a></b> page to report any issues.</p>');
+        $msg.insertBefore('#js-post-category-messages');
+        disable_report_form();
+        hide_responsibility_errors();
```
dracos: Is this line necessary? Isn't it already handled by the removal of the asset/road layer from the category change?

web/cobrands/northamptonshire/assets.js
```diff
@@ -618,4 +619,33 @@ function check_emergency() {
...
+
+    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, northants_defaults.body) > -1;
+    var relevant_cat = $('#form_category').val() == 'Street lighting';
+    var relevant = relevant_body && relevant_cat;
+    var currently_shown = !!$('#ncc_streetlights').length;
+
+  console.log(relevant, currently_shown);
+    if (!relevant) {
+        $('#ncc_streetlights').remove();
+        enable_report_form();
```
dracos: This worries me due to e.g. changing to a category that would need to hide it again (not on a road or something) but this running after, but as it's all hopefully refactoring I'm not too fussed!

web/cobrands/northamptonshire/assets.js
```diff
@@ -618,4 +619,33 @@ function check_emergency() {
...
+        return;
+    }
+
+    if (!currently_shown) {
```
dracos: Might be nicer (slash easier for me to include in refactor :) ) if it was consistent with the other calls which return first if (relevant === currently_shown). Then this line wouldn't be needed, I think? Unless there was a reason for not doing that.

struan:
Right, I see my error here was trying to take into account the emergency state which is irrelevant as by definition if you've selected street lighting then it's not an emergency. Removing that cleans up the code and removes all the weird edge cases I was seeing.

web/cobrands/northamptonshire/assets.js
```diff
@@ -423,7 +423,7 @@ var northants_defaults = $.extend(true, {}, fixmystreet.assets.alloy_defaults, {
       var emergency_state = ncc_is_emergency_category();
 
       disable_report_form();
-      if (!emergency_state.relevant || emergency_state.body) {
+      if ((!emergency_state.relevant || emergency_state.body) && this.visibility) {
```
dracos: Probably need a similar update to Bucks otherwise the not_found there shows a "Pick an item" message :)

struan: I'm not sure I follow where this is a problem. Can you give me an example?

dracos: I guess it instantly gets removed after, doesn't it. My thought process was (on Bucks) if you switch from e.g. Bollard light not working to non-asset category, `layer_visibilitychanged` is called, which now calls asset_not_found which calls `labeled_defaults.actions.asset_not_found` which sets the message back to the default. It doesn't really need to as it then gets removed, but then that doesn't matter if it does. Hopefully might get tidied up too.

dracos:
@struan I've pushed a fix I think for that that also makes the fn almost identical to road not_found, which makes it feel right?

struan:
Yup, that makes sense.

PR #2456, opened 2019-04-08
Message controller

dracos:
The first commit I think is safe and could be taken alone; am working on second commit more/testing needed, but hopefully along the right lines, I think it could sort itself out better but I tried for now to keep the business code as similar as possible, maybe. [skip changelog]

### DIFF
--- a/web/cobrands/buckinghamshire/assets.js
+++ b/web/cobrands/buckinghamshire/assets.js
@@ -465,27 +465,10 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
     }
 }));
 
-function check_rights_of_way() {
-    var relevant_body = OpenLayers.Util.indexOf(fixmystreet.bodies, defaults.body) > -1;
-    var relevant_cat = $('#form_category').val() == 'Rights of Way';
-    var relevant = relevant_body && relevant_cat;
-    var currently_shown = !!$('#row-message').length;
-
-    if (relevant === currently_shown) {
-        // Either should be shown and already is, or shouldn't be shown and isn't
-        return;
-    }
-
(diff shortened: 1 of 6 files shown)


web/cobrands/buckinghamshire/assets.js
```diff
@@ -352,54 +312,31 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
     all_categories: true,
     actions: {
         found: function(layer, feature) {
-            fixmystreet.body_overrides.allow_send(layer.fixmystreet.body);
-            fixmystreet.body_overrides.remove_only_send();
+            var map = {
+                "HE": '#js-not-council-road-he',
+                "HWOA": '#js-not-council-road-other'
+            };
+            var bad_id = map[feature.attributes.feature_ty] || '#js-not-council-road';
```
struan: I am not sure about bad_id as a name. msg_id would make more sense?

web/cobrands/northamptonshire/assets.js
```diff
@@ -617,49 +617,21 @@ function ncc_is_emergency_category() {
     return {relevant: relevant, currently_shown: currently_shown, body: body};
 }
```
struan: Presumably this should be removed in this commit and not the next.

dracos: Can't easily tell what this is referring to sorry; if it's ncc_is_emergency_category, that is still called somewhere, e.g. asset_found, until the last commit of this PR (when it's removed)?

struan: Yes, so it is. I missed that.

PR #2457, opened 2019-04-08
[Northamptonshire] use the correct layer environment for assets

struan:
The tile URL includes a parameter for the client number which we were
not using so configure this based on live or staging.

Please check the following:

Thanks for contributing!

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/alloy.js
+++ b/web/cobrands/fixmystreet-uk-councils/alloy.js
@@ -28,7 +28,7 @@ OpenLayers.Protocol.Alloy = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
     },
 
     getURL: function(coords, options) {
-        return OpenLayers.String.format(options.base, {'layerid': options.layerid, 'layerVersion': options.layerVersion, 'z': 15, 'x': coords[0], 'y': coords[1]});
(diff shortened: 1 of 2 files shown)


web/cobrands/fixmystreet-uk-councils/alloy.js
```diff
@@ -117,7 +117,7 @@ OpenLayers.Strategy.Alloy = OpenLayers.Class(OpenLayers.Strategy.FixMyStreet, {
 
 fixmystreet.assets.alloy_defaults = {
     http_options: {
-      base: "https://alloy-api-tile01.yotta.co.uk/api/render-layer/tile/${layerid}/28/${layerVersion}-/${z}/${x}/${y}",
+      base: "https://alloy-api-tile01.yotta.co.uk/api/render-layer/tile/${layerid}/${environment}/${layerVersion}-/${z}/${x}/${y}",
```
davea: The values that get interpolated in to `${layerVersion}` include the `-` suffix, so you end up with a double-dash in the URL e.g. `https://alloy-api-tile01.yotta.co.uk/api/render-layer/tile/308/28/308.8--/15/16303/10788`. Doesn't seem to matter but might be worth fixing whilst here?

struan: oooh, good spot, will remove that too.

PR #2460, opened 2019-04-11
Miscellaneous Minor Improvements

tomhukins:
These changes should have no negative consequences for the code or its behaviour, but make it simpler or clearer.

### DIFF
--- a/perllib/Utils.pm
+++ b/perllib/Utils.pm
@@ -12,7 +12,6 @@ package Utils;
 use strict;
 use DateTime;
 use Encode;
-use File::Slurp qw();
 use mySociety::GeoUtil;
 use mySociety::Locale;
 use FixMyStreet;
--- a/cpanfile
+++ b/cpanfile
@@ -63,7 +63,6 @@ requires 'Error';
 requires 'FCGI'; # Required by e.g. Plack::Handler::FCGI
 requires 'File::Find';
 requires 'File::Path';
-requires 'File::Slurp';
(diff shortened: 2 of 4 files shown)


perllib/Utils.pm
```diff
@@ -12,7 +12,6 @@ package Utils;
 use strict;
 use DateTime;
 use Encode;
-use File::Slurp qw();
```
dracos: I guess we can remove this from the `cpanfile` as well :)

tomhukins: I wondered about that, but I also wondered whether to remove it from `cpanfile.snapshot` as well, then wondered whether unpinning its version from that file might cause problems.  In the end I decided not to overthink it and take the safe, sub-optimal approach.  If you're happy making that further removal, I cautiously approve though.

dracos: It would cause problems if removed from the snapshot, yes. `cpanfile` is the list of modules used by the project (similar to e.g. a Gemfile); `cpanfile.snapshot` is a cache of the exact versions installed of all modules and their dependencies (like Gemfile.lock). So removing it from `cpanfile` (as we're not using it any more ourselves) is fine but wouldn't change the snapshot, because other modules that we do use still depend upon it.

tomhukins: Thanks for the explanation.  I've worked with carton and cpanfile before, but the details never stick in my head as well as I'd hope.

dracos:
Thank you again!

PR #2466, opened 2019-04-29
Version 2.6

dracos:
- [x] Blog post
- [ ] AMI after


dracos:
Transifex doesn't store "fuzzy" ones (guesses as to new translations based upon old ones), it's easier just to scrub them then try and sort it out each time.

PR #2468, opened 2019-05-02
show open311 failure details in admin report edit page

struan:
Rather than having to go into the database to see any error details
display it in the admin. The error only shows the first few lines by
default and expands on hover.

Please check the following:

Thanks for contributing!

### DIFF
--- a/templates/web/base/admin/report_edit.html
+++ b/templates/web/base/admin/report_edit.html
@@ -87,6 +87,11 @@ class="admin-offsite-link">[% problem.latitude %], [% problem.longitude %]</a>
     <input type="submit" class="btn" name="mark_sent" value="[% loc('Mark as sent') %]">
   [% END %]
 </li>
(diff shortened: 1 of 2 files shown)


templates/web/base/admin/report_edit.html
```diff
@@ -87,6 +87,11 @@
     <input type="submit" class="btn" name="mark_sent" value="[% loc('Mark as sent') %]">
   [% END %]
 </li>
+[% IF c.user.is_superuser AND problem.send_fail_count > 0 %]
+<li class="sm">[% loc('Send Fail Count:') %] [% problem.send_fail_count %]</li>
+<li class="sm">[% loc('Last failure:') %] [% PROCESS format_time time=problem.send_fail_timestamp %]</li>
+<li class="sm truncate_height">[% loc('Reason:') %] [% problem.send_fail_reason %]</li>
```
davea: This should probably have a `| html` in case of rogue 311 endpoints :)

PR #2470, opened 2019-05-03
 [Open311] Allow desc in email template with tokens.

dracos:

### Requirements:

Be able to personalise the email so admins can add the reporter's name to the email .
Include the email template that is stored in FixMyStreet as well as the update that is pulled from the council integration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
           marked private on the site.
         - Add new upload_files flag which sends files/photos as part of the
           POST service request.
+        - Allow description in email template with placeholder.
 
 * v2.6 (3rd May 2019)
     - New features:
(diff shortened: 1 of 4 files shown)


LouiseMySociety:
Requirements: 

Be able to personalise the email so admins can add the reporter's name to the email . 
Include the email template that is stored in FixMyStreet as well as the update that is pulled from the council integration.

dracos:
Have changed this to only allow a `{{description}}` placeholder - the update text is used as the text of an update, it is public, and so can't include someone's name. Also, you may have more than one update, which is bundled up in one email - alert emails are sent on updates, they are independent of the contents of those updates.

If a council wants update emails to contain the reporter's name, then that is to be done in the alert-update template, either for that council or if we want it in general, in the default template. Our current alert-update template (and templates in general) doesn't really have a location to put the user's name.

PR #2473, opened 2019-05-07
Upgrade carton, Catalyst, and a number of other packages

dracos:
Including DateTime, DBIx::Class, Email::MIME, Moo, Moose, Number::Phone, Starman, Template, ...

This also fixes some warnings (in dev console and one test), and tests perl 5.26/5.28 on Travis.

The splitting of the coverage tests has led to a drop in coverage, because codecov does not appear to know whether two 'part-covered' branches merge to one 'covered' line (even though it looks like it sends "1/2" data, I guess it can't be sure it's not the same 1/2...), so if the two possibilities of an `if` are now being tested in different halves, it now says that is partial not full.


.travis.yml
```diff
@@ -35,6 +34,7 @@ env:
   global:
     - "S3_BUCKET=fixmystreet-bundle-cache"
     - "CYPRESS=0"
+    - "CATALYST_NOWARN_DEPRECATE=1"
```
struan: could we add a comment to say why this is here. Also add might be nice to add it to `scripts/test`.

t/app/controller/admin.t
```diff
@@ -25,7 +25,7 @@ my $dt = DateTime->new(
     second => 23
 );
 
-my $report = FixMyStreet::App->model('DB::Problem')->find_or_create(
+my $report = FixMyStreet::DB->resultset('Problem')->find_or_create(
```
struan: I'm not clear why this change is happening.

.travis.yml
```diff
@@ -49,14 +54,16 @@ addons:
...
 before_script:
   - commonlib/bin/gettext-makemo FixMyStreet
-  - 'if [ "$TRAVIS_PERL_VERSION" = "5.24" ]; then export HARNESS_PERL_SWITCHES="-MDevel::Cover=+ignore,local/lib/perl5,commonlib,perllib/Catalyst/[^A],perllib/DBIx,perllib/Email,perllib/Template,^t"; fi'
+  - 'if [ "$COVERAGE_PART" != "0" ]; then export HARNESS_PERL_SWITCHES="-MDevel::Cover=+ignore,local/lib/perl5,commonlib,perllib/Catalyst/[^A],perllib/Email,^t"; fi'
 script:
-  - 'if [ "$CYPRESS" = "0" ]; then script/test --jobs 3 t; fi'
+  - 'if [ "$CYPRESS" = "0" ] && [ "$COVERAGE_PART" = "0" ]; then script/test --jobs 3 t; fi'
+  - 'if [ "$COVERAGE_PART" = "1" ]; then script/test --jobs 3 `find t/app/controller -name "*.t"`; fi'
+  - 'if [ "$COVERAGE_PART" = "2" ]; then script/test --jobs 3 `find t -name "*.t" ! -path "t/app/controller*"`; fi'
   - 'if [ "$CYPRESS" = "1" ]; then PATH=$(npm bin):$PATH bin/browser-tests run ${CYPRESS_RECORD_KEY:+--record}; fi'
```
struan: Again, a comment to say why this is happening would be helpful

t/app/controller/admin.t
dracos: Shouldn't really be using the full App in a test if all we need is the DB schema, for performance, but in particular, it's because an upgraded package somewhere now throws a warning if you use App's find_or_create but use a User created via DB->resultset, saying that seems a bit odd.

PR #2474, opened 2019-05-09
[Buckinghamshire] use buckinghamshire map tiles

struan:
Update Buckinghamshire cobrand to use Buckinghamshire's own map tiles so that house numbers are viewable.

Please check the following:

Thanks for contributing!


perllib/FixMyStreet/Map/Buckinghamshire.pm
```diff
@@ -2,19 +2,88 @@
...
+sub copyright {
+    return '&copy; BCC';
+}
+
+sub map_template { 'buckinghamshire' }
+
 sub map_javascript { [
-    '/vendor/OpenLayers/OpenLayers.wfs.js',
-    '/vendor/OpenLayers.Projection.OrdnanceSurvey.js',
+    '/vendor/OpenLayers/OpenLayers.bristol.js',
```
davea: Could/should this file be renamed to something not Bristol-specific now it's being shared? (And same for `/data/bristol/openlayers.bristol.cfg`)

perllib/FixMyStreet/Map/Buckinghamshire.pm
```diff
@@ -2,19 +2,88 @@
...
+    my $params = {
+        zoom_levels    => scalar $self->scales,
+        default_zoom   => 5,
+        min_zoom_level => 0,
+        id_offset      => 0,
+    };
+    return $params;
+}
+
+sub tile_parameters {
```
dracos: Given how most of this fn is shared with Bristol (plus the reproject, zoom_parameters functions), worth having a parent class used by both?

struan:
I've fixed the roadworks stuff now, plus refactored some of the shared code.

PR #2477, opened 2019-05-14
Various small post NCC go live fixes

struan:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1323

NB: the cobrand specific client side tests need to be run manually at the moment:

./bin/browser-tests run --cobrand northamptonshire --area_id 2234 --coords '52.236251,0.892052'

Please check the following:

Thanks for contributing!

### DIFF
--- a/.cypress/cypress/integration/northamptonshire.js
+++ b/.cypress/cypress/integration/northamptonshire.js
@@ -1,85 +1,72 @@
-/*
- * These don't run by default, need to run the tests with
- * ./bin/browser-tests run --cobrand northamptonshire --area_id 2234 --coords '52.236251,0.892052'
- *
- * NB: most of the non northamptonshire tests will then fail
- */
+it('loads the right front page', function() {
+    cy.visit('http://northamptonshire.localhost:3001/');
+    cy.contains('Northamptonshire');
+});
 
-var cobrand = Cypress.env('cobrand');
-var only_or_skip = (cobrand == 'northamptonshire') ? describe.only : describe.skip;
+it('prevents clicking unless asset selected', function() {
+  cy.server();
+  cy.fixture('trees.json');
+  cy.fixture('trees_none.json');
+  cy.route('**/render-layer/**', 'fixture:trees_none.json').as('empty-trees-layer');
+  cy.route('**/16463/10787**', 'fixture:trees.json').as('trees-layer');
+  cy.route('/report/new/ajax*').as('report-ajax');
+  cy.visit('http://northamptonshire.localhost:3001/');
+  cy.get('[name=pc]').type('NN1 1NS');
+  cy.get('[name=pc]').parents('form').submit();
 
-only_or_skip('Northamptonshire cobrand specific testing', function() {
+  cy.get('#map_box').click();
+  cy.wait('@report-ajax');
 
-    it('loads the right front page', function() {
-        cy.visit('/');
-        cy.contains('Northamptonshire');
-    });
+  cy.get('[id=category_group]').select('Fallen Tree');
 
-    it('prevents clicking unless asset selected', function() {
-      cy.server();
-      cy.fixture('trees.json');
-      cy.fixture('trees_none.json');
-      cy.route('**/render-layer/**', 'fixture:trees_none.json').as('empty-trees-layer');
-      cy.route('**/16463/10787**', 'fixture:trees.json').as('trees-layer');
-      cy.route('/report/new/ajax*').as('report-ajax');
-      cy.visit('/');
-      cy.get('[name=pc]').type('NN1 1NS');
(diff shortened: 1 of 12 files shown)


t/cobrand/northamptonshire.t
```diff
@@ -0,0 +1,43 @@
...
+
+        $mech->follow_link_ok( { text_regex => qr/skip this step/i, },
+            "follow 'skip this step' link" );
+
+        $mech->content_contains('Trees');
+        $mech->content_lacks('Flytipping');
+    };
+};
+
+END {
```
dracos: I know we do this elsewhere, but we shouldn't, says https://perldoc.perl.org/Test/More.html#I-love-it-when-a-plan-comes-together

.cypress/cypress/integration/northamptonshire.js
```diff
@@ -0,0 +1,73 @@
+/*
+ * These don't run by default, need to run the tests with
+ * ./bin/browser-tests run --cobrand northamptonshire --area_id 2234 --coords '52.236251,0.892052
```
dracos: Missing quote

.cypress/cypress/integration/northamptonshire.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * NB: most of the non northamptonshire tests will then fail
```
dracos: You can run e.g. `bin/browser-tests run ... --spec .cypress/cypress/integration/northamptonshire.js` to only run the Northamptonshire file.

dracos: Although, shouldn't the `only_or_skip` below be suppressing all other tests in this instance?

.cypress/cypress/integration/northamptonshire.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    it('prevents clicking unless asset selected', function() {
+      cy.server();
+      cy.fixture('trees.json');
+      cy.fixture('trees_none.json');
+      cy.route('**/render-layer/**', 'fixture:trees_none.json').as('empty-trees-layer');
+      cy.route('**/16463/10787**', 'fixture:trees.json').as('trees-layer');
+      cy.route('/report/new/ajax*').as('report-ajax');
+      cy.visit('report/new?longitude=0.892052&latitude=52.236251');
```
dracos: Nice! Perhaps one of these could go to /around and click the map, or did that prove too tricky?

web/cobrands/fixmystreet-uk-councils/alloy.js
```diff
@@ -9,19 +10,24 @@ OpenLayers.Protocol.Alloy = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
         options = OpenLayers.Util.applyDefaults(options, this.options);
         var all_tiles = this.getTileRange_(options.scope.bounds, options.scope.layer.maxExtent, options.scope.layer.map);
         var rresp;
-        var start = new Date();
         var max = all_tiles.length;
-        options.scope.newRequest(start, max);
+        if (this.currentRequests.length) {
+            for (var j = 0; j < this.currentRequests.length; j++) {
+                this.currentRequests[j].priv.abort();
```
dracos: This looks okay; however, as the last `OpenLayers.Protocol.Response` is returned by this read, `abort()` has already been called on that one by the `OpenLayers.Protocol.HTTP` abort function. Would it be better for this class to override `abort()` and do this abort loop there?

.cypress/cypress/integration/northamptonshire.js
dracos: I guess that is only per-file or something and I didn't realise. Could add similar to the other test files I guess easily enough.
Or better, think of a way of having them all run - I assume I couldn't find a way to change the host name :-/

web/cobrands/fixmystreet-uk-councils/alloy.js
struan: I'm not sure I follow this.

dracos: This `read` function returns the final `rresp` (and final entry in currentRequests) of the loop. OpenLayers.Strategy.BBOX stores that in `this.response` (in triggerRead). In a subsequent call to triggerRead, as `this.response` is set, it calls abort on the layer's protocol with that response: `this.layer.protocol.abort(this.response);` which is Protocol.HTTP which calls `response.priv.abort();` on it. It then gets to this read which will call abort() on all the currentRequests, including the one that has already had abort called on it.

Instead, if you overrode abort() in this class, that could do the loop and calling of al the priv.aborts() and read() wouldn't have to concern itself with aborting anything, which isn't really its job.

PR #2479, opened 2019-05-16
[Buckinghamshire] fix zoom level issues

struan:
Allows assets to display when zoomed in.
Remove very zoomed out views as not useful.

Please check the following:

### DIFF
--- a/web/cobrands/buckinghamshire/assets.js
+++ b/web/cobrands/buckinghamshire/assets.js
@@ -16,8 +16,11 @@ var defaults = {
     },
     format_class: OpenLayers.Format.GML.v3.MultiCurveFix,
     asset_type: 'spot',
-    max_resolution: 4.777314267158508,
-    min_resolution: 0.5971642833948135,
+    max_resolution: {
+      'buckinghamshire': 2.116670900008467,
+      'fixmystreet': 4.777314267158508
+    },
+    min_resolution: 0.00001,
     asset_id_field: 'central_as',
     attributes: {
         central_asset_id: 'central_as',


perllib/FixMyStreet/Map/Buckinghamshire.pm
```diff
@@ -15,11 +15,6 @@ sub layer_names{ [ 'Basemapping2018' ] }
 sub scales {
     my $self = shift;
     my @scales = (
-        '1000000',
-        '500000',
```
dracos: You need all these (bar maybe 1M) for https://fixmystreet.buckscc.gov.uk/reports/Buckinghamshire I think?

web/cobrands/buckinghamshire/assets.js
```diff
@@ -17,7 +17,7 @@ var defaults = {
     format_class: OpenLayers.Format.GML.v3.MultiCurveFix,
     asset_type: 'spot',
     max_resolution: 4.777314267158508,
```
dracos: See e.g. something like 88f8334157416f33d0bd81a38b46b25005cbbc15 for if we don't want the assets to appear wildly more zoomed out than normal on the Bucks cobrand?

perllib/FixMyStreet/Map/Buckinghamshire.pm
```diff
@@ -6,7 +6,7 @@ use base 'FixMyStreet::Map::UKCouncilWMTS';
 
 use strict;
 
-sub default_zoom { 8; }
+sub default_zoom { 4; }
```
dracos: I think now the scales are back, this won't want to be this zoomed out?

web/js/map-wmts-buckinghamshire.js
```diff
@@ -137,11 +137,9 @@ fixmystreet.maps.config = function() {
 
 fixmystreet.maps.marker_size = function() {
     var zoom = fixmystreet.map.getZoom() + fixmystreet.zoomOffset;
-    if (zoom >= 7) {
+    if (zoom >= 1) {
```
dracos: Similarly here, presumably. Could leave 1M in, leave these alone and only change min/max_resolution? :)

struan: yes, good point. done now.

Issue #2475, opened 2019-05-13
Allow categories to appear under multiple groups


struan:
Where categories are grouped it would be helpful to be able to place a category under multiple groups for categories which could be related to more than one. This will enable a category to be listed under all the groups that a user might expect it to appear, hence reducing the likelihood of users trying multiple groups before finding the category they want.

e.g _Flytipping_ might appear under both a _Rubbish and Litter_ group and an _Environmental Issues_ group.

The category would be the same in each location it appeared.

struan:
Parts of plan are:
- Allow multiple groups in open311 services response
- Store multiple groups in FMS DB
- Show category in multiple groups on front end

struan:
For the multiple groups in open311 bit of this it's not totally obvious how to do this.

In theory we should be using the `keywords` bit of the service description for extra group for the extra groups, although the spec is a bit hazy on that. It's still going to be plain text though unless we wanted to extend things. I guess we could just specify CSV escaping and then use that when importing it.

dracos:
The code is live, and activated on staging for BANES, emailed James about it. So considering this 'done' in that it can be switched on for any council that wants it now.

PR #2480, opened 2019-05-16
display category under multiple groups

struan:
The group metadata of a contact can now be a comma separated list of
groups under which to display the category

For #2475

Please check the following:

Thanks for contributing!

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -273,8 +273,13 @@ sub update_contacts : Private {
         if ( $c->get_param('reputation_threshold') ) {
             $contact->set_extra_metadata( reputation_threshold => int($c->get_param('reputation_threshold')) );
         }
-        if ( my $group = $c->get_param('group') ) {
-            $contact->set_extra_metadata( group => $group );
+        if ( my @group = $c->get_param_list('group') ) {
+            @group = grep { $_ } @group;
+            if (scalar @group == 0) {
+                $contact->unset_extra_metadata( 'group' );
+            } else {
+                $contact->set_extra_metadata( group => \@group );
+            }
         } else {
             $contact->unset_extra_metadata( 'group' );
         }
--- a/perllib/Open311/PopulateServiceList.pm
+++ b/perllib/Open311/PopulateServiceList.pm
@@ -2,6 +2,7 @@ package Open311::PopulateServiceList;
 
 use Moo;
 use Open311;
+use Text::CSV;
 
 has bodies => ( is => 'ro' );
 has found_contacts => ( is => 'rw', default => sub { [] } );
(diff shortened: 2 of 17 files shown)


web/cobrands/fixmystreet/admin.js
```diff
@@ -139,6 +139,20 @@ $(function(){
         return true;
     });
 
+    $(".js-group-item-add").on("click", function(e) {
+        e.preventDefault();
+        var $template_item = $(".js-group-item-template");
+        var $new_item = $template_item.clone();
+        $new_item.removeClass("hidden-js js-group-item-template");
+        $new_item.show();
```
dracos: I don't think you need this line, or therefore also the next removeAttr line. You've already removed the hidden class, so then the insertBefore adds it to the document and there it is.

templates/web/base/admin/bodies/contact-form.html
```diff
@@ -132,7 +132,21 @@ <h1>[% contact.category_display | html %]</h1>
     <p>
       <label>
         [% loc('Group') %]
+        [% IF body.get_cobrand_handler.enable_multiple_category_groups %]
+            [% IF contact.extra.group %]
+                [% FOR group IN contact.extra.group %]
+                <input class="form-control" type="text" name="group" value="[% group | html %]" size="30">
+                [% END %]
+            [% ELSE %]
+                <input class="form-control" type="text" name="group" value="[% contact.extra.group | html %]" size="30">
```
dracos: `contact.extra.group` will always be blank if it's got here.

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -273,8 +273,15 @@ sub update_contacts : Private {
...
-        if ( my $group = $c->get_param('group') ) {
-            $contact->set_extra_metadata( group => $group );
+        if ( my @group = $c->get_param_list('group') ) {
+            @group = grep { /.+/ } @group;
+            if (scalar @group == 0) {
+                $contact->unset_extra_metadata( 'group' );
+            } elsif (scalar @group > 1) {
+                $contact->set_extra_metadata( group => \@group );
+            } else {
+                $contact->set_extra_metadata( group => $group[0] );
```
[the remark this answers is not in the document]
struan: er, no, I guess not.

t/app/controller/admin/bodies.t
```diff
@@ -2,8 +2,6 @@ package FixMyStreet::Cobrand::Tester;
 
 use parent 'FixMyStreet::Cobrand::Default';
 
-sub enable_category_groups { 1 }
```
dracos: Don't need this whole package now.

struan: Sadly it's required so we can both have a cobrand for setting the features on and for the `MAPIT_TYPES` override to work so the `Mock::MapIt` call works. If you use any other cobrand then the internal `MAPIT_TYPES` setting of the cobrand overrides the override.

dracos: Surely the 'default' cobrand would be fine as it doesn't set one, like the UK ones? But fine, doesn't matter :)

perllib/Open311/PopulateServiceList.pm
```diff
@@ -276,7 +276,7 @@ sub _set_contact_group {
         }
     }
 
-    if ($old_group ne $new_group) {
+    if ($self->_groups_different($old_group, $new_group)) {
         if ($new_group) {
             $contact->set_extra_metadata(group => $new_group);
```
dracos: Should this always store arrayref to match admin.

perllib/Open311/PopulateServiceList.pm
```diff
@@ -295,6 +295,23 @@ sub _set_contact_group {
     }
 }
 
+sub _groups_different {
+    my ($self, $old, $new) = @_;
+
+    my $diff = 1;
+    if ($old && $new) {
```
dracos: Both old and new at this point could be a string or an arrayref. The comparison won't work if e.g. the current metadata is ["Streets"] and the fetched data is "Streets". Perhaps easier to make sure old/new are both arrayrefs and then compare?

perllib/Open311/PopulateServiceList.pm
dracos: Or I guess you change it back in admin so that if one group it's always a string, not a ref of one element. I thought 'always an arrayref' would be easier, but given my other comment below, perhaps it is not.

perllib/Open311/PopulateServiceList.pm
dracos: I did this.

PR #2481, opened 2019-05-16
Bexley changes

dracos:
[skip changelog]
This changes which categories display which assets, and shows the asset ID on the pin/sidebar.

### DIFF
--- a/web/cobrands/bexley/js.js
+++ b/web/cobrands/bexley/js.js
@@ -55,7 +55,7 @@ fixmystreet.assets.add(defaults, {
     attributes: {
         UnitID: 'Unit_ID'
     },
-    asset_category: ["Damaged sign, bollard etc."],
+    asset_category: ["Keep left bollard"],
     asset_item: 'bollard'
 });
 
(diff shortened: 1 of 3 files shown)


web/cobrands/bexley/js.js
```diff
@@ -128,5 +128,27 @@ fixmystreet.assets.add(labeled_defaults, {
...
+
+fixmystreet.message_controller.register_category({
+    body: defaults.body,
+    category: function() {
+        var cat = $('#form_category').val();
+        if (cat === 'Dead animal') {
+            var where = $('#form_DALocation').val();
+            if (where === 'Garden' || where === 'Other private property') {
+                return true;
+            }
```
struan: Once you select one of these options there's no way to change it without clicking the map again which is a bit annoying.

dracos: Isn't the category box still visible too? I guess I didn't want someone just changing the subcategory in order to get it through - though you raise an important point that the link it mentions won't be visible any more, whoops, so it should somehow leave it visible..!

PR #2482, opened 2019-05-17
Small improvements from Hounslow branch

davea:
A few small things that have come from the Hounslow cobrand work.

 - Support `<keywords>private</keywords>` in Open311 service definitions to mark that contact as non_public
 - Ensure non_public reports aren't included in results when searching by ref from front page
 - Ensure reports fetched via Open311 are marked private if the category they're in is also marked private. NB this situation shouldn't arise through normal use, it's just a belt-and-braces check that we don't expose things that shouldn't be public.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -416,7 +416,7 @@ sub lookup_by_ref : Private {
             external_id => $ref
         ];
 
-    my $problems = $c->cobrand->problems->search( $criteria );
+    my $problems = $c->cobrand->problems->search({ non_public => 0, -or => $criteria });
 
     my $count = try {
         $problems->count;
--- a/perllib/Open311/GetServiceRequests.pm
+++ b/perllib/Open311/GetServiceRequests.pm
@@ -157,6 +157,7 @@ sub create_problems {
         my $state = $open311->map_state($request->{status});
 
         my $non_public = $request->{non_public} ? 1 : 0;
+        $non_public ||= $contacts[0] ? $contacts[0]->non_public : 0;
 
         my $problem = $self->schema->resultset('Problem')->new(
             {
--- a/perllib/Open311/PopulateServiceList.pm
+++ b/perllib/Open311/PopulateServiceList.pm
@@ -165,6 +165,7 @@ sub _handle_existing_contact {
     }
 
     $self->_set_contact_group($contact);
+    $self->_set_contact_non_public($contact);
 
     push @{ $self->found_contacts }, $self->_current_service->{service_code};
 }
@@ -201,6 +202,7 @@ sub _create_contact {
     }
 
     $self->_set_contact_group($contact);
+    $self->_set_contact_non_public($contact);
 
     if ( $contact ) {
         push @{ $self->found_contacts }, $self->_current_service->{service_code};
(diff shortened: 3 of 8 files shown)


perllib/Open311/GetServiceRequests.pm
```diff
@@ -157,6 +157,7 @@ sub create_problems {
         my $state = $open311->map_state($request->{status});
 
         my $non_public = $request->{non_public} ? 1 : 0;
+        $non_public ||= $contacts[0] ? $contacts[0]->non_public : 0;
```
dracos: This is fine, just noting that it means if a Open311 endpoint explicitly *does* say it is public, this will still override it. Presumably that case would never happen, so not worth checking for it being defined at all?

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -416,7 +416,10 @@ sub lookup_by_ref : Private {
             external_id => $ref
         ];
 
-    my $problems = $c->cobrand->problems->search( $criteria );
+    my $problems = $c->cobrand->problems->search({ -and => {
```
dracos: Doesn't matter, but I don't think the `-and` is needed here; `search({ non_public => 0, -or => $criteria })` should be the same.

perllib/Open311/PopulateServiceList.pm
```diff
@@ -283,6 +285,21 @@ sub _set_contact_group {
     }
 }
 
+sub _set_contact_non_public {
+    my ($self, $contact) = @_;
+
+    # We never want to make a private category unprivate.
+    return if $contact->non_public;
+
+    my $keywords = $self->_current_service->{keywords} || '';
```
dracos: For future proofing, do we want to do something like `%keywords = map { $_ => 1 } split /,/, $keywords` and use `$keywords{private}` just in case a keyword called "notatallprivate" came along? Unlikely I realise!

perllib/Open311/PopulateServiceList.pm
```diff
@@ -283,6 +285,21 @@ sub _set_contact_group {
...
+    $contact->update({
```
dracos: It seems a bit odd that one contact can cause multiple changelog entries (e.g. each of name, group, extra qns and private make their own row), but something for later perhaps.

t/open311/populate-service-list.t
```diff
@@ -225,6 +225,129 @@ subtest 'check conflicting contacts not changed' => sub {
...
+    is $contact_count, 1, 'correct number of contacts';
+
+    my $contact = FixMyStreet::DB->resultset('Contact')->search( { body_id => $body->id } )->first;
+    is $contact->email, '100', 'email correct';
+    is $contact->category, 'Cans left out 24x7', 'category correct';
+    is $contact->non_public, 1, 'contact marked as non_public';
+};
+
+subtest 'check existing category marked non_public' => sub {
+    FixMyStreet::DB->resultset('Contact')->search( { body_id => $body->id } )->delete();
```
dracos: Why not just mark the previous contact as public? :)

t/open311/populate-service-list.t
```diff
@@ -225,6 +225,129 @@ subtest 'check conflicting contacts not changed' => sub {
...
+    my $contact_count = FixMyStreet::DB->resultset('Contact')->search( { body_id => $body->id } )->count();
+    is $contact_count, 1, 'correct number of contacts';
+
+    $contact->discard_changes;
+    is $contact->email, '100', 'email correct';
+    is $contact->category, 'Cans left out 24x7', 'category correct';
+    is $contact->non_public, 1, 'contact changed to non_public';
+};
+
+subtest 'check existing non_public category does not get marked non_public' => sub {
```
dracos: Title should end 'marked public'.

t/open311/populate-service-list.t
```diff
@@ -225,6 +225,129 @@ subtest 'check conflicting contacts not changed' => sub {
...
+    FixMyStreet::DB->resultset('Contact')->search( { body_id => $body->id } )->delete();
```
dracos: Again, this could reuse the last contact, looks like.

perllib/Open311/GetServiceRequests.pm
davea: You're right, it's very unlikely to happen! I think it's safer to be able to say 'all fetched reports in a `non_public` category will be marked as `non_public`'.

PR #2483, opened 2019-05-20
Add Roles

dracos:
This PR implements the addition of "roles", being groups of permissions that can be assigned to a user to save having to select individual permissions. We first create the new database tables and add the new module we're going to use, HTML::FormHandler; then the three main bits are the CRUD interface for the roles, the changes to the user admin (editing and searching), and altering the permission checking to actually check these roles.

Please check the following:
- [x] Update Pro manual documentation with information on Roles
- [x] Is new functionality tested?

Todo:
- [x] Add fifth column to user import that adds roles (straightforward)
- [x] Add a way to assign many users to a role at once

Possible future things to do:
* Add a way to create a role based upon a user

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
@@ -1,0 +1,96 @@
+package FixMyStreet::App::Controller::Admin::Roles;
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+use FixMyStreet::App::Form::Role;
+
+sub auto :Private {
+    my ($self, $c) = @_;
+
+    my $user = $c->user;
+    if ($user->is_superuser) {
+        $c->stash(rs => $c->model('DB::Role')->search_rs({}, { join => 'body', order_by => ['body.name', 'me.name'] }));
+    } elsif ($user->from_body) {
+        $c->stash(rs => $user->from_body->roles->search_rs({}, { order_by => 'name' }));
+    }
+}
+
+sub index :Path :Args(0) {
+    my ($self, $c) = @_;
+
+    my $p = $c->cobrand->available_permissions;
+    my %labels;
+    foreach my $group (sort keys %$p) {
+        my $group_vals = $p->{$group};
+        foreach (sort keys %$group_vals) {
+            $labels{$_} = $group_vals->{$_};
+        }
+    }
+
+    $c->stash(
+        roles => [ $c->stash->{rs}->all ],
+        labels => \%labels,
+    );
+}
+
+sub create :Local :Args(0) {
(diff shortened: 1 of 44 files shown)


perllib/FixMyStreet/App/Form/Widget/Field/CheckboxGroup.pm
```diff
@@ -0,0 +1,60 @@
...
+
+use Moose::Role;
+with 'HTML::FormHandler::Widget::Field::CheckboxGroup';
+use namespace::autoclean;
+
+sub render_element {
+    my ( $self, $result ) = @_;
+    $result ||= $self->result;
+
+    my $output = '<ul class="permissions-checkboxes">';
```
davea: If we're rendering this ourselves, any scope to include the all/none shortcuts?
image

perllib/FixMyStreet/App/Controller/Admin/Roles.pm
```diff
@@ -0,0 +1,96 @@
...
+}
+
+sub edit :PathPart('') :Chained('item') :Args(0) {
+    my ($self, $c) = @_;
+    return $self->form($c, $c->stash->{obj});
+}
+
+sub delete :PathPart('delete') :Chained('item') :Args(0) {
+    my ($self, $c) = @_;
+    $c->stash->{obj}->delete;
```
davea: This line crashes if there are users in the role: `ERROR:  update or delete on table "roles" violates foreign key constraint "user_roles_role_id_fkey" on table "user_roles"`

templates/web/base/admin/roles/index.html
```diff
@@ -0,0 +1,36 @@
...
+        <td>[% role.name %]</td>
+        <td><ul class="no-margin no-bullets">
+          [% FOR perm IN role.permissions.sort %]
+            <li>[% labels.$perm %]
+          [% END %]
+        </ul></td>
+        <td>
+          <a href="[% c.uri_for(c.controller.action_for('edit'), [role.id]) %]">Edit</a>
+          &middot;
+          <a href="[% c.uri_for(c.controller.action_for('delete'), [role.id]) %]">Delete</a>
```
davea: This destructive action shouldn't be right alongside the edit link IMO. It'd be better as a big red button with a JS confirmation at the bottom of the role's edit page, like we do for e.g. templates:

image


We should indicate to the user how many users have this role before they delete it, maybe with a link to the pre-filled search page (`/admin/users?search=&role=1`) so they can verify who will be affected or re-assign roles etc.

dracos:
All the requested changes should have been made (might be easier to down/upgrade your schema as I changed it to add unique/cascade), I'll work on the new bits now.

dracos:
Okay new things there now, just adding some tests for those new bits but is ready for review.

PR #2488, opened 2019-05-23
[Hounslow] Go-live branch

davea:
This branch is made up of smaller individually-reviewed PRs:

 - #2482 
 - #2489


perllib/FixMyStreet/Cobrand/Hounslow.pm
```diff
@@ -0,0 +1,151 @@
...
+sub contact_email {
+    my $self = shift;
+    return join( '@', 'enquiries', $self->council_url . 'highways.org' );
+}
+
+sub send_questionnaires { 0 }
+
+sub enable_category_groups { 1 }
+
+sub suggest_duplicates { 1 }
```
dracos: This is a feature flag on master, so this line can be removed (and it added to config).

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -201,6 +201,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
     my $extra_titles_list = $c->cobrand->title_list($c->stash->{all_areas});
 
     my @list_of_names = map { $_->name } values %{$c->stash->{bodies}};
+    my %display_names = map { $_->name => $_->get_cobrand_handler ? $_->get_cobrand_handler->council_name : $_->name } values %{$c->stash->{bodies}};
```
dracos: This breaks for any cobrand that does not have `council_name`, ie all international ones.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -222,6 +223,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
 
     $c->stash->{json_response} = {
         bodies          => \@list_of_names,
+        display_names   => \%display_names,
```
dracos: This adds weight to every single ajax call, for currently one very specific usage. At the least could it be non-existent if empty, and only contain the row for which it needs to be there?

davea: d4b788f28075bb6566808c402b148ea58fac16f8 ensures only names that are different are included in the `display_names` output, and doesn't include that key in the JSON if it would be empty.

PR #2489, opened 2019-05-24
[Hounslow] Add cobrand

davea:
This is the Hounslow cobrand including the various template & functionality tweaks made along the way.

This PR consists of all Hounslow-related work except for the general enquiries form.


perllib/FixMyStreet/Cobrand/Hounslow.pm
```diff
@@ -80,4 +94,13 @@ sub should_skip_sending_update {
 }
 
 
+sub lookup_site_code_config { {
+    buffer => 50, # metres
+    url => "https://tilma.staging.mysociety.org/mapserver/hounslow",
```
struan: This will need to be updated before going live.

web/cobrands/hounslow/assets.js
```diff
@@ -0,0 +1,235 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+var defaults = {
+    http_options: {
+        url: "https://tilma.staging.mysociety.org/mapserver/hounslow",
```
struan: likewise, these will need to be updated.

perllib/FixMyStreet/Cobrand/Hounslow.pm
```diff
@@ -6,7 +6,7 @@ use warnings;
 
 sub council_area_id { 2483 }
 sub council_area { 'Hounslow' }
-sub council_name { 'Hounslow Borough Council' }
+sub council_name { 'Hounslow Highways' }
```
struan: I wonder if at some point we want a display name and a body name but probably not worth it unless we get more of these.

web/cobrands/hounslow/js.js
```diff
@@ -0,0 +1,18 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+var org_id = '5540';
+var body = "Hounslow Borough Council";
+fixmystreet.assets.add($.extend(true, {}, fixmystreet.roadworks.layer_future, {
+    http_options: { params: { organisation_id: org_id } },
```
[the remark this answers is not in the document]
davea: I think this was caused by a special character in the JSON response which was causing a parsing error:

```
Bad JSON: SyntaxError: Unexpected token  in JSON at position 166184
```

I've improved the JSON handling in 76071ab74923a064aac1aa8c1e3f7b34ffaa6f2d.

PR #2490, opened 2019-05-24
Duplicate/asset message interaction fix

dracos:
Fixes a couple of small issues where the new duplicate suggestion code was interacting with the asset messaging code and reshowing the form at some times when it shouldn't.
Also stop focussing on title when duplicates removed, as this bothered me while fixing this, and start to move UK cobrand config out of files.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UKCouncils.pm
+++ b/perllib/FixMyStreet/Cobrand/UKCouncils.pm
@@ -15,6 +15,11 @@ sub is_council {
     1;
 }
 
+sub suggest_duplicates {
+    my $self = shift;
+    return $self->feature('suggest_duplicates');
+}
+
 sub path_to_web_templates {
     my $self = shift;
     return [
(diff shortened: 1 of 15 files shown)


perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -15,6 +15,19 @@ sub is_council {
     1;
 }
 
+sub has_feature {
+    my ($self, $feature) = @_;
+    my $features = FixMyStreet->config('FEATURES');
```
davea: It'd be good to document this in `general.yml-example`.

dracos: Yeah, I wondered about that, if it's UK cobrand only, it might be confusing, but I guess there are other things like tat in there already. Perhaps rename it UKFEATURES or something...

davea: Or promote has_feature up to Default.pm?

dracos: Good idea, have done that!

PR #2491, opened 2019-05-29
[FixaMinGata] cobrand delivery may 2019

joesiltberg:
These are the latest changes for the FixaMinGata-cobrand.

The new Swedish translations should match what is in Transifex right now.


joesiltberg:
I will rebase to a more recent master and see if that fixes the codecov failure...

dracos:
Looks good, thanks!

PR #2492, opened 2019-05-29
[Buckinghamshire] Include Car parking category.

dracos:
[skip changelog]
For https://github.com/mysociety/fixmystreet-freshdesk/issues/58


dracos:
(Added changing district pins to grey)

PR #2496, opened 2019-06-03
[Rutland] limit report summaries to 254 characters

struan:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Rutland.pm
+++ b/perllib/FixMyStreet/Cobrand/Rutland.pm
@@ -12,6 +12,10 @@ sub council_url { return 'rutland'; }
 sub report_validation {
     my ($self, $report, $errors) = @_;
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Rutland.pm
```diff
@@ -12,6 +12,10 @@ sub council_url { return 'rutland'; }
 sub report_validation {
     my ($self, $report, $errors) = @_;
 
+    if ( length( $report->title ) > 254 ) {
+        $errors->{name} = sprintf( _('Summaries are limited to %d characters in length.'), 254 );
```
dracos: Should be `title`, not `name`. Also either "Summaries are limited to %s characters in length. Please shorten your summary" to match the existing string (e.g. in Northamptonshire), or don't mark this string for translation :)

PR #2497, opened 2019-06-03
Access map filters on mobile

### LINKED ISSUE #2366: Way of accessing filters at small widths
On mobile/small widths, you get a full screen map showing only open reports by default, with no way of changing that. There should be a way to see the filters (just bringing them into view should be all that's needed, they're in the page currently already!).

zarino:
Reports list filters can now be accessed while on a narrow screen.

Fixes #2366.

# New "Filter" item in `#sub_map_links`

Screen Shot 2019-06-03 at 14 57 50

# Filters fixed to bottom of screen

Screen Shot 2019-06-03 at 14 57 57

# Small multi-select dropdown

Screen Shot 2019-06-03 at 14 58 03

# Long multi-select dropdown (scrollable)

Screen Shot 2019-06-03 at 14 58 11

The dropdowns close themselves when you tap outside them.

And obviously it all reverts back to the normal styling / layout on wide screens.


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -762,10 +768,14 @@ $.extend(fixmystreet.set_up, {
         if ($('#map_permalink').length === 0) {
             $('#sub_map_links').append('<a href="#" id="map_permalink">' + translation_strings.permalink + '</a>');
         }
+        if ($('#map_filter').length === 0) {
```
dracos: This adds it to e.g. /reports and /my pages where it is not needed.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -762,10 +768,14 @@ $.extend(fixmystreet.set_up, {
...
+            $('#sub_map_links').prepend('<a href="#side" id="map_filter" class="hidden">Filter</a>');
+        }
     }
 
     if ($('.mobile').length) {
         $('#map_permalink').addClass('hidden');
+        $('#map_filter').removeClass('hidden');
```
dracos: Similarly, this then shows it on mobile pages where it doesn't work properly.
I think it needs to do all its things in apply/remove_ui.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -204,6 +206,10 @@ fixmystreet.resize_to = {
     // been put into place by previous mobile UI.
     $('#report-a-problem-sidebar').show();
     $('.rap-notes-trigger').remove();
+
+    // Turn off the mobile map filters.
+    $('#mapForm').removeClass('mobile-filters-active');
+    $('#sub_map_links').css('bottom', '');
```
dracos: Again, I think this should be in remove_ui

dracos:
That's more straightforward, isn't it :) One bug, though - if you click the map, then click "Try again", the filter link no longer works. (The show/hide pins is okay.) And I fell for that 'tap elsewhere to close pop-up menu' again!

zarino:
Ok @dracos this is now looking in much better shape.

I think the only substantial thing left to do is sort out the "tapping outside multiselect menu to close" thing – which I should probably implement as an update to https://github.com/mysociety/jquery-multi-select/ and then include in here as an update to `web/vendor/jquery.multi-select.min.js`.

dracos:
Sounds good to me, and I don't think it's enough of an issue to prevent this going ahead now!

zarino:
@dracos Ok so this branch now includes the updated version of `$.multiSelect` [from this open pull request](https://github.com/mysociety/jquery-multi-select/pull/15), and a commit making use of it, so that taps around the modals on small screens close them.

.cypress/cypress/integration/around_filters.js
```diff
@@ -127,3 +127,84 @@ describe('Around page filtering and push state', function() {
...
+            cy.get('[name=pc]').type(Cypress.env('postcode'));
+            cy.get('[name=pc]').parents('form').submit();
+        });
+
+        // Work around viewport bug https://github.com/cypress-io/cypress/issues/1534
+        // Manually reset the viewport before each it() and then wait
+        // a little while for the viewport resize to take effect.
+        beforeEach(function(done) {
+            cy.viewport(375, 667);
+            setTimeout(done, 500);
```
zarino: I should probably use `cy.wait(500)` here instead. Then there’ll be no need for the `done` callback.

PR #2504, opened 2019-06-06
Move ResponsePriority admin to HTML::FormHandler

dracos:
Also some small Role fixes.

The code looks to be about the same amount really, but it looks a bit clearer.

[skip changelog]


perllib/FixMyStreet/App/Controller/Admin/ResponsePriorities.pm
```diff
@@ -4,98 +4,94 @@ use namespace::autoclean;
...
+        ],
+        body_id => $c->stash->{body}->id,
+    };
+
+    my $form = FixMyStreet::App::Form::ResponsePriority->new(%$opts);
+    $c->stash(template => 'admin/responsepriorities/edit.html', form => $form);
+    $form->process(item => $priority, params => $c->req->params);
+    return unless $form->validated;
+
+    $c->response->redirect($c->uri_for($self->action_for('list')));
```
struan: There is no action for list.

PR #2509, opened 2019-06-10
[Open311] Make sure Open311 parameters are encoded

dracos:
This was previously done automatically, but isn’t when you do file uploads.

Looks like Open311 file uploading isn't tested at all, so don't have a regression test, will add some testing later today.
[skip changelog]


dracos:
Note to self for tomorrow - going by the coverage diff, this is calling out to an external service and won't work offline anymore, whoops.

PR #2513, opened 2019-06-12
Updates the help text for how to use the 'code' for extra fields.

### LINKED ISSUE #2512: Extra fields not available under problem.extra in email templates.
**Describe the bug**
When adding an extra field to a category, there's a help text that says the extra field should be available as 'problem.extra.address' or similar in the templates. I can't get this to work in the email templates.

Maybe I'm missing something, but I'll create a pull request with a solution that seems to work the way I would have expected.

**To Reproduce**
1. Create a new category with an extra field
2. Set the code for the new field to e.g. 'building'
3. Edit the submit.txt template, use ```[% problem.extra.building %]``` in e.g. the subject line
4. Submit a new report and enter something in the extra field
5. Note that the field doesn't show up in the email

**Expected behavior**
All extra fields available through ```problem.extra.<code>``` in the templates.

joesiltberg:


dracos:
Thanks!

PR #2521, opened 2019-06-18
Receive updates over Open311

dracos:
This adds a counterpoint to GetServiceRequestUpdates call, allowing Open311 endpoints to POST updates to FixMyStreet, rather than having them fetched from the endpoint.

It tries to reuse GetServiceRequestUpdates code, which was first refactored to split out a giant loop into functions. In the end, ignoring boilerplate and fetching parameters etc, the actual new code boils down to: find problem, check comment doesn't already exist, process update, format output.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1396

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Open311/Updates.pm
+++ b/perllib/FixMyStreet/App/Controller/Open311/Updates.pm
@@ -1,0 +1,88 @@
+package FixMyStreet::App::Controller::Open311::Updates;
+
+use utf8;
+use Moose;
+use namespace::autoclean;
+use Open311;
+use Open311::GetServiceRequestUpdates;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+=head1 NAME
+
+FixMyStreet::App::Controller::Open311::Updates - Catalyst Controller
+
+=head1 DESCRIPTION
+
+=head1 METHODS
+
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Open311/Updates.pm
```diff
@@ -0,0 +1,87 @@
...
+    my $user = $body->comment_user;
+
+    my $key = $c->get_param('api_key') || '';
+    my $token = $c->cobrand->feature('open311_token') || '';
+    $c->detach('bad_request', [ 'api_key' ]) unless $key && $key eq $token;
+
+    my $request = {
+        media_url => $c->get_param('media_url'),
+    };
+    foreach (qw(service_request_id update_id updated_datetime status description)) {
```
davea: Would be good to reject invalid `status` values. We should support `external_status_code` too, in case there are response templates which are triggered by that.

dracos: We don't reject invalid statuses we pull, so is at least consistent :) I thought it'd be better to accept invalid values in order to get the update, as I think we handle unknown states quite well.
Have added external_status_code.

PR #2523, opened 2019-06-20
[Westminster] Add OpenID Connect single sign-on support

davea:
Extends social login functionality to authenticate against an (MS Active Directory B2C-specific) OpenID Connect identity provider.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -44,13 +44,12 @@ sub general : Path : Args(0) {
 
     # decide which action to take
     $c->detach('code_sign_in') if $clicked_sign_in_by_code || ($data_email && !$data_password);
-    if (!$data_username && !$data_password && !$data_email) {
-        $c->detach('social/facebook_sign_in') if $c->get_param('facebook_sign_in');
-        $c->detach('social/twitter_sign_in') if $c->get_param('twitter_sign_in');
+    if (!$data_username && !$data_password && !$data_email && $c->get_param('social_sign_in')) {
+        $c->forward('social/handle_sign_in');
     }
 
-       $c->forward( 'sign_in', [ $data_username ] )
-    && $c->detach( 'redirect_on_signin', [ $c->get_param('r') ] );
+    $c->forward( 'sign_in', [ $data_username ] )
+        && $c->detach( 'redirect_on_signin', [ $c->get_param('r') ] );
 
 }
 
@@ -180,13 +179,13 @@ sub email_sign_in : Private {
         name => $c->get_param('name'),
         password => $user->password,
     };
-    $token_data->{name} = $c->session->{oauth}{name}
-        if $c->get_param('oauth_need_email') && $c->session->{oauth}{name} && !$token_data->{name};
 
-    $token_data->{facebook_id} = $c->session->{oauth}{facebook_id}
-        if $c->get_param('oauth_need_email') && $c->session->{oauth}{facebook_id};
-    $token_data->{twitter_id} = $c->session->{oauth}{twitter_id}
-        if $c->get_param('oauth_need_email') && $c->session->{oauth}{twitter_id};
+    if ($c->get_param('oauth_need_email')) {
+        $token_data->{name} = $c->session->{oauth}{name}
+            if $c->session->{oauth}{name} && !$token_data->{name};
+        $c->forward('set_oauth_token_data', [ $token_data ]);
+    }
+
     if ($c->stash->{current_user}) {
         $token_data->{old_user_id} = $c->stash->{current_user}->id;
         $token_data->{r} = 'auth/change_email/success';
@@ -215,6 +214,14 @@ sub get_token : Private {
 
     my $data = $token_obj->data;
     return $data;
+}
+
+sub set_oauth_token_data : Private {
+    my ( $self, $c, $token_data ) = @_;
+
+    foreach (qw/facebook_id twitter_id oidc_id/) {
+        $token_data->{$_} = $c->session->{oauth}{$_} if $c->session->{oauth}{$_};
+    }
 }
 
 =head2 token
@@ -275,6 +282,7 @@ sub process_login : Private {
     $user->password( $data->{password}, 1 ) if $data->{password};
     $user->facebook_id( $data->{facebook_id} ) if $data->{facebook_id};
     $user->twitter_id( $data->{twitter_id} ) if $data->{twitter_id};
+    $user->add_oidc_id( $data->{oidc_id} ) if $data->{oidc_id};
     $user->update_or_insert;
     $c->authenticate( { $type => $data->{$type}, $ver => 1 }, 'no_password' );
 
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -180,6 +180,9 @@ sub email_sign_in : Private {
         name => $c->get_param('name'),
         password => $user->password,
     };
+    $token_data->{name} = $c->session->{oauth}{name}
+        if $c->get_param('oauth_need_email') && $c->session->{oauth}{name} && !$token_data->{name};
+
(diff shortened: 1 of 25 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -54,16 +62,10 @@ __PACKAGE__->add_columns(
...
-  "twitter_id",
-  { data_type => "bigint", is_nullable => 1 },
-  "facebook_id",
-  { data_type => "bigint", is_nullable => 1 },
-  "extra",
-  { data_type => "text", is_nullable => 1 },
   "area_ids",
   { data_type => "integer[]", is_nullable => 1 },
+  "oidc_ids",
+  { data_type => "text[]", is_nullable => 1 },
```
dracos: Not that it matters much, but if you generate this on a fresh db schema created from scratch, it wouldn't move the other rows around.

perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -654,4 +658,20 @@ sub in_role {
...
+    my @oidc_ids = uniq ( $oidc_id, @$oidc_ids );
+    $self->oidc_ids(\@oidc_ids);
+}
+
+sub remove_oidc_id {
+    my ($self, $oidc_id) = @_;
+
+    my $oidc_ids = $self->oidc_ids || [];
+    my @oidc_ids = grep { $_ ne $oidc_id } @$oidc_ids;
+    $self->oidc_ids(\@oidc_ids);
```
dracos: To match a new user (as first test), should this set to undef if empty?

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -187,6 +188,8 @@ sub email_sign_in : Private {
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{facebook_id};
     $token_data->{twitter_id} = $c->session->{oauth}{twitter_id}
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{twitter_id};
+    $token_data->{oidc_id} = $c->session->{oauth}{oidc_id}
+        if $c->get_param('oauth_need_email') && $c->session->{oauth}{oidc_id};
```
dracos: Feels like these three could be in a little loop now, given they're all the same.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1191,6 +1191,8 @@ sub tokenize_user : Private {
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{facebook_id};
     $c->stash->{token_data}{twitter_id} = $c->session->{oauth}{twitter_id}
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{twitter_id};
+    $c->stash->{token_data}{oidc_id} = $c->session->{oauth}{oidc_id}
+        if $c->get_param('oauth_need_email') && $c->session->{oauth}{oidc_id};
```
dracos: Ditto

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -47,6 +47,7 @@ sub general : Path : Args(0) {
     if (!$data_username && !$data_password && !$data_email) {
         $c->detach('social/facebook_sign_in') if $c->get_param('facebook_sign_in');
         $c->detach('social/twitter_sign_in') if $c->get_param('twitter_sign_in');
+        $c->detach('social/oidc_sign_in') if $c->get_param('oidc_sign_in');
```
dracos: Should we rename this `sign_in` with three different values - looks like that would help simplify this in a few places in this commit? Or is there a reason they're separate.

perllib/FixMyStreet/App/Controller/Auth/Social.pm
```diff
@@ -142,6 +143,84 @@ sub twitter_callback: Path('/auth/Twitter') : Args(0) {
...
+
+    my %oauth;
+    $oauth{return_url} = $c->get_param('r');
+    $oauth{detach_to} = $c->stash->{detach_to};
+    $oauth{detach_args} = $c->stash->{detach_args};
+    $c->session->{oauth} = \%oauth;
+    $c->res->redirect($url);
+}
+
+sub oidc_callback: Path('/auth/OIDC') : Args(0) {
```
dracos: I realise they're all subtly different, but is there any worth in factoring the (now) three out to shared subclasses or something with variants for their differing parts?

perllib/FixMyStreet/App/Controller/Report/Update.pm
```diff
@@ -408,6 +408,8 @@ sub tokenize_user : Private {
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{facebook_id};
     $c->stash->{token_data}{twitter_id} = $c->session->{oauth}{twitter_id}
         if $c->get_param('oauth_need_email') && $c->session->{oauth}{twitter_id};
+    $c->stash->{token_data}{oidc_id} = $c->session->{oauth}{oidc_id}
+        if $c->get_param('oauth_need_email') && $c->session->{oauth}{oidc_id};
```
dracos: Ditto the first thing. Or a shared fn or something.

perllib/FixMyStreet/App/Controller/Report/Update.pm
```diff
@@ -444,6 +446,8 @@ sub save_update : Private {
             $c->detach('/auth/social/facebook_sign_in');
         } elsif ( $c->get_param('twitter_sign_in') ) {
             $c->detach('/auth/social/twitter_sign_in');
+        } elsif ( $c->get_param('oidc_sign_in') ) {
+            $c->detach('/auth/social/oidc_sign_in');
```
dracos: Ditto ditto the second thing, unless I've missed something.

perllib/OIDC/Lite/Client/WebServer/Azure.pm
```diff
@@ -0,0 +1,47 @@
...
+
+=cut
+
+
+sub get_id_token {
+    my $self = shift;
+
+    my $old_parser = $self->{response_parser};
+
+    $self->{response_parser} = OIDC::Lite::Client::IDTokenResponseParser->new;
```
dracos: This looks fine, but is there any reason you don't set the parser once in a `new` function; it looks like it's only refresh_access_token is other fn that uses it and presumably that's not called.

davea: That's a better idea, thanks.

perllib/FixMyStreet/DB/Result/User.pm
davea: Good to know for next time 👍

perllib/FixMyStreet/App/Controller/Auth.pm
davea: Do you mean the function name or the parameter?

dracos: I think I meant parameter here.
But I think I mentioned function elsewhere, though probably more work/less worth it.

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -44,10 +44,8 @@ sub general : Path : Args(0) {
 
     # decide which action to take
     $c->detach('code_sign_in') if $clicked_sign_in_by_code || ($data_email && !$data_password);
-    if (!$data_username && !$data_password && !$data_email) {
-        $c->detach('social/facebook_sign_in') if $c->get_param('facebook_sign_in');
-        $c->detach('social/twitter_sign_in') if $c->get_param('twitter_sign_in');
-        $c->detach('social/oidc_sign_in') if $c->get_param('oidc_sign_in');
+    if (!$data_username && !$data_password && !$data_email && $c->get_param('social_sign_in')) {
+        $c->detach('social/handle_sign_in');
```
dracos: I think all the calls to `social/handle_sign_in` should be forward, not detach, that way the current behaviour of carrying on if bad data provided is maintained, I think.

PR #2528, opened 2019-06-26
Update UI text when selecting a private category

davea:
This makes it clearer that all details for a report made in a private category will not be shown publicly.

Fixes mysociety/fixmystreet-commercial#1405

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -274,8 +274,10 @@ sub by_category_ajax_data : Private {
         $category ? (list_of_names => $list_of_names) : (),
     };
 
+    my $non_public = $c->stash->{non_public_categories}->{$category};
     my $body = {
         bodies => $list_of_names,
+        $non_public ? (non_public => JSON->true) : (),
     };
 
     if ($generate) {
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -262,12 +262,6 @@ sub category_extras_ajax : Path('category_extras') : Args(0) {
 sub by_category_ajax_data : Private {
     my ($self, $c, $type, $category) = @_;
 
-    my $generate;
-    if (($c->stash->{category_extras}->{$category} && @{ $c->stash->{category_extras}->{$category} } >= 1) or
-            $c->stash->{unresponsive}->{$category} or $c->stash->{report_extra_fields}) {
-        $generate = 1;
-    }
-
     my $bodies = $c->forward('contacts_to_bodies', [ $category ]);
     my $list_of_names = [ map { $_->name } ($category ? @$bodies : values %{$c->stash->{bodies_to_list}}) ];
     my $vars = {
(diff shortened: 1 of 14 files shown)


templates/web/base/report/new/councils_text_all.html
```diff
@@ -2,12 +2,19 @@
 [% DEFAULT list_of_names = default_list %]
 
 <p>
-[%
+<span class="js-hide-if-private-category">[%
```
dracos: Here, you know the category, and whether it's public or private, don't you?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1167,8 +1174,8 @@ fixmystreet.update_report_a_problem_btn = function() {
 
 fixmystreet.update_public_councils_text = function(text, bodies) {
     bodies = bodies.join('</strong> ' + translation_strings.or + ' <strong>');
-    text = text.replace(/<strong>.*<\/strong>/, '<strong>' + bodies + '</strong>');
-    $('#js-councils_text').html(text);
+    bodies = '<strong>' + bodies + '</strong>';
+    $('.js-councils_text_names').html(bodies);
```
dracos: This breaks the existing functionality (bad that it doesn't break a test then!) in that the `text` passed in might change more than just the bodies - e.g. it might now be the unresponsive text where it then says it's published online but not sent anywhere. Only replacing the bodies loses this. But as per my other comments, I think the server can return the correct text and this function can remain the same.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -274,8 +274,10 @@ sub by_category_ajax_data : Private {
         $category ? (list_of_names => $list_of_names) : (),
     };
 
+    my $non_public = $c->stash->{non_public_categories}->{$category};
     my $body = {
         bodies => $list_of_names,
+        $non_public ? (non_public => JSON->true) : (),
```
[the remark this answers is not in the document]
davea: I wanted the Public details" header to change as well, so needed a way to show/hide the `form-section-heading` elements based on the category's `non_public` flag.

templates/web/base/report/new/councils_text_all.html
davea: Yes, but the contents of this template is only rendered and included in `/report/new/ajax` output on a per-category basis for unresponsive bodies/categories AFAICT. I've pushed 6faa885a0e530462c4d393f69b29982e27a116b7 which rejigs things a bit to include `councils_text` for `non_public` categories, does that seem like a better approach?

web/cobrands/fixmystreet/fixmystreet.js
davea: 6faa885a0e530462c4d393f69b29982e27a116b7 should hopefully fix this!

templates/web/base/report/new/councils_text_all.html
dracos: Yeah, that looks better, doesn't it. Perhaps it would be better separate and down next to/with the existing inclusion of that template for unresponsive (general tidying of that fn, why is $generate not next to where it is used? It's performing the same purpose for the category_extras.html template) - with a note saying this template needs to be included whenever it would be changed from the default output, and that is currently unresponsive or non_public.

davea: Good idea. 155a00074f731ff694e02ca1cee24dc755668853 rejigs it a bit.

PR #2536, opened 2019-06-28
[Westminster] Asset handling

dracos:
Will fix https://github.com/mysociety/fixmystreet-commercial/issues/1444
[skip changelog]

### DIFF
--- a/templates/web/westminster/footer_extra_js.html
+++ b/templates/web/westminster/footer_extra_js.html
@@ -1,0 +1,10 @@
+[%~
+IF bodyclass.match('mappage');
+    scripts.push(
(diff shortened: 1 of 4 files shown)


dracos:
I've fixed the small things in fixups, and made the asset_group thing its own commit (d810a3545) back when that was added, and also improved it to be a bit of VectorAsset so not needing a Westminster subclass any more. I've probably broken it in doing so somehow.

dracos:
@davea I've hopefully fixed the 'panning removes the chosen property' issue, it tries to remember what was selected if it's then going to redraw it (seemed easier than trying to stop it redrawing at the mo!). Have added new thing for parent/child separate fetching,  if you could take a look at that, thanks. Think that's it.

PR #2544, opened 2019-07-03
[Zurich] Set envelope for external body emails.

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1424
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/SendReport/Zurich.pm
+++ b/perllib/FixMyStreet/SendReport/Zurich.pm
@@ -78,4 +78,17 @@ sub send_from {
     return [ FixMyStreet->config('CONTACT_EMAIL'), FixMyStreet->config('CONTACT_NAME') ];
 }
 
+# For messages to external body, we also want the sender to be the
+# category/body, so that bounces go there.
+sub envelope_sender {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/SendReport/Zurich.pm
```diff
@@ -78,4 +78,17 @@ sub send_from {
     return [ FixMyStreet->config('CONTACT_EMAIL'), FixMyStreet->config('CONTACT_NAME') ];
 }
 
+# For messages to external body, we also want the sender to be the
+# category/body, so that bounces go there.
+sub envelope_sender {
+    my ($self, $row) = @_;
+
+    if ( $row->external_body ) {
+        my $from = $self->send_from($row);
```
struan: Will this not mean that if the report has a category then the sender will be the email address of the category rather than of the division manager? This might just be me misunderstanding the workflow though.

dracos: Yes, currently all categories have the same email as the body, so I assumed they'd want the same thing as being used by the From line. I'll confirm with them.

Issue #2543, opened 2019-07-03
Duplicate closure fails on skip map page


dracos:
If you skip the map, but get a 'duplicate' message, trying to close it gives a JS error.

PR #2549, opened 2019-07-05
Prevent JS errors on skipped map page

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/2543
Also factored together some shared styles and stylemaps (given one error to fix the above was caused by pin_prefix).


web/cobrands/fixmystreet/assets.js
```diff
@@ -505,15 +480,56 @@ function get_fault_stylemap() {
...
+        popupYOffset: -40,
+        graphicOpacity: 1.0
+    }),
+
     selectedFeature: function() {
         return selected_feature;
     },
 
     add: function(default_options, options) {
+        if (!fixmystreet.map_type) {
```
davea: This breaks assets on certain cobrands (I've tried BANES, Hounslow, Bristol, Lincolnshire, Bexley. Works OK on Bucks though). It looks as though `map_type` isn't set at this point so no asset layers get added to the map.

web/js/map-OpenLayers.js
```diff
@@ -791,6 +794,10 @@ $.extend(fixmystreet.utils, {
 
     $(function(){
 
+        if (!fixmystreet.map_type) {
```
davea: This test fails on Bexley and Lincolnshire and breaks the map for some reason.

web/cobrands/fixmystreet/assets.js
dracos: Okay, turns out map_type isn't always defined, I thought it was; I've switched to checking for the presence of the map div.

web/js/map-OpenLayers.js
dracos: Okay, turns out map_type isn't always defined, I thought it was; I've switched to checking for the presence of the map div.

web/cobrands/fixmystreet/assets.js
dracos: Plus rewriting the code to always make sure map.js loads before assets.js. But then still back to the blank issue as elsewhere.

PR #2559, opened 2019-07-12
[Bexley] Include subcategory in dashboard export.

dracos:
[skip changelog]
I thought about doing it in all cobrands with category groups, but B&NES fetch their CSV file programmatically, so I don't know if they'd handle that or not. (Also they have the option of multiple groups for one category, so might not make sense for them.) So this inserts it just after Category in the export, for Bexley only. I don't know if any others do manual exports; alternatively we could add it to all that have groups and don't have multiple-groups-per-category, meh.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -131,4 +131,31 @@ sub open311_post_send {
     $sender->send($row, $h);
 }
 
+sub dashboard_export_problems_add_columns {
+    my $self = shift;
+    my $c = $self->{c};
+
+    my %groups;
+    if ($c->stash->{body}) {
+        %groups = FixMyStreet::DB->resultset('Contact')->active->search({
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -131,4 +131,31 @@ sub open311_post_send {
...
 
+sub dashboard_export_problems_add_columns {
+    my $self = shift;
+    my $c = $self->{c};
+
+    my %groups;
+    if ($c->stash->{body}) {
+        %groups = FixMyStreet::DB->resultset('Contact')->active->search({
+            body_id => $c->stash->{body}->id,
+        })->group_lookup;
```
struan: This could return an array ref if it's been set through the admin.

dracos: ...or if multi groups ever switched on, Open311 would do it too, yeah, might as well try and cope, ta.

PR #2567, opened 2019-07-18
Island roads assets

struan:
[skip-changelog]


web/cobrands/isleofwight/assets.js
```diff
@@ -0,0 +1,347 @@
...
+    "Abandoned Vehicles": "Abandoned_Vehicles"
+};
+
+for (i = 0; i < point_category_list.length; i++) {
+    cat = point_category_list[i];
+    layer = layer_map[cat] || cat;
+
+    fixmystreet.assets.add($.extend(true, {}, point_asset_defaults, {
+        asset_group: cat,
+        asset_category: [
```
dracos: If asset_group is set, asset_category isn't used, unless it's a `road: true` layer? So I think these all can be removed from point layers.

web/cobrands/isleofwight/assets.js
```diff
@@ -0,0 +1,347 @@
...
+fixmystreet.assets.add($.extend(true, {}, defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "streets"
+        }
+    },
+    always_visible: true,
+    non_interactive: true,
+    asset_type: 'area',
+    road: true,
```
dracos: I don't think you need road: true here, unless there's some only_send interaction I'm not aware of.

web/cobrands/isleofwight/assets.js
struan: I think in this case this is ok because of the non standard setup of the cobrand. The idea here is that you have groups and categories with the same name but they're never shown at the same time and this means that when the groups are showing they display the correct assets.

web/js/map-wmts-isleofwight.js
```diff
@@ -0,0 +1,206 @@
...
+    if (zoom >= 7) {
+        return 'normal';
+    } else if (zoom >= 4) {
+        return 'small';
+    } else {
+        return 'mini';
+    }
+};
+
+fixmystreet.maps.selected_marker_size = function() {
```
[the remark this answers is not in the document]
struan: I didn't want to just randomly add things to Hounslow because I'm not sure if the behaviour there is as intended.

web/cobrands/isleofwight/assets.js
```diff
@@ -188,6 +190,8 @@ var line_category_list = [
 
 var layer_map = {
     "Dog Fouling": "Dog_Fouling",
+    "Drainage": "Drainage_Line",
```
dracos: It was lowercase line in the bit this came from.

web/cobrands/isleofwight/assets.js
```diff
@@ -123,7 +122,6 @@ var point_asset_defaults = $.extend(true, {}, defaults, {
 });
 
 var line_asset_defaults = $.extend(true, {}, defaults, {
-    road: true,
```
dracos: You need this one, though, so that the found/not found are called.

web/cobrands/isleofwight/assets.js
```diff
@@ -99,20 +99,24 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
     })
 }));
 
+function found_item(layer, asset) {
+  var id = asset.attributes.asset_id || '';
```
dracos: This was different in the two calls, central_asset_id/asset_id.

web/cobrands/isleofwight/assets.js
```diff
@@ -99,20 +99,24 @@ fixmystreet.assets.add($.extend(true, {}, defaults, {
...
+  if (id !== '') {
+      var attrib = asset.attributes;
+      var asset_name = attrib.feature_type_name + '; ' + attrib.site_name + '; ' + attrib.feature_location;
+      $('.category_meta_message').html('You have selected ' + asset_name);
+  } else {
+      $('.category_meta_message').html('You can pick an <b class="asset-spot">' + layer.fixmystreet.asset_item + '</b> from the map &raquo;');
```
dracos: This was different in the two calls a/an.

web/cobrands/isleofwight/assets.js
dracos: But there is no piece of code that gets called using asset_category if asset_group is set on the object is my point, unless I've missed something?

web/cobrands/isleofwight/assets.js
```diff
@@ -0,0 +1,347 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+fixmystreet.stopper_after = true;
```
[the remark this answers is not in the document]
dracos: Still isn't :)

web/cobrands/isleofwight/assets.js
struan: it should be `central_asset_id`, not everything has an `asset_id`

PR #2568, opened 2019-07-18
Add/improve anonymous reporting

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1465

### DIFF
--- a/t/app/controller/report_new_anon.t
+++ b/t/app/controller/report_new_anon.t
@@ -3,6 +3,11 @@ use parent 'FixMyStreet::Cobrand::FixMyStreet';
 sub allow_anonymous_reports { 1 }
 sub anonymous_account { { email => 'anon@example.org', name => 'Anonymous' } }
 
+package FixMyStreet::Cobrand::AnonAllowedByButton;
+use parent 'FixMyStreet::Cobrand::FixMyStreet';
+sub allow_anonymous_reports { 'button' }
+sub anonymous_account { { email => 'anonbutton@example.org', name => 'Anonymous Button' } }
+
 package main;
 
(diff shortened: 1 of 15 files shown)


templates/web/westminster/report/form/user.html
```diff
@@ -25,6 +55,9 @@ <h2 class="form-section-heading form-section-heading--private hidden-nojs">
     </button>
   [% END %]
     <button type="button" class="btn btn--block hidden-nojs js-new-report-user-show">Confirm by email</button>
+    <small id="or">or</small>
```
struan: Should this not be outside the `social_auth_enabled` IF?

t/app/controller/report_new_anon.t
```diff
@@ -0,0 +1,92 @@
...
+
+    my $report = FixMyStreet::DB->resultset("Problem")->first;
+    ok $report, "Found the report";
+
+    is $report->state, 'confirmed', "report confirmed";
+    $mech->get_ok( '/report/' . $report->id );
+
+    is $report->bodies_str, $body->id;
+    is $report->name, 'Anonymous';
+    is $report->anonymous, 0; # Doesn't change behaviour here, but uses anon account's name always
```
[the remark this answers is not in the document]
dracos: I think this was so a user of this functionality (prior to the button way of doing it) could set a default name to be used that might not be anonymous e.g. could attribute all reports made like this to the site itself or similar.

templates/web/westminster/report/form/user.html
dracos: It will always be enabled there due to MyWestminster, but yes I guess so to match.

PR #2582, opened 2019-08-06
[Westminster] USRN lookup

davea:
Adds server-side USRN lookup for Westminster reports made via the app (or any which don't already have a USRN for whatever reason).

For https://github.com/mysociety/fixmystreet-commercial/issues/1471
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Westminster.pm
+++ b/perllib/FixMyStreet/Cobrand/Westminster.pm
@@ -4,6 +4,8 @@ use base 'FixMyStreet::Cobrand::Whitelabel';
 use strict;
 use warnings;
 
+use URI;
+
 sub council_area_id { return 2504; }
 sub council_area { return 'Westminster'; }
 sub council_name { return 'Westminster City Council'; }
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -70,6 +75,63 @@ sub open311_config {
...
+    {
+        buffer => 1000, # metres
+        proxy_url => "https://tilma.staging.mysociety.org/resource-proxy/proxy.php",
+        url => "https://westminster.assets/40/query",
+        property => "USRN",
+        accept_feature => sub { 1 }
+    }
+}
+
+sub _fetch_features {
```
dracos: A lot of this is the same as what it's overwriting - a factored out `lookup_site_code_url` function would be better.

davea: Good idea, have added that in 334f4370d95649041120bc0e5e5858ba6edfbf1c.

PR #2586, opened 2019-08-06
Redirect correctly after editing user to remove them from cobrand

davea:
Certain edits to a user (e.g. unsetting `from_body`, editing email address) might result in that user not being visible in the current cobrand any more. This PR fixes a 404 error by redirecting back to the user list after such an edit, instead of to the user profile page.

[skip changelog]

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1486.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Users.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Users.pm
@@ -174,7 +174,7 @@ sub add : Local : Args(0) {
     $c->forward( '/admin/log_edit', [ $user->id, 'user', 'edit' ] );
 
     $c->flash->{status_message} = _("Updated!");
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -174,7 +174,14 @@ sub add : Local : Args(0) {
     $c->forward( '/admin/log_edit', [ $user->id, 'user', 'edit' ] );
 
     $c->flash->{status_message} = _("Updated!");
-    $c->res->redirect( $c->uri_for_action( 'admin/users/edit', $user->id ) );
+    # User may not be visible on this cobrand, e.g. if their from_body
```
dracos: Worth putting these two changes in one function, given they're identical?

PR #2587, opened 2019-08-06
[Westminster] Some small changes

davea:
- Replace ‘gully’ with ‘drain’ in asset UI ([Basecamp](https://3.basecamp.com/4020879/buckets/12002933/todos/1972169811))
 - ~Upload all photos over Open311~ ([Basecamp](https://3.basecamp.com/4020879/buckets/12002933/messages/1969231294)) (supporting code is included in this PR, but not enabled for the Westminster cobrand)
 - Hide 'change password' & 'change email' links on `/my` ([Basecamp](https://3.basecamp.com/4020879/buckets/12002933/messages/1945441618#__recording_1978717754))
 - Remove TfL categories ([Basecamp](https://3.basecamp.com/4020879/buckets/12002933/todos/1978021687))
 - Show 3 months on `/around` by default ([Basecamp](https://3.basecamp.com/4020879/buckets/12002933/messages/1939205134#__recording_1974850946))

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth/Social.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth/Social.pm
@@ -207,6 +207,19 @@ sub oidc_sign_in : Private {
         $oauth{logout_redirect_uri} = $redirect_uri;
     }
 
+    # The OIDC endpoint may provide a specific URI for changing the user's password.
+    if ( my $password_change_uri = $c->cobrand->feature('oidc_login')->{password_change_uri} ) {
+        $oauth{change_password_uri} = $oidc->uri_to_redirect(
+            uri          => $password_change_uri,
+            redirect_uri => $c->uri_for('/auth/OIDC'),
+            scope        => 'openid',
+            state        => 'password_change',
+            extra        => {
+                response_mode => 'form_post',
+            },
+        );
+    }
+
     $c->session->{oauth} = \%oauth;
     $c->res->redirect($url);
 }
@@ -231,6 +244,10 @@ sub oidc_callback: Path('/auth/OIDC') : Args(0) {
             );
             $c->res->redirect($url);
             $c->detach;
+        } elsif ($c->user_exists && $c->get_param('state') && $c->get_param('state') eq 'password_change') {
+            $c->flash->{flash_message} = _('Password change cancelled.');
+            $c->res->redirect('/my');
+            $c->detach;
         } else {
             $c->detach('oauth_failure');
         }
@@ -248,6 +265,11 @@ sub oidc_callback: Path('/auth/OIDC') : Args(0) {
             $c->stash->{detach_args} = $c->session->{oauth}{detach_args};
         }
         $c->detach('oidc_sign_in', []);
+    }
+
+    # User may be coming back here after changing their password on the OIDC endpoint
+    if ($c->user_exists && $c->get_param('state') && $c->get_param('state') eq 'password_change') {
+        $c->detach('/auth/profile/change_password_success', []);
     }
 
     # The only other valid state param is 'login' at this point.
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/App/Controller/Auth/Social.pm
```diff
@@ -231,6 +244,10 @@ sub oidc_callback: Path('/auth/OIDC') : Args(0) {
             );
             $c->res->redirect($url);
             $c->detach;
+        } elsif ($c->get_param('state') &&$c->get_param('state') eq 'password_change' && $c->user_exists) {
```
dracos: Missing space (also, as I'm here, any reason this is in a different order to the same if a few lines down?).

perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -116,4 +116,12 @@ sub _fetch_features_url {
     return $cfg->{proxy_url} . "?" . $uri->as_string;
 }
 
+sub categories_restriction {
```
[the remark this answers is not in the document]
davea: Won't that stop fixmystreet.com users in Westminster being able to send reports to TfL?

perllib/FixMyStreet/SendReport/Open311.pm
```diff
@@ -30,6 +30,7 @@ sub send {
             extended_description    => 1,
             multi_photos            => 0,
             upload_files            => 0,
+            always_upload_files     => 0,
```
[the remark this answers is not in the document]
davea: Yep, good idea.

perllib/FixMyStreet/App/Controller/Auth/Social.pm
davea: No good reason, no :)

perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -116,4 +118,12 @@ sub _fetch_features_url {
...
+    my ($self, $rs) = @_;
+    # Westminster don't want TfL categories on their cobrand
+    # XXX This still shows "These will be sent to TfL or Westminster City Council"
+    # on /report/new before a category is selected...
+    return $rs->search( [ { 'body.name' => 'Westminster City Council' } ], { join => { body => 'body_areas' } });
```
dracos: Not sure what the `[` and `]` is doing here, or the join with body_areas.

perllib/FixMyStreet/Cobrand/Westminster.pm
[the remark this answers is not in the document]
dracos: Oh I assumed not wanting them there meant they wanted to get them all instead. Fair enough.
In which case, my comment becomes - I don't see how the XXX happens because categories_restriction happens before bodies_to_list (which is what is shown) is generated, so it feels like there's a bug here somewhere.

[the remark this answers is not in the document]
dracos: Ah, it looks like it's because the JS report_form_ajax call uses bodies rather than bodies_to_list - I actually changed that as part of my bugfixing work on Wednesday but it broke a test and I couldn't see if there was a reason it differed that I was missing.

perllib/FixMyStreet/Cobrand/Westminster.pm
davea: That'll be a result of me cargo-culting from another cobrand!

perllib/FixMyStreet/SendReport/Open311.pm
dracos: It's still always_upload_files here.

struan:
Why the test or why the behaviour?

dracos:
Either :) I thought I might have missed a reason why you'd want to show the body even if it has no contacts, I guess I can see it'd be confusing if the names disappeared entirely or something, perhaps it's that, if that could happen.

struan:
I assumed it was so people would at least know who to try and get in contact with if we had no contacts.

PR #2589, opened 2019-08-07
Have duplicate suggestion and assets coexist better

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1480


dracos:
"I think the "zoom in to pick a light" text should be hidden when duplicates are shown though:" - yes that is fixed by one part of #2588 :)

davea:
> yes that is fixed by one part of #2588 :)

Excellent!

PR #2599, opened 2019-08-14
Add 'disable_form' flag to category extras

davea:
This replaces the cobrand-JS-specific way of adding messages that disable the report form with a new 'disable form' tickbox in the admin.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1515

### DIFF
--- a/templates/web/base/admin/extra-metadata-form.html
+++ b/templates/web/base/admin/extra-metadata-form.html
@@ -43,6 +43,12 @@
             <input name="metadata[[% loop.index %]].protected" data-field-name="protected" type=checkbox [% meta.protected == 'true' ? 'checked' : '' %]>
         </label>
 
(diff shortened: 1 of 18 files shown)


templates/web/base/admin/extra-metadata-form.html
```diff
@@ -75,6 +81,10 @@
                             [% loc('Name') %]
                             <input class="js-metadata-option-name" name="metadata[[% outer_loop.index %]].values[[% loop.index %]].name" type="text" value="[% option.name | html %]">
                         </label>
+                        <label>
+                            [% loc('Disable form') %]
+                            <input class="js-metadata-option-disable" name="metadata[[% outer_loop.index %]].values[[% loop.index %]].disable" type="checkbox"[% ' checked' IF option.disable %]>
+                        </label>
```
struan: This needs to be added to the `renumber_metadata_options` method in `admin.js` otherwise if you check it when you add the option for the first time it doesn't get set.

PR #2602, opened 2019-08-15
[Westminster] UPRN lookup

davea:
Makes it easier to handle more than one call to `lookup_site_code` for a single report, and looks up the UPRN for reports that have that field in their report extras but not a value. Also restores the frontend UPRN lookup for bins, after the field was renamed.

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1471

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Westminster.pm
+++ b/perllib/FixMyStreet/Cobrand/Westminster.pm
@@ -90,8 +90,18 @@ sub open311_config {
     # display the road layer. Instead we'll look up the closest asset from the
     # asset service at the point we're sending the report over Open311.
     if (!$row->get_extra_field_value('USRN')) {
-        if (my $ref = $self->lookup_site_code($row)) {
+        if (my $ref = $self->lookup_site_code($row, 'USRN')) {
             push @$extra, { name => 'USRN', value => $ref };
+        }
+    }
+
+    # Some categories require a UPRN to be set, so if the field is present
+    # but empty then look it up.
+    my $fields = $row->get_extra_fields;
+    my ($uprn_field) = grep { $_->{name} eq 'UPRN' } @$fields;
+    if ( $uprn_field && !$uprn_field->{value} ) {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -88,24 +88,42 @@ sub open311_config {
...
+        if (my $ref = $self->lookup_site_code($row, undef, 'USRN')) {
             push @$extra, { name => 'USRN', value => $ref };
         }
     }
 
+    # Some categories require a UPRN to be set, so if the field is present
+    # but empty then look it up.
+    my $fields = $row->get_extra_fields;
+    my @uprn_field = grep { $_->{name} eq 'UPRN' } @$fields;
+    if ( scalar @uprn_field && !$row->get_extra_field_value('UPRN')) {
```
dracos: If you replaced `my @uprn_field` with `my ($uprn_field)` then this line could be `if ($uprn_field && !$uprn_field->{value})`? In fact, could you just go off `get_extra_field_value` being undefined as opposed to blank to tell the difference?

perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -88,24 +88,42 @@ sub open311_config {
...
+        if (my $ref = $self->lookup_site_code($row, undef, 'UPRN')) {
```
dracos: Function parameters near not the greatest. It looks like buffer is never called if you want to drop that entirely first?

perllib/FixMyStreet/Cobrand/Westminster.pm
davea: I went with the first option, it's a bit clearer.

perllib/FixMyStreet/Cobrand/Westminster.pm
davea: Good call, have gotten rid of the buffer param.

PR #2607, opened 2019-08-19
Make sure anonymous reports are marked anon irrespective of permissions

davea:
There was a small bug where the 'default_to_body' permission would
override the 'report anonymously' button when staff users were adding
reports to the site. The result of this was that the name of the
anonymous user record would be shown. No harm done, as that user's name
would likely be set to "Anonymous user" in config, but it resulted in
reports pages showing the slightly odd wording along the lines of:

> Reported in the Bin bags category by Anonymous user at 14:57 today

This commit ensures the contributing_as_body flag is mutually exclusive
with the 'report_anonymously' request parameter.

Please check the following:

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -949,6 +949,8 @@ sub process_report : Private {
     my $anon_button = $c->cobrand->allow_anonymous_reports eq 'button' && $c->get_param('report_anonymously');
     if ($anon_button) {
         $c->stash->{contributing_as_anonymous_user} = 1;
+        $c->stash->{contributing_as_body} = undef;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -952,7 +952,7 @@ sub process_report : Private {
     }
 
     # set some simple bool values (note they get inverted)
-    if ($c->stash->{contributing_as_body}) {
+    if ($c->stash->{contributing_as_body} && !$anon_button) {
```
dracos: This works fine here, but note later on, body takes precedence over anonymous_user for setting the extra metadata of contributed_as. I think instead of this change, at line 951 it should make sure body/another_user are unset (given only one of these three should be set at once, perhaps it should be a single flag, not 3?).

PR #2608, opened 2019-08-20
Contact email admin improvements

zarino:
It’s common for people to send contact form messages about their most recent report (usually they’ve included personal details and want them removed) but without actually linking to the report, or using the well hidden "Unsuitable?" link on the report page.

So the person handling the support email for that site has to go manually searching for the user’s profile, or their most recent report, in the admin interface.

After doing this for what seemed like the thousandth time this morning, I figured it’s time to get FMS to do the hard work for me, generating links to the user’s profile page, list of reports, and latest report, if a matching user (and report) is found in the database:

Screenshot_2019-08-20 Screenshot(1)

It also improves the `/_dev/email/contact` email preview endpoint, to show some dummy content, just like the problem/update/alert endpoints.

@dracos @davea @struan what do you think? Useful?

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Contact.pm
+++ b/perllib/FixMyStreet/App/Controller/Contact.pm
@@ -202,11 +202,7 @@ sub prepare_params_for_email : Private {
         $c->stash->{user_admin_url} = $admin_url . '/users/' . $user->id;
         $c->stash->{user_reports_admin_url} = $admin_url . '/reports?search=' . $user->email;
 
-        my $user_latest_problem = $user->problems->search({
-            state => [ FixMyStreet::DB::Result::Problem->visible_states() ]
-        }, {
-            order_by => { -desc => 'id' }
-        })->single;
+        my $user_latest_problem = $user->latest_visible_problem();
         if ( $user_latest_problem) {
             $c->stash->{user_latest_report_admin_url} = $admin_url . '/report_edit/' . $user_latest_problem->id;
         }
--- a/perllib/FixMyStreet/App/Controller/Contact.pm
+++ b/perllib/FixMyStreet/App/Controller/Contact.pm
@@ -197,6 +197,21 @@ sub prepare_params_for_email : Private {
     my $base_url = $c->cobrand->base_url();
     my $admin_url = $c->cobrand->admin_base_url;
 
+    my $user = $c->cobrand->users->find( { email => $c->stash->{em} } );
+    if ( $user ) {
+        $c->stash->{user_admin_url} = $admin_url . '/users/' . $user->id;
+        $c->stash->{user_reports_admin_url} = $admin_url . '/reports?search=' . $user->email;
+
+        my $user_latest_problem = $user->problems->search({
+            state => [ FixMyStreet::DB::Result::Problem->visible_states() ]
+        }, {
+            order_by => { -desc => 'id' }
+        })->single;
+        if ( $user_latest_problem) {
+            $c->stash->{user_latest_report_admin_url} = $admin_url . '/report_edit/' . $user_latest_problem->id;
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


davea:
Oooh I love it! I've found myself frustrated at the same thing. One nice-to-have would be an indication whether that user was logged in, and if they were, whether the email address they've put in the contact field matches their user's email address? I've often wondered how we'd know if someone was impersonating another user (not that there's any indication it's ever happened).

One thing to consider is that replying to the user includes the email from the contact form, which would reveal the admin links to that user. They wouldn't be able to access them, of course, but worth noting.

zarino:
> whether the email address they've put in the contact field matches their user's email address?

Feels like maybe a separate PR to me? But initial my thoughts: I don’t think I care about whether the user was logged in, but I guess I _might_ want to see a little notification if they _were_ logged in and the two email addresses didn’t match.

> replying to the user includes the email from the contact form, which would reveal the admin links to that user.

Yup, I considered adding that to the PR message above, but then wondered whether this actually matters? We’d be exposing is the user’s internal FMS ID, and the ID of their most recent report, to that same user (based on the stated email address matching the email address of the user in the database). Is that a problem?

perllib/FixMyStreet/App/Controller/Contact.pm
```diff
@@ -197,6 +197,22 @@ sub prepare_params_for_email : Private {
     my $base_url = $c->cobrand->base_url();
     my $admin_url = $c->cobrand->admin_base_url;
 
+    my $user = $c->cobrand->users->find( { email => $c->stash->{em} } );
+    if ( $user ) {
+        $c->stash->{user_admin_url} = $admin_url . '/users/' . $user->id;
+        $c->stash->{user_reports_admin_url} = $admin_url . '/reports?search=' . $user->email;
+
+        my $user_latest_problem = $c->cobrand->problems->search({
+            user_id => $user->id
```
dracos: Probably only want to include visible states here? Guess might be of interest if latest report is hidden already, maybe...

perllib/FixMyStreet/App/Controller/Contact.pm
```diff
@@ -197,6 +197,22 @@ sub prepare_params_for_email : Private {
...
+        }, {
+            rows => 1,
+            order_by => { -desc => 'id' }
+        })->first;
```
dracos: If you use single, instead of first, you can drop the rows=>1 line too.

perllib/FixMyStreet/App/Controller/Develop.pm
```diff
@@ -115,6 +115,32 @@ sub email_previewer : Path('/_dev/email') : Args(1) {
...
+        $vars->{ip} = $c->req->address;
+        $vars->{user_agent} = $c->req->user_agent;
+        $vars->{complaint} = sprintf(
+            "Complaint about report %d",
+            $vars->{problem}->id,
+        );
+        $vars->{problem_url} = $c->cobrand->base_url() . '/report/' . $vars->{problem}->id;
+        $vars->{admin_url} = $c->cobrand->admin_base_url . '/report_edit/' . $vars->{problem}->id;
+        $vars->{user_admin_url} = $c->cobrand->admin_base_url . '/users/' . $c->user->id;
+        $vars->{user_reports_admin_url} = $c->cobrand->admin_base_url . '/reports?search=' . $c->user->email;
```
dracos: Wondered if it was worth factoring out the bits of Contact.pm so that this could call it rather than duplicate it all, but probably not?

perllib/FixMyStreet/App/Controller/Contact.pm
zarino: Feels like we shouldn’t link to anything here that the user themselves can’t access. So that would preclude hidden reports.

PR #2609, opened 2019-08-21
[Island Roads] Report triage area

struan:
This adds the following:

 * A new Triage report sender that changes the report status to 'triage'
 * A Triage admin controller that display a list of reports awaiting triage at `/admin/triage`. This also handles displaying and updating reports awaiting triage.
 * An alternative side panel for reports in the triage state showing the current category and the available categories.

The list of current categories is limited to those that have a send method of 'Triage'. All other categories are displayed as categories to update to.

This also assumes that the Triage categories have the same name as the groups that the non Triage categories are in, and uses this to set the post triage category list to the first category in the group.

When a report is updated with a non triage category it marked for resending and the state is set to `confirmed`.

Fixes mysociety/fixmystreet-commercial#1474

Please check the following:


perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
+package FixMyStreet::App::Controller::Admin::Triage;
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'FixMyStreet::App::Controller::Reports'; }
```
dracos: I think I'd prefer the lines in the function to call `/reports/...` than do this, because I'm not sure if this makes other pages available here that you don't want to?

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+
+sub auto : Private {
```
dracos: The auto in Admin.pm should also be being run, so only the triage bit of this function should be needed?

templates/web/base/admin/triage/_list-filters.html
```diff
@@ -0,0 +1,53 @@
+[% select_category = BLOCK %]
```
dracos: I think I factored out the state dropdown in the current Zurich work, in which case this and the base reports/_list-filters.html template are basically the same and could be factored together.

templates/web/base/admin/triage/_inspect.html
```diff
@@ -0,0 +1,117 @@
+[% permissions = c.user.permissions(problem) %]
```
dracos: Again, worth factoring out some bits of _inspect.html into smaller templates so they can be shared? e.g. the top info bit

templates/web/base/admin/triage/_inspect.html
```diff
@@ -1,3 +1,20 @@
+[% BLOCK category_list %]
```
dracos: Probably better to add passing a select_name to admin/report-category.html, defaulting to 'category', instead of duplicating the template here :)

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -479,6 +479,10 @@ sub report_edit : Path('report_edit') : Args(1) {
         $c->forward('users/flag');
         $c->stash->{problem}->discard_changes;
     }
+    elsif ( $c->get_param('triage') ) {
```
dracos: Does this ever get called?

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -42,8 +43,21 @@ sub munge_category_list {
...
     my %bodies = map { $_->body->name => 1 } @$contacts;
-    return unless $bodies{'Hounslow Borough Council'};
-    @$options = grep { ($_->{category} || $_->category) !~ /^Traffic lights$/i } @$options;
+    if ( $bodies{'Hounslow Borough Council'} ) {
+        @$options = grep { ($_->{category} || $_->category) !~ /^Traffic lights$/i } @$options;
+    }
+
+    if ( $bodies{'Isle of Wight Council'} ) {
+        my $user = $self->{c}->user;
+        my $b = $self->{c}->model('DB::Body')->for_areas( COUNCIL_ID_ISLEOFWIGHT )->first;
```
dracos: If you change the %bodies hash above, you can already have the body here, it's already been looked up.

templates/web/isleofwight/admin/triage/_inspect.html
```diff
@@ -1,16 +1,27 @@
...
+                    [% IF group_select == 0 AND problem.category == group.name %]
+                        [% selected = 1; group_select = 1 %]
+                    [% END %]
+                    <option value="[% cat_op.category | html %]"[% ' selected' IF selected OR problem.category == cat_op.category %]>[% cat_op.category_display | html %]</option>
+                    [% selected = 0 %]
+                [%~ END ~%]
+                [% IF group.name %]</optgroup>[% END %]
+            [%~ END =%]
+        [% ELSE %]
+            <optgroup label="[% loc('Available categories') %]">
```
dracos: No need for an optgroup with only one group

templates/web/isleofwight/admin/triage/_inspect.html
```diff
@@ -1,16 +1,27 @@
...
-    <optgroup label="[% loc('Available categories') %]">
-      [% FOREACH cat IN category_options %]
-        <option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %]>[% cat.category_display | html %]</option>
+        [%~ IF category_groups.size ~%]
+            [%~ FOREACH group IN category_groups ~%]
+                [% IF group.name %]<optgroup label="[% group.name %]">[% END %]
+                [% group_select = 0 %]
+                [% selected = 0 %]
+                [%~ FOREACH cat_op IN group.categories ~%]
+                    [% IF group_select == 0 AND problem.category == group.name %]
```
dracos: Instead of the loop vars, could the IF on the option line not be `IF problem.category == cat_op.category OR problem.category == group.name`?

templates/web/isleofwight/admin/triage/_inspect.html
```diff
@@ -0,0 +1,135 @@
...
+            [% INCLUDE 'report/_inspect_reject_button.html' %]
+          [% CATCH file %]
+            <p>
+              <a href="[% c.uri_for( '/contact', { id => problem.id, reject => 1 } ) %]" class="btn btn--block">[% loc('Reject report') %]</a>
+            </p>
+          [% END %]
+        [% END %]
+      </div>
+
+    [% IF permissions.report_edit_category OR permissions.report_inspect OR permissions.triage %]
```
dracos: Not sure why the base template doesn't get the permissions.triage bits.

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+    $c->stash->{filter_problem_states} = { 'for triage' => 1 };
+    return 1;
+}
+
+sub setup_categories : Private {
+    my ( $self, $c ) = @_;
+
+    if ( $c->stash->{problem}->state eq 'for triage' ) {
+        $c->stash->{holding_options} = [ grep { $_->send_method && $_->send_method eq 'Triage' } @{$c->stash->{category_options}} ];
+        $c->stash->{holding_categories} = { map { $_->category => 1 } grep { $_->send_method && $_->send_method eq 'Triage' } @{$c->stash->{category_options}} };
```
dracos: You can use the previous row's result in this row :)

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+        $c->stash->{end_options} = [ grep { !$_->send_method || $_->send_method ne 'Triage' } @{$c->stash->{category_options}} ];
+        $c->stash->{end_categories} = { map { $_->category => 1 } grep { !$_->send_method || $_->send_method ne 'Triage' } @{$c->stash->{category_options}} };
```
dracos: Ditto

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+
+    my $current_category = $c->get_param('holding_category');
+    my $new_category = $c->get_param('category');
+
+    my $problem = $c->stash->{problem};
+    my $changed = $c->forward('/admin/report_edit_category', [ $problem, 1 ] );
+
+    if ( $changed && !$problem->whensent ) {
+        $c->stash->{problem}->update( { state => 'confirmed' } );
+        # TODO add a comment/audit
```
dracos: At the least an admin log entry, yeah :) The front end is showing the public update bit that isn't currently used at all.

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+sub update : Private {
+    my ($self, $c) = @_;
+
+    my $current_category = $c->get_param('holding_category');
+    my $new_category = $c->get_param('category');
+
+    my $problem = $c->stash->{problem};
+    my $changed = $c->forward('/admin/report_edit_category', [ $problem, 1 ] );
+
+    if ( $changed && !$problem->whensent ) {
```
dracos: whensent will always be null here, I think - because the send method changed from a Triage to a non-triage one, which will prompt a resend. If I've understood it right. If whensent *is* set though (which means they didn't change the category, I think is the only way? Or the send_method wasn't set right which is what happened to me in testing which confused me as to why nothing was happening.), basically I think if you've submitted this form out of triage, it should always do something...

perllib/FixMyStreet/SendReport/Triage.pm
```diff
@@ -0,0 +1,20 @@
+package FixMyStreet::SendReport::Triage;
+
+use Moo;
```
dracos: Some help text for this module would be nice.

templates/web/isleofwight/admin/triage/_inspect.html
dracos: In fact, I don't think there's a need at this time for separate base/iow templates for this - if you really wanted the 'group' behaviour to be IoW only, then at least these templates should be sharing everything else.

perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -0,0 +1,108 @@
...
+        $c->stash->{end_groups} = \@category_groups;
+    }
+
+    return 1;
+}
+
+sub update : Private {
+    my ($self, $c) = @_;
+
+    my $current_category = $c->get_param('holding_category');
```
dracos: This is a dropdown in the interface, is read here, but nothing can be done with it - I can't change it to a different holding category. Should the front end be displaying this rather than letting it be edited?

templates/web/base/admin/triage/index.html
```diff
@@ -0,0 +1,42 @@
...
+[% ELSIF c.cobrand.moniker == 'isleofwight' %]
+  [% SET body_name = 'Island Roads' %]
+[% END %]
+
+[%
+    PROCESS "report/photo-js.html";
+    PROCESS "maps/${map.type}.html";
+    SET bodyclass = 'mappage';
+    INCLUDE 'header.html',
+        title = tprintf(loc('%s - Summary reports'), name)
```
dracos: name isn't defined here.

templates/web/isleofwight/admin/triage/_inspect.html
```diff
@@ -0,0 +1,146 @@
...
+
+        <p>
+          <label for="category">CONFIRM Subject</label>
+          [% PROCESS category_list category_groups=end_groups field_name="category" categories_hash=end_categories category_options=end_options %]
+        </p>
+
+        [% IF permissions.report_inspect OR permissions.triage %]
+
+          <p>
+            <label for="state">[% loc('State') %]</label>
```
dracos: This shows a state dropdown, but nothing happens with it.

templates/web/isleofwight/admin/triage/_inspect.html
struan: Won't that select multiple items?

dracos: I think the logic is the same as the existing code? Currently it says if the problem category matches cat_op.category or selected, which can be set if it matches the group name, so it'll also select both if there are one of each with the same name.

struan: If you have a Group called "Roads" and the problem is set to the category "Roads" then it will set `selected` for every category in that group, whereas I only want it to set `selected` for the first.

The idea is that the Confirm categories have groups that are named the same as the Triage categories

dracos: Aha, I see what you mean, sorry for the misunderstanding.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -42,15 +42,14 @@ sub munge_category_list {
     my ($self, $options, $contacts, $extras) = @_;
 
     # No TfL Traffic Lights category in Hounslow
-    my %bodies = map { $_->body->name => 1 } @$contacts;
+    my %bodies = map { $_->body->name => $_ } @$contacts;
```
dracos: $_ is a contact here, not a body.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -42,15 +42,14 @@ sub munge_category_list {
...
     if ( $bodies{'Hounslow Borough Council'} ) {
         @$options = grep { ($_->{category} || $_->category) !~ /^Traffic lights$/i } @$options;
     }
 
     if ( $bodies{'Isle of Wight Council'} ) {
         my $user = $self->{c}->user;
-        my $b = $self->{c}->model('DB::Body')->for_areas( COUNCIL_ID_ISLEOFWIGHT )->first;
-        if ( $user && ( $user->is_superuser || $user->belongs_to_body( $b->id ) ) ) {
+        if ( $user && ( $user->is_superuser || $user->belongs_to_body( $bodies{'Isle of Wight Council'}) ) ) {
```
dracos: belongs_to_body needs to be passed an ID, not an object.

templates/web/base/report/inspect/information.html
```diff
@@ -0,0 +1,54 @@
+      <div class="inspect-section">
```
dracos: Should the base _inspect use this template as well?

templates/web/base/admin/triage/_list-filters.html
struan: Yup, not quite sure how best to use that in this branch at the moment though given the Zurich things aren't merged yet.

dracos: Yeah, thought that'd be in by now. Anyway, can be left to after, no biggie.

templates/web/base/admin/triage/_inspect.html
```diff
@@ -0,0 +1,69 @@
...
```
dracos: I see this went away, and then came back, presumably to show the groups? Showing the categories in their groups sounds like a thing the base report-category.html template should probably be doing anyway too?

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
```diff
@@ -86,6 +86,41 @@ sub munge_category_list {
     @$options = grep { my $c = ($_->{category} || $_->category); $c =~ 'Pick a category' || $seen->{ $c } } @$options;
 }
 
+sub munge_load_and_group_problems {
+    my ($self, $where, $filter) = @_;
+
+    return unless $where->{category};
+
+    my $b = $self->{c}->model('DB::Body')->for_areas( $self->council_area_id )->first;
```
dracos: I think $c->stash->{body} will have to exist here and be equal to this :)

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
```diff
@@ -86,6 +86,41 @@ sub munge_category_list {
...
+            category => $where->{category}
+        }
+    );
+
+    my @cat_names;
+    while ( my $cat = $cats->next ) {
+        if ( $cat->send_method eq 'Triage' ) {
+            my $cat_name = $cat->category;
+            my $len = length $cat_name;
+            my @cats = $self->{c}->model('DB::Contact')->search(
```
dracos: As there won't be many, but this loop might be a few times, wonder if it's easier/nicer code to fetch all the body's contacts first off, and create a lookup of group => categories, then you don't have to worry about RABX string lengths at least!

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
```diff
@@ -70,6 +70,92 @@ sub open311_munge_update_params {
...
+
+sub munge_category_where {
+    my ($self, $where) = @_;
+
+    my $user = $self->{c}->user;
+    my $b = $self->{c}->model('DB::Body')->for_areas( $self->council_area_id )->first;
+    if ( $user && ( $user->is_superuser || $user->belongs_to_body( $b->id ) ) ) {
+        $where = {
+            body_id => $where->{body_id},
+            -or => { send_method => [ { '!=' => 'Triage' }, undef ] },
```
dracos: I think this can just be ` send_method => [ { '!=' => 'Triage' }, undef ]` - not sure what the `-or` wrapper is for.

perllib/FixMyStreet/App/Controller/Around.pm
```diff
@@ -236,10 +236,15 @@ sub check_and_stash_category : Private {
     $csv->combine(@list_of_names);
     $c->{stash}->{list_of_names_as_string} = $csv->string;
 
+    my $where  = { body_id => [ keys %bodies ], };
+
+    my $cobrand_where = $c->cobrand->call_hook('munge_category_where', $where );
```
dracos: I can't see any corresponding change for /reports

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
```diff
@@ -70,6 +70,92 @@ sub open311_munge_update_params {
...
+            category => $categories
+        }
+    );
+
+    my @cat_names;
+    while ( my $cat = $cats->next ) {
+        if ( $cat->send_method eq 'Triage' ) {
+            my $cat_name = $cat->category;
+            my $len = length $cat_name;
+            my @cats = $self->{c}->model('DB::Contact')->not_deleted->search(
```
dracos: As there won't be many, but this loop might be a few times, wonder if it's easier/nicer code to fetch all the body's contacts first off, and create a lookup of group => categories, then you don't have to worry about RABX string lengths at least!

templates/web/base/report/updates.html
```diff
@@ -6,6 +6,9 @@
   [%- END %]
   [%- NEXT %]
 [%- END %]
+[%- IF update.get_extra_metadata('triage_report') == 1 AND ( NOT c.user OR ( NOT c.user.from_body AND NOT c.user.is_superuser ) ) %]
```
dracos: Feels like a 'staff' state for updates might be better here, but something for later perhaps.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -81,6 +82,8 @@ sub send() {
 
             # this is for the new_updates alerts
             next if $row->{non_public} and $row->{user_id} != $row->{alert_user_id};
+            # do not sent triage comments
+            next if $row->{item_extra} && $row->{item_extra} =~ /triage_report,I1:1/;
```
dracos: Would it not make more sense to mark the comment as already alerted on at the time it was created? Then no need to check it here.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -165,10 +165,13 @@ sub ward : Path : Args(2) {
     $c->stash->{stats} = $c->cobrand->get_report_stats();
 
     my @categories = $c->stash->{body}->contacts->not_deleted->search( undef, {
-        columns => [ 'id', 'category', 'extra' ],
+        columns => [ 'id', 'category', 'extra', 'body_id', 'send_method' ],
         distinct => 1,
         order_by => [ 'category' ],
     } )->all;
+
+    $c->cobrand->call_hook('munge_categories', \@categories);
```
dracos: Lots of similar named functions, I'm having trouble keeping track. Perhaps this should be munge_reports_category_filter_list? (and munge_category_where could be munge_around_category_filter_where or somesuch)

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -37,13 +38,118 @@ sub restriction {
...
+    my $cats = $self->{c}->model('DB::Contact')->not_deleted->search(
+        {
+            body_id => $b->id,
+            category => $categories
+        }
+    );
+
+    my @cat_names;
+    while ( my $cat = $cats->next ) {
+        if ( $cat->send_method eq 'Triage' ) {
```
dracos: send_method might be undefined here, perhaps?

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
```diff
@@ -80,9 +80,135 @@ sub open311_munge_update_params {
...
+    my $seen = { map { $_->category => 1 } @$contacts };
+    @$options = grep { my $c = ($_->{category} || $_->category); $c =~ 'Pick a category' || $seen->{ $c } } @$options;
+}
+
+sub munge_category_where {
+    my ($self, $where) = @_;
+
+    my $user = $self->{c}->user;
+    my $b = $self->{c}->model('DB::Body')->for_areas( $self->council_area_id )->first;
+    if ( $user && ( $user->is_superuser || $user->belongs_to_body( $b->id ) ) ) {
```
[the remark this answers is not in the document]
struan: As above, it needs to limit it to only IoW categories.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -37,13 +38,118 @@ sub restriction {
...
+    my ($self, $where) = @_;
+
+    my $user = $self->{c}->user;
+    my @iow = grep { $_->name eq 'Isle of Wight Council' } @{ $self->{c}->stash->{bodies} };
+    return unless @iow;
+
+    # display all the categories on Isle of Wight at the moment as there's no way to
+    # do the expand bit later as we fetch it using ajax which uses a bounding box so
+    # can't determine the body
+    $where = {
```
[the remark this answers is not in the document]
struan: It needs to limit the list to only Isle of Wight categories, otherwise you return all the categories.

templates/web/isleofwight/other-reported.txt
```diff
@@ -0,0 +1,43 @@
+Subject: Your report has been logged: [% report.title %]
```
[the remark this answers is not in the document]
struan: It's to be the template that all the other cobrands use for confirm_report_sent so I was trying to be consistent. It does, as you say, not make a lot of sense though.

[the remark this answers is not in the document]
struan: It's in that commit because the behaviour changes depending on whether the report has been triaged or not.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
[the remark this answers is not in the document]
dracos: Yes, but $where already has that, when passed in? OIC, it overrides if anything if returned. Well, stop doing that and let the function mutate/munge the passed in $where :)

perllib/FixMyStreet/Cobrand/IsleOfWight.pm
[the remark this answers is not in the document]
dracos: The other part of this function does what I mean, adding $where->{send_method} to the existing $where and returning it. I just didn't understand why the first part of this function didn't do the same thing.

templates/web/isleofwight/other-reported.txt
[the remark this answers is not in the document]
dracos: If you want this template to be used for confirm report sent, as I think you imply, then the confirm_report_sent templates need to exist, which they currently don't :) (And a test that these emails arrive would show you whether it was or not.)

PR #2614, opened 2019-08-27
Improve asset selection/duplicate suggestion interactions on mobile

davea:
On mobile the asset selection process is a bit fiddly, and the duplicate suggestion UI makes this harder than it needs to be.

This PR adds three small changes to improve things a bit:

1.  Stops the duplicates code from removing the asset markers from the map if you happen to miss the asset when tapping:

Old | New
--- | ---
mobile_dups | mobile_dups2

2. The duplicates are now dismissed when returning to the map to select the asset marker:
(Although I'm not super happy with this solution, as the user may actually want to go to the map to see the locations of the duplicate markers... So open to removing/rethinking this change.)

mobile_dups_dismissed

3. The "You can pick a _thing_ from the map" message is a button on mobile that takes you to the map:

mobile_asset_button

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -386,6 +386,11 @@ function check_zoom_message_visibility() {
     if (this.relevant()) {
         if ($p.length === 0) {
             $p = $("<p>").prop("id", id).prop('class', 'category_meta_message');
+            if ($('html').hasClass('mobile')) {
+                $p.click(function() {
+                    $("#mob_ok").trigger('click');
+                }).addClass("btn");
+            }
             $p.prependTo('#js-post-category-messages');
         }
 
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -482,7 +482,7 @@ $.extend(fixmystreet.set_up, {
             }
         });
 
-        $(fixmystreet).trigger('report_new:category_change');
+        $(fixmystreet).trigger('report_new:category_change', { check_duplicates_dismissal: true });
     });
   },
 
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1460,6 +1460,9 @@ fixmystreet.display = {
                 $('html').addClass('only-map');
                 $('#mob_sub_map_links').removeClass('map_complete');
                 $('#mob_ok').text(translation_strings.ok);
+                if (fixmystreet.duplicates) {
+                    fixmystreet.duplicates.hide();
+                }
             });
         });
     }
--- a/web/js/duplicates.js
+++ b/web/js/duplicates.js
@@ -209,4 +209,10 @@
         remove_duplicate_list();
     });
 
+    fixmystreet.duplicates = {
+        hide: function() {
+            remove_duplicate_pins();
+            remove_duplicate_list();
+        }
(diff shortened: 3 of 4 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -385,6 +385,11 @@ function check_zoom_message_visibility() {
     if (this.relevant()) {
         if ($p.length === 0) {
             $p = $("<p>").prop("id", id).prop('class', 'category_meta_message');
+            if ($('html').hasClass('mobile')) {
+                $p.click(function() {
+                    $("#mob_ok").trigger('click');
```
dracos: This is nice; additionally I feel we should probably lose the "Try again" overlay entirely, perhaps? So that people can scroll back up the page rather than have to click either this or the "Map" link that they might not realise is a link.

web/js/duplicates.js
```diff
@@ -190,6 +200,10 @@
         if ($('.js-responsibility-message:visible').length) {
             return false;
         }
+        // On mobile only show once per category
+        if ($('html').hasClass('mobile') && dismissed) {
```
dracos: This means that on mobile, if you pick a point, get some dupes, go back up to the map and move the point to somewhere entirely different, you do not look up any duplicates around that new location (or rather they have been looked up, but this then stops them being shown). I understand the reasoning when there are assets, so perhaps this is acceptable, I'm not sure if there's a way that can accommodate it all (could only do this if assets are being shown I guess...).

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1393,6 +1393,9 @@ fixmystreet.display = {
                 $('html').addClass('only-map');
                 $('#mob_sub_map_links').removeClass('map_complete');
                 $('#mob_ok').text(translation_strings.ok);
+                if (fixmystreet.duplicates) {
+                    fixmystreet.duplicates.hide();
```
dracos: I think this does make sense at present, especially given there's no way to then dismiss the duplicates without going back down anyway. Perhaps on mobile it'd be better if duplicates weren't shown on the map at all ever, and instead when you clicked "Read more" under any of them, the expansion included a little inline map there? That would keep them together then at least.

web/js/duplicates.js
```diff
@@ -35,6 +40,11 @@
             url_params.pin_size = 'normal';
         }
 
+        if (params && params.category ) {
+            dismissed = params.category === dismissed_category;
+            dismissed_category = params.category;
+        }
```
dracos: Why not return here if it's going to be dismissed, before making a request to the server?

web/cobrands/fixmystreet/assets.js
davea: Possibly, but let's consider that in a separate issue: #2667.

web/js/duplicates.js
davea: Yes, it's tricky isn't it. I think the prioritisation of assets over duplicates is the right choice in this case. Given the assets have fixed locations, perhaps the duplicates code should only suggest duplicates that are literally on the same asset? A future improvement in any case.

web/cobrands/fixmystreet/fixmystreet.js
davea: 💯 to the inline map: #2668.

davea:
Thanks for the review. I've addressed the code improvements and broken the other suggestions out into individual tickets.

PR #2619, opened 2019-08-30
 Improve user flow when JavaScript is not available

dracos:
This improves the reporting journey to only ask for category, and then category extra questions if appropriate, first, so that if the choice would lead to the form being disabled, this can be shown immediately.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -143,6 +143,7 @@ sub report_new_ajax : Path('mobile') : Args(0) {
 
     $c->forward('setup_categories_and_bodies');
     $c->forward('setup_report_extra_fields');
+    $c->forward('check_for_category');
     $c->forward('process_report');
     $c->forward('process_user');
     $c->forward('/photo/process_photo');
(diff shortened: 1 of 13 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1518,10 +1518,55 @@ sub generate_map : Private {
...
+    my $category = $c->get_param('category') || $c->stash->{report}->category || '';
     $category = '' if $category eq _('Loading...') || $category eq _('-- Pick a category --');
-    $c->stash->{category} = $category || $c->stash->{report}->category || '';
+    $c->stash->{category} = $category;
 
+    if ($c->get_param('submit_category_part_only')) {
+        # If we've clicked the first-part category button (no-JS only probably),
+        # we only want to reshow the form
+        $c->stash->{force_form_not_submitted} = 1;
+    }
```
struan: This does mean if they change their mind once the full form is displayed they can still submit in a blocked category. However, that's possibly a future improvement.

dracos: I've committed 87bba912063d2f2e6b884be58cfc8ffb709a5028 as a fixup which should deal with this - if there's a disable form message to be shown, it knows it shouldn't allow full form submission.

PR #2620, opened 2019-09-03
Asset code ordering issues

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1543

I realise with this fix the stopper code is run more times than seems necessary (e.g. on a map pin click, checkFeature is called immediately, then when category data comes back, that will lead to a category change event firing, so it'll get called again, plus the stopper code runs in its own right on category change anyway) but given we don't know when/where data will be available, that seems safest, if a bit hmm. The stopper code doesn't do very much, at least.
Also due to the above, picking the roads category on Westminster will show the disabled bit before the TfL data comes back at which point the form comes back. I couldn't think of an easy fix except somehow telling the stopper code that it should wait for layer data, which also doesn't seem great; the only alternative I could think of is to load the TfL data all the time (as happens on Bromley) but that then seems a waste if you never need it. Bit of a compromise all round, looks like.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1025,7 +1025,16 @@ sub process_report : Private {
         my $body_string = do {
             if (my $single_body_only = $c->get_param('single_body_only')) {
                 my $body = $c->model('DB::Body')->search({ name => $single_body_only })->first;
-                $body ? $body->id : '-1';
+                if ($body) {
+                    # Drop the contacts down to those in this body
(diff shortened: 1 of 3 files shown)


dracos:
Note: The last two commits (Northamptonshire/HE related) of this PR are currently on live.

dracos:
I just realised that the category extra emptying (which I put in to remove a 'please ring' message) needed something doing server side, otherwise you'd get stuck in a loop of server saying you hadn't filled in a required field, and the client hiding the required field because it was emptying the box. Please check and say if you agree :)

PR #2627, opened 2019-09-06
Island roads training bugfixes

struan:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/User.pm
+++ b/perllib/FixMyStreet/DB/Result/User.pm
@@ -317,7 +317,11 @@ sub body {
 
 sub moderating_user_name {
     my $self = shift;
-    return $self->body || _('an administrator');
+    my $body = $self->body;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/DB/Result/User.pm
```diff
@@ -317,7 +317,11 @@ sub body {
 
 sub moderating_user_name {
     my $self = shift;
-    return $self->body || _('an administrator');
+    my $body = $self->body;
+    if ( $body && $body eq 'Isle of Wight Council' ) {
```
dracos: Okay (will this apply to Hounslow Highways as well?), but would be nicer to be in the cobrand presumably. As this is only used on-session, you can access the cobrand at `$self->result_source->schema->cobrand` I think, looking elsewhere in the same file (though would you need .com fix too? hmm), if that's helpful.

PR #2628, opened 2019-09-09
[Peterborough] Initial cobrand

davea:
The cobrand has been through review with the client and feedback has been addressed.

NB 11199e5ed11f3b01bf90f36e41f98c4bdb9c83c9 is very similar to 086d9debfe1a365f49a97e3e707f2d1c26861f45 in #2611, so some rejigging required to one or another of the PRs after one is merged.

[skip changelog]


templates/web/peterborough/about/faq-en-gb.html
```diff
@@ -0,0 +1,136 @@
...
+    you&rsquo;re from the authority then the emailed version of the problem
+    report might also contain the closest road or postcode to the pin on the
+    map.</dd>
+</dl>
+
+<h2><a name="organisation"></a>Organisation Questions</h2>
+
+<dl>
+    <dt>Who built [% site_name %]?</dt>
+    <dd>This will depend upon the installation of the FixMyStreet Platform,
```
dracos: This part is still the default text (haven't checked it all thoroughly).

templates/web/peterborough/index-steps.html
```diff
@@ -0,0 +1,19 @@
+<h2>[% loc('How to report a problem') %]</h2>
+
+<ol class="big-numbers">
+    <li>[% question %]</li>
+    <li>[% loc('Locate the problem on a map') %]</li>
+    <li>[% loc('Enter more details of the problem') %]</li>
+    [% IF c.cobrand.is_council %]
```
dracos: This will always be true (and it will never be zurich, further down)

web/cobrands/peterborough/layout.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#site-logo {
+    padding: 1em 0;
+}
+
+body.frontpage #front-main, 
+body.twothirdswidthpage #front-main, 
+body.fullwidthpage #front-main, 
+body.authpage #front-main {
```
dracos: You don't have front-main on any page but body.frontpage?

web/cobrands/peterborough/layout.scss
```diff
@@ -0,0 +1,101 @@
...
+    text-align: left;
+    background-color: #fff;
+    border-top: 1px solid #e9e9e9;
+    color: $primary_b;
+    padding: 2.5em 0;
+}
+
+#front-main { 
+    #postcodeForm div {
+        margin: 0;
```
dracos: Might not be needed due to flexbox changes in CE cobrand - I'll pull them out into separate PR to make this easier (will do same with alert index template refactoring as well I guess?)

dracos:
Pulled out flexbox/template changes into #2631 so when that's merged can rejig this on top of that.

dracos:
#2631 is merged, so rebasing on top of current master should simplify that alert template change for you.

web/cobrands/peterborough/js.js
```diff
@@ -0,0 +1,42 @@
...
+    http_options: {
+        url: "https://tilma.staging.mysociety.org/mapserver/peterborough",
+        params: {
+            SERVICE: "WFS",
+            VERSION: "1.1.0",
+            REQUEST: "GetFeature",
+            SRSNAME: "urn:ogc:def:crs:EPSG::3857"
+        }
+    },
+    format_class: OpenLayers.Format.GML.v3.MultiCurveFix, // Not sure needed any more
```
dracos: We have MapServer 7 on the server now, so I assume this *isn't* needed? Would be good to find an example geometry to confirm this for sure.

templates/web/fixmystreet.com/footer_extra_js.html
```diff
@@ -14,6 +14,7 @@
     scripts.push( version('/cobrands/northamptonshire/assets.js') );
     scripts.push( version('/cobrands/hounslow/assets.js') );
     scripts.push( version('/cobrands/westminster/assets.js') );
+    scripts.push( version('/cobrands/peterborough/assets.js') );
```
dracos: This should be js.js, going by elsewhere.

web/cobrands/peterborough/js.js
davea: Yep, looks as though our mapserver is doing the right thing now. See, e.g. geometry of the second `featureMember` here: https://tilma.mysociety.org/mapserver/highways?SERVICE=WFS&VERSION=1.1.0&REQUEST=GetFeature&SRSNAME=urn%3Aogc%3Adef%3Acrs%3AEPSG%3A%3A3857&TYPENAME=Highways&bbox=-35748.263942009,6687326.5660957,-34175.212234679,6688378.3808725

PR #2633, opened 2019-09-12
Rename categories

### LINKED ISSUE #1398: Allow editing of category name in admin
If you want to change the name of a category, you can't; you have to delete it and create a new one for the new name.
This is because it used as a foreign key; it should certainly be allowed if there are no reports though.

dracos:
This PR:
* [BANES] Adds a display name field to their admin so they can have an override name for a category, but keeping the same internal name (fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/83)
* Has the core code necessary for that to work
* Lets categories be renamed in other cobrands (fixes #1398)
* Makes contact edit note optional on staging sites.
* Does a bit of refactor admin body creation/editing
* Stops some warning lines in Westminster/Zurich tests

Please check the following:

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -220,7 +220,7 @@ sub update_contact : Private {
 
     my $category = $self->trim( $c->get_param('category') );
     $errors{category} = _("Please choose a category") unless $category;
-    $errors{note} = _('Please enter a message') unless $c->get_param('note');
+    $errors{note} = _('Please enter a message') unless $c->get_param('note') || FixMyStreet->config('STAGING_SITE');
 
     my $contact = $c->model('DB::Contact')->find_or_new(
         {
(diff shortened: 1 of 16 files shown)


templates/web/bathnes/admin/bodies/_category_field.html
```diff
@@ -0,0 +1,17 @@
+[% IF contact.in_storage %]
+    <h1>[% contact.category | html %]</h1>
+    <input type="hidden" name="category" value="[% contact.category | html %]" >
+
+    <p>
```
davea: An `admin-hint` here would be nice, just to explain what happens when a display name is set/changed.

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -220,145 +201,181 @@ sub check_for_super_user : Private {
...
-            @group = grep { $_ } @group;
-            if (scalar @group == 0) {
-                $contact->unset_extra_metadata( 'group' );
-            } else {
-                $contact->set_extra_metadata( group => \@group );
-            }
-        } else {
+    );
+    if ($current_contact && $contact->id && $contact->id != $current_contact->id) {
+        $errors{category} = _('You cannot rename a category to an existing category');
```
davea: In the BANES cobrand it's possible to use the same display name for multiple contacts, and to set a contact's display name to the same as another contact's actual name. Should this be disallowed?

dracos: Seems unlikely that it'll arise, but the former should be easy enough, and I guess the latter theoretically could be a use case (probably not, but shrug)

PR #2641, opened 2019-09-19
[Open311] Spot new <groups> parameter rather than use CSV.

dracos:
Companion to https://github.com/mysociety/open311-adapter/pull/59 to read in `groups` when given, rather than parse a single entry as CSV (which doesn't then need to be used anywhere, I think).
This means we can drop all the enable multiple groups stuff to deal with comma edge cases, and let anyone have multiple groups if they should want to without worrying about it.


t/open311/populate-service-list.t
```diff
@@ -153,7 +151,7 @@ subtest "set multiple groups with quoted csv" => sub {
         <metadata>false</metadata>
         <type>realtime</type>
         <keywords>lorem, ipsum, dolor</keywords>
-        <group>&quot;sanitation &amp; cleaning&quot;,street</group>
+        <groups><group>sanitation &amp; cleaning</group><group>street</group></groups>
```
struan: The description of this test should probably be changed.

PR #2642, opened 2019-09-19
[Northamptonshire] blue pins and no ID on report page

dracos:
For reports fetched from Alloy. [skip changelog]


dracos:
Altered to also remove ref from defect report pages

PR #2646, opened 2019-09-23
Restore single-line layout for admin index search inputs

zarino:
Fixes a bug that was introduced by 6ac0633d87da551f25c3d7d4cf2c902acf2a9e51, I think.

Screen Shot 2019-09-23 at 11 56 01

[skip changelog] because this is just fixing a bug in a feature that’s still in the "unreleased" section.

### DIFF
--- a/templates/web/base/admin/index.html
+++ b/templates/web/base/admin/index.html
@@ -20,18 +20,22 @@ and to receive notices of updates.
   </p>
 [% END %]
 
-<div class="admin-index-search form-txt-submit-box clearfix">
+<div class="admin-index-search clearfix">
 
 <form method="get" action="[% c.uri_for('reports') %]" accept-charset="utf-8">
-    <p><label for="search_reports">[% loc('Search Reports') %]</label>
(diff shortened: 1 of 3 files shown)


zarino:
> Should the .form-txt-submit-box be removed from the form parent div, then?

Oops, yes, how did I miss that.

@dracos I’ve added a second commit that has a go at fixing the styling of the `<select>` box to match the inputs further up. Ended up having to hard-code the width of the `<select>` to prevent it filling the full width of the parent, not sure why.

dracos:
"Ended up having to hard-code the width of the <select> to prevent it filling the full width of the parent, not sure why." - very long council names, so the select's default width is 100% (or larger). So you could set its max-width to 72%, which would kinda work in terms of giving 72/28, apart from the fact the proportions aren't right, the text boxes are more like 78/21% (I guess as they are smaller and extra space is distributed, I dunno, I probably messed up something). Anyway, the others are smaller and growing whereas this one is larger - you could ask it to shrink but it'd still want to take up 100% of width. So giving it same width as the text boxes then gives it the same flex growth behaviour.

PR #2651, opened 2019-09-23
[IsleOfWight] script to add emergency categories

struan:
One off script that adds an question to each Triage category to ask if
the problem is urgent or not. Sets urgent answer to block form
submission.
[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/island_roads_setup_messages
+++ b/bin/fixmystreet.com/island_roads_setup_messages
@@ -1,0 +1,73 @@
+#!/usr/bin/env perl
+#
+# One off script to transfer the hardcoded JS messages to the database
+
+use strict;
+use warnings;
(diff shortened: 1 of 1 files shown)


bin/fixmystreet.com/island_roads_setup_messages
```diff
@@ -0,0 +1,95 @@
...
+GetOptions(
+    'commit' => \$commit,
+);
+
+if (!$commit) {
+    say "*** DRY RUN ***";
+}
+
+my $iow = FixMyStreet::DB->resultset("Body")->find({ name => 'Isle of Wight Council' });
+if ($iow) {
```
dracos: You could save indenting everything below if you reversed and exited after printing error if no $iow.

Issue #2469, opened 2019-05-02
Improve design of admin body extra fields


dracos:
e.g. Have it display a list of the extra fields, only expanding to show form for one when selected; make each individual form nicer, including sub-form dropdown entry creation.

dracos:
This is all done, docs just done now.

PR #2653, opened 2019-09-25
Improve top-level part of admin category form

dracos:
One part of #2469 (the top-level part, and disable-form field top-level).

### DIFF
--- a/perllib/FixMyStreet/Roles/Extra.pm
+++ b/perllib/FixMyStreet/Roles/Extra.pm
@@ -135,6 +135,56 @@ sub push_extra_fields {
     $self->extra({ %$extra, $META_FIELD => [ @$existing, @fields ] });
 }
 
+=head2 update_extra_field
+
+    $problem->update_extra_field( { ... } );
+
+Given an extra field, will replace one with the same code in the
+existing list of fields, or add to the end if not present.
+
+=cut
+
+sub update_extra_field {
+    my ($self, $field) = @_;
+
+    # Can operate on list that uses code (Contact) or name (Problem),
+    # but make sure we have one of them
+    my $attr;
+    $attr = 'code' if $field->{code};
+    $attr = 'name' if $field->{name};
+    die unless $attr;
+
+    my $existing = $self->get_extra_fields;
+    my $found;
+    foreach (@$existing) {
+        if ($_->{$attr} eq $field->{$attr}) {
+            $_ = $field;
+            $found = 1;
+        }
+    }
+    if (!$found) {
+        push @$existing, $field;
+    }
+
+    $self->set_extra_fields(@$existing);
+}
+
+=head2 remove_extra_field
+
+    $problem->remove_extra_field( $code );
+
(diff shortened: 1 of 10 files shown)


templates/web/base/admin/bodies/contact-form.html
```diff
@@ -22,81 +22,74 @@ <h1>[% contact.category_display | html %]</h1>
...
-    <p>
-        [% loc("Check <strong>private</strong> if reports in this category should <strong>never be displayed on the website</strong>.
-                <br>
-                Normally, categories are not private.
-                <br>
-                This is suitable for issues that you want to allow users to report to the body, but for which there is no public
-                interest in displaying the report. In the UK, we've used this for services like requesting an extra rubbish bin
-                at a specific address.") %]
+    <p class="form-check">
+        <input type="checkbox" name="disable" value="1" id="disable" data-toggle-visibility="#js-disable-form-message-box" [% ' checked' IF contact.disable_form_field %]>
```
davea: Setting the "Disable form when this category is selected" tickbox will wipe out any extra fields that have already been set on this contact - is that intentional? I think it might stem from the use of `name` in the `Extra::update_extra_field` comparison (Extra.pm L153) but `code` on L296 of Bodies.pm, but also possibly a bug in the `update_extra_field` function when the field doesn't already exist.

perllib/FixMyStreet/Roles/Extra.pm
```diff
@@ -135,6 +135,49 @@ sub push_extra_fields {
...
+=head2 update_extra_field
+
+    $problem->update_extra_field( { ... } );
+
+Given an extra field, will replace one with the same code in the
+existing list of fields, or add to the end if not present.
+
+=cut
+
+sub update_extra_field {
```
davea: It'd be good to have some tests of these new functions. I think there's a bug where the extra fields are replaced entirely with the passed `$field` if it doesn't already exist - see other comment.

templates/web/base/admin/bodies/contact-form.html
```diff
@@ -22,81 +22,74 @@ <h1>[% contact.category_display | html %]</h1>
...
-  <p>
-    <input type="checkbox" name="non_public" value="1" id="non_public" [% ' checked' IF contact.non_public %]>
-    <label for="non_public" class="inline">[% loc('Private') %]</label>
-  </p>
 
-  <div class="admin-hint">
-    <p>
-      [% loc("Check <strong>inspection required</strong> if reports in this category <strong>must be inspected</strong> before being sent.") %]
+    <p class="form-group form-group--indented [% 'hidden-js' IF NOT contact.disable_form_field %]" id="js-disable-form-message-box">
+        <label for="disabled-message">[% loc('Message to show when form is disabled') %]</label>
```
davea: These messages are likely to include links to other pages - should we tell the user they should use HTML and/or automatically upgrade plain URLs to hyperlinks somehow?

dracos:
Have rolled this branch back to then so it can be merged now once tests pass, then new bits can be done on top.

PR #2659, opened 2019-09-30
[IsleOfWight] only hide old reports on live site

struan:
[skip changelog]


struan:
force pushed with 977a546cd51fa0e9d88451dcdfe59c68c9017e8d

PR #2661, opened 2019-10-01
[IsleOfWight] update reference number text

struan:
Remove reference to council from ref no on report page and rss link on all reports page
[skip changelog]

### DIFF
--- a/templates/web/isleofwight/report/_council_sent_info.html
+++ b/templates/web/isleofwight/report/_council_sent_info.html
@@ -1,0 +1,17 @@
+[% SET duration_clause = problem.duration_string(c) %]
+[% IF duration_clause || problem.can_display_external_id %]
+    <p class="council_sent_info">
+    [%- IF problem.can_display_external_id %]
+        [%- IF duration_clause %]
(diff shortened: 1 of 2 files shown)


templates/web/isleofwight/report/_council_sent_info.html
```diff
@@ -0,0 +1,17 @@
+[% SET duration_clause = problem.duration_string(c) %]
+[% IF duration_clause || problem.can_display_external_id %]
```
dracos: Doesn't need to call this function here if it has its own template, I guess. Shrug. Does it matter if it still says council on .com, as it currently does.

templates/web/isleofwight/report/_council_sent_info.html
```diff
@@ -0,0 +1,17 @@
...
+    <p class="council_sent_info">
+    [%- IF problem.can_display_external_id %]
+        [%- IF duration_clause %]
+            [%- external_ref_clause = tprintf('Enquiry ref:&nbsp;%s', problem.external_id) %]
```
dracos: If this is only line that changes from parent, might be good to factor out to separate template? But might be changed due to other changes, not sure.

PR #2662, opened 2019-10-01
[IsleOfWight] use appropriate asset server on staging/live

struan:
Change the tilma url based on whether we are using the live or staging
site.
[skip changelog]

### DIFF
--- a/web/cobrands/isleofwight/assets.js
+++ b/web/cobrands/isleofwight/assets.js
@@ -4,9 +4,14 @@ if (!fixmystreet.maps) {
     return;
 }
 
+var is_live = false;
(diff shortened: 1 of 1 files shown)


web/cobrands/isleofwight/assets.js
```diff
@@ -4,9 +4,14 @@ if (!fixmystreet.maps) {
     return;
 }
 
+var is_live = false;
+if ( location.hostname === 'www.fixmystreet.com' || location.hostname == 'fms.islandroads.com' ) {
```
dracos: triple equal second one to be consistent :)

PR #2663, opened 2019-10-01
Restructured admin category extra-metadata-form

zarino:
The second part of #2469 (following on from #2653).

* New "Behaviour" radio group consolidates two previously separate fields (the "Automated" select box, and the "Notice" checkbox) since the features really were mutually exclusive.
* Metadata fields are collapsable (and collapsed by default), to help you get a top-level overview of all the questions / hidden fields attached to a category, in one go.
* Metadata fields are now draggable, so you can reorder fields without typing in "Order" numbers.
* The sub-options for each field (eg: "Question text", "Required" checkbox, etc) are hidden until previous selections show that they are required, making the form a bit less intimidating for new users.
* Some improved accessibility, with `<fieldset>` and `<legend>` elements.
* Simplified `admin.js` somewhat, with less cloning/manipulating of DOM elements for the repeatable components.

@dracos – a few questions for you to consider while reviewing:

1. Right now, without JavaScript, the title in each field’s grey `<header>` is always "New field". Do you think this is ok? Or should the Perl template create the "*[code]* / [behaviour] / [description]" title too? Feels awkward duplicating that title-building logic in both Perl and JavaScript code, but maybe it’s worth it for more consistency?
1. This removes the "Hint" field. Is that ok? Or should it be re-introduced?
2. The `toggle_visibility` stuff is all in `fixmystreet.js`, but obviously so far it’s only used on this admin page. Eventually it’d be nice to see it used across the site, wherever we need to show/hide elements based on user choices – but in the meantime, are you happy for it to stay in `fixmystreet.js`, or should it go into `admin.js`?

## Screenshot – with metadata fields collapsed

collapsed

## Screenshot – with one metadata field expanded

expanded

[skip changelog]


dracos:
I've written all the server side changes needed for this, fixed some bugs, added point 1 above, and got the tests to pass, so probably shouldn't review it any longer :)

templates/web/base/admin/extra-metadata-option.html
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+    <div class="row">
+        <div class="col">
+            <p class="form-check">
+                <input type="checkbox" name="metadata[[% i %]].values[[% j %]].disable" value="1" id="metadata-[% i %]-values-[% j %]-disable" data-toggle-visibility="#js-disable-form-message-box-[% i %]-[% j %]" [% ' checked' IF option.disable %]>
+                <label for="metadata-[% i %]-values-[% j %]-disable">[% loc('Disable form when this option is selected') %]</label>
+            </p>
+            <p class="form-group [% 'hidden-js' IF NOT option.disable %]" id="js-disable-form-message-box-[% i %]-[% j %]">
+                <label for="disabled-message-[% i %]-[% j %]">[% loc('Message to show when form is disabled (HTML permitted)') %]</label>
+                <textarea id="disabled-message-[% i %]-[% j %]" name="metadata[[% i %]].values[[% j %]].datatype_description" class="form-control">[% contact.disable_form_field.description %]</textarea>
```
struan: This doesn't seem to be saved anywhere.

dracos: Whoops! Good opportunity to stop using datatype_description in fact.

PR #2670, opened 2019-10-07
[Westminster] Hide fixmystreet.com updates on Westminster cobrand

davea:
Ensures that comments left on Westminster reports on fixmystreet.com don't show up on the Westminster cobrand.

[skip changelog]

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1509

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -183,9 +183,9 @@ sub load_problem_or_display_error : Private {
 sub load_updates : Private {
     my ( $self, $c ) = @_;
 
-    my $updates = $c->model('DB::Comment')->search(
-        { problem_id => $c->stash->{problem}->id, state => 'confirmed' },
-        { order_by => [ 'confirmed', 'id' ] }
+    my $updates = $c->cobrand->updates->search(
+        { problem_id => $c->stash->{problem}->id, "me.state" => 'confirmed' },
+        { order_by => [ 'me.confirmed', 'me.id' ] }
     );
 
     my $questionnaires_still_open = $c->model('DB::Questionnaire')->search(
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -183,9 +183,9 @@ sub load_problem_or_display_error : Private {
 sub load_updates : Private {
     my ( $self, $c ) = @_;
 
-    my $updates = $c->model('DB::Comment')->search(
-        { problem_id => $c->stash->{problem}->id, state => 'confirmed' },
-        { order_by => [ 'confirmed', 'id' ] }
+    my $updates = $c->cobrand->updates->search(
```
dracos: I guess it probably makes sense to make a similar change to Contact.pm (moderation of another cobrand's update) and Photo.pm (showing photo of another cobrand's update if requested)? (And perhaps Questionnaire.pm for parity with here, though don't think that would have any affect anywhere.)

perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -153,4 +153,13 @@ sub categories_restriction {
         }, { join => 'body' });
 }
 
+sub updates_restriction {
+    my $self = shift;
+
+    # Westminster don't want any fms.com updates shown on their cobrand.
+    return $self->next::method(@_)->search({
+        "me.cobrand" => 'westminster'
```
dracos: All the Open311-imported updates do not have a cobrand at all, which means this will hide all of them. Given any .com created ones will have a cobrand, and you can't create westminster updates, I guess checking for blank string here instead would do?

davea: Oops! I've updated it to specifically exclude only `fixmystreet` cobrand updates, just in case we enable updates on Westminster in the future and wonder why they're not appearing :)

PR #2674, opened 2019-10-08
[Northamptonshire] updated report status backfill script

struan:
Update the script for updating all statuses rather than just the
investigating ones

[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/one-off-status-update
+++ b/bin/fixmystreet.com/one-off-status-update
@@ -19,9 +19,10 @@ It takes a single command line argument which is the path to the csv
 file containing the reports to update, along with details of the current
 alloy state which it uses to determine what state to update them too.
 
-Only reports that match the state in the `fms_state` column of the csv
-file will be updated. Similarly, if it cannot determine what state to
-update the report to it will skip that report.
+Only reports that do not match the state in the spreadsheet will be updated.
+Reports that are closed/fixed on FixMyStreet but open in the spreadsheet
+will also be ignored. It will also skip reports that it cannot determine
+the state for.
 
 Any update made will set the appropriate response template.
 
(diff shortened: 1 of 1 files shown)


bin/fixmystreet.com/one-off-status-update
```diff
@@ -78,13 +78,26 @@ if (!$commit) {
     say "*** DRY RUN ***";
 }
 
-my %alloy_to_fms_map = (
+my %enq_to_fms_map = (
+    'Issued to Inspector' => 'investigating',
+    'In Progress' => 'investigating',
```
dracos: not "in progress"?

bin/fixmystreet.com/one-off-status-update
```diff
@@ -78,13 +78,26 @@ if (!$commit) {
...
+    'Completed' => 'closed',
```
dracos: Do they close rather than mark as fixed?

bin/fixmystreet.com/one-off-status-update
```diff
@@ -106,19 +119,31 @@ if ($northants) {
...
             warn "skipping $id as can't determine new state\n" if $verbose;
             next;
         }
 
+        if ($p->state eq $new_state) {
+            warn "skipping $id as has correct state\n" if $verbose;
+            next;
+        }
+
+        if ( ($p->fixed_states->{$p->state} || $p->closed_states->{$p->state})
```
dracos: The code generally uses `is_fixed` or `is_closed`

bin/fixmystreet.com/one-off-status-update
```diff
@@ -78,13 +78,26 @@ if (!$commit) {
...
+    'Issued to Inspector' => 'investigating',
+    'In Progress' => 'investigating',
+    'Completed' => 'closed',
+);
+
+my %rfc_to_fms_map = (
     'No Action Necessary' => 'unable to fix',
     'Outside NCC Control' => 'not responsible',
     'Highways to Monitor. No Action' => 'unable to fix',
     'Work Instructed' => 'in progess',
```
dracos: This is action scheduled in the live config.

bin/fixmystreet.com/one-off-status-update
```diff
@@ -78,13 +78,26 @@ if (!$commit) {
...
 );
 
+my %defect_to_fms_map = (
+    'Remedied' => 'fixed - council',
+    'Order Raised' => 'in progress',
```
dracos: This is action scheduled in the live config.

bin/fixmystreet.com/one-off-status-update
```diff
@@ -78,13 +78,26 @@ if (!$commit) {
...
+    'Rejected' => 'closed',
+    'Found and Forwarded' => 'closed',
```
dracos: Should these two be in the live config also then? As that defaults to open on these two.

bin/fixmystreet.com/one-off-status-update
struan: Completed is the state for an inspection essentially. It doesn't get marked as fixed unless a defect is raised. And in theory there should be a reason for closure set when an inspection is completed which will override this, but we need a default.

bin/fixmystreet.com/one-off-status-update
struan: No, because in progress in this case means an inspection is in progress, rather than fixing the problem is in progress.

PR #2679, opened 2019-10-10
[IsleOfWight] report archive template

struan:
[skip changelog]


struan:
Good point, I've added in an archive.html if you want to have a quick squiz of it.

davea:
Looks good!

PR #2688, opened 2019-10-16
[Northamptonshire] small tweaks to backfill script

struan:
Fix a couple of issues discovered in testing where the Response to
customer wasn't being detected in the database;
[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/one-off-status-update
+++ b/bin/fixmystreet.com/one-off-status-update
@@ -119,6 +119,9 @@ if ($northants) {
             next;
         }
 
+        # we don't want to update hidden reports
+        next if $p->state eq 'hidden';
+
         my $new_state = get_state( $report );
 
         unless ( $new_state ) {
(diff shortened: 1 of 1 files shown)


bin/fixmystreet.com/one-off-status-update
```diff
@@ -140,7 +143,15 @@ if ($northants) {
         my $text = $report->{Response_to_Customer};
         # do not add a second response to customer if it's already there.
         if ($text) {
-            my $c = $p->comments->search({text => $text});
+            # sometimes there are extra spaces on the end in the database
+            my $search = { text => { ilike => "$text%" } };
+            # there's quite a few occurrences of the text being corrected that match this
+            if ( $text =~ /.+thank you[.\s]*$/i ) {
```
dracos: No need for .+ at the start here. Though, why have the if at all, it won't remove anything if it's not present.

bin/fixmystreet.com/one-off-status-update
```diff
@@ -140,7 +143,15 @@ if ($northants) {
...
+                my $search_text = $text;
+                $search_text =~ s/\sthank you[.\s]*$//;
+                $search = { text => { ilike => "$search_text%" } };
```
dracos: Do you want a % at the start as well, or are you happy that isn't necessary?

struan: I don't especially as there's a few places where they've added updates on the the start of the response which are important.

PR #2689, opened 2019-10-17
Improve fetch scripts

dracos:
This PR combines fetch-comments and fetch-reports into one fetch script, and allows command line selection of start/end times.

It also (March 2020) parallelizes the fetching of comments - first commit providing a config number of workers to run in parallel, the second making a min/max with a timeout so you can start with lower and add more if backends are slow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
         - Refactor Script::Report into an object.
         - Move summary failures to a separate script.
         - Add script to export/import body data.
+        - Add fetch script that does combined job of fetch-comments and fetch-reports.
     - UK:
         - Added junction lookup, so you can search for things like "M60, Junction 2"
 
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
         - Allow cobrands to disable admin resending.
         - Sass variables for default link colour and decoration.
         - Make contact edit note optional on staging sites.
+        - Combine fetch-comments and fetch-reports scripts.
```
chrismytton: I feel like we should explicitly mention that people with an FMS install will need to update any cronjobs to use the new `bin/fetch` script. Relatedly it would also be good to have a bit in the release notes about the mapping between old and new commands. So `bin/fetch-comments` becomes `bin/fetch --updates`, `bin/fetch-comments-24hs` becomes `bin/fetch --updates --start 24` etc (or I guess could just point them at the example crontab, or the relevant commit?).

PR #2692, opened 2019-10-18
[Bexley] Map NSG_REF as site_code for Confirm

chrismytton:
Bexley's Confirm server needs the `NSG_REF` attribute from the Streets layer to appear in the `site_code` field.

We already had a `usrn` configuration for Bexley's assets, so this change includes a modification to the code that handles the `usrn` mapping to allow it to accept an array of configuration objects as well as just a single object.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1638

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -236,14 +236,18 @@ OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorAsset, 
 
     updateUSRNField: function() {
         if (this.fixmystreet.usrn) {
-            var usrn_field = this.fixmystreet.usrn.field;
-            var selected_usrn;
-            if ( this.selected_feature ) {
-                selected_usrn = this.fixmystreet.getUSRN ?
-                    this.fixmystreet.getUSRN(this.selected_feature) :
-                    this.selected_feature.attributes[this.fixmystreet.usrn.attribute];
+            if (!this.fixmystreet.usrn.length) {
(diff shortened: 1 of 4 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -197,14 +197,19 @@ OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorAsset,
 
     updateUSRNField: function() {
         if (this.fixmystreet.usrn) {
-            var usrn_field = this.fixmystreet.usrn.field;
-            var selected_usrn;
-            if ( this.selected_feature ) {
-                selected_usrn = this.fixmystreet.getUSRN ?
+            if (!this.fixmystreet.usrn.length) {
+                this.fixmystreet.usrn = [this.fixmystreet.usrn];
```
dracos: Nice how this will only do it the once :)

web/cobrands/fixmystreet/assets.js
```diff
@@ -197,14 +197,19 @@ OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorAsset,
...
+            }
+            for (var i = 0; i < this.fixmystreet.usrn.length; i++) {
```
dracos: This is fine, just noting as we have jQuery here anyway, it has an `$.each(array, function(idx, val) { })` function. Really should just polyfill `forEach` properly, shouldn't we, except we'd still have the jQuery fn :)

PR #2699, opened 2019-10-23
[Northamptonshire] script to update emergency message

struan:
To avoid having to do it all by hand

[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/add_emergency_message
+++ b/bin/fixmystreet.com/add_emergency_message
@@ -21,10 +21,15 @@ my ($opts, $usage) = describe_options(
     '%c %o',
     ['commit', 'whether to commit changes to the database' ],
     ['body=s', 'name of body to attach question to', { required => 1 } ],
-    ['code=s', 'code to use for question', { required => 1 } ],
-    ['question=s', 'question to ask user first', { required => 1 } ],
-    ['yes=s', 'yes answer to question', { required => 1 } ],
-    ['no=s', 'no answer to question', { required => 1 } ],
+    ['mode' => 'hidden' => { one_of => [
+        ['questions' => 'add an emergency question'],
+        ['category' => 'add an emergency message'],
+    ], required => 1 }],
+    ['update' => 'only update existing messages'],
+    ['code=s', 'code to use for question'],
+    ['question=s', 'question to ask user first'],
+    ['yes=s', 'yes answer to question'],
+    ['no=s', 'no answer to question'],
     ['message=s', 'message to be shown if form disabled', { required => 1 } ],
     ['send_method=s', 'send method to restrict categories to' ],
     ['help|h', "print usage message and exit" ],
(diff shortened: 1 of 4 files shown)


struan:
Sadly, not as the Northants doesn't use the standard mechanism because it predates it. I have updated this though to convert it to the standard one, and added the ability to update this using `add_emergency_message` in future.

perllib/FixMyStreet/Roles/Extra.pm
```diff
@@ -243,4 +243,24 @@ sub get_extra_field_value {
...
+Can use either `name` or `code` to identify the field.
+
+=cut
+
+sub get_extra_field {
+    my ($self, %opts) = @_;
+
+    my @fields = @{ $self->get_extra_fields() };
+
+    my $comparison = $opts{code} ? 'code' : 'name';
```
dracos: If you add this line to `get_extra_field_value` above, isn't it then the same? :)

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,27 +40,44 @@ if (!$opts->commit) {
...
-        {
-            key => 'no',
-            name => $opts->no,
-        }
-    ],
-};
+my $field;
+
+if ( $opts->mode eq 'questions' ) {
+    say STDERR "questions, code, yes and no required"
```
dracos: this should die?

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,27 +40,44 @@ if (!$opts->commit) {
...
+        ],
+    };
+} else {
+    $field = {
+        order => 0,
+        required => 'true',
+        protected => 'true',
+        code => '_fms_disable_',
+        description => $opts->message,
+        datatype => 'notice',
```
dracos: datatype is either string or singlevaluelist (for question, ignored for others now, so I think can be left off).

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,27 +40,44 @@ if (!$opts->commit) {
...
+        variable => 'true',
```
dracos: Think this should be false.

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,27 +40,44 @@ if (!$opts->commit) {
...
+                disable_message => $opts->message,
+            },
+            {
+                key => 'no',
+                name => $opts->no,
+            }
+        ],
+    };
+} else {
+    $field = {
```
dracos: I think this is missing `disable_form`?

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,27 +40,44 @@ if (!$opts->commit) {
...
+        order => 0,
+        required => 'true',
```
dracos: required isn't needed for notices.

bin/fixmystreet.com/add_emergency_message
```diff
@@ -65,12 +87,15 @@ unless ($body) {
 
 my $contacts = $body->contacts->not_deleted;
 $contacts = $contacts->search({ send_method => $opts->send_method }) if $opts->send_method;
+my $field_code = $opts->code || '_fms_disable_';
```
dracos: This could just be $field->code

perllib/FixMyStreet/Roles/Extra.pm
struan: No, because it returns `$field->{value}`. I wanted the whole `$field`.

dracos: Hmm, yes, I see; I'd perhaps argue getting rid of get_extra_field_value then (there is a rare bug on reports that fail to send where if it's present but blank it adds more and more entries instead of updating the extra field see e.g. report 1771619), not sure.

bin/fixmystreet.com/add_emergency_message
```diff
@@ -35,42 +40,60 @@ if (!$opts->commit) {
...
+        code => '_fms_disable_',
+        description => $opts->message,
+        variable => 'false',
+    };
+}
 
 my $body = FixMyStreet::DB->resultset("Body")->find({ name => $opts->body });
 unless ($body) {
-    say STDERR "Could not find body " . $opts->body;
+    die "Could not find body " . $opts->body;
```
dracos: Afraid you changed the wrong one - this one already exits on the next line, it was line 46 I was suggesting needs to die/exit.

struan: doh!

perllib/FixMyStreet/Roles/Extra.pm
struan: I dunno, it seems like quite a useful shortcut.

PR #2700, opened 2019-10-25
[TfL] Initial cobrand design & smaller features

davea:
Mostly cosmetic & configuration changes so far.

(The JSON file for `import_categories` can be found in `crow:~davea/tfl_categories.json`)

[skip changelog]


templates/web/tfl/footer_extra_js.html
```diff
@@ -0,0 +1,11 @@
+[%
+IF bodyclass.match('mappage');
+  scripts.push(
+    version('/vendor/OpenLayers.Projection.OrdnanceSurvey.js'),
+    version('/cobrands/fixmystreet-uk-councils/roadworks.js'),
```
dracos: Technically I guess this line should be in the roadworks commit :)

perllib/FixMyStreet/DB/Result/Body.pm
```diff
@@ -224,6 +224,17 @@ has a council_name, use it in preference to the body name.
 
 sub cobrand_name {
     my $self = shift;
+
+    # Because TfL covers all the boroughs in London, get_cobrand_handler
+    # may return another London cobrand if it is listed before tfl in
+    # ALLOWED_COBRANDS, because one of this body's area_ids will also
```
dracos: Could we just mandate that 'tfl' must always come before any other London cobrands in that list? :)

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -0,0 +1,33 @@
+package FixMyStreet::Cobrand::TfL;
+use parent 'FixMyStreet::Cobrand::UKCouncils';
```
dracos: Want to make this Whitelabel, then can drop the empty front/footer-marketing.html template from this commit.

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -0,0 +1,33 @@
...
+sub area_types  { [ 'LBO' ] }
+
+sub area_check {
+    my ( $self, $params, $context ) = @_;
+
+    my $councils = $params->{all_areas};
+    my $council_match = grep { $councils->{$_} } @{ $self->council_area_id };
+
+    return 1 if $council_match;
+    return (0, "That location is not covered by TfL");
```
dracos: Do we/can we link to .com here like UKCouncils does?

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -30,4 +30,23 @@ sub enter_postcode_text {
     return 'Enter a London postcode, or street name and area';
 }
 
+sub owns_problem {
```
dracos: I think this would be clearer overall if `is_council` is set to 0, which fixes a few other places that would then assume it could call TfL a council etc, and then this function is never called and isn't necessary, as far as I can see. Looks like index-steps.html might be only thing needing to then be added.

bin/tfl/import_categories
```diff
@@ -0,0 +1,97 @@
...
+$db->txn_begin;
+if (!$opt->commit) {
+    say "NOT COMMITTING TO DATABASE";
+}
+
+my $config = decode_json(path($ARGV[0])->slurp_utf8);
+
+my $body = FixMyStreet::DB->resultset('Body')->find({ name => 'TfL' });
+
+say "Found body " . $body->name;
```
dracos: Probably want to die if not found :)

bin/tfl/import_categories
```diff
@@ -0,0 +1,97 @@
...
+        my $child_cat = FixMyStreet::DB->resultset("Contact")->find_or_new({
+            body => $body,
+            category => $cat->{category}
+        });
+        $child_cat->email($cat->{email});
+        $child_cat->state('confirmed');
+        $child_cat->editor($0);
+        $child_cat->whenedited(\'current_timestamp');
+        $child_cat->note($child_cat->in_storage ? 'Updated by import_categories' : 'Created by import_categories');
+        my $groups = $child_cat->groups;
```
dracos: There aren't any child groups in the example data (in case this code isn't needed any more).

bin/tfl/import_categories
```diff
@@ -0,0 +1,97 @@
...
+                description => $cat->{disable} eq 1 ? $config->{disabled_message} : $cat->{disable},
+                order => 0,
+            });
+        }
+        $child_cat->set_extra_fields(@{ $cat->{extra_fields} }) if $cat->{extra_fields};
+        if (my $asset_field = $cat->{asset_field}) {
+            my ($description, $code) = @$asset_field;
+            $child_cat->update_extra_field({
+                code => $code,
+                variable => "true",
```
dracos: Shouldn't need variable for hidden_field.

bin/tfl/import_categories
davea: The `Other` category appears in multiple groups, so might already have had the group set by an earlier iteration.

dracos: Ah yes In which case, only slight annoyance is it creating many history items for that.

davea: Yeah, bit annoying, but no big deal as the script should only be run the once.

templates/web/tfl/report/_council_sent_info.html
```diff
@@ -0,0 +1,17 @@
+[% SET duration_clause = problem.duration_string(c) %]
+[% IF duration_clause || problem.whensent %]
+    <p class="council_sent_info">
+    [%- IF problem.whensent %]
+        [%- IF duration_clause %]
+            [%- external_ref_clause = tprintf(loc('Council ref:&nbsp; FMS%s'), problem.id) %]
+        [%- ELSE %]
+            [%- external_ref_clause = tprintf(loc('%s ref:&nbsp; %s'), problem.external_body, problem.id) %]
```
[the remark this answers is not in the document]
davea: Oops, nope.

templates/web/tfl/tokens/_confirm_problem_council_id.html
```diff
@@ -0,0 +1,2 @@
+<h2>Your issue is on its way to TfL.</h2>
+<p>Your reference for this report is FMS[% report.id %], please quote it in any enquiries.</p>
```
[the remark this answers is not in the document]
davea: 4656524218e80b2c79306e59a76eff521538dab7 in `tfl-cobrand` introduces some tests for this; but the commit it's part of is probably not going to remain as-is. I'll rejig the tests out of there into this branch.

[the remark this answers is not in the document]
davea: Done in 947f09516f45a285a631906e8f95ecf834e48e04

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -69,4 +69,6 @@ sub lookup_by_ref {
     return 0;
 }
 
+sub report_sent_confirmation_email { 'id' }
```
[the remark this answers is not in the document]
davea: Done in 018c2a5e3f13c4a03f721fe7b448633c35e88937

perllib/FixMyStreet/DB/Result/Body.pm
davea: I've got a vague feeling that doing that caused another issue, but in my testing now I can't see anything wrong.

I guess it's a toss-up between including cobrand-specific code in the core modules and introducing a non-obvious requirement for `ALLOWED_COBRANDS` to be ordered. I suppose for the purposes of our deployments enforcing an ordering would be OK.

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -22,7 +22,7 @@ sub area_check {
     my $council_match = grep { $councils->{$_} } @{ $self->council_area_id };
 
     return 1 if $council_match;
-    return (0, "That location is not covered by TfL");
+    return (0, $self->area_check_error_message);
```
dracos: Need to pass in params/context?

davea: 🤦‍♂ Of course.

perllib/FixMyStreet/DB/Result/Body.pm
dracos: Actually on the staging site now I'm seeing "reported to Bexley or Bexley" in the HTML. It's because the Cobrand.pm body_handler check does $areas->{id} which won't work for TfL when id is an array. So it wasn't to do with cobrand order, it would never have returned TfL if I've understood it right. So sorry, perhaps easiest just to put this bit back.

PR #2701, opened 2019-10-25
Two-factor improvements

dracos:
This PR extends two-factor authentication to non-superuser staff accounts (well, anyone in theory, but the setup page is only present for staff users), allows you to mandate 2FA for users on a cobrand (and does so for superusers on .com), asks for a 2FA code during 2FA setup (so we know you've set it up okay), fixes being able to reset your password with 2FA turned on, and improves the Security page layout.


dracos:
"One question I did have: what happens if a cobrand enforces 2FA for from_body users, but the user logs in on .com without having set up 2FA? Presumably this would let them login." - if a UK cobrand wishes to enforce 2FA, then .com needs a matching change to its `must_have_2fa` function also, to force them to do so there as well.

davea:
OK, that's reasonable.

davea:
One thing that just occurred to me is that if a user sets up 2FA when they login as part of making a report, that report gets lost.

e.g.:
 1. Create a user without 2FA set up
 1. Make that user a superuser
 1. In a logged-out session, start a new report
 1. Click the "Or sign in with password to prefill this information." link in the second section of report form
 1. Login with that user's email & password
 1. See prompt to set up 2FA
 1. Set up 2FA
 1. See the confirmation screen that 2FA has been set up, only link to follow is to the 'your account' page; report content has been lost

It's a bit of a stretch to think that people will run into this on a regular basis, but it is possible.

dracos:
Ooh, ta, that flow should work fine, have added a test and fixed the issue meaning it wasn't working in the latest commit -  94eb4a1

dracos:
Can open PR after this is merged, but just to note https://github.com/mysociety/fixmystreet/commit/df24f00927b2a1aa53cf9b8ab6f65d06625d3991 is TfL specific bit on top of this.

PR #2703, opened 2019-10-29
[TfL] Improvements to inspector areas functionality

davea:
Inspectors have been able to be assigned to areas for a long time, but site has never really done anything with those areas.

The main feature this PR adds is to automatically filter the 'All Reports' page to the areas an inspector is assigned to, if any. Inspectors are also taken to this page post-login.

TfL inspectors will be assigned to Borough-level areas, not wards, so there are also several commits to support assigning users to areas specified by the cobrand, as well as changing the areas used for the wards list on /reports.

NB this PR is obviously lacking tests and there are still some rough edges to be polished (e.g. removing the references to 'wards' on the TfL cobrand), but I'm requesting a review for feedback on the general approach taken.

[skip changelog]


perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -209,8 +209,12 @@ sub edit : Path : Args(1) {
     }
 
     $c->forward('/admin/fetch_all_bodies');
-    $c->forward('/admin/fetch_body_areas', [ $user->from_body ]) if $user->from_body;
-    $c->forward('fetch_body_roles', [ $user->from_body ]) if $user->from_body;
+    if ($user->from_body) {
+        unless ($c->cobrand->call_hook(admin_fetch_inspector_areas => $user->from_body)) {
+            $c->forward('/admin/fetch_body_areas', [ $user->from_body ]);
+        }
+        $c->forward('fetch_body_roles', [ $user->from_body ]) if $user->from_body;
```
dracos: No need for if at end of this line

perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -209,8 +209,12 @@ sub edit : Path : Args(1) {
     }
 
     $c->forward('/admin/fetch_all_bodies');
-    $c->forward('/admin/fetch_body_areas', [ $user->from_body ]) if $user->from_body;
-    $c->forward('fetch_body_roles', [ $user->from_body ]) if $user->from_body;
+    if ($user->from_body) {
+        unless ($c->cobrand->call_hook(admin_fetch_inspector_areas => $user->from_body)) {
```
dracos: This is okay, but I think would be better if this hook wasn't here (and below) but inside `admin/fetch_body_areas`, and that way the hook wouldn't have to duplicate the name sorting or setting of fetched_areas_body_id.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -45,6 +45,32 @@ sub my : Path : Args(0) {
...
 
+
+=head2 inspector_redirect
+
+A convenience redirect to the /reports/ page pre-filtered to the
+inspector's body, areas & categories.
+
+=cut
+
+sub inspector_redirect : Local : Args(0) {
```
dracos: Nice

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -401,7 +401,7 @@ sub ward_check : Private {
         $parent_id = $c->stash->{area}->{id};
     }
 
-    my $qw = FixMyStreet::MapIt::call('area/children', [ $parent_id ],
+    my $qw = $c->cobrand->call_hook('all_reports_ward_areas') || FixMyStreet::MapIt::call('area/children', [ $parent_id ],
```
dracos: Instead of being a hook, this could be a function in Default.pm that returns the default, and overridden in TfL, passing in ID and cobrand, then could be shared here and in Body.pm location too.

davea: Is there a situation where the cobrand on which that function is being called would be different to the cobrand being passed in?

t/app/controller/report_inspect.t
```diff
@@ -79,6 +79,74 @@ FixMyStreet::override_config {
...
+            is $mech->uri->path, $test->{destination}, 'redirected to correct destination';
+            is_deeply { $mech->uri->query_form }, $test->{query_form}, 'destination query params set correctly';
+            if ($test->{previous}) {
+                is $mech->res->previous->code, 302, "got 302 for post-login redirect";
+                is $mech->res->previous->base->path, $test->{previous}, "previous URI correct";
+            }
+
+            $mech->get_ok("/");
+            ok $mech->find_link( text => 'All reports', url => $test->{good_link} );
+            ok !$mech->find_link( text => 'All reports', url => $test->{bad_link} );
```
dracos: If the first test passes, could the second test ever fail?

t/Mock/MapIt.pm
```diff
@@ -115,6 +115,11 @@ sub dispatch_request {
             $self->output({2326 => {parent_area => undef, id => 2326, name => "Cheltenham Borough Council", type => "DIS"}});
         } elsif ($areas eq 'UTA') {
             $self->output({2650 => {parent_area => undef, id => 2650, name => "Aberdeen Council", type => "UTA"}});
+        } elsif ($areas eq 'LBO') {
+            $self->output({
+                2482 => {parent_area => undef, id => 2482, name => "Bromley Borough Council", type => "LBO"},
+                2483 => {parent_area => undef, id => 2483, name => "Hounslow Borough Council", type => "LBO"},
```
dracos: Just to note another PR adds this but with slightly different councils, for any future merge/tidying.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -1163,7 +1163,7 @@ sub fetch_all_bodies : Private {
 sub fetch_body_areas : Private {
     my ($self, $c, $body ) = @_;
 
-    my $children = $body->first_area_children;
+    my $children = $c->cobrand->call_hook('admin_fetch_inspector_areas', $body) || $body->first_area_children;
```
dracos: After all the fixups, I think what's fallen out is this hook can be removed? Because it now doesn't do anything, as fetch_area_children does it all anyway.

t/app/controller/report_inspect.t
davea: You'd hope not, but thought belt-and-braces in case some future commit adds a new 'all reports' link or something.

davea:
Merged in 9f1defd8f1759f6d0b7c320b5e525b035b2d0e72.

PR #2712, opened 2019-11-04
[TfL] Inspector form tweaks

davea:
- Removes priority and traffic management fields & admin permissions
 - Ensures only TfL response templates are visible, not all London cobrands

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -375,10 +375,16 @@ sub inspect : Private {
     $c->forward('/admin/categories_for_point');
     $c->stash->{report_meta} = { map { 'x' . $_->{name} => $_ } @{ $c->stash->{problem}->get_extra_fields() } };
 
-    if ($c->cobrand->can('council_area_id')) {
-        my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+    if ($c->cobrand->can('body')) {
+        my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories(
+            $c->stash->{contacts},
+            body_ids => [ $c->cobrand->body->id ]
+        );
         $c->stash->{priorities_by_category} = $priorities_by_category;
-        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories(
+            $c->stash->{contacts},
+            body_ids => [ $c->cobrand->body->id ]
+        );
         $c->stash->{templates_by_category} = $templates_by_category;
     }
 
--- a/perllib/FixMyStreet/Roles/ContactExtra.pm
+++ b/perllib/FixMyStreet/Roles/ContactExtra.pm
@@ -25,8 +25,15 @@ sub for_bodies {
 }
 
 sub by_categories {
-    my ($rs, $area_id, @contacts) = @_;
-    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my ($rs, $contacts, %params) = @_;
+
+    my %body_ids = ();
+    if ( $params{body_id} ) {
+        %body_ids = ( $params{body_id} => 1 );
+    } else {
+        %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $params{area_id} });
+    }
+    my @contacts = @$contacts;
     my @body_ids = keys %body_ids;
     my %extras = ();
     my @results = $rs->for_bodies(\@body_ids, undef);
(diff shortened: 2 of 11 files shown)


templates/web/tfl/report/inspect/extra_details.html
```diff
@@ -0,0 +1,12 @@
+[% IF permissions.report_inspect %]
```
dracos: If you're having this on its own, it might as well be its own template in the base too then can be shared more easily?

perllib/FixMyStreet/Roles/ContactExtra.pm
```diff
@@ -25,8 +25,10 @@ sub for_bodies {
 }
 
 sub by_categories {
-    my ($rs, $area_id, @contacts) = @_;
-    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my ($rs, $area_id, $contacts, $bodies) = @_;
```
dracos: So $area_id isn't used if $bodies is provided? I think this would be clearer if you passed in a hash of arguments (so area_id *or* bodies), then bodies is set from area_id if not given.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -376,9 +376,16 @@ sub inspect : Private {
...
 
     if ($c->cobrand->can('council_area_id')) {
-        my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+        my $priorities_by_category = FixMyStreet::App->model('DB::ResponsePriority')->by_categories($c->cobrand->council_area_id, $c->stash->{contacts});
         $c->stash->{priorities_by_category} = $priorities_by_category;
-        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories(
+            $c->cobrand->council_area_id,
+            $c->stash->{contacts},
+            # The cobrand may wish to provide its own list of bodies independent
```
dracos: Does this not apply the other way also? Why would London borough cobrands want to see TfL response templates?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -267,6 +267,26 @@ sub prefill_report_fields_for_inspector { 1 }
 
 sub social_auth_disabled { 1 }
 
+
+=head2 response_template_body_ids
```
dracos: This is okay, but looking at it like this - is there a reason why we use body areas at all, and not just the body (as this override does) everywhere?

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -376,9 +376,19 @@ sub inspect : Private {
...
+        );
         $c->stash->{priorities_by_category} = $priorities_by_category;
-        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories($c->cobrand->council_area_id, @{$c->stash->{contacts}});
+        my $templates_by_category = FixMyStreet::App->model('DB::ResponseTemplate')->by_categories(
+            $c->stash->{contacts},
+            area_id => $c->cobrand->council_area_id,
+            # The cobrand may wish to provide its own list of bodies independent
+            # of council_area_id - e.g. TfL covers many areas but we only want
+            # to show response templates directly owned by the TfL body.
+            body_ids => $c->cobrand->call_hook('response_template_body_ids')
```
dracos: There is a gotcha here, in that call_hook returns an empty list in a list context like here, so if there were any arguments after this one (e.g. it was `body_ids => $c->cobrand->call_hook(...), foo => 123`), then body_ids would be set to "foo" and "123" to undef. Adding `|| undef` to the end of the line would prevent.

davea: Good catch, thanks.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
davea: That's a good question. Might have been something to with districts for Oxfordshire? Either way, don't think it's in use any more. Have removed the area_ids reference in 1108af9070fe520f4f40f6e84e299fde8d200611 - does that look OK?

dracos: Yeah! No harm in leaving the area_id flow in, I guess, though happy to remove it as well (and ha, I guess doesn't need params any more if it's only ever taking a body ID argument)

perllib/FixMyStreet/Roles/ContactExtra.pm
```diff
@@ -25,8 +25,15 @@ sub for_bodies {
...
 sub by_categories {
-    my ($rs, $area_id, @contacts) = @_;
-    my %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $area_id });
+    my ($rs, $contacts, %params) = @_;
+
+    my %body_ids = ();
+    if ( $params{body_id} ) {
+        %body_ids = ( $params{body_id} => 1 );
+    } else {
+        %body_ids = map { $_->body_id => 1 } FixMyStreet::DB->resultset('BodyArea')->search({ area_id => $params{area_id} });
```
dracos: Yeah, you could drop this, and then the below could be simplified if you knew it was only ever one body. But meh.

PR #2716, opened 2019-11-04
Fix password reset on 2FA accounts again.

dracos:
Just after fixing it, 3d593bc68 broke it again, because it took anyone
who must have 2FA switched on to the must-have-2FA sign up flow, even
if they already had 2FA set up. [skip changelog]


dracos:
Approved in https://github.com/mysociety/fixmystreet/pull/2718 (where it has a test)

PR #2718, opened 2019-11-04
[TfL] Mandate 2FA for non-internal-IP staff users.

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1606
[skip changelog] Matching IP addresses given in server config.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -144,4 +144,18 @@ sub dashboard_export_problems_add_columns {
     };
 }
 
+sub must_have_2fa {
+    my ($self, $user) = @_;
+
+    require Net::Subnet;
+    my $ips = $self->feature('internal_ips');
+    my $is_internal_network = Net::Subnet::subnet_matcher(@$ips);
+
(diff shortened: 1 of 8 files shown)


perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -80,4 +80,17 @@ sub admin_allow_user {
...
+sub must_have_2fa {
+    my ($self, $user) = @_;
+
+    require Net::Subnet;
+    my $ips = $self->feature('internal_ips');
+    my $is_internal_network = Net::Subnet::subnet_matcher(@$ips);
+
+    my $ip = $self->{c}->req->address;
+    return 0 if $is_internal_network->($ip);
+    return 1 if $user->from_body && $user->from_body->name eq 'TfL';
```
davea: We should probably return 1 for superusers too.

dracos: I'd hope they'd all already logged into .com, but sure :)

dracos:
Adding tests showed this was not working at all. This PR now includes #2716 and changes on top to allow skipping of 2FA.

PR #2722, opened 2019-11-06
Admin auditing

dracos:
This adds audit logging to some things that were missing it, and then adds a view that shows you a user's audit log history, shortlist history, problems created, and updates created (with text). With a bit of refactoring.

Todo:
- [x] contributed_by!
- [x] tests
- [x] Create moderation entry if report/update edited in backend admin?
- [ ] Pagination if needed (probably)


dracos:
Added some more tests, spotted a couple of broken links, and changed base to master as nothing TfL specific here.

dracos:
Oh and added category change log event.

dracos:
Yes, I deliberately ignored priorities and states :) Have pushed a commit to hopefully log all those user things.

PR #2723, opened 2019-11-07
[TfL] Red route asset handling

davea:
- Ensures reports can only be made on red routes in specific categories.
 - Ensures a marker must be selected for bus stops/traffic lights layers
 - Only shows roadworks.org markers that are on red routes

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1588

[skip changelog]

### DIFF
--- a/templates/web/tfl/report/new/roads_message.html
+++ b/templates/web/tfl/report/new/roads_message.html
@@ -1,4 +1,9 @@
 <div id="js-roads-responsibility" class="box-warning hidden">
+    <div id="js-not-tfl-road" class="hidden js-responsibility-message">
+        <p>Transport for London does not maintain this road. Please use
+        <a href="https://www.fixmystreet.com/">FixMyStreet.com</a> to report it
+        to the appropriate borough.</p>
+    </div>
     <div id="js-not-an-asset" class="hidden js-responsibility-message">
         <p><strong>Select a marker</strong></p>
         <p>Please select <span class="js-roads-asset" data-original="an item">an item</span> from the map on which to make a report.</p>
--- a/templates/web/tfl/report/new/roads_message.html
+++ b/templates/web/tfl/report/new/roads_message.html
@@ -1,0 +1,6 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <div id="js-not-an-asset" class="hidden js-responsibility-message">
+        <p><strong>Select a marker</strong></p>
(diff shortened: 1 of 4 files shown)


web/cobrands/tfl/assets.js
```diff
@@ -23,7 +23,20 @@ var defaults = {
     body: "TfL"
 };
 
-fixmystreet.assets.add($.extend(true, {}, defaults, {
+var asset_defaults = $.extend(true, {}, defaults, {
+    select_action: true,
+    no_asset_msg_id: '#js-not-an-asset',
+    actions: {
+        asset_found: function() {
```
dracos: This can be `asset_found: fixmystreet.message_controller.asset_found,`

web/cobrands/tfl/assets.js
```diff
@@ -23,7 +23,20 @@ var defaults = {
...
+            fixmystreet.message_controller.asset_found();
+        },
+        asset_not_found: function() {
```
dracos: Annoying that you can't do the same here, I think you could if asset_not_found was using this instead of an argument and argument calling fixed up.

templates/web/tfl/report/new/roads_message.html
```diff
@@ -0,0 +1,6 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
```
dracos: Does/will this need adding to all other London borough cobrands, e.g. bexley? Are they showing bus stop/traffic light assets? Don't think they are at present.

web/cobrands/tfl/assets.js
```diff
@@ -64,6 +64,72 @@ fixmystreet.assets.add(asset_defaults, {
...
+        strokeColor: "#ff0000",
+        strokeOpacity: 0.6,
+        strokeWidth: 2
+    })
+});
+
+
+/* Reports in these categories can only be made on a red route */
+var tlrn_categories = [
+    "All out - three or more  out in a row",
```
dracos: Double space correct here?

templates/web/tfl/report/new/roads_message.html
```diff
@@ -1,4 +1,9 @@
 <div id="js-roads-responsibility" class="box-warning hidden">
+    <div id="js-not-tfl-road" class="hidden js-responsibility-message">
```
dracos: This will need adding to .com roads_message.html if you want it to work there, I think.
Though perhaps not necessary if these categories will only appear once a red route is clicked. But for now...

dracos:
Fixed some suggestions, and hopefully implemented rest in #2735

PR #2724, opened 2019-11-07
[TfL] Direct reports to email addresses on a per-borough basis

davea:
Adds a new cobrand feature that allows the destination email address
for a report to be determined on its MapIt area IDs, not just the
category it was made in.

Connects https://github.com/mysociety/fixmystreet-commercial/issues/1595

[skip changelog]


davea:
Yes, it's assuming `Email::TfL` isn't needed any longer. Guess I'll remove it in this PR too to make that clearer.

PR #2730, opened 2019-11-08
Store email addresses report sent to on the report.

dracos:
If email SendReport is used, this stores the email address(es) sent to on the report as a record. This could be worked out (in normal cases) from looking at the category details at the time the report was sent, but in complicated cases and ones that might change frequently, there doesn't seem any harm in freezing it at the point of sending. This could in future be displayed in admin, CSV export, etc.

### DIFF
--- a/perllib/FixMyStreet/SendReport/Email.pm
+++ b/perllib/FixMyStreet/SendReport/Email.pm
@@ -110,12 +110,19 @@ sub send {
         $params, $sender, $nomail, $cobrand, $row->lang);
 
     unless ($result) {
+        $row->set_extra_metadata('sent_to' => email_list($params->{To}));
         $self->success(1);
     } else {
         $self->error( 'Failed to send email' );
     }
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/SendReport/Email.pm
```diff
@@ -110,6 +110,7 @@ sub send {
         $params, $sender, $nomail, $cobrand, $row->lang);
 
     unless ($result) {
+        $row->set_extra_metadata('sent_to' => email_list($params->{To}));
```
struan: This will be overwritten if we re-send the message. Not sure how much of a concern that is. Given it's in theory for tracking who is currently responsible for it, possibly not a concern.

dracos: Yeah, I thought it best for now to match what happens to e.g. `whensent` and `send_method_used` which both only contain the lastest sending information.

PR #2735, opened 2019-11-11
[TfL] Red route handling for .com/cobrands

dracos:
This PR sets up the UK site and London cobrands so that clicking on a TfL red route will show all TfL categories and any council street cleaning category/group, and clicking elsewhere in Greater London will show all council categories and TfL any-location categories (bus stops, traffic lights). It centralises all TfL handling in one JS file which should handle all the above.

Category groups need to be activated on all London cobrands for this to function correctly.

The only loss I have thought of is Bromley's "Blocked drain" in Street cleaning will no longer go to TfL (they don't have such a category for that to work; we could drop it if Bromley do not want to receive them).

To fix the issue of Other being a stopper category for TfL, but we don't want that on non-red route on non-tfl cobrand, I've pushed code so that TfL would be called "Other (TfL") with a display_name of TfL, which I think then handles as you'd expect everywhere. Hopefully.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1668

### DIFF
--- a/bin/tfl/import_categories
+++ b/bin/tfl/import_categories
@@ -53,6 +53,7 @@ my $groups = $config->{groups};
 for my $group (keys %$groups) {
     my $cats = $groups->{$group};
     for my $cat (@$cats) {
+        $cat->{category} = 'Other (TfL)' if $cat->{category} eq 'Other';
         my $child_cat = FixMyStreet::DB->resultset("Contact")->find_or_new({
             body => $body,
             category => $cat->{category}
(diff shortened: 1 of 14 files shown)


davea:
Haven't had time for a full review yet but two bugs stood out (not sure if they're due to my set up though?)

1. An odd error when you try and report something in a TfL category not on a red route:

bug


2. It doesn't enforce picking an asset for the TfL asset categories:

bug2

dracos:
I can't replicate 1 so that's concerning - will have to find out more what your set up is. It shouldn't be showing you TfL-only categories when clicking not on a red route, unless the category is in both TfL and the council lists, in which case that works for me, see animation below ("Fallen Tree" is TfL only, "Joint category" is in both)

tflwestminjoint

dracos:
As for 2, the TfL staging site does not enforce asset selection, so...

dracos:
The asset selection bug was because the code was doing the following:
* Asset not selected => Show message about clicking an asset
* TfL road not clicked on, but not a TfL road only category => Hide message (ostensibly not-tfl-road message, but in reality any message including asset one)

I've swapped the order of the layers, so that it now does:
* TfL road not clicked on, but not a TfL road only category => Hide messages
* Asset not selected => Show message

I think that doesn't cause any counter-issues.

Issue #2736, opened 2019-11-13
RSS XSL


dracos:
It'd be good if the RSS feeds could have some (simple) XSL again, to explain things to people who click on them without understanding what RSS feeds are. Similar to what the BBC feeds do (https://www.bbc.co.uk/news/10628494) but they seem to be the only obvious ones I've found. We used to have this, but it had the original design and so was removed in 102840bc128085cc731f7283266c21e2f8f8e2f5 in 2012.

PR #2739, opened 2019-11-13
Add cobrand-specific XSL to RSS feeds.

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/2736 - uses the email templates to give cobrand specific branding if present in the emails, and also gives nice consistency between RSS and email alerts, which conceptually are doing the same thing anyway.

Before:
image

After:
image


zarino:
@dracos I’ve added a fixup commit with a few styling and content changes:

Screenshot_2019-11-15 New problems near BS34 7BF on FixMyStreet XML Feed

I’m seeing html-escaped images and "Report on FixMyStreet" links at the end of each report – which didn‘t show up in your screenshot. Is that just me?

PR #2741, opened 2019-11-14
[Tfl] Content changes

dracos:
Fixes all of https://github.com/mysociety/fixmystreet-commercial/issues/1679 except for https://github.com/mysociety/fixmystreet-commercial/issues/1680 which is a bug.

By the end it was heavy going so my final commit is separate because I know I probably should have done it a different way but hey.

### DIFF
--- a/templates/web/base/report/new/form_public_councils_text.html
+++ b/templates/web/base/report/new/form_public_councils_text.html
@@ -1,0 +1,11 @@
+[% BLOCK public_councils_text %]
+<h2 class="form-section-heading js-hide-if-private-category">[% loc( 'Public details' ) %]</h2>
+<h2 class="form-section-heading form-section-heading--private js-hide-if-public-category hidden-js">[% loc( 'Report details' ) %]</h2>
+<div class="form-section-description" id="js-councils_text">
+  [% IF js %]
+    [% PROCESS 'report/new/councils_text_all.html' list_of_names = [ loc('the local council') ] %]
+  [% ELSE %]
+    [% PROCESS 'report/new/councils_text.html' %]
+  [% END %]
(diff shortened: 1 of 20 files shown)


templates/web/tfl/footer.html
```diff
@@ -0,0 +1,20 @@
+                    [% IF pagefooter %]
+                    <footer role="contentinfo">
+                        [% INCLUDE 'front/footer-marketing.html' %]
```
davea: This file is blank, so no need to include it.

templates/web/tfl/report/new/councils_text_all.html
```diff
@@ -0,0 +1,16 @@
+<p>
+[% UNLESS non_public_categories.$category;
+
+    tprintf(
+        'Information you enter here will be sent to us and published for others to see, in accordance with our <a href="%s">privacy policy</a>.',
```
davea: I think this means the `fixmystreet.update_public_councils_text` override in `tfl/js.js` can be dropped now?

templates/web/base/report/new/form_public_councils_text.html
```diff
@@ -0,0 +1,11 @@
+[% IF c.cobrand.moniker != 'tfl' %]
```
davea: I thought it might be easier to explain by showing and not telling, so have pushed a suggestion (e6f78002a8521a34468e7c7667c4fd79a7df8f9a) which uses blocks to remove the `tfl` references from the base template.

dracos: Ta; I've removed one unneeded tfl bit of that and brought it in.

PR #2745, opened 2019-11-15
allow staff to submit in disabled categories

struan:
If a user belongs to a body override the disabled form check upon
submission as mostly staff users can submit in those categories.

Fixes mysociety/fixmystreet-freshdesk#98

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northamptonshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Northamptonshire.pm
@@ -137,4 +137,13 @@ sub report_validation {
     }
 }
 
+sub staff_ignore_form_disable_form {
+    my $self = shift;
+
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Northamptonshire.pm
```diff
@@ -137,4 +137,13 @@ sub report_validation {
     }
 }
 
+sub staff_ignore_form_disable_form {
+    my $self = shift;
+
+    my $c = $self->{c};
+
+    return $c->user_exists
+        && $c->user->belongs_to_body( $c->cobrand->body->id );
```
dracos: $c->cobrand can be $self here, though is the same thing.

PR #2749, opened 2019-11-18
[Cheshire East] Emergency message on report form

dracos:
Also gives change_location a bottom margin as I needed it here, needed it on TfL for their top message, and would improve the Bucks message too. [skip changelog]
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1686

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -496,7 +496,7 @@ small#or:after {
 
 .change_location {
     color: #666;
-    margin: 0;
+    margin: 0 0 1em;
 
     // No need for this element on mobile, where they can't click the map,
     // and there's already a "Report" button in the header that leads to
@@ -2117,10 +2117,8 @@ label .muted {
   display: inline-block;
 }
 
-#problem_form {
-  & > label:first-child {
+#form_category_row > label:first-child {
     margin-top: 0;
-  }
 }
 
 .box-warning,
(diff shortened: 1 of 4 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -496,7 +496,7 @@ small#or:after {
 
 .change_location {
     color: #666;
-    margin: 0;
+    margin: 0 0 1em;
```
zarino: This increases the space above the category picker – not by a _huge_ amount, but it’s more than it really needs:

Screen Shot 2019-11-19 at 16 36 01

If we’re going to give `.change_location` a margin-bottom, maybe we should remove the `margin-top` from `#form_category_label`?

Or, if you don’t mind the extra space, we could just leave it there.

Maybe worth setting a to-do item that we should really look at whether there’s a nicer way to do the vertical spacing in this form, all in one place, rather than handling all these individual elements in different ways.

dracos: Grr, I swear I checked this one, sigh. Yes, that's too much space, I thought the margins would collapse together but clearly something is preventing them doing so. Ideally fixing *that* would be my fix, given that's what margins are meant to do ;) But I guess won't be easy, so taking it off the top label might be easier...

zarino: I too would have expected the margins to collapse here. But I can’t see why they’re not. I _did_ notice, if I replace the `<fieldset>` with a `<div>` in the browser DevTools, the margins collapse. So maybe it’s something "special" about fieldset elements? (Perhaps related to the silly "inset" styling that `<legend>` elements get by default?)

dracos: Yeah, fieldset establishes a new block formatting context, so margins don't collapse.

dracos: I found a bit of CSS that I think was meant to be doing this already, and wasn't: `#problem_form > label:first-child { margin-top: 0; }` so I've changed that to `#form_category_row > label:first-child { margin-top: 0; }`

PR #2750, opened 2019-11-18
Issues/commercial/1490 occ inspector tool

struan:
Reverse hard coded disabling of inspector tool on OCC, and add a script to do this at a staff permission level.

Required for mysociety/fixmystreet-commercial#1490

[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/update_council_user_permissions
+++ b/bin/fixmystreet.com/update_council_user_permissions
@@ -1,0 +1,70 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all staff belonging to a body.
+
+use strict;
+use warnings;
+use v5.14;
+
+BEGIN {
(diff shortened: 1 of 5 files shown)


bin/oxfordshire/one-off-remove-inspector-perms
```diff
@@ -0,0 +1,51 @@
...
+my $commit;
+GetOptions(
+    'commit' => \$commit,
+);
+
+if (!$commit) {
+    say "*** DRY RUN ***";
+}
+
+my $body = FixMyStreet::DB->resultset("Body")->find({ name => 'Oxfordshire County Council' });
```
dracos: Could we make the name, and the list of permissions, command line arguments, then this is a slightly more useful generic tool? And even if time, make it add or delete, so it could also give every staff user at a body a particular permission? Anyway, at least the name/permissions would be good.

PR #2752, opened 2019-11-19
Make all links consistently underlined

wrightmartin:
This is a bit of a catch-all fix and I haven't been able to test it exhaustively. If we spot any links appearing underlined we can add them to the list in the stylesheet I guess

### DIFF
--- a/web/cobrands/tfl/base.scss
+++ b/web/cobrands/tfl/base.scss
@@ -188,12 +188,6 @@ ol.big-numbers>li {
     }
 }
 
-#front-recently {
-    a {
-        text-decoration: none;
-    }
-}
-
 .item-list--reports li > a,
 #key-tools a,
 .problem-back {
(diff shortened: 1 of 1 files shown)


web/cobrands/tfl/base.scss
```diff
@@ -183,6 +188,18 @@ ol.big-numbers>li {
     }
 }
 
+#front-recently {
```
dracos: I don't think this one is needed, the `.item-list--reports li > a` below should also catch it.

wrightmartin: Thanks @dracos, fixed

web/cobrands/tfl/base.scss
```diff
@@ -183,6 +188,12 @@ ol.big-numbers>li {
     }
 }
 
+.item-list--reports li > a,
+#key-tools a,
+.problem-back {
+    text-decoration: none;
```
dracos: Fine for now, I wonder if this should actually be in base so that it works for any cobrand that adds text-decoration.

PR #2755, opened 2019-11-20
[TfL] Add cookies banner and analytics JS

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/1688 and https://github.com/mysociety/fixmystreet-commercial/issues/1673.

[skip changelog]


dracos:
Dunno if this was ready for review, but I switched the order of include as requested on Basecamp, and added jQuery to all pages so it would function on e.g. front page. Looks good to me if nothing else remaining.

PR #2757, opened 2019-11-21
Add OS MasterMap map type

chrismytton:
Adds a MasterMap map type for the Peterborough cobrand which is used at lower zoom levels. Also switches Bexley to use that map type, rather than duplicate code.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1555

image


chrismytton:
Have now combined the shared code from Bexley and Peterborough into a single `MasterMap` map type.

PR #2758, opened 2019-11-21
[Bexley] Switch to new MasterMap URL

chrismytton:
We now have a generic mastermap layer which contains the MasterMap files for both Bexley and Peterborough (and potentially others in the future), so switch to using that.

Related to the parallel change for Peterborough https://github.com/mysociety/fixmystreet/pull/2757

## Notes to deployer

Once this has been deployed we can remove the old "bexley-mm" layer bits in the tilma httpd config.


chrismytton:
> do you want to combine Bexley.pm/bexley.js and Peterborough/etc from the other PR into a shared mastermap code?

Heh, this exact thought crossed my mind about 5 minutes after opening this PR 🙂 Will do, yep 👍

chrismytton:
Closing in favour of the combined version in https://github.com/mysociety/fixmystreet/pull/2757

PR #2759, opened 2019-11-22
Add configuration for setting CSP header.

dracos:
This allows you to output a working Content-Security-Policy header, with optional third-party domains, by setting a new CONTENT_SECURITY_POLICY configuration option.

In the UK, uses a cobrand feature to allow per-cobrand CSP setting.

### DIFF
--- a/conf/general.yml-example
+++ b/conf/general.yml-example
@@ -28,6 +28,10 @@ SECURE_PROXY_SSL_HEADER: ''
 # If you're behind a proxy, set this to a two-element list containing the
 # trusted HTTP header and the required value. For example:
 #   SECURE_PROXY_SSL_HEADER: [ 'X-Forwarded-Proto', 'https' ]
+CONTENT_SECURITY_POLICY: ''
+# To activate CSP, set this to 1. If you have additional third party domains to
+# allow JavaScript on, or want to specify extra script-src CSP directives, set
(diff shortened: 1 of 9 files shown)


conf/general.yml-example
```diff
@@ -28,6 +28,9 @@ SECURE_PROXY_SSL_HEADER: ''
 # If you're behind a proxy, set this to a two-element list containing the
 # trusted HTTP header and the required value. For example:
 #   SECURE_PROXY_SSL_HEADER: [ 'X-Forwarded-Proto', 'https' ]
+CONTENT_SECURITY_POLICY: ''
+# To activate CSP, set this to 1. If you have additional third party domains to
+# allow JavaScript on, set this to space-separated list of domains.
```
davea: It'd be nice to have this as a YAML list to be consistent with other settings e.g. `ALLOWED_COBRANDS`, `MAPIT_TYPES`, etc, unless there's another use case for being able to include arbitrary strings in the header?

dracos: You could theoretically use it if you were happy using e.g. new CSP3 headers. But no reason it couldn't be a list then as well, they're all space separated items anyway. But I thought a string might make it clearer that it can be more than just domains, just I imagined that would be the common case. The content of this string inserts itself into the `script-src` part of the CSP. I'll try and update the help text.

PR #2760, opened 2019-11-22
[Peterborough] Include email categories on cobrand

chrismytton:
They've requested that these be added to the cobrand.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1667


chrismytton:
Thanks for picking this up @struan.

PR #2761, opened 2019-11-22
[TfL] Display more details when roadworks marker is selected

davea:
Shows details about selected roadworks asset in the sidebar:

image

For https://github.com/mysociety/fixmystreet-commercial/issues/1605

[skip changelog]


web/cobrands/tfl/assets.js
```diff
@@ -266,8 +266,20 @@ var rw_stylemap = new OpenLayers.StyleMap({
     })
 });
 
+function format_date(date) {
```
dracos: Make the one in roadworks.js a function that can be called externally?

web/cobrands/tfl/assets.js
```diff
@@ -312,7 +311,51 @@ fixmystreet.assets.add(fixmystreet.roadworks.layer_future, {
...
+            feature.layer = this;
+            var attr = feature.attributes,
+            tooltip = attr.tooltip.replace(/\\n/g, '\n'),
+            desc = attr.works_desc.replace(/\\n/g, '\n');
+
+            var $msg = $('<div class="js-roadworks-message js-roadworks-message-' + this.id + ' box-warning"></div>');
+            var $dl = $("<dl></dl>").appendTo($msg);
+            $dl.append("<dt>Location</dt>");
+            var $summary = $("<dd></dd>").appendTo($dl);
+            tooltip.split("\n").forEach(function(para) {
```
dracos: Probably not worth it, but feels quite a bit of this could be shared with the roadworks.js in some manner.

davea: Agreed on both counts. I did think about factoring it together somehow with more config options but decided not to bother.

web/cobrands/tfl/assets.js
davea: Good idea, will do.

PR #2762, opened 2019-11-22
Banes road nearby

struan:
[skip changelog]


templates/web/bathnes/report/new/roads_message.html
```diff
@@ -0,0 +1,7 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <strong>Not maintained by Bath &amp; North East Somerset Council</strong>
+    <div id="js-not-a-road" class="hidden js-responsibility-message">
+        <p>The location you have selected doesn't appear to be on <span class="js-roads-asset" data-original="a road">a road</span>.</p>
```
davea: This might be better as "...a road that Bath & North East Somerset Council maintain" in case the user clicks something that is definitely a road/pavement that just happens to be outside the council's responsibility.

PR #2765, opened 2019-11-25
Guard against XSS from malicious admin user.

dracos:
The below breaks down into three things:
* Various templates assumed a problem's state would be safe, as it is only set by an admin (the front end state change is whitelisted);
* Various templates assumed a problem's category would be safe, for the same reason;
* In a couple of templates, you could submit a malicious payload to yourself.


dracos:
This would be superseded/unnecessary if https://github.com/mysociety/fixmystreet/pull/2772 lands.

PR #2766, opened 2019-11-25
add bodies_id to stash for around and report/new

struan:
This resolved the problem of checking user body permissions on both
around and report/new by adding a variable to stash specifically for
this purpose. Previously staff users could not see the mark private tick
if the report form was loaded via javascript from the around page as
there was no bodies variable in the stash.


struan:
Merged in b36f62bdaf4ca3775c2d6efc4974f98dbd105850

PR #2771, opened 2019-11-26
[Northamptonshire] update Vegetation asset layers

struan:
Allow all categories to make a report on roads or tree assets.
[skip changelog]


struan:
Oh, no, it should still be. Thanks

dracos:
Oh, I approved, but I note the Northants Cypress test is now failing.

Issue #2702, opened 2019-10-28
Group categories in dropdown on map & inspector UI


davea:
Showing of category groups in category dropdown:
- [ ] Map filter
- [x] Inspector form
- [x] Dashboard
- [x] Admin selecting categories for user/templates
- [x] Admin report edit
- [x] Admin body page, list of categories

For cobrands with a lot of grouped categories, showing all categories in a flat list in the dropdown on the map filters isn't very helpful:

![Report Snapshot](https://user-images.githubusercontent.com/4776/67674214-d022af00-f973-11e9-90c9-9b7b9278692f.JPG)

And it's the same on the inspector tool when changing category:

<img width="776" alt="image" src="https://user-images.githubusercontent.com/4776/67674231-dd3f9e00-f973-11e9-833f-b3dac1c01e82.png">

At the very least it would be good to use some `<optgroup>` hierarchy like we do for non-JS on the new report form:

<img width="541" alt="image" src="https://user-images.githubusercontent.com/4776/67674328-1d9f1c00-f974-11e9-8cab-5fc9348aa878.png">

dracos:
Also dashboard category dropdown, admin selecting categories for user/templates, admin report edit.

dracos:
Chat about the map filter one in Slack: https://mysociety.slack.com/archives/C0260Q88Y/p1572862477014100

davea:
PR for map filter one at https://github.com/mysociety/jquery-multi-select/pull/20

PR #2773, opened 2019-11-27
Group categories on map filters & admin

davea:
Adds category groups to the user and template editing forms in the admin:

image

For https://github.com/mysociety/fixmystreet/issues/2702

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Users.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Users.pm
@@ -394,8 +394,10 @@ sub edit : Chained('user') : PathPart('') : Args(0) {
             id => $_->id,
             category => $_->category,
             active => $active_contacts{$_->id},
+            group => $_->get_extra_metadata('group') // '',
         } } @live_contacts;
         $c->stash->{contacts} = \@all_contacts;
+        $c->forward('/report/stash_category_groups', [ \@all_contacts, 1 ]) if $c->cobrand->enable_category_groups;
     }
 
     # this goes after in case we've delete any alerts
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -661,6 +661,33 @@ sub check_has_permission_to : Private {
     return \%permissions;
 };
 
+
+sub stash_category_groups : Private {
+    my ( $self, $c, $contacts, $combine_multiple ) = @_;
+
+    my %category_groups = ();
+    for my $category (@$contacts) {
+        my $group = $category->{group} // $category->get_extra_metadata('group') // [''];
+        # this could be an array ref or a string
+        my @groups = ref $group eq 'ARRAY' ? @$group : ($group);
+        if (scalar @groups > 1 && $combine_multiple) {
+            @groups = sort @groups;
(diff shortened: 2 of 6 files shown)


perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -394,8 +394,10 @@ sub edit : Chained('user') : PathPart('') : Args(0) {
             id => $_->id,
             category => $_->category,
             active => $active_contacts{$_->id},
+            group => $_->{group} // $_->get_extra_metadata('group') // [''],
```
dracos: I think `$_->get_extra_metadata('group')` will do here (and in Templates.pm the same), as the loop in stash_category_groups sorts out the default for you, and it's always an object here.

templates/web/base/admin/category-checkboxes.html
```diff
@@ -22,4 +14,23 @@
...
+      <p>[% hint %]</p>
+    </div>
+  [% END %]
+    [% loc('Categories:') %]
+  </legend>
+
+  [% IF category_groups %]
+    [% FOR group IN category_groups %]
+      <h3>[% group.name | html %]</h3>
+      [% INCLUDE checkboxes contacts=group.categories %]
```
dracos: If a category is in multiple groups it will now be listed under all of them. Presumably if you want to deselect that means you'd have to untick it from all of them? I realise this won't be that likely, but only listing them under one group and have some extra text saying "also in a, b, c" might be better.

davea: Good idea. I've put them in a "Multiple Groups" pseudogroup at the bottom of the list with a note as you suggest.

templates/web/base/admin/category-checkboxes.html
```diff
@@ -18,8 +10,31 @@
...
+    <div class="admin-hint">
+      <p>[% hint %]</p>
+    </div>
+  [% END %]
+    [% loc('Categories:') %]
+  </legend>
+
+  [% IF category_groups %]
+    [% FOR group IN category_groups %]
+      <h3>[% group.name | html %]</h3>
```
dracos: If there are categories without a group, then these will be listed under a blank heading here; perhaps default to "No Group" in that case?

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -661,6 +661,33 @@ sub check_has_permission_to : Private {
...
+
+sub stash_category_groups : Private {
+    my ( $self, $c, $contacts ) = @_;
+
+    my %category_groups = ();
+    for my $category (@$contacts) {
+        my $group = $category->{group} // $category->get_extra_metadata('group') // [''];
+        # this could be an array ref or a string
+        my @groups = ref $group eq 'ARRAY' ? @$group : ($group);
+        if (scalar @groups < 2) {
```
dracos: This now breaks the display when reporting a problem - because there you do want the existing behaviour of adding it to all the groups, you don't want a "Multiple Groups" option then. I think it has to be a flag passed to this for the two different behaviour.

davea: Oops, of course. We'll get there in the end :)

PR #2780, opened 2019-12-02
Enable category groups on map filter dropdown

davea:
Adds category groups on the map filter UI:
image

Fixes https://github.com/mysociety/fixmystreet/issues/2702


t/app/controller/reports.t
```diff
@@ -18,6 +18,9 @@ my $body_west_id = $mech->create_body_ok(2504, 'Westminster City Council')->id;
 my $body_fife_id = $mech->create_body_ok(2649, 'Fife Council')->id;
 my $body_slash_id = $mech->create_body_ok(10000, 'Electricity/Gas Council')->id;
 
+my $pothole = $mech->create_contact_ok(body_id => $body_edin_id, category => 'Potholes', email => 'potholes@example.org');
+my $graffiti = $mech->create_contact_ok(body_id => $body_west_id, category => 'Graffiti', email => 'graffiti@example.org');
+my $flytipping = $mech->create_contact_ok(body_id => $body_fife_id, category => 'Flytipping', email => 'flytipping@example.org');
```
struan: I'm not sure I understand what adding these lines is doing?

t/app/controller/reports.t
```diff
@@ -128,6 +131,7 @@ $mech->content_contains('2,3,4,4');
 FixMyStreet::override_config {
     ALLOWED_COBRANDS => 'fixmystreet',
     MAPIT_URL => 'http://mapit.uk/',
+    COBRAND_FEATURES => { category_groups => { fixmystreet => 1 } },
```
struan: likewise, why add this?

t/app/controller/around.t
```diff
@@ -232,7 +232,11 @@ subtest 'check category, status and extra filtering works on /around' => sub {
 
     # Create one open and one fixed report in each category
     foreach my $category ( @$categories ) {
-        $mech->create_contact_ok( category => $category, body_id => $body->id, email => "$category\@example.org" );
+        my $contact = $mech->create_contact_ok( category => $category, body_id => $body->id, email => "$category\@example.org" );
+        if ($category ne 'Pothole') {
+            $contact->set_extra_metadata(group => ['Environment']);
+            $contact->update;
+        }
```
struan: I don't understand what the changes in this file are for.

t/app/controller/reports.t
dracos: Nothing I think, actually, but it seemed odd to be creating reports just underneath that didn't have any matching categories existing for them. I can take them back out.

t/app/controller/around.t
dracos: This is so the fetched around page is loading the filter with category groups in it. I should perhaps have checked for that being in the output!

PR #2784, opened 2019-12-05
Staging banner on web and email

dracos:
image
image
image


web/cobrands/sass/_base.scss
```diff
@@ -659,6 +659,18 @@ ul.error {
...
+.dev-site-notice {
+    font-size: 1.5em;
+    color: #fff;
+    background-color: #f00;
+    position: absolute;
+    padding: 1em 5em;
+    top: 1em;
+    left: -5em;
+    transform: rotate(-45deg);
+    z-index: 2;
```
davea: Could stick a `pointer-events: none` to allow clicking through to content underneath maybe?

zarino: Yeah, I almost suggested this myself. Happily, unlike old-school corner badges (where the whole bounding square was rendered unclickable) this banner will only block clicks behind the banner itself. But maybe for a few extra characters, it’s worth adding the `pointer-events` rule in? Seems unlikely that anyone will ever _actually_ want to click on the banner itself.

PR #2790, opened 2019-12-06
Enable custom do-not-reply address & bounce messages for cobrands

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/1689

### DIFF
--- a/bin/handlemail
+++ b/bin/handlemail
@@ -22,6 +22,7 @@ BEGIN {
 use Getopt::Long;
 use Path::Tiny;
 use FixMyStreet;
+use FixMyStreet::Cobrand;
 use FixMyStreet::DB;
 use FixMyStreet::Email;
 use mySociety::HandleMail;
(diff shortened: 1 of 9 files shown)


bin/handlemail
```diff
@@ -180,20 +181,43 @@ sub handle_non_bounce_to_null_address {
...
+# use a cobrand's own reply-autoresponse template.
+sub get_config_for_autoresponse {
+    # cobrand might have been set from command line, so prefer that if so.
+    if ( defined $cobrand ) {
+        return ( $cobrand, FixMyStreet->config('CONTACT_EMAIL'), FixMyStreet->config('CONTACT_NAME') );
+    }
+
+    # Try and find a matching email address in the COBRAND_FEATURES config
+    my $recipient = mySociety::HandleMail::get_bounce_recipient($data{message})->address;
+    my $cobrands = FixMyStreet->config('COBRAND_FEATURES')->{do_not_reply_email} || {};
```
dracos: This will die with an error if COBRAND_FEATURES doesn't exist (and so returns undef), would be good to be careful given it's a script run on email input.

davea:
Please give it a go on staging, yes! I'll fix those bits now...

PR #2793, opened 2019-12-10
[TfL] Server-side red route lookup

davea:
The app can't distinguish between reports being made on red routes or not, so this PR makes the determination server side and returns the appropriate categories in the `/report/new/ajax` call. This fixes the app and also updates the website to use the same method.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1716

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UKCouncils.pm
+++ b/perllib/FixMyStreet/Cobrand/UKCouncils.pm
@@ -289,6 +289,18 @@ sub prefill_report_fields_for_inspector { 1 }
 
 sub social_auth_disabled { 1 }
 
+sub munge_report_new_category_list {
+    my ($self, $options, $contacts, $extras) = @_;
+
+    my %bodies = map { $_->body->name => $_->body } @$contacts;
+    if ( $bodies{'TfL'} ) {
+        # Presented categories vary if we're on/off a red route
+        my $tfl = FixMyStreet::Cobrand->get_class_for_moniker( 'tfl' )->new({ c => $self->{c} });
+        $tfl->munge_red_route_categories($options, $contacts);
+    }
+}
+
+
 =head2 lookup_site_code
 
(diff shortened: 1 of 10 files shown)


web/cobrands/tfl/assets.js
```diff
@@ -109,75 +50,10 @@ var red_routes_layer = fixmystreet.assets.add(defaults, {
...
-    road: true,
     non_interactive: true,
     always_visible: true,
     all_categories: true,
-    nearest_radius: 0.1,
     stylemap: tlrn_stylemap,
-    no_asset_msg_id: '#js-not-tfl-road',
-    actions: {
-        found: function(layer, feature) {
-            fixmystreet.message_controller.road_found(layer, feature);
```
dracos: As discussed, I think the first bit of found/not_found needs to run on the tfl cobrand alone, to keep its current behaviour.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -289,6 +289,19 @@ sub prefill_report_fields_for_inspector { 1 }
 
 sub social_auth_disabled { 1 }
 
+sub munge_report_new_category_list {
+    my ($self, $options, $contacts, $extras) = @_;
+
+    my %bodies = map { $_->body->name => $_->body } @$contacts;
+    if ( $bodies{'TfL'} ) {
```
dracos: I guess you don't want to run this on TfL cobrand itself.

dracos:
Pushed a fix for one test (needed TfL cobrand to be present) and other (TfL cobrand always returns all TfL categories), and then (untested) a change to not_found to only run on TfL cobrand and spot tlrn categories again.

davea:
Thanks for the fixups. They did the trick so I've merged this 👍

PR #2802, opened 2019-12-13
Stored ‘contributed_by’ when staff users make anonymous reports via button

davea:
[skip changelog]


davea:
> Perhaps contributing_as_anonymous_user should always store contributed_by if a staff user?

Yes, makes sense I think. Done in c8d069f72021052cad343a9ad222f6e606cd6458.

davea:
That's the intended/expected behaviour, yes. It means we can tell the difference between anonymous report made by the button vs the "report as" dropdown, though not sure if there's actually any value in knowing that yet :)

PR #2803, opened 2019-12-13
[Greenwich] [TfL] Fix TfL category display, and ignore groups

dracos:
This stops a category that should only be displayed on the TfL cobrand from displaying elsewhere (including Greenwich), and ignores groups from Greenwich as we don't want to use them but we do want to enable category groups for the TfL categories.
[skip changelog]


dracos:
Merged as fbd8ab730 and ea9d9fc2a

PR #2806, opened 2019-12-17
Show GPS marker on map when frontpage geolocation link used

### LINKED ISSUE #2359: Add 'you are here' pin to map when using 'locate me automatically'
Would it be possible to add a 'you are here' pin to the map once someone has clicked 'locate me automatically'? Appreciate this will be the very centre, but I think it would help people make sense of the map.

davea:
When the user clicks the "...or use my location" link to get to the map, their current location is shown on the map:

IMG_1129

The radius of the semi-transparent circle under the marker corresponds to the accuracy of the current location.

image

I opted to put this beneath all the pins on the map, though this does mean in busy locations it can be hard to see:

image

However putting the marker on top looks a little odd and more importantly seems to break the ability to click the existing pins:

image

Fixes #2359.


dracos:
Please compile OpenLayers from https://github.com/mysociety/ol2 otherwise it will break under Content-Security-Policy.

web/js/map-OpenLayers.js
```diff
@@ -375,6 +375,66 @@ $.extend(fixmystreet.utils, {
...
+            layer = new OpenLayers.Layer.Vector('Geolocation');
+            fixmystreet.map.addLayer(layer);
+            layer.setZIndex(fixmystreet.map.getLayersByName("Pins")[0].getZIndex() - 1);
+          }
+
+          function updateGeolocationMarker(e) {
+              if (!layer) {
+                  addGeolocationLayer();
+              }
+              layer.removeAllFeatures();
```
dracos: Could we move the features instead of recreating them every time there's a location update?

zarino: Just to flog a dead horse just a little more, if the blue dot is permanent, rather than constantly being recreated, then that would open the door to a putting a CSS "pulse" animation onto it…

davea:
I'd forgotten about our OpenLayers fork. Will recompile with that. And yes, I'll have a look at moving instead of recreating the features.

PR #2808, opened 2019-12-18
[TfL] Front page text change

dracos:
[skip changelog]


templates/web/tfl/around/_postcode_form_examples.html
```diff
@@ -0,0 +1,3 @@
+<p class="form-hint" id="pc-hint">
+[% tprintf(loc('e.g. ‘%s’ or ‘%s’'), c.cobrand.example_places) %]
```
davea: May as well bin the `loc` call while you're here?

PR #2810, opened 2019-12-18
Per-report OpenGraph images

### LINKED ISSUE #2394: Allow per-report OpenGraph images
Continuing from #613 - that switched to a new default image for all pages, but we think it's desirable to allow report pages to use the report image, if provided, and the default if not.

gbp:
Fixes #2394:

Please check the following:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
         - Set report title autocomplete to off to prevent email autocompleting
         - Add map filter debouncing to reduce server requests. #2675
         - Add XSL to RSS feeds so they look nicer in browsers.
+        - Add per-report OpenGraph images. #2394
         - Display GPS marker on /around map. #2359
     - Admin improvements:
         - Add new roles system, to group permissions and apply to users. #2483
--- a/perllib/FixMyStreet/App/Controller/Photo.pm
+++ b/perllib/FixMyStreet/App/Controller/Photo.pm
@@ -43,7 +43,7 @@ sub during :LocalRegex('^(temp|fulltemp)\.([0-9a-f]{40}\.(?:jpeg|png|gif|tiff))$
     $c->forward( 'output', [ $photo ] );
 }
 
-sub index :LocalRegex('^(c/)?([1-9]\d*)(?:\.(\d+))?(?:\.(full|tn|fp))?\.(?:jpeg|png|gif|tiff)$') {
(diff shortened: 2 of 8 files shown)


gbp:
@dracos not really sure where to start with implement tests for this as there doesn't seem to be much coverage around the image generation yet. Could you confirm if the cobrands which override `header_opengraph_image.html` should also have this feature?

dracos:
"not really sure where to start with implement tests for this as there doesn't seem to be much coverage around the image generation yet." - hmm, yeah, think the tests assume IM isn't present; the shrink/resize just return themselves, which I guess is better than nothing; you could do something similar, perhaps, with the sample image the tests use, check that at least gets output as you'd expect, that'd probably do.

"Could you confirm if the cobrands which override header_opengraph_image.html should also have this feature?" - yeah, might as well be consistent, they can always switch it away if they decide they don't want it.

gbp:
> the shrink/resize just return themselves, which I guess is better than nothing; you could do something similar, perhaps, with the sample image the tests use, check that at least gets output as you'd expect, that'd probably do.

I attempted to:
```diff
diff --git a/t/app/model/photoset.t b/t/app/model/photoset.t
index 29a28d232..88a902395 100644
--- a/t/app/model/photoset.t
+++ b/t/app/model/photoset.t
@@ -96,5 +96,17 @@ subtest 'Correct storage backends are instantiated' => sub {

 };

+subtest 'Photoset get Opengraph image data' => sub {
+    my $photo = path('t/app/controller/sample.jpg')->slurp_raw;
+    my $photoset = FixMyStreet::App::Model::PhotoSet->new({
+        data_items => [ $photo ],
+    });
+
+    my $data = $photoset->get_image_data( num => 0, size => 'og' );
+
+    is $data->{width}, 1200, 'width is 1200';
+    is $data->{height}, 630, 'height is 630';
+};
+

 done_testing();
```

But this isn't working due to [ImageMagick being disabed](https://github.com/mysociety/fixmystreet/blob/master/perllib/FixMyStreet/ImageMagick.pm#L6) in test mode. Not really sure how/where else I should be testing this.

dracos:
I think it'd be fine if it tests the basics by fetching a report page, and checking that e.g. the right image URL is being given in the opengraph line. The only alternative would be to use e.g. Image::Size to read the width/height instead (that's used at the point of image upload), which doesn't feel worth it.

davea:
I've added a couple of simple tests as suggested. Could you give this a review please @dracos?

Issue #2221, opened 2018-08-28
Provide offline FixMyStreet web use


dracos:
Requires #2220.

* While making a report/update, a loss of network when submitting a report could be caught and the report saved as draft, to be submitted when network returns
* If there's no network at an earlier/any point, it could display a different form that used geolocation, let you take a photo, and saved that as a draft for submission when there is signal;
* Other?

lizettal:
@dracos as offline reporting functionality is in place, is there anything left on this ticket to do or can it be closed?

PR #2812, opened 2019-12-19
Progressive Web App basics

dracos:
This PR:
* Adds a [manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) with defaults, but with the potential to read from database config/cobrand-specific directory of icons;
* Replaces the appcache behaviour with a [service worker](https://serviceworke.rs/) that performs the same task - staff users with the planned reports permission can cache reports offline and submit the inspector form when offline. Other users will have a custom offline page, and some static assets are cached, but currently no other functionality. Of course this can then be expanded e.g. #2221.

With these, FixMyStreet should be installable as a simple PWA.


perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -10,19 +14,69 @@ FixMyStreet::App::Controller::Offline - Catalyst Controller
...
 =head1 DESCRIPTION
 
-Offline pages Catalyst Controller.
+Offline pages Catalyst Controller - service worker and appcache.
 
 =head1 METHODS
 
 =cut
 
+sub manifest: Path("/.well-known/manifest.webmanifest") {
```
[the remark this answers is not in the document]
dracos: whoops, missed the link rel, didn't I!
It used to be in the spec, but got taken out: https://github.com/w3c/manifest/pull/275 - but I still quite liked it as a location. Guess it doesn't really make any difference.

web/cobrands/fixmystreet/offline.js
```diff
@@ -289,37 +261,29 @@ fixmystreet.offline = (function() {
     }
 
     // Remove a list of reports from the offline cache
-    function removeReports(urls) {
+    async function removeReports(urls) {
         var pathsRemoved = [];
```
davea: Not part of this PR, admittedly, but this variable doesn't seem to be used anywhere.

dracos:
Since the approval above, changes:
* I force-pushed a change to only the last commit to remove the async/await syntax, as I didn't want to deal with syntax errors in old browsers if councils using them, shouldn't have any functional change, not that complex anyway;
* New commit to move from localstorage to indexeddb - the latter is async and available in a SW, and so...
* New commit to then move the "capture POST on the inspect form" from the front end to the SW which is clearer and more extendable and should work even if you e.g. fetched the page online but then went offline whilst filling the inspect form in.

templates/web/base/offline/service_worker.html
```diff
@@ -0,0 +1,103 @@
...
+  // Handle inspection form submission if offline...
+  if (request.method === 'POST' && RegExp('/report/\\d+$').test(url)) {
+    fetchEvent.respondWith(async function() {
+      const fetchPromise = fetch(request.clone());
+      try {
+        return await fetchPromise;
+      }
+      catch {
+        fetchEvent.waitUntil(async function() {
+          var formData = await request.formData();
```
davea: This is where iOS/macOS Safari is getting hung up, as far as I can tell. Execution never proceeds past this line. No errors though. Here's the output (of aa4b7aec3fd9a90de794da59918f9306b56c1520) when submitting an inspector form when offline:

image

dracos: Appears that Safari does not support FormData in workers. Have switched to URLQueryParams for now, as this is a non-file form, but will be bit harder if/when we come to process POSTs with attachments...

PR #2821, opened 2020-01-07
Hide ‘provide extra information’ preamble when no visible fields are present

### LINKED ISSUE #2811: Remove heading from extra details form when only notice shown
When a category's extra data is only a single notice the "provide extra detail" heading is a bit out of place as there are no extra details being requested:

image

davea:
For categories that use the extra metadata fields to just show a notice
message to the user, the ‘help the council fix the problem by providing
extra information’ text was a bit out of place. This commit hides it
unless there are actual UI elements also included in the markup.

Fixes #2811.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -697,6 +697,8 @@ sub setup_categories_and_bodies : Private {
     my %category_extras  = ();       # extra fields to fill in for open311
     my %category_extras_hidden =
       (); # whether all of a category's fields are hidden
+    my %category_extras_notices =
+      (); # whether all of a category's fields are simple notices and not inputs
     my %non_public_categories =
       ();    # categories for which the reports are not public
     $c->stash->{unresponsive} = {};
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -697,6 +697,8 @@ sub setup_categories_and_bodies : Private {
     my %category_extras  = ();       # extra fields to fill in for open311
     my %category_extras_hidden =
       (); # whether all of a category's fields are hidden
+    my %category_extras_notices =
+      (); # whether all of a category's fields are simple notices and not inputs
```
struan: This is a weird formatting convention we've adopted :)

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -729,6 +731,16 @@ sub setup_categories_and_bodies : Private {
...
                 $category_extras_hidden{$contact->category} = $all_hidden;
             }
+
+            my $all_notices = (grep {
+                ( $_->{variable} || '' ) ne 'false'
+                && !$c->cobrand->category_extra_hidden($_)
+            } @$metas) ? 0 : 1;
+            if (exists($category_extras_notices{$contact->category})) {
+                $category_extras_notices{$contact->category} &&= $all_notices;
+            } else {
```
struan: I'm not sure I understand how you would end up with this already existing?

davea: (Meant to reply earlier, oops) I think it's possible in the situation where two overlapping bodies have the same category name, e.g. district/county or TfL/London boroughs.

PR #2822, opened 2020-01-08
[Bexley] Resend report on most category changes

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1721
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -382,7 +382,7 @@ sub edit_category : Private {
     my ($self, $c, $problem, $no_comment) = @_;
 
     if ((my $category = $c->get_param('category')) ne $problem->category) {
-        my $force_resend = $c->cobrand->call_hook('category_change_force_resend');
+        my $force_resend = $c->cobrand->call_hook('category_change_force_resend', $problem->category, $category);
         my $disable_resend = $c->cobrand->call_hook('disable_resend');
         my $category_old = $problem->category;
         $problem->category($category);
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -255,7 +255,7 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         $c->detach('edit_display') if $done;
     }
 
-    if ( $c->get_param('resend') && !$c->cobrand->call_hook('disable_resend') ) {
+    if ( $c->get_param('resend') && !$c->cobrand->call_hook('disable_resend_button') ) {
```
chrismytton: There's another reference to the `disable_resend` hook in the `edit_category` method in this file, does that need to be changed as well?

https://github.com/mysociety/fixmystreet/blob/bbeafd861b32bf9009d5961053f75e293375ae35/perllib/FixMyStreet/App/Controller/Admin/Reports.pm#L386

dracos: Yeah, so before this commit there was one `disable_resend` hook that hid the Resend button in the admin, plus also made sure no category change would lead to a report being resent, even if th body or send method changed. After this commit, there are now two hooks, `disable_resend` and `disable_resend_button` - the button one just hides the button in the admin, the `disable_resend` one (not now used by anything at present) still prevents a category change ever resending a report. Thought I might as well leave that in case was useful, but could also remove it as not currently used.

chrismytton: Ah I see, that makes sense. Think it's fine to leave it in, seems like something that could well come up in the future again.

PR #2824, opened 2020-01-08
Reduce duplicate Permalink.updateLink calls when zooming map

davea:
fms.com has so many layers now that zooming events were causing enough quick-fire history.replaceState calls (in updateLink) that iOS Safari was raising `SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds` errors, as well as being throttled on Chrome on Android. Zooming in causes lots of layers to come into range simultaneously, and each was triggering the 'changelayer' event. We don't include the layers state string in the permalink URL, so there's no need to listen to those events.

[skip changelog]


web/js/map-OpenLayers.js
```diff
@@ -1093,17 +1093,36 @@ OpenLayers.Control.PermalinkFMS = OpenLayers.Class(OpenLayers.Control.Permalink,
...
+        OpenLayers.Control.Permalink.prototype.initialize.apply(this, arguments);
+
+        // Zoom events on iOS/Android generate a huge number of calls to
+        // updateLink each of which calls replaceState. On iOS calling
+        // replaceState too often soon results in subsequent calls
+        // raising a "SecurityError: Attempt to use history.replaceState()
+        // more than 100 times per 30 seconds" exception, which results
+        // in the map breaking.
+        // Debouncing the call to _updateLink works around this problem.
+        // A better solution would be to stop so many updateLink calls
```
dracos: Turns out this is because zooming in is causing a lot more layers to be in range so they each fire a changelayer event which the permalink is listening for. It doesn't need to listen for those events (or the changebaselayer either), which I think should fix this.

davea:
Seems like a good idea in principle, yep. Something about the new class is losing any extra parameters passed to `/around` though, specifically `geolocate=1` which is breaking that recently added geolocation marker. Had a quick look but couldn't see exactly where it was going wrong.

dracos:
Ah yes, I was stripping off the existing query parameters before constructing the new parameters, rather than after. Hopefully fixed now.

PR #2826, opened 2020-01-08
[UK] do not show unknown banner on fms pro sites

struan:
In theory these are never unknown so do not display this in report
banner.

[skip changelog]

### DIFF
--- a/templates/web/fixmystreet-uk-councils/report/banner.html
+++ b/templates/web/fixmystreet-uk-councils/report/banner.html
@@ -17,12 +17,6 @@
     [% INCLUDE banner, id = 'closed', text = loc('Closed') %]
 [% END %]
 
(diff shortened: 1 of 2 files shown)


templates/web/fixmystreet-uk-councils/report/banner.html
```diff
@@ -0,0 +1,26 @@
...
+
+[% IF problem.is_fixed %]
+    [% INCLUDE banner, id = 'fixed', text = loc('Fixed') %]
+[% END %]
+
+[% IF problem.is_closed %]
+    [% INCLUDE banner, id = 'closed', text = loc('Closed') %]
+[% END %]
+
+[% IF NOT problem.to_body_named('Bromley') %]
```
davea: Bromley has its own `banner.html`, so don't think this line is required.

Issue #2376, opened 2019-01-11
Minor issue in process-inactive-accounts


joesiltberg:
For FixaMinGata, we're running process-inactive-accounts and process-inactive-reports nightly. We're expecting to anonymize the first real users during spring so I did some testing to make sure it works as expected. 

It seems to work fine! 

I noticed one minor thing which perhaps isn't worth fixing, but perhaps it'll be a problem when more users are anonymized.

**Describe the bug**
process-inactive-accounts seems to anonymize already anonymized users. Not sure if this is a problem, but if you're running the script with --verbose to see which users were anonymized you'll eventually get a very long list instead of just the ones that were anonymized this time.

Not sure if we're actually doing any db writes due to these since we'd be writing the same values, but if we do perhaps it will be a performance issue once there are many old accounts.

**To Reproduce**
1. Make sure there's at least one user where last_active is old enough
2. Run the script with high verbosity as the fms user, for instance: 
```/var/www/beta.fixamingata.se/fixmystreet/bin/process-inactive-accounts --verbose --anonymize 12 --email 11```
3. Note that some users are anonymized
4. Do step 2 again
5. Note that the same users are anonymized again

One idea is to check for empty passwords to avoid re-anonymization. But perhaps there could be accounts with empty passwords which would still need to be anonymized? Checking the anonymized email seems a bit ugly...

dracos:
Yes, you're right, it is anonymizing already anonymized users. You can't check for empty passwords as that would include anyone who never set a password, I think. Sadly, I think checking the email might be the only way; bit ugly as you say, but should work okay nonetheless.

PR #2827, opened 2020-01-08
[TfL] Remove old reports after specified time periods

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1608
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1669
Fixes https://github.com/mysociety/fixmystreet/issues/2376

Will be run as `process-inactive-reports --delete 36 --cobrand tfl` and `process-inactive-reports --delete 84 --cobrand tfl` with the code picking the right ones to delete based upon the time period.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -107,6 +107,26 @@ sub report_age { '6 weeks' }
 # We don't want any reports made before the go-live date visible
 sub cut_off_date { '2019-12-09 12:00' }
 
+sub problems_restriction {
+    my ($self, $rs) = @_;
+    return $rs if FixMyStreet->staging_flag('skip_checks');
+    $rs = $self->next::method($rs);
+    my $table = ref $rs eq 'FixMyStreet::DB::ResultSet::Nearby' ? 'problem' : 'me';
+    $rs = $rs->search({
+        "$table.lastupdate" => { '>=', \"now() - '3 years'::interval" }
+    });
+    return $rs;
+}
+
+sub problems_sql_restriction {
+    my ($self, $item_table) = @_;
+    my $q = $self->next::method($item_table);
+    if ($item_table ne 'comment') {
+        $q .= " AND lastupdate >= now() - '3 years'::interval";
+    }
+    return $q;
+}
+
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -127,6 +127,16 @@ sub problems_sql_restriction {
     return $q;
 }
 
+sub inactive_reports_filter {
+    my ($self, $time, $rs) = @_;
+    if ($time == 3*12) {
```
davea: This has the side-effect of selecting all reports unless `$time` is exactly`36` or `84`, which might cause issues if a manual one-off run used e.g. 24 months. Would it be better to have a specific flag for safety-critical reports, or treat anything less than `84` as non-safety critical?

perllib/FixMyStreet/Script/Inactive.pm
```diff
@@ -110,22 +128,39 @@ sub anonymize_reports {
...
+    while (my $problem = $problems->next) {
+        say "Deleting associated data of problem #" . $problem->id if $self->verbose;
+        next if $self->dry_run;
+
+        $problem->comments->delete;
+        $problem->questionnaires->delete;
+        $problem->alerts->delete;
+    }
+    say "Deleting all matching problems" if $self->verbose;
+    return if $self->dry_run;
```
davea: The pattern of wrapping everything in a transaction which is committed or rolled back according to command line flag (see e.g. `bin/tfl/import-categories` and `FixMyStreet::DB::Factories`) might be worth considering to make it less likely one of these would be missed off.

dracos: This script sends email also, though, so can't really do a per-whole-run transaction in case there was e.g. an email failure causing the script to die halfway through.

davea: An excellent point! Nevermind then :)

perllib/FixMyStreet/Cobrand/TfL.pm
dracos: Yep, good idea, done, and added tests for everything now.

PR #2838, opened 2020-01-15
[Peterborough] Ensure 50 char title limit is restricted to Peterborough reports

davea:
For https://github.com/mysociety/fixmystreet-freshdesk/issues/107

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
+++ b/web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
@@ -46,7 +46,12 @@ body_validation_rules = {
           maxlength: 50
         }
     },
-    'Peterborough City Council': confirm_validation_rules,
+    'Peterborough City Council': $.extend({
+        title: {
+          required: true,
+          maxlength: 50
+        }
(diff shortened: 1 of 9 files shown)


web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
```diff
@@ -46,7 +46,10 @@ body_validation_rules = {
           maxlength: 50
         }
     },
-    'Peterborough City Council': confirm_validation_rules,
+    // Use confirm_validation_rules as the prototype for Peterborough, so that
+    // the .title addition below doesn't end up applying to all bodies that
+    // use confirm_validation_rules.
+    'Peterborough City Council': Object.create(confirm_validation_rules),
```
dracos: Well spotted; could we have (iirc) `$.extend({}, confirm_validation_rules)`? IE8 might all be fake bots, I dunno, but we still get a few :)

davea: Good idea, and actually that lets us put the `title` right there instead of lower down.

Issue #2835, opened 2020-01-14
Assets disappear at maximum zoom on OS MasterMap


chrismytton:
Assets disappear at maximum zoom on sites using the MasterMap map layer.

**To Reproduce**

1. Go to https://fix.bexley.gov.uk/report/new?longitude=0.1529159510650684&latitude=51.455160729901294 (works on any site with MasterMap and an assets layer)
2. Select the "Traffic lights" category
3. Zoom in all the way
4. The assets disappear when at maximum zoom, but reappear when you zoom out

**Expected behavior**

Assets should be visible at maximum zoom.

**Screencast**

![asset-zoom-issue](https://user-images.githubusercontent.com/22996/72358459-8466e280-36e4-11ea-96c2-603a61c9c7ad.gif)

**Desktop:**

 - OS: macOS
 - Browser: Reproducible in latest versions of Chrome, Firefox and Safari

chrismytton:
Note: This was actually reported by Peterborough when they were testing the tree asset layer on staging.

chrismytton:
Pull request that should fix this here: https://github.com/mysociety/fixmystreet/pull/2843

chrismytton:
Pull request broke some tests on Northamptonshire, so need to investigate that and try again.

dracos:
I think this was caused because if you specify a min resolution and not a max resolution, it tries to construct all the resolution levels from scaling up from the min resolution (which is not a real resolution if it's 0.00001). When we're setting it to a non-existant low resolution alongside a max resolution, that's for the "check it's in range for display" check, but it is constructing bogus resolutions in the layer, but they're constructed downwards from the (real) max resolution so is okay.
Northants is the only layer not to set max resolution on all layers I think, so doing that would presumably fix it. Or set this default of 0.00001 only if max resolution is set would also work I think, I'll just push that now to run the tests.

chrismytton:
@dracos Thanks, I think I understand now. And looks like that tweak has made the tests pass, hooray!

Do you want to make a PR for that or shall I?

dracos:
Ah, thanks for reminding me, forgot I did this at the end of Friday :) Made a PR now.

PR #2843, opened 2020-01-20
Fix assets disappearing at maximum zoom

chrismytton:
Rather than setting a min_resolution on all of the cobrands just set a
default in web/cobrands/fixmystreet/assets.js. This means that assets
will appear at all zoom levels, which seems like what we want to happen
most of the time.

Fixes https://github.com/mysociety/fixmystreet/issues/2835

### DIFF
--- a/web/cobrands/bathnes/assets.js
+++ b/web/cobrands/bathnes/assets.js
@@ -24,7 +24,6 @@ fixmystreet.maps.banes_defaults = {
     asset_item: "asset",
     asset_type: 'spot',
     max_resolution: 4.777314267158508,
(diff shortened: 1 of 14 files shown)


dracos:
...Sadly the Northamptonshire trees tests fail, so that will need looking at.

chrismytton:
> ...Sadly the Northamptonshire trees tests fail, so that will need looking at.

Oops, missed this when I merged. Will take a look before deploying.

chrismytton:
This fix was reverted on master because of the broken tests. I've re-opened https://github.com/mysociety/fixmystreet/issues/2835 for now while we look at a fix.

PR #2850, opened 2020-01-21
List categories in body admin by group.

dracos:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
         - Add user admin log page.
         - Allow report as another user with only name.
         - Allow staff users to sign other people up for alerts.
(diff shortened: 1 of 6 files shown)


dracos:
I did wonder that, but I thought it was better to see all the groups that it appeared under on the one page, especially if you'd come to the admin from the front end and knew the category you were after was in a particular group (but wouldn't necessarily know it was under multiple groups). Is there an issue with it listing the categories more than once? I guess best would be for it to also say "Also in [other groups]" or similar.

davea:
You're right, perhaps it's fine. I guess the difference on the template/user form is that the categories are associated with input elements and so it doesn't make sense to tick one "Other" but not them all. 👍

PR #2852, opened 2020-01-24
Change Northamptonshire base map

struan:
Update Northamptonshire to use their own tile server. This means adding support for WMS map tiles so includes the code to do this, and to handle displaying the NCC assets on the new projection.

[skip changelog]


web/js/map-wms-base.js
```diff
@@ -0,0 +1,29 @@
...
+    fixmystreet.map_type = OpenLayers.Layer.WMS;
+
+    fixmystreet.map_options = {
+        maxExtent: this.layer_bounds,
+        units: 'm',
+    };
+
+    fixmystreet.layer_options = [];
+    $.each(fixmystreet.wms_config.layer_names, function(i, v) {
+        fixmystreet.layer_options.push({
```
dracos: It seems odd to have a template creating the wms_config data object, and then having the JS having to do lots of work here to use it? Why is the JS here not `fixmystreet.layer_options.push(fixmystreet.wms_config)` or at least trying to map it across with as little work as possible?

web/js/map-OpenLayers.js
```diff
@@ -845,7 +845,19 @@ $.extend(fixmystreet.utils, {
                 // This option is thankfully used by them both
                 numZoomLevels: fixmystreet.numZoomLevels
             }, fixmystreet.layer_options[i]);
-            if (fixmystreet.layer_options[i].matrixIds) {
+            if (fixmystreet.layer_options[i].wms_version) {
```
dracos: Worth sticking fixmystreet.layer_options[i] in another var if it's going to be used a lot more?

templates/web/base/maps/noscript_map_wms.html
```diff
@@ -0,0 +1,32 @@
+<div class="noscript square-map__outer">
+    <div class="square-map__inner">
+        <div id="[% nsm_prefix %]drag">
+            [%- FOR row IN map.tiles -%]
+                [%- FOR tile IN row -%]
+                    [%- top_px = tile.row_offset * map.tile_size -%]
+                    [%- left_px = tile.col_offset * map.tile_size %]
+                    <[% map.img_type %]
```
dracos: I don't know why the safe has disappeared from here compared to the WMTS, it's needed if it's an input type="image".

perllib/FixMyStreet/Map/Northamptonshire.pm
```diff
@@ -0,0 +1,76 @@
...
+        layer_names     => $self->layer_names,
+        wms_version    => '1.1.1',
+        layer_style     => 'default',
+        format          => 'image/png', # appended to tile URLs
+        size            => 256, # pixels
+        dpi             => 96,
+        inches_per_unit => 39.3701,
+        projection      => 'EPSG:27700',
+        origin_x        => -5220400.0,
+        origin_y        => 4470200.0,
```
dracos: These origins are not used.

templates/web/base/maps/wms_config.html
```diff
@@ -0,0 +1,17 @@
...
+        'tile_dpi': [% map.tile_dpi %],
+        'tile_urls': [% map.tile_urls | safe %],
+        'tile_size': [% map.tile_size %],
+        'tile_suffix': '[% map.tile_suffix %]',
+        'layer_names': [% map.layer_names | safe  %],
+        'layer_style': '[% map.layer_style %]',
+        'wms_version': '[% map.wms_version %]',
+        'format': '[% map.format %]',
+        'scales': [% map.scales | safe %],
+        'origin_x': 0,
```
dracos: Does WMS need tileOrigin? Why set these if always going to be 0?

templates/web/base/maps/wms_config.html
```diff
@@ -0,0 +1,17 @@
+<script nonce="[% csp_nonce %]">
+  var fixmystreet = window.fixmystreet || {};
+  fixmystreet.wms_config = {
+        'map_projection': '[% map.map_projection %]',
+        'tile_dpi': [% map.tile_dpi %],
+        'tile_urls': [% map.tile_urls | safe %],
+        'tile_size': [% map.tile_size %],
+        'tile_suffix': '[% map.tile_suffix %]',
```
dracos: There is something wrong here, as this is never set by the server side. Is this line needed (and therefore formatSuffix)?

templates/web/base/maps/wms_config.html
```diff
@@ -0,0 +1,17 @@
+<script nonce="[% csp_nonce %]">
+  var fixmystreet = window.fixmystreet || {};
+  fixmystreet.wms_config = {
+        'map_projection': '[% map.map_projection %]',
+        'tile_dpi': [% map.tile_dpi %],
```
dracos: The DPI is not used anywhere client-side.

templates/web/base/maps/wms_config.html
```diff
@@ -0,0 +1,17 @@
...
+        'tile_urls': [% map.tile_urls | safe %],
+        'tile_size': [% map.tile_size %],
+        'tile_suffix': '[% map.tile_suffix %]',
+        'layer_names': [% map.layer_names | safe  %],
+        'layer_style': '[% map.layer_style %]',
```
dracos: This is not set server side, is it needed?

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+}
+
+# A hash of parameters used in calculations for map tiles
+sub tile_parameters {
+    my $params = {
+        urls         => [ '' ], # URL of the map tiles, up to the /{z}/{x}/{y} part
+        layer_names  => [ '' ],
+        wms_version => '1.0.0',
+        layer_style  => '',
+        matrix_set   => '',
```
dracos: This is not a default, it should be format I guess? And 'suffix' next line is not used anywhere either (see other comment).

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+
+    my $res = $scales[$zoom] /
+        ($self->tile_parameters->{inches_per_unit} * $self->tile_parameters->{dpi});
+
+    return $res;
+}
+
+sub map_tiles {
+    my ($self, %params) = @_;
+    my ($left_col, $top_row, $z) = @params{'x_left_tile', 'y_top_tile'};
```
dracos: $z is not being used here

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+# Given a lat/lon, convert it to OSM tile co-ordinates (nearest actual tile,
+# adjusted so the point will be near the centre of a 2x2 tiled map).
+#
+# Takes parameter for rows/cols.  For even sizes (2x2, 4x4 etc.) will
+# do adjustment, but simply returns actual for odd sizes.
+#
+sub latlon_to_tile_with_adjust {
+    my ($self, $lat, $lon, $zoom, $rows, $cols) = @_;
+    my ($x_tile, $y_tile)
+        = my @ret
```
dracos: What is this line for?

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+sub latlon_to_px($$$$$$) {
+    my ($self, $lat, $lon, $x_tile, $y_tile, $zoom) = @_;
+    my ($pin_x_tile, $pin_y_tile) = $self->reproject_from_latlon($lat, $lon, $zoom);
+    my $res = $self->get_res($zoom);
+    my $pin_x = ( $pin_x_tile - $x_tile ) / $res;
+    my $pin_y = ( $y_tile - $pin_y_tile ) / $res;
+    return ($pin_x, $pin_y);
+}
+
+sub click_to_tile {
```
dracos: This function appears to be doing what in other map types is two functions, click_to_tile and tile_to_latlon, I don't know if there's a reason for that, I'm guessing the tile X/Y co-ords have already been projected? Just making sure I have it clear in my head :)

templates/web/base/maps/noscript_map_wms.html
```diff
@@ -0,0 +1,32 @@
...
+            [% END %]
+        </div>
+        <div id="[% nsm_prefix %]pins">[% FOR pin IN map.pins %][% INCLUDE pin %][% END %]</div>
+        [% INCLUDE 'maps/_compass.html' %]
+    </div>
+</div>
+
+[% BLOCK pin %]
+[%
+    SET pin_top = pin.py ;
```
dracos: I don't think you need any of this - if you only include maps/pin.html, is it not okay? That shifts it 24/64, just like this is doing.

templates/web/base/maps/noscript_map_wms.html
```diff
@@ -0,0 +1,32 @@
+<div class="noscript square-map__outer">
```
dracos: This template is very similar to the WMTS one; could the pin location calculations be factored out?

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+                        dotted_id => $dotted_id,
+                        alt => "Map tile $dotted_id",
+                    }
+                }
+                @col_offsets
+            ]
+        }
+        @row_offsets
+    ];
+}
```
dracos: A lot of this function, and this whole file, is pretty identical/structured same as WMTS base - can we not share these somehow?

perllib/FixMyStreet/Map/WMSBase.pm
```diff
@@ -0,0 +1,274 @@
...
+
+sub tile_base_url {
+    my $self = shift;
+    my $params = $self->tile_parameters;
+    return sprintf '%s?version=%s&format=%s&size=%s&width=%s&height=%s&service=WMS&layers=%s&request=GetMap&srs=%s',
+        $params->{urls}[0], $params->{wms_version}, $params->{format}, $params->{size}, $params->{size},
+        $params->{size}, $params->{layer_names}[0], $params->{projection};
+}
+
+# Given a lat/lon, convert it to OSM tile co-ordinates (nearest actual tile,
```
dracos: It's not OSM tile coordinates, I assume. Rest of comment doesn't appear to match either.

perllib/FixMyStreet/Map/WMSBase.pm
struan: Yeah, the tile coords can be directly reprojected so there's no need for the second tile_to_latlon function/

dracos:
There were two issues - scaling image maps doesn't scale the X/Y co-ordinates sent to the server, so any display bar precisely 100% would give a wrong reading. With that fixed (don't know if it breaks something else, Zurich still broken but will it break its print screen?), pins/etc worked but were a tiny bit out. This is because need to use US Survey Inch of 39.37 instead of 39.3701, but of course. Also noticed compass was broken - this will need an explicit tile_to_latlon function in WMS, even if it doesn't do anything (clearer if it's consistent with the other WMTS/OSM as well anyway). Haven't been able to check WMS is working, but my changes didn't change anything in that area.

struan:
@dracos could you have a once over of 136e0a4d9f89c8d8d875ac2d51fbe8046ad44afd and 58227c819227415a5058a14d335fc2b52448452e to see if they make sense?

dracos:
Look good to me (assuming fid works :) ).

Issue #2853, opened 2020-01-24
Allow multiple questions to disable the new report form


chrismytton:
**Describe the bug**

When there are multiple questions with the **Disable form when this option is selected** checkbox ticked in the admin only the last question seems to actually disable the form.

**To Reproduce**

* Add multiple questions to a category in the admin which have the **Disable form when this option is selected** for one of the options
* Make a new report in that category
* When you select the option that should disable the form for the first question nothing happens

chrismytton:
PR to hopefully fix this here: https://github.com/mysociety/fixmystreet/pull/2854

PR #2854, opened 2020-01-24
Allow multiple question to disable the new report form

chrismytton:
This changes the existing logic to account for the fact that multiple questions might have the disable form checkbox ticked for one of their options. Previously it only worked for one question which disabled the form.

At the moment this just displays one of the messages at a time, with the first one taking priority. So if you have a "Is this an emergency" question and a "Is this on private land" question which both disable the
form, and the user has selected "Yes" to both, then only the message for the first question is displayed. This can potentially be improved in the future, but seemed out of scope for this change.

Fixes https://github.com/mysociety/fixmystreet/issues/2853

Original PR which added the disable for option for question: https://github.com/mysociety/fixmystreet/pull/2599

### DIFF
--- a/.cypress/cypress/integration/peterborough.js
+++ b/.cypress/cypress/integration/peterborough.js
@@ -1,0 +1,34 @@
+describe('new report form', function() {
+
+  beforeEach(function() {
+    cy.server();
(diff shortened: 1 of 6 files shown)


chrismytton:
Cancelling review request, still need to write another test!

PR #2862, opened 2020-01-28
Only trigger one change event on initial popstate.

dracos:
When going back to the initial state with popstate, a change event was
being triggered on every single option of the filter selects. This led
to a lot of change events running on the category/status multi-selects
which then needlessly repeated the same activities over and over. This
locked up the browser for seconds in locations with many categories.

Performance chart before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@
         - Checking of cached front page details against database. #2696
         - Inconsistent display of mark private checkbox for staff users
         - Clear user categories when staff access is removed. #2815
+        - Only trigger one change event on initial popstate.
     - Development improvements:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -60,6 +60,7 @@
         - Checking of cached front page details against database. #2696
         - Inconsistent display of mark private checkbox for staff users
         - Clear user categories when staff access is removed. #2815
+        - Only trigger one change event on initial popstate.
```
zarino: Maybe this should mention it’s for the `filter_categories` dropdown?

dracos: And the status one too :) I think anyone who cares can look at the code/git log, want to keep this high level.

PR #2863, opened 2020-01-28
Improve mobile try again interface

### LINKED ISSUE #2667: Remove "try again" overlay on mobile?
From #2614:
> I feel we should probably lose the "Try again" overlay entirely, perhaps? So that people can scroll back up the page rather than have to click either this or the "Map" link that they might not realise is a link.

dracos:
This PR renames "Try again" when you first click map to "Nearby reports". You don't need to click "try again" in order to... try again, you can click the map again or drag the pin, clicking try again is actually more annoying because it takes you back to around loading nearby pins and starting the whole thing again which you might well not want. Update the heading to explain you can reposition before hitting OK.

Going back up from the form to the map and clicking "Try again" doesn't go back to around either, but only resets the map to allow repositioning of the pin again. You can go back to nearby reports from there if you desire.

This PR also restricts lat/lon in URL to 6dp, and stops history state being pushed on each map drag/click (assuming those are only corrections to the first).

Fixes #2667.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1459,6 +1459,9 @@ fixmystreet.display = {
             height = $map_box.height();
         $map_box.append(
             '<p class="sub-map-links" id="mob_sub_map_links">' +
+            '<a href="#" id="problems_nearby">' +
+                translation_strings.back +
+            '</a>' +
             '<a href="#" id="try_again">' +
                 translation_strings.try_again +
             '</a>' +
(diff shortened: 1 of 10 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1457,33 +1491,18 @@ fixmystreet.display = {
...
-            $('html, body').animate({ scrollTop: 0 }, 1000, function(){
-                $('html').addClass('only-map');
-                $('#mob_sub_map_links').removeClass('map_complete');
-                $('#mob_ok').text(translation_strings.ok);
-                if (fixmystreet.duplicates) {
-                    fixmystreet.duplicates.hide();
-                }
-            });
-        });
+        $('#mob_ok').toggle(fixmystreet.mobile_reporting.confirmed_location, fixmystreet.mobile_reporting.try_again);
```
davea: There's something a bit odd about this button on iOS - sometimes it needs two clicks to activate it. The bug appears consistently when half-scrolling up to the map then clicking the "Try again" button; but generally doesn't happen when clicking the "Map" button with the chevron:



try_again_75_full

davea: Steps to reproduce (on iOS):

1. Enter a location on front page
1. Tap the map to place a pin
1. Tap "OK"
1. Wait for "Report your problem" form to scroll in to view
1. Scroll up until the "Try again" button is visible on map
1. Tap "Try again" button
1. Wait for map to scroll into view and overlay to disappear
1. (optional, doesn't seem to affect bug incidence) Tap map to reposition report.
1. Tap "OK" button
1. Notice that nothing happens, apart from the "OK" button becoming highlighted.
1. Tap "OK" button again
1. "Report your problem" form scrolls in to view as expected.

These steps work as expected:

1. Enter a location on front page
1. Tap the map to place a pin
1. Tap "OK"
1. Wait for "Report your problem" form to scroll in to view
1. Scroll up until the "Map" button is visible
1. Tap "Map" button
1. Wait for map to scroll into view and overlay to disappear
1. (optional, doesn't seem to affect bug incidence) Tap map to reposition report.
1. Tap "OK" button
1. "Report your problem" form scrolls in to view as expected.

Is there some kind of `mouseenter` event or `:hover` CSS class on the "OK" button perhaps. Because the "OK" button is the same DOM element as the "Map" button, I wonder if tapping it to reveal the map counts as the first tap which highlights the button, and the second activates it. Clicking the "Try again" button doesn't activate the hover state on the "Map"/"OK" button, hence two taps required.

dracos:
Okay, have dropped it to 500 (I just copied what it was before); renamed to Back/Continue; and hopefully fixed the double-tap issue.

dracos:
Grr, hopefully got it this time!

PR #2866, opened 2020-01-30
Add cobrand-specific manifest customisation

### LINKED ISSUE #2792: Allow customisation of manifest
Depends on https://github.com/mysociety/fixmystreet/issues/2220

Allow bits of the manifest to be customisable, e.g. the theme colour, app logos, and so on.
This could also be used/expanded for e.g. favicon, HTML theme colour, and then to being able to customise basic logo/styling in admin itself, though having non-static stylesheets more of a challenge!

For starters, going by #2220 work (see manifest commit on pwa branch), an admin page that lets you set, for a cobrand:
* name
* short_name
* theme_colour
* background_colour
* Icons (192 and 512 are defaults, but perhaps any square multiple of 48), saved in `web/theme/*moniker*/icons/` maybe?

I'd guess a theme database table with cobrand/key/value columns? Or cobrand and those things as columns if we don't think it's going to extend a lot...

Then that'd be picked up for the manifest. After that exists, it could set the theme colour in the HTML head (as fixmystreet.com cobrand does manually at present), and also then favicon would be next step, new ticket by then though I'd say!

davea:
Adds admin UI for superusers to manage the `ManifestTheme` model for each cobrand. Allows the names/icon/colours to be customised.

## Desktop Chrome PWA theming:
Screenshot 2020-01-30 at 15 48 36

image

## Installing on iOS:
Installing | Icon
---------- | ----
IMG_1640 | IMG_D33429854842-1

## Managing via cobrand admin
lincolnshire fms davea me_admin_manifesttheme_lincolnshire (2)

## fixmystreet.com can edit all cobrand themes
image

## Guidance tooltips
fms davea me_admin_manifesttheme_tfl

### Possible future improvements
 - Colour picker UI in form
 - Live preview of app/home screen UI alongside form fields

Please check the following:

Fixes #2792.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
@@ -67,6 +67,7 @@ sub form {
 
     if ($c->get_param('delete_theme')) {
         $c->forward('_delete_all_manifest_icons');
+        $c->forward('/offline/_clear_manifest_icons_cache', [ $theme->cobrand ]);
         $theme->delete;
         $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', 'delete' ]);
         $c->response->redirect($c->uri_for($self->action_for('index')));
@@ -82,6 +83,7 @@ sub form {
     return unless $form->validated;
 
     $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', $action ]);
+    $c->forward('/offline/_clear_manifest_icons_cache', [ $theme->cobrand ]);
     $c->response->redirect($c->uri_for($self->action_for('index')));
 }
 
--- a/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
@@ -39,7 +39,14 @@ sub item :PathPart('admin/manifesttheme') :Chained :CaptureArgs(1) {
 
 sub edit :PathPart('') :Chained('item') :Args(0) {
     my ($self, $c) = @_;
-    return $self->form($c, $c->stash->{obj});
+
+    my $form = $self->form($c, $c->stash->{obj});
+
+    # We need to do this after form processing, in case a form POST has deleted
+    # an icon.
+    $c->forward('/offline/_stash_manifest_icons', [ $c->stash->{obj}->cobrand, 1 ]);
+
+    return $form;
 }
 
 
@@ -59,6 +66,7 @@ sub form {
     my ($self, $c, $theme) = @_;
 
     if ($c->get_param('delete_theme')) {
+        $c->forward('_delete_all_manifest_icons');
         $theme->delete;
         $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', 'delete' ]);
         $c->response->redirect($c->uri_for($self->action_for('index')));
@@ -68,13 +76,22 @@ sub form {
     my $action = $theme->in_storage ? 'edit' : 'add';
     my $form = FixMyStreet::App::Form::ManifestTheme->new( cobrand => $c->cobrand->moniker );
     $c->stash(template => 'admin/manifesttheme/form.html', form => $form);
-    $form->process(item => $theme, params => $c->req->params);
+    my $params = $c->req->params;
+    $params->{icon} = $c->req->upload('icon') if $params->{icon};
+    $form->process(item => $theme, params => $params);
     return unless $form->validated;
 
     $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', $action ]);
     $c->response->redirect($c->uri_for($self->action_for('index')));
 }
 
+sub _delete_all_manifest_icons :Private {
+    my ($self, $c) = @_;
 
+    $c->forward('/offline/_stash_manifest_icons', [ $c->stash->{obj}->cobrand, 1 ]);
+    foreach my $icon ( @{ $c->stash->{manifest_icons} } ) {
+        unlink FixMyStreet->path_to('web', $icon->{src});
+    }
+}
 
 1;
--- a/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
@@ -44,7 +44,7 @@ sub edit :PathPart('') :Chained('item') :Args(0) {
 
     # We need to do this after form processing, in case a form POST has deleted
     # an icon.
-    $c->forward('/offline/_stash_manifest_icons', [ $c->stash->{obj}->cobrand, 1 ]);
+    $c->stash->{editing_manifest_theme} = $c->forward('/offline/_find_manifest_theme', [ $c->stash->{obj}->cobrand, 1 ]);
 
     return $form;
 }
@@ -67,7 +67,7 @@ sub form {
 
     if ($c->get_param('delete_theme')) {
         $c->forward('_delete_all_manifest_icons');
-        $c->forward('/offline/_clear_manifest_icons_cache', [ $theme->cobrand ]);
+        $c->forward('/offline/_clear_manifest_theme_cache', [ $theme->cobrand ]);
         $theme->delete;
         $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', 'delete' ]);
         $c->response->redirect($c->uri_for($self->action_for('index')));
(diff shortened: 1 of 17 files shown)


perllib/FixMyStreet/App/Form/ManifestTheme.pm
```diff
@@ -0,0 +1,66 @@
...
+use FixMyStreet::App::Form::I18N;
+extends 'HTML::FormHandler::Model::DBIC';
+use namespace::autoclean;
+
+has 'cobrand' => ( isa => 'Str', is => 'ro' );
+
+has '+widget_name_space' => ( default => sub { ['FixMyStreet::App::Form::Widget'] } );
+has '+widget_tags' => ( default => sub { { wrapper_tag => 'p' } } );
+has '+item_class' => ( default => 'ManifestTheme' );
+has_field 'cobrand' => ( required => 0 );
```
dracos: Could this be a dropdown, from `available_cobrand_classes` like ReportExtraFields does?

perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -77,6 +77,10 @@ sub index : Path : Args(0) {
         $c->detach;
     }
 
+    # TODO: Not sure we want to hammer the FS for every front page request,
```
dracos: I agree, this will hurt performance. Could the stash function check/set something in Memcache, and then an edit/delete of the icons clears it much like it already does for States?

perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
```diff
@@ -0,0 +1,97 @@
...
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+use FixMyStreet::App::Form::ManifestTheme;
+
+sub auto :Private {
+    my ($self, $c) = @_;
+
+    if ( $c->cobrand->moniker =~ /^default|fixmystreet$/ ) {
```
dracos: No working site will have `default` as their moniker, though. Perhaps if the length of `available_cobrand_classes` is equal to 1 rather than the default check? I realise still need `fixmystreet` (or something) to be the "parent" cobrand.

dracos: Also just to note that regex matches anything beginning with default or anything ending with fixmystreet :)

db/downgrade_0071---0070.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+DROP TABLE manifest_theme;
+
+ALTER TABLE admin_log DROP CONSTRAINT admin_log_object_type_check;
+
+ALTER TABLE admin_log ADD CONSTRAINT admin_log_object_type_check CHECK (
```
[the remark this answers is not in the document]
davea: Oops. I'll bring all of the DB changes into the first commit before merging.

perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
davea: If there's only one cobrand available then wouldn't it be better to bypass the list page and only allow editing of that single cobrand's theme?

perllib/FixMyStreet/App/Controller/Root.pm
davea: Good idea, will add that.

perllib/FixMyStreet/App/Form/ManifestTheme.pm
davea: That'd be nicer, yes.

perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
dracos: Yes, that's what I meant, sorry if I wasn't clear - instead of checking for 'default' check if available is equal to 1 and do the same.

templates/web/base/common_header_tags.html
```diff
@@ -2,6 +2,9 @@
 
 <meta http-equiv="content-type" content="text/html; charset=utf-8">
 <link rel="manifest" href="/.well-known/manifest.webmanifest">
+[% FOREACH icon IN manifest_icons %]
+    <link rel="apple-touch-icon" sizes="[% icon.sizes %]" href="[% icon.src %]">
+[% END %]
```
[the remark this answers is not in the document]
davea: One small refactor later and it's there 👍

perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
```diff
@@ -0,0 +1,99 @@
...
+
+    $c->forward('/admin/log_edit', [ $theme->id, 'manifesttheme', $action ]);
+    $c->forward('/offline/_clear_manifest_theme_cache', [ $theme->cobrand ]);
+    $c->response->redirect($c->uri_for($self->action_for('index')));
+}
+
+sub _delete_all_manifest_icons :Private {
+    my ($self, $c) = @_;
+
+    $c->forward('/offline/_stash_manifest_theme', [ $c->stash->{obj}->cobrand, 1 ]);
```
dracos: I guess if it was done on every request, you wouldn't need to call this here? But no harm in it.

perllib/FixMyStreet/App/Controller/Root.pm
```diff
@@ -77,6 +77,8 @@ sub index : Path : Args(0) {
         $c->detach;
     }
 
+    $c->forward('/offline/_stash_manifest_theme', [ $c->cobrand->moniker ]);
```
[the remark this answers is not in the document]
davea: Yes, of course! Sorted in 67469ac689d935aee89d0a59f9ed1cab010ea0a5.

perllib/FixMyStreet/App/Controller/Admin/ManifestTheme.pm
davea: Actually no, because the call in `Root::auto` will include the default theme & icons if there isn't one present in the DB. This call will ignore any default icons on disk and memcache.

dracos: Hmm, so does that mean those admin pages lose the icons/theme-colour in their header? Or if I'm editing a council cobrand's theme on .com, that page is themed with the cobrand? Just feels like something could get confused - it feels like it should be able to have the stashed theme for the page, and a separate theme for being edited? Or have I misunderstood.

davea: You're right - I hadn't thought about the current theme of the page and the theme you're editing as two separate things. I've tidied things up and made it clearer in 6074fdb3362993853b1292f069170d52bb83b36b.

perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -33,42 +33,14 @@ sub manifest: Path("/.well-known/manifest.webmanifest") {
...
-            };
-        }
-    );
-
-    unless (@icons) {
-        push @icons,
-            { src => "/cobrands/fixmystreet/images/192.png", sizes => "192x192", type => "image/png" },
-            { src => "/cobrands/fixmystreet/images/512.png", sizes => "512x512", type => "image/png" };
-    }
+    $c->forward("_stash_manifest_theme", [ $c->cobrand->moniker ]);
```
dracos: This has already been run by the root auto, so isn't needed here.

perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -82,6 +54,72 @@ sub manifest: Path("/.well-known/manifest.webmanifest") {
...
 }
 
+sub _stash_manifest_theme : Private {
+    my ($self, $c, $cobrand) = @_;
+
+    $c->stash->{manifest_theme} = $c->forward('_find_manifest_theme', [ $cobrand ]);
+}
+
+sub _find_manifest_theme : Private {
+    my ($self, $c, $cobrand, $ignore_cache_and_defaults) = @_;
```
dracos: Now the find/stash thing has been tidied, looking at all the calls, I wonder if it'd be nicer to have the default be ignore (so every direct call to _find_manifest can only pass in the cobrand), and it's only the call in stash_manifest_theme that would have to pass a flag to say "use_cache_and_defaults".

perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -82,6 +54,72 @@ sub manifest: Path("/.well-known/manifest.webmanifest") {
...
+        }
+    }
+
+    return $manifest_theme;
+}
+
+sub _clear_manifest_theme_cache : Private {
+    my ($self, $c, $cobrand ) = @_;
+
+    Memcached::set("manifest_theme:$cobrand", "");
```
dracos: This could be a delete now you've added one?

perllib/FixMyStreet/App/Controller/Offline.pm
davea: I'm 50/50 on this. I agree that the 2 of the 3 calls passing the flag is an indication that it should be the default, but I also think for this specific function the default behaviour should be the most efficient/fastest i.e. to use the cache. Is it compounded by the flag actually doing two separate but unrelated things? Maybe it would be better as a `%params` hash? Dunno.

PR #2868, opened 2020-01-31
Make it clearer that full name is required when reporting

zarino:
Requested by the folks at Bucks. I figure it makes sense to change it across the board though [for all UK councils that include the checking code.]

It’s annoying that we even have to mandate a "full name" (who says everyone’s "full name" has a space in it!?) but it’s a requirement of most councils’ case management systems.

At least this way, users get an up-front hint that they need to enter a full name (ie: with at least two words) before they submit the form and the validation runs.

[skip changelog]


zarino:
@chrismytton go ahead!

PR #2870, opened 2020-02-03
[UK] Use feature flag for send_questionnaire.

dracos:
Will need a feature adding to match current behaviour of:
```
send_questionnaire:
  fixmystreet:
    TfL: 0
```
[skip changelog] Connects with https://github.com/mysociety/fixmystreet-freshdesk/issues/111


dracos:
Added a test for that config ;)

PR #2877, opened 2020-02-06
[TfL] Bus stops outside London

dracos:
After this change, you can safely add the TfL body to other councils (namely the ones surrounding London that include TfL bus stops); clicking in those councils on other cobrands will have no change, as the TfL body will be fully ignored on a non-London council, but on the TfL cobrand it will allow it for e.g. the selection of bus stops outside London.

It removes the body rather than just all its categories to prevent a "We do not have details of the other body covering this area" message appearing.

Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/116 [skip changelog]

### DIFF
--- a/.cypress/cypress/fixtures/tfl-bus-stops.xml
+++ b/.cypress/cypress/fixtures/tfl-bus-stops.xml
@@ -1,0 +1,176 @@
+<?xml version='1.0' encoding="UTF-8" ?>
+<wfs:FeatureCollection
+   xmlns:ms="http://mapserver.gis.umn.edu/mapserver"
+   xmlns:gml="http://www.opengis.net/gml"
+   xmlns:wfs="http://www.opengis.net/wfs"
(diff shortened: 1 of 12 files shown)


dracos:
You haven't missed anything; yes, you have to have already picked the Bus stops category first (by clicking somewhere within London first) for it to work at present. Was feeling pretty good I'd got it to work at all, but now I look back at it, it's pretty rubbish!

I'll go back to the drawing board - I guess really we should add all the councils around London to being covered by the TfL body, and remove the LBO-only restriction. This would let people search for anywhere in those councils, click the map in any of those councils - the red route error would still work fine, and the other categories have assets so would make sense too.

We'd not want them to show on .com (confusing double bus stop/traffic lights categories), nor show any of those areas on the all reports page/have them be 'children'. But that's probably a much better solution.

dracos:
Okay, have redone this a totally different way, that feels nicer.

PR #2882, opened 2020-02-10
[Buckinghamshire] Change "Name" to "Full name" on forms

chrismytton:
Since we require the user to enter their full name Buckinghamshire have asked us to change the text to "Full name" to make it clearer that the full name is required to complete the form.

Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/122

There is the wider issue of doing this for all cobrands that have the validation for full name on them, but that will be done separately, as part of https://github.com/mysociety/fixmystreet/pull/2868.

### DIFF
--- a/templates/web/buckinghamshire/report/form/user_name.html
+++ b/templates/web/buckinghamshire/report/form/user_name.html
@@ -1,0 +1,13 @@
+<!-- user_name.html -->
+<label for="form_name">Full name
+[% TRY %]
+    [% INCLUDE 'report/form/after_name.html' %]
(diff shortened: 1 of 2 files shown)


templates/web/buckinghamshire/report/form/user_name.html
```diff
@@ -0,0 +1,13 @@
+<!-- user_name.html -->
+<label for="form_name">[% loc('Full name') %]
```
davea: Don't need the `loc()` here really :)

PR #2883, opened 2020-02-11
Fix category extra AJAX call for categories with ampersands

davea:
The `IF category_extras.$category.size` condition in category_extras.html was using an escaped string for `$category`, meaning the lookup would always fail for categories that had an ampersand in their name. The result of this was an empty `<div id="category_meta">`, and no extra fields shown to the user. This caused issues when one or more of the fields were required, as they were always empty and the error message wasn't being rendered and shown to the user. This meant the user would see their report form over and over again when submitting instead of useful feedback.

Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/126

[skip changelog]

### DIFF
--- a/templates/web/base/report/new/category_extras.html
+++ b/templates/web/base/report/new/category_extras.html
@@ -1,5 +1,6 @@
 [% SET default_list = [] %][% FOR b IN bodies_to_list.values %][% default_list.push(b.cobrand_name) %][% END %]
 [% DEFAULT list_of_names = default_list %]
+[% category = mark_safe(category) %]
 
 <div id="category_meta">
   [%- IF unresponsive.$category %]
--- a/t/app/controller/report_new_open311.t
+++ b/t/app/controller/report_new_open311.t
@@ -360,6 +360,31 @@ subtest "Category extras includes description label for user" => sub {
     };
 };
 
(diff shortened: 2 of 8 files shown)


templates/web/base/report/new/category_extras.html
```diff
@@ -1,5 +1,6 @@
 [% SET default_list = [] %][% FOR b IN bodies_to_list.values %][% default_list.push(b.cobrand_name) %][% END %]
 [% DEFAULT list_of_names = default_list %]
+[% category = mark_safe(category) %]
```
dracos: Does this need to be earlier in the template process - e.g. the councils_text.html templates etc also look up `unresponsive.$category` which presumably could fail for the same reason.

davea: I've made sure they're all SafeStrings in d6a7f98c81a19baedc5305cb94483a052fd74b4f - does that look OK? Doesn't seem to matter if something is run through `mark_safe` twice, and makes it obvious in the template what's happening.

dracos: It does double wrap, which it could perhaps not, but that doesn't have any effect on output, no.
Alternative, if you're always going to trust category anyway, would be to mark it safe in the perl code rather than in the templates? Meh, don't know if that'd be any easier/better.

PR #2884, opened 2020-02-11
Include explanation of private reports in other-reported email

dracos:
If a report is made using "Report as other" and marked as private, the email that gets sent to the user telling them this should explain that it's a private report. We also make the link in that email work (even if not logged in) but don't log them in, in case they forward the email on (could change that easily enough). Fixes https://github.com/mysociety/fixmystreet-freshdesk/issues/125

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -525,6 +525,31 @@ sub tokenised_url {
     return "/M/". $token->token;
 }
 
+has view_token => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my $self = shift;
+        my $token = FixMyStreet::App->model('DB::Token')->create({
+            scope => 'alert_to_reporter',
+            data => { id => $self->id }
+        });
+    },
+);
+
+=head2 view_url
+
+Return a url for this problem report that will always show it
+(even if e.g. a private report) but does not log the user in.
+
+=cut
+
+sub view_url {
+    my $self = shift;
+    return $self->url unless $self->non_public;
+    return "/R/" . $self->view_token->token;
+}
+
 =head2 is_hidden
 
 Returns 1 if the problem is in an hidden state otherwise 0.
--- a/templates/email/default/other-reported.html
+++ b/templates/email/default/other-reported.html
@@ -12,12 +12,20 @@ INCLUDE '_email_top.html';
   [% start_padded_box | safe %]
   <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
   <p style="[% p_style %]">Your report to [% report.body %] has been logged on [% site_name %].</p>
+
+[% IF report.non_public %]
+<p style="[% p_style %]">It has been marked as private and will not be visible
+to the general public; you may view it using the link below, or if you sign in
+using the email address associated with the report.</p>
+[% END %]
+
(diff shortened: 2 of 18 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -525,6 +525,27 @@ sub tokenised_url {
...
 }
 
+=head2 view_url
+
+Return a url for this problem report that will always show it
+(even if e.g. a private report) but does not log the user in.
+
+=cut
+
+sub view_url {
```
davea: Calling this twice in the HTML and plaintext versions of the email creates two tokens. Not a big deal but seems like it could be avoided.

templates/email/default/other-reported.html
```diff
@@ -12,14 +12,22 @@
   [% start_padded_box | safe %]
   <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
   <p style="[% p_style %]">Your report to [% report.body %] has been logged on [% site_name %].</p>
+
+[% IF report.non_public %]
+<p style="[% p_style %]">It has been marked as private and will not be visible
+to the general public; you may view it using the link below, or if you sign in
+using the email address associated with the report.</p>
```
davea: This link stops working after a month. We should a) indicate this to the user and b) make it clear that after a month they will *need* to login with their email address to view the report.

dracos: At that point, the website does say "If you are the problem reporter, or a member of staff, please <a href="%s">sign in</a> to view this report." which I thought covered it well enough?

dracos:
"attempting to leave an update on the report will fail if the user's not logged in." - think I must have missed something - the update form will be the logged-out form, so it'll ask you for email, or for you to sign in, same as usual? Did it do something different for you?

templates/email/default/other-reported.html
davea: The message shown is actually: 
image

I can imagine a scenario where a user refers back to the email each time they want to view the report, so it'd be good to manage their expectations around the time limited nature of the link.

davea:
I get an access denied message (yes, using the same email address as the report):
signin_update

templates/email/default/other-reported.html
dracos: Ah, right, sorry, yep, perhaps it shouldn't expire the token (given it doesn't log you in). Or (easier) not have a token at all, scrub that bit of code, and always just give the "please log in" page. That would fix the leaving an update issue as well.

dracos:
I guessed I had logged in whilst leaving the update, but that gives the same error you get, so I dunno how it had ever worked for me! I think perhaps dropping the token entirely makes it clearer, if I tweak the text to say you'll have to sign in after clicking it.

templates/email/default/other-reported.html
```diff
@@ -12,14 +12,22 @@
   [% start_padded_box | safe %]
   <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
   <p style="[% p_style %]">Your report to [% report.body %] has been logged on [% site_name %].</p>
+
+[% IF report.non_public %]
```
davea: Looks like the cobrand-templates check missed this, but this block and `view_url` should probably be added to the overridden templates in `tfl`, `hounslow`, and `buckinghamshire` (and maybe `fixamingata`?).

PR #2886, opened 2020-02-12
Ensure that Vagrant box packages are kept updated

sagepe:
This adds a script, `bin/install_packages`, that will install packages
listed in versions of `conf/packages*` and calls it from the bootstrap
script in a way appropriate to the environment it is being run under.

This should ensure that, for example, changes to dependencies will be
applied to Vagrant machines in-between tagged releases.

Please check the following:


sagepe:
How about this - just default to `packages.generic` unless we think we' in Docker.

If that's OK, I'll tidy up the commits and rebase against master.

dracos:
Yep, looks good, thanks!

PR #2887, opened 2020-02-13
Allow anonymous reporting on per-category basis

dracos:
This passes in the category to `allow_anonymous_reports` function so it can change its answer on a per-category basis should it wish to. It also passes the result to the client JS so it can show/hide the button as appropriate. In a cobrand where it is always allowed for all categories, this does mean it's adding `"allow_anonymous": true` to every category of the output, but with gzip that shouldn't make much difference.

Also fixes a bug where the report anonymously button wasn't being shown in non-JS scenario.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -272,9 +272,11 @@ sub by_category_ajax_data : Private {
     }
 
     my $non_public = $c->stash->{non_public_categories}->{$category};
+    my $anon_button = ($c->cobrand->allow_anonymous_reports($category) eq 'button');
     my $body = {
         bodies => [ map { $_->name } @bodies ],
         $non_public ? ( non_public => JSON->true ) : (),
+        $anon_button ? ( allow_anonymous => JSON->true ) : (),
     };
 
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -272,9 +272,11 @@ sub by_category_ajax_data : Private {
     }
 
     my $non_public = $c->stash->{non_public_categories}->{$category};
+    my $anon_button = ($c->cobrand->allow_anonymous_reports($category) eq 'button');
```
davea: I wonder if there's a way to pass a contact here instead of just the category name - it'd allow for more expressive things down the line e.g. anonymous reports on .com for a specific body, storing the anonymous-allowed flag in contact `extra` (and admin UI to control it), only allowing anonymous reports if the send_method is email, all that sort of thing.

dracos: See what you mean, but annoyingly contact is elusive, given technically (though unlikely/never going to happen) it could be an overlap category with two contacts from two different bodies.

davea: Ah well, nevermind :)

PR #2890, opened 2020-02-17
[HighwaysEngland] road number attribute now lower case

struan:
The data has been updated and it seems the case of the attributes has
changed.

[skip changelog]


dracos:
Sorry, yes, it had changed but I'd put a fix in tilma so the output would be the same but then forgot to deploy it.

PR #2891, opened 2020-02-18
Staff-only categories

dracos:
Adds a new "staff" contact state. This state will appear when reporting if you are logged in as a staff member of the relevant body, but otherwise not appear on the site anywhere.

(Also factors some code, fixes a bug in createsuperuser)


templates/web/base/admin/bodies/contact-form.html
```diff
@@ -40,6 +40,11 @@
             <label for="state-deleted">[% loc('Deleted') %]</label>
             <p class="form-hint" id="state-deleted-hint">[% loc('Prevent new reports from using this category, <em>and</em> also remove it from map filters.') %]</p>
         </div>
+        <div class="form-check form-check--inline">
+            <input type="radio" name="state" id="state-deleted" aria-describedby="state-deleted-hint" value="staff"[% ' checked' IF contact.state == 'staff' %]>
+            <label for="state-deleted">[% loc('Staff only') %]</label>
+            <p class="form-hint" id="state-deleted-hint">[% loc('Only staff users will be able to add reports in this category.') %]</p>
```
davea: Looks like you've accidentally reused `deleted` on these three lines instead of e.g. `staff`.

Also - should it be made clear that reports will be publicly visible unless the "Hide reports made in this category" box is also ticked? Would it be going too far to tick that box automatically (or even make it mandatory) when the 'staff' option is selected?

t/app/controller/report_new_staff.t
```diff
@@ -224,4 +224,26 @@ for my $test (
...
+    }, sub {
+        $inspector->update({ is_superuser => 1 });
+        $mech->log_in_ok('inspector@example.org');
+
+        $mech->get_ok('/admin/body/' . $body_ids{2651} . '/Trees');
+        $mech->submit_form_ok({ with_fields => { state => 'staff' } }, 'mark Trees as staff-only');
+        $edin_trees->discard_changes;
+        is $edin_trees->state, 'staff', 'category is staff only';
+
+        my $extra_details = $mech->get_ok_json( '/report/new/ajax?latitude=55.952055&longitude=-3.189579' );
```
davea: It'd be good to have a test that non-superuser staff can see their categories, and also that they can't see any staff categories from another body.

templates/web/base/admin/bodies/contact-form.html
dracos: Hmm, I realise it's been asked for here in terms of making private reports, but I could see there being categories that were staff only that could nonetheless be normal public reports, so I think I'll leave it for now (sure you'll be right and they'll not do it right).

PR #2892, opened 2020-02-24
[Bromley] Fix shortlist button styling for inspectors

zarino:
The shortlist "star" icons were losing their styling, because Bromley’s default button styling was more specific.

Bromley’s existing styles only affected "primary" buttons anyway (ie: elements with `.btn-primary` rather than just regular `.btn`) so buttons at the bottom of the first stage of the reporting form, for example, were still showing the default FMS white/grey gradient. I’ve not added any `.btn` styling in this pull request, so that’s not changed.

### Before

Screenshot_2020-02-24 Viewing a location

### After

Screenshot_2020-02-24 Viewing a location(1)

Fixes mysociety/fixmystreet-commercial#1659.

[skip changelog]

### DIFF
--- a/web/cobrands/bromley/_colours.scss
+++ b/web/cobrands/bromley/_colours.scss
@@ -2,7 +2,7 @@
 
 $menu-image: 'menu-black';
 
-$bromley_blue: rgb(91,120,147);
+$bromley_blue: #647890;
 $bromley_green: #235e1c;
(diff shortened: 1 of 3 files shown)


zarino:
@dracos ok, completely new version this time round, with that nasty `input[type="submit"]` styling gone. Do you think I should add a changelog entry for the improvement to `button-variant()`?

PR #2895, opened 2020-02-25
Optional server-side tracking for unsuccessful location searches

zarino:
If a SQLite file exists at `../data/analytics.sqlite` with a table named `location_searches_with_no_results`, then a row will be created for each `/around` search that returns no results.

Assuming a search for "heath field close" on fixmystreet.com, that returned no results, it would log:

| datetime | cobrand | geocoder | url | user_input |
| --- | --- | --- | --- | --- |
| 2020-02-25 14:57:48 | fixmystreet | OSM | https://nominatim.openstreetmap.org/search?countrycodes=gb&format=json&q=heath+field+close | heath field close |

Part of mysociety/fixmystreet-commercial#1663.

@dracos I attempted to make the whole thing "optional" by wrapping it in a `try catch` block – not sure whether that’s a good idea or not?

Also, do you think it’s worth adding tests for this? Maybe not worth testing the actual analytics.sqlite logging behaviour, but I’m wondering whether we should have a test that checks the extra bit we added doesn’t break the site if you, say, don’t have an `analytics.sqlite` file in place?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,7 @@
         - Make front page cache time configurable.
         - Better working of /fakemapit/ under https.
         - Improve Open311 error output on failing GET requests.
+        - Optionally log failed geocoder searches.
     - Backwards incompatible changes:
         - If you wish the default for the showname checkbox to be checked,
           add `sub default_show_name { 1 }` to your cobrand file.
--- a/perllib/FixMyStreet/App/Controller/Location.pm
+++ b/perllib/FixMyStreet/App/Controller/Location.pm
@@ -6,6 +6,7 @@ BEGIN {extends 'Catalyst::Controller'; }
 
 use Encode;
 use FixMyStreet::Geocode;
+use Try::Tiny;
 use Utils;
 
 =head1 NAME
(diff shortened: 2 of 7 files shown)


zarino:
Sigh, just noticed there’s a typo in the commit message (`goecoder`).

And will obvs also need a changelog update.

dracos:
"I’m wondering whether we should have a test that checks the extra bit we added doesn’t break the site if you, say, don’t have an analytics.sqlite file in place?" - that'll already be covered somewhere, code coverage run shows it's already being run by something. I've made some small error tweaks to log error better.

dracos:
"I’m wondering whether we should have a test that checks the extra bit we added doesn’t break the site if you, say, don’t have an analytics.sqlite file in place?" - that'll already be covered somewhere, code coverage run shows it's already being run by something. I've made some small error tweaks to log error better, and as discussed added a URL field.

zarino:
Your changes look good to me @dracos!

Can you handle the merging, deploying, and placing of the suitably empty SQLite file in the data directory on the live server?

zarino:
mysociety/fixmystreet-commercial#1663 now includes documentation on how to query the database.

Issue #2905, opened 2020-03-03
Add a generic script for copying response templates


chrismytton:
We've had to copy response templates that clients have created on staging over to the live site a few times now. The scripts we've been using are currently in `heron:/home/bci/export-import-templates*`.

It would be good to "upgrade" these to a generic script for copying response templates that can be used for future sites and add them to this repo.

PR #2910, opened 2020-03-06
Add script to export/import body data.

dracos:
This covers response templates, roles, and staff users.
Needs tests.
Fixes https://github.com/mysociety/fixmystreet/issues/2905

### DIFF
--- a/bin/export-import-data
+++ b/bin/export-import-data
@@ -1,0 +1,153 @@
+#!/usr/bin/env perl
+
+use v5.14;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
+use FixMyStreet::DB;
+use Getopt::Long::Descriptive;
+use JSON::MaybeXS;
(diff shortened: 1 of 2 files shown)


bin/export-import-data
```diff
@@ -0,0 +1,153 @@
...
+    [ 'name=s', "Name of body" ],
+    [ 'import=s', "File to import" ],
+    [ 'commit', "Actually commit changes to the database" ],
+    [ 'help', "print usage message and exit", { shortcircuit => 1 } ],
+);
+print($usage->text), exit if $opt->help;
+$usage->die unless $opt->name;
+die "Please specify a file to import\n" if $opt->import && ! -e $opt->import || -d $opt->import;
+
+my $J = JSON::MaybeXS->new(utf8 => 1, pretty => 1);
```
struan: Also need to add something like

`use open ':std', ':encoding(UTF-8)';`

so that the output is UTF-8 encoded otherwise you get an error on import.

dracos: Hmm, I think that'll double decode it then, doesn't stop the wide character warning anyway. I changed the slurp_utf8 to just slurp, as JSON is doing the utf8 decoding for us here so don't need to do it twice.

PR #2918, opened 2020-03-11
Add Highways England junction lookup

chrismytton:
This is taken from the [sort-my-sign repo](https://github.com/mysociety/sort-my-sign). It allows the user to search for motorways names and junction numbers, e.g. "M60, Junction 2", and then geocodes that using a SQLite database created from Highways England markerpost locations.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1797

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
         - order unsent reports by confirmed date
     - Bugfixes
 	- Application user in Docker container can't install packages. #2914
+    - UK:
+        - Added junction lookup, so you can search for things like "M60, Junction 2"
 
 * v3.0 (4th March 2020)
     - Security:
--- a/perllib/HighwaysEngland.pm
+++ b/perllib/HighwaysEngland.pm
@@ -1,0 +1,39 @@
+package HighwaysEngland;
+
+use strict;
+use warnings;
+
+use FixMyStreet;
+use DBI;
+use Utils;
+
+sub database_file { FixMyStreet->path_to('../data/roads.sqlite') }
+
+sub junction_lookup {
+    my $s = shift;
+    if ($s =~ /^\s*(?<road>[AM][0-9MT]*)[\s,.]*(junction|junc|j)\s*(?<junction>.*?)\s*$/i
+          || $s =~ /^\s*(junction|junc|j)\s*(?<junction>.*?)[,.\s]*(?<road>[AM][0-9MT]*)\s*$/i
+    ) {
+        return _lookup_db($+{road}, 'junction', $+{junction}, 'name') || undef;
+    } elsif ($s =~ /^\s*(?<road>[AM][^ ]*)\s*(?<dist>[0-9.]+)\s*$/i
+          || $s =~ /^\s*(?<dist>[0-9.]+)\s*(?<road>[AM][^ ]*)\s*$/i
+    ) {
+        return _lookup_db($+{road}, 'sign', $+{dist}, 'distance') || undef;
+    }
+}
+
+sub _lookup_db {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
         - order unsent reports by confirmed date
     - Bugfixes
 	- Application user in Docker container can't install packages. #2914
+    - Added junction lookup, so you can search for things like "M60, Junction 2"
```
dracos: Looking at previous releases, I guess this should be under a "UK:" bullet point.

docs/install/manual-install.md
```diff
@@ -166,6 +166,24 @@ manually).
 $ bin/update-all-reports
 {% endhighlight %}
 
+#### Junction lookup
```
dracos: Thanks for adding this doc :) I wonder if this is the best place, it feels more like it goes under customising than installing to me, though not strongly (e.g. I imagine it won't apply to most people following the manual install doc, I mean). Perhaps under the "geocoder" part of customising?

perllib/HighwaysEngland.pm
```diff
@@ -0,0 +1,35 @@
...
+    } elsif ($s =~ /^\s*(?<road>[AM][^ ]*)\s*(?<dist>[0-9.]+)\s*$/i
+          || $s =~ /^\s*(?<dist>[0-9.]+)\s*(?<road>[AM][^ ]*)\s*$/i
+    ) {
+        return _lookup_db($+{road}, 'sign', $+{dist}, 'distance') || undef;
+    }
+}
+
+sub _lookup_db {
+    my ($road, $table, $thing, $thing_name) = @_;
+    my $db = DBI->connect("dbi:SQLite:dbname=".database_file, undef, undef) or return;
```
dracos: `database_file()`? as I got confused about it being some named constant for a second :)

perllib/HighwaysEngland.pm
```diff
@@ -0,0 +1,35 @@
+package HighwaysEngland;
+
+use strict;
```
dracos: Pedantically, needs a use FixMyStreet; use DBI; use Utils; as all are used by it.
(I guess ideally it wouldn't have FMS, something else would pass it in as a config, but meh, not needed now!)

t/app/controller/report_new.t
```diff
@@ -2185,4 +2185,33 @@ subtest "extra google analytics code displayed on email confirmation problem cre
...
+    if ($road eq 'M6' && $thing eq '11') {
+        return { latitude => 52.65866, longitude => -2.06447 };
+    } elsif ($road eq 'M5' && $thing eq '132.5') {
+        return { latitude => 51.5457, longitude => 2.57136 };
+    }
+});
+
+subtest 'junction lookup' => sub {
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => [ { fixmystreet => '.' } ],
```
dracos: This can just be ALLOWED_COBRANDS => 'fixmystreet',

t/app/controller/report_new.t
```diff
@@ -2185,4 +2185,33 @@ subtest "extra google analytics code displayed on email confirmation problem cre
     };
 };
 
+my $he = Test::MockModule->new('HighwaysEngland');
```
dracos: Tests are fine, but I don't think report_new.t is the right file (plus it's long enough!), it's testing the geocoder on the around page, so around.t I'd say? (Or if the test was lower level calling the function in UK.pm directly or HE directly, a uk cobrand file, or geocoder test file, etc)

PR #2919, opened 2020-03-11
[Northamptonshire] handle further investigation state

struan:
If an update has an external_status code of further we display 'Under
futher investigation' on the site. This is for acknowledging comments
from users.

[skip changelog]


chrismytton:
I think the test failure is related to https://github.com/mysociety/fixmystreet-commercial/issues/1796 (ugh!) so can be ignored.

perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -265,16 +265,18 @@ sub meta_line {
 };
 
 sub problem_state_processed {
-    my $self = shift;
+    my ( $self, $c ) = @_;
     return 'fixed - user' if $self->mark_fixed;
     return 'confirmed' if $self->mark_open;
-    return $self->problem_state;
+    my $cobrand_state = $c->cobrand->call_hook(problem_state_processed => $self);
```
dracos: Doesn't matter/could be later tidying, but just a note that you can access the current cobrand in any DB function call within a web request (and some cron) in `$self->result_source->schema->cobrand`, no need to pass around `$c`.

PR #2921, opened 2020-03-11
external status code of response template takes precedence

struan:
Order the response templates query by external status code so if you
have an update with an external status code the response template
matching that is always used. Previously with the database left to
determine the order it could pick a response template with a matching
state rather than a matching external status.


t/open311/getservicerequestupdates.t
```diff
@@ -1009,6 +1007,83 @@ foreach my $test ( {
...
+    title => "Acknowledgement 1",
+    text => "Thank you for your report. We will provide an update within 48 hours.",
+    auto_response => 1,
+    state => "in progress"
+});
+
+for my $test (
+    {
+        external_code => '090',
+        description => 'check numeric external status code in responset template override state',
```
dracos: typo in description, and one underneath too

perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -262,17 +262,22 @@ sub comment_text_for_request {
...
     my $template;
     if ($state_changed || $ext_code_changed) {
+        my $order;
         my $state_params = {
             'me.state' => $state
         };
         if ($ext_code) {
             $state_params->{'me.external_status_code'} = $ext_code;
+            # make sure that empty string comes last. Can't use DESC as that orders
+            # empty string before codes starting with 0 and explicit ASC will order
```
dracos: This sort works fine, but I don't understand the comment - DESC sorts "0" before "" as far as I can see?

```
bci=> select state,external_status_code,external_status_code is null from response_templates
where auto_response order by external_status_code desc nulls last;
...
                  | 0015                 | f
                  | 0010                 | f
                  | 0010                 | f
investigating     |                      | f
duplicate         |                      | f
action scheduled  |                      | f
```

struan: weird, I definitely had issues with this when I was testing. I will see if I can work out what's going on.

struan: hmmm, I have no idea why it was failing like that for me :|

perllib/Open311/GetServiceRequestUpdates.pm
```diff
@@ -262,17 +262,22 @@ sub comment_text_for_request {
...
+            # empty string before codes starting with characters
+            $order = { order_by => \"me.external_status_code NULLS LAST" };
```
dracos: You do need it to be DESC though :)

struan: doh!

Issue #2788, opened 2019-12-06
SendReport::Open311 should be able to save fields


dracos:
By default, SendReport::Open311 discards any changes made to the report during the sending process. In general, this is because it is adding/removing things needed only for that process (e.g. duplicating the title/description as an extra field, including a URL, removing an "urgent" question the backend does not need to see, etc), but sometimes it is fetching actual new info (e.g. a street's site code if not picked up client-side). Having a way to say store some data but discard others/the rest would be ideal.

PR #2923, opened 2020-03-12
[Open311] Allow save/drop of row changes during send.

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/2788


perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -97,8 +99,6 @@ sub open311_config {
             push @$extra, { name => 'UPRN', value => $ref };
         }
     }
-
-    $row->set_extra_fields(@$extra);
 }
```
struan: Should this not return undef or similar?

PR #2924, opened 2020-03-16
Sending dæmon

dracos:
This PR adds a daemon using Parallel::ForkManager. Then it does a noddy thing for updates, only locking per body, not per update, which seemed easier given the current code, but discussion there welcome too. The daemon will spin up some number of children each of which will just repeatedly look for reports(/updates) to send.

Todo:
- [x] More logging - print to stderr. Have USR1 cycle through logging levels
- [x] Improve fetching of reports as currently if N are being skipped/failing, it'll just keep trying to process those. And similar for updates, it'll just look at same body over and over :-/

### DIFF
--- a/bin/send-daemon
+++ b/bin/send-daemon
@@ -21,6 +21,7 @@ use FixMyStreet;
 use FixMyStreet::DB;
 use FixMyStreet::Script::Reports;
 use FixMyStreet::Queue::Item::Report;
+use Open311::PostServiceRequestUpdates;
 
 my ($opts, $usage) = describe_options(
     '%c %o',
@@ -76,6 +77,7 @@ while (!$exit) {
         while (!$exit) {
             $0 = "fmsd (running queue)";
             $db->txn_do(\&look_for_report);
+            $db->txn_do(\&look_for_update);
             $0 = "fmsd";
             sleep(5 + rand(10));
         }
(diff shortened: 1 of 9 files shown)


bin/send-daemon
```diff
@@ -0,0 +1,115 @@
...
+        send_comments => 1,
+    }, {
+        for => \'UPDATE SKIP LOCKED',
+        rows => 1,
+    } )->single or return;
+
+    my $cobrand = $body->get_cobrand_handler;
+    next if $cobrand && $cobrand->call_hook('open311_post_update_skip');
+
+    say "Sending updates for " . $body->name;
```
davea: This would benefit from `if $debug_mode` as above.

dracos:
Has basically been rewritten due to #2947

sagepe:
Just a final thought as I'm preparing the internal bits for deployment - I don't think we strictly need an `-example` file after all: the daemon configuration gets generated separately and this doesn't check for an example file.

Having the example file as-is will result in a copy of the unit file being processed and added to `conf/` but it's useful to have an example for re-users to refer to. If we wanted to avoid generating the copy in `conf/` we could rename the example to end in `.example` rather than `-example`.

dracos:
I've deployed this on staging and no file in conf/ was generated from the systemd file that I can see.

dracos:
Ooh it did on the second deploy (presumably checks what is already there in the checkout, hmm).

sagepe:
> I've deployed this on staging and no file in conf/ was generated from the systemd file that I can see.

I'll check what's happening here, but maybe we just switch to `.example` instead.

sagepe:
> I'll check what's happening here, but maybe we just switch to `.example` instead.

I think we should make this switch, the local template used on our deploy contains variables only available during the daemon generation phase, so the config file isn't needed later and isn't being generated properly.

PR #2926, opened 2020-03-16
[Highways England] Initial cobrand

chrismytton:
Work done so far on the Highways England cobrand.


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -98,9 +98,19 @@ sub munge_report_new_bodies {
     my %bodies = map { $_->name => 1 } values %$bodies;
     if ( $bodies{'TfL'} ) {
         # Presented categories vary if we're on/off a red route
-        my $tfl = FixMyStreet::Cobrand->get_class_for_moniker( 'tfl' )->new({ c => $self->{c} });
+        my $tfl = FixMyStreet::Cobrand::TfL->new({ c => $self->{c} });
```
davea: Does this also apply for the Highways England bit below? Even if not, probably worth being consistent or leaving a comment to explain the difference.

PR #2931, opened 2020-03-20
[fixmystreet.com] Survey banner

### LINKED ISSUE #2928: Add 2020 survey to FMS.com
We need to collect data for the FMS user research so we'd like to add a banner for the survey link to FMS.com to show to 40% of the site visitors. We have done similar to this on PMO sites and on WTT.

The survey is on SurveyGizmo already. @JenMysoc can provide link when necessary

zarino:
Big purple survey banner, which is hidden by default, and shown on 40% of pageloads, by JavaScript.

CTA button opens the surveygizmo survey in a new window/tab.

Banner can be hidden using the cross in the top right corner.

If the user clicks either the main CTA button, or the close button, a cookie is set that prevents the banner ever showing again.

Works on both the static pages, and the map pages. Here’s how it looks:

Screenshot_2020-03-20 FixMyStreet

Screenshot_2020-03-20 Pothole in cycle lane - Viewing a problem

Screen Shot 2020-03-20 at 14 33 15

Fixes #2928.

[skip changelog]

### DIFF
--- a/templates/web/fixmystreet.com/before_wrapper.html
+++ b/templates/web/fixmystreet.com/before_wrapper.html
@@ -1,0 +1,26 @@
+<div class="survey-banner">
+    <p>
+        <span class="survey-banner__text">Share your views and help us improve FixMyStreet for&nbsp;everyone</span>
+        <a class="survey-banner__cta" href="https://www.surveygizmo.com/s3/5506345/46e9932d0fbf" target="_blank">Take our short survey</a>
+    </p>
+    <button class="survey-banner__close">Close</button>
+</div>
+<script nonce="[% csp_nonce %]">
+(function(){
+    var live_site = location.hostname === 'www.fixmystreet.com';
+    var hasnt_opted_out = document.cookie.indexOf('__has_closed_march_2020_survey') === -1;
+    var randomly_chosen = Math.random() > 0.6;
+    if ( live_site && hasnt_opted_out && randomly_chosen ) {
+        document.body.className = ((document.body.className) ? document.body.className + ' has-survey-banner' : 'has-survey-banner');
+        document.querySelector('.survey-banner').style.display = 'block';
+    }
+    function hide_survey(){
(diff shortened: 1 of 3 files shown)


zarino:
Gah! Didn’t spot that it breaks the mobile reporting flow. Good catch.

I guess, ideally, the banner wouldn’t be shown on mobile apart from:

1. on a non-mappage page (like the homepage)
2. on a mappage _if you’ve just loaded it afresh_ (ie: not got there via the ajaxy behaviour)

Maybe we should add a close button too, reusing the same button/click/cookie stuff that the `top_banner--country` does in `front.js`.

As for the delay, yeah, it’s annoying. I’ve just thought, maybe we include the JavaScript as an inline `<script>` tag right after the element itself? @dracos will that just work? (I know we do some clever security stuff around inline script tags.)

dracos:
It will work as long as you use `<script nonce="[% csp_nonce %]">` so that it's allowed to run.

templates/web/fixmystreet.com/before_wrapper.html
```diff
@@ -0,0 +1,18 @@
+<div class="survey-banner">
+    <p>
+        <span class="survey-banner__text">Share your views and help us improve FixMyStreet for&nbsp;everyone</span>
+        <a class="survey-banner__cta" href="https://www.surveygizmo.com/s3/5506345/46e9932d0fbf" target="_blank">Take our short survey</a>
+    </p>
+    <button class="survey-banner__close">Close</button>
+</div>
+<script nonce="[% csp_nonce %]">
+if ( document.cookie.indexOf('has_closed_march_2020_survey') === -1 && Math.random() > 0 ) {
```
dracos: presumably not 0 :)
This will somehow need to know not to run in test because its appearance is breaking all the map-click tests, I'm afraid.

zarino:
@dracos I think this should be good to go now? If so, I’ll squash it and merge it.

PR #2934, opened 2020-03-25
[Bristol] Update URLs for assets on staging

chrismytton:
Note: This is only on staging currently. If the client is happy with it then we will need to change `perllib/FixMyStreet/Map/Bristol.pm` to use the new tiles on the live site.

Part of https://github.com/mysociety/fixmystreet-freshdesk/issues/138


chrismytton:
Bristol are happy with the map tiles, but still doing testing on the assets layer. I've broken out the map tile change into a separate PR here: https://github.com/mysociety/fixmystreet/pull/2938

PR #2936, opened 2020-03-26
Fixes the To header when sending emails about inactive accounts.

### LINKED ISSUE #2935: process-inactive-accounts sends emails with incorrect To header
When emailing inactive users because their accounts are about to be removed, the To header isn't set correctly (email address and the user's name are listed as separate recipients).

We started getting warnings about this from our cron job after upgrading to 3.0. It doesn't look like the relevant code in Inactive.pm has changed recently so I'm not sure why we're seeing warnings now, possibly because of upgraded perl modules?

Will create a pull request with a fix...

joesiltberg:
Without the brackets the user's name was interpreted as a separate
recipient.

Fixes #2935

### DIFF
--- a/perllib/FixMyStreet/Script/Inactive.pm
+++ b/perllib/FixMyStreet/Script/Inactive.pm
@@ -191,7 +191,7 @@ sub email_inactive_users {
                 user => $user,
                 url => $self->base_cobrand->base_url_with_lang . '/my',
             },
-            { To => [ $user->email, $user->name ] },
+            { To => [ [ $user->email, $user->name ] ] },
             undef, 0, $self->base_cobrand,
         );
 


perllib/FixMyStreet/Script/Inactive.pm
```diff
@@ -191,7 +191,7 @@ sub email_inactive_users {
                 user => $user,
                 url => $self->base_cobrand->base_url_with_lang . '/my',
             },
-            { To => [ $user->email, $user->name ] },
+            { To => [ [ $user->email, $user->name ] ] },
```
joesiltberg: Compare with for instance ArchiveOldEnquiries.pm

PR #2938, opened 2020-03-26
[Bristol] Update URL for map tiles

chrismytton:
As requested by the client, this updates the URL for Bristol's map tiles.

Part of https://github.com/mysociety/fixmystreet-freshdesk/issues/138


chrismytton:
Bah, test failure is because of https://github.com/mysociety/fixmystreet-commercial/issues/1796, so can be ignored (though I should really try and fix that soon, or delete that test).

PR #2945, opened 2020-04-03
Push Test

MoodAhmed:
Please check the following:

Thanks for contributing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
+LINK TO PROJECT BOARD: https://trello.com/b/M5aa7YDr/03-fix-mystreet-philly
+
+*Team Members:*
+Marcus Heimann	
+Alex Liu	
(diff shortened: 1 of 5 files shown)


dracos:
Hi, I don't think you wanted to open this PR? :)
Thanks for your interest in the FixMyStreet project. Please see our documentation on customising an installation at https://fixmystreet.org/customising/ - and specifically the parts on design https://fixmystreet.org/customising/css/ and templates https://fixmystreet.org/customising/templates/ . You probably do not want to edit base templates, unless you e.g. need to add a new template hook, which is doubtful. You should also not edit the example configuration file (our tests use it, for one thing), but have a copy called general.yml that you include what you need in. Otherwise you will find it harder to update the core software in future.
Hope that's helpful.

PR #2946, opened 2020-04-06
Removes documentation about restricting access to /admin

### LINKED ISSUE #2937: The documentation still says access to /admin should be restricted
We haven't started letting staff users into the admin interface yet, and still use htaccess to restrict access to /admin as the documentation advises (docs/running/index.md).

But I'm guessing this isn't still necessary? The privilege system in fixmystreet handles access to the admin interface now, right? 

Or is there still some reason left to restrict access with htaccess?

joesiltberg:
Fixes #2937


dracos:
Thanks!

PR #2950, opened 2020-04-07
[HE] message about being able to use report ids in search

struan:
[skip changelog]


templates/web/highwaysengland/around/_postcode_form_examples.html
```diff
@@ -1 +1,2 @@
 <p class="form-hint" id="pc-hint">[% tprintf('eg ‘%s’ or ‘%s’', c.cobrand.example_places) %]</p>
+<p class="form-hint">If you have previously made a report then please enter your reference, for example FMS1234567, into the search box to locate the latest status.</a>
```
dracos, line 2: Typo for </p> at the end.
Wonder if the confirmation page should more strongly suggest recording the reference number? Hohum.

PR #2955, opened 2020-04-09
[Peterborough] Improve extra question handling.

dracos:
Ignore (for Open311-sending purposes) any questions whose code starts
"PCC-", so that it is easier to use the admin to add extra questions,
and move the removal of the extra questions to a more normal location
so answers are not lost when extra fields is reset post-send.

For freshdesk 544 and 545.

[skip changelog]


dracos:
On staging.

PR #2959, opened 2020-04-15
Liu brancch

warbl:
Please check the following:

Thanks for contributing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
+LINK TO PROJECT BOARD: https://trello.com/b/M5aa7YDr/03-fix-mystreet-philly
+
+*Team Members:*
+Marcus Heimann	
+Alex Liu	
+Stephen Lucas	
+Mahmood	
+
+
 # Welcome to FixMyStreet Platform
 
 FixMyStreet Platform is an open source project to help people run websites for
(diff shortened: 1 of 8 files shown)


dracos:
[cf #2945]
Thanks for your interest in the FixMyStreet project. I can see you have started on your own cobrand templates, which is good :) As before, you almost certainly do not want to edit base templates files, the example configuration file (it breaks the tests), or the README, we can't accept the PR like that.

Some hopefully helpful comments on your cobrand .pm file:
* You should not use "Whitelabel" as a parent in your cobrand file, it is for lots of UK-specific code (as you can see, Whitelabel's own parent is "UKCouncils"); as you are just beginning, I recommend your parent should be Default.
* You do not need "council_url"; you probably don't need "council_area" either, though if your installation only covers the one body I guess it might work okay.
* You do not have a "Philadelphia" geocoder so you do not need get_geocoder; you should set parameters for e.g. the OSM geocoder using configuration options as explained in the documentation.
* "MasterMap" is a UK-specific map type, so you don't want that either (also MAP_TYPE can be set in general.yml, you do not need to set it in code); please get it working with the normal defaults first.
* You have set a path_to_pin_icons but have no pin images there, you do not need that function.
* I am not sure why you have set disable_resend_button or bypass_password_checks, neither of those should be necessary.

Hope that's helpful.

warbl:
Hello! I'm sorry for the general clumsiness in the changes and I will be
working on a forked repo from now on. A few classmates and I are working on
this for a school project. I was wondering if you could shed some light on
an issue we ran into in setting up the Cobrand:
[image: image.png]

Thanks!

dracos:
Hi, I'm afraid your image did not attach properly. Try leaving the update on fhe github PR instead of by email I guess.

warbl:
Made a pull request with the error

PR #2963, opened 2020-04-16
Add category staff restriction options

dracos:
Adds assigned_users_only option to categories and assigned_categories_only option to users.
Users with assigned_categories_only will only see staff features on a report page in their assigned categories.
Users will only see staff features on a report page in a category with assigned_users_only if it is in their assigned categories.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1857


dracos:
Ooh, ta, I'd checked all the templates for from_body/belongs_to_body/etc but missed 'permissions'. There's also moderation too, looking at that, and signing up alert as another user (probably not important).

This is only report pages, so I guess in list pages you will still see shortlist buttons, hmm.

dracos:
I've added a new commit (before the existing one, you'll see why) refactoring the report page permissions so that it does the lookup once at the start and then uses that permissions object everywhere in the template. Then if the assigned stuff applies, it blanks out that object, so that should fix it properly anywhere they're checked in the templates without me having to check I've got everywhere. And then added a commit to handle the around/reports list pages display of shortlist buttons. Which I think then covers it well enough?

dracos:
Ooh, thanks for spotting. It's only the full-page request that looks up categories (for the filter) and thus I hooked into those categories to find out which were assigned_users_only. The ajax request didn't (before) need to know a list of categories, it could just fetch the reports and return then, so bit annoying if we have to now look up the categories too, but I can't think of another solution, hmm.

dracos:
All reports page now always looks up body's categories on both ajax/non-ajax calls so that should fix it. I think I'm going to ignore the same issue on /around (doing it there would seem to involve a mapit call to then get the bodies to get their categories!) and see if it comes up.

Or I guess another solution, given presumably hardly used at all, is to just always fetch every category that has assigned_users_only set, and see if any apply... Meh.

PR #2964, opened 2020-04-16
[Bromley] Fix heatmap category filter.

dracos:
As it has switched to using category groups, the special subcategory code needs updating to update the correct variable. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -132,20 +132,25 @@ sub tweak_all_reports_map {
     # A place where this can happen
     return unless $c->action eq 'dashboard/heatmap';
 
-    # Bromley only subcategory stuff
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -134,18 +134,21 @@ sub tweak_all_reports_map {
...
-    my $filter = $c->stash->{filter_categories};
-    my @new_contacts;
-    foreach (@$filter) {
-        push @new_contacts, $_;
-        foreach (@{$subcats{$_->id}}) {
-            push @new_contacts, {
-                category => $_->{key},
-                category_display => (" " x 4) . $_->{name},
-            };
+    my $groups = $c->stash->{category_groups};
```
struan, line 137: I do wonder if a bit more of a comment here to explain this might be helpful as I was initially a bit confused till I went and looked at the Bromley config

Issue #2902, opened 2020-02-27
Allow inspectors to post a photo along with their update


chrismytton:
(From the FMS Pro Feature identification session in London, February 2020.)

Related to https://github.com/mysociety/fixmystreet/issues/2901, which is about making the inspector UI less confusing.

At the moment the inspector can upload a photo with a regular update, but they can't upload a photo through the form that's specific to inspectors.

davideaton2:
This is something TfL have asked for

chrismytton:
> This is something TfL have asked for

Forgive me if this is a silly question: Why can't their inspectors use the regular update form to upload a photo?

davideaton2:
I'm guessing because they are all logged in as inspectors and don't want to log back out to post updates with photos?

chrismytton:
Hmm, I must be misunderstanding something then, on the Oxfordshire version it seems like you can upload a photo as an inspector:

<img width="883" alt="image" src="https://user-images.githubusercontent.com/22996/78541367-6c879200-77ed-11ea-99f0-f84625637691.png">

dracos:
Yeah, as discussed on call last week, I think they literally don't realise they can do this now and this could be "sorted" (ignoring improving it in #2901) by just telling them how they can do this *now* if they want to.

Issue #2901, opened 2020-02-27
Improve inspector UI to make it less confusing


chrismytton:
> Improve inspector UI to make the dual forms (updates, and inspector) less confusing.

> It’s technically possible for Inspectors to add updates (using the normal update form) with a photo, but it’s not the same part of the form that they use to mark a report as inspected. So maybe that could be improved.

(From the FMS Pro Feature identification session in London, February 2020.)

chrismytton:
This issue is a bit vague at the moment. It came out of the discussions in London in February, but I'm afraid I don't remember any specifics.

@mysociety/fixmystreet What do we want to change about the inspector UI to make it less confusing? Do we want to combine the inspector form with the regular update form? Or perhaps separate the inspector form out onto its own page? Or is this simply about making the current interface a bit clearer? Feels like perhaps this should be a design ticket, if it's about changing the UI/UX?

As a reminder, this is the interface we're talking about, as seen on the Oxfordshire CC cobrand. The "inspector form" is the green column in the middle. The regular update form is on the left.

<img width="1680" alt="Screenshot 2020-04-02 at 16 34 58" src="https://user-images.githubusercontent.com/22996/78537968-31369480-77e8-11ea-9755-e5f530b8c22a.png">

dracos:
Feature-wise, the two forms, if you are a staff user who can see all the bits anyway:
* On the "public" update form (hidden behind "Provide an update" button), you can attach a photo, update text, response templates, change state, and leave update as yourself (though I think body name is actually displayed?), anonymous, or as another user.
* On the inspector form, you can change category, state, extra questions, internal note field, priority, "traffic management", and optionally leave public update text. All updates are left with name of the the body.

So assuming an inspector doesn't need to leave updates anonymously/as another user, you could say that if the ability to attach photos was added to the inspector form, then you could entirely hide the public update form from people who see the inspector form, and then there's no confusion about what form to fill in. (There are smaller versions of the inspector form that would need checking too e.g. if you only have the report_edit_category or _priority permission).

That would seem the easiest solution to me anyway at this point, in terms of something straightforward to do rather than consider larger scale changes?

PR #2965, opened 2020-04-17
Add photo field to inspector form

chrismytton:
This adds a photo upload field to the inspector form.

Fixes https://github.com/mysociety/fixmystreet/issues/2902
Related to https://github.com/mysociety/fixmystreet/issues/2901

Screenshot 2020-04-17 at 06 12 26


templates/web/base/report/inspect/public_update.html
```diff
@@ -7,8 +7,34 @@
               [% loc('Save with a public update') %]
             </label>
           </p>
+          <div id="public_update_form_fields">
+          [% IF c.cobrand.allow_photo_upload %]
```
dracos, line 11: This would seem a good time to factor out this whole IF statement (which also occurs identically in at least three other places) into its own template?

t/app/controller/report_inspect.t
```diff
@@ -589,7 +593,27 @@ FixMyStreet::override_config {
...
+        $report->update;
+        $mech->get_ok("/report/$report_id");
+        $mech->submit_form_ok({ button => 'save', with_fields => {
+            public_update => "This is a public update.", include_update => "1",
+            state => 'action scheduled',
+            photo1 => [ [ $sample_file, undef, Content_Type => 'image/jpeg' ], 1 ],
+        } });
+        $report->discard_changes;
+        my $comment = $report->comments(undef, { rows => 1, order_by => { -desc => "id" }})->first;
+        isnt $comment->photo, '', 'photo added to comment';
```
dracos, line 612: I think photo would be null if no photo was added, though, in which case this would pass (as only undef can match undef)? Probably best to check it has the hash you expect.

t/app/controller/report_inspect.t
```diff
@@ -589,7 +593,27 @@ FixMyStreet::override_config {
...
+        $mech->get_ok("/report/$report_id");
+        $mech->content_contains("/photo/c/" . $comment->id . ".0.jpeg");
```
dracos, line 614: ...though I guess this does show that it did get added :)

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -533,6 +533,13 @@ sub inspect : Private {
 
         $c->cobrand->call_hook(report_inspect_update_extra => $problem);
 
+        $c->forward('/photo/process_photo');
+        if ( my $photo_error  = delete $c->stash->{photo_error} ) {
+            $valid = 0;
+            $c->stash->{errors} ||= [];
```
dracos, line 539: I don't think this line is necessary (it will autovivify into an array if undefined), but I see a few other clauses above use it too, and some don't.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -560,6 +567,10 @@ sub inspect : Private {
                     anonymous => 0,
                     %update_params,
                 } );
+                if ( my $fileid = $c->stash->{upload_fileid} ) {
+                    $update->photo($fileid);
+                    $update->update;
```
dracos, line 572: I don't think this needs to be a separate db call, you could put it in the main add call, e.g. `photo => $c->stash->{upload_fileid} || undef,` or similar.

chrismytton:
While demo-ing the inspector tool on staging @dracos noticed that this change actually breaks the offline "save submitted form offline and submit to server when back online" functionality. Looks like that doesn't handle multipart forms correctly (it's assuming it's just a `foo=1&bar=2` string).

Also worth noting that `formData()` is not available in service workers in Safari.

davea:
The most recent two commits should sort the iOS issue 👍 @dracos do they look ok?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -660,6 +660,12 @@ $.extend(fixmystreet.set_up, {
   },
 
   dropzone: function($context) {
+    if ('Dropzone' in window) {
```
dracos, line 663: Ah, does this mean eb16dc0425f9f4bf89e0b5ad73bcbdad5a9a4b35 isn't needed any longer? Added in response to https://github.com/mysociety/fixmystreet/pull/2965#pullrequestreview-450683391 but if it's not loading the widget offline at all (fair enough), it presumably doesn't need to try and initialise it?

davea: That commit is still required as the bug was occurring during normal online usage not offline. This fixup commit (20a220d9b8ee177a9d9dc2df550e9c97ab5d1df7) disables the Dropzone widget when offline by reordering the `autoDiscover = true` code to be before `return` when `navigator.onLine == true`.

web/cobrands/fixmystreet/offline.js
```diff
@@ -302,6 +302,14 @@ fixmystreet.offline = (function() {
         $('.moderate-display.segmented-control, .shadow-wrap, #update_form, #report-cta, .mysoc-footer, .nav-wrapper').hide();
         $('.js-back-to-report-list').attr('href', '/my/planned');
 
+        // On iOS we want to hide the photo fields on the offline inspector
```
dracos, line 343: Pedantic, but I guess this should be in the following commit, as in this commit, offline support is just broken for everything because the changes in the following commit have not yet been made.

PR #2969, opened 2020-04-20
East sussex salesforce

struan:
Cobrand for back end only integration

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/EastSussex.pm
+++ b/perllib/FixMyStreet/Cobrand/EastSussex.pm
@@ -6,4 +6,36 @@ use warnings;
 
 sub council_area_id { return 2224; }
 
+sub open311_pre_send {
+    my ($self, $row, $open311) = @_;
+
+    my $contact = $row->category_row;
+    my $fields = $contact->get_extra_fields;
+    for my $field ( @$fields ) {
+        if ($field->{variable} && !$field->{automated}) {
+            my $text = $row->detail;
+            my $q = $row->get_extra_field_value( $field->{code} ) || '';
+            $text .= "\n\n" . $field->{description} . "\n" . $q;
+            $row->detail($text);
+        }
+    }
+}
+
+sub open311_post_send {
+    my ($self, $row, $h, $contact) = @_;
+
+    my $fields = $contact->get_extra_fields;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/EastSussex.pm
```diff
@@ -6,4 +6,37 @@ use warnings;
...
+            $text .= "\n\n" . $field->{description} . "\n" . $q;
+            $row->detail($text);
+        }
+    }
+}
+
+sub open311_post_send {
+    my ($self, $row, $h, $contact) = @_;
+
+    $contact = $row->category_row;
```
chrismytton, line 27: It's slightly confusing having `$contact` as an argument and then ignoring it but reusing the variable. Is there a reason for doing it like that?

web/cobrands/eastsussex/assets.js
```diff
@@ -0,0 +1,98 @@
...
+    read: function(json, type, filter) {
+        if (typeof json == "string") {
+            obj = OpenLayers.Format.JSON.prototype.read.apply(this,
+                                                              [json, filter]);
+        } else {
+            obj = json;
+        }
+
+      var results = [];
+      for (var i=0, len=obj.length; i<len; i++) {
```
chrismytton, line 18: Could use `$.each` here, but looking at the existing JS it seems that loops are done in both ways in lots of places. So not hugely important 🙂

web/cobrands/eastsussex/assets.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+
+OpenLayers.Format.EastSussex = OpenLayers.Class(OpenLayers.Format.JSON, {
+    read: function(json, type, filter) {
+        if (typeof json == "string") {
+            obj = OpenLayers.Format.JSON.prototype.read.apply(this,
+                                                              [json, filter]);
+        } else {
+            obj = json;
```
chrismytton, line 14: Doesn't look like this `obj` variable is defined with `var obj` anywhere?

struan:
This has been merged as of 13e9cbb95d6ba619c4046bbbdcd7e7cb7b0f9e67

PR #2970, opened 2020-04-20
Problem with setting up Cobrand

warbl:
image


dracos:
Please do not open a Pull Request simply to post an error message as an image. Pull Requests are for sending code changes, such as bug fixes or feature requests, or indeed new cobrands, upstream that have been tested and that are working successfully. For problems running the software, we have a mailing list https://fixmystreet.org/community/ which would be more appropriate.

However, this is not an issue with FixMyStreet, but an issue in your own code - Perl module files need to end with a true statement (ie. "1;"), see all the existing files. If you do not have some knowledge of Perl, I suggest you do not try and customise your cobrand much using a .pm file until you have got the site running without that. 

(Your .pm file still has things in it that make no sense - there is no Philadelphia geocoder, you have no cobrand-specific pin icons. Please do not try and run before you can walk, it makes things harder on yourselves.)

PR #2974, opened 2020-04-22
[TfL] Update homepage message

chrismytton:
Request from client to update the text on the homepage.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1851

image


chrismytton:
> Only minor point is four weeks from when?

Hmm yes, good point! Will mention that in my reply.

PR #2980, opened 2020-04-24
Disable staff private tickbox if category private.

### LINKED ISSUE #2961: Private categories, new report form should have private tickbox disabled
If a category is marked as private, then when a relevant staff user is making a new report in that category, it currently still shows the "Private" tickbox, defaulting to unticked, even though the setting of that is immaterial. It should probably show the field, preticked and disabled.

[Update https://mysocietysupport.freshdesk.com/a/tickets/553 when fixed]

dracos:
If a category is marked private, new reports are always marked private regardless of the setting of the private checkbox shown to relevant staff when making a new report. So tick/disable this box in that situation to prevent any confusion. Fixes #2961

### DIFF
--- a/templates/web/base/report/new/form_user_loggedin.html
+++ b/templates/web/base/report/new/form_user_loggedin.html
@@ -63,7 +63,9 @@
 
 [% IF c.user.has_permission_to("report_inspect", bodies_ids) OR c.user.has_permission_to("report_mark_private", bodies_ids) %]
     <div class="checkbox-group">
-        <input type="checkbox" name="non_public" id="form_non_public" value="1"[% ' checked' IF report.non_public %]>
+        <input type="checkbox" name="non_public" id="form_non_public" value="1"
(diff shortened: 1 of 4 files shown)


templates/web/base/report/new/form_user_loggedin.html
```diff
@@ -63,7 +63,9 @@
 
 [% IF c.user.has_permission_to("report_inspect", bodies_ids) OR c.user.has_permission_to("report_mark_private", bodies_ids) %]
     <div class="checkbox-group">
-        <input type="checkbox" name="non_public" id="form_non_public" value="1"[% ' checked' IF report.non_public %]>
+        <input type="checkbox" name="non_public" id="form_non_public" value="1"
+            [%~ ' checked' IF report.non_public OR non_public_categories.$category %]
+            [%~ ' disabled' IF non_public_categories.$category %]>
```
davea, line 68: I wonder if a tooltip explaining why this box is ticked would be helpful?

PR #2981, opened 2020-04-24
Disable category rename on Open311 categories.

### LINKED ISSUE #2957: Do not allow category rename on Open311-controlled categories
If a category is renamed in the admin, but is controlled by Open311 service fetching, it will get renamed back as soon as populate service list next runs. So in that scenario, we should not allow category renaming (it can cause an issue as e.g. it updates all matching report categories).

dracos:
If you rename an Open311 category, it will only get reinstated/deleted
by the Open311 populate script shortly after, and can cause issues due
to the problem category renaming that occurs. Fixes #2957.

### DIFF
--- a/templates/web/base/admin/bodies/contact-form.html
+++ b/templates/web/base/admin/bodies/contact-form.html
@@ -74,7 +74,8 @@
         <span class='form-hint'>[% loc('Use this where you do not want problem reporters to be able to reopen their fixed or closed reports when leaving an update.') %]</span>
     </p>
 
-    [% IF body.send_method == 'Open311' %]
+    [% IF contact.sent_by_open311 %]
+
       <p class="form-check">
           <input type="checkbox" name="open311_protect" value="1" id="open311_protect"[% ' checked' IF contact.get_extra_metadata('open311_protect') %]>
           <label for="open311_protect">[% loc("Protect this category's name and group(s) from Open311 changes") %]</label>
--- a/web/cobrands/fixmystreet/staff.js
+++ b/web/cobrands/fixmystreet/staff.js
@@ -335,6 +335,21 @@ fixmystreet.staff_set_up = {
     });
   },
 
+  open311_category_edit: function() {
+    var protect_input = document.getElementById('open311_protect');
+    if (!protect_input) {
+        return;
+    }
+    protect_input.addEventListener('change', function() {
+        var cat = document.getElementById('category');
+        cat.readOnly = !this.checked;
+        cat.required = this.checked;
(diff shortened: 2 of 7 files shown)


davea:
I think this functionality has some conflict/overlap with https://github.com/mysociety/fixmystreet/pull/2986 (my fault, I realise, for not reviewing this first!). In light of the new tickbox for protecting the name, does this PR need reworking? Some initial thoughts, it could maybe:
 a. Disable renaming unless the protection box is ticked; or
 b. Automatically enable protection if category is renamed; or
 c. Something else?

dracos:
I guess either would work; the issue that led me to this PR was that Open311 renamed it back and all the problems couldn't match up any more. (b) might be a bit magic though. I guess (a) would work along with a bit of JS to automatically make it editable as soon as it's ticked?

davea:
(a) sounds like the one to go for, agreed

web/cobrands/fixmystreet/staff.js
```diff
@@ -335,6 +335,17 @@ fixmystreet.staff_set_up = {
     });
   },
 
+  open311_category_edit: function() {
+    document.getElementById('open311_protect').addEventListener('change', function() {
```
davea, line 339: Yay for slowing reducing the jQuery dependency :) Sadly this will crash if `#open311_protect` doesn't exist on the page though.

web/cobrands/fixmystreet/staff.js
```diff
@@ -335,6 +335,17 @@ fixmystreet.staff_set_up = {
...
+        var cat = document.getElementById('category');
+        cat.readOnly = !this.checked;
+        cat.required = this.checked;
+        if (!this.checked) {
+            cat.value = cat.getAttribute('value');
```
davea, line 344: Neat way of restoring the original value 👍

dracos:
Thanks, hopefully added a server side check now, was slightly tricky!

templates/web/base/admin/bodies/contact-form.html
```diff
@@ -68,7 +68,7 @@
         <label for="updates_disallowed">[% loc('Disable updates on reports in this category') %]</label>
     </p>
 
-    [% IF body.send_method == 'Open311' %]
+    [% IF body.send_method == 'Open311' OR (body.can_be_devolved AND contact.send_method == 'Open311') %]
```
davea, line 71: Could `contact.sent_by_open311` be used here now? (Also, nice catch of something I'd not considered when adding the protection field!)

PR #2985, opened 2020-04-27
Respect SKIP_PACKAGES_INSTALL when not in Docker too

davea:
When working locally it’s not convenient to always run `install_packages` as part of `./script/update`, so this commit allows you to define the `SKIP_PACKAGES_INSTALL` environment variable to stop that.

[skip changelog]

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -5,16 +5,16 @@ cd "$(dirname "$0")/.."
 
 git submodule --quiet update --init --recursive --rebase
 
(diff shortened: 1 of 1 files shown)


sagepe:
If we're going to do this for all cases, maybe just move the check for `SKIP_PACKAGES_INSTALL` to before the Docker check, short-circuiting the `if` block entirely?

davea:
Duh, good idea. How's that?

PR #2986, opened 2020-04-27
Add contact flag to protect category/group names from Open311 changes

davea:
Adds a new tickbox to the contact editing form that protects the category
and group names from being changed by Open311.

image

For mysociety/fixmystreet-commercial#1776

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -272,6 +272,11 @@ sub update_contact : Private {
     } else {
         $contact->unset_extra_metadata( 'photo_required' );
     }
+    if ( $c->get_param('open311_protect') ) {
+        $contact->set_extra_metadata( open311_protect => 1 );
+    } else {
+        $contact->unset_extra_metadata( 'open311_protect' );
+    }
     if ( my @group = $c->get_param_list('group') ) {
         @group = grep { $_ } @group;
         if (scalar @group == 0) {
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -272,6 +272,11 @@ sub update_contact : Private {
     } else {
         $contact->unset_extra_metadata( 'photo_required' );
     }
+    if ( $c->get_param('open311_protect') ) {
+        $contact->set_extra_metadata_if_undefined( open311_protect => 1 );
```
dracos, line 276: Is this for the weird case where we've manually set open311_protect to 0 or something? Just think `set_extra_metadata` on its own might be more obvious (I realise it's copying photo_required above). Perhaps I've missed something.

davea: Pure cargo culting. Will remove `_if_undefined` :)

davea:
Good question about `non_public`. The [existing code](https://github.com/mysociety/fixmystreet/blob/813d8a0273764b6215e739f268dc7243d768a205/perllib/Open311/PopulateServiceList.pm#L313-L314) already prevents a private category being made public which is the biggest risk. I can't think of a situation where Open311 would want the category to be private and FMS would want it to be public. My gut feeling is that the protection of the name and the protection of the `non_public` flag are two slightly separate things so shouldn't be lumped under a single tickbox, happy to be persuaded otherwise though.

PR #2998, opened 2020-04-29
[East Sussex] FIx issue sending notices, tidy up detail reset.

dracos:
Fix issue sending notices. Store the original detail so we can reset back to it easily. (Also use a function not also used by update sending, in case that's used in future and it would break.) [skip changelog]

### DIFF
--- a/t/cobrand/eastsussex.t
+++ b/t/cobrand/eastsussex.t
@@ -1,0 +1,42 @@
+use FixMyStreet::TestMech;
+use FixMyStreet::Script::Reports;
+use CGI::Simple;
+
+my $mech = FixMyStreet::TestMech->new;
+
+my $body = $mech->create_body_ok(2224, 'East Sussex Council',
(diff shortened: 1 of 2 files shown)


dracos:
Added some tests (+0.12% woo), which happily also showed that the current code isn't working correctly :)

t/cobrand/eastsussex.t
```diff
@@ -0,0 +1,38 @@
...
+
+    $p->discard_changes;
+    ok $p->whensent, 'Report marked as sent';
+    is $p->send_method_used, 'Open311', 'Report sent via Open311';
+    is $p->external_id, 248, 'Report has right external ID';
+
+    my $req = $test_data->{test_req_used};
+    my $c = CGI::Simple->new($req->content);
+    like $c->param('description'), qr/Is it urgent\?\r\nno/, 'Correct description';
+    unlike $c->param('description'), qr/This is a notice/, 'No notice text';
```
struan, line 35: Should this not also test that the extra information has not been saved?

PR #3000, opened 2020-04-29
Makes sure cached recent reports are displayed in order.

### LINKED ISSUE #2999: Recent reports on front page are not always ordered
**Describe the bug**
When the recent reports are fetched from the database they're ordered by confirmed date, but when they're found in memcached they're unordered.

**To Reproduce**
Typically it's enough to just view the front page and note the timestamps of the recent reports. If the cache timeout had expired however they'll be correct.

**Expected behavior**
The order should be the same regardless of whether the cache is used or not.

Will supply PR with suggested fix shortly...

joesiltberg:
Fixes #2999

### DIFF
--- a/perllib/FixMyStreet/DB/ResultSet/Problem.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Problem.pm
@@ -159,7 +159,7 @@ sub _recent {
             $probs = [ $rs->search({
                 id => [ map { $_->id } @$probs ],
                 %$query,
-            })->all ];
+            }, $attrs)->all ];
         } else {
             $probs = [ $rs->search( $query, $attrs )->all ];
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -159,7 +159,7 @@ sub _recent {
             $probs = [ $rs->search({
                 id => [ map { $_->id } @$probs ],
                 %$query,
-            })->all ];
+            }, $attrs)->all ];
```
joesiltberg, line 162: I don't see a reason why this search shouldn't include $attrs in the same way as the search below does.

Sorry if I'm missing something...

joesiltberg:
Woohoo #3000 ! 🥳🍾🎉

:-)

PR #3006, opened 2020-05-01
Simplify #geolocate_link styling

zarino:
Fixes mysociety/fixmystreet-commercial#1835.

Reduces the amount of styling on the `#geolocate_link` in core, because we’re seeing more cobrands now (eg: bexley, bristol, bucks, lincs) where the very opinionated "dark grey button with rounded bottom corners" styling just didn’t fit and had to be overwritten, separately, again and again.

Rather than keep the grey button styling in just a few cobrands, I‘ve removed it everywhere. Also took the opportunity to increase the font size, so it’s a bit more tappable on touchscreens.

Before, and after, at narrow widths:

fmscom

island-roads

highways

While I was at it, I also reduced the specificity of all the `#geolocate` selectors (in core and the cobrands). No need for them to be as specific as they were.

### DIFF
--- a/web/cobrands/cheshireeast/layout.scss
+++ b/web/cobrands/cheshireeast/layout.scss
@@ -31,12 +31,7 @@ body.frontpage .content {
 }
 
 a#geolocate_link {
-    color: #2e3191; // override default `#front-main a`
     margin-top: 1em;
-
-    &:hover {
-        text-decoration: none;
-    }
 }
 
 .ce-footer {
--- a/web/cobrands/cheshireeast/layout.scss
+++ b/web/cobrands/cheshireeast/layout.scss
@@ -29,8 +29,11 @@ body.frontpage .content {
     margin: 0;
     width: 30em;
 }
-#front-main a#geolocate_link {
-    color: #2e3191;
+
+a#geolocate_link {
+    color: #2e3191; // override default `#front-main a`
+    margin-top: 1em;
+
     &:hover {
         text-decoration: none;
     }
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -11,6 +11,9 @@ $link-visited-color: $link-color !default;
 $link-text-decoration: none !default;
 $link-hover-text-decoration: underline !default;
 
+$primary_link_decoration: underline !default;
+$primary_link_hover_decoration: $primary_link_decoration !default;
+
 $itemlist_item_background: #f6f6f6 !default;
 $itemlist_item_background_hover: #e6e6e6 !default;
 $col_big_numbers: #666 !default;
@@ -2371,6 +2374,12 @@ label .muted {
           background:#333;
         }
       }
+    }
+  }
+  a {
+    text-decoration: $primary_link_decoration;
+    &:hover {
+      text-decoration: $primary_link_hover_decoration;
     }
   }
 }
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -2373,26 +2373,13 @@ label .muted {
       }
     }
   }
-  a#geolocate_link {
-    @include inline-block;
-    vertical-align:top;
-    background:#1a1a1a;
-    color:#C8C8C8;
-    padding:0.5em;
-    font-family: $meta-font;
-    font-size: 0.8125em;
-    @include border-radius(0 0 0.25em 0.25em);
-    &:hover {
-      text-decoration:none;
-      background:#2a2a2a;
-    }
-  }
-  a#geolocate_link.loading {
-      background: #1a1a1a url("/cobrands/fixmystreet/images/spinner-black.gif") flip(100%,0) 50% no-repeat;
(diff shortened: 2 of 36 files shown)


dracos:
Do you think it's obvious enough as a link? Feels like use my location is more a button-y thing than a link-y thing maybe and as there are no other links in view I wouldn't even realise it was a link. Dunno.

Why so many `override default `#front-main a` - why is there a default colour (inherit) set for that at all?

zarino:
> Do you think it's obvious enough as a link? Feels like use my location is more a button-y thing than a link-y thing maybe and as there are no other links in view I wouldn't even realise it was a link. Dunno.

Yeah, I’m not sure this will be the "final" style for this button, but at least this PR makes the button easier to style in future. I think the link styling will do for now, while we have discussions about all the other changes we want to make to this form. Happy to be persuaded otherwise though, if you feel strongly against it.

> Why so many override default #front-main a` - why is there a default colour (inherit) set for that at all?

A good question. I assume it was there because some cobrand(s) at some point wanted to include links above the search box, and the default blue link colour didn’t have good enough contrast with the #front-main background colour. My instinct is we could just remove it, since it seems to be causing more issues than it solves right now. There’s a ticket for "considering" what to do about it: https://github.com/mysociety/fixmystreet/issues/3007.

dracos:
Could we underline it at least? On mobile, I'm just not sure how you'd tell the difference between it and any other text (not saying the old one was brilliant or anything, but it does at least usually look a bit like the main form "Go" button...).

Sorry, had missed #3007.

I note when you click the geolocation link, before it finishes, it gets a black background inherited from base, which would be good to fix as part of this!

web/cobrands/sass/_base.scss
```diff
@@ -2373,26 +2373,13 @@ label .muted {
...
-    font-family: $meta-font;
-    font-size: 0.8125em;
-    @include border-radius(0 0 0.25em 0.25em);
-    &:hover {
-      text-decoration:none;
-      background:#2a2a2a;
-    }
-  }
-  a#geolocate_link.loading {
-      background: #1a1a1a url("/cobrands/fixmystreet/images/spinner-black.gif") flip(100%,0) 50% no-repeat;
```
dracos, line 2391: Can remove the GIF, I think this was its only use.

zarino: I’m wondering whether to just leave it there, in case a future cobrand wants to use it? Given we’ve already got half a dozen other `spinner-*.gif` colour options there. Or is that just messy?

dracos: Meh, don't mind.

web/cobrands/sass/_layout.scss
```diff
@@ -13,6 +13,11 @@ $header-top-border: $header-top-border-width solid $primary !default;
 
 $container-max-width: 60em !default;
 
+$primary_link_color: null !default;
+$primary_link_decoration: null !default;
```
dracos, line 17: Do you need to define the two _decoration ones in here, they're not used

web/cobrands/tfl/layout.scss
```diff
@@ -66,12 +66,6 @@ h1 {
   #postcodeForm div {
     margin: 0;
   }
-  a#geolocate_link {
-    color: $beck-blue;
-    font-family: $heading-font;
-    text-decoration: underline;
```
dracos, line 72: TfL used to be underlined.

web/cobrands/cheshireeast/layout.scss
```diff
@@ -29,11 +29,10 @@ body.frontpage .content {
...
 }
-#front-main a#geolocate_link {
-    color: #2e3191;
-    &:hover {
-        text-decoration: none;
-    }
+
+a#geolocate_link {
+    margin-top: 1em;
+    text-decoration: none; // cheshireeast uses border-bottom for links
```
dracos, line 35: Isn't this covered by the default now then?

web/cobrands/sass/_layout.scss
zarino: Removed now! Thanks!

web/cobrands/tfl/layout.scss
zarino: Fixed now! Thanks!

zarino:
Ok @dracos:

* Underlined links in #front-main is now the default, with `cheshireeast` and `oxfordshire` overriding `$primary_link_decoration` to `none` – they can’t use `null` because then the `underline !default` from `base.scss` would get applied.
* Links now have an underline in `tfl` :+1:
* `$primary_link_hover_decoration` is now set to the value of `$primary_link_decoration` by default (not a big change, but seemed like it would save people some hassle if they, like cheshireeast, want the links to _always_ be undecorated).
* Un-used `$primary_link_decoration` and `$primary_link_hover_decoration` removed from `_layout.scss`.
* Issue numbers now in Changelog.
* I haven’t squashed the two commits because they still seem fairly well self-contained to me, and the first commit isn’t _broken_ or anything without the second one.

Issue #2991, opened 2020-04-28
Disallow updates on certain categories


dracos:
Currently the default is to disallow updates on any report that has its `closed_updates` flag set, otherwise allowed. In the UK there are feature flags to allow updates e.g only on open reports, staff only, etc.

Add a tickbox in the category admin interface that when selected disallows updates on reports in that category. Question as to whether separate page to view/set option for all categories at once, but for now add to category admin.

Issue #2989, opened 2020-04-28
Enable per category anonymous reporting to be configured in admin


struan:
This is present in some cobrands in code but it would be easier if the config was in the admin.

This should be a tickbox in admin interface for a category that when selected allows users to make fully anonymous reports on that category.

chrismytton:
Looks like this was done in https://github.com/mysociety/fixmystreet/pull/3028

PR #3012, opened 2020-05-01
Add way to disallow report reopening on reports in a category

### LINKED ISSUE #2992: Allow reopening only on certain categories
In general, the problem reporter can reopen a fixed or closed report.
Add an admin option to disallow reopening for that category.
Either a tickbox in the category admin, or its own "which categories allow reopening" page? (with info then displayed on category admin page).

dracos:
Fixes #2992. Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1872.
Does need a doc update but I will do that at same time as #2991 doc given it'll be same screenshot, same place. In fact, might as well wait for #2989 to only do the screenshot once!

Includes same first commit as mentioned in #3011

### DIFF
--- a/templates/web/base/report/update/form_state_checkbox.html
+++ b/templates/web/base/report/update/form_state_checkbox.html
@@ -1,5 +1,7 @@
 [% IF (problem.is_fixed OR problem.is_closed) AND ((c.user_exists AND c.user.id == problem.user_id) OR alert_to_reporter) %]
 
+    [% RETURN IF c.cobrand.reopening_disallowed(problem) ~%]
+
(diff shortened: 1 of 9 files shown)


templates/web/base/report/update/form_state_checkbox.html
```diff
@@ -1,5 +1,7 @@
 [% IF (problem.is_fixed OR problem.is_closed) AND ((c.user_exists AND c.user.id == problem.user_id) OR alert_to_reporter) %]
 
+    [% RETURN IF c.cobrand.reopening_disallowed(problem) ~%]
```
davea, line 3: Should this change be included in the `fixmystreet.com` version of this template or is it a deliberate thing to limit that flag to cobrands?

PR #3014, opened 2020-05-04
[BANES] Add script to close redundant reports

chrismytton:
B&NES have sent us a spreadsheet with a list of reports along with their
IDs that they want marking as "No further action" along with a specific
update.

This script takes a comma-separated list of report IDs and closes those
reports, leaving the appropriate comment text.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1867

### DIFF
--- a/bin/fixmystreet.com/banes-close-reports
+++ b/bin/fixmystreet.com/banes-close-reports
@@ -1,0 +1,69 @@
+#!/usr/bin/env perl
+#
+# B&NES have sent a spreadsheet of reports that they want closing. This script
+# accepts a comma separated list of report IDs, which will be extracted from
+# the spreadsheet, marks those reports as "No further action" and leaves an
+# update on them.
+
+use utf8;
+use v5.14;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use constant COUNCIL_NAME => 'Bath and North East Somerset Council';
+use constant CLOSURE_TEXT => <<'CLOSURE_TEXT';
+We’re sorry we've been unable to respond to your report.
+
+We’ve made a few changes recently to improve the way we respond to reports. We’re closing down any reports we think are now out of date, including this one, so that these changes can take effect.
+
+If you think this issue still requires attention, please report it again.
+CLOSURE_TEXT
+
+use FixMyStreet::DB;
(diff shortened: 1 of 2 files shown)


bin/fixmystreet.com/banes-close-reports
```diff
@@ -0,0 +1,66 @@
...
+CLOSURE_TEXT
+
+use FixMyStreet::DB;
+use FixMyStreet::Script::ArchiveOldEnquiries;
+use Getopt::Long::Descriptive;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['report-ids=s', "comma-separated list of report IDs to close", { required => 1 } ],
+    ['commit|c', "actually close reports and send emails. Omitting this flag will do a dry-run"],
```
davea, line 35: Technically the archive script isn't sending emails (though they may be sent later as alerts on the problems that got updated) so maybe `add comments` instead of `send emails` here.

bin/fixmystreet.com/banes-close-reports
```diff
@@ -0,0 +1,66 @@
...
+    user => $body->comment_user->id,
+    user_name => $body->comment_user->name,
+    closure_text => CLOSURE_TEXT,
+    retain_alerts => 1,
+    commit => $opts->commit,
+    closed_state => 'unable to fix',
+});
+
+if ($opts->{commit}) {
+    printf("Closing %d old reports: ", $q->count);
```
davea, line 61: Is it worth a sanity check here that bails out if the results from the DB differ in number to the IDs from the command line?

chrismytton: I'm only getting back reports in the open state, in case any of the listed reports are already closed, so it's expected that the number of results might not match the number of arguments.

I have added a sanity check to ensure we're not getting _more_ reports back than we have report IDs. If that happens then something probably has gone quite wrong!

dracos:
This has broken master tests. The tests never passed on this branch. Please do not merge until the tests have passed.

perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -11,6 +11,7 @@ use FixMyStreet::Email;
 
 my $opts = {
     commit => 0,
+    close_state => 'closed',
```
dracos, line 14: typo for closed_state.

dracos:
All changes made after a review need to be re-reviewed before merge.

chrismytton:
> This has broken master tests. The tests never passed on this branch. Please do not merge until the tests have passed.

Oops, sorry.

PR #3024, opened 2020-05-07
display user detail for contributed_by in report_edit

struan:
Include name, email and body rather than just user id.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -234,6 +234,26 @@ sub edit : Path('/admin/report_edit') : Args(1) {
             push @fields, { name => 'Duplicates', val => join( ',', @{ $problem->get_extra_metadata('duplicates') } ) };
             delete $extra->{duplicates};
         }
+
+        if ( $extra->{contributed_by} ) {
+            my $u = $c->cobrand->users->find({id => $extra->{contributed_by}});
+            if ( $u ) {
+                my $uri = $c->uri_for_action('admin/users/index', { search => $u->email } );
+                push @fields, {
+                    name => _('Created By'),
+                    val => FixMyStreet::Template::SafeString->new( "<a href=\"$uri\">@{[$u->name]} (@{[$u->email]})</a>" )
+                };
+                if ( $u->from_body ) {
+                    push @fields, { name => _('Created Body'), val => $u->from_body->name };
+                } elsif ( $u->is_superuser ) {
+                    push @fields, { name => _('Created Body'), val => _('Superuser') };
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -234,6 +234,18 @@ sub edit : Path('/admin/report_edit') : Args(1) {
             push @fields, { name => 'Duplicates', val => join( ',', @{ $problem->get_extra_metadata('duplicates') } ) };
             delete $extra->{duplicates};
         }
+
+        if ( $extra->{contributed_by} ) {
+            my $u = $c->cobrand->users->find({id => $extra->{contributed_by}});
```
davea, line 239: If `$u` isn't found then this causes a crash on the next line. This is unlikely to happen, but is possible if e.g. a superuser contributes a report and then tries to view the report admin page on the cobrand site. Because `UKCouncils::users_restriction` doesn't include superusers it won't be found.

perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -234,6 +234,18 @@ sub edit : Path('/admin/report_edit') : Args(1) {
...
+            push @fields, { name => 'Created By', val => "@{[$u->name]} (@{[$u->email]})" };
+            if ( $u->from_body ) {
+                push @fields, { name => 'Created Body', val => $u->from_body->name };
+            } elsif ( $u->is_superuser ) {
+                push @fields, { name => 'Created Body', val => 'SuperUser' };
```
davea, line 244: Nitpick, sorry, but we use "Superuser" elsewhere.

perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -234,6 +234,18 @@ sub edit : Path('/admin/report_edit') : Args(1) {
             push @fields, { name => 'Duplicates', val => join( ',', @{ $problem->get_extra_metadata('duplicates') } ) };
             delete $extra->{duplicates};
         }
+
+        if ( $extra->{contributed_by} ) {
+            my $u = $c->cobrand->users->find({id => $extra->{contributed_by}});
+            push @fields, { name => 'Created By', val => "@{[$u->name]} (@{[$u->email]})" };
```
davea, line 240: The HTML escaping might make this more trouble than it's worth, but could this link to the the admin user search results for their email to make it easier to find their other reports or view their timeline for example?

dracos: The descriptions also need marking for translation in some way.

PR #3025, opened 2020-05-07
Blocks /mapit from web crawlers

joesiltberg:
I noticed that https://global.mapit.mysociety.org/ disallows all web crawling, but I guess bots will still go through fixmystreet.com/mapit?

Perhaps there's a good reason to allow that?

### DIFF
--- a/web/robots.txt
+++ b/web/robots.txt
@@ -3,3 +3,6 @@ Crawl-delay: 3
 
 User-agent: bingbot
 Crawl-delay: 3
+
+User-agent: *
+Disallow: /mapit


dracos:
global.mapit doesn't have a robots.txt file, which means anything allowed as far as I know? Our .com proxy only allows proxying of the URLs needed (ie. area polygons). Nowadays (as in for some years now) the proxy is probably entirely unnecessary and we could call directly with CORS and no proxy at all, so perhaps that is a better solution.

joesiltberg:
When I get https://global.mapit.mysociety.org/robots.txt this is what it says:

```
# robots.txt served from varnish
User-agent: *
Disallow: /
```

Are the area polygons needed for proper indexing by the search engines?

dracos:
Weird, I got a 404 a couple of times, then that. Wonder what's going on there. I don't think that should be being served.

Sorry, I don't understand your question. The area polygons are used solely to display the boundary on a body's All Reports page, they're entirely optional. And like I say, were only served via a proxy because old browsers could not perform cross-site requests for data; pretty sure they all can now do so.

joesiltberg:
Oh ok. Sorry if I was unclear but I think you answered the question :-)

I was looking through our access logs in order to get an idea of how many requests we get, in case we want to switch to mySociety's instance at some point. A large part of our requests now are from bots so I'm thinking that would influence our price.

Maybe that's something to consider before switching to CORS, if it would affect your billing for the service?

Closing this for now, thanks for your help!

dracos:
Hmm, yes, hadn't thought about that side of things! Perhaps will leave it alone and continue to proxy, then - but to be clear, we proxy in the web server directly, not using the FMS code proxy, that proxy should probably also be restricted to the area geojson lookup given that's all it needs, one for the future.

PR #3028, opened 2020-05-07
admin interface for per category anonymous reporting

struan:
Add an interface to enable a category to accept anonymous reports, plus
the code to handle permitting this.

It's only available on single body cobrand sites in the default
configuration.

### DIFF
--- a/templates/web/base/admin/bodies/contact-form.html
+++ b/templates/web/base/admin/bodies/contact-form.html
@@ -74,8 +74,14 @@
         <span class='form-hint'>[% loc('Use this where you do not want problem reporters to be able to reopen their fixed or closed reports when leaving an update.') %]</span>
     </p>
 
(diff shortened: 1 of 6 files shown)


templates/web/base/admin/bodies/contact-form.html
```diff
@@ -74,6 +74,11 @@
         <span class='form-hint'>[% loc('Use this where you do not want problem reporters to be able to reopen their fixed or closed reports when leaving an update.') %]</span>
     </p>
 
+    <p class="form-check">
```
davea, line 77: It might be an idea to hide or disable this field unless the body's cobrand has `anonymous_account` set, otherwise clicking the 'report anonymously' button when making a report results in a "please enter your email" error.

struan: yes, good catch.

PR #3034, opened 2020-05-14
reset asset select message by id when asset unselected

struan:
Resetting using the class could cause an incorrect "you can pick an
$item from the map" message to be display when asset_groups were being
used. This was happening when the unselect event was firing after the
category change event and then updating the new message using the
message from the unselected layer.

Fixes mysociety/fixmystreet-commercial#1888

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -833,7 +833,19 @@ fixmystreet.assets = {
     named_select_action_not_found: function() {
         var message = this.fixmystreet.asset_item_message;
         message = message.replace('ITEM', this.fixmystreet.asset_item);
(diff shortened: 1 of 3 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -833,7 +833,20 @@ fixmystreet.assets = {
     named_select_action_not_found: function() {
         var message = this.fixmystreet.asset_item_message;
         message = message.replace('ITEM', this.fixmystreet.asset_item);
-        $('.category_meta_message').html(message);
+        var $p = $('.category_meta_message');
```
dracos, line 836: This line isn't needed, if you `var $p` the use in your if.

PR #3035, opened 2020-05-15
[Oxfordshire] Add instruct defect functionality back in

chrismytton:
In the past Oxfordshire had a way for inspectors to "instruct a defect", which generated a CSV which got emailed to Oxfordshire with a list of defects.

They now want to bring this functionality back, but instead of emailing a CSV they want it to call an API.

As a first step towards that goal, this adds back some of the instruct defect functionality that was removed in 68e18ff.

Part of https://github.com/mysociety/fixmystreet-commercial/issues/1846

image

### DIFF
--- a/bin/browser-tests
+++ b/bin/browser-tests
@@ -11,7 +11,19 @@ my ($cobrand, $coords, $area_id, $name, $mapit_url, $coverage);
 
 BEGIN {
     $config_file = 'conf/general.yml-example';
-    $cobrand = [ 'borsetshire', 'fixmystreet', 'northamptonshire', 'bathnes', 'buckinghamshire', 'hounslow', 'isleofwight', 'peterborough', 'tfl', 'hackney' ];
+    $cobrand = [qw(
+        bathnes
+        borsetshire
+        buckinghamshire
+        fixmystreet
+        hackney
+        hounslow
+        isleofwight
+        northamptonshire
+        oxfordshire
+        peterborough
+        tfl
+    )];
     $coords = '51.532851,-2.284277';
     $area_id = 2608;
     $name = 'Borsetshire';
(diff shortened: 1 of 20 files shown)


bin/browser-tests
```diff
@@ -11,7 +11,18 @@ my ($cobrand, $coords, $area_id, $name, $mapit_url);
...
+        borsetshire
+        fixmystreet
+        northamptonshire
+        bathnes
+        buckinghamshire
+        hounslow
+        isleofwight
+        peterborough
+        tfl
+        oxfordshire
```
dracos, line 24: Yes, certainly worth doing this now, but perhaps alphabetise it at the same time? :)

templates/web/base/report/_inspect.html
```diff
@@ -69,6 +80,15 @@ <h2 class="inspect-form-heading">[% loc('Inspect report') %]</h2>
           [% INCLUDE 'report/inspect/public_update.html' %]
         [% END %]
 
+        [% IF problem.get_extra_metadata('inspected') %]
```
dracos, line 83: Could we get this in Oxfordshire template only? Just worried that it might confuse current users (in case they e.g. have "all permissions") to see this when nothing would be happening for them.

templates/web/base/report/_inspect.html
```diff
@@ -33,6 +33,17 @@ <h2 class="inspect-form-heading">[% loc('Inspect report') %]</h2>
             <label for="state">[% loc('State') %]</label>
             [% INCLUDE 'report/inspect/state_groups_select.html' %]
           </p>
+          [% IF permissions.report_instruct AND NOT problem.get_extra_metadata('inspected') %]
```
dracos, line 36: ditto the below I guess, given it wouldn't be doing this for anyone else

t/app/controller/report_inspect.t
```diff
@@ -225,22 +225,27 @@ FixMyStreet::override_config {
...
         $user->update;
     };
 
     subtest "test update is required when instructing" => sub {
+        $report->unset_extra_metadata('inspected');
         $report->update;
+        $report->inspection_log_entry->delete;
         $report->comments->delete_all;
         $mech->get_ok("/report/$report_id");
         $mech->submit_form_ok({ button => 'save', with_fields => { public_update => undef, include_update => "1" } });
```
dracos, line 244: Should this line have raise_defect => 1 in it? Otherwise it doesn't appear to be testing it.

chrismytton: Have added a separate test to check that public update is required when raising a defect. I've also given this test a clearer title, because it doesn't actually have anything to do with instructing.

PR #3041, opened 2020-05-21
[fms.com] add toggle between aerial and roads maps

struan:
Adds a button to the map controls on FMS.com that toggles between the roads and aerial views:

image

aerial

aerial_mobile

[skip changelog]


struan:
See https://github.com/mysociety/fixmystreet-commercial/issues/1870 for details

dracos:
Reviewed; then made some changes to bring the Bing URLs up to date, add attribution for the aerial layer, move it all to the base Bing so it can have this as well, and got it so the toggle works on report pages. @struan Could you review my commits?

PR #3045, opened 2020-05-27
Mark user as active when sent an email alert

dracos:
If a user is sent email alerts they've signed up for, but otherwise doesn't log in to the site, we probably don't want to delete their account as 'inactive'. We could/would send them the "about to be deleted" warning email, but that could be confusing because their account isn't "inactive", as dur, they're getting email alerts. (If an email alert bounces, they'll be unsubscribed and then not get emails and account will get deleted as normal.) So when an alert is sent, take that as activity on the user's account and update the last active time.

Also update .com privacy page to say that is what we are doing. Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Releases
 
 * Unreleased
+    - Changes:
+        - Mark user as active when sent an email alert.
     - Bugfixes:
         - Fix issue with dashboard report CSV export. #3026
         - bin/update-schema PostgreSQL 12 compatibility. #3043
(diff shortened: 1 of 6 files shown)


dracos:
In terms of running this on .com, plan is:
1. Manually update people's last active time to match their last email alert sent time (if newer).
2. Manually set anyone active in the last 23-24 months to have a last active time of 23 months ago.
3. Run script that emails all 23-24 month users, and anonymizes all 24+ month users.
4. Run that script daily.
Step 2 is so that the people emailed at step 3 definitely have a month to act on the email, rather than in some cases only a day, as this is just starting.

dracos:
I've pushed a couple of small related things to here:
* "Ignore body users when anonymizing." does what it says, just excludes those so they are not anonymized by mistake (reported by Sweden);
* 'Store staff user when staff make anonymous update.' - I just noticed some code in New.pm that was missing from the identical code in Update.pm

PR #3047, opened 2020-05-29
Add Open Location Codes support to search box.

dracos:


dracos:
Bit annoying they don't show you the full plus code as well (which for that one is 9C7RXR26+Q2).

To make it simpler to share, one can provide a plus code relative to another base and then only give the string for the base rather than its code. Then to reverse you geocode the string and take the nearest "XR26+Q2" to that. So in this case "Edinburgh" is "9C7RXR36+" and so you then get the full code from that. It means if geocoders give different results for the same string, you could conceivably get different results, so they recommend using smaller neighbourhoods that are less likely to be different (but presumably also less likely to be found at all?!). https://github.com/google/open-location-code/wiki/Guidance-for-shortening-codes has a bit more.

The library I've added does also handle shortening/expanding, apart from the actual geocoding part. When I have a mo, I'll look at adding that, I guess it'll have to call the external geocoder, take the first result and hope that it's right.

dracos:
Actually having written that down, it seemed quite straightforward, so have done that now too, in a new commit :)

PR #3050, opened 2020-06-01
 [fixmystreet.com] Add option for recaptcha.

dracos:
This shows a recaptcha to logged out users from non-UK IP addresses on alert and report/update forms. For https://github.com/mysociety/sysadmin/issues/1302.


chrismytton:
Oh and there are quite a few red crosses in the check section, would be good to reduce those a bit!

PR #3051, opened 2020-06-02
[Bexley] Show report ID in report confirmation email

chrismytton:
As requested by the client, this adds the report ID to the report confirmation email that is sent when an unregistered user creates a report.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1892

image

### DIFF
--- a/templates/email/bexley/_problem-confirm_extra.txt
+++ b/templates/email/bexley/_problem-confirm_extra.txt
@@ -1,0 +1,2 @@
+  The report's reference number is [% report.id %].
+  Please quote this if you need to contact us about this report.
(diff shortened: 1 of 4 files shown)


templates/email/default/problem-confirm.html
```diff
@@ -18,7 +18,7 @@ <h1 style="[% h1_style %]">Please confirm your&nbsp;report</h1>
 of problem, so it will instead be sent to [% report.body %].
 [% END %]
 
-[% TRY %][% INCLUDE '_problem-confirm_extra.html' %][% CATCH file %][% END %]
+[% TRY %][% INCLUDE '_problem-confirm_extra.html' report=report %][% CATCH file %][% END %]
```
dracos, line 21: I don't think this is necessary; variables are already available to included templates, you only need this if you need to pass in something under a different name.

templates/email/bexley/_problem-confirm_extra.txt
```diff
@@ -0,0 +1,2 @@
+  The report's reference number is [% report.id %].
```
dracos, line 1: It would be good to have a test that this is indeed being included.

chrismytton:
Thanks for picking this up, @dracos!

dracos:
They never confirmed they wanted it.

PR #3054, opened 2020-06-03
[Oxfordshire] display asset layers on map

struan:
This adds asset layers to the map for OCC for:
 - drains
 - grit bins
 - streetlights
 - traffic lights
 - trees

It also adds a roads layer for access to the USRN.

It includes the asset id and USRN in the description sent over Open311.

[skip changelog]

### DIFF
--- a/web/cobrands/oxfordshire/assets.js
+++ b/web/cobrands/oxfordshire/assets.js
@@ -151,4 +151,35 @@ fixmystreet.assets.add(defaults, {
     asset_item: 'grit bin'
 });
 
+var highways_style = new OpenLayers.Style({
+    fill: false,
+    strokeColor: "#5555FF",
+    strokeOpacity: 0.1,
+    strokeWidth: 7
+});
+
+fixmystreet.assets.add(defaults, {
+    stylemap: new OpenLayers.StyleMap({
+        'default': highways_style
+    }),
+    wfs_url: proxy_base_url + 'nsg/',
+    wfs_feature: "WFS_LIST_OF_STREETS",
+    srsName: "EPSG:27700",
+    geometryName: null,
+    usrn: {
+        attribute: 'USRN',
+        field: 'usrn'
+    },
+    non_interactive: true,
+    road: true,
+    no_asset_msg_id: '#js-not-a-road',
+    asset_item: 'road',
+    asset_type: 'road',
+    actions: {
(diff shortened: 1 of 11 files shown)


web/cobrands/oxfordshire/assets.js
```diff
@@ -0,0 +1,176 @@
...
+});
+
+var streetlight_select = $.extend({
+    label: "${UNITNO}",
+    fontColor: "#FFD800",
+    labelOutlineColor: "black",
+    labelOutlineWidth: 3,
+    labelYOffset: 65,
+    fontSize: '15px',
+    fontWeight: 'bold'
```
davea, line 51: There's something about this which makes it quite hard to determine what's going on in busy areas:

image

I tried changing the font and outline sizes, but didn't make much difference. I think it's all the yellow. Could the asset markers perhaps be a different, more contrasting, colour? With a bigger font size too it pops a bit more:

image

zarino: I agree with the idea of adding stronger colour contrast between the number and the background (which will often be covered with dots). Or even just generally between the "selected" asset marker, and the background markers.

Blue for the dots looks ok – although we’ll need to make it’s not too easily confused with the blue "here’s your location" dot that was recently added to FMS. If we change the colour, we’ll also need update the yellow dot in the sidebar "pick an asset from the map" hint, assuming that’s still there.

struan: Spot the location marker:

Screenshot 2020-06-15 at 11 56 45

I will try some other colours.

struan: The dark green from the header seems to be ok:

Screenshot 2020-06-15 at 12 01 23

struan: and against some dots:

Screenshot 2020-06-15 at 12 02 06

dracos:
Made a few small changes before merging to reviewed branch: used fixmystreet.staging to pick tilma host rather than URLs; set a default stylemap for Tree/Traffic lights so all consistent; renamed asset_* to owned_* for clarity; included hover on streetlight stylemap; renamed Pavement to Pavements to match category name.

dracos:
More changes needed: couldn't select owned asset once non-owned asset selected; only request required data from WFS servers; owned code not running on grit bins; layer name fixed and attribute renamed.

PR #3056, opened 2020-06-05
[Oxfordshire] Send raise_defect flag to open311 for inspected reports

chrismytton:
Part of https://github.com/mysociety/fixmystreet-commercial/issues/1847

This deals with the USRN aspect of the data to be sent for a defect, but not the various item type codes.

### DIFF
--- a/t/cobrand/oxfordshire.t
+++ b/t/cobrand/oxfordshire.t
@@ -4,6 +4,7 @@ use CGI::Simple;
 use FixMyStreet::TestMech;
 use FixMyStreet::Script::Alerts;
 use FixMyStreet::Script::Reports;
+use Open311;
 my $mech = FixMyStreet::TestMech->new;
 
 my $oxon = $mech->create_body_ok(2237, 'Oxfordshire County Council');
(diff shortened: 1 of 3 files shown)


t/cobrand/oxfordshire.t
```diff
@@ -183,6 +184,47 @@ FixMyStreet::override_config {
...
+        is $cgi->param('attribute[street_descriptor]'), '12345 - Street Avenue, Town', 'Street description sent';
+
+        # Now set a USRN on the problem (found at submission)
+        $comment->problem->push_extra_fields({ name => 'usrn', value => '12345' });
+        $comment->problem->update;
+
+        $o->post_service_request_update($comment);
+        $cgi = CGI::Simple->new($o->test_req_used->content);
+        is $cgi->param('attribute[usrn]'), 12345, 'USRN sent with update';
+        is $cgi->param('attribute[raise_defect]'), 1, 'Defect flag sent with update';
```
struan, line 224: Given that the results of these two tests are the same, how would we know if one wasn't working?

dracos: I'm not sure what you mean, sorry - if the report already has a USRN (from the JS if it's the road/pavement categories) then it looks up only that USRN for the extra data. If it doesn't, then it looks up the closest USRN and its data. In both cases, the result should be the same, in that it has success. If you mean the mock is the same in both, I didn't think that mattered to the test, but could have a different mock if you think it does. Or let me know what I've misunderstood :)

struan: Yeah, it's me being picky about the mock using the same data. I guess I think it should be possible for the test to fail if it doesn't use the USRN which isn't the case here.

PR #3058, opened 2020-06-09
Fix double encoding of per-category templates.

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1917


dracos:
It only happens in the inspect form update form, not the main column update form.

chrismytton:
> It only happens in the inspect form update form, not the main column update form. (As that form is getting its per-category template data from a hidden data field.)

Ah yep, I see now!

PR #3060, opened 2020-06-10
[Hackney] Add asset layers & improve category import script

davea:
[skip changelog]

### DIFF
--- a/bin/import_categories
+++ b/bin/import_categories
@@ -46,10 +46,9 @@ if (!$opt->commit) {
 my $config = decode_json(path($ARGV[0])->slurp_utf8);
 
 my $body = FixMyStreet::DB->resultset('Body')->find({ name => $opt->body });
+die "Couldn't find body " . $opt->body unless $body;
 
 $body->contacts->delete_all if $opt->delete;
-
-die "Couldn't find body" unless $body;
 
 my $groups = $config->{groups};
 if ($groups) {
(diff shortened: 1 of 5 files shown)


bin/import_categories
```diff
@@ -72,12 +71,14 @@ sub make_categories {
             category => $cat->{category}
         });
         $child_cat->email($cat->{email});
+        $child_cat->send_method("Email") if $cat->{devolved};
```
dracos, line 74: Perhaps `$child_cat->send_method($cat->{devolved}) if $cat->{devolved};` and set that to Email so it could be something else in future?

web/cobrands/hackney/assets.js
```diff
@@ -0,0 +1,275 @@
...
+  attributes: {}
+});
+
+fixmystreet.assets.add(wfs_defaults, {
+  http_options: {
+      params: {
+          TYPENAME: "amenity:public_toilet",
+      }
+  },
+  asset_category: "Toilets",
```
dracos, line 116: This appears identically again further down, same asset and category.
Plus can it not be combined with "Public toilets"?

web/cobrands/hackney/assets.js
```diff
@@ -0,0 +1,275 @@
...
+      }
+  },
+  asset_category: "Cycle Hangars",
+  attributes: {}
+});
+
+fixmystreet.assets.add(wfs_defaults, {
+  http_options: {
+      params: {
+          TYPENAME: "greenspaces:park_bench",
```
dracos, line 133: Duplicated right underneath

web/cobrands/hackney/assets.js
```diff
@@ -0,0 +1,275 @@
...
+  },
+  {
+    "categories": ["Illuminated Bollards", "Non-illuminated Bollards"],
+    "item_name": "bollard",
+    "layer_name": "Bollards",
+    "styleid": "5d308d57fe2ad8046c67cdb9",
+    "layerid": "layers_streetLightingAssets"
+  },
+  {
+    "categories": ["Benches"],
```
dracos, line 180: Just to check, does the displaying of multiple asset layers for the same category work okay selection-wise?

bin/import_categories
davea: Nice, will do.

web/cobrands/hackney/assets.js
davea: Yep, it appears to! Right now we don't do anything with the feature attributes for the WFS layers, but having tested it locally it works as expected.

PR #3069, opened 2020-06-12
 [TfL] Hide unassigned categories from filters.

dracos:
If a staff user has assigned categories only, or categories have assigned users only, only show those categories in the map filter for the appropriate staff. [skip changelog]
Would hopefully fix https://github.com/mysociety/fixmystreet-commercial/issues/1891


dracos:
This has never been approved, am closing and can reopen if it ever is.

PR #3070, opened 2020-06-15
[Bucks] Spot lights ending with "A"

dracos:
And also a small refactor of that message display code so that the specific cobrand lines aren't needed (I think).
[skip changelog]


dracos:
@davea Think I've fixed the test failure now, hopefully, and also added East Sussex tidy as well as it wasn't using the shared functions it could.

PR #3078, opened 2020-06-18
[Hackney] A few bug fixes that have shaken out of testing

davea:
- Allow more cobrand control over geocoder parameters
 - Some Hackney-specific geocoder tweaks
 - Work around an issue with Hackney WFS server returning features outside of the BBOX param
 - Show array contents instead of `ARRAY(0x5587da81ffb0)` in report admin extra data

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hackney.pm
+++ b/perllib/FixMyStreet/Cobrand/Hackney.pm
@@ -15,9 +15,17 @@ sub disambiguate_location {
     my $self    = shift;
     my $string  = shift;
 
+    my $town = 'Hackney';
+
+    # Teale Street is on the boundary with Tower Hamlets and
+    # shows the 'please use fixmystreet.com' message, but Hackney
+    # do provide services on that road.
+    ($string, $town) = ('E2 9AA', '') if $string =~ /^teale\s+st/i;
+
     return {
         %{ $self->SUPER::disambiguate_location() },
-        town   => 'Hackney',
+        string => $string,
+        town   => $town,
         centre => '51.552267,-0.063316',
         bounds => [ 51.519814, -0.104511, 51.577784, -0.016527 ],
     };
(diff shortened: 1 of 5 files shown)


templates/web/base/admin/reports/edit.html
```diff
@@ -131,7 +131,22 @@
 <li><label class="inline-text" for="category">[% loc('Category:') %]</label>
     [% INCLUDE 'admin/report-category.html' %]
 </li>
-<li>[% loc('Extra data:') %] [% IF extra_fields.size %]<ul>[% FOREACH field IN extra_fields %]<li><strong>[% field.name %]</strong>: [% field.val %]</li>[% END %]</ul>[% ELSE %]No[% END %]</li>
+<li>[% loc('Extra data:') ~%]
+    [%~ IF extra_fields.size ~%]
+        <ul>
+        [%~ FOREACH field IN extra_fields ~%]
+            <li><strong>[%~ field.name ~%]</strong>: [% IF field.name == loc("Created By") ~%]
+                    [%# We can't use the .list.join() approach here because field.val
```
dracos, line 139: Instead of testing for a specific field, how about `IF field.val.0.defined` which returns true if field.val is a list (unless its first item is undef, I guess!) and then you can do the join, and just print it out if false.

perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -27,6 +35,27 @@ sub get_geocoder {
     return 'OSM'; # default of Bing gives poor results, let's try overriding.
 }
 
+sub geocoder_munge_query_params {
+    my ($self, $params) = @_;
+
+    $params->{addressdetails} = 1;
```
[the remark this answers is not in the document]
davea: Possibly, yes, though I've not looked into whether the same fields are more widely useful for other cobrand areas.

templates/web/base/admin/reports/edit.html
davea: I knew there must be a more elegant way of doing it! Thanks, that's much neater.

PR #3079, opened 2020-06-18
[Hounslow] Closed as well as fixed on front page.

dracos:
Change front page stat to include closed+fixed [skip changelog]

### DIFF
--- a/t/cobrand/hounslow.t
+++ b/t/cobrand/hounslow.t
@@ -38,6 +38,15 @@ my ($report) = $mech->create_problems_for_body(1, $hounslow_id, 'A brand new pro
     cobrand => 'fixmystreet'
 });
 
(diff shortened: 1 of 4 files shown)


t/cobrand/hounslow.t
```diff
@@ -38,6 +38,15 @@ my ($report) = $mech->create_problems_for_body(1, $hounslow_id, 'A brand new pro
     cobrand => 'fixmystreet'
 });
 
+subtest "showing the front page" => sub {
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => 'hounslow',
+    }, sub {
+        $mech->get_ok('/');
+        $mech->content_contains('completed in past month');
```
davea, line 46: Might be good to test that the value is correct too, while we're here?

PR #3080, opened 2020-06-18
[Bromley] waste updates only on open reports

dracos:
Hopefully all these small PRs like this are straightforward, rather than annoying :)
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -397,6 +397,19 @@ sub munge_report_new_contacts {
     $self->SUPER::munge_report_new_contacts($categories);
 }
 
+sub updates_disallowed {
+    my $self = shift;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -397,6 +397,19 @@ sub munge_report_new_contacts {
     $self->SUPER::munge_report_new_contacts($categories);
 }
 
+sub updates_disallowed {
+    my $self = shift;
+    my ($problem) = @_;
+
+    # Only open waste reports
+    if (my $contact = $problem->contact) {
+        my $waste = grep { $_ eq 'Waste' } @{$problem->contact->groups};
```
davea, line 406: This pattern has cropped up a few times, I wonder if we should have a `Contact::has_group` method for simplicity's sake?

PR #3081, opened 2020-06-19
[Bromley] Disallow opening if existing missed/request report

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1909 [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -541,6 +541,7 @@ sub bin_services_for_address {
             request_open => $open_request,
             request_containers => $containers,
             request_max => $quantity_max{$_->{ServiceId}},
+            enquiry_open_events => $open->{enquiry},
             service_task_id => $servicetask->{Id},
             service_task_name => $servicetask->{TaskTypeName},
             service_task_type_id => $servicetask->{TaskTypeId},
@@ -580,6 +581,8 @@ sub _parse_events {
             $open->{request}->{$container} = 1;
         } elsif (2095 <= $event_type && $event_type <= 2103) { # Missed collection
             $open->{missed}->{$service_id} = 1;
+        } else { # General enquiry of some sort
+            $open->{enquiry}->{$event_type} = 1;
         }
     }
     return $open;
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -517,6 +517,9 @@ sub bin_services_for_address {
     my $result = $echo->GetServiceUnitsForObject($property->{uprn});
     return [] unless $result;
 
+    my $events = $echo->GetEventsForObject($property->{id});
+    my $open = _parse_events($events);
+
     my @out;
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -621,7 +618,7 @@ working days (excluding weekends and bank holidays) after that date.
 
 =cut
 
-sub report_allowed {
+sub report_allowed_time {
```
struan, line 646: I'm not sure this name is an improvement tbh. Not that I can think of a better one :|

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -477,6 +477,28 @@ sub bin_services_for_address {
     my $result = $echo->GetServiceUnitsForObject($property->{uprn});
     return [] unless $result;
 
+    my $events = $echo->GetEventsForObject($property->{id});
+    my $open;
+    foreach (@$events) {
+        next if $_->{ResolvedDate} || $_->{ResolutionCodeId}; # XXX?
```
struan, line 483: is this comment still relevant? If so it's not very helpful :)

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -477,6 +477,28 @@ sub bin_services_for_address {
...
+        my $event_type = $_->{EventTypeId};
+        my $service_id = $_->{ServiceId};
+        if ($event_type == 2104) { # Request
+            my $data = $_->{Data}{ExtensibleDatum};
+            my $container;
+            foreach (@$data) {
+                if ($_->{ChildData}) {
+                    foreach (@{$_->{ChildData}{ExtensibleDatum}}) {
+                        $container = $_->{Value} if $_->{DatatypeName} eq 'Container Type';
```
struan, line 492: does it make sense to have a `last` here once you match container? I assume there can be only one.

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -477,6 +477,28 @@ sub bin_services_for_address {
...
+                    }
+                }
+            }
+            $open->{request}->{$container} = 1;
+        } elsif (2095 <= $event_type && $event_type <= 2103) { # Missed collection
+            $open->{missed}->{$service_id} = 1;
+        }
+    }
+
```
struan, line 501: It does feel like this function has got quite large now. Also, `$result` is miles away from where it's used now.

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: within_two_working_days I guess describes what it does, the fact that gets assigned to "report_allowed" is immaterial

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: yeah, ta, have reordered, and pulled out the schedule parsing and event parsing to their own functions.

PR #3083, opened 2020-06-22
Change company name

abimysoc:
[skip changelog]

UKCOD is now formally known as mySociety, this PR hopefully sweeps up all the existing mentions of the old name.

Sidenote: some of these cobrands (and they are the older ones) are using very old FAQ text and should maybe be prompted to review/update?


dracos:
Looking at old FAQs is ticket https://github.com/mysociety/fixmystreet-commercial/issues/1774

PR #3087, opened 2020-06-23
[Open311] Use devolved data on update sending.

dracos:


perllib/Open311/PostServiceRequestUpdates.pm
```diff
@@ -41,18 +39,25 @@ sub fetch_bodies {
...
+    my $o = Open311->new($self->open311_params($body, $comment));
+    return $o;
 }
 
 sub open311_params {
-    my ($self, $body) = @_;
+    my ($self, $body, $comment) = @_;
+
+    my $conf = $body;
+    if ($comment) {
```
struan, line 51: When would this ever not have a comment?

dracos: Ah, well, umm, the test suite in t/open311/post-service-request-updates.t calls this fn directly to check its cobrand output for various bodies, and it didn't seem worth passing in a fake comment even though the code always does. I guess I could factor out second half of this function to a different one and use that but didn't seem worth it.

struan: yup, makes sense.

PR #3092, opened 2020-06-25
Rewrite open311-update-reports

dracos:
This rewrites the standard fetch-reports way of updating reports to use the same code as the extension which fetches updates directly. This means it gains access to all the functionality concerning e.g. response templates. Also factor out the Open311 construction because that simplifies it everywhere used.


t/open311/getupdates.t
```diff
@@ -236,15 +244,22 @@ subtest 'test translation of auto-added comment from old-style Open311 update' =
...
     };
     my @parts = uri_split($o->test_uri_used);
     is $parts[2], '/requests.xml', 'path matches';
     my @qs = sort split '&', $parts[3];
     is_deeply(\@qs, [ 'jurisdiction_id=mysociety', 'service_request_id=638346' ], 'query string matches');
 
     is $problem3->comments->count, 1, 'added a comment';
-    is $problem3->comments->first->text, "Stängd av kommunen", 'correct comment text';
+    is $problem3->comments->first->problem_state, 'fixed - council';
+    is $problem3->comments->first->text, '', 'correct comment text';
```
davea, line 262: This doesn't seem to be testing the translation any more, AIUI, so does the test need a more accurate name?

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -838,87 +838,6 @@ sub local_coords {
...
-    }
-
-    if ( $request->{status} eq 'closed' ) {
-        if ( $self->state ne 'fixed' ) {
-            $self->state('fixed');
-            $update->mark_fixed(1);
-
-            if ( !$status_notes ) {
-                # FIXME - better text here
-                $status_notes = _('Closed by council');
```
davea, line 907: Does the removal of this text have a knock-on effect on the behaviour of FixaMinGata that they'll need to be notified of?

dracos: The update left will still lead to the default "State changed to: fixed" being displayed as the update, so think should be okay.

PR #3093, opened 2020-06-25
[Bromley] Code for fetching updates

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1898. [skip changelog]
As we already had an Echo integration present here, seemed easier to use it directly rather than go via open311-adapter requesting all the IDs. Also given I was here anyway and it didn't seem to take much longer, wrote the push-notification way of doing it should we use that method.
Wondered about having a F::C::Bromley;:Waste package for all this maybe, what do you think?

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Hercules.pm
+++ b/perllib/FixMyStreet/App/Controller/Hercules.pm
@@ -494,6 +494,63 @@ sub setup_categories_and_bodies : Private {
     @$contacts = grep { grep { $_ eq 'Waste' } @{$_->groups} } @$contacts;
 }
 
+sub receive_echo_event_notification : Path('/waste/echo') : Args(0) {
+    my ($self, $c) = @_;
+    $c->stash->{format} = 'xml';
+    $c->response->header(Content_Type => 'text/xml');
+
+    $c->detach('soap_error', [ 'POST' ]) unless $c->req->method eq 'POST';
+
+    my $echo = $c->cobrand->feature('echo');
+    $c->detach('soap_error', [ 'config', 500 ]) unless $echo;
+
+    # Make sure we log entire request for debugging
+    $c->detach('soap_error', [ 'body' ]) unless $c->req->body;
+    my $soap = join('', $c->req->body->getlines);
+    $c->log->info($soap);
+
+    my $action = $c->req->header('SOAPAction');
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Hercules.pm
```diff
@@ -494,6 +494,51 @@ sub setup_categories_and_bodies : Private {
     @$contacts = grep { grep { $_ eq 'Waste' } @{$_->groups} } @$contacts;
 }
 
+sub receive_echo_event_notification : Path('/waste/echo') : Args(0) {
+    my ($self, $c) = @_;
+    $c->response->header(Content_Type => 'text/xml');
+
+    require SOAP::Lite;
+
+    $c->detach('soap_error', [ 'POST' ]) unless $c->req->method eq 'POST';
```
struan, line 503: Should we also check the SOAP action somewhere and bail if it's not one we expect?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -678,4 +677,150 @@ sub within_working_days {
...
+            my $name = $_->{Name};
+            my $id = $_->{ResolutionCodeId};
+            $data->{resolution}{$id} = $name;
+        }
+    }
+    return $data;
+}
+
+# Check latest received update to see if state the same, skip if so
+sub waste_check_last_update {
```
struan, line 807: Possibly explain we do this as no history available so have to use current state

perllib/FixMyStreet/App/Controller/Hercules.pm
```diff
@@ -494,6 +494,51 @@ sub setup_categories_and_bodies : Private {
     @$contacts = grep { grep { $_ eq 'Waste' } @{$_->groups} } @$contacts;
 }
 
+sub receive_echo_event_notification : Path('/waste/echo') : Args(0) {
```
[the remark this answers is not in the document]
dracos: Yeah I assume we'll move it all to something, and this was in case I had to give it to someone prior to that :)

perllib/FixMyStreet/App/Controller/Hercules.pm
dracos: I wanted to use SOAP::Transport::HTTP (which does checks like this) but then it wants to generate the output and captures the $c->detaches and so on. So, yes, will do.

PR #3100, opened 2020-07-02
[Hackney] Custom domain for email envelope

davea:
Adds cobrand hook for setting domain of VERP emails, enables this for Hackney using a cobrand feature.


dracos:
(well, vworp apart from the test failures)

PR #3106, opened 2020-07-06
[Hackney] skip hackney categories for TfL red routes

struan:
Because Hackney doesn't have any matching categories for TfL red routes
skip sending the body, otherwise you get a "no contact details" message
when you click on a red route in Hackney

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -437,6 +437,13 @@ sub munge_surrounding_london {
         # Don't send any TfL categories
         %$bodies = map { $_->id => $_ } grep { $_->name ne 'TfL' } values %$bodies;
     }
+
+    # Hackney doesn't have any of the council TfL categories so don't show
+    # any Hackney categories on red routes
+    my %bodies = map { $_->name => $_->id } values %$bodies;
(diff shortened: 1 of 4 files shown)


t/cobrand/tfl.t
```diff
@@ -864,6 +875,21 @@ for my $test (
...
+        expected => [
+            'Bus stops',
+            'Flooding',
+            'Grit bins',
+            'Pothole',
+            'Timings',
+            'Traffic lights',
+            'Trees'
+        ],
+    },
```
dracos, line 892: Do we want also to test the message we don't want to be shown isn't being shown? 🤷

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -437,6 +437,13 @@ sub munge_surrounding_london {
         # Don't send any TfL categories
         %$bodies = map { $_->id => $_ } grep { $_->name ne 'TfL' } values %$bodies;
     }
+
+    # Hackney doesn't have any of the council TfL categories so don't show
+    # any Hackney categories on red routes
+    my %bodies = map { $_->name => $_->id } values %$bodies;
+    if ( $bodies{'Hackney Council'} && $self->report_new_is_on_tlrn ) {
```
dracos, line 444: Might be nice to cache the TLRN lookup as now it'll be doing it twice, but quick enough it'll be cached presumably.

PR #3111, opened 2020-07-10
Centralise update creation to include fields.

dracos:
Given the user, we can infer the name if not provided, and the extra
data if a staff user. We can also provide defaults for various other
fields. Always have superuser take precedence over from_body.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Questionnaire.pm
+++ b/perllib/FixMyStreet/App/Controller/Questionnaire.pm
@@ -206,16 +206,12 @@ sub submit_standard : Private {
         $update = $c->model('DB::Comment')->new(
             {
                 problem      => $problem,
-                name         => $problem->name,
                 user         => $problem->user,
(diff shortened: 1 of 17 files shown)


perllib/FixMyStreet/App/Controller/Questionnaire.pm
```diff
@@ -206,16 +206,12 @@ sub submit_standard : Private {
         $update = $c->model('DB::Comment')->new(
             {
                 problem      => $problem,
-                name         => $problem->name,
```
struan, line 209: On the off chance they've used a different name subsequent to making the report removing this will change the behaviour. I'm not sure it's worth being concerned about though.

dracos: Yeah, I spotted that, then I thought might be better anyway to use the name currently associated with the account filling it in at the time it's filled in 👍

PR #3113, opened 2020-07-13
Do not have bootstrap run sudo commands.

### LINKED ISSUE #2930: script/update no longer working for some installations
I'm not sure how much of a problem this is, but perhaps it will cause problems for some installations.

Running script/update after upgrading to v3.0 will now require sudo access (since bootstrap will run the new bin/install_packages), and only work on systems with APT.

I think it's great if fixmystreet helps make sure its dependencies are installed, but should it be done automatically when running script/update?

Also, if bootstrap installs the needed packages, the installation instructions for manual installation (https://fixmystreet.org/install/manual-install/) can be simplified.

dracos:
Your site user may not have access to sudo, and your admin user may cause permission issues if used to do the git checkout, or the Perl module installation. Document the separate script to install system packages instead.

Fixes #2930. I think it causes too much confusion to have it run sudo-level commands, when it's a user-level script. I realise this was introduced because a new Perl module needed a new library package, I've added some documentation about updating.

I've left in the Docker fms user being able to run bin/install_packages, don't see why not, and have changed the preinit to run this as well. I'm not sure if that's everything.

### DIFF
--- a/bin/docker.preinit
+++ b/bin/docker.preinit
@@ -34,6 +34,7 @@ fi
 su $FMS_DB_USER -c "${FMS_ROOT}/bin/docker-cobrand"
 
 # Ensure things are up to date - schema, CSS, etc
+${FMS_ROOT}/bin/install_packages docker
 su $FMS_DB_USER -c "${FMS_ROOT}/script/update"
 
 # Update reports
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
         - Improve compatibility with G Suite OpenID Connect authentication. #3032
         - Fix duplicate asset message after dismissing duplicate suggestions.
         - Improve moderation diff display in a few small ways. #3105
+        - Do not have bootstrap run sudo commands. #2930
     - Admin improvements:
         - Display user name/email for contributed as reports. #2990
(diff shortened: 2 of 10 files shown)


dracos:
Sounds good. Have squashed together; do you want to do a check of all this and then we can do a 3.0.2 release perhaps?

bin/docker.preinit
```diff
@@ -34,6 +34,7 @@ fi
 su $FMS_DB_USER -c "${FMS_ROOT}/bin/docker-cobrand"
 
 # Ensure things are up to date - schema, CSS, etc
+su $FMS_DB_USER -c "${FMS_ROOT}/bin/install_packages docker"
```
sagepe, line 37: This is going to need sudo, which is fine as there is already a sudo rule in place to allow it.

sagepe: Missed that the first time around, sorry!

dracos: Have replaced whole line with `${FMS_ROOT}/bin/install_packages docker` assuming that'll work :)

sagepe: 👍 test build from this commit for both Docker image and AMI look good.

PR #3115, opened 2020-07-14
[Bromley] Completion date display

dracos:
And disallow reporting missed collection if it's day of collection and not yet marked completed.

[skip changelog]


perllib/Integrations/Echo.pm
```diff
@@ -72,24 +72,27 @@ sub call {
...
+            Key => 'Handle',
+            Type => "Task",
+            Value => [
+                { 'msArray:anyType' => $ref->[0] },
+                { 'msArray:anyType' => $ref->[1] },
+            ],
+        );
+        push @refs, \%a;
+    }
+    my $res = $self->call('GetTasks',
```
struan, line 97: This could probably do with a comment to explain what's going on.

PR #3116, opened 2020-07-14
[Hounslow] geocoder hint for incorrect OSM data

struan:
Snowy Fielder Waye is in OSM as Snowy Fielder Way

[skip changelog]


dracos:
Have renamed in OSM :)

struan:
Updated in OSM so not required

PR #3118, opened 2020-07-14
[UK] Show message if site-wide update disallowed.

dracos:
And fix bug in .com showing of original reporter only message I just spotted.
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1886 [skip changelog]


templates/web/fixmystreet-uk-councils/admin/bodies/_updates_disallowed_hint.html
```diff
@@ -0,0 +1,9 @@
+[% cfg = c.cobrand.feature('updates_allowed') ~%]
+<span class="form-hint">
```
struan, line 2: I wonder if we should make this text stand out a bit more?

PR #3120, opened 2020-07-15
Allow deep linking with both category+group, and fix various template lookups

### LINKED ISSUE #3110: Enable deep linking to a category group and category
Now that we allow a category to exist in multiple groups it's not possible to create a deep link that will pre select a category in only one of the groups it is in.

e.g if you have the following:

* Roads
  * Potholes
  * Flooding
* Pavements
  * Potholes
  * Damaged paving

And wish to select only Roads -> Potholes it is currently not possible.

Setting filter_category to Potholes will select it in both the Road and Pavements groups. Adding a filter_group=Roads will select all the categories in the Roads group as well.

I suggest that `filter_group=Roads&filter_category=Potholes` would only select Roads -> Potholes.

dracos:
Fixes #3110. And while fixing, spotted a variety of places where category lookup on a hash would fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
         - Improve Bing geocoder results.
         - Add option of checking passwords against Have I Been Pwned.
         - Add aerial maps option to Bing maps.
+        - Select matches for both filter category and group. #3110
     - Changes:
         - Mark user as active when sent an email alert.
     - Bugfixes:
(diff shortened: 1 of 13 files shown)


dracos:
The first issue was the selection of the category from the filter (as opposed to old_category) was always selecting the last, not respecting a group. Factored that to be treated same as old category in fixup 2, more consistent now. And fixup 1 for the /report/new which adds a group check to the HTML selection of category so if present will only match a category in a matching group (similar to the filter).

PR #3124, opened 2020-07-17
Front end changes for Peterborough bartec integration

struan:
This adds a few changes required for the Bartec integration:

 * Sends over address details from the geo lookup
 * Adds bin asset layers
 * Enabled category display name editing in the admin

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -32,6 +32,8 @@ sub disambiguate_location {
 
 sub get_geocoder { 'OSM' }
 
+sub contact_extra_fields { [ 'display_name' ] }
+
 sub geocoder_munge_results {
     my ($self, $result) = @_;
     $result->{display_name} = '' unless $result->{display_name} =~ /City of Peterborough/;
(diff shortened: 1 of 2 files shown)


web/cobrands/peterborough/assets.js
```diff
@@ -162,4 +162,42 @@ fixmystreet.assets.add(light_defaults, {
...
+    },
+    attributes: {
+        asset_details: function() {
+            var a = this.attributes;
+            return a.Reference + ", " + a.Location;
+        },
+        central_asset_id: 'OBJECTID'
+    },
+    asset_id_field: 'OBJECTID',
+    asset_type: 'spot',
```
davea, line 180: trailing comma needs removing here (and L190 & L200)

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -51,6 +51,15 @@ around open311_extra_data_include => sub {
             $_->{value} .= "\n\nSkanska CSC ref: $ref->{value}" if $ref;
         }
     }
+    if ( $row->contact->email =~ /Bartec/ ) {
+        my $address = $row->geocode->{resourceSets}->[0]->{resources}->[0]->{address};
```
davea, line 57: This should check that `$row->geocode` exists first (~10% of reports have a NULL geocode field which is quite surprising!)

PR #3125, opened 2020-07-20
[Hounslow] skip disabled contacts in enquiry form

struan:
This is largely to resolve an issue with the TfL other category
appearing in the contact form even though it is disabled, and hence has
no contact details.

[skip changelog]


dracos:
Whilst this works fine, it will break again if e.g. TfL make Other non disabled, or add a General Enquiries group or some other body covers London etc. Is a better solution not to filter the contacts in `setup_general_enquiries_stash` to only be the Hounslow ones? That way it'd only ever use those ones.

PR #3126, opened 2020-07-23
Ncc alloy v2 assets

struan:
This was partially reviewed earlier but never merged.

f3f65a7, e01cb6e and fd5eb18 are new.

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/alloy.js
+++ b/web/cobrands/fixmystreet-uk-councils/alloy.js
@@ -4,6 +4,7 @@ OpenLayers.Protocol.Alloy = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
     currentRequests: [],
 
     tileSize: 512,
+    resolution: 2.388657133579254,
     abort: function() {
         if (this.currentRequests.length) {
             for (var j = 0; j < this.currentRequests.length; j++) {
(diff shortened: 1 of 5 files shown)


test.log
```diff
@@ -0,0 +1,44 @@
+2020/02/12-16:10:49 Starman::Server (type Net::Server::PreFork) starting! pid(36517)
```
dracos, line 1: Don't need to commit this? :)

web/cobrands/northamptonshire/assets.js
```diff
@@ -253,9 +254,10 @@ fixmystreet.alloy_add_layers(northants_defaults, layers);
 
 // NCC roads layers which prevent report submission unless we have selected
 // an asset.
-var northants_road_defaults = $.extend(true, {}, fixmystreet.assets.alloyv2_defaults, {
+var northants_road_defaults = $.extend(true, {}, fixmystreet.alloyv2_defaults, {
     protocol_class: OpenLayers.Protocol.AlloyV2,
     http_options: {
+        base: "https://tilma.staging.mysociety.org/resource-proxy/proxy.php?https://northants.assets/${layerid}/${x}/${y}/${z}/cluster?styleIds=${styleid}",
```
dracos, line 260: This should presumably use the is_live flag? Also presumably the "26" in the line below can't work, so does that want swapping with a placeholder or something? I see other commits have removed the need for is_live, is it still needed here? Ditto other places this changes.

web/cobrands/fixmystreet-uk-councils/alloy.js
```diff
@@ -169,8 +167,9 @@ OpenLayers.Format.AlloyV2 = OpenLayers.Class(OpenLayers.Format.GeoJSON, {
 
 OpenLayers.Protocol.AlloyV2 = OpenLayers.Class(OpenLayers.Protocol.Alloy, {
     tileSize: 128,
+    resolution: 4.777314267158508,
```
dracos, line 170: Does this affect Hackney's use?

struan: Yes in as much as Hackney is likely to not work without this.

PR #3127, opened 2020-07-28
[Bromley] Few more checks

dracos:
Deal with blocks of flats sharing missed bin reporting (so if one reports, the others cannot); not allowing missed reporting if the task was labelled as not completed or completed but with excess waste, and treat out-of-stock requests as open (which I hope is how they do it, I have no example data to look at).


t/app/controller/waste.t
```diff
@@ -64,12 +64,12 @@ FixMyStreet::override_config {
     };
     subtest 'Report a missed bin' => sub {
         $mech->content_contains('service-101', 'Can report, last collection was 27th');
-        $mech->content_contains('service-537', 'Can report, last collection was 27th');
+        $mech->content_lacks('service-537', 'Cannot report, last collection was 27th but the service unit has a report');
```
struan, line 67: Only thing I wonder is if we should check that it says someone has already reported it? Mostly as I wanted to double check it did this because saying that for flats seems particularly important.

PR #3129, opened 2020-07-29
Full text search index

dracos:
Adds a full text search index to both problems and updates, and updates the admin search to use that index.

If DB_FULL_TEXT_SEARCH_CONFIG is set prior to the database schema change, that value will be used, otherwise it will default to english.


dracos:
Just to note recreated the indexes with coalesce for those columns that could be null. I knew I had to do this but somehow forgot.

PR #3131, opened 2020-07-30
[Docs] Add instructions & example for including HTML in notice fields

davea:
Also fixes `jekyll serve` when running locally.

Please check the following:

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1905


docs/_includes/admin-tasks-content.md
```diff
@@ -653,6 +653,50 @@ selected. This can be used to provide extra information about the category,
...
+<img alt="" src="/assets/img/pro-user-guide/blocked-drains-guidance.png" class="admin-screenshot" />
+
+The **Notice text** field allows you to include HTML tags to apply text
+formatting, add links, or include images in the notice shown to the user. The
+above notice was generated from this HTML:
+
+```html
+We take different action depending on how a drain is blocked.<br>
+
+<span style="font-weight: normal">A blockage on the surface of a drain could be caused by leaves or litter. These blockages can usually be cleared by carrying out street sweeping.</span><br>
```
dracos, line 671: I wonder if we should make the extra question labels not be bold, then this sort of reset wouldn't be needed.

docs/_includes/admin-tasks-content.md
```diff
@@ -653,6 +653,50 @@ selected. This can be used to provide extra information about the category,
...
+ - `b`
+ - `i`
+ - `strong`
+ - `em`
+ - `a`
+   - The following attributes are allowed: `href`, `style`, `target="_blank"`, `title`
+ - `img`
+   - The following attributes are allowed: `src`, `alt`, `width`, `height`, `hspace`, `vspace`, `align`, `sizes`, `srcset`
+ - `font`
+   - The following attributes are allowed: `color`
```
dracos, line 693: I think we only allowed this because someone was already using it. They could equivalently use span style=color... which is probably better (can specify background then too), so I'd drop this from the docs.

docs/_includes/admin-tasks-content.md
davea: Possibly, but I'd say that's beyond the scope of this PR. The example was taken from an existing cobrand so there's clearly a need there. I've ticketed this as https://github.com/mysociety/fixmystreet/issues/3142.

PR #3133, opened 2020-07-31
Full text admin user search

dracos:
One "bug" fix - forcing postgresql to use the index in situations where it does not want to.

And then adding full text search to the user admin search as well. Plus side - is quicker. Down side - can no longer search for part of name/phone/email only whole words. Perhaps alternative solution would be to search whole fields first (for exact match) and only do ilike search if no results to that. Or just say is okay as is. Review the first bug commit regardless please :)

[skip changelog]

### DIFF
--- a/bin/update-schema
+++ b/bin/update-schema
@@ -215,6 +215,7 @@ else {
 # (assuming schema change files are never half-applied, which should be the case)
 sub get_db_version {
     return 'EMPTY' if ! table_exists('problem');
+    return '0074' if index_exists('users_fulltext_idx');
     return '0073' if index_exists('problem_fulltext_idx');
     return '0072' if constraint_contains('contacts_state_check', 'staff');
     return '0071' if table_exists('manifest_theme');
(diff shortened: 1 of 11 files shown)


dracos:
Just a side effect I hadn't noticed, hmm. Could translate `@` to space and then it'd index local and domain separately, which I think would work (assuming you did the same on query too)?

dracos:
Have changed it to treat @ and . as spaces in both indexing and query, so a domain search should work (possible false positives now I guess, e.g. a search for "gmail.com" would match "gmail@example.com" but that doesn't seem too bad. Changed the index to wrap `id || ' ' || coalesce(name,'') || ' ' || coalesce(email,'') || ' ' || coalesce(phone,'')` in `translate(..., '@.', '  ')`, and changed FullTextSearch.pm to allow variable translation and apply it to the query too (otherwise you'd search for "gmail.com" and it wouldn't find anything because only "gmail" and "com" were indexed).

PR #3134, opened 2020-08-04
Allow inspectors to change asset of report

dracos:
Couple of these commits have questions I'm not sure of the answer to. The first does fix a bug in that if you currently go to e.g. https://www.fixmystreet.com/report/new?longitude=-0.604850&latitude=51.710853&category=Street+light+dim you'll see the selection of other assets is a bit weird - a click doesn't change asset, it deselects then snap-to-nearest kicks in so it sort of works but the next click then gets confused, and so on. With this change, all the asset layers are underneath so it's better. I'm not sure if it affects anything else, though.

The second made my code a bit simpler, I wasn't sure why the update was restricted to only the inspector form. I realise the around page may have a hidden form with fields in that would match and be updated also, but if you ever then made a new report they'd be updated by the map click etc. But I wasn't sure if there was some way this could have an effect somewhere else also.

Ignoring those, the actual addition hopefully makes sense, a change asset button and its code, and related. Not sure about `             `$inspect_fields` but seemed easiest way to handle that.

Animation (of slightly older code) at https://github.com/mysociety/fixmystreet-commercial/issues/1919
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1919

### DIFF
--- a/web/cobrands/fixmystreet/staff.js
+++ b/web/cobrands/fixmystreet/staff.js
@@ -189,6 +189,7 @@ fixmystreet.staff_set_up = {
         populateSelect($priorities, priorities_data, 'priorities_type_format');
         updateTemplates({'category': category});
         $priorities.val(curr_pri);
+        update_change_asset_button();
     });
 
     function state_change(state) {
@@ -251,6 +252,46 @@ fixmystreet.staff_set_up = {
             });
         }
     }
+
+    function get_value_and_group(slr) {
+        var elt = $(slr)[0];
+        var group = $(elt.options[elt.selectedIndex]).closest('optgroup').prop('label');
+        return { 'value': $(elt).val(), 'group': group || '' };
+    }
+
+    function update_change_asset_button() {
+        var category = get_value_and_group('#category'); // The inspect form category dropdown only
+        var found = false;
+        if (fixmystreet.assets) {
+            for (var i = 0; i < fixmystreet.assets.layers.length; i++) {
+                var layer = fixmystreet.assets.layers[i];
+                if ((layer.fixmystreet.asset_category || layer.fixmystreet.asset_group) && layer.relevant(category.value, category.group)) {
+                    found = true;
+                    break;
+                }
+            }
+        }
+        if (found) {
+            $('.btn--change-asset').show();
+        } else {
+            $('.btn--change-asset').hide();
+        }
+    }
+    update_change_asset_button();
+
+    $('.btn--change-asset').on('click', function(e) {
+        e.preventDefault();
+        $(this).toggleClass('asset-spot');
+        if ($(this).hasClass('asset-spot')) {
+            var v = get_value_and_group('#category');
+            $('#inspect_form_category').val(v.value);
+            $('#inspect_category_group').val(v.group);
+        } else {
+            $('#inspect_form_category').val('');
+            $('#inspect_category_group').val('');
+        }
+        $(fixmystreet).trigger('inspect_form:asset_change');
+    });
 
     // Make the "Provide an update" form toggleable, hidden by default.
     // (Inspectors will normally just use the #public_update box instead).
--- a/web/cobrands/fixmystreet/staff.js
+++ b/web/cobrands/fixmystreet/staff.js
@@ -246,17 +246,8 @@ fixmystreet.staff_set_up = {
         // triage pages may not show geolocation button
         if (el) {
             fixmystreet.geolocate(el, function(pos) {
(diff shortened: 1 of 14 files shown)


web/cobrands/fixmystreet/staff.js
```diff
@@ -248,6 +249,46 @@ fixmystreet.staff_set_up = {
...
+        return { 'value': $(elt).val(), 'group': group || '' };
+    }
+
+    function update_change_asset_button() {
+        var category = get_value_and_group('#category');
+        var found = false;
+        if (fixmystreet.assets) {
+            for (var i = 0; i < fixmystreet.assets.layers.length; i++) {
+                var asset_layer = fixmystreet.assets.layers[i];
+                if (asset_layer.relevant(category.value, category.group)) {
```
davea, line 264: This results in a crash when viewing the inspector report page if the layer doesn't have `asset_group` or `asset_category` (e.g. Lincolnshire's NSG layer).

web/js/map-OpenLayers.js
```diff
@@ -146,6 +146,23 @@ $.extend(fixmystreet.utils, {
         };
       },
 
+      update_pin_input_fields: function(lonlat, wrapper) {
+        wrapper ||= 'document';
+        var bng = lonlat.clone().transform(
+            new OpenLayers.Projection("EPSG:4326"),
+            new OpenLayers.Projection("EPSG:27700") // TODO: Handle other projections
```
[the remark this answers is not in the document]
dracos: Zurich? :) I more meant if an international user wished to use this functionality in some way, it's not configurable.

web/cobrands/fixmystreet/staff.js
dracos: NSG layer shouldn't have been loaded because I skipped all VectorNearest layers? Hmm.

dracos: Ah, LLPG layer, which is loaded as Asset because no usrn/road provided.

dracos: Have pushed fix for that hopefully

dracos:
Thanks for the detailed stuff.
1. Hmm, will investigate, that's pretty basic isn't it. At least that one should be fixable though.
2/3. That text for 2 is coming from Peterborough's asset handling, so that's a bit wrong. I imagine this is affecting 3 as well, if multiple attribute field settings are running.
4. I guess the category change event is kicking off the new-report look-for-duplicates code. I need to separate things out so only the asset layer stuff is run.
5. Yeah, I guess I assumed it'd only be used when it already had an asset. Could try and disable snapping, so it would only do something if asset selected, that probably makes sense. Nothing is saved unless they submit, at least.
6. Probably. Ah, what's your browser? I think e.g. Firefox remembers form fields sometimes (to save you if you hit refresh by mistake), so if you'd used the change asset button, then hit refresh (as opposed to clicking in URL bar then hitting return), then the hidden category field populated when you clicked change asset would still be populated which would kick things off as you say.
Mobile: Yeah, scroll would make sense. Changing the label to show everything that would be set (can it have newlines?) but some other overlay probably better as you suggest.

davea:
> 6. Probably. Ah, what's your browser? I think e.g. Firefox remembers form fields[...]

Yep, this was with Firefox.

dracos:
Asking for rereview for your issues 1-5 (5 optional). Will look at mobile stuff tomorrow hopefully.

dracos:
Okay, hopefully that fixed and nicer mobile experience and couple of other bits. New commits are:
* https://github.com/mysociety/fixmystreet/pull/3134/commits/82ad3f1db8debc125e8d54a372b26379a54d63db - a general bugfix for the admin drag which as a side effect fixes the asset layer not being selectable sometimes (here and on live site if e.g. no asset snap when making new report). 
* https://github.com/mysociety/fixmystreet/pull/3134/commits/a78ffca120498653b52ccbe51703b526d6e7b4ab - looks like the unexpected around view was always happening; have added a fix so it doesn't happen whilst changing asset.
* https://github.com/mysociety/fixmystreet/pull/3134/commits/cf29fa12e4e34dad0e35a1577f701f22e73817cf - stuck in missing hover styles as I thought I'd broken it but turned out I was in one of these areas;
* https://github.com/mysociety/fixmystreet/pull/3134/commits/71e3de0f305449182d63319bcc1b3a86b857d62b - better time on mobile.

Regarding site code - hmm, yes, ideally, it should load VectorNearest layers with a `usrn` and let that code run too. As you say, perhaps one to leave for now, but I do expect it to probably be needed in future.

PR #3136, opened 2020-08-06
De-Cify Result::Problem and Result::Comment and _add_columns

dracos:
Start of work for offline CSV generation, needed some of these functions to not require C to exist, given it only does in web context. The cobrand is set on the schema on a request and by offline things that need it, any other things now passed directly in (e.g. a user for meta_line). One commit per function changed, will squash :) [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -631,13 +631,14 @@ meta data about the report.
 =cut
 
 sub meta_line {
-    my ( $problem, $c ) = @_;
+    my ( $problem, $user ) = @_;
+    my $cobrand = $problem->result_source->schema->cobrand;
 
     my $date_time = Utils::prettify_dt( $problem->confirmed );
     my $meta = '';
 
     my $category = $problem->category_display;
-    $category = $c->cobrand->call_hook(change_category_text => $category) || $category;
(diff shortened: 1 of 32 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -918,11 +919,11 @@ sub as_hashref {
     };
     $out->{is_fixed} = $self->fixed_states->{ $self->state } ? 1 : 0 if !$cols || $cols->{is_fixed};
     $out->{photos} = [ map { $_->{url} } @{$self->photos} ] if !$cols || $cols->{photos};
-    $out->{meta} = $self->confirmed ? $self->meta_line( $c->user ) : '' if !$cols || $cols->{meta};
-    $out->{created_pp} = $c->cobrand->prettify_dt( $self->created ) if !$cols || $cols->{created_pp};
+    $out->{meta} = $self->confirmed ? $self->meta_line : '' if !$cols || $cols->{meta};
```
davea, line 922: The removal of `$c->user` means the body contribution detail is omitted later on (I think?) - does that matter/is it intentional?

dracos: Nothing uses the meta line, as far as I can tell. I thought it a bit odd that hashref output would include random extra user details here depending on who requested it, too, so I was happy for it to not do that.

PR #3141, opened 2020-08-07
[Northamptonshire] add in V2 PRoW and Traffic Light assets

struan:
[skip changelog]

### DIFF
--- a/web/cobrands/northamptonshire/assets.js
+++ b/web/cobrands/northamptonshire/assets.js
@@ -11,45 +11,6 @@ var base_url = fixmystreet.staging ?
 var url_with_style = base_url + '?styleIds=${styleid}';
 
 var layers = [
-/*
-{
-  "name": "Posts",
-  "id": "5d480a68fe2ad809d85a78cf"
-}
-{
-  "name": "Grips",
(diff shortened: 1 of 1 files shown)


web/cobrands/northamptonshire/assets.js
```diff
@@ -109,59 +87,13 @@ var layers = [
...
-      "Restricted Visibility"
-  ],
-  "layer_name": "Tree",
-  "styleid": '5d481376fe2ad8059ce44ef2',
-},
-{
-  "categories": [ "Safety Bollard - Damaged/Missing" ],
-  "layer_name": "Safety Bollard",
-  "styleid": "5d481446fe2ad8059ce44f02",
+  "styleid": "5d481a4ffe2ad809d85a7b90&styleIds=5d481742fe2ad809d85a7b05"
```
dracos, line 90: Is this adding another layer also?

struan: This is fetching two styles in the same call and displaying them on the same layer. In Alloy the assets are grouped into layers in which individual asset types are styles and you can request multiple styles in one call.

PR #3143, opened 2020-08-07
Allow HTML in response templates and updates from staff

davea:
Response templates can now include HTML which is sanitized and rendered when shown on the front end and in update emails. This also extends to updates left by staff/superusers.

Not 100% happy with the approach but at this point would be good to have a second pair of eyes before I dig any further into the rabbit hole.

For https://github.com/mysociety/fixmystreet-commercial/issues/1928

### DIFF
--- a/docs/_includes/admin-tasks-content.md
+++ b/docs/_includes/admin-tasks-content.md
@@ -829,7 +829,11 @@ above, ‘Creating a template’. Additionally you can delete the template from 
 #### HTML content in templates
 
 HTML tags are permitted in response templates, which makes it possible to include
-images or rich text formatting in the updates which are added to reports.
+hyperlinks or rich text formatting in the updates which are added to reports.
+
+Be aware that response templates are emailed to users as well as being shown on
+the site, so it's best to keep any HTML formatting quite light-touch due to the
+quirks of email clients' rendering of HTML message.
 
 Refer to the section ["HTML Content in notices"](#html-content-in-notices) above for details of
 what tags and attributes are allowed.
--- a/docs/_includes/admin-tasks-content.md
+++ b/docs/_includes/admin-tasks-content.md
@@ -825,6 +825,15 @@ Click on ‘Templates’ in the admin menu. You will see a table of existing tem
 beside the status you wish to change. You may alter any of the fields as described in the section
 above, ‘Creating a template’. Additionally you can delete the template from this page.
 
+
+#### HTML content in templates
+
+HTML tags are permitted in response templates, which makes it possible to include
+images or rich text formatting in the updates which are added to reports.
+
+Refer to the section ["HTML Content in notices"](#html-content-in-notices) above for details of
+what tags and attributes are allowed.
+
 </div>
 
 <div class="admin-task" markdown="1" id="view-statistics">
--- a/perllib/FixMyStreet/App/View/Web.pm
+++ b/perllib/FixMyStreet/App/View/Web.pm
@@ -126,29 +126,34 @@ sub staff_html_markup_factory {
 
     return sub {
         my $text = shift;
-        unless ($staff) {
-            return FixMyStreet::Template::html_paragraph(add_links($text));
-        }
-
-        $text = FixMyStreet::Template::sanitize($text);
-
-        # Apply Markdown-style italics
-        $text =~ s{\*(\S.*?\S)\*}{<i>$1</i>};
-
-        # Mark safe so add_links doesn't escape everything.
-        $text = FixMyStreet::Template::SafeString->new($text);
-
-        $text = add_links($text);
-
-        # If the update already has block-level elements then don't wrap
-        # individual lines in <p> elements, as we assume the user knows what
-        # they're doing.
-        unless ($text =~ /<(p|ol|ul)>/) {
-            $text = FixMyStreet::Template::html_paragraph($text);
-        }
-
-        return $text;
+        return _staff_html_markup($text, $staff);
     }
+}
+
+sub _staff_html_markup {
+    my ( $text, $staff ) = @_;
+    unless ($staff) {
+        return FixMyStreet::Template::html_paragraph(add_links($text));
+    }
+
+    $text = FixMyStreet::Template::sanitize($text);
+
+    # Apply Markdown-style italics
+    $text =~ s{\*(\S.*?\S)\*}{<i>$1</i>};
+
+    # Mark safe so add_links doesn't escape everything.
+    $text = FixMyStreet::Template::SafeString->new($text);
+
+    $text = add_links($text);
+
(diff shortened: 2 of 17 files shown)


perllib/FixMyStreet/App/View/Web.pm
```diff
@@ -100,7 +101,7 @@ sub add_links {
     my $text = shift;
     $text = FixMyStreet::Template::conditional_escape($text);
     $text =~ s/\r//g;
-    $text =~ s{(https?://)([^\s]+)}{"<a href=\"$1$2\">$1" . _space_slash($2) . '</a>'}ge;
+    $text =~ s{([^"'])(https?://)([^\s]+)}{"$1<a href=\"$2$3\">$2" . _space_slash($3) . '</a>'}ge;
```
dracos, line 104: This won't catch a link if it's the first thing in `$text`. Something like `(?<!["'])` is the same and is zero-width, so you then don't need the $1 in the replacement either.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -132,6 +133,18 @@ sub send() {
                 }
             }
 
+            # This seems a bit inefficient to do for every single update, just
+            # so we can later determine if it came from a staff/superuser. It
+            # is run as a batch job though so maybe that doesn't matter?
+            # Would a JOIN in the query be better option, or maybe adding something
+            # to the comment's extra when it's created? The latter approach would
+            # also solve the theoretical problem of a user leaving an HTML update
```
dracos, line 141: All updates left by staff should have extra data of is_body_user or is_superuser so that it doesn't matter if the user loses their staff status later. Also means you don't need to get the user here, so that's handy!

templates/email/default/_email_comment_list.txt
```diff
@@ -1,7 +1,7 @@
 [% FOR row IN data -%]
 [% row.item_name _ ' : ' IF row.item_name AND NOT row.item_anonymous -%]
 [% '(' _ cobrand.prettify_dt(row.confirmed) _ ') ' IF cobrand.include_time_in_update_alerts -%]
-[% row.item_text %]
+[% row.item_text | strip_html %]
```
dracos, line 4: Hmm, strip_html won't make much sense if e.g. they've used `a` or `img`. Guess it needs some HTML to text parser. Googling gives HTML::FormatText::WithLinks which looking at the author, umm, we can probably use that for starters ;) Doesn't look like it handles lists though. And images presumably need to become alt text...
Alternatively, perhaps it could say this update had HTML, please view on the site.

perllib/FixMyStreet/Template.pm
```diff
@@ -157,4 +157,42 @@ sub sanitize {
...
+}
+
+=head2 email_html_sanitize
+
+Intended for use in the _email_comment_list.html template to allow HTML
+in updates from staff/superusers. Also applies Markdown italics formatting.
+
+This would be better as a series of filters much like we have in
+FixMyStreet::App::View::Web, however adding those filters to
+FixMyStreet::App::View::Email didn't appear to work.
```
dracos, line 180: ::App::View::Email is only used by web-flow generated emails.

templates/email/default/_email_comment_list.html
```diff
@@ -5,7 +5,7 @@
         <img style="[% list_item_photo_style %]" src="[% inline_image(update.get_first_image_fp) %]" alt="">
       </a>
     [%~ END %]
-      [% update.item_text | html_para | replace('<p>', '<p style="' _ list_item_p_style _ '">') %]
+      [% email_html_sanitize(update) | html_para | replace('<p>', '<p style="' _ list_item_p_style _ '">') %]
```
dracos, line 8: Guess one issue is things like links won't have the right link style, I assume (like we fix the `<p>` in this line) - could fix that one too at least, I guess, not sure if anything else would matter. Probably minor.

docs/_includes/admin-tasks-content.md
```diff
@@ -825,6 +825,19 @@ Click on ‘Templates’ in the admin menu. You will see a table of existing tem
...
+
+HTML tags are permitted in response templates, which makes it possible to include
+hyperlinks or rich text formatting in the updates which are added to reports.
+
+Be aware that response templates are emailed to users as well as being shown on
+the site, so it's best to keep any HTML formatting quite light-touch due to the
+quirks of email clients' rendering of HTML message.
+
+Refer to the section ["HTML Content in notices"](#html-content-in-notices) above for details of
+what tags and attributes are allowed.
```
dracos, line 839: As html_para is applied to updates, do we want to limit this to inline tags only? Otherwise if they use `<ul>` etc it's going to get wrapped in `<p>`s on top, unless they happen to do the HTML all on one line...

Alternative would be to not do html_para if there is HTML and then mandate they must use valid HTML including start/end `<p>` (or whatever), but then we only have styling for `item-list__update-text p` for example. 

Ideally I assume we'd want to allow list items, no reason why you wouldn't want to, in which case I'd say you'd probably want to have the staff_sanitize function do combined what it does currently, markup, and html_para, depending on whether it needed to or not - I guess if the output had no p/ul/ol in it, then run html_para, but don't if it already does?

perllib/FixMyStreet/Script/Alerts.pm
davea: That is handy indeed, I'll use that 👍

templates/email/default/_email_comment_list.html
davea: Doesn't look like we apply a `class` or `style` to links, just a top-level directive for `a` in the document CSS. I suppose if the user has put their own `style` attribute it might clash with the top-level rule... Probably OK, like you say, but I guess we could strip the `style` only in email HTML?

perllib/FixMyStreet/App/View/Web.pm
```diff
@@ -26,7 +26,7 @@ __PACKAGE__->config(
         add_links => \&add_links,
         escape_js => \&escape_js,
         markup => [ \&markup_factory, 1 ],
```
dracos, line 28: Think you need to remove this line.

templates/email/default/_email_comment_list.txt
dracos: Oh, this comment still applies - if we want to get it live, perhaps the message I suggest? Or that module?

davea: Sigh, overlooked this! Will take a look at that module next week.

davea:
Thanks for https://github.com/mysociety/fixmystreet/pull/3143/commits/4a4ee013f02571ec9ed49fbcfb1f02cbb5e63b1f - looks good to me 👍

dracos (approved):

PR #3144, opened 2020-08-10
[central-beds] New cobrand for Central Bedfordshire

zarino:
New cobrand for Central Beds, following the styling conventions of pages like http://my.centralbedfordshire.gov.uk

Homepage, map page, and FAQ page all look good to me at mobile and desktop widths. I‘ve also added some email templates and styles. It’s been a while since I’ve done one of these, so let me know if there’s something I’ve forgotten.

Coordinates in `perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm` were pulled from the MapIt data for Central Bedfordshire. Someone with more expertise than me should probably check they’re right, before we accidentally deploy a cobrand with the wrong bounding box ;-)

Screenshot_2020-08-10 Report It

Screenshot_2020-08-10 Report It(1)

[skip changelog]


perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -0,0 +1,24 @@
+package FixMyStreet::Cobrand::CentralBedfordshire;
+use parent 'FixMyStreet::Cobrand::Whitelabel';
+
+use strict;
+use warnings;
+
+sub council_area_id { return 21070; }
```
davea, line 7: Perl uses the value of the last expression in a function as the return value in the absence of the `return` keyword, so this (and the ones below) could more concisely be
```suggestion
sub council_area_id { 21070 }
```

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -0,0 +1,24 @@
...
+sub send_questionnaires { 0 }
+
+sub disambiguate_location {
+    my $self    = shift;
+    my $string  = shift;
+
+    return {
+        %{ $self->SUPER::disambiguate_location() },
+        centre => '52.006697,-0.436005',
+        bounds => [ 51.805087, -0.702181, 52.190913, -0.143957 ],
```
davea, line 20: There's also `bin/fixmystreet.com/generate_council_location` to save you from having to look these values up manually.

zarino: 😲

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
zarino: I just copied another cobrand file ;-) Looks like about 50% of them use returns, and 50% don’t.

PR #3146, opened 2020-08-10
Offline CSV generation

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1926

From the command line, can call with export=2 instead of 1 to get the 202 flow, otherwise is as before.

### DIFF
--- a/perllib/FixMyStreet/Reporting.pm
+++ b/perllib/FixMyStreet/Reporting.pm
@@ -1,0 +1,337 @@
+package FixMyStreet::Reporting;
+
+use DateTime;
+use Moo;
+use Text::CSV;
+use Types::Standard qw(ArrayRef CodeRef Enum HashRef InstanceOf Int Maybe Str);
+use FixMyStreet::DB;
+
+# What are we reporting on
+
+has type => ( is => 'ro', isa => Enum['problems','updates'] );
+has on_problems => ( is => 'lazy', default => sub { $_[0]->type eq 'problems' } );
+has on_updates => ( is => 'lazy', default => sub { $_[0]->type eq 'updates' } );
+
+# Filters to restrict the reporting to
+
+has body => ( is => 'ro', isa => Maybe[InstanceOf['FixMyStreet::DB::Result::Body']] );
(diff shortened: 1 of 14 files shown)


t/app/controller/dashboard.t
```diff
@@ -252,7 +259,30 @@ FixMyStreet::override_config {
...
+        (my $name = $f->basename) =~ s/-part$//;;
+        $mech->log_in_ok( $counciluser->email );
+        $mech->get_ok('/dashboard/status');
+        $mech->content_contains('/dashboard/csv/www.example.org-body-' . $body->id . '-start_date-2014-01-02.csv');
+        $mech->content_like(qr/$name\s*<br>0KB\s*<i>In progress/);
+
+        my $token = 'access_token=' . $counciluser->id . '-1234567890abcdefgh';
+        $mech->get_ok("/dashboard/csv/$name?$token");
+        is $mech->res->code, 202;
+    }
```
struan, line 285: Should we not also check that if all files are generated it reflects that too. If for no other reason that to check we don't get endless request loops if someone leaves the tab open.

perllib/FixMyStreet/Reporting.pm
```diff
@@ -305,6 +306,46 @@ sub generate_csv {
 
 # Output code
 
+sub cache_dir {
+    my $self = shift;
+
+    my $cfg = FixMyStreet->config('PHOTO_STORAGE_OPTIONS');
+    my $dir = $cfg ? $cfg->{UPLOAD_DIR} : FixMyStreet->config('UPLOAD_DIR');
+    $dir = path($dir, "dashboard_csv")->absolute(FixMyStreet->path_to());
+    my $subdir = $self->user ? $self->user->id : 0;
```
struan, line 315: Should there not always be a user here?

dracos: No, kiitc uses the cobrand hook to allow non-authenticated access to the CSV export.

dracos: Though they will still be using export=1 and not be large, so wouldn't hit this. But still, might as well be safe.

PR #3153, opened 2020-08-17
[Northamptonshire] script to update external ids

struan:
The migration from V1 to V2 changes the IDs so take a CSV of the changes
and apply it, saving the old id just in case.

[skip changelog]

### DIFF
--- a/bin/northamptonshire/update_ids_for_v2
+++ b/bin/northamptonshire/update_ids_for_v2
@@ -1,0 +1,59 @@
+#!/usr/bin/env perl
+
+# The migration from Alloy V1 to V2 causes the IDs to change. This takes a CSV
+# with a mapping of old ids to new ones and applies it to the external_id, saving
+# the old one in extra in case it's relevant.
+
+use strict;
+use warnings;
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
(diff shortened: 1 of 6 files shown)


bin/northamptonshire/update_ids_for_v2
```diff
@@ -0,0 +1,57 @@
...
+    'file=s' => \$file,
+);
+
+if (!$commit) {
+    say "*** DRY RUN ***";
+}
+
+my $csv = Text::CSV->new({ binary => 1 });
+open my $fh, "<:encoding(utf-8)", $file or die "Failed to open $file: $!\n";
+my @cols = @{$csv->getline($fh)};
```
dracos, line 38: Could use $csv->header here instead, e.g. like Admin/Users.pm?

bin/northamptonshire/update_ids_for_v2
```diff
@@ -0,0 +1,57 @@
...
+$csv->column_names (@cols);
+
+my $count = 0;
+my $northants = FixMyStreet::DB->resultset("Body")->find({ name => 'Northamptonshire County Council' });
+if ($northants) {
+    while (my $report = $csv->getline_hr( $fh ) ) {
+        my $p = FixMyStreet::DB->resultset('Problem')->search({ external_id => $report->{$OLD_ID}, bodies_str => $northants->id })->first;
```
dracos, line 45: Probably want to use `->to_body(...)` rather than direct bodies_str lookup in case there's a report sent to both county and district (doesn't look like there is though).

PR #3154, opened 2020-08-17
[Bromley] Some small waste fixes

dracos:
A fix for the script not looking at all reports when fetching updates, better message on closed waste reports, create the alert for reports so users get updates(!), and show the last collection resolution string. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -475,6 +475,15 @@ sub add_report : Private {
     my $report = $c->stash->{report};
     $report->confirm;
     $report->update;
+
+    $c->model('DB::Alert')->find_or_create({
+        user => $report->user,
+        alert_type => 'new_updates',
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -475,6 +475,15 @@ sub add_report : Private {
...
     $report->confirm;
     $report->update;
+
+    $c->model('DB::Alert')->find_or_create({
+        user => $report->user,
+        alert_type => 'new_updates',
+        parameter => $report->id,
+        cobrand => $report->cobrand,
+        lang => $report->lang,
+    })->confirm;
```
struan, line 485: Do we want to offer a way to not get alerts?

dracos: We don't on .com for normal reports :) and I'd say you're more likely to want to know about these type of reports.

PR #3158, opened 2020-08-18
[Highways England] Work across whole network.

dracos:
This adjusts the cobrand so it allows reporting on any HE road, and adjusts the sending so it can utilise configured per-area emails for any area, not just 7, or just fall back as now to the  if none is present. [skip changelog]


davea (approved):

dracos:
This is awaiting approval, can be reopened once that happens.

PR #3163, opened 2020-08-21
[Bromley] Waste tweaks

dracos:
[skip changelog]
7b29f3e for https://3.basecamp.com/4020879/buckets/16062166/todos/2941906765
d3745c4 for https://3.basecamp.com/4020879/buckets/16062166/todos/2942227397
c501fe2 and 12df555 for https://3.basecamp.com/4020879/buckets/16062166/uploads/2945618773

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -582,6 +582,7 @@ sub bin_services_for_address {
             }
 
             my $row = $task_ref_to_row{$ref};
+            $row->{last}{state} = $state;
             $row->{last}{completed} = $completed;
             $row->{last}{resolution} = $resolution;
             $row->{report_allowed} = within_working_days($row->{last}{date}, 2);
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -570,9 +570,17 @@ sub bin_services_for_address {
             my $ref = join(',', @{$_->{Ref}{Value}{anyType}});
             my $completed = construct_bin_date($_->{CompletedDate});
             my $state = $_->{State}{Name} || '';
+
+            my $resolution_id = $_->{Resolution}{Ref}{Value}{anyType} || 'x';
```
davea, line 574: Is the `x` a placeholder so the subsequent `me.external_status_code` fails? If so, would it be better/clearer to just skip the template lookup altogether if `$_->{Resolution}{Ref}{Value}{anyType}` doesn't exist?

dracos: Not sure why I did this! Will do.

davea (approved):
🗑️ ✔️

PR #3165, opened 2020-08-24
[Hackney] Allow editing & addition of contacts with special destination addresses

davea:
This adds a cobrand hook to validate the email address provided when editing a contact
in the admin.

[skip changelog]


dracos (approved):
Hmm, how did I miss that.

dracos:
How did I also miss that this means every contact passes as being in this special format.

PR #3169, opened 2020-08-25
[Oxfordshire] Hide duplicate ‘Other’ category from map filters

davea:
Two rows would be returned because the district and county `Other` contacts have different `extra` fields, so this adds a `WHERE` for the Oxfordshire cobrand to only fetch the county contact.

This probably isn't the best way to fix the problem.

[skip changelog]

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1964

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Around.pm
+++ b/perllib/FixMyStreet/App/Controller/Around.pm
@@ -254,6 +254,9 @@ sub check_and_stash_category : Private {
             distinct => 1
         }
     )->all_sorted;
+    # Ensure only uniquely named categories are shown
+    my %seen;
+    @categories = grep { !$seen{$_->category_display}++ } @categories;
     $c->stash->{filter_categories} = \@categories;
     my %categories_mapped = map { $_->category => 1 } @categories;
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
" This probably isn't the best way to fix the problem." - in fact this may make it worse elsewhere as you're now splitting up rows with the same category+extra but different bodies into multiple rows.

::My looks to have the same issue. (Reports is okay, as only one body.)

Really, the code here in Around and My should be de-duplicating. It looks like the only reason extra is fetched is so optional help text can be shown (on KiitC, and there's no interface for it), so a small loop that just takes the first of a particular category and drops subsequent of the same would do?

davea:
Yep, that's a better idea. Hopefully e51c372e83b19efdc08bfd03a841a86dc2661f45 does the trick.

dracos (approved):
It'd be good to have a test, though :)

davea:
Test added in ffe7f1bbfe91062d0f4dad076c4823e30102eeb6 - does that look OK?

dracos:
Looks good!

PR #3170, opened 2020-08-28
Non map refactoring bits

dracos:
Hopefully no actual changes, just some small tidyings from the Bromley waste branch, and factoring out some functions so they can be used later (e.g. working day calculator, and non-map report creation).
[skip changelog]


davea (approved):
Nice bit of tidying up 👍 

(some tests for `FixMyStreet::WorkingDays` would be good though, couldn't see any in `zurich.t`)

dracos:
"(some tests for FixMyStreet::WorkingDays would be good though, couldn't see any in zurich.t)" - it is covered by the check overdue part of that, but no harm in some tests just for it, will do.

PR #3173, opened 2020-09-03
[Bromley] Update wizard stuff with updates from parallel branch

dracos:
This takes the current form wizard commit from the noise branch (where I had originally taken the code from this branch and then adapted it lots) and refactors it back into this branch as I think it's nicer. Each form gets its own package, the meta saved-data/process stuff is all self contained within the Wizard package, and so the controller is much smaller (as the first page of each form has dynamic fields, there's still a bit of code for that first Page, but hopefully that is clear). It's possible it might change further, but this seemed an okay point to cross it back.
[skip changelog]


struan (approved):
👍

PR #3174, opened 2020-09-03
[Hackney] Noise prototype

dracos:
Uses same govuk stuff as waste for ease/sharing, but that can be changed/refactored if needed.
The new test script should show the main processes of what it's doing.
Code-wise, after development first on waste and then here, I think this is now quite a nice basic solution - it adds a new Wizard form, which is then subclassed to provide the pages/fields required for the form, most things are self contained in there.
The controller has a few little bits but is mostly just passing things to the form and loading up the next form if it validates, and dealing with the final result (currently it just saves everything, so that will need work once we know more what it should do).
[skip changelog]


chrismytton (approved):
📢 

Looks good! Thought the code might be a bit more complex, but it's actually very readable! 👍

PR #3176, opened 2020-09-10
[TfL] auto close reports

struan:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1955

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/TfL/AutoClose.pm
+++ b/perllib/FixMyStreet/Script/TfL/AutoClose.pm
@@ -1,0 +1,129 @@
+package FixMyStreet::Script::TfL::AutoClose;
+
+use v5.14;
+
+use Moo;
+use CronFns;
+use FixMyStreet;
+use FixMyStreet::Cobrand;
+use FixMyStreet::DB;
+use Types::Standard qw(InstanceOf Maybe);
+
+has commit => ( is => 'ro', default => 0 );
+
+has verbose => ( is => 'ro', default => 0 );
+
+has body => (
+    is => 'lazy',
+    isa => Maybe[InstanceOf['FixMyStreet::DB::Result::Body']],
+    default => sub {
+        my $body = FixMyStreet::DB->resultset('Body')->find({ name => 'TfL' });
+        return $body;
+    }
+);
+
+has days => (
+    is => 'ro',
+    default => 28
+);
+
+sub close {
+    my $self = shift;
+
+    die "Can't find body\n" unless $self->body;
+    warn "DRY RUN: use --commit to close reports\n" unless $self->commit;
+    my $categories = $self->categories;
+    $self->close_reports($categories);
+}
+
+has newest => (
+    is => 'lazy',
+    isa => InstanceOf['DateTime'],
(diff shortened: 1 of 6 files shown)


dracos (changes requested):

perllib/FixMyStreet/Script/TfL/AutoClose.pm
```diff
@@ -0,0 +1,102 @@
...
+        category => { -in => [ keys %$categories ] },
+        'me.state' => 'action scheduled',
+        bodies_str => $self->body->id,
+        'comments.state' => 'confirmed',
+        'comments.problem_state' => 'action scheduled',
+    },
+    {
+        group_by => 'me.id',
+        join => [ 'comments' ],
+        having => \[ 'MIN(comments.confirmed) < ?', $dtf->format_datetime($self->newest) ]
```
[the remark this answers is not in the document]
struan: I had originally just used the lastupdate of the problem which is problematic in the other direction.

fetching all the the updates does seem like the only way to resolve this though. I guess with this query as the basis then at least you are only considering reports where there is a chance that they are valid so it's not too awful.

perllib/FixMyStreet/Script/TfL/AutoClose.pm
```diff
@@ -84,6 +84,20 @@ sub close_reports {
 
     my $count = 0;
     for my $r ( $reports->all ) {
+        my $comments = FixMyStreet::DB->resultset('Comment')->search(
+            { problem_id => $r->id },
+            { order_by => 'confirmed' }
+        );
+        my $earliest;
+        while ( my $c = $comments->next ) {
+            warn $c->problem_state . "\n";
```
dracos, line 93: Debug

perllib/FixMyStreet/Script/TfL/AutoClose.pm
```diff
@@ -84,6 +84,20 @@ sub close_reports {
...
+            if ( $c->problem_state ne 'action scheduled' ) {
+                $earliest = undef;
+                next;
+            }
+            $earliest = $c->confirmed unless defined $earliest;
+        }
+        next unless defined $earliest && $earliest < $self->newest;
```
dracos, line 100: newest could be a lazy has attribute, so that it's only calculated once?

dracos (changes requested):
Something odd happening in the tests, now they're running - failing sometimes, I think the test is assuming the order it will get a result in or something like that. Other than that, only one debug and one suggestion :)

perllib/FixMyStreet/Script/TfL/AutoClose.pm
```diff
@@ -36,6 +36,17 @@ sub close {
     $self->close_reports($categories);
 }
 
+has newest => (
+    is => 'lazy',
+    isa => Maybe[InstanceOf['DateTime']],
```
dracos, line 41: Why Maybe, isn't it always a DateTime? :)

dracos (approved):
Looks good; one query about why something that I think is always set is a Maybe is all :)

PR #3180, opened 2020-09-15
"Search tips" message when there are no location search results

zarino:
Improvements to the message you get when your location search returns no results (usually because you’ve made a spelling mistake, or have used too ambiguous a street name), on fixmystreet.com and the UK council cobrands.

Screenshot_2020-09-21 Screenshot(1)

I’ve included a customised version for each cobrand, to match the layout of their respective homepages, eg:

Screenshot_2020-09-21 Screenshot(2)

The easiest way to test this out locally is to force all location searches to return as if there had been no results, eg, by updating the `determine_location_from_pc` function in `Location.pm` so that it returns early like this:

```
sub determine_location_from_pc : Private {
    my ( $self, $c, $pc ) = @_;

    # temporary redirect for testing out the "search tips" message
    $c->stash->{location_error_pc_lookup} = 1;
    $c->stash->{location_error} = 'Sorry, we could not find that location.';
    return;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
         - Rewrite open311-update-reports to share code and improve functionality.
     - UK:
         - Add option for recaptcha. #3050
+        - Display search tips when location search returns no results. #3180
 
 * v3.0.1 (6th May 2020)
     - New features:
--- a/templates/web/fixmystreet.com/around/location_error.html
+++ b/templates/web/fixmystreet.com/around/location_error.html
@@ -1,0 +1,31 @@
+[% IF location_error_pc_lookup %]
+
+    <div class="search-help">
+        <h2 class="search-help__header" role="alert">
+            [% location_error | safe %]
+        </h2>
+
+        <div class="search-help__tips">
+            <div class="search-help__tips__category">
+                <h3>[% loc('Searching by postcode?') %]</h3>
+                <ul>
+                    <li>[% loc('Check you <strong>haven’t swapped numbers and letters</strong>. <code>O</code>, <code>0</code>, <code>I</code> and <code>1</code> aren’t the same.') %]</li>
+                    <li>[% loc('<strong>Don’t forget the space</strong> in your postcode.') %]</li>
+                    <li>[% loc('<strong>Don’t mix postcodes and street names.</strong>') %]</li>
+                </ul>
+            </div>
+            <div class="search-help__tips__category">
+                <h3>[% loc('Searching by street name?') %]</h3>
+                <ul>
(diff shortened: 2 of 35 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
         - Add aerial maps option to Bing and OSM maps.
         - Select matches for both filter category and group. #3110
         - Add an extra zoom level to most map types. #3130
+        - Display search tips when location search returns no results. #3180
```
dracos, line 15: Probably should be in UK section further down now.

web/cobrands/zurich/_colours.scss
```diff
@@ -26,3 +26,5 @@ $mobile_menu_tab_bg_col: #FFFFFF; // the white border and tab on mobile site
 $mobile_header_blue: #366AB6; // close match to Zurich logo_portal.jpg
 
 $header-top-border: solid 0.25em $mobile_menu_tab_bg_col;
+
+$search-help-margin: 0 -1em;
```
dracos, line 30: Is this needed if it's not included on Zurich

templates/web/fixmystreet.com/around/location_error.html
```diff
@@ -0,0 +1,31 @@
...
+        <h2 class="search-help__header" role="alert">
+            [% location_error | safe %]
+        </h2>
+
+        <div class="search-help__tips">
+            <div class="search-help__tips__category">
+                <h3>[% loc('Searching by postcode?') %]</h3>
+                <ul>
+                    <li>[% loc('Check you <strong>haven’t swapped numbers and letters</strong>. <code>O</code>, <code>0</code>, <code>I</code> and <code>1</code> aren’t the same.') %]</li>
+                    <li><strong>[% loc('Don’t forget the space</strong> in your postcode.') %]</li>
```
dracos, line 13: The translation should include the opening tag as well - e.g. if a translation needed to move the strong to the end of the phrase. But as this is .com you could not have any translation strings at all at present.

dracos (changes requested):
Looks good; couple of tiny suggested changes.

CHANGELOG.md
zarino: Good idea!

web/cobrands/zurich/_colours.scss
zarino: Nope, you’re right, not needed any more. Will remove.

templates/web/fixmystreet.com/around/location_error.html
zarino: Ooops, typo. Will fix.

dracos (approved):
🚩

PR #3188, opened 2020-09-22
[Central Beds] Add highways layer and prevent reports made off CBC roads

davea:
Adds an invisible highways layer which is used to check ownership of roads and prevents reports that aren't on CBC-responsible roads.

[skip changelog]

### DIFF
--- a/web/cobrands/centralbedfordshire/assets.js
+++ b/web/cobrands/centralbedfordshire/assets.js
@@ -1,0 +1,62 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+
+var defaults = {
+    http_options: {
+        url: fixmystreet.staging ? "https://tilma.staging.mysociety.org/mapserver/centralbeds" : "https://tilma.mysociety.org/mapserver/centralbeds",
+        params: {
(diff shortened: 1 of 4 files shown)


web/cobrands/centralbedfordshire/assets.js
```diff
@@ -0,0 +1,62 @@
...
+    always_visible: true,
+    road: true,
+    all_categories: true,
+    usrn: {
+        attribute: 'streetref1',
+        field: 'NSGRef'
+    },
+    actions: {
+        found: function(layer, feature) {
+            fixmystreet.message_controller.road_found(layer, feature, function(feature) {
```
struan, line 47: Could we filter this at the server side to only return adopted roads and then could use the standard road_found function? Not sure it makes much difference either way I guess.

struan (approved):
🛣️

web/cobrands/centralbedfordshire/assets.js
davea: That's a good idea, the less data transferred the better. The current filters and behaviour are subject to change based on client feedback, so when we've got an idea of the final requirements I'll revisit the server-side filters.

PR #3189, opened 2020-09-22
[Central Beds] Include extra params for Symology when POSTing Open311 updates

davea:
[skip changelog]


davea:
(Not sure what that test failure is about; it's OK locally)

dracos (approved):
Probably not worth doing yet, but perhaps a note to say is same as Bexley's, in terms of potentially having a Symology Role at some point later (assuming it's the same for that reason I mean :) )

PR #3190, opened 2020-09-22
[Noise] Add working address lookup.

dracos:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Noise.pm
+++ b/perllib/FixMyStreet/App/Form/Noise.pm
@@ -152,7 +152,8 @@ has_page address => (
             $options = $form->previous_form->addresses;
         } else {
             my $saved_data = $form->saved_data;
-            $options = $form->c->cobrand->addresses_for_postcode($saved_data->{postcode});
+            my $data = $form->c->cobrand->addresses_for_postcode($saved_data->{postcode});
+            $options = $data->{addresses};
             push @$options, { value => 'missing', label => 'I can’t find my address' };
         }
         return { address => { options => $options } };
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -59,6 +61,41 @@ sub geocoder_munge_results {
...
+    $ua->default_header(x_api_key => $key);
+
+    my $pages = 1;
+    my @addresses;
+    for (my $page = 1; $page <= $pages; $page++) {
+        my $res = $ua->get($url . '&page=' . $page);
+        my $data = decode_json($res->decoded_content);
+        $pages = $data->{data}->{page_count};
+        foreach my $address (@{$data->{data}->{address}}) {
+            next unless $address->{locality} eq 'HACKNEY';
```
struan, line 82: I wonder if there's something to be said for flagging when the postcode is outside hackney. Rather than just saying "we can't find the address"

t/app/controller/noise.t
```diff
@@ -25,11 +29,13 @@ FixMyStreet::override_config {
         $mech->submit_form_ok({ with_fields => { name => "Test McTest", email => $user->email, phone => '01234 567890' } });
         $mech->submit_form_ok({ with_fields => { best_time => [['day', 'evening'], 1], best_method => 'email' } });
         $mech->submit_form_ok({ with_fields => { postcode => 'SW1A 1AA' } });
```
struan, line 31: Should this not check that the results don't include the ELSEWHERE address?

struan: Plus a test for not matching any addresses.

perllib/FixMyStreet/App/Form/Noise.pm
```diff
@@ -303,23 +295,23 @@ has_page source_known_address => (
...
-            # Need to perform same lookup as UPRN form...
-            $options = [
-                { value => 'house1', label => 'House 1' },
-                { value => 'house2', label => 'House 2' },
-                { value => 'house3', label => 'House 3' },
-                { value => 'missing', label => 'I can’t find the address' },
-            ];
+            my $saved_data = $form->saved_data;
+            $options = $form->c->cobrand->addresses_for_postcode($saved_data->{source_postcode});
+            push @$options, { value => 'missing', label => 'I can’t find my address' };
```
struan, line 300: I notice if they select this option then the confirmation form at the end has "missing" as their address rather than what they put in the box on the What is your Address page.

struan (changes requested):
Few places where the tests could do with expanding, plus some questions/suggestions.

struan (approved):
👍

PR #3191, opened 2020-09-23
Improved email phone in form

dracos:
This does three things:
1. only shows username errors on the form where they were triggered (ie. sign in form or 'by email' form) which should fix https://github.com/mysociety/fixmystreet-commercial/issues/1993
2. If SMS_AUTHENTICATION is set, changes the new report form so that it shows email/phone, asks you to pick between them and uses your choice as the 'username' for account purposes, which should fix https://github.com/mysociety/fixmystreet-commercial/issues/1996.

image

And then one I am not sure about:
3. It splits the dual-use username field into username (signing in form) and username_register (confirm by code). I think this might help with obscure errors where e.g. if new report form submission to server has a problem, it would display the username in both fields client-side, which would then cause an issue on next submission if you only altered the one you could see, not realising there was another. I did not rename username to username_sign_in (I did get halfway through, sigh) because it starts to interact with the auth code e.g. with 2FA and change password etc and I didn't want to affect them (I mean, hopefully I haven't!).


templates/web/base/report/form/user_loggedout_by_email.html
```diff
@@ -25,19 +23,44 @@
...
+    <p class="segmented-control segmented-control--radio">
+        <input type="radio" name="update_method" id="update_method_email"
+             data-show="#js-optional-phone" data-hide="#js-optional-email"
+            value="email"[% ' checked' IF update_method == 'email' %]>
+        <label class="btn" for="update_method_email">[% loc('Email') %]</label>
+        <input type="radio" name="update_method" id="update_method_phone"
+             data-show="#js-optional-email" data-hide="#js-optional-phone"
+            value="phone"[% ' checked' IF update_method == 'phone' %]>
+        <label class="btn" for="update_method_phone">[% loc('Phone') %]</label>
+    </p>
```
struan, line 43: This doesn't change the form validation so you can select phone here, not fill in a phone number but fill in an email and it submits fine. Similarly if you don't fill in anything it says you need an email even though the label says optional.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -126,8 +126,10 @@ sub report_new_ajax : Path('mobile') : Args(0) {
 
     # Apps are sending email as username
     # Prepare for when they upgrade
-    if (!$c->get_param('username')) {
-        $c->set_param('username', $c->get_param('email'));
+    my $username_field = ( $c->get_param('submit_sign_in') || $c->get_param('password_sign_in') )
+        ? 'username': 'username_register';
```
struan, line 130: username_register is confusing as I expect that to be the one to use when they are registering which is the opposite of the case. maybe username_logged_in?

struan (changes requested):
I've not fully reviewed this but there's at least one bug so I might as well point that out first.

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: It matches password_register? It is the one they are using when they are registering (whilst making a new report).

templates/web/base/report/form/user_loggedout_by_email.html
dracos: I can't reproduce this - if I select phone, leave phone blank, fill in email, the page reloads and the error is "Please enter your email" as per your second point.

perllib/FixMyStreet/App/Controller/Report/New.pm
struan: It's also what I see when I'm already logged in though which is what's confusing me.

templates/web/base/report/form/user_loggedout_by_email.html
dracos: Oh you mean submits fine as in it doesn't change the HTML5 required attributes, right.

struan: Yes, I could have been a bit clearer there :)

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: Have set it back to username for logged in form.

templates/web/base/report/form/user_loggedout_by_email.html
dracos: Have added JS bit that should set required as needed.

struan: I'm not sure what's happening with this but that doesn't seem to be working. The code is definitely running but doesn't seem to be having an effect. Also, if the JS validation doesn't kick in then the non JS validation is getting quite confused:

Screenshot 2020-09-25 at 11 34 32

Ah, it looks like `fem.attr('required', true);` is the correct incantation.

I also think the display of the validation for the "How would you like to receive updates" bit needs some design love:

Screenshot 2020-09-25 at 11 25 50

struan (changes requested):
Few design things, some JS issues, and non JS validation issues

templates/web/base/report/form/user_loggedout_by_email.html
dracos: Oh, this is me changing it to be jquery in the JS rather than direct in the HTML but not checking it after that, sorry.

The double error - yeah, I saw that happened but I didn't really want to change it, it seemed... hard. It's technically not confused because it dates from 01311af63412f39b75ff3b12bb6dc4051eab13b3 where a staff user using report-as-another to only enter a bad phone number in which case the error message needs to be attached to 'phone' field, but if someone is logging in with a bad phone number, needs to attach to 'username' field - and so here as it displays username error at top and then phone error by phone you get it twice. I couldn't see an easy way to fix this.

Will fix the required thing, whoops.

struan (changes requested):
Sorry, one final thing, there's not a template set for sms in `redirect_or_confirm_creation` so you get an error page as it tries to display the `report/new` template and doesn't have all the information.

struan (approved):
The error problem with SMS failing is fixed in https://github.com/mysociety/fixmystreet/pull/3177/commits/fe39d461ed6952ccb75cab613f134f32e2bc3bc5 so the rest of this looks ok.

PR #3193, opened 2020-09-24
[IslandRoads] use standard confirm validation

struan:
[skip changelog]


davea (approved):
`[IsleOfWight]` seems to be the predominant tag for commit messages, otherwise 👍

PR #3194, opened 2020-09-29
do not trigger duplicate check with stopper check

struan:
The stopper checking code is triggered using the category_change event
which, by default also calls the duplicate check code. This meant that
when an extra question dropdown with an associated stopper was updated
the duplicate check re-ran and the duplicate reports panel was re-shown.

The invocation of category_change on the extra questions now includes
the skip_duplicates parameter to avoid this.

Fixes mysociety/fixmystreet-commercial#2007

### DIFF
--- a/templates/web/borsetshire/footer_extra_js.html
+++ b/templates/web/borsetshire/footer_extra_js.html
@@ -2,4 +2,8 @@
     version('/cobrands/fixmystreet-uk-councils/js.js'),
     version('/cobrands/borsetshire/google_maps_style.js'),
     version('/cobrands/borsetshire/js.js'),
-) %]
+);
+IF bodyclass.match('mappage');
+   scripts.push( version('/cobrands/fixmystreet/assets.js') );
+END;
+%]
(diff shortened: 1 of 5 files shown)


templates/web/borsetshire/footer_extra_js.html
```diff
@@ -2,4 +2,5 @@
     version('/cobrands/fixmystreet-uk-councils/js.js'),
     version('/cobrands/borsetshire/google_maps_style.js'),
     version('/cobrands/borsetshire/js.js'),
+    version('/cobrands/fixmystreet/assets.js'),
```
dracos, line 5: I think you only want to be adding this on map pages, see other cobrands, otherwise it throws a JS error on other pages.

dracos: I'm not clear why this addition was needed, either.

dracos (changes requested):
All looks fine, but addition of assets.js breaks the non-map pages.

templates/web/borsetshire/footer_extra_js.html
struan: It's there because otherwise the stopper code isn't included and I can't test the bug :)

dracos: Oh, so, should we think about having this code factored out and be able to be included without assets? Seems odd if it's possible to use entirely non-asset related, and people can set "disable the form" in the admin but wouldn't know this had to be included... Something to ticket anyway.

dracos (approved):

PR #3197, opened 2020-09-30
[Lincolnshire] Add bus stops asset layer

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/2011

[skip changelog]

### DIFF
--- a/web/cobrands/lincolnshire/assets.js
+++ b/web/cobrands/lincolnshire/assets.js
@@ -26,6 +26,12 @@ fixmystreet.assets.add(defaults, {
 });
 
 fixmystreet.assets.add(defaults, {
+    wfs_feature: "Public_Transport_Stops",
+    asset_category: [ "Bus stop", "Bus shelter" ],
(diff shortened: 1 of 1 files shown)


web/cobrands/lincolnshire/assets.js
```diff
@@ -25,6 +25,13 @@ fixmystreet.assets.add(defaults, {
     asset_item: 'bollard'
 });
 
+fixmystreet.assets.add(defaults, {
+    wfs_url: "https://tilma.staging.mysociety.org/mapserver/lincs",
```
davea, line 29: This line will be removed once the layers are on live tilma obvs :)

dracos: If staging has all the layers on as well, you could copy what other assets.js do and set the server based upon fixmystreet.staging, then you don't need to remember (well, remember this, you do need to remember to put the assets on live as well then).

dracos (approved):
One suggestion about staging is all

Issue #3195, opened 2020-09-29
Improve response priority in inspect form


dracos:
Currently deleted priorities are shown in the front end inspector form dropdown, marked as disabled. I think it would be preferable not to show disabled priorities at all (unless the report is assigned a priority that is now deleted).
There are also small bugs e.g. outputting `selecteddisabled` if a priority is both deleted and the default/selected, and outputting multiple `selected` if the current priority is not the default one.

PR #3198, opened 2020-10-01
Do not display deleted priorities in inspect form.

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/3195


davea (approved):
It's vaguely weird that moving a report from a deleted to a not-deleted priority isn't undoable as the option will have disappeared (although it wasn't possible before as it was disabled I suppose!). But I think that's not a very common occurrence so not worth doing anything about.

👍

PR #3199, opened 2020-10-02
Record whether report made on desktop or mobile.

dracos:
Have two submit buttons, one of which displays at <48em, one of which displays at >=48em. Record desktop/mobile with report depending on which button is clicked.
If neither button is clicked, e.g. if they submit the form by hitting return (although spec says button should be clicked then? whatever), have a bit of fallback JS that sets the form field in the browser resize function.
If no-JS and no form button clicked, carry on storing nothing as now.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -29,8 +29,8 @@ Create a new report, or complete a partial one.
 
 submit_problem: true if a problem has been submitted, at all.
 submit_sign_in: true if the sign in button has been clicked by logged out user.
-submit_register: true if the register/confirm by email button has been clicked
-by logged out user.
+submit_register(_mobile): true if the register/confirm by email button has been clicked
+by logged out user, or submit button clicked by logged in user.
 
 =head2 location (required)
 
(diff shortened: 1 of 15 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1046,7 +1046,13 @@ sub process_report : Private {
     $report->detail( $detail );
 
     # mobile device type
-    $report->service( $params{service} ) if $params{service};
+    if ($c->get_param('submit_register')) {
```
zarino, line 1049: Small point, but should we handle these via the `$params` hash, as we do for everything else at this point?

zarino (approved):
Looks good to me.

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: You'll see that, in the main, params is for all the variables, and get_param is used for submit buttons (e.g. report_anonymously, submit_sign_in bit above). Don't think it's totally consistent, but whatever :)

dracos:
Reordered New.pm because just realised that mobile app does use submit_register at present, so that would have then overridden the service field actually provided, so using that in preference if it's set first.

PR #3200, opened 2020-10-02
Don’t strip all spaces from Open311 categories.

### LINKED ISSUE #3167: Don't strip all spaces from 'email' in category admin
The line `$email =~ s/\s+//g;` means that every time someone edits a category that has an Open311 code with a space in it, the spaces would get stripped. Should that only be stripping from beginning/end?

dracos:
If a category’s send method is Open311, only strip spaces from the
ends of the code. We are aware of active Open311 servers that have
codes with spaces in the middle.

Fixes #3167.


struan (approved):
👾

PR #3201, opened 2020-10-02
avoid caching the service worker for logged in users

struan:
As the contents vary if the user is logged in and depending on their
permissions we should not be caching the service worker. This sets the
Vary header to 'Cookie' to avoid this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
         - Do not strip spaces from middle of Open311 category codes. #3167
         - Show all category history even if category renamed.
         - Fix email alert on initial update template.
+        - Do not cache the service worker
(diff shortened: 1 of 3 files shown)


dracos (approved):
I think this is fine, though caching always confuses me :) An alternative would be to Cache-Control: private or Cache-Control: max-age=0 (the latter apparently officially endorsed for SW: https://developers.google.com/web/updates/2019/09/fresher-sw ). But I think caching it for non-logged in people for a period of time (our reverse proxy would be an hour) is okay for our case, can always change in future. I assume it fixes your issue :)

struan:
Will squash this down to one commit before merging

dracos (approved):

PR #3202, opened 2020-10-02
[Central Beds] SocietyWorks footer with FMS Pro logo

zarino:
@LouiseMySociety @davideaton2 How does this look as a consistent way we could display branding at the bottom of council cobrands? The downside is we’re adding _another_ footer to the pages. But the upside is we don’t have to mess around with the content of the council’s existing footer, or with the homepage content. I think it’s worth it.

Screenshot_2020-10-02 FixMyStreet

I’ve written the styling such that the logo can be easily swapped out for a SocietyWorks one, if FMS Pro isn’t appropriate.

[skip changelog]


davideaton2:
I totally agree it's worth it, looks great

davea (approved):
Looks good!

PR #3203, opened 2020-10-05
[Bromley] Nuanced template resolution codes.

dracos:
Have the response template external status code be a combination
of resolution code, task type ID, and task state. [skip changelog]


davea (approved):
`%p`/`%P` aside, looks good 👍

PR #3205, opened 2020-10-05
[Central Beds] Include Symology area code param for Open311 reports

davea:
Adds a new Central Beds-specific `COBRAND_FEATURE` that maps between MapIt area codes and CBC's three highways teams areas.

[skip changelog]


dracos (approved):
🗺️

PR #3206, opened 2020-10-06
[UK] Prevent duplicate extra fields when blank.

dracos:
If a report had an existing, but empty, extra field that was looked up
on the server when being sent by Open311, the looked up information
would be added as an additional extra field, rather than replacing the
blank one. [skip changelog]


chrismytton (approved):
🚀

PR #3207, opened 2020-10-06
[Noise] check if the noise follows a pattern

struan:
Add an extra question to the when screen to check if there's a pattern
to when the noise occurs, and only then ask about when, otherwise ask
for more information about when. To avoid people filling in all the day
and time boxes if there's no pattern.

For mysociety/fixmystreet-commercial#2015

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Noise.pm
+++ b/perllib/FixMyStreet/App/Form/Noise.pm
@@ -414,9 +414,19 @@ has_field radius => (
 );
 
 has_page when => (
-    fields => ['happening_now', 'happening_days', 'happening_time', 'continue'],
+    fields => ['happening_now', 'happening_pattern', 'continue'],
     title => 'When does the noise happen?',
-    next => 'more_details',
+    next => sub { $_[0]->{happening_pattern} ? 'happening_time' : 'happening_description' },
+    post_process => sub {
+        my $form = shift;
+        my $saved_data = $form->saved_data;
+        if ($saved_data->{happening_pattern}) {
+            delete $saved_data->{happening_description};
+        } else {
+            delete $saved_data->{happening_days};
+            delete $saved_data->{happening_time};
+        }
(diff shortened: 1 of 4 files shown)


templates/web/base/noise/index.html
```diff
@@ -21,6 +21,12 @@
     if (id) { id.className += ' hidden-js'; }
     id = document.getElementById('form-estates-row');
     if (id) { id.className += ' hidden-js'; }
+    id = document.getElementById('form-happening_days-row');
+    if (id) { id.className += ' hidden-js'; }
+    id = document.getElementById('form-happening_time-row');
+    if (id) { id.className += ' hidden-js'; }
+    id = document.getElementById('form-happening_description-row');
+    if (id) { id.className += ' hidden-js'; }
```
dracos, line 29: Might be worth having an array and a loop now :)

perllib/FixMyStreet/App/Form/Noise.pm
```diff
@@ -414,7 +414,7 @@ has_field radius => (
 );
 
 has_page when => (
-    fields => ['happening_now', 'happening_days', 'happening_time', 'continue'],
+    fields => ['happening_now', 'happening_pattern', 'happening_days', 'happening_time', 'happening_description', 'continue'],
```
dracos, line 417: If we're asking a question, and the answer leads to two entirely different questions appearing, I think this would be better done as two different pages, so yes to happening_pattern would take you to a page with days and time on, and no would take you to a page with description on, and both would then take you to more_details; that way it's smaller per-page, and doesn't confuse if the JS doesn't work and all the questions get shown.

dracos (changes requested):
I suggest moving the subsequent questions to new pages rather than with JS on the same page, but that's all!

dracos (approved):
Looks good

PR #3208, opened 2020-10-07
Show all category history even if category renamed

### LINKED ISSUE #3181: Contact history in admin goes off name, not ID
If a category is renamed, it does not show the history prior to the renaming.

dracos:
Fixes #3181.


davea (approved):
🔀 ⏳ 🆗

PR #3209, opened 2020-10-08
Fix email alert on initial update template.

dracos:
A report's confirmation timestamp uses current_timestamp, and so
includes microseconds. An initial update text, to fit in with the
Open311 handling of updates, uses a DateTime object, which does not.
This means if a report is created when logged in, the initial update
can have a timestamp earlier than the report, and so is not alerted on.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1736,7 +1736,10 @@ sub create_related_things : Private {
         my $request = {
             service_request_id => $problem->id,
             update_id => 'auto-internal',
-            comment_time => DateTime->now,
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1736,7 +1736,7 @@ sub create_related_things : Private {
         my $request = {
             service_request_id => $problem->id,
             update_id => 'auto-internal',
-            comment_time => DateTime->now,
+            comment_time => DateTime->now->add( seconds => 1 ),
```
struan, line 1739: I think a comment to indicate what's going on here would be helpful.

struan (approved):

PR #3210, opened 2020-10-08
[Noise] ask directly for location rather than if know it

struan:
Instead of having a separate flow for known/unknown location have a
location search box and then show a list of addresses if they enter a
Hackney postcode, or the map page if we can't get a list of addresses.

For https://github.com/mysociety/fixmystreet-commercial/issues/2013

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Noise.pm
+++ b/perllib/FixMyStreet/App/Form/Noise.pm
@@ -223,9 +223,9 @@ has_field kind_other => (
 );
 
 has_page where => (
-    fields => ['where', 'estates', 'address_known', 'continue'],
+    fields => ['where', 'estates', 'source_location', 'continue'],
     title => 'Where is the noise coming from?',
-    next => sub { $_[0]->{address_known} ? 'source_known_postcode' : $_[0]->{latitude} ? 'map': 'address_unknown' },
+    next => sub { $_[0]->{source_addresses} ? 'source_known_address' : $_[0]->{latitude} ? 'map': 'address_unknown' },
 );
 
 has_field where => (
@@ -251,21 +251,41 @@ has_field estates => (
     ]
 );
 
-has_field address_known => (
-    type => 'Select',
-    widget => 'RadioGroup',
+has_field source_location => (
     required => 1,
-    label => 'Do you know the address of the source of the noise, and is it in Hackney?',
-    options => [
-        { label => 'Yes', value => 1 },
-        { label => 'No', value => 0 },
-    ],
-);
-
-has_page source_known_postcode => (
-    fields => ['source_postcode', 'find_address'],
-    title => 'The source of the noise',
-    next => 'source_known_address',
+    type => 'Text',
+    label => 'Postcode, or street name and area of the source',
+    tag => { hint => 'If you know the postcode please use that' },
+    validate_method => sub {
+        my $self = shift;
+        my $c = $self->form->c;
+        return if $self->has_errors; # Called even if already failed
+        my $value = $self->value;
+        $value =~ s/[^A-Z0-9]//i;
+        my $pc = mySociety::PostcodeUtil::canonicalise_postcode($value);
+        if (mySociety::PostcodeUtil::is_valid_postcode($pc)) {
+            my $data = $self->form->c->cobrand->addresses_for_postcode($pc);
+            if ($data->{addresses} && @{$data->{addresses}}) {
+                my $options = $data->{addresses};
+                push @$options, { value => 'missing', label => 'I can’t find my address' };
+                my $saved_data  = $self->form->saved_data;
+                $saved_data->{source_addresses} = 1;
+                $self->form->addresses($options);
+                return 1;
+            }
+        }
+        my $ret = $c->forward('/location/determine_location_from_pc', [ $self->value ]);
+        if (!$ret) {
+            if ( $c->stash->{possible_location_matches} ) {
+                $self->add_error('Multiple matches - deal somehow');
+            } else {
+                $self->add_error($c->stash->{location_error});
+            }
+        }
+        my $saved_data = $self->form->saved_data;
+        $saved_data->{latitude} = $c->stash->{latitude};
+        $saved_data->{longitude} = $c->stash->{longitude};
+    },
 );
 
 sub check_postcode {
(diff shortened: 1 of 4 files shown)


t/app/controller/noise.t
```diff
@@ -61,7 +79,40 @@ FixMyStreet::override_config {
...
+        $mech->submit_form_ok({ with_fields => { latitude => 51.549239, longitude => -0.054106, radius => 'medium' } });
+        $mech->submit_form_ok({ with_fields => {
+            happening_now => 0,
+            happening_days => [['tuesday'], 1],
+            happening_time => [['morning','evening'], 1],
+        } });
+        $mech->submit_form_ok({ with_fields => { more_details => 'Details' } });
+        # Check going back skips the geocoding step
+        $mech->submit_form_ok({ form_number => 3, fields => { goto => 'where' } });
+        $mech->submit_form_ok({ with_fields => { where => 'residence', source_location => 'A street' } });
```
dracos, line 102: This could be a location with multiple results, to show that that works (it currently does not), to be the reverse of the other test.

perllib/FixMyStreet/App/Form/Noise.pm
```diff
@@ -283,13 +306,6 @@ sub check_postcode {
     $self->form->addresses($data);
 }
 
-has_field source_postcode => (
```
dracos, line 286: You've removed this, fine, but the source_known_address page still uses it to do the lookup if need be. So e.g. entering a Hackney postcode as source, pick one of the results, hit submit, currently get an error "'200002791497' is not a valid value" as it's not looked them up.

perllib/FixMyStreet/App/Form/Noise.pm
```diff
@@ -329,10 +345,41 @@ has_field source_address => (
...
+        my $form = shift;
+        my $saved_data = $form->saved_data;
+        my $locations = $saved_data->{possible_location_matches};
+        my $options = [];
+        for my $location ( @$locations ) {
+            push @$options, { label => $location->{address}, value => $location->{latitude} . "," . $location->{longitude} }
+        }
+        return { possible_location_matches => { options => $options } };
+    },
+    post_process => sub {
```
dracos, line 362: process is always called on a form, both on 'arrival' and submission, so this goes a bit odd when possible_location_matches is the list, and not the single result, putting "ARRAY" in lat and null in lon. Guess it doesn't matter as is overwritten on next page, but wonder if there was a cleaner way. I used Amhurst Road as my test location.

dracos (changes requested):
Some bugs to fix, looks like, but yes, I think this is a clearer flow.
I guess test didn't pick up the postcode submission bug because it always returns results even if given new data, could return nothing in that case.

dracos (changes requested):
This works well apart from one scenario, sorry - if you enter a source address (not a postcode) and go through that flow and then go back to there (or earlier) in the flow, when you get back to the address entry and don't change anything, it resets and you have to pick a place (if multiple locations) and pick from the map again. It should remember the location and keep using it if you haven't changed the search text, I guess.

(Postcode works okay I think because the uprn is still stored in the `source_address` saved data and so your choice is still selected when you go to that page.)

(You'll have to mock out the address lookup for future dev as it's currently down.)

struan:
I'm not sure I understand what the issue is. I think this does the correct thing:


Screen Recording 2020-10-28 at 11 50 16 2020-10-28 11_59_27

OTOH, it does seem that if you go back and enter a postcode it ignores that so I will fix that.

dracos:
Sorry I wasn't clear. "I'm not sure I understand what the issue is. I think this does the correct thing:" - if you moved the pin on the pin page, and then went back and clicked through (doesn't need to be a multiple-result search), the pin resets to the location result of the search result, not where you've subsequently placed the pin.

dracos (changes requested):
Afraid it still doesn't work properly, sorry. Three examples:
1. Enter Danby House. Pin on Danby House, in the right place. Pick an area size, click Go, get error that I haven't given a latitude or longitude.
2. Enter Well Street. Pick first result. Click "search for a different location". Enter Danby House (which has 1 result). Pin is still on Well Street, not Danby House.
3. Enter Danby House. Pin on Danby House. Click "search for a different location". Enter Well Street. Get taken straight to map, no disambiguation page, pin still on Danby House.

Feels like this is more complex than it needs to be. I sort of see the point of the source_lat/lon fields, but are they necessary? If it had its previous behaviour of storing lat/lon on one result, or chosen result, and if someone goes back and has a new search value in source_location or pc, that clears the values, is that not enough? When do you need to maintain the lat/lon if they've searched for a new search term? I may have missed something.

Could the `pc` field be dropped and that page reuse the `source_location` field? I know the postcode behaviour is different, but if that helps simplify it I don't think it would matter much.

struan:
Only thing I'm not sure on is at the moment if you put in an address that requires disambiguation, e.g Hackney Marshes, then on re-editing you don't get an option to change that, which I'm not sure is the right decision.

dracos:
Oh yes, I see what you mean; though you can click Back from the map page easily enough rather than "Search for a new location" I guess. But if you get to the end, then go back to edit, you mean. Hmm. How about if you always go to `source_known_address` or `choose_location` from `where` page, never straight to `map` (and presumably if you haven't changed the text your choice on `choose_location` will be remembered from `location_matches` field?) and then only if the `location_matches` changes would you delete lat/lon and lose the previous location. Does that work?

struan:
Ok, tweaked that.

perllib/FixMyStreet/App/Form/Noise.pm
```diff
@@ -329,32 +353,60 @@ has_field source_address => (
...
+        my $form = shift;
+        my $saved_data = $form->saved_data;
+        # if we previously used a postcode lookup we want to remove the location data
+        if ($saved_data->{source_address}) {
+            delete $saved_data->{source_address};
+        }
+        if ( my $location = $saved_data->{location_matches} ) {
+            my ($lat, $lon) = split ',', $location;
+            if ( !$saved_data->{latitude} && !$saved_data->{longitude} ) {
```
dracos, line 379: This if block can just be removed I think, the same stuff happens underneath.

dracos (changes requested):
🎉 Sorry for all the issues, but looks like it's all good now. One tidy up I think. And the PR test implies might need bit of hand-holding to merge okay.

dracos (approved):
📣 📣

PR #3211, opened 2020-10-08
Oxfordshire defect feedback

dracos:
After feedback, we construct the extra details string from various other bits being asked, with a JS preview.
Traffic management only applies to Oxfordshire, so put it in there until otherwise.
Couple of other minor UI improvements.

[skip changelog]


davea (approved):
👍

PR #3217, opened 2020-10-15
[Central Beds] Don't allow users to change report status

davea:
Removes the "this problem has been fixed" tickbox from the update form. Factors out the same behaviour for TfL and Hounslow into a cobrand feature, and adds tests for it.

For https://github.com/mysociety/fixmystreet-commercial/issues/2017
For https://github.com/mysociety/fixmystreet-commercial/issues/2018

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -304,6 +304,16 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+sub body_disallows_state_change {
+    my $self = shift;
+    my ($problem) = @_;
+    my $c = $self->{c};
+
+    my ($disallowed, $body) = $self->per_body_config('update_states_disallowed', $problem);
+    $disallowed //= 0;
+    return $disallowed;
+}
+
 sub problem_state_processed {
     my ($self, $comment) = @_;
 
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -304,6 +304,17 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+sub body_disallows_state_change {
+    my $self = shift;
+    my ($problem) = @_;
+    my $c = $self->{c};
+
+    my ($allowed, $body) = $self->per_body_config('update_states_allowed', $problem);
+    $allowed //= 0;
```
dracos, line 313: I assume you want to default to allowed, not disallowed (I think this is what causes the test failure in report_display.t

dracos (changes requested):
Fix for the default, I think.
Also, do we want to have a fixmystreet-uk-councils form_state_checkbox.html template that calls a similar function in UKCouncils.pm, so that be done in the same configuration rather than need a new blank template?

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
davea: Oops :)

davea:
Thanks for the review. I've rejigged things as you suggested to remove the need for an empty template file. It's now controlled by the `update_states_disallowed` cobrand feature flag.

dracos (approved):
Need to note somewhere (well, here I am doing it!), needs config adding when this is merged/deployed, which if I've understood right would be:
```
update_states_disallowed:
  fixmystreet:
    Central Bedfordshire: 1
    Hounslow: 1
    TfL: 1
  centralbedfordshire: 1
  highwaysengland: 1
  hounslow: 1
  tfl: 1
```

PR #3218, opened 2020-10-16
Fix unicode issue &c

dracos:
Issue in the new staff HTML thing in emails if extra happens to have Unicode in it anywhere.
And some other small fixes.
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Template.pm
+++ b/perllib/FixMyStreet/Template.pm
@@ -175,6 +175,7 @@ sub email_sanitize_text : Fn('email_sanitize_text') {
 
     my $text = $update->{item_text};
     my $extra = $update->{item_extra};
+    utf8::encode($extra) if $extra;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Template.pm
```diff
@@ -175,6 +175,7 @@ sub email_sanitize_text : Fn('email_sanitize_text') {
 
     my $text = $update->{item_text};
     my $extra = $update->{item_extra};
+    utf8::encode($extra);
```
davea, line 178: This (and L238) causes a `Use of uninitialized value in subroutine entry` warning when running tests in `alert_new.t` for some reason.

davea (approved):
One new warning aside, looks good 👍

perllib/FixMyStreet/Template.pm
dracos: Will add `if $extra` to end.

PR #3222, opened 2020-10-19
[Bexley] Add private comments field to report form

chrismytton:
This allows staff to add private comments to a report when creating it.

Your user will need to have the `contribute_as_body` permission in order to see this box.

open311-adapter changes are in https://github.com/mysociety/open311-adapter/pull/156

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1850

---

image

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -139,6 +139,11 @@ sub open311_extra_data_include {
         }
     }
 
+    # Add private comments field
+    push @$open311_only,
+        { name => 'private_comments', description => 'Private comments',
+          value => $row->get_extra_metadata('private_comments') || '' };
+
     return $open311_only;
 }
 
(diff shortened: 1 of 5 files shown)


davea:
There doesn't seem to be a way of viewing or editing these notes after the report is made (apart from viewing them in the **Extra data:** section of the report admin page), is that intentional? There is already the "Extra details" field on the inspect form (which is stored in `detailed_information`) - are the use cases different enough that they should be separate? Perfectly fine if they are of course, just want to make sure I'm understanding the problem being solved :)

chrismytton:
@davea They've only asked for it to appear in the backend, so I think it's fine that it can't be viewed/edited anywhere on the FMS side.

> These comments would be concatenated into to the main description boxes in Uniform/Symology and the box itself would have to be placed on the "tell us about you" page and only visible to back-end staff that are logged in.

(From https://github.com/mysociety/fixmystreet-commercial/issues/1850)

davea (approved):

perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -139,6 +139,11 @@ sub open311_extra_data_include {
         }
     }
 
+    # Add private comments field
+    push @$open311_only,
+        { name => 'private_comments', description => 'Private comments',
+          value => $row->get_extra_metadata('private_comments') };
```
dracos, line 145: This needs to default to the empty string if not present, otherwise it looks lke Open311.pm warns every time a private comment is undefined.

chrismytton:
Thanks for sorting that out, @dracos!

PR #3224, opened 2020-10-21
Enable per-category hint customisation.

dracos:
Lets you set the summary/detail hint message per category in the admin, which is then used in preference to the defaults in the form. A couple of small changes alongside (one a bugfix really, the other just putting parent categories alongside category rather than the other end of the form). Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2024


perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -283,6 +283,13 @@ sub update_contact : Private {
             $contact->unset_extra_metadata($_);
         }
     }
+    foreach (qw(title_hint detail_hint)) {
+        if ( my $value = $c->get_param($_) ) {
+            $contact->set_extra_metadata( $_ => $value );
```
davea, line 288: This allows you to hide the hint entirely by entering a string containing only whitespace - is that desirable? I'm not sure if it is so we should maybe trim the string before the test.

davea (changes requested):
Code-wise this is great, but needs some docs to point people at please.

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
dracos: I guess someone could decide they didn't want a hint to be shown? Hmm, no probably want to always have something, will do.

davea (approved):

PR #3225, opened 2020-10-21
[Oxfordshire] show defects on map

chrismytton:
Show defects from Oxfordshire's WFS feed on the /around page and as defects when creating a new report.

Part of https://github.com/mysociety/fixmystreet-commercial/issues/1836


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -284,4 +289,98 @@ sub dashboard_export_problems_add_columns {
...
+sub pins_from_wfs {
+    my ($self, $bbox) = @_;
+
+    my $wfs = $self->defect_wfs_query({
+        BBOX => $bbox,
+    });
+
+
+    my @features = grep {
+        $_->{properties}->{APPROVAL_STATUS_NAME} eq 'With Contractor';
```
davea, line 327: WFS allows server-side filtering of features which would remove the need for this grep (and hopefully speed up the WFS response time too). If you added a `filter` param to the WFS request that looked something like this it'd have the desired effect, I think:

```xml
<ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
    <ogc:And>
        <ogc:PropertyIsEqualTo matchCase="true">
            <ogc:PropertyName>APPROVAL_STATUS_NAME</ogc:PropertyName>
            <ogc:Literal>With Contractor</ogc:Literal>
        </ogc:PropertyIsEqualTo>
        <ogc:BBOX>
            <ogc:PropertyName>SHAPE_GEOMETRY</ogc:PropertyName>
            <gml:Envelope xmlns:gml="http://www.opengis.net/gml" srsName="EPSG:4326">
                <gml:lowerCorner>-1.3134050585937 51.80585466836</gml:lowerCorner>
                <gml:upperCorner>-1.2090349414063 51.84765563379</gml:upperCorner>
            </gml:Envelope>
        </ogc:BBOX>
    </ogc:And>
</ogc:Filter>
```

Might have to remove the newlines and other whitespace :) `filter` is mutually exclusive with the `BBOX` param, hence it being included within the `<And>` in the filter. You could lose `APPROVAL_STATUS_NAME` from the `propertyName` param too.

web/js/map-OpenLayers.js
```diff
@@ -513,7 +513,37 @@ $.extend(fixmystreet.utils, {
...
+        var problem_id = feature.attributes.id;
+
+        // Oxfordshire: Handle clicks on defects, which have negative fake IDs.
+        if (problem_id < 0) {
+            defect_popup = new OpenLayers.Popup.FramedCloud(
+                "occDefects",
+                feature.geometry.getBounds().getCenterLonLat(),
+                null,
+                feature.attributes.title.replace("\n", "<br />"),
+                { size: new OpenLayers.Size(0, 0), offset: new OpenLayers.Pixel(0, 0) },
```
davea, line 538: The overlapping of the popover and the marker is a bit odd to my eye:

image

If this was instead `{ size: new OpenLayers.Size(0, 0), offset: new OpenLayers.Pixel(6, -46) },` then the bubble could almost be imagined to be coming from the person in the icon:

image

web/js/duplicates.js
```diff
@@ -179,6 +179,43 @@
...
+            }
+        }
+
+        // Handle clicks on defect pins
+        var select_defect = new OpenLayers.Control.SelectFeature(
+            fixmystreet.markers,
+            {
+                hover: true,
+                clickFeature: function (feature) {
+                    close_defect_popup();
```
davea, line 199: There's quite a lot of duplication here and in `map-OpenLayers.js` - could it be factored together somehow? There's also the issue that this is quite Oxfordshire-specific but is in the main codebase. Having said that, I don't have a good suggestion for hooking in here from OCC JS because we're in a closed scope and not a function that the cobrand JS could override.

I think `fixmystreet.cobrand` would be available when this code is called - what about a cobrand check that then calls the factored-together code in the Oxfordshire JS?

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -284,4 +289,98 @@ sub dashboard_export_problems_add_columns {
...
+
+    my $uri = URI->new("https://tilma.mysociety.org/proxy/occ/nsg/");
+    $uri->query_form(
+        %$params,
+        REQUEST => "GetFeature",
+        SERVICE => "WFS",
+        SRSNAME => "urn:ogc:def:crs:EPSG::4326",
+        TYPENAME => "WFS_DEFECTS_FOR_QUERYING",
+        VERSION => "1.1.0",
+        propertyName => 'ITEM_CATEGORY_NAME,ITEM_TYPE_NAME,REQUIRED_COMPLETION_DATE,ACTUAL_COMPLETION_DATE,APPROVAL_STATUS_NAME,SHAPE_GEOMETRY',
```
davea, line 303: `ACTUAL_COMPLETION_DATE` isn't used anywhere and can presumably be removed.

davea (changes requested):
This is looking really good, just a few suggestions for making it snappier and a bit neater.

I noticed that the popup doesn't always disappear when clicking the map to place the report pin:

lingering_popup


Also something odd is going on with the date formatting as it's only showing the day of the week for completion in the UI (despite the tests that look for the correct format passing?):

image

image

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
chrismytton: Fixed in 75ca14a8cb5b11f2a48e3bc501159de4f83337b2

web/js/map-OpenLayers.js
chrismytton: Love the idea of it being a speech bubble!

Fixed in d170b80c0865b2c67e3f51d0647afa2283cc7498 and 0d45ab49e3ea6480340e9c839d70f842d2324d56.

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
chrismytton: Oh this is nice, thanks for the tip-off!

Done in 276f841b24e7dc947cc14b0cbb38491872aa7a2b

web/js/duplicates.js
chrismytton: I decided to use jQuery events so that other cobrands can hook into the same places in the future if necessary.

Done in 0b58e43af4d7997f44e2b37f1fb1f3ac255d21b9

chrismytton:
Thanks for the review @davea, really helpful.

I fixed the problem with the markers not always closing in 6f4dbc8fbc6c9ca3e9280a67c7cac9fb3a952073 and the date formatting in 67d529691609701193fd2e5451252342ea1b52fb.

Have pushed the changes as !fixup commits so it's hopefully a bit clearer what I've actually changed. I'll squash those down before merging.

Think this is ready for another review now.

davea (changes requested):
Looks great! I forgot to test on mobile before (sorry!) and I'm afraid there is one small issue I spotted - when tapping a defect pin the popup bubble is shown and then disappears shortly after moving into the centre of the map:

occ_mobile2


I think what's happening is the pan to fit in the bubble is triggering a refresh of the shown markers, and then as part of showing those the popup is removed. A few potential solutions I can think of:

- Increase the `ratio` of the pins layer to load a buffer of pins outside the viewport - would reduce the likelihood of the pins refreshing upon panning but not guaranteed. Also would potentially introduce a problem where not all pins within the current viewport would be shown (e.g. the 100 most recent reports would be spread over a wider area including off-screen)
 - Somehow recreate the popup when the layer finishes panning, if the associated defect marker is still shown (I think we did something similar with the `fault_feature` stuff in `asset_selected` in `fixmystreet/assets.js`.)
 - Prevent the pan from refreshing the markers when the popup is shown, and refresh when it's closed

chrismytton:
@davea I've fixed the disappearing popup problem in 38cccd6e6f08b9dc3c513807df4a44dab22f39df by simply removing the code that closes popups after panning the map. That code was introduced in 0ebd9c37bd8c637342c0236af1404cfd07c37a11 but from what I can tell the code I removed will always cause issues with popups on mobile, so is probably best removed.

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -284,4 +289,108 @@ sub dashboard_export_problems_add_columns {
...
+        outputformat => "application/json"
+    );
+
+    try {
+        my $response = get($uri);
+        my $json = JSON->new->utf8->allow_nonref;
+        return $json->decode($response);
+    } catch {
+        # Ignore WFS errors.
+        return [];
```
davea, line 331: `pins_from_wfs` is expecting a hashref, so this change would prevent a crash on L365 if the WFS response isn't valid JSON:

```suggestion
        return {};
```

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -284,4 +289,108 @@ sub dashboard_export_problems_add_columns {
...
+
+    return map {
+        [ $_->{latitude}, $_->{longitude}, $_->{colour},
+          $_->{id}, $_->{title}, "normal", JSON->false
+        ]
+    } @$pins;
+}
+
+sub extra_around_pins {
+    my ($self, $bbox) = @_;
```
davea, line 386: `$bbox` might be `undef` if the user isn't using JS (e.g. they've gone to a URL like `/around?pc=witney` without the `js=1`param), so we should return in that case to prevent an invalid request to the WFS server (e.g. https://tilma.mysociety.org/proxy/occ/nsg/?REQUEST=GetFeature&SERVICE=WFS&SRSNAME=urn%3Aogc%3Adef%3Acrs%3AEPSG%3A%3A4326&TYPENAME=WFS_DEFECTS_FOR_QUERYING&VERSION=1.1.0&filter=%3Cogc%3AFilter+xmlns%3Aogc%3D%22http%3A%2F%2Fwww.opengis.net%2Fogc%22%3E%3Cogc%3AAnd%3E%3Cogc%3APropertyIsEqualTo+matchCase%3D%22true%22%3E%3Cogc%3APropertyName%3EAPPROVAL_STATUS_NAME%3C%2Fogc%3APropertyName%3E%3Cogc%3ALiteral%3EWith+Contractor%3C%2Fogc%3ALiteral%3E%3C%2Fogc%3APropertyIsEqualTo%3E%3Cogc%3ABBOX%3E%3Cogc%3APropertyName%3ESHAPE_GEOMETRY%3C%2Fogc%3APropertyName%3E%3Cgml%3AEnvelope+xmlns%3Agml%3D%22http%3A%2F%2Fwww.opengis.net%2Fgml%22+srsName%3D%22%22%3E%3Cgml%3AlowerCorner%3EEPSG%3A4326+%3C%2Fgml%3AlowerCorner%3E%3Cgml%3AupperCorner%3E+%3C%2Fgml%3AupperCorner%3E%3C%2Fgml%3AEnvelope%3E%3C%2Fogc%3ABBOX%3E%3C%2Fogc%3AAnd%3E%3C%2Fogc%3AFilter%3E&propertyName=ITEM_CATEGORY_NAME%2CITEM_TYPE_NAME%2CREQUIRED_COMPLETION_DATE%2CSHAPE_GEOMETRY&outputformat=application%2Fjson)

davea (changes requested):
That fix is 💯, nicely done.

Spotted another couple of small bugs, but simple fixes hopefully.

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
chrismytton: Oops! Fixed in c0b845fd55d5772003030ae374f82de8c69c9899

perllib/FixMyStreet/Cobrand/Oxfordshire.pm
chrismytton: Ah good catch. Fixed in 88bd91ec9dad55b0553230ce8de062c18d209ab2

davea (approved):
💥 👍

PR #3226, opened 2020-10-23
Some hopeful Cypress Travis fixes

dracos:
Three hopeful fixes, for three different things that sometimes but not always fail on Travis. [skip changelog]

The first fix, while a nice thing to do anyway, did not help, for some reason, looking at the video, on Travis the Inspector part of the test is submitting the form ok, must be logging in okay, redirecting to /my/planned, but that then thinks not logged in and redirects back to /auth?r=my/planned (can't see how else you end up at that URL instead anyway). Some cookie issue with the hostnames, or something? I dunno. I've added another commit to use links to log out and back in instead of using cy.visit(), shouldn't make any difference but will see. As intermittent, it passing first time doesn't mean it is fixed :-/ Cypress has many open cookie issues, e.g. https://github.com/cypress-io/cypress/issues/7133 

I have now rerun the tests on push and PR 4 or 5 times and it has passed every time. Still can't be sure, but I think this certainly helps, so feel free to merge and rebase things on top if that would be useful.

### DIFF
--- a/.cypress/cypress.json
+++ b/.cypress/cypress.json
@@ -1,7 +1,7 @@
 {
   "baseUrl": "http://fixmystreet.localhost:3001",
   "projectId": "y8vvs1",
-  "blacklistHosts": ["gaze.mysociety.org", "*.openstreetmap.org", "portal.roadworks.org", "tilma.mysociety.org", "tilma.staging.mysociety.org"],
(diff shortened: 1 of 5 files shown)


davea (approved):
Looks good! 🤞

PR #3227, opened 2020-10-23
prevent editing of category names with hardcoded flag

struan:
If a category has hardcoded set to 1 in it's extra metadata then prevent
the name being edited in the admin. This is to avoid issues where the
name of the category is used in e.g. layers or other configuration and
changing it breaks things.

Also includes admin interface for setting this that is restricted to
super users only.

Fixes mysociety/fixmystreet-commercial#1992


davea (changes requested):
One picky thing, sorry!

davea (approved):
👍

PR #3230, opened 2020-10-30
[Central Beds] Add asset layers & fix mobile header

davea:
Adds street lighting and gullies asset layers. Also fixes an issue where the header prevented the "continue" button being shown on mobile when making a report.

[skip changelog]


struan (approved):
💡

PR #3232, opened 2020-11-02
[CheshireEast] allow cleaning reports to be made anywhere

struan:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2050

[skip changelog]

### DIFF
--- a/web/cobrands/cheshireeast/assets.js
+++ b/web/cobrands/cheshireeast/assets.js
@@ -46,6 +46,20 @@ fixmystreet.assets.add(labeled_defaults, {
     asset_item: 'street light'
 });
 
+var cleaning_cats = [
+  'Dead Animals',
+  'Dog Fouling',
+  'Fly Posting',
+  'Fly Tipping',
+  'Graffiti',
+  'Litter and Bins',
+  'Needles, Syringes or Glass',
+  'Play Area Safety Issue',
+  'Street Sweeping',
+  'Abandoned Vehicles - Not Taxed',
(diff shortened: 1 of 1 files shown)


web/cobrands/cheshireeast/assets.js
```diff
@@ -62,7 +76,15 @@ fixmystreet.assets.add(defaults, {
     all_categories: true,
     actions: {
         found: fixmystreet.message_controller.road_found,
-        not_found: fixmystreet.message_controller.road_not_found
+        not_found: function(layer) {
+            var cat = $('select#form_category').val();
+          console.log(cat, cleaning_cats);
```
dracos, line 81: debug

web/cobrands/cheshireeast/assets.js
```diff
@@ -62,7 +76,15 @@ fixmystreet.assets.add(defaults, {
...
+            if (OpenLayers.Util.indexOf(cleaning_cats, cat) != -1) {
+              fixmystreet.message_controller.road_found();
```
dracos, line 83: Should perhaps pass layer in (though I realise not used)?

dracos (changes requested):
Looks fine. I guess one question (presumably considered) is that if the highways layer now doesn't apply to all categories, why not list the categories it does apply to and stop it being `all_categories`, instead of providing an exemption list? Is that list much longer?

struan:
The list of categories it applies to is much longer so this seemed the sensible way round.

dracos (approved):

PR #3234, opened 2020-11-04
[Bromley] Cope if last instance has been rescheduled to the future

dracos:
The last instance is set by its original scheduled time, so if it is rescheduled to be in the future, it can still appear as the last instance. If this happens, treat it as a next instance instead (so then there will probably not be a last instance).

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -690,10 +690,19 @@ sub _parse_schedules {
 
         my $last = $schedule->{LastInstance};
         $d = construct_bin_date($last->{CurrentScheduledDate});
-        if ($d && (!$max_last || $d > $max_last->{date})) {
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -690,10 +690,19 @@ sub _parse_schedules {
 
         my $last = $schedule->{LastInstance};
         $d = construct_bin_date($last->{CurrentScheduledDate});
-        if ($d && (!$max_last || $d > $max_last->{date})) {
+        if ($d && $d->strftime("%F") gt $today && (!$min_next || $d < $min_next->{date})) {
```
struan, line 693: This could really do with a comment to explain what's going on.

struan (approved):
📆

dracos:
Has been merged with comment adding.

PR #3236, opened 2020-11-09
[Oxfordshire] script to bulk close a list of reports

struan:
Script that takes a CSV file with a list of report IDs and closes them
with a message, suppressing the alerts on the older ones.

Fixes mysociety/fixmystreet-commercial#2053

[skip changelog]


bin/fixmystreet.com/bulk-report-close
```diff
@@ -0,0 +1,106 @@
...
+        die "Found more reports than expected, bailing";
+    }
+    printf("Closing %d old reports with no alert ", $no_message->count);
+    FixMyStreet::Script::ArchiveOldEnquiries::close_problems($no_message);
+    printf("done.\n");
+} else {
+    printf("Would close %d old reports with no alert. Run with --commit to actually close reports.\n", $no_message->count);
+}
+
+# Provide some variables to the archiving script
```
dracos, line 87: This doesn't change any of the options, should it be called with retain_alerts = 1?

dracos (changes requested):
This lacks tests. Presumably the banes script can be removed now, if this supersedes it, and would it not be better to combine these with the archive-old-enquiries script? It looks like this basically does the same, but with configurable message, state, and retain_alerts - if those were all options to archive-old-enquiries, would that not then do the same (if run with no time period of email sending and the relevant options)? You'd also get configurable email cut off then.

bin/oxfordshire/archive-old-enquiries
```diff
@@ -19,11 +32,16 @@ my ($opts, $usage) = describe_options(
...
-    ['closure-cutoff=s',     "Anything before this will be closed with no email", { required => 1 } ],
-    ['email-cutoff=s',     "Anything before this will be closed with an email sent to the reporter", { required => 1 } ],
-    ['limit|l=s',     "limit to a certain number of reports/users to be closed"],
-    ['help|h',       "print usage message and exit" ],
+    ['cobrand=s',    "which cobrand template to use for sent emails" ],
+    [ "mode" => hidden => { one_of => [
+        ['closure-cutoff=s',  "Anything before this will be closed with no email" ],
+        ['reports=s',          "csv file with a single column of report ids" ],
+    ], required => 1 } ],
+    ['email-cutoff=s',     "Anything before this will be closed with an email sent to the reporter" ],
```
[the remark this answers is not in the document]
dracos: Also text help doesn't make sense for ID behaviour, I think.

perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -44,6 +46,63 @@ sub archive {
...
+        lastupdate => { '<', $opts->{email_cutoff} },
+    });
+
+    my $with_message = $rs->search({
+        lastupdate => { '>=', $opts->{email_cutoff} },
+    });
+
+    die "Found more reports than expected\n" if $no_message->count + $with_message->count > $max_reports;
+
+    close_problems($with_message);
```
dracos, line 75: I think retain_alerts is 0 by default, so it's not being set to 1 for this.

perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -44,6 +46,63 @@ sub archive {
...
+    my $reports = get_ids_from_csv();
+    my $max_reports = scalar @$reports;
+
+    my $rs  = FixMyStreet::DB->resultset("Problem")->search({
+        id => $reports,
+        state => [ FixMyStreet::DB::Result::Problem->open_states() ],
+    })->to_body( $opts->{body});
+
+    my $no_message = $rs->search({
+        lastupdate => { '<', $opts->{email_cutoff} },
```
dracos, line 66: Yeah, I'm a bit confused by this. So for the normal archive usage, it doesn't do anything for later than email_cutoff, sends an Archive email for earlier than email_cutoff, later than closure_cutoff, and doesn't send anything for earlier than closure_cutoff.

But with the ID list, you're saying send any normal alerts for later than email cutoff, and don't send any normal alerts for earlier than email cutoff. Which is sort of the reverse, iyswim. I think it would make more sense to be using closure_cutoff here as a cutoff between email/no email, and ignore email_cutoff. I think?

dracos (changes requested):
Couple of suggestions and some confusion, but looking good.

dracos (approved):
Looks good. Hope all my requested changes haven't ended up making it wrong in some way.

struan:
First issue is if in e.g.  Northants if I click off a road, select Kerbs -> Damaged/Loose it then tells me I need to pick a road maintained by the council but it's not clear what I should do. I realise I should click the back button but it feels like we could be a bit more helpful.

PR #3237, opened 2020-11-10
[Zürich] Send admin emails, if any, when report is confirmed

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/2056

[skip changelog]


dracos (approved):
That's a much nicer way than I'd thought of of doing it :)

PR #3241, opened 2020-11-12
[CheshireEast] send address string over open311

struan:
[skip changelog]


dracos (commented):
Is this needed? Does the code in SendReport::Open311 not automatically do this if there's a "closest_address" field present? (And indeed, would override what this function does?)

struan:
The version there has a lot of cruft, plus this should always be sent for Cheshire East so might as well do it here.

dracos:
In which case, like I say, it won't work because the version with a lot of cruft will always override whatever this function does, as far as I can tell.

PR #3244, opened 2020-11-14
Add GitHub Actions.

dracos:
[skip changelog] 

Diff failure is because of the changes added to deal with perlmagick being installed now and it trying to shrink the 1x83 test image which fails. OTOH is now testing e.g. the static_map stitch code which it wasn't on Travis.
Tried using the perl version docker images, but had weird test failures (slim base builds? Couldn't work it out), so instead am using GitHub's default, perlbrew to install a specific version and caching that.


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: Run tests (with coverage, part 1)
+      if: matrix.part == 1
+      run: |
+        script/test --jobs 3 $(find t/app/controller -name "*.t")
+      env:
+        PERL5LIB: local-coverage/lib/perl5
+        HARNESS_PERL_SWITCHES: "-MDevel::Cover=+ignore,local/lib/perl5,commonlib,perllib/Catalyst/[^A],perllib/Email,Test.pm,^t"
+
+    - name: Run tests (with coverage, part 2)
```
struan, line 49: Why the two parts?

struan (approved):
🤖

.github/workflows/coverage.yml
dracos: So it finishes quicker overall. It's a bit lopsided as you can see from the runs, but was at least an easy place to split them. Now it's in its own workflow, we could consider splitting further...
Originally done in f7873a97d3be13251da30f12705c3dafa5468995

PR #3246, opened 2020-11-17
[TfL] Use Street Manager for roadworks.

dracos:
[skip changelog] Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2045


davea (changes requested):
image

I think we'd better do something about LineStrings - these are almost impossible to see and harder to click. I think the centre point would do the job, as you suggested elsewhere.

davea (approved):
🎉

PR #3247, opened 2020-11-19
Cypress coverage

dracos:
This pins the Ubuntu version on GitHub Actions, and adds coverage to the Cypress tests, posting to codecov same as the Perl tests, which is nice (even though it drops overall coverage to begin with!). [skip changelog]


struan (approved):
📔

PR #3248, opened 2020-11-19
Fix non-JS form when all extra questions answered.

dracos:
The code for deciding whether there were extra questions to answer
would always run, even if all the extra questions had been answered.


chrismytton (approved):
👍

PR #3251, opened 2020-11-26
[Hackney] Use different variable for split sending

dracos:
Reusing sent_to can cause an issue if a non-
split email report is marked for resending.
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hackney.pm
+++ b/perllib/FixMyStreet/Cobrand/Hackney.pm
@@ -81,9 +81,9 @@ sub open311_extra_data {
     ];
 
     # Make sure contact 'email' set correctly for Open311
-    if (my $sent_to = $row->get_extra_metadata('sent_to')) {
-        $row->unset_extra_metadata('sent_to');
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -81,9 +81,9 @@ sub open311_extra_data {
     ];
 
     # Make sure contact 'email' set correctly for Open311
-    if (my $sent_to = $row->get_extra_metadata('sent_to')) {
-        $row->unset_extra_metadata('sent_to');
-        my $code = $sent_to->{$contact->email};
+    if (my $split_match = $row->get_extra_metadata('split_match')) {
+        $row->unset_extra_metadata('split_match');
```
davea, line 85: I wonder if `unset_extra_metadata` should work like `delete` and return the removed value. It'd make this and L177/178 a bit terser (whether that's a good thing or not I don't know!)

davea (approved):
`s/📨/🪓/g` :)

PR #3255, opened 2020-12-02
[Central Beds] Post go-live tweaks

davea:
A few things not picked up in testing:

 - Show external Symology ID on report page
 - Send email to user with Symology ID after report is sent
 - Hide reports made before the cobrand launched.

[skip changelog]


struan (approved):
👍

PR #3256, opened 2020-12-02
Multi page reporting base fallout

dracos:
The work for multi-page reporting has touched a wide array of parts of the codebase, and this is a random assortment of things that have been dealt with along the way. Sorry for their randomness. [skip changelog]

Remove sidebar.html - only used by Zurich really.
[UK] Some additional front-end tests - cover some parts not currently tested (Bromley assets, Bucks gritting, HE cobrand, more visibility checking, Westminster assets)
Add missing preventDefault()s - prevent a couple of '#' in the URL needlessly
Factor together photo upload template - tidying
[Cypress] Exempt some analytics domains - most are in a BASE_URL check, but some are not
Call begin_report from drag for consistency - so that when multi-page reporting is here, drag can also reset to first 'page' same as clicking etc.
[fixmystreet.com] Flag to skip 2FA on staging site - very annoying in dev when you want to log in as admin@
Remove duplicate event handler - it is being called twice on live, this should fix it - more complicated than I thought it would be for some reason I didn't track down
Only hide relevant responsibility message - otherwise if you had a road layer and a point asset layer in play, and you clicked somewhere not on an asset, and the point layer was added first, it would add its "click an asset" message, but the road layer would then remove it.


web/cobrands/buckinghamshire/assets.js
```diff
@@ -321,6 +321,7 @@ fixmystreet.assets.add(defaults, {
         }
     },
     no_asset_msg_id: '#js-not-a-road',
+    no_asset_msg_ids: '#js-not-a-road,#js-not-council-road-he,#js-not-council-road-other,#js-not-council-road',
```
davea, line 324: I'm a little confused why we have both of these. Is it because `no_asset_msg_id` is sometimes shown, but `no_asset_msg_ids` are only ever hidden? A comment explaining the difference, or a slightly more distinct name would be good for clarity when looking back later on.

davea (commented):
This is looking really good, nice to tidy things up and improve robustness. One question is all!

web/cobrands/buckinghamshire/assets.js
dracos: Yes, so if there is no `no_asset_msg_ids` it will just hide `no_asset_msg_id` by default. It's the group of IDs that could be shown (ie. for this Bucks asset road_not_found will show js-not-a-road but road_found will show one of the other IDs) so they need to be hidden at the same time because we don't know which one had been shown. I guess could rename to `no_asset_possible_msg_ids`, but comment here and in fixmystreet/assets.js probably clearer.

dracos: I've switched it to a class which I think is hopefully clearer.

davea (approved):
🎒 ✏️ ✅

PR #3257, opened 2020-12-04
[Peterborough] only display roadworks for some categories

struan:
[skip changelog]

### DIFF
--- a/.cypress/cypress/integration/peterborough.js
+++ b/.cypress/cypress/integration/peterborough.js
@@ -38,4 +38,15 @@ describe('new report form', function() {
     cy.get('.category_meta_message').should('contain', 'You can pick a tree from the map');
   });
 
+  it('displays nearby roadworks', function() {
+    cy.fixture('peterborough_roadworks.json');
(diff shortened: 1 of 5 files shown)


.cypress/cypress/integration/peterborough.js
```diff
@@ -38,4 +38,15 @@ describe('new report form', function() {
     cy.get('.category_meta_message').should('contain', 'You can pick a tree from the map');
   });
 
+  it('displays nearby roadworks', function() {
+    cy.fixture('peterborough_roadworks.json');
+    cy.route('/streetmanager.php**', 'fixture:peterborough_roadworks.json').as('roadworks');
+    cy.wait('@roadworks');
+    cy.get('select:eq(4)').select('Pothole');
+    cy.contains('Roadworks are scheduled near this location');
+    cy.contains('Parapet improvement');
```
davea, line 47: These should probably be suffixed with `.should('be.visible')` after Matthew's recent comment.

davea (approved):
One little suggested test improvement but overall ✅

PR #3258, opened 2020-12-04
[Buckinghamshire] Red claims

struan:
Initial Bucks red claims work.

This is the main workflow with all the fields and the framework for the Red Claims screens.

It doesn't have photo upload or location by map, or saving the report.

[skip changelog]


perllib/FixMyStreet/App/Controller/Claims.pm
```diff
@@ -0,0 +1,594 @@
...
+                },
+            ]
+        },
+        {
+            stage => 'when',
+            title => "When did the incident happen?",
+            fields => [
+                {
+                    desc => "Date",
+                    value => 'incident_date'
```
dracos, line 156: Error message on this for too early a date is `'2015' is not a valid value` which is not very helpful, as it doesn't say what is.

perllib/FixMyStreet/App/Form/Claims.pm
```diff
@@ -0,0 +1,635 @@
...
+
+has_field 'attention_date.year' => ( type => 'Year' );
+has_field 'attention_date.month' => ( type => 'Month' );
+has_field 'attention_date.day' => ( type => 'MonthDay' );
+
+has_field gp_contact => (
+    required => 0,
+    type => 'Text',
+    widget => 'Textarea',
+    label => 'Please give the name and contact details of the GP or hospital where you recieved medical attention',
```
dracos, line 564: received

templates/web/base/claims/summary.html
```diff
@@ -0,0 +1,412 @@
...
+    <dt class="govuk-summary-list__key">
+      [% summary_title %]
+    </dt>
+    <dd class="govuk-summary-list__value">
+    </dd>
+    <dd class="govuk-summary-list__actions">
+    <form method="post">
+        <input type="hidden" name="saved_data" value="[% form.fif.saved_data %]">
+        <input type="hidden" name="goto" value="[% step1 %]">
+        <input type="submit" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-0" value="Change answers">
```
dracos, line 371: This button has nothing associated with it

templates/web/base/claims/summary.html
```diff
@@ -0,0 +1,412 @@
...
+[% BLOCK format_value -%]
+    [% IF field.block == 'yes_no' %]
+        [% field.value == 1 ? 'Yes' : 'No' %]
+    [% ELSIF field.block == 'date' %]
+        [% field.value.day %]/[% field.value.month %]/[% field.value.year %]
+    [% END %]
+[%- END %]
+
+[%
+    summary = [
```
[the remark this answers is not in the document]
dracos: Currently summary includes "What are you claiming for? 1"

dracos (changes requested):
All seems to hang together well, I think we can consolidate and improve some things now hopefully :)

perllib/FixMyStreet/App/Form/Field/DOBYear.pm
```diff
@@ -0,0 +1,20 @@
+package FixMyStreet::App::Form::Field::DOBYear;
+use Moose;
+extends 'HTML::FormHandler::Field::IntRange';
```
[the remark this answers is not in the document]
struan: These are all text fields so I'm not sure what the issue is?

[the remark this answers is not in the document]
dracos: Sorry, got confused. Just the error message on it I mention below.

templates/web/base/govuk/fields.html
```diff
@@ -168,13 +168,21 @@ <h2 class="govuk-fieldset__heading">
       [% IF loop.last AND field.get_tag('last_differs') %]
```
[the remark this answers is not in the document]
struan: This only adds in things that aren't in the waste version, and it's identical to the noise one at the moment.

perllib/FixMyStreet/App/Controller/Claims.pm
```diff
@@ -0,0 +1,81 @@
...
+
+    my $report_id = $data->{report_id};
+
+    my $detail = "";
+
+    for my $stage ( @{ $form->fields_for_display } ) {
+        next if $stage->{hide};
+        for my $field ( @{ $stage->{fields} } ) {
+            next if $field->{hide};
+            $detail .= "$field->{desc}: " . $form->format_for_display($field->{name}, $field->{value}) . "\n";
```
dracos, line 38: As this is called both here and in template the same way, perhaps `fields_for_display` should return this as the value directly (or in another field to be used)?

templates/web/base/claims/summary.html
```diff
@@ -0,0 +1,57 @@
...
+            <input type="hidden" name="goto" value="[% page.stage %]">
+            <input type="submit" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-0" value="Change answers">
+        </form>
+        </dd>
+      </div>
+      [% FOR field IN page.fields %]
+      [% NEXT IF field.hide %]
+        <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key govuk-summary-list__key--sub">[% field.desc %]</dt>
+            <dd class="govuk-summary-list__value">[% form.format_for_display(field.name, field.value) | html %]</dd>
```
dracos, line 45: Don't need to escape any longer.

dracos (changes requested):
Looks really good, a few tiny comments is all, and thanks for very clear fixup commits.

perllib/FixMyStreet/App/Controller/Claims.pm
struan: I didn't want to so that in case it turns out that we need the underlying value.

perllib/FixMyStreet/App/Form/Claims.pm
```diff
@@ -0,0 +1,826 @@
...
+    label => 'What day did the incident happen?',
+    messages => {
+        datetime_invalid => 'Please enter a valid date',
+    },
+);
+
+has_field 'incident_date.year' => (
+    type => 'Year',
+    messages => {
+        select_invalid_value => 'The incident must be within the last five years',
```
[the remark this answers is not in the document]
dracos: Doesn't seem to be fixed, I'm afraid: 
image
Not sure why not, though, what you have looks like it should.

perllib/FixMyStreet/App/Controller/Claims.pm
dracos: Yeah, fair enough, just wondered if you returned format_for_display result as a `pretty` field so the template/here would be bit simpler then.

dracos (commented):
One fix doesn't appear to fix it, sorry if I've misunderstood!
And explaining what I meant hopefully more clearly on one other thing, but that's no matter really.

perllib/FixMyStreet/App/Form/Claims.pm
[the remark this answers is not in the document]
struan: ah, me not reading the docs carefully enough...

perllib/FixMyStreet/App/Controller/Claims.pm
dracos: You could use `$field->{pretty}` here as well :)

dracos (approved):
One other place you could use `pretty`, I think? :)
🧧

PR #3259, opened 2020-12-08
User edit admin link

dracos:
This moves the ban/flag buttons to the user edit admin, which seems a better place for them; and adds a link to the user admin from a report/update admin page. And tries to clarify that changing the name/user on a report/update changes it for that report/update, it isn't changing the user. And a fix to overflowing category dropdown when there are very long categories on report admin page.

Some of this has been done with inline styling that should be done properly, however that may be, help appreciated there :)

### DIFF
--- a/templates/web/base/admin/reports/edit.html
+++ b/templates/web/base/admin/reports/edit.html
@@ -131,6 +131,7 @@ class="admin-offsite-link">[% problem.latitude %], [% problem.longitude %]</a>
 <li><label class="inline-text" for="category">[% loc('Category:') %]</label>
     [% INCLUDE 'admin/report-category.html' %]
 </li>
+
 <li>[% loc('Extra data:') ~%]
     [%~ IF extra_fields.size ~%]
         <ul>
@@ -154,8 +155,6 @@ class="admin-offsite-link">[% problem.latitude %], [% problem.longitude %]</a>
     <input type='text' class="form-control" name='name' id='name' value='[% problem.name | html %]'></li>
 <li><label for="username">[% loc('User:') %]</label>
     <input type='text' class="form-control" id='username' name='username' value='[% problem.user.username | html %]'>
-    [% PROCESS abuse_button %]
-    [% PROCESS flag_button user=problem.user %]
 </li>
 [% IF problem.user.phone_display != problem.user.username %]
 <li>[% loc('Phone:') %] [% problem.user.phone_display | html %]</li>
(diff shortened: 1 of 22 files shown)


templates/web/base/admin/users/form.html
```diff
@@ -9,6 +9,29 @@
...
+
+          <label>
+            [% loc('Flagged:') %]
+            <input type="checkbox" id="flagged" name="flagged"[% user.flagged ? ' checked' : '' %]>
+          </label>
+          [% IF username_in_abuse %]
+           <small>[% loc('User in abuse table') %]</small>
+           <input class="btn" name="unban" type="submit" value="[% loc('Unban') %]">
+          [% ELSE %]
+           <input type="submit" class="btn" name="banuser" value="[% loc('Ban user') %]">
```
zarino, line 31: My first thought was, shouldn’t this "Ban" button be down next to all the other "Danger" buttons, at the end of the form?

But then I thought, really, isn’t this just a state (banned / not banned), and shouldn’t it just be toggled like all the other states (flagged / not flagged, verified / not verified)?

templates/web/base/admin/reports/edit.html
```diff
@@ -181,21 +167,49 @@
...
+<li><label class="inline-text" for="anonymous">[% loc('Anonymous:') %]</label>
+<select class="form-control" name="anonymous"  id="anonymous">
+<option [% 'selected ' IF problem.anonymous %]value="1">[% loc('Yes') %]</option>
+<option [% 'selected ' IF !problem.anonymous %]value="0">[% loc('No') %]</option>
+</select></li>
+<li><label class="inline-text" for="name">[% loc('Name:') %]</label>
+    [% loc('(for this report)') %]
+    <input type='text' class="form-control" name='name' id='name' value='[% problem.name | html %]'></li>
 
+<li><label class="inline-text" for="username">[% loc('User:') %]</label>
```
zarino, line 196: Should this be `[% loc('Username:') %]`?

templates/web/base/admin/reports/edit.html
```diff
@@ -181,21 +167,49 @@
...
     </li>
   [% END %]
 </ul>
 </li>
 [% END %]
+</ul>
 
-<li><label class="inline-text" for="closed_updates">[% loc('Closed to updates') %]:</label>
-    <input type="checkbox" id="closed_updates" name="closed_updates"[% ' checked' IF problem.extra.closed_updates %]></li>
+<div class="clearfix full-width" style="background-color: #eee; padding: 1em 1em 0;">
```
zarino, line 181: FWIW I’m totally fine with these inline CSS styles. Barely seems worth breaking them out into a separate class, given this is just an admin page, and the whole form is already quite rough-and-ready, it feels like we can get away with it here.

zarino (commented):
Generally thumbs up 👍

Just a few comments / questions.

templates/web/base/admin/users/form.html
dracos: I think it is because it's held separately internally, and don't want every user edit to have to check the table if it's not ticked, if you see what I mean.

templates/web/base/admin/reports/edit.html
dracos: We don't use the word username anywhere, so probably shouldn't start now.

PR #3261, opened 2020-12-08
Banes housing layer restriction

chrismytton:
Adds a new asset layer containing housing association information. When a pin is dropped onto one of these housing association areas and it's in the category whitelist a warning message will be shown telling the user to email the housing association instead.

Part of https://github.com/mysociety/fixmystreet-commercial/issues/2051

### DIFF
--- a/templates/web/fixmystreet.com/report/new/roads_message.html
+++ b/templates/web/fixmystreet.com/report/new/roads_message.html
@@ -19,4 +19,9 @@
     <div id="js-not-an-asset" class="hidden js-responsibility-message">
         <p>Please select <span class="js-roads-asset" data-original="an item">an item</span> from the map on which to make a report.</p>
     </div>
+    <div id="js-curo-group-restriction" class="js-responsibility-message hidden">
+        <strong>Maintained by Curo Group</strong>
+        <p>The location you have selected is maintained by Curo Group.</p>
+        <p>You can report this issue to Curo Group directly by emailing <span class="js-roads-asset"></span>.</p>
+    </div>
 </div>
--- a/web/cobrands/bathnes/assets.js
+++ b/web/cobrands/bathnes/assets.js
@@ -6,7 +6,8 @@ if (!fixmystreet.maps) {
 
 fixmystreet.maps.banes_defaults = {
     http_options: {
-        url: "https://isharemaps.bathnes.gov.uk/getows.ashx",
+        // Use tilma proxy for staging sites to avoid CORS issues.
+        url: fixmystreet.staging ? "https://tilma.mysociety.org/proxy/bathnes/isharemaps/" : "https://isharemaps.bathnes.gov.uk/getows.ashx",
         params: {
             mapsource: "BathNES/WFS",
             SERVICE: "WFS",
@@ -234,5 +235,59 @@ fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
     attribution: " © Crown Copyright. All rights reserved. 1000233344"
 });
 
+// List of categories which are Curo Group's responsibility
+var curo_categories = [
+    'Allotment issue',
+    'Obstructive vegetation',
+    'Play area safety issue',
+    'Trees and woodland',
+    'Dead animals',
+    'Dog fouling',
+    'Excessive or dangerous littering',
+    'Household bins left out',
+    'Needles'
+];
+
+fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Curo_Land_Registry"
+        }
+    },
+    asset_type: 'area',
+    stylemap: fixmystreet.assets.stylemap_invisible,
+    non_interactive: true,
+    always_visible: true,
+    all_categories: true, // Not really, but we deal with that in the found action handler.
+    nearest_radius: 0.1,
+    road: true,
+    no_asset_msg_id: '#js-curo-group-restriction',
+    actions: {
+        found: function(layer) {
+            var category = fixmystreet.reporting.selectedCategory().category;
+            if (curo_categories.indexOf(category) === -1) {
+                fixmystreet.message_controller.road_found(layer);
+                return;
+            }
+
+            fixmystreet.message_controller.road_not_found(layer);
+            $('#js-roads-responsibility > strong').hide();
+
+            var domain = '@curo-group.co.uk';
+            var email = 'estates';
+            if (category === 'Household bins left out early or after collection day') {
+                email = 'tennancycompliance&support';
+            }
+            var email_string = $(layer.fixmystreet.no_asset_msg_id).find('.js-roads-asset');
+            if (email_string) {
+                email_string.html('<a href="mailto:' + email + domain + '">' + email + domain + '</a>');
+            }
+        },
+        not_found: function(layer) {
+            $('#js-roads-responsibility > strong').show();
+            fixmystreet.message_controller.road_found(layer);
+        }
+    }
+});
 
 })();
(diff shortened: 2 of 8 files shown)


web/cobrands/bathnes/assets.js
```diff
@@ -234,5 +234,52 @@ fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
...
+        found: function(layer) {
+            var category = $('select#form_category').val();
+            if (curo_categories.indexOf(category) === -1) {
+                fixmystreet.message_controller.hide_responsibility_message();
+                return;
+            }
+
+            var email = 'estates@curo-group.co.uk';
+            if (category === 'Household bins left out early or after collection day') {
+                email = 'tennancycompliance&support@curo-group.co.uk';
```
dracos, line 273: Typo? Also would be good to have local parts separate, just to prevent the obvious spam.

templates/web/fixmystreet.com/report/new/roads_message.html
```diff
@@ -20,3 +20,9 @@
         <p>Please select <span class="js-roads-asset" data-original="an item">an item</span> from the map on which to make a report.</p>
     </div>
 </div>
+
+<div id="js-housing-association-restriction" class="js-responsibility-warning box-warning hidden">
```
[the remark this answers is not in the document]
dracos: (Also as I think it needs styling with a logo?)

web/cobrands/bathnes/assets.js
```diff
@@ -6,7 +6,7 @@ if (!fixmystreet.maps) {
 
 fixmystreet.maps.banes_defaults = {
     http_options: {
-        url: "https://isharemaps.bathnes.gov.uk/getows.ashx",
+        url: "https://tilma.mysociety.org/proxy/bathnes/isharemaps/",
```
dracos, line 9: Do we want the live service to use the proxy, presumably not?

web/cobrands/bathnes/assets.js
```diff
@@ -234,5 +234,52 @@ fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
...
+];
+
+fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Curo_Land_Registry"
+        }
+    },
+    asset_type: 'area',
+    stylemap: fixmystreet.assets.stylemap_invisible,
```
dracos, line 257: You might need a nearest_radius of e.g. 0.1 (like TfL red routes) to prevent nearby clicks from also matching to this

web/cobrands/bathnes/assets.js
```diff
@@ -234,5 +234,52 @@ fixmystreet.assets.add(fixmystreet.maps.banes_defaults, {
...
+            if (category === 'Household bins left out early or after collection day') {
+                email = 'tennancycompliance&support@curo-group.co.uk';
+            }
+            var email_string = $(layer.fixmystreet.no_asset_msg_id).find('.js-email');
+            if (email_string) {
+                email_string.html('<a href="mailto:' + email + '">' + email + '</a>');
+            }
+            fixmystreet.message_controller.show_responsibility_message(layer);
+        },
+        not_found: fixmystreet.message_controller.hide_responsibility_message,
```
dracos, line 281: I would prefer this to use the existing road_found/road_not_found, given my current work, which I think should be possible. `not_found` here would match to message_controller.road_found fine, and `found` would match to road_found if not a matching category, and could set asset_item before calling road_not_found (much as the streets asset layer in the same file already does). Then no need for the new message_controller functions.
It will need to behave differently in multi-page anyway, I'm hoping keeping the same mechanism will mean less to have to change and it might carry across fine. And then any tweaks can be made.

dracos (changes requested):
Given the multi-page work (and assuming this can't wait until that is done), I would really like it to use the existing functions (as they've already been updated), and it really needs some front end tests so it doesn't get broken later on, thanks.

web/cobrands/bathnes/assets.js
chrismytton: This was the address we were given. Will double check with them that it's the one they're expecting.

web/cobrands/bathnes/assets.js
chrismytton: Have changed this so staging sites use tilma and live sites use their one.

web/cobrands/bathnes/assets.js
chrismytton: There were a couple of reasons that I didn't use `road_found`/`road_not_found`.

Firstly, I want to be able to change the text of the `<strong>` tag in `#js-roads-responsibility` to have different text, rather than use the "Not maintained by Bath &amp; North East Somerset Council" that's there at the moment. But `#js-roads-res

The text for `asset_item` says `a <b class="asset-…`, which isn't appropriate for an email address.

Looking at the multi-page forms stuff I can see that making any changes to that stuff right now isn't really a good idea, so I've tried to work around the above issues in the found: handler. Bit fragile, but I think better than meddling with the core code.

dracos: On the strong, your sentence doesn't end; but I think in the BANES template you could move that strong inside the inner div and then it would work as you want, you wouldn't need the special show/hide then?

Ah yes, the "a..." makes it more annoying, doesn't it.Your solution to change the email looks good, thanks for adapting the code!

dracos (changes requested):
Changes look good! Is missing the nearest radius still, and still would like a test of it please.

chrismytton:
Oops, had done the nearest radius change but hadn't pushed it. That's in https://github.com/mysociety/fixmystreet/pull/3261/commits/9145e78878d5e8b68424515ff1493b5b1f205714.

Have started writing some tests but getting a bit tangled up trying to stub out the response from their WFS server in a way that lets me test that the stoppers are appearing at the right time.

I can look at this when I'm back on the 11th January, but if someone else wants/needs to pick it up before then please do.

chrismytton:
@dracos I've tried adding tests in 1921e4fcd1c183aaaaf97d6f973756612acfabb7 but for some reason the GeoJSON isn't getting added to the map. I uncommented the bit that hides the layer and even tried just passing the request through to the real server but nothing seems to work!

Any ideas what I've done wrong?

dracos:
You need to specify the co-ordinates you have used in t/Mock/MapIt.pm, something like:
`[ '?', 51.379009, -2.359276, 2551, 'Bath and North East Somerset Council', 'UTA' ],`
and you need to create a BANES body in the fixture, something like:
`{ area_id => 2551, categories => [ 'Dog fouling' ], name => 'Bath and North East Somerset Council' }`
The former is needed because otherwise the `/report/new` URL associates itself with the default mock MapIt area and fixture body (otherwise how was it showing any categories), and that will not match the asset layer's body when called in the `is_body_only` checks for both showing the Curo layer, and for showing the disabled message. And the latter is then needed because now you've matched those co-ordinates with area 2551, you need a body associated with area 2551, otherwise it will show the "We don't have any contacts for this area" message.

chrismytton:
@dracos Ahhh, thank you! I'd tried adding the fixture but didn't seem to work, this explains why, I was missing the mapit mock!

Have force-pushed a fixup commit with the tests in, will request another review once they're passing.

dracos (approved):

chrismytton:
@dracos I've added 4a971abbe4343d8711621a3d878196a797331a27 which addresses the bug with the button not being disabled for the drains category. Can you have another look and check it's a sensible approach please?

dracos (approved):
Yes, that looks like a sensible way to do it. Could you update the comment on the enable function to explain what it's doing? Given it's not now unconditional like disabling.
Really, I guess, the enable form function should have all its logic as to whether it can do so or not in it, so as well as the bit you've added, the stopper bit conceivably, but not suggesting you do that now!

PR #3262, opened 2020-12-09
Fix typo

abimysoc:
'some of this section may not <strike>not</strike> apply'


dracos:
Merged as 06dd2144eb3dd0feac90c684d4e3b73a54cd4c1c, thanks

PR #3263, opened 2020-12-09
fix alert/no alert queries for archiving a list of reports

struan:
Also fix the bad tests that said it worked

[skip changelog]


dracos (approved):
Oh, whoops, hope it didn't cause a problem

PR #3265, opened 2020-12-10
Upgrade Modernizr.

dracos:
File shrinks from 1407/815 to 1175/648 (is inlined in header).

The file is generated by compiling a slightly altered Modernizr with
nothing except `mq` and then compressed with closure-compiler.

Modernizr alterations are available in https://github.com/mysociety/Modernizr/commit/92580e1c8329b9b50b0a6eb11d3dcd5892e5eb0a:
* src/isSVG.js: isSVG forced to false
* src/ModernizrProto.js: Everything removed apart from version
* src/computedStyle.js: much simplified/console removed
* src/createElement.js: only uses first element
* lib/build.js: remove scriptGlobalObject, compact banner
* src/generate.js: Remove testRunner, _q loop

[skip changelog] cc @sagepe


davea (approved):
:shipit:

PR #3266, opened 2020-12-11
Combine map layer and hide pins toggles into new .map-links toolbar

zarino:
A new `.map-links` element, positioned near pan/zoom controls, replaces the old `.big-hide-pins-link`, and acts as a place to store controls to do with the display of the map.

map-links-smaller

As a result, this commit removes the map layer toggle and hide pins toggle from `#sub_map_links`. In most cobrands, this will mean `#sub_map_links` doesn’t appear at all for regular users.

Zurich already had its own road/aerial toggle, so that’s been incorporated into the new core functionality (the only difference being that on Zurich, aerial is the default layer, layer ID `0`, while roads is layer ID `1`).

Fixes mysociety/fixmystreet-commercial#2029.


templates/web/base/maps/openlayers.html
```diff
@@ -47,8 +69,12 @@
     [% END %]
     <img id="loading-indicator" class="hidden" aria-hidden="true" src="/i/loading.svg" alt="Loading...">
 
-[% IF NOT around_page AND map_sub_links %]
+<div class="map-links">
+    [% map_links | safe %]
+</div>
+
+[% IF map_sub_links %]
```
dracos, line 76: I don't think there are any any more, now Zurich is gone'

dracos: This was not true, there is the "add all to shortlist" button.

dracos (approved):
I wrote a few bits of this and fixed couple of bugs, but I approve it all!

PR #3268, opened 2020-12-14
Move emergency-banner.html include out of pre-steps.html

chrismytton:
I originally put it in pre-steps.html so that cobrands that already had
a banner wouldn't also show the emergency message. The idea being that
we'd migrate the existing emergency banners over.

However we've had a support ticket from Bucks saying that they've tried
to set an emergency message but of course it isn't showing because
they've already got a custom pre-steps.html with a banner in.

I think it's going to be easier to show both the existing banners and
the new emergency message banner. Then if it seems necessary we can
migrate the custom banners into the new emergency banner system
gradually over time, rather than having to do it all at once.

Related to https://github.com/mysociety/fixmystreet-commercial/issues/2071

### DIFF
--- a/templates/web/base/front/pre-steps.html
+++ b/templates/web/base/front/pre-steps.html
@@ -1,1 +1,0 @@
-[% TRY %][% PROCESS 'front/emergency-message.html' %][% CATCH file %][% END %]
--- a/templates/web/base/index.html
+++ b/templates/web/base/index.html
@@ -6,6 +6,8 @@
     <p class="form-error">[% error %]</p>
 [% END %]
 
+[% TRY %][% PROCESS 'front/emergency-message.html' %][% CATCH file %][% END %]
+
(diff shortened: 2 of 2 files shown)


dracos (commented):
I guess this is fine, but I don't understand "we can migrate the custom banners into the new emergency banner system gradually over time, rather than having to do it all at once" - why would you have to do it all at once? You could add one cobrand's message to the admin, and then add any CSS/remove its pre-steps.html, without affecting any of the others. I assume they would only come back and point out the styling is different anyway, at which point you have to do the same thing :) This way, you're going to at some point briefly have a duplicate message or no message during the transfer, I thought preventing that was the reason you'd done it as you had (which made sense).

chrismytton:
@dracos Yeah, you're totally right. I've opened https://github.com/mysociety/fixmystreet/pull/3269 to just migrate the Bucks emergency message over to the new system.

PR #3269, opened 2020-12-14
[Buckinghamshire] Migrate emergency banner to new system

chrismytton:
This migrates the Buckinghamshire emergency banner over to the new system that allows them to update the content of the banner.

I've already added the text from the banner in the admin, so this should work when it's put live.

Should fix https://github.com/mysociety/fixmystreet-commercial/issues/2071


dracos (approved):
🦧

PR #3270, opened 2020-12-14
Make navigation appear next to menu rather than scroll

dracos:
image

Lets you scroll it if the window is short:
image

Some aside commits to tidy some things up so I don't have to change as much; then one commit that implements it without moving it in the DOM, using JS to make it open/close, then a second commit that moves it in the DOM to right after the button, which I think makes it more keyboard navigable and more sense. Probably also needs a skip to content adding then as well.


web/cobrands/sass/_base.scss
```diff
@@ -712,15 +712,23 @@ ul.error {
...
+  font-weight: normal;
+  margin: 0;
+}
+
 #nav-link {
   width: 3em;
   height: 0;
   padding-top: 3em;
   overflow: hidden;
-  background: transparent url(../fixmystreet/images/#{$menu-image}.png) center center no-repeat;
```
zarino, line 720: If we’re not using the `menu-*.png` any more, should we `git rm` them from `/web/cobrands/fixmystreet/images/`?

web/cobrands/sass/_base.scss
```diff
@@ -712,15 +712,23 @@ ul.error {
...
-  background-image: inline-image("../fixmystreet/images/#{$menu-image}.svg"), none;
+  background: transparent inline-image("../fixmystreet/images/#{$menu-image}.svg") center center no-repeat;
   background-size: 22px 18px;
   #{$right}: 0;
+
+  html.js-nav-open & {
+    background-image: inline-image("../fixmystreet/images/cross-grey.svg");
```
zarino, line 730: Should really make a custom SVG for this, that matches the `menu-*.svg` images in `/web/cobrands/fixmystreet/images/`. I’ll sort that.

web/cobrands/sass/_base.scss
```diff
@@ -747,21 +755,44 @@ ul.error {
...
 
+html.mobile.js-nav-open #js-menu-open-modal {
+    position: absolute;
+    bottom: -1000px; // Just to be sure
+    left: 0;
+    right: 0;
+    background-color: rgba(0, 0, 0, 0.4);
+}
+
 // The nav list has this sort of "full width" look,
```
zarino, line 766: I think if we change the `left: 1em; right: 1em;` below to `left: 0; right: 0;` then we can remove the `margin: 0 -1em;` and this comment too.

zarino: Would mean we could _also_ drop a `margin: 0` from the `#main-nav` ruleset in `_layout.scss` too.

web/cobrands/sass/_base.scss
zarino: Also, aside from this line, it looks like `cross-grey.svg` isn’t currently used anywhere, so maybe we should just delete it.

zarino: Following up on my own comment – `cross-grey.svg` and `cross-grey.png` were added in bdaf17a566b235865ad004e427c6f2230df39f24 for toggling the visibility of the `#report-a-problem-sidebar`, and then all references to the two files were removed in `44c97f31635cda9cdd0ce2d37d186bd5af487e0c`, when we consolidated all of the inspector/admin-facing button icons into a single sprite. But for some reason I didn’t remove the `cross-*` image files. So I’m going to add a commit that removes them now. :+1:

web/cobrands/sass/_base.scss
zarino: I’ve added a commit that removes the `menu-*.png` files.

dracos:
Added skip to content link if you could check that over, then I think this might be ready to go.

zarino (approved):
Looks good to me! 👍

PR #3273, opened 2020-12-17
[Highways England] Disable contact form & point links to HE site

davea:
- Remove links from help pages to /contact
 - Switch on abuse_reports_only
 - Respect contact_email cobrand feature instead of global CONTACT_EMAIL

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
+++ b/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
@@ -4,6 +4,8 @@ use parent 'FixMyStreet::Cobrand::UK';
 use strict;
 use warnings;
 
+sub council_name { 'Highways England' }
+
 sub council_url { 'highwaysengland' }
 
 sub site_key { 'highwaysengland' }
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -88,6 +88,18 @@ sub allow_anonymous_reports { 'button' }
 
 sub admin_user_domain { 'highwaysengland.co.uk' }
 
+sub abuse_reports_only { 1 }
+
+sub contact_name {
```
dracos: If you look near the top of the file, you'll see a number of functions-that-should-use-UKCouncils-code being passed through, would be good to match those rather than repeat the code?

dracos (changes requested):
One suggestion for the contact name/email to match other usage in the file. And then I thought better to use the council FAQ/privacy than the base with its generic message so copied them over and deleted irrelevant bits, so fixed the cobrand.pm suggestion too. Do check that over!

davea:
Thanks for the review and the improvements. Your changes look good to me!

PR #3279, opened 2021-01-08
Open311 JSON fixes

dracos:
Stop the JSON output using arrays where not needed (was fixed for requests back in 2017, looks like, but this was missed), and also include JSON links on the index page.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2106


davea (approved):
Looks good!

PR #3284, opened 2021-01-13
[Hackney] Send email on noise update

dracos:
Also improve display of report detail/update text in emails (former, use html_para so it gets spacing okay, latter, fix issue where it was double calling html_para on update email lists). I imagine not all integrations might want to send an email but at this stage didn't seem worth cobranding it. [skip changelog]

### DIFF
--- a/templates/email/default/_email_comment_list.html
+++ b/templates/email/default/_email_comment_list.html
@@ -5,7 +5,7 @@
         <img style="[% list_item_photo_style %]" src="[% inline_image(update.get_first_image_fp) %]" alt="">
       </a>
     [%~ END %]
-      [% email_sanitize_html(update) | html_para | replace('<p>', '<p style="' _ list_item_p_style _ '">') %]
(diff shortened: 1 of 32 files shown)


templates/email/default/_email_comment_list.html
```diff
@@ -5,7 +5,7 @@
         <img style="[% list_item_photo_style %]" src="[% inline_image(update.get_first_image_fp) %]" alt="">
       </a>
     [%~ END %]
-      [% email_sanitize_html(update) | html_para | replace('<p>', '<p style="' _ list_item_p_style _ '">') %]
+      [% email_sanitize_html(update) | replace('<p>', '<p style="' _ list_item_p_style _ '">') %]
```
struan, line 8: Why are some of these using the new filter and some not?

struan (approved):
One small question but otherwise 👍

templates/email/default/_email_comment_list.html
dracos: turns out email_sanitize_html is already calling html_para, so I can't pass it to the new filter (which would call html_para again). So didn't seem worth doing anything about.

PR #3289, opened 2021-01-19
[Peterborough] vary sender for environment categories

struan:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2069

[skip changelog]


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -98,6 +98,124 @@ around 'open311_config' => sub {
...
+            state => 'closed'
+        });
+        FixMyStreet::DB->resultset('Comment')->create({
+            user_id => $self->body->comment_user_id,
+            problem => $problem,
+            state => 'confirmed',
+            cobrand => $problem->cobrand,
+            cobrand_data => '',
+            problem_state => 'closed',
+            # XXX need message here
```
davea, line 194: Does this comment still apply?

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -98,6 +98,124 @@ around 'open311_config' => sub {
...
+            text => '',
+        });
+    }
+}
+
+sub _fetch_features_url {
+    my ($self, $cfg) = @_;
+    my $uri = URI->new( $cfg->{url} );
+    if ( $cfg->{arcgis} ) {
```
davea, line 203: Should this be `$cfg->{type} eq "arcgis"`, based on L114 & L126?

davea (changes requested):
A couple of queries about the first commit, haven't yet looked at the second. Some tests would be good too.

web/cobrands/peterborough/assets.js
```diff
@@ -163,42 +189,62 @@ fixmystreet.assets.add(light_defaults, {
...
+      url: url_base + '3/query?',
       params: {
-        TYPENAME: 'DogBins'
+        outFields: '',
       }
     },
-    asset_category: 'Dog bin',
-    asset_item: 'dog waste bin'
+    stylemap: fixmystreet.assets.stylemap_invisible,
+    asset_category: ['General fly tipping', 'Hazardous fly tipping', 'Offensive graffiti', 'Non offensive graffiti'  ],
```
davea, line 236: Would be good to factor this together with the `PCC Property Combined` one above, if only to make it easier to update these lists of categories in the future.

davea (changes requested):
One little refactoring suggestion, and it looks like the message is missing from the FMS cobrand:

image

davea (approved):
LGTM!

templates/web/fixmystreet.com/report/new/flytipping_text.html
```diff
@@ -0,0 +1 @@
+As this is private land Peterborough Council are unable to clear the fly-tip, this is the land owners responsibility. Your report will still be sent to Peterborough Council’s environmental enforcement team who will investigate. Unfortunately, they can’t provide updates on these investigations so your report will be marked as closed once it has been received.
```
davea, line 1: Should these be `Peterborough City Council`? And missing an apostrophe in `owner's` I think. (Missing in the cobrand version too).

davea (changes requested):
Sorry, one style nitpick that I didn't spot before :)

davea (approved):

web/cobrands/peterborough/assets.js
```diff
@@ -163,42 +189,62 @@ fixmystreet.assets.add(light_defaults, {
...
+    asset_item: 'road',
+    asset_type: 'road',
 });
 
-fixmystreet.assets.add(bin_defaults, {
+// PCC Property Combined
+fixmystreet.assets.add(flytipping_defaults, {
     http_options: {
-      params: {
-        TYPENAME: 'DogBins'
```
dracos, line 197: Just to note to check whether this PR should be removing the dog/litter bin asset layers.

struan:
Extra commit to fix issues in https://mysocietysupport.freshdesk.com/a/tickets/874

templates/web/peterborough/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <div id="js-not-an-asset" class="hidden js-responsibility-message">
+        <p>Please select <span class="js-roads-asset" data-original="an item">an item</span> from the map on which to make a report.</p>
+    </div>
+</div>
+<div id="js-environment-message" class="hidden box-warning js-responsibility-message">
```
dracos, line 6: No need for the js-responsibility-message class then.

web/cobrands/peterborough/assets.js
```diff
@@ -225,9 +225,9 @@ fixmystreet.assets.add(flytipping_defaults, {
                     return;
                 }
             }
-            $("#js-roads-responsibility").removeClass("hidden");
-            $("#js-roads-responsibility .js-responsibility-message").addClass("hidden");
+            $("#js-roads-responsibility").addClass("hidden");
```
dracos, line 250: This line, what is it for now? If this is using its own message to display, why is this concerned with js-roads-responsibility at all? (Or if for some reason I've missed it is needed, why was it removed from the found function above.)

web/cobrands/peterborough/assets.js
```diff
@@ -225,9 +225,9 @@ fixmystreet.assets.add(flytipping_defaults, {
...
             $('#js-environment-message').removeClass('hidden');
+            fixmystreet.message_controller.check_for_stopper();
```
dracos, line 252: check_for_stopper is called on category change anyway, why do we need to additionally call it here?

web/cobrands/peterborough/assets.js
```diff
@@ -15,14 +32,45 @@ var defaults = {
...
+
+var waste_categories = ['General fly tipping', 'Hazardous fly tipping', 'Offensive graffiti', 'Non offensive graffiti' ];
+
+OpenLayers.Layer.PeterboroughVectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorNearest, {
+    getNearest: function(lonlat) {
+        var point = new OpenLayers.Geometry.Point(lonlat.lon, lonlat.lat);
+        var feature = this.getFeatureAtPoint(point);
+        var radius = this.fixmystreet.nearest_radius || 10;
+        if ( OpenLayers.Util.indexOf(waste_categories, fixmystreet.reporting.selectedCategory().category) != -1) {
+            radius = 2;
```
dracos, line 59: Is there a reason why this isn't just 2 for everything? Given the shapes are vectors, not lines, does it matter if it was 2 everywhere instead of 10? Wouldn't need this subclass at all then.

web/cobrands/peterborough/assets.js
```diff
@@ -163,42 +211,69 @@ fixmystreet.assets.add(light_defaults, {
...
+});
+
+// PCC Property Leased Out NOT Responsible
+fixmystreet.assets.add(flytipping_defaults, {
+    http_options: {
+      url: url_base + '3/query?',
+    },
+    actions: {
+        found: function() {
+            $("#js-roads-responsibility").addClass("hidden");
```
dracos, line 264: Ditto what is js-roads-responsibility doing here

web/cobrands/peterborough/assets.js
```diff
@@ -163,42 +211,69 @@ fixmystreet.assets.add(light_defaults, {
...
-    asset_item: 'dog waste bin'
+    actions: {
+        found: function(layer) {
+            $("#js-environment-message").addClass("hidden");
+        },
+        not_found: function() {
+            for ( var i = 0; i < fixmystreet.assets.layers.length; i++ ) {
+                var layer = fixmystreet.assets.layers[i];
+                if ( layer.fixmystreet.name == 'Adopted Highways' && layer.selected_feature ) {
+                    $("#js-roads-responsibility").addClass("hidden");
```
dracos, line 246: Should this be js-environment-message now?

web/cobrands/peterborough/assets.js
```diff
@@ -163,42 +211,69 @@ fixmystreet.assets.add(light_defaults, {
...
+fixmystreet.assets.add(flytipping_defaults, {
+    http_options: {
+      url: url_base + '3/query?',
+    },
+    actions: {
+        found: function() {
+            $("#js-roads-responsibility").addClass("hidden");
+            $('#js-environment-message').removeClass('hidden');
+        },
+        not_found: function() {
```
dracos, line 267: Is this function needed? Seems like the other layer would do what's wanted here, and hide the message on a road/show otherwise, this function is just repeating the same thing. I might have missed something. A test would make it easy to tell :)

dracos (changes requested):
A number of queries and things that need changing, I think. Tests would make it much more confident it was working as expected.

web/cobrands/peterborough/assets.js
struan: My understanding was these weren't being used.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -97,6 +97,153 @@ around 'open311_config' => sub {
...
+        if ( $features && scalar @$features ) {
+            return $self->SUPER::get_body_sender($body, $problem);
+        }
+
+        # neither of those so just send email for records
+        my $emails = $self->feature('open311_email');
+        if ( $emails->{flytipping} ) {
+            my $contact = $self->SUPER::get_body_sender($body, $problem)->{contact};
+            $problem->set_extra_metadata('flytipping_email' => $emails->{flytipping});
+            $problem->update;
```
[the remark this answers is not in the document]
struan: We do, but in the case I'm also using this to flag that we should auto close the report and add a comment to report after it's been sent which means it needs a bit more persistence.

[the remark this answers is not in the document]
dracos: If you stored it on $self->flytipping_email it'd still be accessible in both munge_sendreport_params and post_report_sent as far as I can see?

.cypress/cypress/integration/peterborough.js
```diff
@@ -18,7 +20,7 @@ describe('new report form', function() {
     cy.get('.js-post-category-messages:visible').should('contain', 'Please phone customer services to report this problem.');
     cy.get('.js-reporting-page--next:visible').should('be.disabled');
     cy.get('#form_emergency').select('no');
-    cy.get('.js-post-category-messages:visible').should('not.contain', 'Please phone customer services to report this problem.');
+    cy.get('.js-post-category-messages:hidden').should('not.contain', 'Please phone customer services to report this problem.');
```
dracos, line 23: I didn't follow the reason for this change

dracos (commented):
Looks good! Dealt with everything I raised. I have one suggestion on that flytipping email storing, sorry.
Also, ideally there'd be a test about the interaction of these layers with the Adopted Highways layer, I guess, but that's probably a nice to have if we're happy it appears to be working as expected.

.cypress/cypress/integration/peterborough.js
struan: I'm not sure either, I've reverted it.

perllib/FixMyStreet/Cobrand/Peterborough.pm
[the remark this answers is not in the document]
struan: The cobrand object in `munge_sendreport_params` isn't the same object in the other two so I don't think this works. At least it doesn't look like it to me. It's generated afresh in `SendReport::Email::send` so I don't think data persists across to it.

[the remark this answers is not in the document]
dracos: Oh, sorry. Last chance - how about you don't do the update (I can't see a discard changes anyway, at least!), then unset it after last use?

dracos (approved):
Phew, well done :)

PR #3291, opened 2021-01-20
Issues/commercial/2048 geolocation UI tidy up

mySocietyClive:
@zarino, @dracos 

Would you mind having a look at this, styled the current location ahref to be like a button, in base.scss and also for just Peterborough.

If you're happy with it shall I make the change for the cobrands that have some custom styling for their geolocation link?


dracos (commented):
Tech wise looks fine (will mean we can drop the Oxfordshire special case, I assume, always a plus), will let Zarino comment on design :)

web/cobrands/sass/_base.scss
```diff
@@ -2499,8 +2499,16 @@ label .muted {
...
 a#geolocate_link {
-  display: inline-block;
-  padding: 0.5em;
+  display: inline-block !important;
+  border-radius: 6px;
+  border: 1px solid #333333;
+  color: #222222;
+  text-decoration: none;
+  font-weight: bold;
+  background: transparent url("/i/locate-me-arrow-black.svg") 4% 50% no-repeat;
```
zarino, line 2508: Your `background` and `background-image` lines here are the wrong way round – the svg image will never get used, because it’s overwritten with the png.

Also, we have a mixin to output the two background-image lines. Here’s how you’d use it.

```background: transparent;
background-position: 0 0;
background-repeat: no-repeat;
background-size: 24px 24px;
@include svg-background-image('/i/locate-me-arrow-black');
```

A shame we have to include all the other `-position`, `-repeat`, etc bumph – maybe something we could improve in future!

Another thing I wonder is whether we could inline the SVG, into the HTML of the button itself. That way, if we give the SVG `{ fill: currentColor }` (or `fill="currentColor"`) then it’d automatically inherit the same colour as the button text, which would save us generating a new image for each cobrand with slightly different button text colours?

web/cobrands/sass/_base.scss
```diff
@@ -2499,8 +2499,16 @@ label .muted {
 }
 
 a#geolocate_link {
-  display: inline-block;
-  padding: 0.5em;
+  display: inline-block !important;
```
zarino, line 2502: Would be nice to have a comment on this line, explaining why we’re using an `!important` – what selector is it overriding? Makes it easier to refactor stuff later, if we ever want to.

web/cobrands/sass/_base.scss
```diff
@@ -2499,8 +2499,16 @@ label .muted {
...
+  border-radius: 6px;
+  border: 1px solid #333333;
```
zarino, line 2504: I wonder whether we should use some existing Sass variables for this border colour and the text color on the line below? Or take the opportunity to define some new variables just for this?

Also, if we’re styling this geolocation button as a regular button, should it use our `button-size()` and/or `button-variant()` mixins instead?

web/cobrands/sass/_base.scss
```diff
@@ -2499,8 +2499,16 @@ label .muted {
...
+  color: #222222;
+  text-decoration: none;
+  font-weight: bold;
+  background: transparent url("/i/locate-me-arrow-black.svg") 4% 50% no-repeat;
+  padding: 15px 15px 15px 40px;
```
zarino, line 2509: The core styles in fixmystreet are designed to work in both left-to-right (LTR) and right-to-left (RTL) languages, and we tend to make sure things like horizontal paddings and margins switch to suit the language direction.

We use a Sass function called `flip()` for this. Do a quick search of the Sass code for `flip(` to see examples of how it works. Give us a shout if it’s not obvious.

zarino (changes requested):
A good start. I’ve left a few comments on how to handle the SVGs, and how we support RTL languages – both things you’ve probably not encountered yet on an FMS cobrand ticket :-)

Since this changes the core templates and styles, we’ll need to check the new button looks ok in _every_ cobrand. I notice a few (eg: Cheshire East) look a bit broken at the moment, because they impose their own special padding on the button, for example.

mySocietyClive:
@zarino I've added flip(), and button-size() and rolled out style to all cobrands, and corrected custom paddings/styles.

Does this warrant an update to CHANGELOG?

templates/web/base/around/postcode_form.html
```diff
@@ -33,7 +33,10 @@
             IF c.get_param('filter_category'); link_params.filter_category = c.get_param('filter_category'); END;
             IF c.get_param('filter_group'); link_params.filter_group = c.get_param('filter_group'); END;
         %]
-        <a href="[% c.uri_for('/around', link_params) | html %]" id="geolocate_link"> [% loc('Use my current location') %]</a>
+        
+        <a href="[% c.uri_for('/around', link_params) | html %]" id="geolocate_link"><svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="M19.83 4.17c.17.17.218.429.12.65l-6.594 14.834a.583.583 0 01-1.069-.008L9.91 14.091l-5.557-2.38a.583.583 0 01-.006-1.068L19.18 4.05a.583.583 0 01.648.12h.001z" fill="#000" fill-rule="nonzero"/></svg>
```
zarino, line 37: Just had a thought! If we did the SVG as an `[% INCLUDE %]`, then it’d be much easier for cobrands to override with their own icon, without needing to fork the entire `postcode_form.html` template.

web/cobrands/sass/_mixins.scss
```diff
@@ -1,5 +1,17 @@
 $direction: 'left' !default;
 
+// Padding
+$pt: .75em;
```
zarino, line 4: These 8 variables have very generic names for something in the global scope!! I’d err towards longer, more obvious variable names, to aid with understanding/debugging later on, rather than super short names. Something like `$button-margin-top` etc.

If these are defaults, I’d also give them a `!default` flag, to make them easier for cobrands to override.

web/cobrands/sass/_mixins.scss
```diff
@@ -10,20 +22,18 @@ $direction: 'left' !default;
     $nav_background_colour: $nav_fg_hover !global;
 }
 
-@mixin button-size() {
-    display: inline-block;
-    margin: 0;
-    padding: 0.75em 1em;
+@mixin button-size($mt:$mt, $mr:$mr, $mb:$mb, $ml:$ml) {
```
zarino, line 25: If we’re allowing custom margin values here, should we also allow custom padding values?

web/cobrands/sass/_mixins.scss
```diff
@@ -10,20 +22,18 @@ $direction: 'left' !default;
...
+    display: inline-block !important; //Override inline display: none on #geolocation_link
+    margin: $mt $mr $mb $ml;
```
zarino, line 27: Given you’ve (correctly!) used `flip()` for the padding, I wonder whether it should also be used here, for the margin?

web/cobrands/tfl/base.scss
```diff
@@ -144,7 +144,8 @@ input.form-error, textarea.form-error {
 
 a#geolocate_link {
     font-family: $heading-font;
-    font-size: 1.125em;
+    font-size: 1em;
+    border-radius: 22.5px;
```
zarino, line 148: This is a suspiciously specific measurement! Are you doing it because you want a circular side on a 45px high button? If so, I’d use some really big value like `border-radius: 10em` – the browser will scale it down to fit at most half the height (or width) of the button.

zarino (commented):
Looking good. A couple of comments while I was passing by.

zarino:
> Does this warrant an update to CHANGELOG?

Yeah, why not! Maybe in a new `- Development improvements:` section, like the previous set of `#geolocation_link` improvements in `v3.1`?

mySocietyClive:
@zarino updated, please check again.

Thanks

web/cobrands/sass/_mixins.scss
```diff
@@ -131,12 +143,12 @@ $right: right;
   @include experimental(box-shadow, $shadows, -moz, -webkit, not -o, not -ms, official);
 }
 
-@mixin experimental($property, $value, $moz: true, $webkit: true, $o: true, $ms: true, $official: true) {
-  @if $webkit   { -webkit-#{$property} : #{$value}; }
-  @if $moz      {    -moz-#{$property} : #{$value}; }
-  @if $ms       {     -ms-#{$property} : #{$value}; }
-  @if $o        {      -o-#{$property} : #{$value}; }
-  @if $official {         #{$property} : #{$value}; }
+@mixin experimental($button-padding-rightoperty, $value, $moz: true, $webkit: true, $o: true, $ms: true, $official: true) {
```
zarino, line 146: Oops! Looks like a find-and-replace error. `$property` shouldn‘t have been changed on these 6 lines.

web/cobrands/sass/_mixins.scss
```diff
@@ -176,8 +188,8 @@ $right: right;
   }
 
   background: -webkit-gradient(linear, 0% 0%, 0% $webkit-end-point, $webkit-gradient-stops);
-  @each $prefix in -moz-, -o-, -webkit-, "" {
-    background: #{$prefix}linear-gradient($stops);
+  @each $button-padding-rightefix in -moz-, -o-, -webkit-, "" {
```
zarino, line 191: Another find-and-replace error here. I think maybe you just did a search for `$pr` and didn’t notice that it was also picking up the `$pr` at the start of `$property` and `$prefix`? Anyway, `$prefix` shouldn’t have been changed in these two lines.

templates/web/base/around/geolocate_link.html
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="M19.83 4.17c.17.17.218.429.12.65l-6.594 14.834a.583.583 0 01-1.069-.008L9.91 14.091l-5.557-2.38a.583.583 0 01-.006-1.068L19.18 4.05a.583.583 0 01.648.12h.001z" fill="#000" fill-rule="nonzero"/></svg>
```
zarino, line 1: Seems a bit funky that there’s so much whitespace around the "arrow" in this image. The arrow itself is about 17px square, but the canvas is 24px. If we leave the SVG at its default 24px height in the button, it ends up causing odd vertical spacing above and below the button label text.

I’m thinking maybe either the arrow should be bigger, so it extends to the edges of the canvas (or thereabouts) or the canvas should be smaller.

web/cobrands/sass/_base.scss
```diff
@@ -2499,61 +2537,71 @@ label .muted {
...
 
 a#geolocate_link {
-  display: inline-block;
-  padding: 0.5em;
+  border: 1px solid $geolocation-link-border;
+  color: $geolocation-link;
+  @include button-size($button-margin-top: 15px);
+  svg {
+    display: inline-block;
+    vertical-align: text-bottom;
```
zarino, line 2545: FWIW these days, when it comes to vertically centering two horizontal siblings in an element (like an icon and some text in a button), I tend to use `display: flex; align-items: center;` on the parent, rather than faffing around with `vertical-align` like this. No matter what `vertical-align` you pick, it always ends up looking wrong at some font-size. At least `align-items: center` just does what it’s meant to.

We have some mixins that provide browser fallbacks for the flex properties – see the end of `_mixins.scss`.

Up to you whether you want to use it here. I think it’s probably worth it.

zarino (changes requested):
A few more comments, inline. Sorry!

You also need to `git rm` the icon files we no longer need: `web/i/locate-me-arrow*`.

I notice the diff on `sass/_base.scss` looks quite a lot bigger than it really is, because you’ve (intentionally, or unintentionally) run some sort of linter over the file, which has fixed up spacing, quote styles, and hex code capitalisation on a bunch of lines unrelated to the rest of the commit ("Roll out geolocation link button style change to all cobrands ", 5dace580dc6a75b795a11fd6a99af661262387ff). There’s nothing wrong with tidying this sort of stuff up, but ideally the tidying would go in a separate commit, to make browsing / debugging the history easier in future. That said, it feels like it’d be more hassle than it’s worth, now, to split commit 5dace580dc6a75b795a11fd6a99af661262387ff into two (one commit for the actual geolocation-related changes, one for the `_base.scss` tidying). What do you think @dracos ?

dracos:
Sorry for the hassle, but yes, it should be two commits, in case someone needs to see the change in future (or in worse case revert, in which case having them in one would make that much harder).

zarino:
> Sorry for the hassle, but yes, it should be two commits, in case someone needs to see the change in future (or in worse case revert, in which case having them in one would make that much harder).

@mySocietyClive Let me know if you want to unpick this commit together, sometime this week.

CHANGELOG.md
```diff
@@ -19,7 +19,9 @@
         - Improve layout of some admin pages.
     - Security:
         - Increase minimum password length to eight.
-
+* v3.1.1 (4th February 2021)
```
dracos, line 22: We don't give a version number/date until release, new rows should be added to the "Unreleased" section above; this would go under Front end improvements, I think, not Development (which is for internal non-user facing stuff).

mySocietyClive:
@zarino I've made amends, let me know if they are ok.

@dracos I've updated changelog.

web/cobrands/sass/_mixins.scss
```diff
@@ -10,20 +22,20 @@ $direction: 'left' !default;
     $nav_background_colour: $nav_fg_hover !global;
 }
 
-@mixin button-size() {
-    display: inline-block;
-    margin: 0;
-    padding: 0.75em 1em;
+
+@mixin button-size($button-margin-top:$button-margin-top, $button-margin-right:$button-margin-right, $button-margin-bottom:$button-margin-bottom, $button-margin-left:$button-margin-left, $button-padding-top:$button-padding-top, $button-padding-right:$button-padding-right, $button-padding-bottom:$button-padding-bottom, $button-padding-left:$button-padding-left) {
+    @include flex-container-inline();
```
dracos, line 27: I think @zarino's comment about flex inline was for applying to the geolocate link svg bit - https://github.com/mysociety/fixmystreet/pull/3291/files#diff-3ae05a9e3e785d418edb2ae5a8abad31d99b8844cef4fbcee18e4a932779633bR2474 - not all button-sizes, unless you wish to go through and check this hasn't affected anything else (I think this is used by every button on the site :) ).

dracos:
I've rebased this on current master, you'll want to reset to it before adding anything more.

web/cobrands/sass/_mixins.scss
zarino: Yeah, @mySocietyClive you’ve added `@include flex-align($alignment: center);` to the `button-size()` mixin, but I don’t think we want it there, just on the geolocation button, after you `@include flex-container-inline();`.

So, yeah, remove the `flex-align` line from the `button-size` mixin in `_mixins.scss`, and put it into the `a#geolocate_link` ruleset in `_base.scss` :+1:

zarino:
Aside from the `flex-align()` / `button-size()` issue, only other thing on my mind is whether we should tidy up the loading state for this button? eg, here’s how it currently looks in the fixmystreet.com cobrand:

Screenshot 2021-03-02 at 15 29 34

My instinct is that maybe 2021 is the year to finally replace this loading gif with an CSS spinner. The spinner could get inserted through a CSS `:before` pseudoelement on the button, and if we hid the geolocation arrow icon at the same time, it’d look like the one just replaced the other, during the loading state.

mySocietyClive:
I've added the animated svg now. 

Screenshot 2021-03-08 at 09 21 34

CHANGELOG.md
```diff
@@ -7,8 +7,12 @@
         - Send text alerts for report updates to only-phone-verified users.
         - Add options for user to set global notification preferences.
         - Pop over mobile navigation menu. #3270
+<<<<<<< HEAD
         - Add support for the OS Maps API. #3328
         - Speed up /report/new/ajax call. #3335
+=======
+        - Improve `#geolocate_link` display, especially for smaller screens. #2048
+>>>>>>> Improve display of geolocation button.
```
dracos, line 15: You've committed git conflict markers. Not sure what tool you use, but you should be able to have something that alerts you to this.

web/cobrands/sass/_mixins.scss
```diff
@@ -24,7 +24,6 @@ $button-margin-left: 0 !default;
 
 
 @mixin button-size($button-margin-top:$button-margin-top, $button-margin-right:$button-margin-right, $button-margin-bottom:$button-margin-bottom, $button-margin-left:$button-margin-left, $button-padding-top:$button-padding-top, $button-padding-right:$button-padding-right, $button-padding-bottom:$button-padding-bottom, $button-padding-left:$button-padding-left) {
-    display: inline-block !important; //Override inline display: none on #geolocation_link
```
dracos: The "important" was added earlier in this PR, and presumably not needed, but now the whole row is being removed? Can I just check, does this need to be inline-block or not? It's been inline-block since at least 2016.

dracos (changes requested):
One git conflict committed, one question about a line removal, but mostly - the spinner hasn't changed?
image
The code to show the old spinner is still present, and the CSS for the new spinner doesn't work because the loading class is put on the link parent not the SVG itself.

web/cobrands/sass/_mixins.scss
mySocietyClive: I think I removed it in error, I'll add it back in now.

Thanks

CHANGELOG.md
mySocietyClive: Ok thanks I've removed git markers. I'm using VS Code - any plugins you can recommend for alerts?

dracos: As I understand it, VS Code should be highlighting conflicts for you. When I do a merge with conflicts, I get something like this in the VS Code sidebar:
image

Opening one of those files gives
image

which I can resolve manually or using the helper buttons. If I try and commit using VS Code before resolving the conflicts, it gives me an error: "you have unmerged files". So it seems to do what you'd want already. So I guess the question is what are you doing when presented with a file with conflicts to resolve the conflict but leave the conflict markers in the file - are you somehow "git adding" it from the command line before resolving the conflict in the file?

mySocietyClive:
Please review again.

Thanks

zarino (approved):

mySocietyClive:
Thanks @zarino 

I've merged in your changes and swashed commits down to 2, and just rebased latest from master on top of my topic branch.

Please review again.

zarino (approved):
Looks good to me! 👍

PR #3292, opened 2021-01-20
Fix class adding going Back on desktop reporting.

dracos:
If you e.g. look somewhere in Bristol on desktop, pick Bridges, click Continue, then click Back, the "Pick a bridge" message has now moved to the bottom of the page. We only want the message to look like that on mobile, it shouldn't have changed on desktop.

[skip changelog]


davea (approved):
🗣️ 🔛 🥫 ✅

PR #3293, opened 2021-01-21
[UK] Various cookie banner changes.

dracos:
Fixes most of https://github.com/mysociety/fixmystreet-commercial/issues/1819 [skip changelog]


zarino (approved):
Looks good to me.

Minor nitpick – in the Oxfordshire popup, you’re missing some space between the alert text and row of buttons, because the `margin-bottom` defined by the `.occlss-u-bspace-sm` utility class in `_oxfordshire-cookiepopup.scss` is getting overridden by the `margin: 0.625rem 0 0 0` on `.occlss-alert__text` later in that file. On the Oxfordshire site, the utility classes are defined _later_ in the file, so they override the more generic classes. I tend to take a more bootstrappy approach, and add `!important` to all my little utility classes – the logic being, if I’ve added a class to an element to add some margin or whatever, then I _really_ want that margin, so an `!important` flag isn’t out of the question.

So, up to you, I’d either add `!important` flags to the `.occlss-u-*` classes in `_oxfordshire-cookiepopup.scss`, or just move those `.occlss-u-*` rulesets to the end of the file.

Issue #3287, opened 2021-01-18
Can't click floating navigation links on /faq on desktop browsers


davea:
Looks like a layer ordering problem - double clicking selects the text in the main body of the page. Tested on Safari/Chrome/Firefox with the same result on each. Works fine on mobile.

dracos:
It's due to my moving the navigation to the top in the HTML, but I don't yet see how. If I remove the div.nav-wrapper and the a name=content then it becomes clickable again, but neither of those two things cover the aside box.

PR #3297, opened 2021-01-25
Fix clicking of /about sidebar on desktop

davea:
Removes the z-index values from the sticky-sidebar and aside.
This doesn’t appear to affect anything else...

For #3287

[skip changelog]


dracos (approved):
AIUI the z-index dated from when the content had a box-shadow, so that the sidebar would go underneath the shadow. The shadow is long gone, looks like, so presumably these can be as well. As to why this fixes it or what caused it, though, still confused - if I remove the moved menu and the <a name="content"> it suddenly becomes clickable again, but not sure why that would affect what is lower down, might want @zarino  to check/confirm

PR #3298, opened 2021-01-28
Moving category dropdown to radios (part 1; that doesn't do that)

dracos:
So, we want to change the category drop down to be a list of radio icons. You'd think that wouldn't be too hard to do, but think of how it works currently - HTML is one `select` with `optgroup`s (if any category groups), that JS then turns into one top-level `select` and multiple second level `select`s. So I thought at first I could turn this into one big list of radio icons and nested fieldsets with legends to then do the same... but that didn't seem ideal (though after the below, perhaps it would have been), and the more visual display than a drop-down made me worry there'd be a flash of this list showing before it changed (not really as noticeable if entries in a drop-down change, I mean, unless you already have it open).

So I thought I would change it so that it outputs the two levels of `select` in the HTML, reducing what the JS has to do, and then that would be easier to turn into radio inputs after that. Anyway, getting that far is this part 1, so I'm hoping I won't be wrong, guess we'll see - it is probably better anyway, as it is less JS, though not as much less as I'd hoped.

I also made templates/code always use category groups whether they exist or not, which simplifies things a bit. And I had to have special code for the native app that won't understand the new way of doing it at present (but could pass in the same flag if it did).

So the upshot of this PR is it shouldn't change anything, really! Hopefully the centralising of the Cypress test category picker and JS category getter will be useful in part 2 when they are radio inputs. Most of the changes are that or knock-on effects to tests etc, the actual main changes are in Report.pm/New.pm/fixmystreet.js and templates. [skip changelog]

[edit: Not sure I like the 5 lines in IoW assets.js code coverage lost, but can't see why they would have been. I think as less JS to create the category dropdown, it's being too quick to have the asset layer load (positive look), or something has broken and it's not loading it when it should be (negative look) - manually it appears to be working okay]


perllib/FixMyStreet/DB/ResultSet/Contact.pm
```diff
@@ -7,6 +7,17 @@ use POSIX qw(strcoll);
 
 __PACKAGE__->load_components('Helper::ResultSet::Me');
 
+
+sub fake {
```
davea, line 11: A comment explaining why this code is needed would be helpful.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -730,6 +730,24 @@ sub stash_category_groups : Private {
         }
     }
 
+    # New reporting, top level mixes groups and categories not in a group
+    if ($opts->{mix_in}) {
+        my $no_group = delete $category_groups{""};
+        my @list = map { [ $_->category_display, $_ ]  } @$no_group;
+        foreach (keys %category_groups) {
+            (my $id = $_) =~ s/[^a-zA-Z]+//g;
+            push @list, [ $_, { id => $id, name => $_, categories => $category_groups{$_} } ];
```
davea, line 739: The same `id` key needs to be included on L759 & L761–L763, I believe, otherwise `group.id` in `subcategories.html` doesn't render correctly (eg  at `/around?pc=pier+walk%2C+greenwich`):

image

davea (changes requested):
What a beast, looking good overall though. The only broken thing I've noticed is non-JS reporting.

Clicking "Submit" on the **Category** page results in errors on the console: 
```
An invalid form control with name='category.RoadsPavements' is not focusable.
An invalid form control with name='category.StreetCleansing' is not focusable.
```
 There appears to be one error for each group in the contacts for that location. These are the contacts I have set up:

image


This occurs in Chrome and Safari on desktop.
It s

perllib/FixMyStreet/DB/ResultSet/Contact.pm
dracos: At the end of what I have done, I am not sure it is worth it, it only saved what, a line or two of code (and then in Part 2 I removed one of the two places it's used!). I guess it is slightly more consistent but not entirely so, e.g. admin/templates still calls stash_category_groups with a hash rather than an object.

perllib/FixMyStreet/App/Controller/Report.pm
dracos: Hmm, I guess on an around page it shouldn't be trying to show these at all, because they'll only be overridden as soon as you click on a map! (It's using the around map filter list, not where a report would be, too, which is confusing). When it was one select, it was within a IF js loop, so it probably should be too, which should fix this. 
So I don't think you need the ID in those lines, because it shouldn't get there with report/new list, if that makes sense.

perllib/FixMyStreet/DB/ResultSet/Contact.pm
dracos: Have ditched it

perllib/FixMyStreet/App/Controller/Report.pm
dracos: Have added `IF NOT js` around the include of `report/new/subcategories.html` in `form_report.html` which should fix this

dracos:
"Clicking "Submit" on the Category page results in errors on the console:" - ah, this is why I shouldn't use required on the subcategory selects. Fixed by changing it to class=required.
(Alternative would be not to output the subcategory dropdowns when showing category, and only outputting the relevant subcategory rather than all of them, but that seemed a bit more complex.)

davea (approved):
Looks good!

PR #3299, opened 2021-01-29
#2137 Update Westminster cobrand to match new branding

mySocietyClive:
Hi @zarino 

Update to Westminster cobrand:

- Updated font
- colour palette 
- Button styles
- Footer styles

Thanks
Clive

[skip changelog]


zarino (changes requested):
This looks lovely!

I think we could probably get it a bit closer to the official site by setting a custom container width, instead of the FMS default `60em`. Looks like `$container-max-width: 1084px;` in `westminster/_colours.scss` would do it.

It’s tempting, also, to just increase the vertical padding in the header a smidge, to more closely match the official site. `#site-header { padding-top: 1.5em; padding-bottom: 1.5em; }` in `westminster/layout.scss` would do it, alongside `$mappage-header-height: 5em;` in `westminster/_colours.scss`. Once you’ve done that, you’ll see everything lines up … although you may then notice that our version of the Westminster logo is a few pixels smaller than theirs – I’ll leave fixing that as an excercise for you, depends whether it annoys you enough to change it ;-)

A small point – we usually try to end text files with a newline. It’s a convention that helps simplify `git diff`s / `git blame`s in some situations. You’ll notice that’s why Github shows you a little red error icon at the end of its diffs for `header_extra.html`, `westminster/_buttons.scss`, `westminster/base.scss`, etc. It doesn‘t matter so much in the SVG images, but it’s something to watch out for when you’re editing HTML templates, Sass styles, and source code.

mySocietyClive:
@zarino please could you review latest changes.

Thanks

zarino (approved):
A thing of beauty!

I’d say just squash the two commits into one before you merge.

Might want to ask @wrightmartin or @dracos about how to skip changelog and review conflicting templates on Github, if you’ve not been introduced to that before.

PR #3300, opened 2021-02-01
 Moving category dropdown to radios (part 2; that does do that)

dracos:
Follows on from #3298
image


templates/web/base/report/new/category.html
```diff
@@ -28,14 +28,14 @@
     >
         <legend id="form_category_legend">[%~ form_category_label OR loc('Category') ~%]</legend>
         [%~ FOREACH group_or_cat IN category_groups ~%]
-            <div>
+            <div class="govuk-radios__item">
           [% IF group_or_cat.name %]
-                <input required id="category_[% group_or_cat.id %]" type="radio" name="category" value="[% group_or_cat.name %]" data-subcategory="[% group_or_cat.id %]"[% ' checked' IF filter_group == group_or_cat.name %]>
-            <label class="inline" for="category_[% group_or_cat.id %]">[% group_or_cat.name %]</label>
+                <input class="govuk-radios__input" required id="category_[% group_or_cat.id %]" type="radio" name="category" value="[% group_or_cat.name %]" data-subcategory="[% group_or_cat.id %]"[% ' checked' IF filter_group == group_or_cat.name %]>
```
zarino, line 33: If the user attempts to Continue without selecting a radio option, the client-side validation message gets injected before the first radio input, and it looks a bit weird:

Screenshot_2021-02-04 Reporting a problem

Do you think we could move it so that it displays _before_ the first `.govuk-radios__item`? (or, same thing but different way of thinking about it, immediately _after_ the `fieldset legend`?)

web/cobrands/sass/_base.scss
```diff
@@ -2867,3 +2867,18 @@ $emergency-message-border: 1px solid #525252 !default;
...
 @import "_top-banner";
 @import "_search-help";
+
+.govuk-radios__label{font-weight:400;line-height:1.25;margin-top:-2px;margin-bottom:0;display:inline-block;padding:13px 15px 13px 1px;cursor:pointer;touch-action:manipulation;float:left}
+.govuk-radios__label:before{content:"";box-sizing:border-box;position:absolute;top:8px;left:0;width:24px;height:24px;border:2px solid currentColor;border-radius:50%;background:transparent}
+.govuk-radios__label:after{content:"";position:absolute;top:15px;left:7px;width:0;height:0;border:5px solid currentColor;border-radius:50%;opacity:0;background:currentColor}
+.govuk-radios__item{font-weight:400;line-height:1.25;display:block;position:relative;min-height:0;margin-bottom:0;padding-left:34px;clear:left;float:left}
+.govuk-radios__item:after{content:"";display:block;clear:both}
+.govuk-radios__input{cursor:pointer;position:absolute;z-index:1;top:-2px;left:-10px;width:44px;height:44px;margin:0;opacity:0}
+.govuk-radios__input:focus + .govuk-radios__label:before{border-width:4px;box-shadow:0 0 0 4px #fd0}
```
zarino, line 2877: Might be premature optimisation (given this minified CSS code is obviously meant to look temporary) but do you think we should make this `#ffdd00` (bright yellow) `:focus` colour into a Sass variable, so it can be overridden more easily?

It’s the one bit of styling here that screams "GOV.UK" so it might be nice to pick a different shade for fixmystreet.com (perhaps something closer to the FMS yellow `#ffd000`) and I imagine there are at least a few more modern council cobrands that already customise their button/input focus colours (but I haven’t checked – could be wrong!).

Just a thought.

zarino (approved):
Mostly looks good to me. Just two small points in the comments, but not deal-breakers if you want to get this out quickly.

Tested and works fine in IE9–11, Edge, Firefox, and iOS Safari.

templates/web/base/report/new/category.html
dracos: Yeah, the validation code already has a special case for segmented-control which I think I can change to instead check for type==radio which would cover all the segmented-control cases plus this one.

web/cobrands/sass/_base.scss
dracos: "given this minified CSS code is obviously meant to look temporary" - well, hmm, it wasn't ;)
I'll add a variable and set it to default to $primary?
Question - should I rename the govuk class if we want this to be default everywhere? We use radios in admin, contact page, those are the main two, both would benefit from something similar to this.

dracos: I have moved the CSS to external file (will have to put the sass that generates it somewhere), and used a variable, and left class names alone for now.

dracos:
This was merged into master.

PR #3301, opened 2021-02-01
[Hackney] Move to new asset layer server.

dracos:
The server we were using is gone. [skip changelog]
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2143


chrismytton (approved):
👍

PR #3302, opened 2021-02-01
[Hackney] Hide noise category from normal flow.

dracos:
[skip changelog]
I guess we should add a new state or flag for contacts that hides them from normal flow which could then work for waste etc as well, but just for now to stop any more staff-made reports.


struan (approved):
🎺

Issue #2900, opened 2020-02-26
Allow selection of all a group's categories


dracos:
The filter on a map page (around or all reports) lists all the categories, now grouped by group if groups are in use, but there's no way to select all the categories in a particular group at once. Perhaps using jquery-multi-select group options (they'd all appear at the top, which might be nice?)

joesiltberg:
I have a related, but easier issue I'd like to solve (commenting here because I'm not sure if I should create a separate issue or try to tackle this one).

We would like to more easily select all categories except one (or a few). So it would be nice if by clicking "All", all checkboxes were selected, and then the user can deselect a few (like how the status selection works).

I have a naïve fix for this by just setting the data-all-options attribute in the select for categories, similar to how it's done for the status filter.

However seeing this issue I'm thinking maybe you wouldn't want that patch? I could try to implement a more complete fix with presets both for all and groups. I guess from your description above it's still not decided how it _should_ work though..?

By the way, I'm not sure how to easily test category groups. I didn't see any way to configure category groups in the admin interface, is it only for groups configured through Open 311?

Thanks!

joesiltberg:
Nevermind the last question about testing category groups, I found an explanation on the mailing list, I'll try that!

joesiltberg:
Created a PR for just my (simpler) problem for now, in case you want it.

PR #3303, opened 2021-02-02
Makes it possible to select all categories in a map page by clicking "All"

joesiltberg:
This makes the category selection behave more like the status selection. By clicking "All", all checkboxes are checked.

This makes it easier for a user to select "all except..."

I've tested this with and without category groups, but I've only tested on v3.0 (the changes in this file after 3.0 don't seem to conflict with this change though).

See also #2900


dracos:
Thanks for this; don't have as much time as normal at present (having to homeschool!), but hopefully will be able to take a look at this at some point.

joesiltberg:
No problem @dracos ! Take care and stay safe 💪

dracos:
I've just written https://github.com/mysociety/jquery-multi-select/pull/22 which if that's acceptable and works would mean we could pass in the all flag instead of having to construct a variable consisting of all the options solely to be read back out again.

joesiltberg:
Sounds good!

If I understand this correctly (sorry it was a while since I looked at this), this would mean we could skip my changes in this PR, and instead import a new version of web/vendor/jquery.multi-select.min.js to the project, and also update web/cobrands/fixmystreet/fixmystreet.js (in the make_multi function) to make use of the new all flag..?

Would you like me to try something like that in this PR? Or just close this?

dracos:
Yep, that sounds about right! That PR has been merged, so yes, if you'd like to try it out and see if it works, please do.

joesiltberg:
@dracos sorry for the delay. Now this PR is updated according to what we said above (using the new version of jquery-multi-select)

dracos:
No need to apologise! This looks good, I've just been testing it and there's one issue I hadn't thought of (sorry), which is that if e.g. Everything is picked in the category dropdown, all the categories get added to the URL, rather than the nicer none-of-them (because in this situation none selected is same as all selected). I think this should be easy enough to fix at the FMS end by the thing that updates the URL on such a change spotting if everything is picked, or something like that, I can try and take a look soon.
(I only worry because I imagine passing all the categories to the backend is harder on the DB than having none passed, iyswim).

joesiltberg:
Ahh yeah that would be nice if it could be solved. It might not help too much for the original use case for this issue however (selecting "all except..."), if the next thing the user does is deselect a category.

For that use case the preset alone should still be an improvement, since previously the users would manually click all the other categories, generating lots of requests.

dracos:
I've just pushed an extra commit to this that should remove the categories from the URL/ajax if they are all selected, do let me know if that works for you.

davea (approved):
This looks great!

joesiltberg:
Looks good to me too! @dracos did you want me to test the URL thing before you merge this?

dracos:
Sorry for the delay, this is now merged, thanks!

PR #3305, opened 2021-02-03
[Open311] Add support for waste_only service keyword

davea:
Stores a flag in the contact’s extra indicating whether it should only appear on the waste workflow, and not the main report flow.

This will be used by the Peterborough cobrand as part of the waste product.

[skip changelog]


chrismytton (approved):
🗑️ 👍

PR #3309, opened 2021-02-05
Shrink GOV.UK stylesheet to only used components.

dracos:
This switches to using a generated CSS file created by the script PRed in https://github.com/mysociety/fixmystreet.com/pull/19 - should contain everything the waste/noise forms currently use, but is quite a bit smaller. [skip changelog]


wrightmartin (commented):
All makes perfect sense, nice work

PR #3313, opened 2021-02-08
[Open311] Trim leading/trailing whitespace from service groups

davea:
Fixes an issue preventing reports being made in categories whose group had trailing whitespace.

[skip changelog]


dracos (approved):
👾

PR #3315, opened 2021-02-12
Send email when staff member subscribes user to alerts

chrismytton:
This adds a new email that is sent when someone is subscribed to email alerts by a member of staff. Councils sometimes use this when someone phones to report a problem and they enter it into the system.

The email that is sent contains an unsubscribe link so that users can stop receiving updates if they wish.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2049

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Alert.pm
+++ b/perllib/FixMyStreet/App/Controller/Alert.pm
@@ -196,13 +196,18 @@ sub create_alert : Private {
         $alert->insert();
     }
 
+    $c->stash->{alert} = $alert;
+
     if ( $c->user_exists && ($c->user->id == $alert->user->id || $c->stash->{can_create_for_another})) {
         $alert->confirm();
+        if ($c->stash->{can_create_for_another} && $c->user->id != $alert->user->id) {
+            # User has been subscribed to the new_updates alert email by a staff member. Send them an email to let them know.
+            $c->forward('send_subscribed_email');
+        }
     } else {
         $alert->confirmed(0);
     }
 
-    $c->stash->{alert} = $alert;
 }
 
 =head2 set_update_alert_options
@@ -303,6 +308,30 @@ sub send_confirmation_email : Private {
 
     $c->stash->{email_type} = 'alert';
     $c->stash->{template} = 'email_sent.html';
+}
+
+=head2 send_subscribed_email
+
+Send an email to an address that has been subscribed to alerts by a staff user.
+
+=cut
+
+sub send_subscribed_email : Private {
+    my ( $self, $c ) = @_;
+
+    my $user = $c->stash->{alert}->user;
+
+    my $token = $c->model("DB::Token")->create( {
+        scope => 'alert',
+        data  => {
+            id => $c->stash->{alert}->id,
+            type => 'unsubscribe',
+        }
+    } );
+    $c->stash->{unsubscribe_url} = $c->cobrand->base_url( $c->stash->{alert_options}->{cobrand_data} ) . '/A/' . $token->token;
+    $c->stash->{problem_url} = $c->cobrand->base_url_for_report($c->stash->{problem}) . $c->stash->{problem}->url;
+
+    $c->send_email( 'alert-subscribed.txt', { to => $user->email } );
 }
 
 =head2 prettify_pc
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Alert.pm
```diff
@@ -196,13 +196,18 @@ sub create_alert : Private {
         $alert->insert();
     }
 
+    $c->stash->{alert} = $alert;
+
     if ( $c->user_exists && ($c->user->id == $alert->user->id || $c->stash->{can_create_for_another})) {
         $alert->confirm();
+        if ($c->stash->{can_create_for_another}) {
```
dracos, line 203: This will send the email to a staff user with this permission who is creating an alert for themselves as well. I guess it should check the alert user is not the same as the logged in user.

templates/email/default/alert-subscribed.html
```diff
@@ -0,0 +1,21 @@
...
+
+INCLUDE '_email_top.html';
+
+%]
+
+<th style="[% td_style %][% only_column_style %]">
+  <h1 style="[% h1_style %]">Subscribed to [% site_name %] alerts</h1>
+  <p style="[% p_style %]">You have been subscribed to [% site_name %] alerts for <a href="[% problem_url %]">[% problem.title %]</a>.</p>
+
+  [% IF unsubscribe_url %]
```
dracos, line 16: This always exists?

templates/email/default/alert-subscribed.txt
```diff
@@ -0,0 +1,22 @@
...
+[% problem_url %]
+
+Email alerts are a free service from [% site_name %].
+
+This email was sent automatically, from an unmonitored email account - so
+please do not reply to it.
+
+[% signature %]
+
+[% IF unsubscribe_url %]
```
dracos, line 14: Ditto

templates/email/default/alert-subscribed.txt
```diff
@@ -0,0 +1,22 @@
...
+Unsubscribe?
+
+We currently email you whenever someone leaves an update on the
```
dracos, line 17: Probably doesn't need this paragraph, that was covered by the main body.

templates/email/default/alert-subscribed.html
```diff
@@ -0,0 +1,21 @@
...
+  <p style="[% p_style %]"><a href="[% unsubscribe_url %]">Unsubscribe from alerts about new [% site_name %] reports</a></p>
```
dracos, line 17: It's not an alert for new reports, so this needs rewording

perllib/FixMyStreet/App/Controller/Alert.pm
```diff
@@ -305,6 +310,30 @@ sub send_confirmation_email : Private {
...
+
+    my $token = $c->model("DB::Token")->create( {
+        scope => 'alert',
+        data  => {
+            id => $c->stash->{alert}->id,
+            type => 'unsubscribe',
+        }
+    } );
+    $c->stash->{unsubscribe_url} = $c->cobrand->base_url( $c->stash->{alert_options}->{cobrand_data} ) . '/A/' . $token->token;
+    $c->stash->{problem_url} = $c->cobrand->base_url() . '/report/' . $c->stash->{problem}->id;
```
dracos, line 332: I think this should be something like `$c->cobrand->base_url_for_report($c->stash->{problem}) . $c->stash->{problem}->url` - 2nd part just as it's a helper function, 1st part just in case there's some way it might be different (I don't think there is but I might well have forgotten something).

dracos (changes requested):
Some small changes needed. It'd also be good to update Develop.pm to say this new email template is `with_problem` so that it passes a problem through for the dev preview page.

templates/email/default/alert-subscribed.html
chrismytton: Ooops yeah, bit of cargo culting going on there!

perllib/FixMyStreet/App/Controller/Alert.pm
chrismytton: Ooh didn't know about `base_url_for_report`, thanks!

dracos (approved):
Changes all look good!

PR #3316, opened 2021-02-15
add reports failed count to sendreports output

struan:
Also have a list of ids and order output so it's consistent to make
it a bit easier to scan.

### DIFF
--- a/t/sendreport/failures.t
+++ b/t/sendreport/failures.t
@@ -1,0 +1,47 @@
+use FixMyStreet;
+use FixMyStreet::DB;
+use FixMyStreet::Script::Reports;
+use FixMyStreet::TestMech;
+
+use Test::Output;
+use DateTime;
+
+ok( my $mech = FixMyStreet::TestMech->new, 'Created mech object' );
+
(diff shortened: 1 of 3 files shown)


t/sendreport/failures.t
```diff
@@ -0,0 +1,49 @@
...
+use FixMyStreet::DB;
+use FixMyStreet::Script::Reports;
+use FixMyStreet::TestMech;
+
+use Test::Output;
+use DateTime;
+
+ok( my $mech = FixMyStreet::TestMech->new, 'Created mech object' );
+
+my $e = FixMyStreet::SendReport::Email->new();
```
dracos, line 11: Missing import of this

dracos (changes requested):
Looks fine, assuming the test is fixed and passes :)
I've wondered about having a summary at the top by body, e.g. of the $count problems, how many from each body so you can see easily which is having an issue, dunno if that's easy to include in this.

dracos (approved):
`@ids` still defined but no longer used

Issue #1860, opened 2017-09-28
Add configuration option to show staff user name as well as body name


jonkri:
I'm wondering about the purpose of `$can_view_contribute` (`$c->user_exists && $c->user->has_permission_to('view_body_contribute_details', $self->problem->bodies_str_ids)`) in Comment.pm. What is the rationale behind only displaying the name of the user if this condition is fulfilled?

Thanks!

dracos:
If an update is left by a staff user, then that update is displayed by default with the name of the body, rather than the name of the user. If the logged-in user has the `view_body_contribute_details` permission, then that staff user can also see the name of the staff user.

jonkri:
Ah, "staff user" - I've been wondering what to call that in English. :)

We are considering to show the names of staff users. I guess we could just put a FixaMinGata if statement in there, in that case?

Thanks! :)

dracos:
It'd be best if there was a cobrand function that could be called providing a configuration option for whether this happens or not.

jonkri:
Ah, I see. We have actually also been discussing, very briefly, about perhaps making sure staff updates are never anonymous. I'll investigate where we stand and will report back on this. Thanks!

jonkri:
Hi, again!

We have decided *not* to make a change related to how staff updates work, in terms of anonymity.

However, it would be great if there was a configuration option for whether or not to display the staff user name in addition to the body name. :)

PR #3319, opened 2021-02-16
Let the cobrand function always_view_body_contribute_details take contributed_as into account

joesiltberg:
Some background:

For FixaMinGata we've shown the names of staff users in their updates. This has been a wish from some of our municipalities, although some have disagreed (leading some staff users to change their names to just first name or something anonymous).

We've started using the functionality now where a staff user can create update "as body" rather than "as myself", I was hoping this would solve the problem by giving the user the choice of who's shown as the source of the comment.

I had assumed when creating an update "as body", only the body name was shown, but because of our override in our cobrand's `always_view_body_contribute_details` , both body name and staff user name was shown regardless of how the update was made.

However removing our override instead leads to only the body name showing (even when creating an update "as myself").

Now I'm starting to feel like I don't really understand the purpose of creating updates "as body" versus "as myself", so perhaps my proposed solution is a bad idea.

Anyway, by sending in contributed_as to the cobrand function, we can get the behaviour I expected; when creating an update as "myself" both body name and staff user name is shown, and when creating an update "as body" only the body name is shown.

I'm just a bit concerned that I'm misunderstanding mySociety's intention with this feature...what's the intended difference between "as body" and "as myself" if it doesn't show up in the comment's meta line? Perhaps I'm setting myself up for trouble down the line if I'm working against the intended way of using FixMyStreet :-)

Sorry for the long PR description, the changes are at least quite small :-)

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixaMinGata.pm
+++ b/perllib/FixMyStreet/Cobrand/FixaMinGata.pm
@@ -182,7 +182,8 @@ sub state_groups_inspect {
 }
 
 sub always_view_body_contribute_details {
-    return 1;
+    my ( $self, $contributed_as ) = @_;
+    return $contributed_as eq '';
 }
 
 # Average responsiveness will only be calculated if a body
--- a/perllib/FixMyStreet/DB/Result/Comment.pm
+++ b/perllib/FixMyStreet/DB/Result/Comment.pm
@@ -298,7 +298,7 @@ sub meta_line {
                 $body = 'Island Roads';
             }
         }
-        my $cobrand_always_view_body_user = $cobrand->call_hook("always_view_body_contribute_details");
+        my $cobrand_always_view_body_user = $cobrand->call_hook(always_view_body_contribute_details => $contributed_as);
         my $can_view_contribute = $cobrand_always_view_body_user ||
             ($user && $user->has_permission_to('view_body_contribute_details', $self->problem->bodies_str_ids));
         if ($self->text) {
--- a/t/cobrand/fixamingata.t
+++ b/t/cobrand/fixamingata.t
@@ -107,7 +107,7 @@ subtest "Test ajax decimal points" => sub {
     };
 };
 
-subtest "check user details always shown" => sub {
+subtest "check user details shown when staff user comments as themselves" => sub {
     FixMyStreet::override_config {
         ALLOWED_COBRANDS => [ 'fixamingata' ],
     }, sub {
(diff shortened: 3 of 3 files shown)


dracos:
Sorry for delay in looking at this. Yes, updates left by staff members, it currently assumes they will only be using the site in their official capacity, and so will use the body name as the public name of the update. Your override to change that was from https://github.com/mysociety/fixmystreet/issues/1860 and always showed the user's name as well.

The difference between create as body/myself is, as far as I can tell, "as body" stores the body's name on report/update rather than the user's name, won't ever be anonymous, and won't create an update alert for a new report. However, for updates it looks like it always uses the user's name (rather than the update's name) for display, whereas on reports it does use the report's name (and so the difference between body/myself there is more obvious!). So bit inconsistent, sorry, not entirely sure why either, and does make it a bit useless for updates. Bit hard to change too, as we couldn't start showing personal user names of users that it was previously displaying body name by default (plus people might not even have those permissions and just be relying on its default behaviour on updates...).

So basically, I think your solution might be the best one I can think of. Could do with a test so we don't break it somehow in future, if you're happy to do that or I can look at doing it at some point hopefully soon.

joesiltberg:
Thanks for the feedback @dracos ! I think we'll go ahead with this solution for FixaMinGata then, and I think the feature will turn out to be useful for updates as well then.

I'll see if I can figure out how to add a test case for this, haven't tried that before. That would be the backend tests I run with `script/test`? The test would end up somewhere under `t/app/controller`? Are there any docs/hints for writing those tests? Or any examples that might be similar to what I would test?

dracos:
Yes, that's right. As it's testing FixaMinGata specific code, the test would go in t/cobrand/fixamingata.t, I think. I guess you'd want the test to set up a problem+update, then check the right thing was displayed for the different options. The existing test has already created a report in `$report`, there's a `$mech->create_comment_for_problem` function you can use to create a comment, then you would use e.g. `$mech->get_ok('/report/' . $report->id);` to visit the report page, and `$mech->content_contains` or `$mech->content_like` to check it contains something (or lacks/unlike to check it doesn't contain something). Update the comment to have a different `contributed_as`, refetch the page, check it now shows the right thing, and so on.

joesiltberg:
@dracos I added a test now, I'd appreciate it if you could take a look at it as I'm not familiar with this testing framework. Seems to work though :-)

dracos (approved):
Test looks great :)

PR #3323, opened 2021-02-19
Sort output in export script.

dracos:
Makes it easier to compare. [skip changelog]

### DIFF
--- a/bin/export-import-data
+++ b/bin/export-import-data
@@ -43,6 +43,8 @@ sub export {
     if (@categories) {
         my @contacts = $body->contacts->search({
             category => { -in => \@categories },
+        }, {
+            order_by => 'category',
         })->all;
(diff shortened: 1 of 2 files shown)


bin/export-import-data
```diff
@@ -43,7 +43,7 @@ sub export {
     if (@categories) {
         my @contacts = $body->contacts->search({
             category => { -in => \@categories },
-        })->all;
+        })->search(undef, { order_by => 'category' })->all;
```
davea, line 46: Is there a reason we need two `search`es here? Would this work instead:
```suggestion
            order_by => 'category',
        })->all;
```

dracos: order_by has to be in the second parameter to search(), it's not a query parameter, but yes, not sure why I hadn't spotted the search above!

davea: Oops, of course.

davea (approved):

PR #3328, opened 2021-02-23
Add support for OS Maps API

dracos:
OS Maps key, layer, URL, and whether it's premium or not can be provided on a per-cobrand basis. In order to do that I had to switch the OSM map classes to use Moo (so e.g. number of zoom levels could look at configuration to work itself out), which hopefully tidies it up a bit at the same time.

The other base change was to store the cobrand being used to set a map class, not just the map class. This is so that, if you are in the scenario where someone is generating a map on a OS-Maps-API-using cobrand, for a report that was generated on a non-OS-Maps-API-using cobrand, it would still be able to access the right information for the API key etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
         - Send text alerts for report updates to only-phone-verified users.
         - Add options for user to set global notification preferences.
         - Pop over mobile navigation menu. #3270
+        - Add support for the OS Maps API. #3328
     - Bugfixes:
         - Fix non-JS form when all extra questions answered. #3248
         - Improve display of disabled fields in iOS.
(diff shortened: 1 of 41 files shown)


perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -41,6 +41,11 @@ sub path_to_email_templates {
     return $paths;
 }
 
+sub map_type {
```
davea, line 44: Should this be in `UK.pm` so it's available in the FMS cobrand?

davea (approved):
One question, but looks great overall! 👍

Issue #3288, opened 2021-01-19
Examples on Open311 page could be customised


dracos:
The examples on `/open311` are mostly hard-coded to Fiksgatami examples (as they wrote the code for that bit). It would be good if it could use per-cobrand examples, either automatically or just by specifying them as with the jurisdiction ID.

PR #3333, opened 2021-03-01
[Zurich] Open311 docs changes

chrismytton:
Makes the open311 documentation at https://www.zueriwieneu.ch/open311 more Zurich-specific.

Fixes https://github.com/mysociety/fixmystreet/issues/3288
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2150


dracos (changes requested):
Looks good. One other order by, I think, and it would be good to have a test that visits this page on default/zurich and check they differ correctly.

dracos (approved):
🧭

PR #3334, opened 2021-03-01
[Bromley] Add spinner when form submitted

dracos:
We have had a few double submissions; the first goes through fine, the second isn't accepted because there already is one (until the first one gets resolved, at which point it would be accepted if not caught beforehand). Backend service is quite slow at the point you're submitting, so I think most obvious cause is someone thinking it hasn't worked and submitting the form again. So disable the button/show a spinner when it is submitted. [skip changelog]
(Some other bits of tidying/speeding up development)


web/js/waste.js
```diff
@@ -0,0 +1,6 @@
+$(function() {
+    $('form.waste').on('submit', function(e) {
+        $('input[type=submit]', this).prop("disabled", true);
+        $('#form-submit-row').addClass('loading');
```
struan, line 4: There's quite a few other button classes used here that might as well have this give we are disabling the button, e.g form-go-row on the address lookup page and also the continue buttons on the enquiry form.

struan (approved):
Looks good although might be nice to be consistent about showing the spinner on all the buttons where there's a wait.

web/js/waste.js
zarino: Yeah, maybe `$('input[type="submit"], this).parents('.govuk-form-group').addClass('loading')` ?

web/cobrands/sass/_waste.scss
```diff
@@ -20,6 +20,10 @@ body.waste {
         font-size: 19px; font-size: 1.1875rem; line-height: 1.31579;
       }
     }
+
+    #form-submit-row.loading {
+        background: url("/cobrands/fixmystreet/images/spinner-white.gif") flip(0%,100%) 100% no-repeat;
```
zarino, line 25: Gets hidden behind the disabled button in the SocietyWorks Showcase cobrand:

Screenshot 2021-03-02 at 14 34 07

zarino (commented):
Aside from the SocietyWorks Showcase issue, I wonder whether we need to consider accessibility here? Will it be obvious to a non-sighted user what’s happening, when they press submit, the button gets disabled, and then nothing happens? Should we add something with an aria role to communicate the loading status? Looks like this can be a whole can of worms, but here are a few links I found in a quick bit of research:

* https://github.com/Shopify/polaris-react/issues/1212
* https://stackoverflow.com/questions/38518164/accessibility-page-loader-indicator-using-aria-live
* https://getbootstrap.com/docs/4.5/components/spinners/#buttons

Annoyingly, doesn‘t seem that the GOV.UK design system has any pre-built components to handle this … unless I’ve just just found it yet.

dracos:
"Will it be obvious to a non-sighted user what’s happening, when they press submit, the button gets disabled, and then nothing happens" - the form submits, like normal? There's no change to the worse here, unless I'm missing something.

web/cobrands/sass/_waste.scss
dracos: Need this deployed to prevent duplicates, so this can be fixed later

PR #3335, opened 2021-03-02
Pre-lookup some data for /report/new/ajax call.

dracos:
The call was doing repeated loops of contacts to look up bodies and
hint texts, repeated database calls to check anonymous allowed, and
making lists of body names it did not need to in templates.


davea (approved):
Worth a mention in CHANGELOG, I reckon. 👍

PR #3337, opened 2021-03-04
Ignore deleted bodies when generating responsiveness stats for /reports

davea:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2185

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/UpdateAllReports.pm
+++ b/perllib/FixMyStreet/Script/UpdateAllReports.pm
@@ -267,7 +267,7 @@ sub calculate_top_five_bodies {
 
     my(@top_five_bodies);
 
-    my $bodies = FixMyStreet::DB->resultset('Body')->search;
+    my $bodies = FixMyStreet::DB->resultset('Body')->active;
     while (my $body = $bodies->next) {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -267,7 +267,7 @@ sub calculate_top_five_bodies {
 
     my(@top_five_bodies);
 
-    my $bodies = FixMyStreet::DB->resultset('Body')->search;
+    my $bodies = FixMyStreet::DB->resultset('Body')->search({ deleted => 0 });
```
dracos, line 270: `->active` exists for this, looks like.

dracos (changes requested):
The corresponding reports/index.html template presumably needs to check `b.deleted` as well as `NOT b.area_count` now (in case a body gets deleted but is still associated with areas).

Maybe instead (as well?) UpdateAllReports.pm should skip any body not associated with any areas. Just worried in case I've missed some other case that differentiates between no-areas and deleted! And that way we can continue for recently-gone bodies having them accessible (e.g. on the dashboard). Ugh, I dunno.

PR #3339, opened 2021-03-09
Make GitHub detect the Perl correctly.

lancew:
Minor commit that makes GitHub's linguist language detection correctly identify all the .pm and .t files as Perl rather than Raku.

Added as a nicety, mainly as trying to identify why ./script/test has failing tests in my vagrant vm, but this commit when tested via CI on github actions all tests pass.


dracos (approved):
Thanks!

PR #3341, opened 2021-03-10
[TfL] Ensure all relevant ‘action scheduled’ reports are auto-closed

davea:
- Don't skip reports that have comments with empty `problem_state` value
 - Ensure `problem_state` is always set when leaving comments via the inspector form

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2208
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -501,8 +501,6 @@ sub inspect : Private {
 
                 $update_params{problem_state} = $problem->state;
 
-                my $state = $problem->state;
-
                 # If an inspector has changed the state, subscribe them to
                 # updates
                 my $options = {
@@ -590,6 +588,7 @@ sub inspect : Private {
                     confirmed => $timestamp,
                     user => $c->user->obj,
                     photo => $c->stash->{upload_fileid} || undef,
+                    problem_state => $problem->state,
                     %update_params,
                 } );
             }
--- a/perllib/FixMyStreet/Script/TfL/AutoClose.pm
+++ b/perllib/FixMyStreet/Script/TfL/AutoClose.pm
@@ -92,15 +92,16 @@ sub close_reports {
             { problem_id => $r->id },
             { order_by => 'confirmed' }
         );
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -590,6 +588,7 @@ sub inspect : Private {
                     confirmed => $timestamp,
                     user => $c->user->obj,
                     photo => $c->stash->{upload_fileid} || undef,
+                    problem_state => $problem->state,
```
dracos, line 591: Feels a bit odd to have it set update_params for the state change to decide whether to add this comment that then will set it again anyway, but it's as good an indicator as anything else I guess.

perllib/FixMyStreet/Script/TfL/AutoClose.pm
```diff
@@ -92,15 +92,16 @@ sub close_reports {
...
-        my $earliest;
+        my $old_state = '';
+        my $last_change;
         while ( my $c = $comments->next ) {
-            if ( ($c->problem_state || '') ne 'action scheduled' ) {
-                $earliest = undef;
-                next;
+            my $new_state = $c->problem_state || '';
+            if ( $new_state eq 'action scheduled' && $new_state ne $old_state) {
+                $last_change = $c->confirmed;
```
dracos, line 100: I think this will go wrong if the sequence of comment problem_states is "action scheduled", "", "action scheduled" - e.g. normal comment, inspector comment pre this PR, normal comment. Then this loop will take the last comment time as the action scheduled time, rather than the first comment as it should do.
I think it needs to retain the previous old_state if the new state is blank, rather than update it to blank.

dracos (changes requested):
Afraid I don't think the new loop gives the right result always yet.

perllib/FixMyStreet/App/Controller/Report.pm
davea: Agreed, couldn't think of a simpler way than what we already had.

dracos (approved):

PR #3343, opened 2021-03-12
[Northamptonshire] Update header and footer

mySocietyClive:
Add the new double logo for North and West Northamptonshire to match new branding, and update copyright year to 2021.
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2203

[skip changelog]


templates/web/northamptonshire/footer_extra.html
```diff
@@ -1,5 +1,5 @@
 <footer class="northamptonshire">
-    <p>© Northamptonshire County Council 2019</p>
+    <p>© Northamptonshire County Council 2021</p>
```
wrightmartin, line 2: How do we feel about replacing this with a programatically `current year`?

web/cobrands/northamptonshire/base.scss
```diff
@@ -15,13 +15,27 @@
     width: 190px;
     height: 41px;
     padding: 11px 0;
-    background: url(/cobrands/northamptonshire/img/new-ncc-logo.png) 0 50% no-repeat;
+    background: url(/cobrands/northamptonshire/img/NorthAndWestLogo.png) 0 50% no-repeat;
+    background-size: contain;
 }
 
+#nav-link {
```
zarino, line 22: I think you can replace all of this custom `#nav-link` styling with just `$menu-image: 'menu-black';` in `web/cobrands/northamptonshire/_colours.scss` (replacing the existing `$menu-image: 'menu-white';`).

web/cobrands/northamptonshire/layout.scss
```diff
@@ -1,6 +1,24 @@
 @import "_colours";
 
 @import "../sass/layout";
+@import "./docs/theme/sass/_variables";
```
zarino, line 4: Why are we including this Sass file from the documentation site?

web/cobrands/northamptonshire/base.scss
```diff
@@ -15,13 +15,27 @@
     width: 190px;
     height: 41px;
     padding: 11px 0;
```
zarino, line 17: If we’re changing stuff about this logo, why keep this odd padding rule? Surely just a width and height is required, no need for padding if we’re using just the background-image?

web/cobrands/northamptonshire/layout.scss
```diff
@@ -1,6 +1,24 @@
...
+    width: 450px;
+    
+    // Prevents logo overlapping nav items at medium screen widths    
+    @media (max-width: 60em) {
+        width: 260px;
+    }
+}
+
+#site-header {
+    padding: 20px 0;
```
zarino, line 16: Looks like somebody forgot to check what effect this would have on the map page header ;-)

Screenshot_2021-03-16 Viewing a location

You can set a custom `$mappage-header-height` in `web/cobrands/northamptonshire/_colours.scss` to control the height of the header on the map page. It defaults to `60px`, but given the padding you’ve added to `#site-header`, it looks like you need something like `104px`.

web/cobrands/northamptonshire/base.scss
zarino: I’d be tempted to remove both the `padding` and `height` here, and leave it to inherit the `height: 60px` from the core styles. It’s only 3px off the 63px you’re currently using, and it’d mean less custom code :-)

zarino (changes requested):
A few comments inline.

templates/web/northamptonshire/footer_extra.html
mySocietyClive: Good point, will do it as a separate commit as I'd imagine there's a few cobrands with copyright in footer.

mySocietyClive:
Update bits, please review again.

web/cobrands/northamptonshire/base.scss
```diff
@@ -13,15 +13,21 @@
 
 #site-logo {
     width: 190px;
-    height: 41px;
-    padding: 11px 0;
-    background: url(/cobrands/northamptonshire/img/new-ncc-logo.png) 0 50% no-repeat;
+    // height: 41px;
+    // padding: 11px 0;
```
zarino, line 17: Any reason to keep these commented-out lines?

zarino (approved):
@mySocietyClive looks good to me. One comment. Feel free to address or ignore.

web/cobrands/northamptonshire/base.scss
mySocietyClive: Good spot, I'll remove them. Thanks

Issue #3340, opened 2021-03-09
Test failures inside the vagrant VM


lancew:
When running the tests inside the development VM, there are four failing test files. When running tests via Github actions all test pass.

**To Reproduce**
1. `vagrant up` - Linux host, virtualbox provider
2. `vagrant ssh`
3. `cd fixmystreet`
4. `./script/test`

TAP output:

```
Test Summary Report
-------------------
t/app/controller/admin/templates.t        (Wstat: 256 Tests: 25 Failed: 1)
  Failed test:  24
  Non-zero exit status: 1
t/app/controller/waste.t                  (Wstat: 65280 Tests: 10 Failed: 3)
  Failed tests:  8-10
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
t/cobrand/bromley.t                       (Wstat: 65280 Tests: 28 Failed: 6)
  Failed tests:  22-24, 26-28
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
t/cobrand/tfl.t                           (Wstat: 65280 Tests: 21 Failed: 2)
  Failed tests:  20-21
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
Files=166, Tests=2826, 588 wallclock secs ( 0.42 usr  0.89 sys + 324.57 cusr 228.60 csys = 554.48 CPU)
Result: FAIL
```

Any assistance working out why these tests pass on GitHub actions CI and not when run inside the VMM I'd appreciate it. Perhaps I have missed a step?

lancew:
This PR has passing tests, for example.
https://github.com/mysociety/fixmystreet/pull/3339

dracos:
From a quick look, I guess the `templates.t` and `tfl.t` failures are because they are testing the `internal_ips` feature which the test is assuming to be 127.0.0.1, but presumably it is not actually that inside the VM somehow.

Not sure about the other two (though the Bromley failures are to do with waste tests, so presumably they are somehow linked) - can you run those tests alone (e.g. `script/test t/app/controller/waste.t` ) so that it gives you a fuller output of why they have failed?

dracos:
Ah, yes, the tests do require SOAP::Lite to be installed, but it's not installed by default as it's not needed for the base code to operate. `script/update` won't install it, as that runs `carton install --deployment --without uk --without zurich --without kiitc`. 
[...]

PR #3345, opened 2021-03-12
Skip UK specific tests, if modules not present.

lancew:
These 4 test files rely on modules only included in the UK build,
which is not done by default by script/update.

* t/app/controller/admin/templates.t
* t/app/controller/waste.t
* t/cobrand/bromley.t
* t/cobrand/tfl.t

This prevents the need to run the following, which installs missing modules
to make a fresh VM pass all tests:

 vendor/bin/carton install --deployment --without zurich --without kiitc

This commit addresses: https://github.com/mysociety/fixmystreet/issues/3340

### DIFF
--- a/t/app/controller/admin/templates.t
+++ b/t/app/controller/admin/templates.t
@@ -326,6 +326,9 @@ subtest "category groups are shown" => sub {
 };
 
 subtest "TfL cobrand only shows TfL templates" => sub {
+    eval "use Net::Subnet";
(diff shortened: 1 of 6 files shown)


dracos (commented):
These will work, but I think e.g. in templates.t it is only one test that won't pass (the TfL templates one), not the whole file, it'd be good to limit to only the necessary tests. Hopefully we can take a look soon at narrowing it down

PR #3348, opened 2021-03-19
[Bromley] Add garden waste container request.

dracos:
[skip changelog]
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2228

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -152,7 +152,7 @@ sub construct_bin_request_form {
                 type => 'Checkbox',
                 apply => [
                     {
-                        when => { "quantity-$id" => sub { $_[0] > 0 } },
+                        when => { "quantity-$id" => sub { $max > 1 && $_[0] > 0 } },
                         check => qr/^1$/,
                         message => 'Please tick the box',
                     },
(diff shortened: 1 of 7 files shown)


struan (approved):
👍

dracos:
@struan Actually, after sleeping on it, this isn't ideal because if two reports at Echo end, they're both going to be pushing updates, so we really need two reports at this end don't we. So will change this to create two reports (like it was two different types of container, but both same type with different stuff to do).

dracos:
@struan Okay, this should now create two reports for the damaged case (one Deliver, one Remove); Open311 end I'll change to accept first of submitted data rather than last, as discussed. Hopefully ready to review now.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -213,6 +218,14 @@ sub process_request_data : Private {
         $data->{detail} = "Quantity: $quantity\n\n$address";
         $c->set_param('Container_Type', $id);
         $c->set_param('Quantity', $quantity);
+        if ($reason eq 'damaged') {
+            $c->set_param('Action', 2); # Remove XXX
+            $c->set_param('Reason', 3); # Damaged XXX
+            $c->forward('add_report', [ $data ]) or return; # Additional remove report
+            $c->set_param('Action', 1); # Deliver
+        } elsif ($reason eq 'stolen' || $reason eq 'taken') {
+            $c->set_param('Reason', 1); # Missing / Stolen XXX
```
struan, line 227: might be nice to have a test for this bit too.

struan (approved):
🗑️

PR #3349, opened 2021-03-19
[Central Beds] Ignore invalid features when doing server-side NSGRef lookup

davea:
Sometimes the nearest feature has a NULL `streetref1` property but there is an overlapping feature that correctly has a `streetref1` value a very small distance away. To avoid choosing the feature with an empty `streetref1` we reject those features, forcing selection of the nearest feature that has a valid value.

[skip changelog]

For https://github.com/mysociety/fixmystreet-commercial/issues/2231


dracos (approved):
Would be good for the test mock to have an empty ref entry too?

PR #3350, opened 2021-03-22
Allow email alert radius to be specified.

### LINKED ISSUE #68: Allow distance to be set on local email alerts, not just RSS
Consider allowing email subscription to the more local RSS available feeds from

```
The simplest alert is our geographic one:
Problems within 38.1km of this location (a default distance which covers roughly 200,000 people)
(alternatively the RSS feed can be customised, within 2km / 5km / 10km / 20km) 
```

dracos:
Fixes #68 (!)


davea (approved):
As easy as that! Looks good 👍

PR #3354, opened 2021-03-25
[Northamptonshire] Changes for unitary changes

dracos:
[skip changelog] See https://github.com/mysociety/fixmystreet-commercial/issues/2215
These are the changes needed in order to rename the Northamptonshire body. Then the new unitary bodies can be created with their categories and configuration as given on the ticket.


struan (approved):
👍

PR #3356, opened 2021-03-29
[Cypress] Pin code coverage version.

dracos:
[skip changelog]

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -35,7 +35,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install -y gettext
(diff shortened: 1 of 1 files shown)


dracos:
This was a bug in code-coverage 3.9.3 which I reported to them and they've already deployed a fix in 3.9.4, so this PR isn't strictly necessary, but might as well pin to stop this happening again in future?

davea:
If it's fixed in 3.9.4, should we pin that version?

dracos:
Looking at the devdependency, we should pin something many versions old ;-) But yes, could pin that one too, I don't think there's any real change.

PR #3358, opened 2021-03-30
Update URL on /my when map moves.

dracos:
Wasn't sure if there was a reason we didn't do this when we did around/reports back in https://github.com/mysociety/fixmystreet/pull/2301 but a user has asked if they can have a "home location" when viewing their previous reports and this would be the easiest way of doing so, I think.


davea (approved):
I don't recall a reason why `/my` was skipped, no. LGTM!

PR #3360, opened 2021-03-31
[Docs] Document emergency message feature

davea:
We didn't have anywhere to point users who wanted to use this feature. Now we do.

[skip changelog]


docs/_includes/admin-tasks-content.md
```diff
@@ -926,3 +926,24 @@ particular date range:
...
+
+<div class="admin-task" markdown="1" id="emergency-message">
+
+### Displaying an emergency message
+
+<span class="admin-task__permissions">Permissions required: User must be marked as staff and ‘Add/edit emergency message’ must be ticked.</span>
+
+This features allows you to set an emergency message which will be displayed on the homepage.
+
+HTML tags can be used, but for security are limited to the same set as described in ["HTML content in notices"](#html-content-in-notices).
```
dracos, line 939: The emergency message is passed to html_para, which always adds <p>/<br> around 2+/1 newlines. So probably want to say only use inline HTML, I guess, and separate out paragraphs with newlines if need be, as per your example.

dracos (changes requested):
Looks good, thanks, one suggestion because it's not exactly the same treatment of HTML as the category notices.

PR #3363, opened 2021-03-31
[Westminster] Street Entertainment custom text

chrismytton:
Adds an asset layer for street entertainment pitches. When a pitch is selected the name and terms and conditions are displayed.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2222

## Todo

- [x]  Fix bug with mobile version going in a loop
- [x] Bug when you don't quite manage to click the asset the whole layer and the extra question disappears from view and there's just a continue button?!
- [x] Tests are written but not passing because of the mobile bug

## Demo

https://user-images.githubusercontent.com/22996/113185311-363cf100-924e-11eb-99b3-263fa246ca89.mov

### DIFF
--- a/.cypress/cypress/fixtures/westminster-street-entertainment.json
+++ b/.cypress/cypress/fixtures/westminster-street-entertainment.json
@@ -1,0 +1,26 @@
+{
+  "type": "FeatureCollection",
+  "crs": {
+    "type": "name",
+    "properties": {
+      "name": "EPSG:4326"
+    }
+  },
+  "features": [
+    {
+      "type": "Feature",
+      "geometry": {
+        "type": "Point",
+        "coordinates": [
+          -0.12690110595622592,
+          51.507530517722408
+        ]
+      },
+      "properties": {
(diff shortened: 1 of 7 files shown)


chrismytton:
This is the bug when you miss the pitch asset and click on another part of the map:

https://user-images.githubusercontent.com/22996/113187033-49e95700-9250-11eb-855d-9cd3749cb2b4.mov

chrismytton:
> "This is the bug when you miss the pitch asset and click on another part of the map:" - this is due to the same category issue, I think, in that because the category is now blank, when you click elsewhere there is no longer a matching asset layer and so it disappears.

You were right, fixing the category issue also fixed that bug, yay!

chrismytton:
@davea Ideally I'd like to get this over to the client today, so let me know if you don't think you'll have time to review and I can ask someone else.

davea (approved):
🎸 🎷 🥁 🆗

dracos:
Just to note I added UPRN field to the fetched data and stick it in a hidden field if present.

PR #3364, opened 2021-04-01
[Hackney] noise missing field in email/ subject change for noise reports

dracos:
Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2255
[skip changelog]


struan (approved):
🎺

PR #3365, opened 2021-04-06
check if staging flags are defined

struan:
Avoids the odd warning in tests

[skip changelog]


dracos (approved):
I did wonder about finding the tests and adding the appropriate flags, but this seems easier :)

PR #3366, opened 2021-04-06
Map control updates

dracos:
Removes the HTML compass from report pages because there isn't really any point in panning the map (I guess you might want to zoom in/out, which could maintain if wanted, but easier to drop both), and fixes the road map toggle button if at an aerial=1 URL.


davea (approved):
Looks slightly weird with the aerial button floating by itself but not a big deal.

image

PR #3367, opened 2021-04-06
Email alert text updates

dracos:
Improves the text on update alert signup for phone-only users (stops referring to email alerts), and fixes an issue whereby a phone-only user could sign up for a local alert but wouldn't ever get anything, so instead ask for email and do a merge, like the verify email step would do, if that makes sense. [skip changelog]


struan (approved):
☎️

PR #3370, opened 2021-04-08
[Bromley] update subscription costs as values change

struan:
Add some javascript to make the payments costs update as you change the
number of bins when creating or modifying a subscription.

[skip changelog]


chrismytton (approved):
🗑️

PR #3371, opened 2021-04-08
[Bromley] use a single category for subscriptions

struan:
Apart from cancellation all subscription actions use the same event code
in Echo so don't have different categories for each one

[skip changelog]


chrismytton (approved):
🏡

PR #3372, opened 2021-04-09
[Greenwich] Re-enable open311 groups

chrismytton:
Greenwich now have a new open311 endpoint which returns useful groups, so stop blanking out groups when fetching contacts for Greenwich.

Part of https://github.com/mysociety/fixmystreet-commercial/issues/2058


dracos (approved):
👥

PR #3373, opened 2021-04-09
[Bromley] handle service tasks with multiple schedules

struan:
When a service task is cancelled and then later renewed you end up with
multiple service task schedules which we need to handle by using the one
with the latest end date for the current schedule

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -627,7 +627,7 @@ sub bin_services_for_address {
     foreach (@$result) {
         next unless $_->{ServiceTasks};
 
-        my $servicetask = $_->{ServiceTasks}{ServiceTask};
+        my $servicetask = _get_current_service_task($_);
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -768,6 +768,28 @@ sub bin_services_for_address {
...
+    my $task = $service->{ServiceTasks}{ServiceTask};
+    my $type = ref $task;
+    return $task if $type eq 'HASH';
+
+    my ($current, $last_date);
+    if ( $type eq 'ARRAY' ) {
+        foreach ( @$task ) {
+            my $end = construct_bin_date($_->{EndDate});
+
+            next if $end && $end < $last_date;
```
dracos, line 783: should this be checking $last_date as well, or it'll have an undefined warning on first pass of loop.

dracos (changes requested):
Just a warning check I think.

dracos (approved):

PR #3374, opened 2021-04-09
[Bromley] style garden waste renewal pages

struan:
plus add in JS to auto-update costs

[skip changelog]

### DIFF
--- a/templates/web/base/waste/garden/renew.html
+++ b/templates/web/base/waste/garden/renew.html
@@ -1,7 +1,95 @@
-[% BLOCK 'content' %]
+[% extra_js = [
+    version('/js/waste.js')
+] ~%]
+[% INCLUDE 'waste/header.html' %]
+
+[% PROCESS 'govuk/fields.html' %]
+[% PROCESS title title=form.title %]
+[% PROCESS errors %]
+
+<div class="waste-service-hint-banner">
+    <div class="waste-service-hint-banner__image">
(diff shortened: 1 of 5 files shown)


templates/web/base/waste/garden/renew.html
```diff
@@ -1,7 +1,129 @@
...
-[% END %]
 
-[% INCLUDE 'waste/garden/property_page.html' %]
+<form method="post">
+  <fieldset class="govuk-fieldset">
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+      <h2 class="govuk-fieldset__heading">How many bins do you need us to collect?</h2>
+    </legend>
+
+  <!-- [% PROCESS form override_fields=['current_bins'] %] -->
```
dracos, line 32: It looks like this has been c'n'p from the subscription templates rather than refactored if they're basically the same? Though, why is the current bins number editable, if there's a separate additional bins box, for a renewal? It seems rather confusing as is. If a renewal only lets you keep same number or more, then I'd make current bins uneditable; if it's meant to also let you reduce the number of bins, then I'd only have the one box with number of bins required, because otherwise it reads like we'd want "2" in already on property and "-1" in additional bins required?

Smaller bug - if you leave either box blank, the total becomes `NaN`.

dracos (changes requested):
Think the renewal form needs a bit of thought.

dracos (approved):
Still feels the templates could share a bin row template partial of some short, but guess can leave that to later.

PR #3376, opened 2021-04-12
[Central Beds] Send problem title/detail/URL as separate Open311 attributes

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/2231

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
@@ -78,14 +78,24 @@ sub open311_extra_data_include {
         }
     }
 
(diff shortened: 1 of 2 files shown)


dracos (commented):
Does the Open311 adapter end accept these new fields? I think if it was Confirm it would, but this is Symology, so they'll need adding/handling there if so.

Would be good to have a test for the area_code_mapping stuff that apparently is not tested at all :)

davea:
Indeed, I'm working on a companion PR for open311-adapter. Agree that some tests of the area code thing would be good but was trying to be as in-and-out with this little ticket as possible.

dracos (approved):

PR #3377, opened 2021-04-12
[Bromley] Some waste fixes 

dracos:
Fix admin template setting by making sure blank external status is actually blank; show more task states than just in progress (e.g. delayed should be shown), and prefer templates over description (don't want to just blank out description as don't know if templates will be created for all states, though I guess they probably will be). [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -950,7 +950,9 @@ sub admin_templates_external_status_code_hook {
     my $task_type = $c->get_param('task_type') || '';
     my $task_state = $c->get_param('task_state') || '';
 
-    return "$res_code,$task_type,$task_state";
+    my $code = "$res_code,$task_type,$task_state";
+    $code = '' if $code eq ',,';
+    return $code;
 }
 
 sub call_api {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -950,7 +950,9 @@ sub admin_templates_external_status_code_hook {
     my $task_type = $c->get_param('task_type') || '';
     my $task_state = $c->get_param('task_state') || '';
 
-    return "$res_code,$task_type,$task_state";
+    my $code = "$res_code,$task_type,$task_state";
+    $code = '' if $code eq ',,';
```
struan, line 955: I assume there's no chance on only one of these being blank?

struan (approved):
one small comment but otherwise 💯

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: As long as the resolution code is set, the looking code in Bromley.pm looks for any of "RES,TYPE,STATE", "RES,TYPE,", "RES,,STATE", "RES,,". And the waste update code sets the code to "RES,,". So I think it's okay, though I get confused.

dracos:
Rereview  just for the last commit which is new, and fixes if the time is rescheduled but stays on the same day.

struan (approved):
👍

PR #3384, opened 2021-04-13
[Oxfordshire] Show defects on All Reports page

chrismytton:
This adds defect pins from Oxfordshire's WFS server to the All Reports page.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2175

## Todo/questions

- [x] There are a lot of pins, will this cause performance issues?
- [x] The popup offset is correct when viewing a normal size pin, but incorrect when viewing a mini pin. Is there anything we can do about that?


dracos (changes requested):
Definitely need to use the bounding box if available (in an ajax call, similar to around), but that doesn't help with the initial page load which yes, causes performance issues on my computer anyway, the fan has come on and it is rather slow when hovering etc.
Couple of possible solutions - perhaps the defects could only appear once you have zoomed in to a particular zoom level, at which point there won't be so many to show and you can continue to show all of them. Or perhaps it could show 100, same as the pagination (eg page 1 shows 100 FMS reports and 100 WFS reports, page 2 shows second 100 of both), you'd have to fetch P*100 from the WFS (it has limit but not offset, IIRC) but could then send only last 100 to the client. You could sort the response with `sortBy` of some sort.

templates/web/base/maps/pin.html
```diff
@@ -1,5 +1,5 @@
 [% DEFAULT pin_style = 'top:' _ (pin.py - 64) _ 'px; left:' _ (pin.px - 24) _ 'px; position: absolute;' -%]
-[% IF pin.id %]
+[% IF pin.id AND pin.problem %]
```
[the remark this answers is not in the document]
chrismytton: Oops, thanks, fixed in https://github.com/mysociety/fixmystreet/pull/3384/commits/b02dd89b718f0ca9e7cbbb1a3db168591561e539.

chrismytton:
@dracos I decided to go with the minimum zoom level approach in the end. The problem with loading 100 pins on the initial non-AJAX load is that if you hit a page like [this one with zoom/lat/long params](https://fixmystreet.oxfordshire.gov.uk/reports/Oxfordshire?zoom=17&lat=51.8093&lon=-1.3086) then it's unlikely that the 100 we load in will be on that bit of the map.

This means that there are no pins on the initial page load, because without the bounding box information from the AJAX call it's tricky to figure out which pins to load. Instead I'm triggering the AJAX call on page load to refresh the markers.

chrismytton:
> * The popup offset is correct when viewing a normal size pin, but incorrect when viewing a mini pin. Is there anything we can do about that?

This is solved for now by only showing defects at zoom 15 or greater, when the pins are the right size for the offset. If they want to show them at different zoom levels then I'll need to adjust the offset of the popup based on the zoom or something, eugh.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -588,6 +588,12 @@ sub load_and_group_problems : Private {
         }
     }
 
+    my $bbox = $c->get_param('bbox');
```
dracos, line 591: $c is available inside the cobrand as $self->{c}, so easier to look up these two vars there and not bother the controller about it?

web/js/map-OpenLayers.js
```diff
@@ -1267,16 +1267,17 @@ OpenLayers.Protocol.FixMyStreet = OpenLayers.Class(OpenLayers.Protocol.HTTP, {
...
             page = $('.pagination:first').data('page');
             this.use_page = false;
         } else if (this.initial_page) {
             page = 1;
         } else {
-            var qs = fixmystreet.utils.parse_query_string();
             this.initial_page = page = qs.p || 1;
         }
         options.params.p = page;
+        options.params.zoom = qs.zoom;
```
dracos, line 1280: This is getting the zoom parameter from the URL before the zoom takes effect, and so is incorrect (you can see this is you start at zoom 15 and zoom out, you still see defect pins because 15 is sent to the server). I also wonder if there's a case where there might not be a zoom in the URL for some reason. Anyway, I think `fixmystreet.map.getZoom()` returns the current zoom level, no need to look at the URL.

dracos (changes requested):
There's a bug in which zoom level is sent to the server, and a tidying suggestion.

"The problem with loading 100 pins on the initial non-AJAX load is that if you hit a page like this one"- hmm, I guess you'd want it to be showing the correct normal reports there as well in that case. Perhaps if the URL has a lat/lon in it, the server should be doing the same as around in this case when given a lat/lon and make up a little bounding box and using that. Anyway, not needed for now, this looks good.

perllib/FixMyStreet/App/Controller/Reports.pm
chrismytton: Ah didn't know that! Yes much easier to just do it in the cobrand, ta.

dracos (approved):

PR #3387, opened 2021-04-13
Nothing links these things

dracos:
Remove Hackney/Northants analytics; hopefully fix occasional Cypress test failure in bathnes.js; stop sending questionnaires for reports made as body or anonymous user (fixes https://github.com/mysociety/fixmystreet-commercial/issues/2267); stop a "0" being printed on scrubbed users in admin report edit. Couldn't face 4 PRs, they're all tiny. [skip changelog]


t/app/controller/admin/users.t
```diff
@@ -685,7 +685,9 @@ subtest "Removing account from admin" => sub {
     $user->discard_changes;
     is $user->name, '', 'Name gone';
     is $user->password, '', 'Password gone';
-    is $user->email, 'removed-' . $user->id . '@example.org', 'Email gone'
+    is $user->email, 'removed-' . $user->id . '@example.org', 'Email gone';
+    $mech->get_ok( '/admin/report_edit/' . $user->problems->first->id );
+    $mech->content_contains("name='username' value=''>")
```
davea, line 690: Missing the trailing semicolon here - might not matter syntactically but saves a future headache if it's not spotted.

davea (approved):
One tiny optional thing.

t/app/controller/admin/users.t
dracos: 👍 WEIRD how I sort of copied that from what it was before!

PR #3389, opened 2021-04-14
CSV export, improve Site Used and add Device Type.

dracos:
Since we started recording mobile/desktop website use in service, Site Used now contains that, rather than the cobrand used. Rename Site Used to Device Type (and only show service column), and add a new Site Used column that always contains the cobrand used. (This was already done for the Highways England cobrand.)


davea (approved):
Looks good. I wonder if we'd need to notify existing users about the change so any stats they're generating from the CSV aren't adversely affected.

PR #3390, opened 2021-04-14
No questionnaires if reported as other, no details

dracos:
This is similar to reporting as anonymous, in that it uses the staff user account, except you could have provided a different name. (Also see previous commit de8084885daffd3f4203cc46cfa3edf4d0a1dc0d)


struan (approved):
❔

PR #3393, opened 2021-04-15
[Bromley] tidy up echo api mocking in tests

struan:
Remove a load of duplication and be more explicit about setting data
where we are relying on it

[skip changelog]


dracos (approved):
I guess at the end we can have another look at this type of thing in a final tidying.

PR #3394, opened 2021-04-16
configure FLOC disabling header on a per cobrand basis

struan:
Add a cobrand feature to send a header to remove the site from being
used as part of FLOC tracking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
     - Development improvements:
         - Include failure count in send report error output, #3316
         - Sort output in export script. #3323
+        - Add config option for disabling FLOC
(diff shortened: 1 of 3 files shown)


dracos:
Looks okay, but I think this is probably better suited to the web server or proxy server than the code (the reason we added CSP in the code is because it has to set a nonce that is included in the HTML output as well) - we could include this header for everywhere (or whatever subset) in the webserver directly.

dracos:
And we do already have a cobrand add_response_headers function someone could use if they wanted it on in their cobrand and couldn't change their webserver for some reason. Being covered in https://github.com/mysociety/sysadmin/issues/1437

PR #3398, opened 2021-04-16
[Buckinghamshire] make it clear only bucks streelights are displayed

struan:
plus a small tweak to enable multiple asset item placeholders in message string.

[skip changlog]

### DIFF
--- a/web/cobrands/buckinghamshire/assets.js
+++ b/web/cobrands/buckinghamshire/assets.js
@@ -83,7 +83,18 @@ fixmystreet.assets.add(labeled_defaults, {
         'Street light dim',
         'Street light intermittent',
         'Street light not working' ],
-    asset_item: 'street light'
+    asset_item: 'street light',
+    asset_item_message: 'You can pick a <b class="asset-ITEM">ITEM</b> from the map &raquo;<br>If there is no yellow dot showing then this ITEM is not maintained by Transport for Buckinghamshire',
+    construct_selected_asset_message: function(asset) {
+        var id = asset.attributes[this.fixmystreet.feature_code] || '';
+        if (id === '') {
+            return;
+        }
+        var data = this.fixmystreet.construct_asset_name(id);
+        var extra = '. Only ITEMs maintained by Transport for Buckinghamshire are displayed.';
+        extra = extra.replace(/ITEM/g, data.name);
+        return 'You have selected ' + data.name + ' <b>' + data.id + '</b>' + extra;
+    }
 });
 
 fixmystreet.assets.add(labeled_defaults, {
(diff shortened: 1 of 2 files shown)


web/cobrands/buckinghamshire/assets.js
```diff
@@ -83,7 +83,8 @@ fixmystreet.assets.add(labeled_defaults, {
         'Street light dim',
         'Street light intermittent',
         'Street light not working' ],
-    asset_item: 'street light'
+    asset_item: 'street light',
+    asset_item_message: 'You can pick a <b class="asset-ITEM">ITEM</b> from the map &raquo;<br>If there is no yellow dot showing then this ITEM is not maintained by Transport for Buckingham'
```
dracos, line 87: Buckinghamshire I assume this should be

web/cobrands/fixmystreet/assets.js
```diff
@@ -485,7 +485,8 @@ function get_asset_pick_message() {
     var message;
     if (typeof this.fixmystreet.asset_item_message !== 'undefined') {
         message = this.fixmystreet.asset_item_message;
-        message = message.replace('ITEM', this.fixmystreet.asset_item);
+        var re = new RegExp('ITEM', 'g');
+        message = message.replace(re, this.fixmystreet.asset_item);
```
dracos, line 489: This is fine, but note that `replace(/ITEM/g, ...)` would work just as well :)

dracos (changes requested):
Typo in text, I think

web/cobrands/fixmystreet/assets.js
struan: I couldn't find anything definitive about browser support for that notation so decided to play it safe.

dracos: It's the same :)

dracos (approved):

struan:
re-review because extra message added when a light is selected.

web/cobrands/buckinghamshire/assets.js
```diff
@@ -84,7 +84,17 @@ fixmystreet.assets.add(labeled_defaults, {
...
+    asset_item_message: 'You can pick a <b class="asset-ITEM">ITEM</b> from the map &raquo;<br>If there is no yellow dot showing then this ITEM is not maintained by Transport for Buckinghamshire',
+    construct_selected_asset_message: function(asset) {
+        var id = asset.attributes[this.fixmystreet.feature_code] || '';
+        if (id === '') {
+            return;
+        }
+        var data = this.fixmystreet.construct_asset_name(id);
+        var extra = ' Only ITEMs maintained by Transport for Buckinghamshire are displayed.';
+        extra = extra.replace(/ITEM/g, data.name);
+        return 'You have selected ' + data.name + ' <b>' + data.id + '</b>' + extra;
```
dracos, line 96: Does it need some punctuation between the ID and the "Only..." text?

dracos (changes requested):
One text suggestion is all

dracos (approved):

PR #3400, opened 2021-04-20
do not allow blank category on triage

struan:
Prevent the user from unsetting the category when triaging a report.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Triage.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Triage.pm
@@ -111,6 +111,14 @@ sub update : Private {
     my $current_category = $problem->category;
     my $new_category = $c->get_param('category');
 
+    if (!$new_category) {
+        my $errors = $c->stash->{errors} || [];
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Admin/Triage.pm
```diff
@@ -111,6 +111,14 @@ sub update : Private {
     my $current_category = $problem->category;
     my $new_category = $c->get_param('category');
 
+    if (!$new_category) {
+        my $errors = $c->stash->{errors} || [];
+        push @$errors, _"Please choose a category";
+        $c->stash->{errors} = $errors;
+
+        $c->detach('/report/id', [ $problem->id ] );
```
dracos, line 119: I see this appears to work, but I think it's a bit confusing to go back to this (and it only not infinite-loop because it's the midpoint of a chain); would a $c->detach just to stop processing here not do the same job? (Note, I haven't actually checked this, sorry)

dracos (approved):

PR #3404, opened 2021-04-22
OpenLayers updates

dracos:
Turns out the WMS file had never been updated from initial creation, so missed all the touch event fixes, the CSP fixes, the removal of Permalink and so on. Hopefully all up to date with the others now (which I also updated to remove unneeded getViewportElement).


struan (approved):
I think this looks ok. Not the easiest thing in the world to review :|

PR #3405, opened 2021-04-22
Add image redaction facility

dracos:
First adds per-photo moderation to the front end rather than it being all/nothing, which seemed like a prerequisite.
Then adds a way to redact sections of photos, replacing the photo. The original is kept as with a removal.

The diff coverage is actually higher than it says, I think, because JS coverage does not include coverage of the Perl bits tested.


davea:
It's possible to end up with the redacted and unredacted versions of a photo displayed - wasn't sure if this was intentional/desired?

1. Moderate a report, redact a portion of the photo, save moderation changes.
2. Re-moderate the report and click "restore photo" below now-hidden original photo, save moderation changes.
3. Notice that redacted and unredacted version of photo is shown.


Screen recording of the above steps

dracos:
> It's possible to end up with the redacted and unredacted versions of a photo displayed - wasn't sure if this was intentional/desired?

I guess *ideally* you'd want it to keep a record of which photo it was originally that has been redacted and then note that upon restoration to automatically remove the redacted version. But that seemed like quite a faff to do (as it is, it just sees the redacted photo as a new photo on the report once it exists) and it seemed like clicking "Restore photo" for the original and at same time/later untick "Show photo" on the redacted was easy enough, as I assumed this would happen exceedingly rarely. And I guess it's *possible* you'd want to reshow the photo but retain the picture of what had previously been redacted?

davea (changes requested):
Looks good!

Some cobrands need the text colour checking though, e.g. Oxfordshire: 
image
(sorry, I didn't look at all of them!)

davea:
One other thing that's just occurred to me (sorry!) is that this PR should update the docs too so we have something to point users at.

dracos:
"Some cobrands need the text colour checking though, e.g. Oxfordshire:" - ah, this is fixed by setting titlePosition:'outside' which then always sets a text colour on that text.

dracos:
Have added some docs :)

davea (approved):
Nice!

PR #3406, opened 2021-04-22
[CheshireEast] Use new selectedCategory method in JS

chrismytton:
Cheshire East have been in touch to say that the change in https://github.com/mysociety/fixmystreet/pull/3232 to allow reports in cleaning categories to be made anywhere isn't working.

After a bit of investigation it turns out that the original code was written and reviewed before the change that introduced the `selectedCategory` helper function, but was merged into master without being re-reviewed (my bad).

So this changes the code to use the new helper function so it works as expected.

Follows on from the original PR that introduced this change: https://github.com/mysociety/fixmystreet/pull/3232

Related to https://github.com/mysociety/fixmystreet-commercial/issues/2239


struan (approved):
🗑️

PR #3410, opened 2021-04-26
[Bromley] cache bin service lookups for speed

dracos:
Factored out from other PRs. Already checked, but I changed it to cache inside call_api given both calls were doing the same thing. Hoping to put this live now rather than wait.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -422,6 +422,15 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+sub clear_cached_lookups {
+    my ($self, $id) = @_;
+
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -567,7 +577,7 @@ sub bin_services_for_address {
     }
     push @to_fetch, GetTasks => \@task_refs if @task_refs;
 
-    my $calls = $self->call_api(@to_fetch);
+    my $calls = $self->call_api('bin_services_for_address' . $property->{id}, @to_fetch);
```
struan, line 580: This is missing a ':' at the end of the key.

struan (changes requested):
missing : aside this looks good.

struan (approved):
👍

PR #3411, opened 2021-04-26
Process update file

struan:
Script to process a bulk update file, plus some changes to support it.

[skip changelog]


bin/fixmystreet.com/process_update_file
```diff
@@ -0,0 +1,36 @@
...
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use Getopt::Long::Descriptive;
+use FixMyStreet::Script::ProcessUpdateFile;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['file|f:s', 'file containing updates' ],
```
dracos, line 22: I think if you put f=s and b=s, and { required => 1}, that will make it mandatory and require the argument, so you don't need to do the file/body checking below.

perllib/FixMyStreet/Script/ProcessUpdateFile.pm
```diff
@@ -0,0 +1,71 @@
...
+);
+
+sub process {
+    my $self = shift;
+
+    die "Problem loading body\n" unless $self->body;
+
+    my $updates = Open311::GetServiceRequestUpdates->new(
+        verbose => $self->verbose,
+        body => $self->body_name,
```
dracos, line 64: This line isn't needed, I think, only fetch uses this variable?

dracos (changes requested):
Couple of tiny suggestions. Feels like the code could potentially be rejigged to separate the open311 bits a bit better maybe, but is fine.

dracos (approved):
🆙 📅

PR #3412, opened 2021-04-27
[Westminster] Rename Street Entertainment category

chrismytton:
Westminster have renamed this category at their end, so update the code to match.

[skip changelog]


struan (approved):
🎺

PR #3414, opened 2021-04-28
[Zürich] Fix overdue_alerts script

davea:
Should stop cron errors about ‘Undefined subroutine
&FixMyStreet::Cobrand::Zurich::is_public_holiday’

[skip changelog]


dracos (approved):
Looks good

PR #3415, opened 2021-04-28
Bromley waste phase2 small fixes

struan:
[skip changelog]

### DIFF
--- a/perllib/Integrations/SCP.pm
+++ b/perllib/Integrations/SCP.pm
@@ -100,12 +100,36 @@ sub pay {
                 }
             },
         },
-        'sale' => {
+        'sale' => ixhash(
             'scpbase:saleSummary' => ixhash(
                 'scpbase:description' => $args->{description},
                 'scpbase:amountInMinorUnits' => $args->{amount},
-            )
-        },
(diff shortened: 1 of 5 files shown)


perllib/Integrations/SCP.pm
```diff
@@ -100,12 +100,36 @@ sub pay {
...
+                        ),
+                        'scpbase:tax' => {
+                            'scpbase:vat' =>ixhash(
+                                'scpbase:vatCode' => $self->config->{scp_vat_code},
+                                'scpbase:vatRate' => $self->config->{scp_vat_rate} || 0,
+                                'scpbase:vatAmountInMinorUnits' => $args->{vat} || 0,
+                            ),
+                        },
+                        'scpbase:lgItemDetails' => ixhash(
+                            'scpbase:fundCode' => $self->config->{scp_fund_code},
```
davea, line 125: Will this (and `scp_vat_code` above) always be present in config? What happens if they're not there?

davea (approved):
One question is all, but 👍

perllib/Integrations/SCP.pm
struan: at the moment they should always be there.

PR #3417, opened 2021-04-28
Make anonymous comments clearer in alert emails

chrismytton:
When showing a list of comments in HTML emails we previously didn't show anything for updates that were left anonymously. This led to some confusion about which text related to which update.

This change means that anonymous comments will show the text "Posted anonymously" next to them, which should make things a bit clearer.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
         - Improve `#geolocate_link` display, especially for smaller screens. #2048
         - Allow email alert radius to be specified. #68
         - Update URL on /my when map moves. #3358
+        - Make anonymous updates clearer in email alerts. #3417
     - Bugfixes:
         - Fix non-JS form when all extra questions answered. #3248
         - Improve display of disabled fields in iOS.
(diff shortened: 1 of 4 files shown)


chrismytton:
Manually testing this was a bit awkward. I ended up changing this line to have `item_anonymous => 1` so all updates appeared as anonymous on `/_dev/email/alert-update`. There may be a better way!

https://github.com/mysociety/fixmystreet/blob/04de4f97ff05e5662965d71e1bdae6c639f7d039/perllib/FixMyStreet/App/Controller/Develop.pm#L102

davea (approved):
Nice and neat. Agreed that it's a bit awkward to test manually. I guess you could have the `$q` on L106 fetch updates from the report being displayed and fallback to the existing behaviour if there weren't any.

chrismytton:
> I guess you could have the `$q` on L106 fetch updates from the report being displayed and fallback to the existing behaviour if there weren't any.

@davea This works really well, thanks! I've added an extra commit to the end of this PR which does this, would you mind taking a look?

davea (approved):
Looks good! :)

(don't forget to check that `fixamingata` template and ack the cobrand-templates check)

PR #3418, opened 2021-04-28
Handle first map page asset not found better.

davea:
On mobile, the message controller code moves an asset not found message to within the map box (given that takes up the whole page) and displays it based upon a map-page class. This class was added to subsequent map pages (e.g. asset selection), but not the first map page, which meant if the asset selection failure happened then (e.g. you have to click on a road to continue), the Continue button was hidden but no message was shown, which was confusing.

[skip changelog]

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/2325


dracos (approved):
Oh, nice, didn't realise there was already a non-mobile test for it, thanks :)

PR #3420, opened 2021-04-30
[Bromley] check for pending DD sub on bin days page

struan:
If there is one then display a message and prevent them trying to sign
up for another one.

[skip changelog]


struan:
this was done in another branch

PR #3422, opened 2021-05-05
[Buckinghamshire] Red claims changes

dracos:
[skip changelog]
Starts with some small fixups and refactorings, and a bug fix or two, then adds the changes from https://github.com/mysociety/fixmystreet-commercial/issues/2067 - Other cause, start page text tweaks, some JS improvements, preventing progression if not fixed and not reported, always need to map location, actually storing the report and sending by Open311/email, back button, multiple photos, and attaching the documents to the email.


perllib/FixMyStreet/App/Form/Claims.pm
```diff
@@ -478,12 +478,22 @@ has_field what_cause => (
     required => 1,
     label => 'What was the cause of the incident?',
     options => [
-        { label => 'Bollard', value => 'bollard' },
-        { label => 'Cats Eyes', value => 'catseyes' },
-        { label => 'Debris', value => 'debris' },
+        { label => 'Bollard', value => 'bollard', data_hide => '#form-what_cause_other-row' },
+        { label => 'Cats Eyes', value => 'catseyes', data_hide => '#form-what_cause_other-row' },
+        { label => 'Debris', value => 'debris', data_hide => '#form-what_cause_other-row' },
+        { label => 'Other', value => 'other', data_show => '#form-what_cause_other-row' },
```
struan, line 489: FWIW, I'm not sure this list is comprehensive, I only pulled enough in from the prototype to test but I don't know where Martin got the full list in the prototype.

perllib/FixMyStreet/SendReport/Open311.pm
```diff
@@ -82,17 +82,21 @@ sub send {
...
+                $self->open311_test_req_used($open311->test_req_used);
+            }
         }
 
         # make sure we don't save extra changes from above
         $row->set_extra_fields( @$original_extra );
 
-        if ( $resp ) {
+        if ( $skip || $resp ) {
             $row->external_id( $resp );
```
struan, line 100: this isn't going to set external_id to anything in the case of skip which I guess is fine?

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -410,6 +411,60 @@ sub lookup_site_code_config { {
...
+            my $filename = $_->{filenames}[0];
+            my $id = $_->{files};
+            my $dir = FixMyStreet->config('PHOTO_STORAGE_OPTIONS')->{UPLOAD_DIR};
+            $dir = path($dir, "claims_files")->absolute(FixMyStreet->path_to());
+            my $data = path($dir, $id)->slurp_raw;
+            push @attachments, {
+                body => $data,
+                attributes => {
+                    filename => $filename,
+                    #content_type => $image->{content_type},
```
struan, line 457: unwanted comment?

struan (commented):
I think this looks ok, a few small comments and I'll have another look tomorrow.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
dracos: Wasn't sure if we wanted the read the content type from somewhere somehow, but guess can see if it's okay as is, so that's why I left it.

perllib/FixMyStreet/SendReport/Open311.pm
dracos: Yeah, will set to undef.

perllib/FixMyStreet/App/Form/Claims.pm
dracos: Ah, right, thanks, have checked against the prototype and updated the missing options here and on where cause, and fixed a few missing hints too

templates/web/base/claims/index.html
```diff
@@ -17,8 +15,14 @@
   [% PROCESS form %]
 </form>
 
+<button id='form-back' class="hidden-nojs govuk-button govuk-button--secondary">Back</button>
```
struan, line 18: This puts a back button on the first page which is a bit confusing.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -410,6 +411,60 @@ sub lookup_site_code_config { {
...
+                    encoding => 'base64', # quoted-printable ends up with newlines corrupting binary data
+                    name => $filename,
+                },
+            };
+        }
+
+        $params->{_attachments_} = \@attachments;
```
struan, line 464: I did manage to get an "email too large error" which still registered the report as sent by attaching two 5 Mb PDFs.

struan (changes requested):
Looks pretty good other than the email error and back button placement.

Also, the confirmation screen is still a debug one, not sure if that's something for later?

struan (approved):
👍

PR #3425, opened 2021-05-05
[Bromley] waste direct debit payment processing

struan:
Fetch a list of recent Direct Debit payments and match them to up to the
relevant report in FixMyStreet and then process accordingly.

For new and ad-hoc payments (the latter is for adding a bin half way
through) find the matching report and confirm it.

For renewals and cancellations we need to find the original subscription
report and then use that to create a new report that then cancels or
renews the subscription.

Issues warnings for reports not handled.

[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/bromley-reconcile-dd
+++ b/bin/fixmystreet.com/bromley-reconcile-dd
@@ -1,0 +1,15 @@
+#!/usr/bin/env perl
+
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
+use FixMyStreet::Cobrand::Bromley;
+
+my $cb = FixMyStreet::Cobrand::Bromley->new;
+$cb->waste_reconcile_direct_debits;
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -329,7 +329,8 @@ sub direct_debit_modify : Path('dd_amend') : Args(0) {
         my $dt = $c->cobrand->waste_get_next_dd_day;
 
         my $one_off_ref = $i->one_off_payment( {
-                payer_reference => 1, # XXX
+                # this will be set when the initial payment is confirmed
+                payer_reference => $c->stash->{orig_sub}->get_extra_metadata('payerReference'),
                 amount => sprintf('%.2f', $ad_hoc / 100),
                 reference => $p->id,
                 comments => '',
@@ -337,7 +338,7 @@ sub direct_debit_modify : Path('dd_amend') : Args(0) {
     }
 
     my $update_ref = $i->amend_plan( {
-        payer_reference => 1, # XXX
+        payer_reference => $c->stash->{orig_sub}->get_extra_metadata('payerReference'),
         amount => sprintf('%.2f', $total / 100),
     } );
 }
@@ -350,7 +351,7 @@ sub direct_debit_cancel_sub : Path('dd_cancel_sub') : Args(0) {
     my $i = Integrations::Pay360->new( { config => $c->cobrand->feature('payment_gateway') } );
 
     my $update_ref = $i->cancel_plan( {
-        payer_reference => 1, # XXX
+        payer_reference => $c->stash->{orig_sub}->get_extra_metadata('payerReference'),
     } );
 }
 
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -574,6 +574,20 @@ sub bin_payment_types {
     };
 }
 
+sub waste_subscription_types {
+    return {
+        New => 1,
+        Renew => 2,
+        Amend => 3,
+    };
+}
+
+sub waste_container_actions {
+    return {
+        deliver => 1,
+        remove => 2
+    };
+}
 
 sub bin_services_for_address {
     my $self = shift;
@@ -590,10 +604,7 @@ sub bin_services_for_address {
         45 => 'Wheeled Bin (Food)',
     };
 
-    $self->{c}->stash->{container_actions} = {
-        deliver => 1,
-        remove => 2
-    };
+    $self->{c}->stash->{container_actions} = $self->waste_container_actions;
 
     my %service_to_containers = (
         535 => [ 1 ],
@@ -617,11 +628,7 @@ sub bin_services_for_address {
 
     $self->{c}->stash->{quantity_max} = \%quantity_max;
 
-    $self->{c}->stash->{garden_subs} = {
-        New => 1,
-        Renew => 2,
-        Amend => 3,
-    };
+    $self->{c}->stash->{garden_subs} = $self->waste_subscription_types;
 
     my $result = $self->{api_serviceunits};
     return [] unless @$result;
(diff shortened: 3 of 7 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+                my $sub_type = $cur->get_extra_field_value('Subscription_Type');
+                if ( $sub_type eq $self->waste_subscription_types->{New} ) {
+                    $p = $cur;
+                } elsif ( $sub_type eq $self->waste_subscription_types->{Renew} ) {
+                    # already processed
+                    next RECORD if $cur->get_extra_metadata('dd_date') && $cur->get_extra_metadata('dd_date') eq $date;
+                }
+            }
+            if ( $p ) {
+                my $service = $self->waste_get_current_garden_sub( $p->get_extra_field_value('property_id') );
```
dracos, line 1231: Is property_id always going to be same as $uprn?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+                $renew->set_extra_metadata('dd_date', $date);
+                $renew->insert;
+                $handled = 1;
+            }
+        } elsif ( $category eq 'Cancel' ) {
+            next unless $payment->{Status} eq 'Processed';
+            $rs = $rs->search({ category => { -in => ['Garden Subscription', 'Cancel Garden Subscription'] } });
+            my $p;
+            while ( my $cur = $rs->next ) {
+                my $sub_type = $cur->get_extra_field_value('Subscription_Type');
```
dracos, line 1268: Will this be undefined on cancel reports, so needs an `|| ''`?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+
+    RECORD: for my $payment ( @$recent ) {
+
+        my $date = $payment->{CollectionDate};
+        my ($category, $type) = $self->waste_payment_type ( $payment->{Type}, $payment->{YourRef} );
+
+        next unless $category && $date;
+
+        my $payer = $payment->{PayerReference};
+        my $reference = $payer;
```
dracos, line 1203: Any reason to have both $payer and $reference?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+                ) {
+                    next RECORD;
+                }
+
+                if ( $type eq $self->waste_subscription_types->{New} ) {
+                    if ( !$cur->get_extra_metadata('payerReference') ) {
+                        $cur->set_extra_metadata('payerReference', $payer);
+                    }
+                }
+                $cur->set_extra_metadata('dd_date', $date);
```
dracos, line 1334: If someone has multiple New reports (signed up, cancelled, sign up again, presumably), and we've got a payment with no YourRef, it feels like all those New reports will get here - the newest one will be unconfirmed, and the older ones will presumably not match on `dd_date`. Then they'll all have their dd_date updated. I might have misunderstood.

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+                    },
+                );
+                $cancel->set_extra_metadata('dd_date', $date);
+                $cancel->insert;
+                $handled = 1;
+            }
+        } else {
+            next unless $payment->{Status} eq 'Paid';
+            # we fetch the confirmed ones as well as we explicitly want to check for
+            # processed reports so we can warn on those we are missing.
```
dracos, line 1312: I don't see how this comment is implemented - either a next RECORD line is hit, or handled is set to 1, I can't see how it will ever get to the unhandled warning.

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1146,6 +1153,255 @@ sub garden_waste_cost {
...
+        my $rs = FixMyStreet::DB->resultset('Problem')->search({
+            extra => { like => '%uprn,T5:value,I' . $len . ':'. $uprn . '%' },
+        },
+        {
+                order_by => { -desc => 'created' }
+        });
+
+        my $handled;
+
+        if ( $type && $type eq $self->waste_subscription_types->{Renew} ) {
```
dracos, line 1217: This is quite a big block; I see the commit message, but might be worth a bit more explanation? It's not that easy to follow. So currently, if I've understood it right:
1. If the payment is a renew, skip if we find a Renew report with the same date, but if we find a New report, take the oldest one (should that be the newest one? Or I guess will they all be identical anyway...), and create a copy of it as a Renew report. (That copy means next time it'll see that and skip)
2. If the payment is a processed cancel, skip if we find a Cancel report (why is this a different category to the others?) with the same date, if we find a New report (again, the oldest one?), create a duplicate Cancel report.
3. Otherwise, if a paid payment, fetch all reports (except cancel ones), skip any if reference/date matches, or a New report where date matches, otherwise set payerReference if not set, set dd_date, and confirm the report (whatever it is).

Not sure, but would it perhaps be simpler to do one loop of the matching reports? Dunno, just feels like it could be made more straightforward somehow.

dracos (changes requested):
Some small questions and suggestions, not sure I've understood the payment bit for new/ad-hoc.

perllib/Integrations/Pay360.pm
```diff
@@ -112,6 +112,65 @@ sub amend_plan {
     }
 }
 
+sub get_payers {
```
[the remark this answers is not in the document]
struan: This and the other unused function I've left in because it's useful for getting debugging info.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -317,15 +317,15 @@ sub direct_debit_modify : Path('dd_amend') : Args(0) {
         my $dt = $c->cobrand->waste_get_next_dd_day;
 
         my $one_off_ref = $i->one_off_payment( {
-                payer_reference => 1, # XXX
+                payer_reference => $p->get_extra_metadata('payerReference'),
```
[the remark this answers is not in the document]
struan: This is set when the initial DD is processed and the subscription isn't created in Echo until then so it's not possible to amend it before this is set.

perllib/FixMyStreet/Cobrand/Bromley.pm
struan: I don't understand this comment.

dracos: You've set $uprn a few lines up, but then are fetching this property_id here - if it's always going to be the same as $uprn, you could use that? Oh, as I have written that I realise that it is the ID not the UPRN, so can ignore me.

perllib/FixMyStreet/Cobrand/Bromley.pm
struan: Cancel is a different category as it is a different event code in Echo.

bin/fixmystreet.com/bromley-reconcile-dd
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use v5.14;
```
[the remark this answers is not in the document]
dracos: Ah, it doesn't do use warnings though!

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1256,24 +1258,40 @@ sub waste_reconcile_direct_debits {
...
+                    }
                 }
             }
-            if ( $p ) {
+            if ( $r ) {
+                my $service = $self->waste_get_current_garden_sub( $r->get_extra_field_value('property_id') );
+                # if there's not a service then it's fine as it's already been cancelled
+                if ( $service ) {
+                    $r->set_extra_metadata('dd_date', $date);
+                    $r->state('confirmed');
```
dracos, line 1289: So if somehow a service has already been cancelled, but this $r report is still unconfirmed, the $payment will be processed each time it's fetched with no outcome? Should we mark the report as 'hidden' or something, if it's a report that doesn't need to be confirmed or do anything?

dracos (changes requested):
Looking good, just one query on the new cancelled report thingy.

dracos (approved):

PR #3427, opened 2021-05-10
[Bromley] Improve max-per-day waste lookup.

dracos:
[skip changelog]


struan (approved):
Might be nice to add a comment explaining what the rules are. I read the code as allowing a maximum number of property looks ups per day, but with as many look ups of the same property?

PR #3431, opened 2021-05-12
[Hackney] Noise closing script

dracos:
Refactors the Bucks script to be shared (and removes Bucks district code). Also stores all MapIt areas in noise like normal reports. [skip changelog]

### DIFF
--- a/bin/fixmystreet.com/hackney-noise
+++ b/bin/fixmystreet.com/hackney-noise
@@ -1,0 +1,37 @@
+#!/usr/bin/env perl
+#
+# If a noise report within Hackney has not been updated
+# in four weeks and is open, close it with a message.
+
+use v5.14;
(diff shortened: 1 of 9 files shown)


bin/fixmystreet.com/hackney-noise
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env perl
+#
+# If a noise report within Hackney has not been updated
+# in four weeks and is open, close it with a message.
```
chrismytton, line 4: This says 4 weeks but from what I can tell you need to supply the `--days` parameter to this script.

Do you think it should default to 28 days if there's no `--days` argument?

chrismytton (approved):
One minor comment, but otherwise looks good!

PR #3433, opened 2021-05-13
Bromley waste phase2 phase1 fixes

struan:
Small fixes to changes made for phase 1.
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1143,8 +1143,6 @@ sub add_report : Private {
 
     $c->cobrand->call_hook( clear_cached_lookups => $c->stash->{property}{id} );
 
-    $c->cobrand->call_hook( 'clear_cached_lookups' => $c->stash->{property}{id} );
-
     return 1;
 }
 
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -260,6 +260,7 @@ sub confirm_subscription : Private {
     );
     $p->set_extra_metadata('payment_reference', $reference) if $reference;
     $p->confirm;
+    $c->forward( '/report/new/create_related_things', [ $p ] );
     $c->stash->{property_id} = $p->get_extra_field_value('property_id');
     $p->update;
     $c->stash->{template} = 'waste/garden/subscribe_confirm.html';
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1129,7 +1130,14 @@ sub add_report : Private {
 
     $c->forward('setup_categories_and_bodies') unless $c->stash->{contacts};
     $c->forward('/report/new/non_map_creation', [['/waste/remove_name_errors']]) or return;
-    $c->forward('/report/new/redirect_or_confirm_creation');
+    if ( $no_confirm ) {
+        my $report = $c->stash->{report};
+        $report->state('unconfirmed');
```
dracos, line 1135: This is presumably for if you are logged in, or the report was created by a staff member and so already confirmed by this point? Do you want to leave a comment about that?

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1129,7 +1130,14 @@ sub add_report : Private {
...
+        $report->confirmed(undef);
+        $report->update;
+    } else {
+        $c->forward('/report/new/redirect_or_confirm_creation');
+    }
 
     $c->cobrand->call_hook( clear_cached_lookups => $c->stash->{property}{id} );
```
dracos, line 1142: Same hook twice?

dracos (changes requested):

perllib/FixMyStreet/App/Controller/Waste.pm
struan: doh, this is a rebase fail :(

perllib/FixMyStreet/App/Controller/Waste.pm
struan: No, this is because we don't want to confirm a report that we've not processed the payment for, so yes, I should totally add a comment.

struan: I suspect there's some more work to be done to confirm email addresses elsewhere but with this I just want to make it work as is.

dracos: Yes, sorry, that's what I meant - it would only have already been marked confirmed by this point if they were already logged in or a staff member.

dracos (approved):

PR #3434, opened 2021-05-13
[Rutland] Show group and category hint text from Salesforce

chrismytton:
Rutland have started including some HTML hint text in their Salesforce instance that they want to display on FixMyStreet.

I've added the Salesforce bits to open311-adapter in https://github.com/mysociety/open311-adapter/pull/188, so this pull request follows on/is linked to that pull request.

Part of https://github.com/mysociety/societyworks/issues/2206

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Rutland.pm
+++ b/perllib/FixMyStreet/Cobrand/Rutland.pm
@@ -40,6 +40,20 @@ sub open311_extra_data_include {
     ];
 }
 
+sub open311_contact_meta_override {
+    my ($self, $service, $contact, $meta) = @_;
+
+    my ($hint) = grep { $_->{code} eq 'hint' } @$meta;
+    my ($group_hint) = grep { $_->{code} eq 'group_hint' } @$meta;
+    @$meta = grep { $_->{code} ne 'hint' && $_->{code} ne 'group_hint' } @$meta;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Rutland.pm
```diff
@@ -40,6 +40,19 @@ sub open311_extra_data_include {
     ];
 }
 
+sub open311_contact_meta_override {
+    my ($self, $service, $contact, $meta) = @_;
+
+    my ($hint) = grep { $_->{code} eq 'hint' } @$meta;
+    my ($group_hint) = grep { $_->{code} eq 'group_hint' } @$meta;
```
dracos, line 47: Do you want to delete these rows from $meta, so that they're not also extra attributes that don't do anything?

dracos (changes requested):
The Open311 population code needs a test. The existing test is failing because the rutland.t file creates a body in 2482 (Bromley), not Rutland (2600). Change that and it fails less - it still fails on the group display because groups that only have one category are not shown as groups, the category is auto-promoted to top level and so the group text won't be seen, you'd need two categories in the group to test that.

The code itself looks fine, think it should do what you want :)

dracos (approved):

PR #3438, opened 2021-05-18
[Bromley] Set description when setting next/last.

dracos:
This way, it will still be set for a future-dated schedule, but not set if a scheduled fails to have next/last instances. [skip changelog]


struan (approved):
👍

PR #3442, opened 2021-05-21
Upgrade jQuery to 3.6.0.

dracos:
Upgrading to jQuery 2.2.4 (+3.5.0 security patch) was done in #3017 after I took this commit out so the change wasn't as big.

I think only a couple of changes here are actually needed (e.g. `complete` to `always`), the others are warnings given by jquery migrate but still work fine, so if any have been missed it shouldn't matter (hopefully haven't though). The :first/:last/etc selectors were deprecated in 3.4 but still working (currently scheduled to be removed in 4, I think). But e.g. `first()` is quicker than `:first` so no harm in doing these. `click()` etc were technically deprecated in 3.3 - https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/ - but the docs still don't say they are and there are frequent issues on the migrate repo saying why does it warn and they say "PR the docs" and have been since at least 2018, so who knows what's happening there. Again, no harm in switching now though.


dracos:
@zarino you already reviewed this last year, and the 2.2 part of that is live now, so this is just the 3 changes, which aren't anything much it seems.

zarino (approved):
Looks good to me.

I notice lots of changes inside `web/vendor/fancybox/jquery.fancybox-1.3.4.js`. Should we just be using the latest version ([3.5.7](http://fancyapps.com/fancybox/3/)) of `fancybox` instead?

dracos:
That version is only available under a non-commercial licence.

dracos:
I think on the old PR I did say someone should go and evaluate all the new JS packages out there and pick the smallest new one to switch to :)

zarino:
Yup, not a dealbreaker. Just thought I’d check we had a reason for not using the latest. Agree it’s something we can sort in a subsequent PR.

PR #3444, opened 2021-05-24
[Bromley] improve handling of expired subscriptions

struan:
Don't skip recently expired garden waste subs from the bin days list so
they can be renewed if people are a bit forgetful.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1032,7 +1032,8 @@ sub process_garden_renew : Private {
     my $payment = $c->cobrand->garden_waste_cost($total_bins);
     $data->{bin_count} = $total_bins;
 
-
+    my $current_bins = $c->cobrand->get_current_garden_bins;
+    $data->{new_bins} = $total_bins - $current_bins;
     if ( !$service || $c->cobrand->waste_sub_overdue( $service->{end_date} ) ) {
         $data->{category} = 'Garden Subscription';
         $data->{title} = 'Garden Subscription - New';
@@ -1041,10 +1042,6 @@ sub process_garden_renew : Private {
         $data->{category} = 'Garden Subscription';
         $data->{title} = 'Garden Subscription - Renew';
         $c->set_param('Subscription_Type', $c->stash->{garden_subs}->{Renew});
-
-        # only override the new bin count if we know the current bin number
-        my $current_bins = $c->cobrand->get_current_garden_bins;
-        $data->{new_bins} = $total_bins - $current_bins;
     }
 
     $c->set_param('payment', $payment);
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -594,6 +594,19 @@ sub waste_container_actions {
     };
 }
 
+sub waste_recently_expired {
+    my ($self, $schedule) = @_;
+
+    my ($y, $m, $d) = ( $schedule->{end_date} =~ /(\d+)-(\d+)-(\d+)/);
+    my $end = DateTime->new( year => $y, month => $m, day => $d)->set_time_zone(FixMyStreet->local_time_zone);
+    my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
+    if ( $now > $end ) {
```
dracos, line 603: Is the end date inclusive? (I assumed it was.) So this is true on the day a schedule expires, should it be?

dracos: Could perhaps waste_sub_overdue (which seems to have a fix for this in this commit), be given a day or week argument, then it returns true if more than that overdue (and so 0 by default?)

dracos (changes requested):

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -594,6 +594,14 @@ sub waste_container_actions {
     };
 }
 
+sub waste_recently_expired {
+    my ($self, $schedule) = @_;
+
+    my ($y, $m, $d) = ( $schedule->{end_date} =~ /(\d+)-(\d+)-(\d+)/);
+    my $end = DateTime->new( year => $y, month => $m, day => $d)->set_time_zone(FixMyStreet->local_time_zone);
+    return $self->waste_sub_overdue($end, weeks => 4);
```
dracos, line 602: waste_sub_overdue is given a string, not a DateTime, so can you drop waste_recently_expired entirely, and only call `waste_sub_overdue($schedules->{end_date}, weeks => 4)` or have I missed something?

struan: for some reason I was convinced that DateTime::Format::W3CDTF didn't parse that date but it turns out I was wrong :|

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -936,6 +937,12 @@ sub get_original_sub : Private {
         });
     }
 
+    if ( $c->stash->{report} ) {
+        $p = $p->search({
+            id => { '<>' => $c->stash->{report}->id }
```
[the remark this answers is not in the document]
struan: Hmmm, it looks like this could only ever happen if you are a staff user in which case you the payment method is moot so this is probably not necessary.

dracos (approved):
woo

PR #3446, opened 2021-05-24
[Oxfordshire] Replace contact form with a link to OCC's own form

chrismytton:
Oxfordshire have requested that the contact form be replaced with a link to their own form.

Fixes https://github.com/mysociety/societyworks/issues/2374

image


struan (approved):
✉️

dracos (changes requested):
You cannot do this (exactly like this, I mean; sorry, was just replying quickly in case it got merged :) ), it breaks reporting abusive reports which uses the same template. Shouldn't be much more work to factor out the form part of the base template to its own file and then replace that with the link.

chrismytton:
@dracos Oooh good catch, thanks! Will factor the form out as you say.

dracos (approved):

PR #3451, opened 2021-05-26
[Zürich] Add Bikeable.ch banner to front page

davea:
image

For https://github.com/mysociety/societyworks/issues/2351

[skip changelog]

### DIFF
--- a/templates/web/zurich/header.html
+++ b/templates/web/zurich/header.html
@@ -2,6 +2,8 @@
     SET html_att = ' lang="' _ lang_code _ '"';
     # For a right-to-left language, use the following line in your own header:
     # SET html_att = html_att _ ' dir="rtl"';
+
+    USE date;
 -%]
 <!doctype html>
 <!--[if IE 9]>   <html class="no-js ie9"[% html_att | safe %]><![endif]-->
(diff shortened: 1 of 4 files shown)


templates/web/zurich/header.html
```diff
@@ -86,6 +86,9 @@
 
                 <div class="content[% " $mainclass" | html IF mainclass %]" role="main">
                 [% IF NOT admin %]
+                  [% IF bodyclass.match('frontpage') %]
```
dracos, line 89: Do we want to add a and date <= 2021-07-12 style statement here so that it automatically disappears after the end date?

web/cobrands/zurich/_bikeable_banner.scss
```diff
@@ -0,0 +1,36 @@
...
+
+        padding: 1em;
+        width: 150px;
+        height: 150px;
+        border-radius: 150px;
+        top: 120px;
+        left: 6px;
+    }
+
+    @media screen and (min-width: 1025px) {
```
dracos, line 28: If being really pinickity, this changes at a point where the circle then goes off the edge of the screen for a little bit, which would be fixed if this min-width was a few pixels more, I think. But don't think it matters

dracos (commented):
Looks nice :) Couple of tiny suggestions I think might improve it, but fine.

templates/web/zurich/header.html
davea: That is a smart idea, will do.

PR #3453, opened 2021-05-28
[Noise] Prevent duplicate form submission.

dracos:
Store a random value in both session and form, and mandate that they
always match; on successful submission, delete the value from the
session so that any resubmission of the same form will error.
[skip changelog]


struan (approved):
👍

PR #3459, opened 2021-06-01
[Oxfordshire] Archive script marks comments sent for Open311 reports

davea:
This prevents a huge backlog of unsent updates building up that don’t actually need to be sent anywhere.

[skip changelog]


chrismytton (approved):
📤

PR #3462, opened 2021-06-02
[Peterborough] Don’t delete inactive Open311 categories

davea:
Peterborough have a number of Open311 categories marked as ‘inactive’ on the site
pending the launch of a new integration. For some reason these categories were
omitted from the service list from Open311 adapter for a short period, which
resulted in the inactive categories being deleted. When they subsequently
reappeared in the service list, they were reinstated as ‘confirmed’ categories
instead of going back into the ‘inactive’ state. This meant users could unexpectedly
raise reports in these new categories.

For https://github.com/mysociety/societyworks/issues/2391

[skip changelog]


struan (approved):
👍

PR #3464, opened 2021-06-03
[Bromley] handle waste events with no data

struan:
Default to an empty array if the event has no data to avoid errors.

[skip changelog]


struan:
Fixes https://github.com/mysociety/societyworks/issues/2394

davea (approved):

PR #3466, opened 2021-06-04
[Bromley] waste payment failed option for CSC staff

struan:
[skip changelog]

### DIFF
--- a/t/app/controller/waste.t
+++ b/t/app/controller/waste.t
@@ -1591,6 +1591,46 @@ FixMyStreet::override_config {
 
     };
 
+    subtest 'staff create new subscription - payment failed' => sub {
+        $mech->clear_emails_ok;
+        $mech->get_ok('/waste/12345/garden');
+        $mech->submit_form_ok({ form_number => 2 });
+        $mech->submit_form_ok({ with_fields => { existing => 'no' } });
+        $mech->content_like(qr#Total to pay now: £<span[^>]*>0.00#, "initial cost set to zero");
+        $mech->content_lacks('password', 'no password field');
+        $mech->submit_form_ok({ with_fields => {
+                current_bins => 0,
+                new_bins => 1,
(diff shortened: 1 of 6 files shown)


templates/email/default/waste/csc_payment_failed.html
```diff
@@ -0,0 +1,25 @@
...
+INCLUDE '_email_top.html';
+
+%]
+
+<th style="[% td_style %][% primary_column_style %]" id="primary_column">
+  [% start_padded_box | safe %]
+  <h1 style="[% h1_style %]">There was a problem with your payment</h1>
+  <p style="[% p_style %]">There was a problem taking payment for your [% report.title %]</p>
+
+  <p style="[% p_style %]">Please contact the Customer Service Centre to try again.</p>
```
davea, line 16: Might be nice to inform the user how they can do this, especially as it's sent from a no-reply email address.

t/app/controller/waste.t
```diff
@@ -1591,6 +1591,48 @@ FixMyStreet::override_config {
...
+                current_bins => 0,
+                new_bins => 1,
+                name => 'Test McTest',
+                email => 'test@example.net'
+        } });
+        $mech->content_contains('Test McTest');
+        $mech->content_contains('£20.00');
+        $mech->content_contains('1 bin');
+        # external redirects make Test::WWW::Mechanize unhappy so clone
+        # the mech for the redirect
```
davea, line 1611: Can this comment (and on L1696) be removed now as the clone doesn't seem to happen?

davea (approved):
Couple of small things but 👍

templates/email/default/waste/csc_payment_failed.html
struan: I need to get Bromley to confirm all the text in the emails so this is mostly a place holder.

PR #3469, opened 2021-06-04
[Peterborough] Final waste tweaks

davea:
A few more things that needed doing before go-live.

[skip changelog]


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -363,6 +363,7 @@ sub bin_services_for_address {
     my $schedules = $bartec->Features_Schedules_Get($property->{uprn});
     my %schedules = map { $_->{JobName} => $_ } @$schedules;
     my $open_requests = $self->open_service_requests_for_uprn($property->{uprn}, $bartec);
+    $self->{c}->stash->{open_service_requests} = $open_requests;
```
dracos, line 366: This (and disabling) is fine, but would be good for v similar Bromley code to change to match, or this changed to match that (it uses `service.enquiry_open_events`, looks like, and doesn't show them rather than showing them disabled - think showing them disabled for Bromley would be okay).

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -390,13 +390,14 @@ sub bin_services_for_address {
...
+            # is there already an open bin request for this container?
             request_open => @request_service_ids_open ? 1 : 0,
             # can this collection be reported as having been missed?
             # allowed if we're within 24 hours of the last collection
             report_allowed => DateTime->now < $last->add(hours => 24),
-            # is there already a missed collection report open for this container?
-            report_open => @report_service_ids_open ? 1 : 0,
+            # is there already a missed collection report open for this container
+            # (or a missed assisted collection for any container)?
+            report_open => ( @report_service_ids_open || $open_requests->{492} ) ? 1 : 0,
```
dracos, line 402: Do we want to create a reverse of container_service_ids somewhere and use that instead of a magic number.

perllib/FixMyStreet/Cobrand/Peterborough.pm
dracos: Also, umm, none of this appears to have any tests?

dracos (changes requested):
Couple of small comments, but more importantly this all appears to lack any tests either before or after.

PR #3471, opened 2021-06-07
[Bromley] Improve waste page navigation.

dracos:
Tweaks to navigation on waste pages to not include unnecessary links.
[skip changelog]


dracos:
There's also an uncommitted change on raven to the httpd conf to redirect most FMS URLs to the fix subdomain.

templates/web/bromley/header_logo.html
```diff
@@ -62,7 +62,11 @@
         <!-- Breadcrumb --><!-- googleoff:all -->
         <ul>
             <li class="first-child"><a href="https://www.bromley.gov.uk">Home</a></li>
+          [% IF bodyclass == 'waste' %]
```
chrismytton, line 65: Is the ` OR (problem AND problem.cobrand_data == 'waste')` bit below needed here as well?

chrismytton (approved):
One question, but otherwise looks good! :compass:

Edit: And the change on raven looks good as well!

templates/web/bromley/header_logo.html
dracos: You can't see this breadcrumb on a report page as it's a map page, so it doesn't matter, but it would keep it consistent IF wise, I guess, so might as well, yeah.

PR #3472, opened 2021-06-08
Send contact form email from fixmystreet.com email if dmarc

chrismytton:
Previously we were sending emails from the recipient's address if we detected dmarc on the email entered into the contact form. This caused issues if the other end also had dmarc, since it would detect the incoming emails as spoofing attempts and reject them.

To avoid these issues change the contact form to send from the do-not-reply email address if the sending email address domain uses dmarc.

Fixes https://github.com/mysociety/societyworks/issues/2344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@
         - Consistent protected field ordering.
     - Security:
         - Increase minimum password length to eight.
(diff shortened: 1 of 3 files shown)


dracos (approved):
Looks good :)

PR #3476, opened 2021-06-11
Only trigger one refresh going Back to list view.

dracos:
If going back from a report page to a list page, a marker refresh would
be triggered twice, once by the popstate, once by display_around. As
only one of those had use_page set, it would at best make two identical
requests, at worst return to page 1 of results whatever page it should
have been showing.

If we're already on a list page, trigger the refresh in popstate as
currently; if not, skip that trigger and rely on the refresh that will
be called in display_around.

An alternative solution would have been to change the refresh in
dispaly_around to use the same debounce as the popstate trigger, which
would fix the issue in that it would only be called once, but introduce
a delay waiting for the end of the debounce time.

Also set use_page when going back to initial state in case that wasn't
the first page of results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
         - Fix Gaze sometimes being called twice on /around. #3324
         - Improved alert signup for phone-only user. #3367
         - Fix moderation of update text.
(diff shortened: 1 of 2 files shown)


chrismytton (approved):
🔄

PR #3481, opened 2021-06-15
Bromley waste phase2 text updates

struan:


templates/web/base/waste/garden/cancel.html
```diff
@@ -1,6 +1,10 @@
 [% BLOCK 'content' %]
   <p>
-  This is the text to say you will not get your money back, will carry on to end of current season, etc.
+      Please be aware refunds or part refunds will not be given for early termination during any 12-month subscription period.
+  <p>
```
dracos, line 4: Should be `</p>`?

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -769,6 +769,13 @@ sub check_if_staff_can_pay : Private {
     return 1;
 }
 
+sub garden_terms : Path('garden_terms') : Args(0) {
```
dracos, line 772: Do we want to use an /about page for this to reuse that mechanism, bit simpler?

dracos (changes requested):
Couple of small suggestions

dracos (approved):

PR #3484, opened 2021-06-16
Show expired garden waste subs on the bin list page

struan:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -553,7 +553,7 @@ sub available_bin_services_for_address {
 
     my $services = {};
     for my $service ( @$result ) {
-        my $name = $self->service_name_override->{$service->{ServiceId}} || $service->{ServiceName};
+        my $name = service_name_override($service);
         $name =~ s/ /_/g;
         $services->{$name} = {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -813,13 +813,19 @@ sub _get_current_service_task {
...
+    my $service_name = service_name_override()->{$service->{ServiceId}} || $service->{ServiceName};
+    my $today = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
     my ($current, $last_date);
-    foreach ( @$servicetasks ) {
-        my $end = construct_bin_date($_->{EndDate});
-
-        next if $last_date && $end && $end < $last_date;
-        $last_date = $end;
-        $current = $_;
+    foreach my $task ( @$servicetasks ) {
```
dracos, line 819: Not for now, but feels like we could take the result of the GetServiceUnits API call and filter out blank/out-dated service tasks first off and then not have to call this function twice in `bin_services_for_address`. But it's only another loop through the same data.

dracos (changes requested):
One tiny change.
And assuming the servicetaskschedule end dates match the servicetask enddates :)

dracos (approved):

PR #3485, opened 2021-06-17
[Bromley] explicitly ask how many bins a customer wants

struan:
This means they can reduce the number of bins, and it makes it a bit
more explicit how many bins they want, rather than having to do sums.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden.pm
@@ -10,14 +10,16 @@ has_field is_staff => ( type => 'Hidden' );
 sub details_update_fields {
     my $form = shift;
     my $data = $form->saved_data;
+    my $c = $form->{c};
 
     my $existing = $data->{existing_number} || 0;
     $existing = 0 if $data->{existing} eq 'no';
-    my $cost = $existing == 0 ? 0 : $form->{c}->cobrand->garden_waste_cost($data->{existing_number});
+    my $bin_count = $c->get_param('bins_wanted') || $existing;
+    my $cost = $bin_count == 0 ? 0 : $form->{c}->cobrand->garden_waste_cost($bin_count);
     $form->{c}->stash->{payment} = $cost / 100;
     return {
         current_bins => { default => $existing },
-        new_bins => { default => 0 },
+        bins_wanted => { default => $bin_count },
     };
 }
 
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -122,13 +133,16 @@ has_field current_bins => (
...
-has_field new_bins => (
+has_field bins_wanted => (
     type => 'Integer',
-    label => 'Additional bins required (0-3)',
+    label => 'Total number of bins required (0-3)',
     tags => { number => 1 },
     required => 1,
     range_start => 0,
     range_end => 3,
+    tags => {
```
dracos, line 143: This is overriding the previous tags => number => 1 (whatever that is doing :) )

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -10,14 +10,25 @@ has_field is_staff => ( type => 'Hidden' );
...
+    $wanted ||= $existing;
+
+    my $cost = $bin_count == 0 ? 0 : $form->{c}->cobrand->garden_waste_cost($bin_count);
     $form->{c}->stash->{payment} = $cost / 100;
     return {
         current_bins => { default => $existing },
-        new_bins => { default => 0 },
+        bins_wanted => { default => $wanted },
     };
 }
```
dracos, line 33: Not sure I understand the logic here. If someone managed to submit with existing=no and existing_number>0, this would set it to 0. But if they'd set bins_wanted as well, it would have left existing_number alone, even though existing=no? And don't wanted/bin_count always end up the same value? Does this have the same outcome:
```
    my $existing = $data->{existing_number} || 0;
    $existing = 0 if $data->{existing} eq 'no';
    my $bin_count = $c->get_param('bins_wanted') || $existing;
    my $cost = $bin_count == 0 ? 0 : $form->{c}->cobrand->garden_waste_cost($bin_count);
    return {
        current_bins => { default => $existing },
        bins_wanted => { default => $bin_count },
    };
```

dracos (changes requested):
One bug (I think), and hopeful simplification of the bin logic if I've understood it right!

perllib/FixMyStreet/App/Form/Waste/Garden.pm
struan: yes, that is much better, thanks.

dracos (approved):
As long as you agree my logic works ;)

PR #3486, opened 2021-06-17
[Bromley] waste phase2 staff only

dracos:
Adds garden_disabled feature to disable it entirely, or garden_waste_staff_only to restrict to staff. Only in terms of what is displayed to the user. [skip changelog]

### DIFF
--- a/templates/web/base/waste/bin_days.html
+++ b/templates/web/base/waste/bin_days.html
@@ -14,7 +14,7 @@
 
 [% SET any_report_allowed = 0 ~%]
 [% SET any_request_allowed = 0 ~%]
-[% IF pending_subscription %]
+[% IF NOT waste_features.garden_disabled AND pending_subscription %]
     <p>You have a pending [% pending_subscription.title %]</p>
 [% END %]
 [% FOR unit IN service_data %]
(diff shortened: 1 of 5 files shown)


templates/web/base/waste/bin_days.html
```diff
@@ -96,7 +96,7 @@ <h3>More services</h3>
             [% IF any_request_allowed %]
               <li><a href="[% c.uri_for_action('waste/request', [ property.id ]) %]">Request a new container</a></li>
             [% END %]
-            [% IF !pending_subscription AND !services_available.Garden_Waste.is_active %]
+            [% IF NOT waste_features.garden_disabled AND NOT pending_subscription AND NOT services_available.Garden_Waste.is_active %]
```
struan, line 99: garden_disabled is never set here if there isn't a garden subscription so this will still show.

struan (changes requested):
one thing with the subscribe link

templates/web/base/waste/bin_days.html
dracos: Oh yes, good point

struan (approved):
👍

PR #3490, opened 2021-06-18
Bromley waste phase2 cancel staff emails

struan:
[skip changelog]


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -619,6 +619,10 @@ sub waste_container_actions {
     };
 }
 
+sub waste_staff_source {
+    return 3;
```
dracos, line 623: Note master has a _set_user_source function that does this and more (9 for Officer, 3 for CSC), that might be good to use here.

dracos (commented):
Fine, but there's a function on master that could be used here too

dracos (approved):

PR #3494, opened 2021-06-18
Add category restrictions to roles

chrismytton:
Adds category restrictions to roles, which should work in the same way as category restrictions for users.

This includes adding an admin interface for assigning categories to roles, as well as a couple of tweaks to the logic that checks user's categories to take a role's categories into account.

Fixes https://github.com/mysociety/societyworks/issues/2382

## Todo

- [x] Write documentation for this feature in the manual

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
@@ -55,7 +55,17 @@ sub item :PathPart('admin/roles') :Chained :CaptureArgs(1) {
 
 sub edit :PathPart('') :Chained('item') :Args(0) {
     my ($self, $c) = @_;
-    return $self->form($c, $c->stash->{obj});
+
+    my $role = $c->stash->{obj};
+
+    $c->stash->{body} = $role->body;
+    $c->forward('/admin/fetch_contacts');
+    $c->forward('/admin/stash_contacts_for_template', [
+        \@{$role->get_extra_metadata('categories') || []},
+    ]);
+    $c->forward('/report/stash_category_groups', [ $c->stash->{contacts}, { combine_multiple => 1 } ]);
+
+    return $self->form($c, $role);
 }
 
 sub form {
@@ -92,7 +102,18 @@ sub form {
     my $action = $role->in_storage ? 'edit' : 'add';
     my $form = FixMyStreet::App::Form::Role->new(%$opts);
     $c->stash(template => 'admin/roles/form.html', form => $form);
-    $form->process(item => $role, params => $c->req->params);
+
+    my %form_params;
+    if ($action eq 'edit') {
+        # Check which contacts should be used for this role's category restrictions.
+        my @live_contacts = $c->stash->{live_contacts}->all;
+        my @live_contact_ids = map { $_->id } @live_contacts;
+        my @new_contact_ids = grep { $c->get_param("contacts[$_]") } @live_contact_ids;
+
+        $form_params{categories} = \@new_contact_ids;
+    }
+
+    $form->process(item => $role, params => $c->req->params, %form_params);
     return unless $form->validated;
 
     $c->forward('/admin/log_edit', [ $role->id, 'role', $action ]);
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -360,6 +360,21 @@ sub should_skip_sending_update {
     return $update->user->from_body && !$update->text && !$private_comments;
 }
 
+sub munge_report_new_category_list {
+    my ($self, $category_options, $contacts) = @_;
+
+    my $user = $self->{c}->user;
+    if ($user && $user->belongs_to_body($self->body->id) && $user->get_extra_metadata('assigned_categories_only')) {
+        my %user_categories = map { $_ => 1} @{$user->categories};
+        my $non_bromley_or_assigned_category = sub {
+            $_->body_id != $self->body->id || $user_categories{$_->category}
+        };
+
+        @$category_options = grep &$non_bromley_or_assigned_category, @$category_options;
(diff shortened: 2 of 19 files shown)


perllib/FixMyStreet/App/Controller/Admin/Roles.pm
```diff
@@ -55,7 +55,22 @@ sub item :PathPart('admin/roles') :Chained :CaptureArgs(1) {
...
+    $c->forward('/admin/fetch_contacts');
+    $c->forward('/admin/stash_contacts_for_template', [
+        \@{$role->get_extra_metadata('categories') || []},
+    ]);
+    $c->forward('/report/stash_category_groups', [ $c->stash->{contacts}, { combine_multiple => 1 } ]);
+
+    my @live_contacts = $c->stash->{live_contacts}->all;
+    my @live_contact_ids = map { $_->id } @live_contacts;
+    my @new_contact_ids = grep { $c->get_param("contacts[$_]") } @live_contact_ids;
+    $c->stash->{new_contact_ids} = \@new_contact_ids;
```
struan, line 71: I don't totally get why this needs to perform this check given that the values in contacts[] comes for live_contacts in the first place, so maybe a comment to explain what's happening here would be nice.

struan (approved):
I did find the having to tick the assigned categories only box for the role categories to apply a bit non intuitive, but otherwise 👍

perllib/FixMyStreet/App/Controller/Admin/Roles.pm
chrismytton: Yeah having this code here was a bit confusing. I've moved it to where it's actually being used in 37106224520e77ed3f1d3e2abed7454f4be3b8c8 and added a comment, so hopefully that makes things a bit clearer.

chrismytton:
> I did find the having to tick the assigned categories only box for the role categories to apply a bit non intuitive

Yeah. I'm going to leave it like that for now and see if it turns out to actually be a problem in practice.

chrismytton:
Thinking about it I should probably add some documentation about this to the manual so people know how to use this feature.

chrismytton:
Bromley want to restrict which categories a user with a certain role can _create_ reports in, as well as view reports in. I've added an extra Bromley-specific commit which hides any categories that the user isn't assigned to. I've also added an extra commit to update the documentation.

@dracos would you mind giving this a review?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -481,6 +481,13 @@ sub munge_report_new_contacts {
     } else {
         @$categories = grep { grep { $_ ne 'Waste' } @{$_->groups} } @$categories;
     }
+
+    my $user = $self->{c}->user;
+    if ($user && $user->from_body($self->body) && $user->get_extra_metadata('assigned_categories_only')) {
+        my %user_categories = map { $_ => 1} @{$user->categories};
+        @$categories = grep { $user_categories{$_->category} } @$categories;
```
dracos, line 488: Does this need the additional Bromley subcategory stuff, or has that already been dealt with? (e.g. like elsewhere in this file that also looks at `$c->user->get_extra_metadata('subcategories')`).

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -481,6 +481,13 @@ sub munge_report_new_contacts {
     } else {
         @$categories = grep { grep { $_ ne 'Waste' } @{$_->groups} } @$categories;
     }
+
+    my $user = $self->{c}->user;
+    if ($user && $user->from_body($self->body) && $user->get_extra_metadata('assigned_categories_only')) {
```
dracos, line 486: from_body is a getter/setter, not a check; I think you want `belongs_to_body` (with the body ID, note).

dracos (changes requested):
Couple of issues is all

perllib/FixMyStreet/Cobrand/Bromley.pm
chrismytton: We agreed with Bromley that we were going to ignore the subcategory stuff for this piece of work because we're going to be turning them into "normal" subcategories as part of the next piece of work for them.

perllib/FixMyStreet/Cobrand/Bromley.pm
chrismytton: Gah I even looked at the definition of `belongs_to_body` to check it did the right thing, but obviously forgot to, you know, actually update the code 🤦‍♂️

dracos (approved):
🧻

chrismytton:
Bromley came back and said they were happy with how this is working except for this error message that appears:

image

After a bit of digging it turns out this is because the `munge_report_new_contacts` where I was removing contacts gets called quite early in the setup of the report/new action.

https://github.com/mysociety/fixmystreet/blob/98cc8512e06d8e3263b8b5572696c9ae6f04274a/perllib/FixMyStreet/App/Controller/Report/New.pm#L745

This means that `$bodies_to_list` doesn't get TfL added to it, because the TfL categories have all been removed.

https://github.com/mysociety/fixmystreet/blob/98cc8512e06d8e3263b8b5572696c9ae6f04274a/perllib/FixMyStreet/App/Controller/Report/New.pm#L776

Therefore TfL gets added to the `@missing_details_bodies` list here.

https://github.com/mysociety/fixmystreet/blob/98cc8512e06d8e3263b8b5572696c9ae6f04274a/perllib/FixMyStreet/App/Controller/Report/New.pm#L850

And that causes the error message in the screenshot above to show.

To work around this problem I've moved the code that munges the list of contacts into the `munge_report_new_category_list` hook, which is called later in the report/new action's setup.

@dracos would you mind checking over my fixup commit please?

chrismytton:
> I wondered if contacts (the second argument) should be reduced similarly, it's not totally clear what that then might be used for, but I don't think it matters if it isn't.

Have done this, just to be safe.

PR #3496, opened 2021-06-21
[Bromley] small changes to dd reconciliation scripts

struan:
fixes paths and module includes

[skip changelog]


dracos (changes requested):
Feels like it therefore needs a test? :)

dracos (approved):

PR #3500, opened 2021-06-23
[Bromley] ask for current container count on garden renewal page

struan:
This is largely to get round the problem of occasional inaccurate data
by asking and hence hopefully not having incorrect container requests.

[skip changelog]


dracos (approved):
Looks fine, if you want to tidy things up at some point, I note data->{bin_number} is doing the same job as data->{total_bins} in e.g. subscribe form, and that the subscribe_details/renew templates are now basically the same template.

PR #3498, opened 2021-06-21
[Zürich] Update brand colour

mySocietyClive:
For https://github.com/mysociety/societyworks/issues/2383

[skip changelog]

screenshot-zurich 127 0 0 1 nip io_3000-2021 06 21-15_59_46
screenshot-zurich 127 0 0 1 nip io_3000-2021 06 21-15_58_31

PR #3502, opened 2021-06-24
Issues/freshdesk/1106 zurich brand colour

mySocietyClive:
[skip changelog]

Includes fix for overlapping header on map pages.

### DIFF
--- a/web/cobrands/zurich/layout.scss
+++ b/web/cobrands/zurich/layout.scss
@@ -21,7 +21,7 @@ $mappage-header-height: 7em;
     }
 
     .header__logobar {
-        padding-left: calc((100vw - 60em) / 2);
+        padding: 1.25em 0 1.25em calc((100vw - 60em) / 2);
         margin-left: 0;
     }
 }
@@ -32,9 +32,15 @@ body.mappage {
         @include box-shadow(0 0 6px 1px #000);
         z-index: 2; // One more than #zurich-main-nav so it's on top
         .container {
-            margin: 0 1em; // White padding left/right
+            margin: 0;
+        }
+        .header__logobar {
+            // Aligns logo to left hand content on map pages
+            padding-left: 5px;
+            padding-right: 0;
         }
     }
+
     // Except on admin pages where there's an admin nav directly underneath it
     &.admin #site-header {
         @include box-shadow(none);
@@ -57,8 +63,8 @@ body.mappage #site-logo-text {
 
 #site-logo {
     width: 460px;
-    height: 120px;
-    background-position: center left;
+    height: 72px;
+    background-position: 10% 0;
     margin-left: 0;
     background-size: inherit;
 }
@@ -94,7 +100,7 @@ body.frontpage {
     }
 
     .table-cell {
-        background-position: 50% 117px;
+        background-position: 50% 112px;
         background-repeat: no-repeat;
         background-image: url(mapbg-1024.jpg);
         @media all and (min-width: 1025px) {


dracos:
This includes all the changes from #3498 that have already been merged. Could this please be rebased on top of the current code, so it contains only the changes (as you see, GitHub is complaining about the conflict here) - that will make it clearer what the new change is.

(commercial-staging is the active branch for the fixmystreet.com and councils staging site, not any other installation of the fixmystreet codebase. We don't open PRs on to the staging branches themselves, only to the branch it's eventually going to end up on or possibly a 'reviewed' branch. I've updated the base of the PR.)

mySocietyClive:
Ok thanks for clarifying the process - I've rebased now, please let me know if it's ok.

web/cobrands/zurich/layout.scss
```diff
@@ -30,9 +30,16 @@ body.mappage {
     #site-header {
         // The header on a map page needs a shadow too
         @include box-shadow(0 0 6px 1px #000);
-        z-index: 2; // One more than #zurich-main-nav so it's on top
+        z-index: auto; // So that #zurich-main-nav rises to the top
         .container {
-            margin: 0 1em; // White padding left/right
+            margin: 0; // White padding left/right
```
dracos, line 35: The comment here no longer matches

web/cobrands/zurich/layout.scss
```diff
@@ -30,9 +30,16 @@ body.mappage {
...
+        }
+        .header__logobar {
+            padding-left: 0;
+            margin-left: 0;
```
dracos, line 39: This line is a duplicate of higher up.

web/cobrands/zurich/layout.scss
```diff
@@ -30,9 +30,16 @@ body.mappage {
     #site-header {
         // The header on a map page needs a shadow too
         @include box-shadow(0 0 6px 1px #000);
-        z-index: 2; // One more than #zurich-main-nav so it's on top
+        z-index: auto; // So that #zurich-main-nav rises to the top
```
dracos, line 33: It feels like this change shouldn't be needed (also the comment isn't right, it's so it's not on top, presumably?). Why is the header bigger than the space available to it?

If we look at https://www.stadt-zuerich.ch/portal/de/index.html and compare against a map page here, the bar is basically the same size. If on desktop the header_logobar had padding top/bottom of 20px, the site-logo position was top left rather than centred vertically, and its height was 72px, I think that would look more like the main site, and you wouldn't need to change this line.
The only other change needed would be the 117px would change to 112px in the front page table cell to match the slight change there.

web/cobrands/zurich/layout.scss
```diff
@@ -30,9 +30,16 @@ body.mappage {
...
         .container {
-            margin: 0 1em; // White padding left/right
+            margin: 0; // White padding left/right
+        }
+        .header__logobar {
+            padding-left: 0;
+            margin-left: 0;
+        }
+        #site-logo {
+            background-position: 10% 50%;
```
dracos, line 42: This would then be 10% 0% to match my previous comment.

web/cobrands/zurich/layout.scss
```diff
@@ -56,7 +63,7 @@ body.mappage #site-logo-text {
 }
 
 #site-logo {
-    width: 460px;
+    width: 463px;
     height: 120px;
     background-position: center left;
```
dracos, line 68: I'd change the height here to 72px and drop the background-position line.

dracos (changes requested):
Thanks, bit easier to look at now :) This fixes the overhang issue, but it still feels a bit odd as the header on a map page is bigger than the space it has, and I couldn't see a reason why. I've made some suggestions to shrink it to the available space, let me know what you think.

dracos:
Also just to note you don't need to assign someone to a pull request for them to review it, reviewers are a whole separate thing to assignee (as you can see, when you're done with my comments, there's a "rereview" button :) ).

web/cobrands/zurich/layout.scss
dracos: As well as removing the duplicate line, you've removed the padding line which was bringing the logo on a map page over to the side to match the content, was that deliberate?

dracos (changes requested):
Looks better :) One issue regarding the logo position on map pages I thought I'd check.

web/cobrands/zurich/layout.scss
mySocietyClive: Just added padding back in - to align logo left hand content.

dracos (approved):
Looks good 👍

PR #3510, opened 2021-06-29
[CheshireEast] update highways notices and phone number

struan:
Remove various highways notices while there are no highways categories.

Also update the daytime emergency phone number.

Fixes mysociety/societyworks#2424

[skip changelog]


chrismytton (approved):
☎️

PR #3511, opened 2021-06-30
[Greenwich] turn off questionnaires

struan:
Re-opening reports is turned off so we should also turn off
questionnaires.

Fixes mysociety/societyworks#2425

[skip changelog]


chrismytton (approved):
Looks good!

As you mentioned in Slack it might be good to make it so turning off updates also turns off questionnaires. Do you think it's worth creating a ticket for that?

struan:
Good shout, ticket is https://github.com/mysociety/societyworks/issues/2427

PR #3514, opened 2021-07-01
add more options to process update file script

struan:
Adds the option to suppress alerts, dry run by default and also print out summary statistics

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/ProcessUpdateFile.pm
+++ b/perllib/FixMyStreet/Script/ProcessUpdateFile.pm
@@ -15,6 +15,8 @@ use Open311::GetServiceRequestUpdates;
 
 has verbose => ( is => 'ro', default => 0 );
 
+has commit => ( is => 'ro', default => 0 );
+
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Script/ProcessUpdateFile.pm
```diff
@@ -61,13 +61,17 @@ sub process {
 
     die "Problem loading body\n" unless $self->body;
 
+    print "Dry run, not adding comments. Use --commit to add\n";
```
chrismytton: Does this need to be displayed conditionally, based on the value of `$self->commit`?

chrismytton (approved):
🔔

perllib/FixMyStreet/Script/ProcessUpdateFile.pm
chrismytton: Oops ignore this, I wrote it before I saw the next commit!

PR #3515, opened 2021-07-01
[Bromley] increase maximum garden bins to 6

struan:
[skip changelog]


dracos (approved):
Is fine; I can't really see an easy way to make it a variable!

PR #3516, opened 2021-07-01
[Bromley] Remove trees asset layer

chrismytton:
Bromley have requested that we remove the trees asset layer for the time being.

Fixes https://github.com/mysociety/societyworks/issues/2426

[skip changelog]


struan (approved):
🌳

PR #3517, opened 2021-07-02
[Bromley] let staff with enough permission filter on waste reports

struan:
If a staff user can view private reports then let them see the Waste
categories in the all reports filter.

Fixes mysociety/societyworks#2430

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -441,8 +441,12 @@ sub munge_around_category_where {
 
 sub munge_reports_category_list {
     my ($self, $categories) = @_;
-    return if $self->{c}->action eq 'dashboard/heatmap';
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -441,8 +441,12 @@ sub munge_around_category_where {
 
 sub munge_reports_category_list {
     my ($self, $categories) = @_;
-    return if $self->{c}->action eq 'dashboard/heatmap';
-    @$categories = grep { grep { $_ ne 'Waste' } @{$_->groups} } @$categories;
+    my $c = $self->{c};
+    return if $c->action eq 'dashboard/heatmap';
+
+    unless ( $c->user && $c->user->from_body && $c->user->has_permission_to('report_mark_private', $self->body->id) ) {
```
dracos, line 447: I think user_exists is what we generally use for the first condition (probably doesn't matter though).

dracos (approved):

Issue #3308, opened 2021-02-04
JavaScript error when category extra data has a space in the key (2h)


chrismytton:
When a category has a piece of extra data which has a space in the key and that category triggers a display of the `#js-roads-responsibility` div, selecting that category and then selecting another category that doesn't display `#js-roads-responsibility` causes an error:

```
Syntax error, unrecognized expression: [name=public safety]
    at Function.m.error (jquery-1.7.2.min.js?1568106838:3)
    at Function.m.filter (jquery-1.7.2.min.js?1568106838:3)
    at m (jquery-1.7.2.min.js?1568106838:3)
    at Function.m [as find] (jquery-1.7.2.min.js?1568106838:3)
    at init.find (jquery-1.7.2.min.js?1568106838:3)
    at HTMLInputElement.<anonymous> (fixmystreet.js?1611765217:570)
    at Function.each (jquery-1.7.2.min.js?1568106838:2)
    at init.each (jquery-1.7.2.min.js?1568106838:2)
    at HTMLSelectElement.<anonymous> (fixmystreet.js?1611765217:569)
    at HTMLDivElement.dispatch (jquery-1.7.2.min.js?1568106838:3)
```

In this case the extra data key was `public safety`. The line that's causing the error is this one:

https://github.com/mysociety/fixmystreet/blob/08a1469538660e7dba37a54d191b7869aed40eb5/web/cobrands/fixmystreet/fixmystreet.js#L570

## Steps to reproduce the behavior

1. Choose a category that shows the `#js-roads-responsibility` message, e.g. one of the "Blocked drain" categories on B&NES.
2. Add a piece of extra data to that category which has a space in the key, e.g. `public safety`. (I had it set to a question, but not sure if that makes a difference.)
3. Start creating a new report and select the category, e.g. "Blocked drains"
4. Now select a different category which doesn't show the `#js-roads-responsibility` message and doesn't have any spaces in the extra data keys
5. You should see the error in the console

## Environment

 - OS: macOS
 - Browser: Google Chrome
 - Version: 88

## Video

Sound on 🔈 

https://user-images.githubusercontent.com/22996/106915224-b0913f00-66fd-11eb-9486-9a9bb9366de8.mov

PR #3518, opened 2021-07-02
Handle spaces in code names of extra data.

ludovic-tc:
An uncaught exception was being thrown by extra data which
included a space in the code (name), e.g. 'public safety',
when the user chose a report category.

+ **fixmystreet.js**: fix applied (single quotes around code name)
+ **fixture**: BathNES has new categories and extra data for testing
+ **bathnes.js**: additional Cypress test to trigger uncaught exception by clicking between road responsibility categories
+ **cypress.json**: fix to ensure Cypress tests run (is this appropriate?)

Fixes mysociety/fixmystreet#3308

Please check the following:

### DIFF
--- a/.cypress/cypress/integration/bathnes.js
+++ b/.cypress/cypress/integration/bathnes.js
@@ -39,3 +39,12 @@ it('uses the Curo Group housing layer correctly', function() {
     cy.pickCategory('Dog fouling');
     cy.contains('Maintained by Curo Group').should('be.visible');
 });
+
+it('handles code names with spaces without error', function() {
+    cy.visit('http://bathnes.localhost:3001/report/new?longitude=-2.359276&latitude=51.379009');
+    cy.contains('Bath & North East Somerset Council');
+    cy.get('input[value="Abandoned vehicles"]').click();
+    cy.get('input[value="Blocked drain"]').click();
+    cy.get('input[value="Abandoned vehicles"]').click();
+    cy.contains('Not maintained by Bath & North East Somerset Council');
+});
--- a/bin/fixmystreet.com/fixture
+++ b/bin/fixmystreet.com/fixture
@@ -116,7 +116,7 @@ if ($opt->test_fixtures) {
         { area_id => 2566, categories => [ 'General fly tipping', 'Fallen branch', 'Light Out', 'Light Dim', 'Fallen Tree', 'Damaged Tree', 'Pothole' ], name => 'Peterborough City Council' },
         { area_id => 2498, categories => [ 'Incorrect timetable', 'Glass broken', 'Mobile Crane Operation', 'Roadworks' ], name => 'TfL' },
         { area_id => 2237, categories => [ 'Flytipping', 'Roads', 'Parks' ], name => 'Oxfordshire County Council' },
-        { area_id => 2551, categories => [ 'Dog fouling' ], name => 'Bath and North East Somerset Council' }
+        { area_id => 2551, categories => [ 'Abandoned vehicles', 'Dog fouling', 'Blocked drain' ], name => 'Bath and North East Somerset Council' }
     ) {
         $bodies->{$_->{area_id}} = FixMyStreet::DB::Factory::Body->find_or_create($_);
         my $cats = join(', ', @{$_->{categories}});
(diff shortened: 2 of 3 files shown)


ludovic-tc:
@chrismytton NB: the test seems to break when it should work and vice versa! 🤦‍♂️

dracos:
The switch from blacklistHosts to blockHosts in the commit [1] breaks the tests on github because the version of cypress in the CI testing still uses blacklistHosts. So the tests running here were not blocking any external hosts and were calling those services because they were no longer blocked (you can see that e.g. in our tilma logs for requests from GitHub IPs at the time this occurred). If you drop that change, I'd hope the tests then pass (assuming that is the only issue).

[1] This shouldn't be in that commit anyway, because it is unrelated to the purpose of that commit.

.cypress/cypress/integration/bathnes.js
```diff
@@ -39,3 +39,12 @@ it('uses the Curo Group housing layer correctly', function() {
     cy.pickCategory('Dog fouling');
     cy.contains('Maintained by Curo Group').should('be.visible');
 });
+
+it('handles code names with spaces without error', function() {
+    cy.visit('http://bathnes.localhost:3001/report/new?longitude=-2.368929&latitude=51.386969');
```
chrismytton, line 44: The lat/long that you've used here will need to be added to `t/Mock/MapIt.pm`, which is the mocked version of MapIt that's used in the tests.

You can see the lat/long that's used by the test above here:

https://github.com/mysociety/fixmystreet/blob/6f35c67e76ef8efdbca3d2e6fdec17d1625cabbd/t/Mock/MapIt.pm#L68

So I think you can basically duplicate that line and change the lat/long to the ones you're using here.

.cypress/cypress.json
```diff
@@ -1,7 +1,7 @@
 {
   "baseUrl": "http://fixmystreet.localhost:3001",
   "projectId": "y8vvs1",
-  "blacklistHosts": ["gaze.mysociety.org", "*.openstreetmap.org", "portal.roadworks.org", "tilma.mysociety.org", "tilma.staging.mysociety.org", "isharemaps.bathnes.gov.uk", "consent.cookiebot.com", "assets.adobedtm.com", "www.googletagmanager.com", "www.googleadservices.com", "*.virtualearth.net"],
+  "blockHosts": ["gaze.mysociety.org", "*.openstreetmap.org", "portal.roadworks.org", "tilma.mysociety.org", "tilma.staging.mysociety.org", "isharemaps.bathnes.gov.uk", "consent.cookiebot.com", "assets.adobedtm.com", "www.googletagmanager.com", "www.googleadservices.com", "*.virtualearth.net"],
```
chrismytton, line 4: As Matthew mentioned in another comment, you shouldn't need to change this if you're running cypress 3.x. As I mentioned in Slack I think you can get the appropriate versions installed with this command:

```
npm install -g cypress@3.8.3 @cypress/code-coverage@3.9.4 codecov
```

chrismytton (changes requested):
I think this is nearly there! Just a couple of comments.

.cypress/cypress/integration/bathnes.js
chrismytton: Or alternatively you could re-use the same lat/long as the Curo test, think that would also work.

.cypress/cypress/integration/bathnes.js
```diff
@@ -39,3 +39,12 @@ it('uses the Curo Group housing layer correctly', function() {
     cy.pickCategory('Dog fouling');
     cy.contains('Maintained by Curo Group').should('be.visible');
 });
+
+it('handles code names with spaces without error', function() {
+    cy.visit('http://bathnes.localhost:3001/report/new?longitude=-2.3659276latitude=51.379009');
```
chrismytton, line 44: There's a missing `&` between the two parameters in the URL. I think if you add that the tests should pass 🤞 

```suggestion
    cy.visit('http://bathnes.localhost:3001/report/new?longitude=-2.3659276&latitude=51.379009');
```

chrismytton (approved):
👾 

Let's have a chat about getting this deployed to production this afternoon 🚀

PR #3520, opened 2021-07-05
[Bromley] only process one row for a dd record

struan:
Avoid problems if someone has tried to create a direct debit payment and
failed, and then succeeded by hiding any matching records that were not
confirmed and not matched.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -1607,6 +1607,14 @@ sub waste_reconcile_direct_debits {
         } elsif ( $p ) {
             my $service = $self->waste_get_current_garden_sub( $p->get_extra_field_value('property_id') );
             unless ($service) {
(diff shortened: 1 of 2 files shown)


dracos (commented):
The size of this function is a bit unwieldy! I understand the first change, wanting to hide other unconfirmed records (which is the commit message), but the second change I don't quite follow - if there's a cancellation at the DD end, no existing service, and there's ever been a cancellation report, don't provide a warning? What if there's a cancellation, a restart, then a cancellation, does that matter?

struan:
There should only ever be a hidden report if someone has cancelled on the front end, we've called the API to cancel the DD and then it's been cancelled in Echo in the interim. In which case the DD has been cancelled and the service has been cancelled and we should be happy that nothing has gone wrong. I guess I could limit that query by date to the last 5 days which is a realistic window for that to happen in.

dracos (approved):

PR #3521, opened 2021-07-08
[Bromley] cope with direct debit renewals that started as a renewal

struan:
It's possible for a user to create a direct debit that starts as a
renewal if the first time they use the system they renew an existing
subscription using direct debit. Previously we'd assumed that there had
to be a New subscription as the first direct debit interaction so this
resolves that.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -1412,13 +1412,21 @@ sub waste_reconcile_direct_debits {
             next unless $payment->{Status} eq 'Paid';
             $rs = $rs->search({ category => 'Garden Subscription' });
             my $p;
+            # loop over all matching records and pick the most recent new sub or renewal
+            # record. This is where we get the details of the renewal from. There should
+            # always be one of these for an automatic DD renewal. If there isn't then
+            # something has gone wrong and we need to error.
             while ( my $cur = $rs->next ) {
+                # only confirmed records are valid.
+                next unless FixMyStreet::DB::Result::Problem->visible_states()->{$cur->state};
                 my $sub_type = $cur->get_extra_field_value('Subscription_Type');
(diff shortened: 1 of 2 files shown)


dracos (commented):
Won't this then create another waste report, that's a duplicate of the one that we've just checked and already exists? What is it doing that for?

This does feel like the sort of function where a flow diagram or something to help explain how it works in overview would be useful to have and to make sure we try and cover all these cases, as otherwise I foresee yet more arising.

struan:
I don't understand why you think it will create a duplicate report.

There are two possible scenarios:

1: Someone has signed for a direct debit subscription which created a report with a Subscription_Type of New. This is matched as the original subscription by the first bit of the if and we use that as the basis of the subscription.

2: Someone has an existing subscription (not created in FixMyStreet) and when they renewed they set up a direct debit. This is the bit that's caught by the second half. If we've already processed this DD then there will be a renew record with the same dd date and we skip the DD payment entirely.

At the end of both there's a matching record with the right dd date which means we'll never process this DD payment again.

I do agree about the documentation.

dracos:
> I don't understand why you think it will create a duplicate report.

If `$cur` is a report, category Garden Subscription, Subscription_Type Renew, DD date does not match (or not present), state is not "unconfirmed", and there is an existing service, then it gets to the `_duplicate_waste_report` line and creates another report, category Garden Subscription, Subscription_Type Renew, with DD date now equal to $date. My assumption was `$cur` was the report created by the filling in of the renewal form, yes? So why create another report instead of updating `$cur` - what is `$cur` for now and what will happen to it? I am probably just confused but it is not clear to me what is going on here.

struan:
There is no report from filling in the form. This is for the second and subsequent direct debit payments which happen automatically so there's no interaction with the site by the user. $cur is just the loop variable in this case.

dracos (approved):
Typo in the code comment is all

PR #3522, opened 2021-07-08
[Bromley] hidden direct debit subs should not prevent renewal

struan:
This allows us to hide failed direct debit subs and not have them effect
renewal by other means, or by trying again.

Fixes mysociety/societyworks#2441

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -989,7 +989,8 @@ sub get_original_sub : Private {
     my $p = $c->model('DB::Problem')->search({
         category => 'Garden Subscription',
         title => 'Garden Subscription - New',
-        extra => { like => '%property_id,T5:value,I_:'. $c->stash->{property}{id} . '%' }
+        extra => { like => '%property_id,T5:value,I_:'. $c->stash->{property}{id} . '%' },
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -989,7 +989,8 @@ sub get_original_sub : Private {
     my $p = $c->model('DB::Problem')->search({
         category => 'Garden Subscription',
         title => 'Garden Subscription - New',
-        extra => { like => '%property_id,T5:value,I_:'. $c->stash->{property}{id} . '%' }
+        extra => { like => '%property_id,T5:value,I_:'. $c->stash->{property}{id} . '%' },
+        state => { '!=' => 'hidden' },
```
dracos, line 993: Should this be visible_states, or do you sometimes need to fetch unconfirmed?

dracos (commented):
One qn

perllib/FixMyStreet/App/Controller/Waste.pm
struan: It needs to fetch unconfirmed as the reports are unconfirmed while we wait for the DD to process.

dracos (approved):

PR #3525, opened 2021-07-09
Remove revalidation of verified landline number

### LINKED ISSUE #3295: Staff editing verified phone numbers (0.5d)
Staff users can, when reporting as another user, create a verified phone account with a landline number. Editing that later in the admin, it currently won't let you edit anything because verified numbers have to be mobile. The admin should either let staff put whatever they want as a verified number, or at least, if you haven't edited an existing user's phone number, it shouldn't be erroring if you try and edit something else for that user.

MorayMySoc:
Fixes #3295 

PROBLEM: Admin can create a user and put in a landline phone number and verify it if reporting on behalf of a user. But phone numbers are expected to be mobile numbers (for confirming number by SMS). Editing a user created by an admin in this way leads to the phone number being revalidated and the edit is rejected, whatever it is, if the number is a landline.

SOLUTION: When submitting an 'edit user' form, we will no longer revalidate the phone number if the phone number is the same as the original. If the phone number is changed, revalidation will take place.

TESTS: Test if changing the user name on a user with a verified landline succeeds. Test if changing the phone number from one landline to another fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
         - Fix moderation of update text.
         - Only trigger one refresh going Back to list view. #3476
         - Fix checked order of updates in dashboard export.
+        - Fix unable to edit user with verified landline #3295
     - Admin improvements:
         - Enable per-category hint customisation.
         - Move ban/unban buttons to user edit admin page.
--- a/t/app/controller/admin/users.t
+++ b/t/app/controller/admin/users.t
@@ -550,6 +550,27 @@ FixMyStreet::override_config {
         $mech->content_contains( 'Updated!' );
     };
 
+    subtest "Test edit user name with phone number unchanged who has verified landline" => sub {
+        my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User', phone => '01184960017', phone_verified => 1 );
+	$mech->get_ok( '/admin/users/' . $existing_user->id );
+	$mech->submit_form_ok( { with_fields => {
+            name => 'Existing E. User',
+        } } );
+	$mech->content_contains( 'Updated!' );
+    };
+
+    subtest "Test edit user change phone number who has verified landline" => sub {
+        my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User', phone => '01184960017', phone_verified => 1 );
+	$mech->get_ok( '/admin/users/' . $existing_user->id );
+	$mech->submit_form_ok( { with_fields => {
+            phone => '01184960018',
+	} } );
+	$mech->content_lacks( 'Updated!' );
+	$mech->content_contains( 'Please check your phone number is correct' );
+    };
+
+
+
     subtest "Test changing user to an existing one" => sub {
         my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User');
         $mech->create_problems_for_body(2, 2514, 'Title', { user => $existing_user });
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
         - Fix moderation of update text.
         - Only trigger one refresh going Back to list view. #3476
         - Fix checked order of updates in dashboard export.
+	- Fix unable to edit user with verified landline #3295
```
struan, line 28: The spacing isn't quite correct here.

t/app/controller/admin/users.t
```diff
@@ -550,6 +550,26 @@ FixMyStreet::override_config {
         $mech->content_contains( 'Updated!' );
     };
 
+    subtest "Test edit user name who has verified landline" => sub {
+        my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User', phone => '01184960017', phone_verified => 1 );
+	$mech->get_ok( '/admin/users/' . $existing_user->id );
+	$mech->submit_form_ok( { with_fields => {
+            name => 'Existing E. User',
+				 } } );
```
struan, line 558: these brackets would be better aligned with `$mech`

t/app/controller/admin/users.t
```diff
@@ -550,6 +550,26 @@ FixMyStreet::override_config {
...
+	$mech->content_contains( 'Updated!' );
+    };
+
+    subtest "Test edit user number who has verified landline" => sub {
+        my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User', phone => '01184960017', phone_verified => 1 );
+	$mech->get_ok( '/admin/users/' . $existing_user->id );
+	$mech->submit_form_ok( { with_fields => {
+            phone => '01184960018',
+	} } );
+	$mech->content_lacks( 'Updated!' );
```
struan, line 568: Would be nice to check for the expected error message here too.

t/app/controller/admin/users.t
```diff
@@ -550,6 +550,26 @@ FixMyStreet::override_config {
         $mech->content_contains( 'Updated!' );
     };
 
+    subtest "Test edit user name who has verified landline" => sub {
```
struan, line 553: might be good to add in a "with phone number unchanged" to the description here to make it a bit clearer what's being tested

t/app/controller/admin/users.t
```diff
@@ -550,6 +550,26 @@ FixMyStreet::override_config {
...
+        my $existing_user = $mech->create_user_ok('existing@example.com', name => 'Existing User', phone => '01184960017', phone_verified => 1 );
+	$mech->get_ok( '/admin/users/' . $existing_user->id );
+	$mech->submit_form_ok( { with_fields => {
+            name => 'Existing E. User',
+				 } } );
+	$mech->content_contains( 'Updated!' );
+    };
+
+    subtest "Test edit user number who has verified landline" => sub {
```
struan, line 562: likewise, the description here could be expanded/changed a bit to be clearer what's being tested.

struan (changes requested):
This looks good. Some small things but pretty much all cosmetic.

Also, be good to squash this all down into a single commit.

CHANGELOG.md
MorayMySoc: Done

t/app/controller/admin/users.t
MorayMySoc: Done

t/app/controller/admin/users.t
MorayMySoc: Done

t/app/controller/admin/users.t
MorayMySoc: Done

t/app/controller/admin/users.t
MorayMySoc: Done

struan (approved):
This looks good. Only needs to be squashed down to a single commit before merging.

struan (approved):

PR #3527, opened 2021-07-09
[Highways England] Add cut off date for problems

chrismytton:
Highways England have requested that reports made before 9th November
aren't visible on their cobrand.

Fixes https://github.com/mysociety/societyworks/issues/2445

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
+++ b/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
@@ -23,7 +23,7 @@ sub body {
 }
 
 # Copying of functions from UKCouncils that are needed here also - factor out to a role of some sort?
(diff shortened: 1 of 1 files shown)


dracos (commented):
Fine in principle (easier than trying to get the "Show older" tickbox working on Reports as well as Around I assume); is this something they are going to ask to change regularly? Should it be a value read from a configuration feature to make it easier to change in future?

chrismytton:
Thanks. Have asked them how often they think it's going to change, so will change this accordingly.

chrismytton:
> is this something they are going to ask to change regularly? Should it be a value read from a configuration feature to make it easier to change in future?

Not clear how regularly this is going to change, if at all. Going to leave it as-is for now, can move to configuration if/when they ask for it to be changed again.

PR #3533, opened 2021-07-14
[Bromley] prevent garden sub modification during renewal window

struan:
At his point the user should renew and change the number of bins while
renewing. This avoids complications with modifications and then
renewals.

[skpi changelog]


dracos (approved):
There is a typo in the commit message is only thing :)

PR #3534, opened 2021-07-14
Stop sending UK Council questionaires if users can't update

MorayMySoc:
Don't send questionnaires for closed reports if cobrand prevents re-opening closed reports. [1d] https://github.com/mysociety/societyworks/issues/2427

* There is a setting to not send questionnaires but this can be forgotten.
* If UK Council doesn't allow updates by users, don't send questionnaires.
* Adds method to test if users allowed to edit.
* Skips sending questionnaire if method returns true
* Adds tests
* Updates changelog

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UKCouncils.pm
+++ b/perllib/FixMyStreet/Cobrand/UKCouncils.pm
@@ -501,6 +501,21 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+# Report if cobrand denies updates by user
+sub deny_updates_by_user {
+    my ($self, $row) = @_;
+    my $cfg = $self->feature('updates_allowed') || '';
+    if ($cfg eq 'none' || $cfg eq 'staff') {
+        return 1;
+    } elsif ($cfg eq 'reporter-open' && !$row->is_open) {
+        return 1;
+    } elsif ($cfg eq 'open' && !$row->is_open) {
+        return 1;
+    } else {
+        return;
+    }
+};
+
 sub extra_contact_validation {
     my $self = shift;
     my $c = shift;
--- a/t/app/model/questionnaire.t
+++ b/t/app/model/questionnaire.t
@@ -13,7 +13,6 @@ use FixMyStreet;
 use FixMyStreet::TestMech;
 
 my $user = FixMyStreet::DB->resultset('User')->find_or_create( { email => 'test@example.com' } );
-
 my $problem = FixMyStreet::DB->resultset('Problem')->create(
     {
         postcode     => 'EH99 1SP',
@@ -105,11 +104,9 @@ for my $test (
         $problem->state( $test->{state} );
         $problem->send_questionnaire( 1 );
         $problem->update;
-
         $problem->questionnaires->delete;
 
         $mech->email_count_is(0);
-
         FixMyStreet::DB->resultset('Questionnaire')
           ->send_questionnaires( { site => 'fixmystreet' } );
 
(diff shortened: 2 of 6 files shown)


t/cobrand/fixmystreet.t
```diff
@@ -208,4 +208,59 @@ FixMyStreet::override_config {
...
+FixMyStreet::override_config {
+    COBRAND_FEATURES => {
+        updates_allowed => { bexley => undef},
+    },
+}, sub {
+    subtest 'Cobrand allows updates by user' => sub {
+        my $cobrand = FixMyStreet::Cobrand::Bexley->new;
+	    ok(!$cobrand->deny_updates_by_user, 'Report updates allowed with updates_allowed set to undef');
+    };
+};
```
struan, line 264: I think these could all be covered by a loop rather than individual tests.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -501,6 +501,14 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+# Report if cobrand denies updates by user
+sub deny_updates_by_user {
+    my $self = shift;
+    my $cfg = $self->feature('updates_allowed') || '';
+    if ($cfg eq 'none' || $cfg eq 'staff') {
+        return 1;
```
struan, line 509: I should have mentioned this but we probably need to handle when updates_allowed is set to 'open' which means they can leave an update if the report is open but not if it's fixed or closed.

There's also a reporter-open and reporter settings, the first of which is probably relevant.

t/app/model/questionnaire.t
```diff
@@ -144,4 +141,61 @@ for my $test (
     };
 }
 
+for my $test (
+    {
+        state => 'fixed',
+        send_email => 1,
+        user => 'user',
```
struan, line 148: the setting for this is 'reporter' rather than user.

struan (changes requested):
There's a few cases which I think I missed explaining to handle, but this looks good otherwise.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -501,6 +501,21 @@ sub updates_disallowed {
     return $self->next::method(@_);
 }
 
+# Report if cobrand denies updates by user
+sub deny_updates_by_user {
+    my ($self, $is_open) = @_;
```
struan, line 506: I wonder if passing in the row here and then doing the is_open check might be a bit more future proof

struan (changes requested):
sorry, one tiny change I should have thought of earlier.

t/cobrand/fixmystreet.t
```diff
@@ -208,4 +208,120 @@ FixMyStreet::override_config {
...
+        my $problem = FixMyStreet::DB->resultset('Problem')->create(
+        {
+        postcode     => 'EH99 1SP',
+        latitude     => 1,
+        longitude    => 1,
+        areas        => 1,
+        title        => 'to be sent',
+        detail       => 'detail',
+        used_map     => 1,
+        user_id      => 1,
```
struan, line 252: This is causing the tests to fail as it needs a user. There's a `$mech->create_user_ok` method you can use to set one up and then use the id of the created user here.

struan: Although actually it's probably easier to use `$mech->create_problems_for_body` to create the problem as it'll do most of the work.

t/cobrand/fixmystreet.t
```diff
@@ -208,4 +208,120 @@ FixMyStreet::override_config {
...
+        problem_state => 'confirmed',
+    }
+) {
+    subtest 'Cobrand set to allow updates' => sub {
+        FixMyStreet::override_config {
+            COBRAND_FEATURES => {
+            updates_allowed => { birmingham => $test->{update_permission} },
+            },
+    }, sub {
+        my $problem = FixMyStreet::DB->resultset('Problem')->create(
```
struan, line 301: probably only need one problem for both these tests to save on repetition.

struan (changes requested):
Code looks fine but few test issues.

t/cobrand/fixmystreet.t
struan: It probably works if you run them in isolation as it only creates the one user which will have Id 1 but when run as part of the suite it cleans out the database between tests which means the auto increment takes effect and the user created at the top of the script is no longer Id 1.

struan (approved):

MorayMySoc:
Merged and deployed

PR #3535, opened 2021-07-15
Cheshire east inspector tool upgrade

ludovic-tc:
- [x] 1.  New "assign reports" permission set up in admin. 
- [x] 2.  Update report screen to show who a report is assigned to.  
- [x] 3.  Update inspector sidebar with a dropdown to allow reports to be assigned to staff. 
- [x] 4.  Update report list to show who a report is assigned to. 
- [x] 5.  Update report list to allow bulk assignment of reports. 
- [x] 6.  (Tests.) 

Depends on https://github.com/mysociety/fixmystreet/pull/3562
Fixes https://github.com/mysociety/societyworks/issues/2455

_Please check the following:_
 
- [x] _Whether this PR should include changes to any documentation, or the FAQ;_
- [x] _Is new functionality tested? CodeCov will warn you about the diff coverage, but won’t complain about e.g. new files;_
- [ ] _Are the changes tested for accessibility?_
- [x] _Have you updated the changelog? If this is not necessary, put square brackets around this: skip changelog_

### DIFF
--- a/perllib/FixMyStreet/App/Controller/My.pm
+++ b/perllib/FixMyStreet/App/Controller/My.pm
@@ -270,6 +270,41 @@ sub shortlist_multiple : Path('planned/change_multiple') {
     $c->res->body(encode_json({ outcome => 'add' }));
 }
 
+sub bulk_assign : Path('planned/bulk_assign') {
+    my ($self, $c) = @_;
+    $c->forward('/auth/check_csrf_token');
+
+    # check here
+    $c->detach('/page_error_403_access_denied', [])
+        unless  ($c->user->has_body_permission_to('assign_report_to_user'));
+
+    my @bulk_reports = $c->get_param_list('bulk-assign-reports');
+    my $assignee = $c->get_param('inspector');
+
+    if (@bulk_reports) {
+        if ($assignee eq 'unassigned') {
+            # take off shortlist
+            my @problems = $c->model('DB::Problem')->search({ id => { -in => [ @bulk_reports ]} });
+            foreach my $problem (@problems) {
+                $problem->user->remove_from_planned_reports($problem);
+            }
+        } else {
+            # add to shortlist
+            my $inspector = $c->model('DB::User')->find({ id => $assignee });
+            # ... if actually an inspector:
+            if ($inspector->has_body_permission_to('report_inspect')) {
+                foreach my $report (@bulk_reports) {
+                    $c->forward( '/report/load_problem_or_display_error', [ $report ] ); 
+                    $inspector->add_to_planned_reports($c->stash->{problem});
+                }
+            }
+        }
+    }
+
+    $c->stash->{body} = $c->user->from_body;
+    $c->detach('/reports/redirect_body');
+}
+
 sub by_shortlisted {
     my $a_order = $a->get_extra_metadata('order') || 0;
     my $b_order = $b->get_extra_metadata('order') || 0;
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -521,6 +521,15 @@ sub inspect : Private {
             }
         }
 
+        # set assignment
+        my $assigned = ($c->get_param('assignment'));
+        if ($assigned && $assigned eq 'unassigned') {
+            # take off shortlist
+            $problem->user->remove_from_planned_reports($problem);
+        } elsif ($assigned) {
+            my $assignee = $c->model('DB::User')->find({ id => $assigned });
+            $assignee->add_to_planned_reports($problem);
+        }
         $problem->non_public($c->get_param('non_public') ? 1 : 0);
         if ($problem->non_public) {
             $problem->get_photoset->delete_cached(plus_updates => 1);
--- a/t/app/controller/report_inspect.t
+++ b/t/app/controller/report_inspect.t
@@ -660,6 +660,28 @@ FixMyStreet::override_config {
         $mech->content_contains('Shortlisted by Inspector Ian');
     };
 
+    subtest "reports list shows assignees' names" => sub {
+        $mech->get_ok("/reports");
+
+        use HTML::Selector::Element qw(find);
+        my $root = HTML::TreeBuilder->new_from_content($mech->content());
+
+        $mech->content_contains('unassigned');
+        my @assigned_to = $root->find("li#report-$report_id div.assigned-to span.assignee")->content_list;
+        like($assigned_to[0], qr/Inspector Ian/, "report $report_id assigned to Ian");
+
+        my $toggle_shortlist = sub {
+            $mech->form_id("add_remove_shortlist_$report_id");
+            $mech->click();
+            $mech->get_ok("/reports");
+            $root = HTML::TreeBuilder->new_from_content($mech->content());
+            @assigned_to = $root->find("li#report-$report_id div.assigned-to span.assignee")->content_list;
+        };
+        $toggle_shortlist->();
+        like($assigned_to[0], qr/Body User/, 'assignment by shortlist-add button still works' );
+        $toggle_shortlist->();
+        like($assigned_to[0], qr/unassigned/, 'unassignment by shortlist-remove button still works' );
+    };
     $user->user_body_permissions->delete;
 };
 
--- a/t/app/controller/report_inspect.t
+++ b/t/app/controller/report_inspect.t
@@ -635,6 +635,34 @@ FixMyStreet::override_config {
 
 };
 
+FixMyStreet::override_config {
+    MAPIT_URL => 'http://mapit.uk/',
+    ALLOWED_COBRANDS => 'oxfordshire',
+}, sub {
+    my $ian = $mech->create_user_ok('inspector@example.com', name => 'Inspector Ian', from_body => $oxon);
+    $user->user_body_permissions->create({ body => $oxon, permission_type => 'assign_report_to_user' });
+    $user->user_body_permissions->create({ body => $oxon, permission_type => 'planned_reports' });
+    $user->update;
+    $ian->user_body_permissions->create({ body => $oxon, permission_type => 'report_inspect' });
+    $ian->user_body_permissions->create({ body => $oxon, permission_type => 'planned_reports' });
+    $ian->update;
+
+    subtest "assign report by dropdown in report page" => sub {
+        $mech->get_ok("/report/$report_id");
+        $mech->content_contains('Assign to:');
+        $mech->content_contains('<select class="form-control" name="assignment" id="assignment">');
+
+        $mech->submit_form_ok({ button => 'save', with_fields => { include_update => 0, assignment => 'unassigned' } });
+        $mech->get_ok("/report/$report_id");
+        $mech->content_lacks('Shortlisted by');
+
+        $mech->submit_form_ok({ button => 'save', with_fields => { include_update => 0, assignment => $ian->id } });
+        $mech->content_contains('Shortlisted by Inspector Ian');
+    };
+
+    $user->user_body_permissions->delete;
+};
+
 foreach my $test (
     { cobrand => 'fixmystreet', limited => 0, desc => 'detailed_information has no max length' },
     { cobrand => 'oxfordshire', limited => 1, desc => 'detailed_information has max length'  },
--- a/templates/web/base/admin/_report-assignment.html
+++ b/templates/web/base/admin/_report-assignment.html
@@ -1,0 +1,11 @@
+[%  IF c.user.has_permission_to('report_inspect', problem.bodies_str_ids ) OR c.user.has_permission_to('assign_report_to_user', problem.bodies_str_ids) %]
+    [%# see report's assigned user %]
+    <span>[% loc('Assigned to:') %]</span>
+    <span class="assignee">
+        [% IF problem.shortlisted_user %]
+            [% problem.shortlisted_user.name OR problem.shortlisted_user.username %]
+        [% ELSE %]
+            [% loc('(unassigned)') %]
+        [% END %]
+    </span>
+[% END %]
--- a/templates/web/base/admin/problem_row.html
+++ b/templates/web/base/admin/problem_row.html
@@ -12,8 +12,8 @@
         <a href="[% uri %]" class="admin-offsite-link">[% problem.id %]</a>
         [%- ELSE %]
         [%- problem.id %]
-        [%- END -%]</td> 
-        <td>[% PROCESS value_or_nbsp value=problem.title %]</td> 
+        [%- END -%]</td>
+        <td>[% PROCESS value_or_nbsp value=problem.title %]</td>
         <td>
             [% PROCESS value_or_nbsp value=problem.name %]
             <br>[% PROCESS value_or_nbsp value=problem.user.email %]
@@ -29,6 +29,13 @@
                 [%- PROCESS value_or_nbsp value=problem.bodies_str -%]
             [%- END -%]
             <br>[% problem.cobrand %]<br>[% problem.cobrand_data | html %]
+            <br>
+            [% IF c.user.has_body_permission_to('assign_report_to_user') %] [%# see report's assigned user %]
+                <span>[% loc('Assigned to') %]: </span>
+                <span>
+                    [%- problem.shortlisted_user.name OR problem.shortlisted_user.username OR loc('unassigned') -%]
+                </span>
+            [%- END -%]
         </td>
         <td>[% prettify_state(problem.state, 1) %]<br><small>
             [% loc('Created') %]:&nbsp;[% PROCESS format_time time=problem.created %]
--- a/templates/web/base/report/_item.html
+++ b/templates/web/base/report/_item.html
@@ -119,7 +119,7 @@ END;
     [%# The around page list is already contained within the new report form %]
     [% item_action.replace('("shortlist-[^"]*)', '$1-' _ problem.id) | safe %]
   [% ELSE ~%]
-    <form method="post" action="/my/planned/change">
+    <form method="post" id="add_remove_shortlist_[% problem.id %]" action="/my/planned/change">
         <input type="hidden" name="id" value="[% problem.id %]">
         <input type="hidden" name="token" value="[% csrf_token %]">
         [% item_action | safe %]
--- a/templates/web/base/report/_item.html
+++ b/templates/web/base/report/_item.html
@@ -43,9 +43,13 @@ END;
   [% PROCESS 'report/_item_extra_class.html' %]
 [%~ CATCH file ~%]
 [%~ END ~%]
-
 <li class="item-list__item item-list--reports__item [% item_extra_class %]"
     data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]" id="report-[% problem.id | html %]">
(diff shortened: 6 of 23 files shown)


chrismytton (commented):
This looks like it's heading in the right direction!

I think I'd probably focus on the inspector tool first, rather than the reports list in the admin.

I'd look at adding a dropdown to the inspect form in the blue sidebar (see image below) to allow assigning a report to a user and handle that in Controller/Report.pm.

For the bulk assignment in the list of reports I wonder if we need to add a checkbox to each report row then have a GitHub-Issues-esqe actions bar at the top of the list of reports where you can assign all checked reports to a certain user.

Hope that makes sense. Happy to chat more about this tomorrow (Friday) if that's helpful.

image

struan (changes requested):
I've not gone over all the code in detail but it certainly works as I'd expect. Couple of small things to pick up though.

t/app/controller/report_inspect.t
```diff
@@ -50,7 +50,7 @@ $user->update( { from_body => $oxon } );
 
 my $sample_file = file(__FILE__)->parent->file("sample.jpg")->stringify;
 ok -e $sample_file, "sample file $sample_file exists";
-
+# goto HERE;
```
struan, line 53: is this left over debug?

templates/web/base/report/_item_small.html
```diff
@@ -30,3 +30,19 @@
...
+        [% IF problem.shortlisted_user.name %]
+            [% problem.shortlisted_user.name %]
+        [% ELSE %]
+            [% problem.shortlisted_user.username %]
+        [% END %]
+        </strong>
+    [% ELSE %]
+        <strong>(unassigned)</strong>
+    [% END %]
+[% END %]
```
struan, line 47: this is the same as `report-assignment.html`, I think?

templates/web/base/admin/report-assignment.html
```diff
@@ -0,0 +1,15 @@
+[%  IF c.user.has_permission_to('report_inspect', c.cobrand.body.id ) %] 
+    [%# see report's assigned user %]
+    <label class="inline-text" for="assignment">[% loc('Assigned to:') %]</label>
```
struan, line 3: I'm not sure this should be a label given there's no input element.

struan (changes requested):
Some small changes.

One thing I'm not sure about is how we best handle things like user's assigned to categories with "assigned categories only". At the moment you can assign a report to them if it's not in one of their categories and it will happen, but they then can't do anything with it. Possibly that's ok.

templates/web/base/report/_item_small.html
ludovic-tc: Yep, there were some slight differences in the presentation that were at odds with each other but I've resolved that and created a new mini-template.

t/app/controller/report_inspect.t
ludovic-tc: sorted

perllib/FixMyStreet/Cobrand/CheshireEast.pm
```diff
@@ -172,6 +172,15 @@ sub open311_post_send {
...
+ 
+sub available_permissions {
+    my $self = shift;
+
+    my $perms = $self->next::method();
 
+    $perms->{Problems}->{assign_report_to_user} = "Assign problem reports to users";
+
+    return $perms;
+}
```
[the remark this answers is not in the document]
ludovic-tc: fixed

templates/web/base/report/inspect/assignment.html
```diff
@@ -0,0 +1,24 @@
...
+                [% END %]
+                </option>
+            [% END %]
+        [% END %]
+        <option value='unassigned' 
+        [%- UNLESS problem.shortlisted_user -%] selected="selected"[%- END -%] 
+        >unassigned</option>
+    </select>
+    </p>
+[% END %]
```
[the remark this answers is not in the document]
ludovic-tc: actioned

templates/web/base/admin/report-assignment.html
dracos: A label has to be associated with a form control as it says at the link you provide, there's no form control present here, so it should not be a label.

templates/web/base/report/_item.html
```diff
@@ -43,9 +43,13 @@
   [% PROCESS 'report/_item_extra_class.html' %]
 [%~ CATCH file ~%]
 [%~ END ~%]
-
 <li class="item-list__item item-list--reports__item [% item_extra_class %]"
     data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]" id="report-[% problem.id | html %]">
+[% IF relevant_staff AND (c.user.has_permission_to('assign_report_to_user', problem.bodies_str_ids)) %]
+    [% UNLESS ((c.req.uri.path.match('\/my\/')) OR  (c.req.uri.path.match('\/around'))) %]
+        <input type='checkbox' class='bulk-assign' name='bulk-assign-reports' form='bulk-assign-form' value='[%- problem.id -%]' style='float:right; margin:2em'>
```
dracos, line 50: These checkboxes appear to unselectable (at least in Chrome/Firefox), trying to do so loads the report detail page instead. The assignment stars appear to still work okay.

templates/web/base/report/_item.html
```diff
@@ -43,9 +43,13 @@
   [% PROCESS 'report/_item_extra_class.html' %]
 [%~ CATCH file ~%]
 [%~ END ~%]
-
 <li class="item-list__item item-list--reports__item [% item_extra_class %]"
     data-report-id="[% problem.id | html %]" data-lastupdate="[% problem.lastupdate %]" id="report-[% problem.id | html %]">
+[% IF relevant_staff AND (c.user.has_permission_to('assign_report_to_user', problem.bodies_str_ids)) %]
+    [% UNLESS ((c.req.uri.path.match('\/my\/')) OR  (c.req.uri.path.match('\/around'))) %]
```
dracos, line 49: There is a `page` stash variable (used elsewhere in this template) you can check being equal to `around` or `my`.

web/cobrands/fixmystreet/staff.js
```diff
@@ -562,6 +568,19 @@ $.extend(fixmystreet.utils, {
...
+    update_unassigned: function($li, $ul) {
+        $li.find('strong').addClass('oldassign').hide();
+        $li.find('strong').after('<strong class="newassign">(unassigned)</strong>');
+    },
+    update_assignee: function($li, $ul) {
+        var user_name = $ul.data('userName');
+        var user_email = $ul.data('userEmail');
+        var user_handle = user_name ? user_name : user_email;
+        var $a2label = $li.find('span.assigned-to-label'); 
+        $li.find('strong').before('<span class="inline-text assigned-to-label" for="assignment" style="display: inline;">Assigned to: </span>');
```
dracos, line 581: Every time you click the star, an extra "Assigned to" appears.

templates/web/base/report/_get_assigned_user.html
```diff
@@ -0,0 +1,6 @@
+[% IF problem.user_planned_reports.user.name OR problem.shortlisted_user %] [%# this report is on a shortlist %]
```
dracos, line 1: I don't understand what's happening here, sorry. Why is checking `problem.shortlisted_user` not enough?

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -521,6 +521,20 @@ sub inspect : Private {
             }
         }
 
+        # set assignment
+        my $assigned = ($c->get_param('assignment'));
+        if ($assigned && $assigned eq 'unassigned') {
+            # take off shortlist...
```
dracos, line 527: This should use the standard `user->remove_from_planned_reports(report)` function, unless there is a reason it is not that's not mentioned.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -261,13 +261,36 @@ sub shortlist_multiple : Path('planned/change_multiple') {
     $c->forward('/auth/check_csrf_token');
 
     my @ids = $c->get_param_list('ids[]');
-
-    foreach my $id (@ids) {
-      $c->forward( '/report/load_problem_or_display_error', [ $id ] );
-      $c->user->add_to_planned_reports($c->stash->{problem});
+    my @bulk_reports = $c->get_param_list('bulk-assign-reports');
```
dracos, line 264: I'm not sure why this uses the same controller method as the shortlist-all-pins functionality - they don't appear to share anything bar the CSRF check and fetching a list of ids (which aren't even used in the new assign code). It would be clearer if it was it owns method/form endpoint.

templates/web/base/admin/problem_row.html
```diff
@@ -38,6 +38,22 @@
             [%- IF problem.is_closed %]<br>[% prettify_state('closed') %]: [% PROCESS format_time time=problem.lastupdate %][% END -%]
             [%- IF problem.is_open %]<br>[% loc('Last&nbsp;update:') %] [% PROCESS format_time time=problem.lastupdate %][% END -%]
         </small></td>
+        [%- IF c.user.has_permission_to('planned_reports', c.cobrand.body.id ) %] 
+        [%# see report's assigned user %]
+        <td>
+            <span>
+            [%- IF problem.user_planned_reports -%] [%# this report is on a shortlist %]
+                [%- IF problem.shortlisted_user.name -%]
```
dracos, line 46: Similar to before, could be `[% problem.shortlisted_user.name OR problem.shortlisted_user.username %]`

templates/web/base/admin/problem_row.html
```diff
@@ -38,6 +38,22 @@
...
+                    [%- problem.shortlisted_user.name -%]
+                [%- ELSE -%]
+                    [%- problem.shortlisted_user.username -%]
+                [%- END -%]
+            [%- ELSE -%]
+                unassigned
```
dracos, line 52: Any bare string like this in a core template needs marking as a translation string, so this would be `[% loc('unassigned') %]`. I haven't checked for all occurrences.

templates/web/base/admin/problem_row.html
```diff
@@ -38,6 +38,22 @@
             [%- IF problem.is_closed %]<br>[% prettify_state('closed') %]: [% PROCESS format_time time=problem.lastupdate %][% END -%]
             [%- IF problem.is_open %]<br>[% loc('Last&nbsp;update:') %] [% PROCESS format_time time=problem.lastupdate %][% END -%]
         </small></td>
+        [%- IF c.user.has_permission_to('planned_reports', c.cobrand.body.id ) %]
```
dracos, line 41: You can see other users of has_permission_to in this same template check `problem.bodies_str_ids`, not `c.cobrand.body.id` - good to be consistent. (Also, you could theoretically be looking at a problem from a different body on one body and thus not have permission.)

templates/web/base/admin/report-assignment.html
```diff
@@ -0,0 +1,7 @@
+[%  IF c.user.has_permission_to('report_inspect', c.cobrand.body.id ) %]
```
dracos, line 1: I think anyone who can see the report in backend admin interface should be able to see this, no need for a permission check.

templates/web/base/admin/reports/index.html
```diff
@@ -14,6 +14,7 @@
         <th>[% loc('Name') %]</th>
         <th>[% loc('Body') %]</th>
         <th>[% loc('State') %]</th>
+        [% IF c.user.has_permission_to('assign_report_to_user', c.cobrand.body.id) %]<th>[% loc('Assigned to') %]</th>[% END %]
```
dracos, line 17: Hmm, is this a big enough feature that it gets its own column in what is already quite a wide table?
Would it be okay as an extra row in the "Body" column perhaps, given it's associated with the body?
Not sure.

Also the permission here - I think the usual way we do this is `c.user.has_body_permission_to(permission)` which does a `user.from_body` check.

templates/web/base/report/_item_small.html
```diff
@@ -30,3 +30,11 @@
     [% ELSIF problem.bodies_str_ids.size == 0 %] [% loc('(not sent to council)') %]
     [% END %]
 [% END %]
+<div class='assigned-to' style='margin-top:0.25em'>
+[%  IF c.user.has_permission_to('report_inspect', c.cobrand.body.id ) %]
```
dracos, line 34: other permission checks on this page use `problem.bodies_str_ids`.

Thinking aloud - what if a user has the assign permission but not the report_inspect permission? You can assign but not see assignments? Do we want to make it work, or just assume anyone with the former permission will also have the latter.

templates/web/base/reports/_bulk-assign.html
```diff
@@ -0,0 +1,11 @@
+[%  IF c.user.has_permission_to('assign_report_to_user', c.cobrand.body.id ) %]
```
dracos, line 1: probably another `has_body_permission_to`

dracos (changes requested):
Lot of good work gone into this I can see, in various parts of the code. I'm afraid I have made a number of suggestions and required changes - most notably of which is the checkboxes don't appear to work, sorry. Plus some small changes, ideas to improve database performance, simplify bits, some permission things - there's quite a few, but hopefully each individually is quite small and clear, shout if anything isn't. As this was quite a big PR (well, regardless of size, but even more in this case), please be sure to fix things as fixup commits so it's easier to rereview, thanks.

templates/web/base/admin/report-assignment.html
ludovic-tc: I am using the same template in admin and in the frontend reports list. I am moving out of 'admin' as part of the changes relating to your earlier comment on that template.

templates/web/base/reports/_bulk-assign.html
```diff
@@ -0,0 +1,11 @@
...
+<form id='bulk-assign-form' name='bulk-assign-form' method='post' action='/my/planned/change_multiple'>
+    <label style='font-weight:normal;font-size:small;display:inline;vertical-align:sub;'>Assign to:</label>
+    <input type="hidden" name="token" value="[% csrf_token %]">
+    <select class="form-control multi-select-button" name="inspector" id="inspector" style="height:1.8em;font-size:0.75em;display:inline;padding:0;">
+        [% INCLUDE 'report/inspect/_assignment-options.html' %]
+    </select>
+    <button id='bulk-assign-btn' class='btn' type='submit' name='bulk-assign-btn' value='bulk-assign'
+        style='font-weight:normal;font-size:0.75em;height:2em;vertical-align:bottom;margin-bottom:0.5em'>Assign</button>
```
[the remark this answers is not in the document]
ludovic-tc: I have amended it, but I'm still not really happy with it either. Can you give me an example of how it ought to look? If I don't style it, it just looks ridiculously big. I have adapted a style from nearby controls but I don't think it looks button-like enough.

dracos:
Sorry to add something new, but I have only just noticed - the list of reports on the front page of the site is showing assignment and checkboxes which probably shouldn't be shown here.

templates/web/base/reports/_bulk-assign.html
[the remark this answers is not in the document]
dracos: Sorry, you appear to be overthinking this. Right above where you have your form, you already have a form that is displaying items and buttons that look fine. It seems to me all you need to do is copy what already exists, put a `<p class="report-list-filters">` inside your form and have a plain input type="submit" with no styling at all and that looks fine.

templates/web/base/admin/reports/index.html
```diff
@@ -14,7 +14,7 @@
         <th>[% loc('Name') %]</th>
         <th>[% loc('Body') %]</th>
         <th>[% loc('State') %]</th>
-        [% IF c.user.has_permission_to('assign_report_to_user', c.cobrand.body.id) %]<th>[% loc('Assigned to') %]</th>[% END %]
+        [% IF c.user.has_permission_to('assign_report_to_user', problem.bodies_str_ids) %]<th>[% loc('Assigned to') %]</th>[% END %]
```
dracos: There is no `problem` at this point (so lucky you dropped the column!).

templates/web/base/admin/_report-assignment.html
```diff
@@ -1,4 +1,4 @@
-[%  IF c.user.has_permission_to('report_inspect', problem.bodies_str_ids ) %]
+[%  IF c.user.has_body_permission_to('report_inspect') OR c.user.has_body_permission_to('assign_report_to_user') %]
```
dracos, line 1: The template that includes this e.g. item_small uses `c.user.has_permission_to('report_inspect', problem.bodies_str_ids)` so I would keep that if you have a problem to check against, rather than switch to has_body_permission_to.

dracos (changes requested):
Thanks, looks like you've covered almost everything I asked for, great.

There are still a couple of issues I'd like fixing before putting on staging (the ordering of users, and the bug in the assigned user display, as otherwise they both might be noticed by testing), but the rest don't have to be done before that, they're only internal improvements.

On a more general note - fixup commits ideally should be able to be done using `git commit --fixup <ID>` to fix a previously made commit. The fact this can't be done here is sort of a tacit acknowledgement that the original code has too much information in one commit. "Cheshire East Inspector Tool Upgrade" is not one thing, it is many things as its own commit message goes on to explain (if you're explaining what a commit does as "A and B and C", then that's probably 3 commits). I would expect perhaps e.g. one commit to add the new assignment permission (and anything involved with that), one to add the dropdown to the inspector tool, one to add the bulk assignment capability, one or two for displaying the assignments, and so on. That would make each part more self-contained and easy to check (perhaps independently), and easier to make fixups against accordingly. Of course, development doesn't always/ever work neatly like that, but the parts I mention all sound quite distinct and e.g. I don't think it would take long once all these fixups are squashed together to split up the one commit into a few separate ones, or to have done that as you were going along.

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -263,13 +263,46 @@ sub shortlist_multiple : Path('planned/change_multiple') {
...
-      $c->forward( '/report/load_problem_or_display_error', [ $id ] );
-      $c->user->add_to_planned_reports($c->stash->{problem});
+        $c->forward( '/report/load_problem_or_display_error', [ $id ] );
+        $c->user->add_to_planned_reports($c->stash->{problem});
     }
 
     $c->res->body(encode_json({ outcome => 'add' }));
 }
 
+sub bulk_assign : Path('planned/bulk_assign') {
```
dracos, line 273: This function has no authentication - anyone could change anyone's assigned reports.

dracos (changes requested):
There is one security issue to resolve. Plus three of my comments from my previous review (the ones I have left unresolved) are still extant (one template, and no discussion on why the existing code could not be used).

dracos (approved):
Thanks, this all looks good now code-wise. It will need some tidying up of commits, potentially as I discussed previously at https://github.com/mysociety/fixmystreet/pull/3535#pullrequestreview-755385196

dracos (approved):
Breaking up into commits looks good :) I realise it is sometimes hard/not possible, but do try to think about this final situation as you're starting/going along, as it makes it easier to deal with fixups, reviewing, and tidying up at the end.

I'm not sure what's going on with Codecov, which says you haven't tested the bulk assign action but I can see that you have, so guess we'll just have to ignore that!

PR #3536, opened 2021-07-16
[Bromley] Add contributing user's role to open311 report_title

chrismytton:
Bromley have requested that we add the role into the open311 report_title field so that they can use it to update the point of contact field in their Confirm instance.

Fixes https://github.com/mysociety/societyworks/issues/2450


dracos (approved):
🎢

Issue #3544, opened 2021-07-27
Swedish language pack should be part of install for tests to pass on linux


MorayMySoc:
**Describe the bug**
The Swedish language pack is not installed by default on a linux install. This means the test suite fails when testing the translation feature.

**To Reproduce**
Do a manual install on fresh Ubuntu/Debian and run the test suite

**Expected behavior**
Test suite should run with no errors

**Desktop (please complete the following information):**
 - Ubuntu 20.04
 - 
**Additional context**

PR #3546, opened 2021-07-27
Adds language-pack-sv and language-pack-de to install

MorayMySoc:
Swedish and German language packs should be part of install for tests to pass on linux #3544

This makes Swedish and German language packs part of the manual install for debian/ubuntu linux instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@
         - Include failure count in send report error output, #3316
         - Sort output in export script. #3323
         - Show relevant updates in alert-update email preview. #3417
+        - Add German and Swiss language packs to default
+        linux install #3544
         - Upgrade jQuery. #3017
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
If this is to be added (and I see no reason not to, you're right, the tests should pass by default you would hope), I think we also need to add language-pack-de (at least, the automated tests add that one as well), and should probably add to packages.docker as well to maintain consistency there.

MorayMySoc:
Great - was about to get in touch to make sure this was worth pursuing and to say if so should add de too. I'll add that and have a look at the Docker file get the branch back up to date with master.

dracos (approved):
🇩🇪 🇸🇪

MorayMySoc:
Merged locally and pushed

PR #3549, opened 2021-07-28
script to transfer response templates

struan:
Enables transferring of response templates between two installs.

### DIFF
--- a/bin/export-import-data
+++ b/bin/export-import-data
@@ -2,6 +2,8 @@
 
 use v5.14;
 use warnings;
+# avoid wide character warnings
+no warnings 'utf8';
 
 BEGIN {
     use File::Basename qw(dirname);
@@ -20,6 +22,7 @@ my ($opt, $usage) = describe_options(
     [ 'name=s', "Name of body" ],
     [ 'import=s', "File to import" ],
     [ 'commit', "Actually commit changes to the database" ],
+    [ 'update-templates', "Update existing templates (default is to skip)" ],
     [ 'categories=s', "pipe-separated list of categories to export contacts for and filter templates by" ],
     [ 'help', "print usage message and exit", { shortcircuit => 1 } ],
 );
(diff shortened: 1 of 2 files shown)


davea (changes requested):
Not sure if the same(ish) file being present twice is deliberate? Given the strong overlap with `export-import-data` I think this would be better implemented as a flag (e.g. `--update-templates`) on that existing script.

davea (changes requested):
Sorry, I might be missing something but it looks like the flag isn't taken into account on L138 of `export-import-data` so existing templates are always skipped. Also `bin/transfer-updates-response-templates` is still hanging around.

davea (changes requested):
This is crashing when trying to e.g. reimport a file I just exported:

```
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st execute failed: ERROR:  duplicate key value violates unique constraint "response_templates_body_id_title_key"
DETAIL:  Key (body_id, title)=(21076, Action Scheduled - Blocked Drain) already exists.
[...]
 at ./bin/export-import-data line 148
```

bin/export-import-data
```diff
@@ -130,11 +134,11 @@ sub import {
 
     foreach (@{$out->{templates}}) {
         my $existing = $body->response_templates->search({ title => $_->{title} })->single;
-        if ($existing) {
+        if ($existing && !$update_templates) {
             warn "Template with title $_->{title} already exists, skipping";
             next;
         }
-        my $template = $body->response_templates->new({
+        my $template = $body->response_templates->find_or_new({
```
davea, line 141: This won't update the text or status of any existing templates unfortunately. `find_or_new` will (I believe) only search on unique columns, meaning it'll match on `body_id` and `title` then skip over the values for `text` and `state`.

davea (changes requested):

bin/export-import-data
struan: sigh, yes :(

davea (approved):
🌟

PR #3550, opened 2021-07-28
[Bromley] make sure duplicated report has a name

struan:
It's possible for the user for a subscription not to have a name so fall
pack to the one on the report to avoid DB errors.

[skip changelog]


chrismytton (approved):
📛

PR #3553, opened 2021-07-29
[Peterborough] update flytipping category names

struan:
[skip changelog]


chrismytton (approved):

dracos:
It looks like Peterborough are using the display name for this category, so I'm guessing this isn't needed?

PR #3556, opened 2021-07-30
[Buckinghamshire] Add help text to time field

MorayMySoc:
Add tooltip / label on IE 11 Red Claims Time Picker [0.5d] 

https://github.com/mysociety/societyworks/issues/2471

Ticket created from https://mysocietysupport.freshdesk.com/a/tickets/1280

Client tester says confused by form in IE11 requiring date in AM/PM format rather than 24 hour clock.

Seems this is a feature of the base HTML 5 time element - it honours the computers locale settings for how to enter time.
I have observed the AM/PM input on all browsers through browswerstack, so think it's not a browser issue.

But adding hint text seems to be a good all round solution.

### DIFF
--- a/perllib/FixMyStreet/App/Form/Claims.pm
+++ b/perllib/FixMyStreet/App/Form/Claims.pm
@@ -339,6 +339,7 @@ has_field incident_time => (
     required => 1,
     type => 'Text',
     html5_type_attr => 'time',
+    tags => { hint => 'For example 06:30 PM or 18:30 depending on your settings' },
     label => 'What time did the incident happen?',
 );
 
--- a/web/cobrands/sass/_claims.scss
+++ b/web/cobrands/sass/_claims.scss
@@ -12,7 +12,7 @@ body.formflow {
     }
 
     input[type="time"].govuk-input {
-      max-width: 12ex;
+      max-width: 20ex;
     }
 
     input[type="text"].govuk-input, input[type="time"].govuk-input {


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
         - Update URL on /my when map moves. #3358
         - Make anonymous updates clearer in email alerts. #3417
         - Add Maidenhead Locator support to search box.
+        - Add helptext to Red Claims time about 12/24 hour times #2471
```
struan, line 17: You don't need a change log for commercial things :)

perllib/FixMyStreet/App/Form/Claims.pm
```diff
@@ -339,6 +339,7 @@ has_field incident_time => (
     required => 1,
     type => 'Text',
     html5_type_attr => 'time',
+    tags => { hint => 'For example 06:30 PM or 18:30 depending on your settings' },
```
struan, line 342: Probably need to make the control wider too as, in chrome at least, you can't see the am/pm bit:

Screenshot 2021-08-03 at 11 40 11

struan (changes requested):
couple of small changes, also, I think this is linking to the wrong ticket in the description.

perllib/FixMyStreet/App/Form/Claims.pm
MorayMySoc: I've had a look at this and seems like our time field size is set by the `base.css`. Though I could change this, is does seem like that could have a wider system effect and perhaps I should raise the issue with the design team?

struan: looks like there's something for it in _claims.scss

struan (approved):
looks good.

MorayMySoc:
Just to highlight for future self and others that this was merged in, not through github

PR #3566, opened 2021-08-13
Fixes issue with not being able to order dropdowns in extra questions

MorayMySoc:
Ordering of dropdown list options is now sorted numerically on the index position. This means the order will be that expected by the user when creating the options.

https://github.com/mysociety/societyworks/issues/2511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
         - Fix checked order of updates in dashboard export.
         - Fix unable to edit user with verified landline #3295
         - Fix 'sites' page to reflect active fixmystreet sites #2481
+        - Fix ordering of dropdown lists in extra questions #3566
     - Admin improvements:
         - Enable per-category hint customisation.
         - Move ban/unban buttons to user edit admin page.
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -507,10 +507,8 @@ sub fetch_body_areas : Private {
 
 sub update_extra_fields : Private {
     my ($self, $c, $object) = @_;
-
     my @indices = grep { /^metadata\[\d+\]\.code/ } keys %{ $c->req->params };
     @indices = sort map { /(\d+)/ } @indices;
-
     my @extra_fields;
     foreach my $i (@indices) {
         my $meta = {};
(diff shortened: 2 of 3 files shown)


perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -544,6 +542,7 @@ sub update_extra_fields : Private {
                         $disable ? (disable => 1, disable_message => $disable_message) : (),
                     }) if $name;
                 }
+                @{$meta->{values}} = sort { $a->{key} <=> $b->{key} } @{$meta->{values}};
```
dracos, line 545: I don't think we can do this, because the order in general is as it comes from the Open311 server, and doing this would change that order. For example, we have many Yes/No answers which have Yes followed by No (as both key and name) , and this would reorder them to No/Yes. We cannot assume that it should be shown in key order.
I think as long as the order shown on the admin web page is maintained, that would be okay, and someone could reorder/rename if needed - the issue is that the sort on line 533 was sorting alphabetically, so the results from the web page would then go 1, 10, 11, ... 2, 3, 4, .... rather than what was entered. If that sort was reinstated but made numerical, then whatever order the user had on the admin page would be kept (and any order created by Open311 import would be maintained).

dracos: Also the key might not be numerical, it can be anything the council wants. And a regression test to show the admin no longer reorders something with many values would be great.

dracos (changes requested):
I think this needs to be done a different way, I'm afraid!

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
         - Only trigger one refresh going Back to list view. #3476
         - Fix checked order of updates in dashboard export.
         - Fix unable to edit user with verified landline #3295
+        - Fix ordering of dropdown lists in extra questions #2511
```
dracos, line 29: This is a public file, so any issue numbers mentioned can only be public issue IDs (if there isn't a public bug issue (no reason there couldn't have been in this case, but doesn't matter now), then generally we put the PR number here instead so the line can be followed back to the relevant change).

dracos (changes requested):
Code change looks good, test looks good, thanks, only issue is the ID referenced in the changelog :)

dracos (approved):

PR #3569, opened 2021-08-19
[Bromley] Private comments for reports

chrismytton:
Adds private comments field to the report form for people with the `can_contribute_as_body` permission. This is then sent over open311 in the description field.

Bromley also want this field adding to updates, but I'm going to do that as a separate change.

Part of https://github.com/mysociety/societyworks/issues/2293


dracos (approved):
🤫

dracos (approved):
Consistent with other after templates, looks good

PR #3571, opened 2021-08-26
[Bucks] Integrate Kaarbontech feed with drains

MorayMySoc:
https://github.com/mysociety/community/issues/111

* Uses Kaarbontech wps feed to map gullies
* Uses information in the feed to create a user message about gully cleaning status

[skip changelog]

### DIFF
--- a/web/cobrands/buckinghamshire/assets.js
+++ b/web/cobrands/buckinghamshire/assets.js
@@ -4,9 +4,13 @@ if (!fixmystreet.maps) {
     return;
 }
 
+var wfs_host = fixmystreet.staging ? 'tilma.staging.mysociety.org' : 'tilma.mysociety.org';
+var tilma_url = "https://" + wfs_host + "/mapserver/bucks";
+var proxy_base_url = "https://" + wfs_host + "/proxy/bcc/";
+
 var defaults = {
     http_options: {
-        url: "https://tilma.mysociety.org/mapserver/bucks",
+        url: tilma_url,
         params: {
             SERVICE: "WFS",
             VERSION: "1.1.0",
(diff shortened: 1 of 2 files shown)


web/cobrands/buckinghamshire/assets.js
```diff
@@ -169,14 +173,43 @@ fixmystreet.assets.add(labeled_defaults, {
...
+
+        function munge_date(date_field) {
+            var RegExDate = /([0-9]{4})-([0-9]{2})-([0-9]{2})/;
+            var myMatch = RegExDate.exec(date_field);
+            if (myMatch) {
+                return myMatch[3] + '/' + myMatch[2] + '/' + myMatch[1];
+            } else {
+                return '';
+            }
+        }
```
chrismytton, line 207: I'm not sure if we have a specific policy on this, but I'd prefer this `munge_date` function to be at the top level, rather than nested inside the calling function.

web/cobrands/buckinghamshire/assets.js
```diff
@@ -169,14 +173,43 @@ fixmystreet.assets.add(labeled_defaults, {
...
+            if (last_clean != '') { message += 'This gulley was last cleaned on ' + last_clean; }
+            if (next_clean != '' && last_clean != '') {
+                message += '. It will be cleaned again before ' + next_clean;
+            } else if (next_clean != '') {
+                message += 'This gulley will be cleaned before ' + next_clean;
+            }
+        return message || "This gulley has no cleaning history";
+
+        function munge_date(date_field) {
+            var RegExDate = /([0-9]{4})-([0-9]{2})-([0-9]{2})/;
```
chrismytton, line 200: The capital `R` made me think this was referencing a class for a moment. Perhaps `regExDate` would be clearer? Again I don't think we actually have a policy on this!

chrismytton (changes requested):
This looks good! Couple of minor suggestions from me.

The latest message on the Freshdesk thread also sounds like it might involve a code change to switch from the `junctions` layer to the `junction_inspections` layer.

chrismytton (approved):
Looks good to me @MorayMySoc! ✨ 

My one (minor) concern is that the gulley cleaned text is a bit subtle. Took me a few minutes to figure out I needed to scroll to the bottom of the categories to see the last cleaned date! I guess that's up to the client to decide. Let's get this on staging!

MorayMySoc:
Feedback on the ticket https://mysocietysupport.freshdesk.com/a/tickets/1083

chrismytton (approved):
Looks good to me!

PR #3572, opened 2021-08-26
WIP on 2462-limit-bin-request

MorayMySoc:
The idea is to limit users from repeating a report of a missed bin within a bin cycle - so if they've reported it missed, they have to wait until next cycle to report another missed bin.

When testing in `bromley.t` I've mocked out `getEventsForObjects` to return an empty array which allows me to test that the 'report within two days' feature is working despite the mock data coming from `Echo.pm` reporting as completed.

I'm trying to repeat this on `waste.t`, but finding that the report of completed still exists so I don't seem to be influencing it. I'm pretty sure I'm making a dumb error that I can't see or understand. Thought I had it after `bromley.t` but seems I'm not understanding after all.


MorayMySoc:
Not right way to go about this - closed without merging

PR #3576, opened 2021-09-02
[Bromley] Update DD lead time text after garden waste subscription

ludovic-tc:
Updated text in direct_debit_in_progress email templates to match true lead times. 

For https://mysocietysupport.freshdesk.com/a/tickets/1401

[skip changelog]


dracos (approved):
Looks good to me

PR #3579, opened 2021-09-03
[Bromley] email amendments

ludovic-tc:
Adds bin days link with optional personalisation to new garden waste subscription email
for https://mysocietysupport.freshdesk.com/a/tickets/1410

[skip changelog]

### DIFF
--- a/t/cobrand/bromley.t
+++ b/t/cobrand/bromley.t
@@ -7,7 +7,10 @@ use Test::Output;
 use FixMyStreet::TestMech;
 use FixMyStreet::SendReport::Open311;
 use FixMyStreet::Script::Reports;
+use FixMyStreet::Script::Alerts;
 use Open311::PostServiceRequestUpdates;
+use List::Util 'any';
+use Regexp::Common 'URI';
 my $mech = FixMyStreet::TestMech->new;
 
 # disable info logs for this test run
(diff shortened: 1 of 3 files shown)


templates/email/bromley/_council_reference_alert_update.html
```diff
@@ -0,0 +1,3 @@
+[% IF title == 'Garden Subscription - New' %]
+<a href="/waste[% IF property_id %]/[% property_id %][% END %]">Check your bin collections day</a>
```
dracos, line 2: I don't think `property_id` is available at this point? I can't see what would be setting it. You need to look it up in the template from `report`, see how the link I mentioned on the report page does it.

templates/email/bromley/_council_reference_alert_update.html
```diff
@@ -0,0 +1,3 @@
+[% IF title == 'Garden Subscription - New' %]
```
dracos, line 1: It would seem appropriate to include this in any waste email, not just new garden subscriptions? `report.cobrand_data` will be `waste` for any waste report.

dracos (changes requested):
I don't think the `property_id` lookup will ever work, I'm afraid; this needs a small test to show that it is outputting correctly.

dracos (changes requested):
I asked you to write a test, which you haven't; that would have shown you that this doesn't work, because you have used `problem` instead of `report`, and the property_id IF in the txt is incorrect.

t/cobrand/bromley.t
```diff
@@ -1775,6 +1778,83 @@ subtest 'check direct debit reconcilliation' => sub {
...
+
+        FixMyStreet::Script::Alerts::send();
+
+        my $alerts = FixMyStreet::App->model('DB::Alert')->search({
+            alert_type => 'new_updates',
+            parameter => $new_sub->id,
+        });
+        is $alerts->count, 1, "alert found for GGW update";
+
+        $mech->email_count_is(1, 'GGW update alert email found');
```
dracos, line 1842: This line, the line below, and the for loop can be replaced (the for loop itself, I mean, not its contents) as `my $email = $mech->get_email` - when used in a scalar context, it checks for 1 email and returns it directly.

t/cobrand/bromley.t
```diff
@@ -1775,6 +1778,83 @@ subtest 'check direct debit reconcilliation' => sub {
...
+        my @emails = $mech->get_email;
+        for my $email (@emails) {
+            my $text_body = $mech->get_text_body_from_email($email);
+            like $text_body, qr/Check your bin collections day/, 'has bin day link text in text part';
+            my @links = $mech->get_link_from_email($email, 'get_all_links');
+            my $found = any { $_ =~ m[recyclingservices\.bromley\.gov\.uk/waste] } @links;
```
dracos, line 1848: Why isn't this looking for the property ID in the URL? Doing so would show that I'm afraid that it still isn't working and including the property ID which I assume you expected to be included by the test data. Your test data is what is wrong, property_id is not a key in an extra hash, it is a field. See how `setup_dd_test_report` does it. In fact, why not use `setup_dd_test_report` rather than your own create_problems_for_body call. You could edit it to set cobrand_data = 'waste', then call it as `my $new_sub = setup_dd_test_report({ 'property_id' => '54323' });` and that's all you'd need.

t/cobrand/bromley.t
```diff
@@ -1775,6 +1778,83 @@ subtest 'check direct debit reconcilliation' => sub {
...
+};
+
+subtest 'Garden Waste new subs alert update emails contain bin collection days link' => sub {
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => 'bromley',
+    }, sub {
+        $mech->clear_emails_ok;
+        my $contact = $mech->create_contact_ok(body => $body, category => 'Garden Subscription', email => 'garden@example.com');
+
+        my $new_sub = ($mech->create_problems_for_body( 1, $body->id, 'Garden Subscription - New', {
```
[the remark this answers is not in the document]
dracos: (And see my later comment for using the existing function in this test file for creating a report.)

dracos (changes requested):
Looking better, but the test needs a few changes to simplify it and test the property ID is actually being output.

t/cobrand/bromley.t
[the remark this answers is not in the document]
ludovic-tc: Thanks, this is a much neater usage.

t/cobrand/bromley.t
ludovic-tc: What was I thinking?

t/cobrand/bromley.t
ludovic-tc: I did initially utilise setup_dd_test_report() and am now reverting to do that. I thought extra fields should be handled differently, but then I found some code where a problem was being created with a hashref called 'extra', and borrowed that. I thought this must mean that it would get turned into the extra_fields. It appeared to work (at least in the test), so I didn't look into it any more deeply.

t/cobrand/bromley.t
```diff
@@ -1833,27 +1815,19 @@ subtest 'Garden Waste new subs alert update emails contain bin collection days l
...
-            my $found = any { $_ =~ m[recyclingservices\.bromley\.gov\.uk/waste] } @links;
-            ok $found, 'Found bin day URL in text part of alert email';
-
-            my $html_body = $mech->get_html_body_from_email($email);
-            like $html_body, qr/Check your bin collections day/, 'has bin day link text in HTML part';
-            my @uris = $html_body =~ m/$RE{URI}/g;
-            $found = any { $_ =~ m[recyclingservices\.bromley\.gov\.uk/waste] } @uris;
-            ok $found, 'Found bin day URL in HTML part of alert email';
-        }
+        $mech->email_count_is(1, 'Garden waste update alert email found');
```
dracos, line 1818: You don't need this line, because the line below performs the same check in `get_email`.

t/cobrand/bromley.t
dracos: Yes, you can also store extra data in a hashref in extra, but that's not how Open311 extra question data (which property_id is treated as) are stored, they're in a _fields list (see Role::Extra for the underlying).

dracos (approved):
One line can be removed, think everything is covered :)

PR #3580, opened 2021-09-03
[Merton] New cobrand for Merton

mySocietyClive:
Adds new cobrand for Merton Council, I copied perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm, so .pm file needs to be checked for correct map ID, and disambiguate_location etc.

For https://github.com/mysociety/societyworks/issues/2552

[skip changelog]


chrismytton (commented):
Thanks @mySocietyClive, this is looking great!

I've pushed a fixup commit to remove the bits of `Merton.pm` that aren't needed and update the `disambiguate_location` method.

The only thing that I've noticed is the example place text isn't visible currently:


https://user-images.githubusercontent.com/22996/132024320-87b4ec67-66b4-49a3-a2c2-2df54222b3ee.mov

Note that to get the example places to show up you'll need to add them into `general.yml`, in `example_places` under `COBRAND_FEATURES`, something like this:

```yaml
COBRAND_FEATURES:
  example_places:
    merton: [ 'SM4 5DX', 'High Street' ]
```

mySocietyClive:
> Thanks @mySocietyClive, this is looking great!
> 
> I've pushed a fixup commit to remove the bits of `Merton.pm` that aren't needed and update the `disambiguate_location` method.
> 
> The only thing that I've noticed is the example place text isn't visible currently:
> 
>  example_places.mov 
> Note that to get the example places to show up you'll need to add them into `general.yml`, in `example_places` under `COBRAND_FEATURES`, something like this:
> 
> ```yaml
> COBRAND_FEATURES:
>   example_places:
>     merton: [ 'SM4 5DX', 'High Street' ]
> ```

Hi @chrismytton could you have another look please, postcode hint wasn't visible as the text colour was white, on a white background... have fixed that now.

Cheers

chrismytton (approved):
Nice one @mySocietyClive! All looks good to me, thanks for this.

I'll take it from here. Will put it on the staging site and let the client know.

mySocietyClive:
> Nice one @mySocietyClive! All looks good to me, thanks for this.
> 
> I'll take it from here. Will put it on the staging site and let the client know.

Cheers @chrismytton

PR #3583, opened 2021-09-07
[Bromley] Remove unneeded cancellation code.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2559 [skip changelog]
Didn't have to change any tests, dunno if that was good or bad thing!


struan (approved):
There's one bit of comment about cancellations above the if statement for payments that might be better moved but otherwise this is good.

I think I must have moved all the tests before. Not sure why I didn't just delete the cancel code at the same time :|

PR #3587, opened 2021-09-13
Do not email inactive body comment users.

dracos:
Just had such an email. They are not deleted, so do not email them either.


MorayMySoc (approved):
Can't see any issues with this.

PR #3588, opened 2021-09-13
[Open311] Move test handling out of core code.

dracos:
Instead, patch Open311 in the test code to store requests and inject responses


chrismytton (approved):
Great to get all that testing code out of core!

Looks like the changelog check is failing. Do you think it's worth adding a changelog entry for this? Guess it might be useful for reusers to know about it.

PR #3589, opened 2021-09-13
[Bexley] Remove COVID-19 messages

ludovic-tc:
This removes the banner previously added in the following places:

- The homepage
- Problem confirm email
- Page when you create a problem as an unregistered user
- Page when you create a problem as a registered user

[skip changelog]

for https://mysocietysupport.freshdesk.com/a/tickets/1456


chrismytton (approved):
🦠 1️⃣ 9️⃣ 👍

PR #3592, opened 2021-09-15
Fix issue searching for numeric IDs.

dracos:
There is an issue with Number::Phone that means it is matching e.g. 402609 as a Jersey phone number. Reported and behaviour improved at https://github.com/DrHyde/perl-modules-Number-Phone/issues/101, but until a new release, move the searches around so a ref search takes precedence. [skip changelog] For FD#1398


chrismytton (approved):
☎️

PR #3596, opened 2021-09-20
[Buckinghamshire] Give Bucks same pins as Oxfordshire CC

ludovic-tc:
Bucks Council would like to have the same pins as Oxfordshire County Council. 

for https://mysocietysupport.freshdesk.com/a/tickets/1452

[skip changelog]


dracos (approved):
🚥

PR #3599, opened 2021-09-22
[Highways England] Fix litter responsibilties for HE

MorayMySoc:
https://github.com/mysociety/societyworks/issues/1268

* Use new layer to see if Highway England responsible for litter on A-road
* If they are responsible make sure HE Litter category available
* If they are not responsible make sure council litter category available
* Also Switch to EPSG:4326 for site code lookup in report sending as
these changes break test otherwise

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -759,6 +759,8 @@ sub stash_category_groups : Private {
             $a->[0] cmp $b->[0];
         } @list;
         @list = map { $_->[1] } @list;
+
+        $c->cobrand->call_hook(munge_mixed_category_groups => \@list);
         $c->stash->{category_groups} = \@list;
         return;
     }
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -174,6 +174,12 @@ sub munge_unmixed_category_groups {
     }
 }
 
+sub munge_mixed_category_groups {
+    my ($self, $list) = @_;
+    my $nh = FixMyStreet::Cobrand::HighwaysEngland->new({ c => $self->{c} });
+    $nh->national_highways_cleaning_groups($list);
+}
+
 sub munge_load_and_group_problems {
     my ($self, $where, $filter) = @_;
 
--- a/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
+++ b/perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
@@ -158,6 +158,65 @@ sub munge_report_new_contacts {
     }
 }
 
+sub munge_mixed_category_groups {
+    my ($self, $list) = @_;
+
+    for my $cat_item (@$list) {
+        my $clean_name = $cat_item->category_display;
+        $clean_name =~ s/ \(NH\)//;
+        $cat_item->set_extra_metadata('display_name', $clean_name);
+    }
+}
+
+sub national_highways_cleaning_groups {
+    my ($self, $category_groups) = @_;
+
+    # We only care if there is a National Highways group
+    my ($nh_group) = grep { ($_->{name} || '') eq $self->council_name } @$category_groups;
+    return unless $nh_group;
+
+    my $c = $self->{c};
+    my $not_he_litter = $c->stash->{report_new_is_on_he_road_not_litter};
+    if (!defined $not_he_litter) {
+        # We delayed working it out because every /around page calls
+        # report_new_is_on_he_road but do it now we need to know
+        my ($x, $y) = ($c->stash->{longitude}, $c->stash->{latitude});
+        $not_he_litter = $self->_report_new_is_on_he_road_not_litter($x, $y);
+    }
+
+    for my $cat_item (@{$nh_group->{categories}}) {
+        my $clean_name = $cat_item->category_display;
+        $clean_name =~ s/ \(NH\)//;
+        $cat_item->set_extra_metadata('display_name', $clean_name);
+    }
+
+    # Don't change anything else unless we're on a HE non-litter road
+    return unless $not_he_litter;
+
+    # NH do not want flytipping or litter reports on these roads, take them out
+    @{$nh_group->{categories}} = grep { $_->category_display !~ /Flytipping|Litter/ } @{$nh_group->{categories}};
+
+    # Put any council street cleaning categories we can find into the NH group,
+    # so they'll still appear if "on the HE road" is picked
+    my %cleaning_cats = map { $_ => 1 } @{ $self->_cleaning_categories };
+    for my $cat (@$category_groups) {
+        if ($cat->{name}) {
+            next if $cat->{name} eq $self->council_name;
+            foreach my $c (@{$cat->{categories}}) {
+                if ($cleaning_cats{$cat->{name}}) {
+                    my $full_title = "$cat->{name}: " . $c->category_display;
+                    $c->set_extra_metadata(display_name => $full_title);
+                    push @{$nh_group->{categories}}, $c;
+                } elsif ($cleaning_cats{$c->category_display}) {
+                    push @{$nh_group->{categories}}, $c;
+                }
+            }
+        } elsif ($cleaning_cats{$cat->category_display}) {
+            push @{$nh_group->{categories}}, $cat;
+        }
+    }
+}
+
 sub report_new_is_on_he_road {
     my ( $self ) = @_;
 
@@ -171,11 +230,35 @@ sub report_new_is_on_he_road {
         srsname => "urn:ogc:def:crs:EPSG::4326",
         typename => "Highways",
         filter => "<Filter><DWithin><PropertyName>geom</PropertyName><gml:Point><gml:coordinates>$x,$y</gml:coordinates></gml:Point><Distance units='m'>15</Distance></DWithin></Filter>",
+        accept_feature => sub { 1 },
     };
 
     my $ukc = FixMyStreet::Cobrand::UKCouncils->new;
     my $features = $ukc->_fetch_features($cfg, $x, $y);
+    my $nearest = $ukc->_nearest_feature($cfg, $x, $y, $features);
+    # National Highways responsible for litter on Motorways and AM roads
+    # And doesn't matter if we are not on a NH road
+    if ($nearest && $nearest->{properties}->{'ROA_NUMBER'} =~ /^(M|A\d+M)/) {
+        $self->{c}->stash->{report_new_is_on_he_road_not_litter} = 0;
+    } elsif (!scalar @$features) {
+        $self->{c}->stash->{report_new_is_on_he_road_not_litter} = 0;
+    }
     return scalar @$features ? 1 : 0;
+}
+
+sub _report_new_is_on_he_road_not_litter {
+    my ( $self, $x, $y ) = @_;
+
+    my $cfg = {
+        url => "https://tilma.mysociety.org/mapserver/highways",
+        srsname => "urn:ogc:def:crs:EPSG::4326",
+        typename => "highways_litter_pick",
+        filter => "<Filter><DWithin><PropertyName>geom</PropertyName><gml:Point><gml:coordinates>$x,$y</gml:coordinates></gml:Point><Distance units='m'>15</Distance></DWithin></Filter>",
+    };
+    my $ukc = FixMyStreet::Cobrand::UKCouncils->new;
+    my $features = $ukc->_fetch_features($cfg, $x, $y);
+    # If we've matched then litter is their responsibility, so return reverse of that
+    return scalar @$features ? 0 : 1;
 }
 
 sub dashboard_export_problems_add_columns {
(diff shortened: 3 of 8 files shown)


dracos (changes requested):
The inline comments cover a range of technical issues I've found, but more importantly, overall I'm afraid that this doesn't appear to actually work, sorry. The main obvious issues are:
* If you're on the HE cobrand and you click on a non-HE-litter-HE-road (e.g. the green next to spaghetti junction), and you select Litter, the whole form just disappears (this is because as you've picked a non-HE category, it turns off the highways layer, which then confuses it). helitter1
* If you're on .com and you click on a motorway, you incorrectly get the "no details" warning for the council (because it's no longer returning any council categories), and if you then select "Somewhere else" there are no categories to pick at all. helitter1
* If you're on .com and you click an HE non-litter road, and select On the HE road, there's no category to pick from either. If you pick Somewhere else, it shows council litter cats + HE (shouldn't be any HE). 
helitter1

Fundamentally the changes at the server don't match what the client is expecting, so that needs to change at one or both ends.

Currently, on .com, if you click on an HE road, the client gets sent the council and HE categories, with the HE ones all under the HE parent category. Then if you pick HE/not-HE, it shows either just the HE subcategory or hides that HE option. If you click not on an HE road, the server doesn't bother returning the HE categories. Currently, on the HE cobrand, it returns only all the HE categories.

What you want, as I understand it, on .com is:
* Click HE litter road - current behaviour: return all council + HE categories, client decides which to show;
* Click HE non-litter road - need to return categories such that council ones shown if "not HE road" picked by user, HE+council litter shown if "HE road" picked;
* Click not HE road - current behaviour: return all council categories.

On HE cobrand, what you want is:
* Click HE litter road - same as now, HE categories
* Click HE non-litter road - show HE categories + council litter
* Click non HE road - same as now, link to .com

Compared to what you have done, I think you need to keep returning council categories on .com (so they can be shown if "not HE road" picked by user), and somehow work out how to show the right categories depending. Perhaps duplicating the council litter categories in the HE subcategory would work? Not sure.

And on HE cobrand, need to work out how to get the council litter categories to work when selected.

Then this definitely feels like it needs additional Cypress tests to show this is working okay for all the above, and won't regress later.

dracos:
After good discussion this morning:
* HE cobrand - think we agreed there is no need to do anything with categories here. Instead, we need to add the new litter layer client side (either separate or preferably combined with the existing HE layer to add an attribute), and then use that at category selection to, if Litter/Flytipping selected on an HE non-litter road, show the same "go to .com" message currently shown when you click not on HE road or click a DBFO road.
* On .com - we think my proposed solution to (if HE non-litter road) copy the council street cleaning categories into the Highways England group (and remove the HE Litter/Flytipping), should work, if we give them a special marker in their category name (and set their `display_name` to not have the marker) and then the client side can look for that marker to know whether to call only_send/do_not_send when a street cleaning category is chosen, as it already does for HE/non-HE choice.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -118,8 +118,8 @@ sub munge_report_new_bodies {
         my $c = $self->{c};
         my $he = FixMyStreet::Cobrand::HighwaysEngland->new({ c => $c });
         my $on_he_road = $c->stash->{on_he_road} = $he->report_new_is_on_he_road;
-
-        if (!$on_he_road) {
+        my $on_he_road_for_litter = $c->stash->{on_he_road_for_litter} = $he->report_new_is_on_he_road_for_litter;
```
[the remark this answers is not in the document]
dracos: Still two calls, but I have moved it to be on report/new only at least.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -761,8 +759,8 @@ sub stash_category_groups : Private {
             $a->[0] cmp $b->[0];
         } @list;
         @list = map { $_->[1] } @list;
-        my $nh = FixMyStreet::Cobrand::HighwaysEngland->new;
-        $nh->national_highways_cleaning_groups(\@list, $c);
+
+        $c->cobrand->call_hook(munge_mixed_category_groups => \@list);
```
dracos, line 763: We don't want every instance of FixMyStreet to be running this code. Here I have replaced cobrand-specific code in core code with a cobrand hook.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1671,9 +1671,7 @@ sub check_for_category : Private {
     my ( $self, $c, $opts ) = @_;
 
     my $category;
-    if ($c->get_param('category.NationalHighways') && $c->get_param('category.NationalHighways') !~ /\(NH\)/) {
-        $category = $c->get_param('category.NationalHighways');
-    } elsif (!$opts->{with_group}) {
+    if (!$opts->{with_group}) {
```
dracos, line 1674: I did not see what this code was for - I *think* it was because the client-side code ended up incorrectly deselecting the top level "National Highways" item which would then confuse both client and server code. That being fixed, I don't think any change is needed here.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -162,6 +162,12 @@ sub munge_report_new_contacts {
 
 }
 
+sub munge_mixed_category_groups {
+    my ($self, $list) = @_;
+    my $nh = FixMyStreet::Cobrand::HighwaysEngland->new({ c => $self->{c} });
+    $nh->national_highways_cleaning_groups($list);
+}
```
dracos, line 169: This replaces the cobrand-in core code - we need this both here and in UKCouncils to run everywhere it should.

perllib/FixMyStreet/SendReport/Email/Highways.pm
```diff
@@ -30,15 +30,15 @@ sub build_recipient_list {
 sub _lookup_site_code_config { {
     buffer => 15, # metres
     url => "https://tilma.mysociety.org/mapserver/highways",
-    srsname => "urn:ogc:def:crs:EPSG::4326",
+    srsname => "urn:ogc:def:crs:EPSG::27700",
```
dracos, line 33: The tests are okay without changing this, as far as I can see, unaffected tests should not need to be changed, a sign something else must be wrong.

t/cobrand/fixmystreet.t
```diff
@@ -344,6 +344,17 @@ FixMyStreet::override_config {
     };
 };
 
+FixMyStreet::override_config {
```
dracos, line 347: Put back this test removed presumably by mistake.

t/cobrand/fixmystreet.t
```diff
@@ -355,15 +366,45 @@ FixMyStreet::override_config {
         $mech->create_contact_ok(body_id => $hampshire->id, category => 'Trees', email => 'foo@bexley');
         $mech->create_contact_ok(body_id => $he->id, category => 'Litter (NH)', email => 'litter@he', group => 'National Highways');
         $mech->create_contact_ok(body_id => $he->id, category => 'Potholes (NH)', email => 'potholes@he', group => 'National Highways');
-        my $he_mod = Test::MockModule->new('FixMyStreet::Cobrand::HighwaysEngland');
+
+        our $he_mod = Test::MockModule->new('FixMyStreet::Cobrand::UKCouncils');
```
dracos, line 370: Nothing was being mocked, so the live tilma was being queried by this test. Mocked it out with a variety of options

web/cobrands/highways/assets.js
```diff
@@ -55,9 +55,6 @@ fixmystreet.assets.add(defaults, {
     actions: {
         found: function(layer, feature) {
             var category = fixmystreet.reporting.selectedCategory().category;
-            if (category.match('NH')) {
```
dracos: I could not work out what this code was for, or why you would want to deselect when a NH category was being picked.

web/cobrands/highways/assets.js
```diff
@@ -70,7 +67,7 @@ fixmystreet.assets.add(defaults, {
                 // already shown the HE message)
                 if ($('#js-highways:checked').length) {
                     if (category && !category.match('NH')) {
-                        non_he_selected();
+                        he_council_litter_cat_selected();
```
dracos, line 70: We need a new function in this case; using non_he_selected meant the right subcategory was picked but the top-level National Highways was unselected, which confused future category changes (e.g. reportedCategory would be blank next category change and the NH would always toggle). This new function does the same as non_he_selected but maintains the top level selection.

web/cobrands/highwaysengland/assets.js
```diff
@@ -105,11 +105,11 @@ fixmystreet.assets.add(defaults, {
...
         },
         not_found: function(layer) {
-            if (window.fixmystreet.assets.layers[0].selected_feature) {
-                var road_number = window.fixmystreet.assets.layers[0].selected_feature.attributes.ROA_NUMBER;
+            if (fixmystreet.assets.layers[0].selected_feature) {
+                var road_number = fixmystreet.assets.layers[0].selected_feature.attributes.ROA_NUMBER;
                 if ( $("#js-not-he-road").is(":hidden") ) {
                     var category = fixmystreet.reporting.selectedCategory().category;
-                    if ((category === 'Flytipping' || category === "Litter") && (road_number && !road_number.match(/^M/)) ) {
+                    if ((category === 'Flytipping' || category === "Litter") && (road_number && !road_number.match(/^(M|A\d+M)/)) ) {
```
dracos, line 112: Match the server-side check for the road name.

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -285,70 +299,72 @@ sub dashboard_export_problems_add_columns {
 # select distinct category from contacts where category ilike '%litter%' or category ilike '%clean%' or category ilike '%fly%tip%';
 # search to find categories in all contacts and then manually edited
 sub _cleaning_categories { [
-    #Bus Station Cleaning - Windows,
```
dracos: Alphabetised and tidied this list, and added a couple more I found.

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -198,32 +220,24 @@ sub report_new_is_on_he_road {
...
         typename => "Highways",
         filter => "<Filter><DWithin><PropertyName>geom</PropertyName><gml:Point><gml:coordinates>$x,$y</gml:coordinates></gml:Point><Distance units='m'>15</Distance></DWithin></Filter>",
+        accept_feature => sub { 1 },
     };
+
     my $ukc = FixMyStreet::Cobrand::UKCouncils->new;
     my $features = $ukc->_fetch_features($cfg, $x, $y);
-    $cfg->{accept_feature} = sub {
-        my $feature = shift;
-        return $feature->{properties}->{feature_ty} = 'ROAD'
```
dracos: Not sure what this was for, removed.

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -198,32 +220,24 @@ sub report_new_is_on_he_road {
...
-        };
     my $nearest = $ukc->_nearest_feature($cfg, $x, $y, $features);
     # National Highways responsible for litter on Motorways and AM roads
+    # And doesn't matter if we are not on a NH road
     if ($nearest && $nearest->{properties}->{'ROA_NUMBER'} =~ /^(M|A\d+M)/) {
         $self->{c}->stash->{report_new_is_on_he_road_not_litter} = 0;
-    } elsif (scalar @$features) {
-        $self->{c}->stash->{report_new_is_on_he_road_not_litter} = $self->_report_new_is_on_he_road_not_litter;
```
dracos: This was being run on every /around page; I kept most of this the same but moved the setting of this to in national_highways_cleaning_groups so it would only run on /report/new when needed.

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -159,28 +159,50 @@ sub munge_report_new_contacts {
...
-                $clean_name =~ s/ \(NH\)//;
-                $cat_item->set_extra_metadata('display_name', $clean_name);
-            }
-            if ($c->stash->{report_new_is_on_he_road_not_litter}) {
-                @{ $cat->{categories} } = grep {$_->category_display !~ /Flytipping|Litter/} @{ $cat->{categories} };
-            }
-        }
+    my ($self, $category_groups) = @_;
+
+    # We only care if there is a National Highways group
```
dracos, line 164: Return early if we don't actually need to do anything

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -159,28 +159,50 @@ sub munge_report_new_contacts {
...
+    my ($nh_group) = grep { ($_->{name} || '') eq $self->council_name } @$category_groups;
+    return unless $nh_group;
+
+    my $c = $self->{c};
+    my $not_he_litter = $c->stash->{report_new_is_on_he_road_not_litter};
+    if (!defined $not_he_litter) {
```
dracos, line 170: This is where we now fetch the litter data

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -159,28 +159,50 @@ sub munge_report_new_contacts {
...
+        $cat_item->set_extra_metadata('display_name', $clean_name);
     }
+
+    # Don't change anything else unless we're on a HE non-litter road
+    return unless $not_he_litter;
+
+    # NH do not want flytipping or litter reports on these roads, take them out
+    @{$nh_group->{categories}} = grep { $_->category_display !~ /Flytipping|Litter/ } @{$nh_group->{categories}};
+
+    # Put any council street cleaning categories we can find into the NH group,
```
dracos, line 189: This is expanded on what was there, to also cover cleaning subcategories and cleaning category groups.

dracos (commented):
This was good progress on this thorny issue - the NH cobrand looks like it works, and I found it easier to work out the bugs in .com than I probably would have done to write the changes in the first place :) Given the complication of testing it properly, it seemed to be easiest to fix things as I found them, which is what I've done, happy to go through all the comments I've left explaining my changes. The main parts of new code were: a hook rather than cobrand-specific stuff in core code; doing the litter lookup on /report/new only; adding mocks; matching subcategories as well as top-level categories; and fixing the bug in the JS which was confusing category selection. We might finally get there!

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -171,13 +230,37 @@ sub report_new_is_on_he_road {
...
+        $self->{c}->stash->{report_new_is_on_he_road_not_litter} = 0;
+    }
     return scalar @$features ? 1 : 0;
 }
 
+sub _report_new_is_on_he_road_not_litter {
+    my ( $self, $x, $y ) = @_;
+
+    my $cfg = {
+        url => "https://tilma.staging.mysociety.org/mapserver/highways",
```
davea, line 253: Don't forget to make this point to live tilma (with the relevant layers in place) before merging.

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
```diff
@@ -223,4 +306,75 @@ sub dashboard_export_problems_add_columns {
...
+    'Littering',
+    'Littering and cleanliness',
+    'Rubbish or fly tipping on the roads',
+    'Street Cleaning',
+    'Street Cleansing',
+    'Street cleaning and litter',
+    'Street cleaning',
+    'Street cleansing',
+    'Sweeping & Cleansing Hazard',
+    #Bench/cycle rack/litter bin/planter,
```
davea, line 344: Can all these commented out entries be removed?

web/cobrands/highwaysengland/assets.js
```diff
@@ -21,7 +21,7 @@ highways_style.addRules([rule_motorway, rule_a_road]);
 var highways_stylemap = new OpenLayers.StyleMap({ 'default': highways_style });
 
 var defaults = {
-    wfs_url: "https://tilma.mysociety.org/mapserver/highways",
+    wfs_url: "https://tilma.staging.mysociety.org/mapserver/highways",
```
davea, line 24: And here too.

davea (commented):
I *think* this is working OK, but I'd appreciate some more instructions for how to get my local env set up correctly to test so I'm more certain.

I've ended up with a slightly odd situation (URL was https://fms.davea.me/report/new?longitude=-2.350070&latitude=51.500742#subcategory):
image

But before I dig into it, there's every chance I've not got my DB set up properly yet.

davea:
Added the ` (NH)` suffix to the relevant categories and the double category disappeared. Happy for this to be merged once the above comments are dealt with 👍

perllib/FixMyStreet/Cobrand/HighwaysEngland.pm
dracos: I thought it was nice to keep the list in case we have to make any changes in the near future.

PR #3600, opened 2021-09-22
Add feature for always using Reply-To in emails.

dracos:
This means we can choose in configuration to set a cobrand or cobrands to always use Reply-To for the user's details and our own email (matching the SMTP envelope) in the From, regardless of DMARC settings.

[skip changelog] For Freshdesk https://mysocietysupport.freshdesk.com/a/tickets/1472 (whereby a recipient is performing a DMARC check regardless of actual DMARC settings)


chrismytton (approved):
📧

PR #3601, opened 2021-09-22
[Waste] Move bin day check to controller.

dracos:
This only moves some cobrand specific code out so that it applies to all users.
[skip changelog] Fixes https://github.com/mysociety/societyworks/issues/2182


MorayMySoc (approved):
This looks good to me

PR #3605, opened 2021-09-28
[UK] Factor couple of functions for all cobrands.

dracos:
Random bucket of small things - typo in waste, HE want bit of text in skip text, updated docs to fix security warning, and thing I actually wanted, making all cobrands have the do_not_reply_email function so can just use config from now on. [skip changelog]


perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -480,6 +480,16 @@ sub contact_email {
...
 }
 
+sub do_not_reply_email {
+    my $self = shift;
+    return $self->feature('do_not_reply_email') || $self->next::method();
+}
+
+sub verp_email_domain {
+    my $self = shift;
+    return $self->feature('verp_email_domain');
```
chrismytton, line 490: Why does the method above need the extra `$self->next::method()` but this one doesn't?

chrismytton (approved):
Looks good!

perllib/FixMyStreet/Cobrand/UKCouncils.pm
dracos: There is a Default.pm entry for do_not_reply_email, it's not an optional hook used with call_hook, whereas this one has been implemented as a hook so doesn't have a default.

PR #3607, opened 2021-09-28
[Hart] Fix colour accessibility issues

zarino:
We no longer use the Hart primary green colour on a white background, or white on the primary green background, to pass WCAG 2.1.

I also modernised some of the header and nav menu layout logic while I was in the area.

home

map

home-mobile

Menu icon and Home icon on mobile remain white, to match the official site:

Screen Shot 2021-09-28 at 19 59 52


web/cobrands/fixmystreet/images/chevron-black-right.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="43" height="33"><path fill="#000" d="M0 0h13l13 16-13 16H0l13-16"/></svg>
```
dracos, line 1: Was going to say do you want to swap the Westminster cobrand to use this as well, but I see that's off-black.

web/cobrands/hart/layout.scss
```diff
@@ -17,117 +13,127 @@ body.twothirdswidthpage .content {
...
+#postcode-intro {
+    padding: 2em 0;
+}
+
+#front_stats big {
+    color: #000;
+}
+
+.big-green-banner {
+    background: $col_click_map inline-image("../fixmystreet/images/chevron-black-#{$right}.svg") $right center no-repeat;
```
dracos, line 138: Not sure Hart will ever be shown in a RTL language, but guess doesn't hurt :)

web/cobrands/hart/_colours.scss
```diff
@@ -11,21 +11,21 @@ $primary_text: #ffffff;
...
+$nav_background_colour: #fff;
+$nav_hover_background_colour: #fff;
 
 $nav_background_colour: $primary;
-$mappage-header-height: 173px + 32px;
+$mappage-header-height: 205px;
 $header-top-border: false;
 
-$container-max-width: 60em;
+$container-max-width: 80em;
```
dracos, line 28: This change doesn't seem to match the rest of the changes (not about tidying or colour changes, I mean). Just to match their site more?

web/cobrands/hart/_colours.scss
```diff
@@ -11,21 +11,21 @@ $primary_text: #ffffff;
...
-/* Unused here */
-$nav_colour: #fff;
-$nav_hover_background_colour: #444;
+$nav_colour: $link-color;
+$nav_background_colour: #fff;
+$nav_hover_background_colour: #fff;
 
 $nav_background_colour: $primary;
-$mappage-header-height: 173px + 32px;
+$mappage-header-height: 205px;
```
dracos, line 25: Do you think this should be 165+40 instead? Was just trying to show that it was the height of the header + the nav bar, I think.

dracos (approved):
Good to do some tidying :)
Few tiny comments, none needed though.
I think Hart is the only one, but wonder if adding a nav-menu--transparent class or something (or can it be transparent by default?) might be useful at some point.

web/cobrands/hart/_colours.scss
zarino: Oops, yes, sorry. Just to make it a bit less jarring, if you follow a link between the two sites.

PR #3608, opened 2021-09-29
Highways England ➡️ National Highways rebrand

davea:
Renames Highways England to National Highways everywhere in the UI. Requires the body to be renamed when deploying too.

I left filenames alone for the sake of simplicity but happy to rename if that's a better idea.

[skip changelog]

For https://mysocietysupport.freshdesk.com/a/tickets/1500


.cypress/cypress/integration/highwaysengland.js
```diff
@@ -1,4 +1,4 @@
-describe('Highways England cobrand tests', function() {
+describe('National Highwayscobrand tests', function() {
```
dracos, line 1: missing space here and below

dracos (approved):
One spacing typo in test name is all :) Note logo.png has lost a pixel in height, probably doesn't matter!

dracos:
I'm undecided about file names - how long we do think this name will last? Think it's probably okay to stay as is.

davea:
No idea on the lifespan of the name. Let's leave them as they are for now then.

PR #3612, opened 2021-10-06
Block well-behaved robots fromfull-size images

sagepe:
There's no real reason for bots to crawl these so we might as well
save ourselves the bandwidth and cache space where possible.

As robots.txt matches prefixes, this rule should cover any full size
image regardless of its format (the majority are jpeg but some are png).
The simple wildcard matching should be supported by the major search
engine bots that make up the bulk of bot traffic.

[skip changelog]


dracos (approved):
📴

PR #3614, opened 2021-10-11
[Shropshire] Shropshire styles and templates and initial Shropshire.pm file

MorayMySoc:
Adding styles and templates
Adding Shropshire PM file

Please check the following:

[skip changelog]


templates/web/shropshire/index-steps.html
```diff
@@ -0,0 +1,18 @@
+<h2>[% loc('How to report a problem') %]</h2>
+
+<ol class="big-numbers">
+    <li>[% question %]</li>
+    <li>Plot the location on the map</li>
+    <li>[% loc('Enter details of the problem') %]</li>
+    [% IF c.cobrand.is_council %]
```
dracos, line 7: It is.

templates/web/shropshire/index-steps.html
```diff
@@ -0,0 +1,18 @@
...
+        <li>Confirm the report and [% c.cobrand.council_name %] will investigate</li>
+    [% ELSIF c.cobrand.moniker == 'zurich' %]
```
dracos, line 9: It is not.

dracos (changes requested):
A template can be simplified, and the example places can be removed again :)

templates/web/shropshire/index-steps.html
dracos: (so this IF can be removed, I meant)

dracos (changes requested):
One IF can still be removed.

More generally, this PR now doesn't really work because other branches have been merged into it that are nothing to do with it (the PR description no longer matches, for example, I assume I'm only reviewing the fixup but that's not clear). Other branches should have been merged into something like  `shropshire-reviewed`, not this branding branch that is the subject of its own PR (and then this PR should also be merging into shropshire-reviewed, and then the whole thing can be tidied and merged to master when ready).

templates/web/shropshire/index-steps.html
MorayMySoc: This fix is now in master as was merged from another branch

dracos:
As above, this was merged elsewheres.

PR #3618, opened 2021-10-18
[Merton] Enable multi_photos in open311 config

chrismytton:
Merton want to receive multiple photos to their open311 endpoint, so
enable the multiple photos feature for open311 reports.

Part of https://github.com/mysociety/societyworks/issues/2560


ludovic-tc:
:+1:

PR #3623, opened 2021-10-20
[Merton] add USRN and service attributes to Open311 requests

ludovic-tc:
Adds `USRN` and `service` (i.e. user's platform) to Open311 service requests.

Fixes mysociety/societyworks/issues/2673

- [ ] provide test(s)


templates/web/merton/footer_extra_js.html
```diff
@@ -0,0 +1,4 @@
+[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 validation=1 tfl=1 roadworks=1 %]
+[% scripts.push(
```
dracos, line 2: The footer_extra_js_base.html includes /cobrands/merton/assets.js because of the `cobrand_js=1` parameter, so this addition of the same JS file is unnecessary.

chrismytton:
Merged to Merton reviewed.

PR #3636, opened 2021-11-01
[Shropshire] Add assets file and configuration 

MorayMySoc:
* Add asset.js and footer
* Add no quesionairre setting
* Add site config setting
* Add fake site config while data is broken

[skip changelog]


MorayMySoc:
Created in error. Closing

PR #3637, opened 2021-11-01
[Shropshire] Assets information

MorayMySoc:
* Adds asset file and trigger footer
* Adds no_questionnaire setting
* Adds site_config setting

[skip changelog]

### DIFF
--- a/templates/web/shropshire/footer_extra_js.html
+++ b/templates/web/shropshire/footer_extra_js.html
@@ -1,1 +1,1 @@
-[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 validation=1 %]
+[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 highways=1 validation=1 %]
--- a/templates/web/shropshire/footer_extra_js.html
+++ b/templates/web/shropshire/footer_extra_js.html
@@ -1,0 +1,1 @@
+[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 validation=1 %]
--- a/web/cobrands/shropshire/assets.js
+++ b/web/cobrands/shropshire/assets.js
@@ -1,0 +1,189 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+var wfs_host = fixmystreet.staging ? 'tilma.staging.mysociety.org' : 'tilma.mysociety.org';
+var tilma_url = "https://" + wfs_host + "/mapserver/shropshire";
+
+var defaults = {
+    http_options: {
+        url: tilma_url,
+        params: {
+            SERVICE: "WFS",
+            VERSION: "1.1.0",
+            REQUEST: "GetFeature",
+            SRSNAME: "urn:ogc:def:crs:EPSG::27700"
+        }
+    },
+    asset_type: 'spot',
+    asset_id_field: 'central_as',
+    attributes: {
+        central_asset_id: 'central_as',
+        site_code: 'site_code'
+    },
+    geometryName: 'msGeometry',
+    srsName: "EPSG:27700",
+    body: "Shropshire Council",
+    strategy_class: OpenLayers.Strategy.FixMyStreet
+};
+
+var streetlight_stylemap = new OpenLayers.StyleMap({
+    'default': fixmystreet.assets.style_default,
+    'hover': fixmystreet.assets.style_default_hover,
+    'select': fixmystreet.assets.construct_named_select_style("${feature_id}")
+  });
+
+var labeled_defaults = $.extend(true, {}, defaults, {
+    select_action: true,
+    feature_code: 'feature_id',
+    stylemap: streetlight_stylemap,
+    construct_asset_name: function(id) {
+        var code = id.replace(/[O0-9]+[A-Z]*/g, '');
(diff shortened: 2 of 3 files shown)


web/cobrands/shropshire/assets.js
```diff
@@ -0,0 +1,189 @@
...
+    actions: {
+        asset_found: fixmystreet.assets.named_select_action_found,
+        asset_not_found: fixmystreet.assets.named_select_action_not_found
+    }
+});
+
+fixmystreet.assets.add(defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Traffic_Signal_Areas"
```
dracos, line 97: It look me ages to find anything in this layer, not sure if that was me or if it e.g. needs to zoom in more or something else to be more noticeable. Is having them as areas noticeably better than just having them as points?

templates/web/shropshire/footer_extra_js.html
```diff
@@ -0,0 +1 @@
+[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 validation=1 %]
```
dracos, line 1: You need highways=1 if you want the Highways England handling code.

dracos (changes requested):
A number of issues with the assets, some things not working, some suggestions as what needs adding.

web/cobrands/shropshire/assets.js
```diff
@@ -32,23 +32,8 @@ var defaults = {
 var streetlight_stylemap = new OpenLayers.StyleMap({
```
dracos, line 32: This style map isn't used.

web/cobrands/shropshire/assets.js
```diff
@@ -0,0 +1,189 @@
...
+    stylemap: new OpenLayers.StyleMap({
+         'default': highways_style
+     }),
+    always_visible: true,
+    non_interactive: true,
+    road: true,
+    asset_item: 'road',
+    asset_type: 'road',
+    all_categories: true,
+});
```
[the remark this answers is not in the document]
dracos: No, asset_item and asset_type are used in messages such as "Zoom in to pick an asset_item" (and asset_type is used in the CSS class). By attribute, I meant a `usrn:` attribute on the object instead of `road: true` (which is for calling found/not found code on road layers, which you don't have here), see examples in other cobrand files and `updateUSRNField` in fixmystreet/assets.js for how that attribute is used.

web/cobrands/shropshire/assets.js
```diff
@@ -0,0 +1,189 @@
...
+    }
+});
+
+fixmystreet.assets.add(defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Parish_Street_Lights"
+        }
+    },
+    asset_category: ["3+ Consecutive Lights Out", "Hanging Lantern", "Smoking/Sparking equipment",
```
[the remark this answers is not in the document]
MorayMySoc: Not sure how to do this. Will have to look into it.

web/cobrands/shropshire/assets.js
```diff
@@ -0,0 +1,189 @@
...
+    actions: {
+        asset_found: fixmystreet.assets.named_select_action_found,
+        asset_not_found: fixmystreet.assets.named_select_action_not_found
+    }
+});
+
+fixmystreet.assets.add(defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Grit_Bins"
```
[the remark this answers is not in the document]
MorayMySoc: Going to have to request the layer as can't seem to download it

web/cobrands/shropshire/assets.js
```diff
@@ -0,0 +1,161 @@
...
+    fill: false,
+    strokeColor: "#5555FF",
+    strokeOpacity: 0.1,
+    strokeWidth: 7
+});
+
+fixmystreet.assets.add(defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Street_Gazetteer"
```
dracos, line 47: The .map file will need updating to match.

dracos (changes requested):
The USRN layer isn't set up correctly, I'm afraid.
For the union thing I mentioned, you could look at all the "CONNECTIONTYPE UNION" layers in e.g. bucks.map or iow.map which combine multiple data files in one layer for sending out.

templates/web/shropshire/footer_extra_js.html
MorayMySoc: Locally, this is now showing the option of is this a highway, but not showing the categories when 'On the ...' selected. Not sure if that's my setup or I need something else.

web/cobrands/shropshire/assets.js
```diff
@@ -52,16 +47,20 @@ fixmystreet.assets.add(defaults, {
      }),
     always_visible: true,
     non_interactive: true,
+    usrn: {
+        attribute: 'USRN',
+        field: 'site_code'
+    },
     road: true,
```
dracos, line 54: `usrn:` looks right, yep, but you still want to drop this line (although all it's doing is looking for hooks to call that aren't there, or setting body_send which doesn't matter here, I think, but best to be clear you're not using the functionality).

dracos (approved):
You can drop the road variable, but not necessary right now, looks good, hope it works :)

PR #3645, opened 2021-11-05
[Shropshire] Add road layer found/not-found messages

MorayMySoc:
If a report is made on a private road will give a private road message
If a report is made not on a public road will give a not Shropshire Council message

https://github.com/mysociety/societyworks/issues/2663

[skip changelog]

### DIFF
--- a/templates/web/shropshire/report/new/roads_message.html
+++ b/templates/web/shropshire/report/new/roads_message.html
@@ -1,0 +1,10 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <strong>Not maintained by Shropshire Council</strong>
+    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-shropshire">
+        <p>The location you have selected doesn't appear to be on <span class="js-roads-asset" data-original="a road">a public road</span> that Shropshire Council maintain.</p>
+        <p>Please select <span class="js-roads-asset" data-original="a road">a road</span> on which to make a report.</p>
+    </div>
+    <div id="js-not-council-road" class="js-responsibility-message hidden js-roads-shropshire">
+        <p>The road you have selected is a <strong>Private road</strong> and not maintained by Shropshire Council</p>
+    </div>
+</div>
--- a/web/cobrands/shropshire/assets.js
+++ b/web/cobrands/shropshire/assets.js
@@ -45,8 +45,6 @@ fixmystreet.assets.add(defaults, {
     stylemap: new OpenLayers.StyleMap({
          'default': highways_style
      }),
-    always_visible: true,
-    non_interactive: true,
     usrn: {
         attribute: 'USRN',
         field: 'site_code'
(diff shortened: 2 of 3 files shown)


web/cobrands/shropshire/assets.js
```diff
@@ -45,16 +45,35 @@ fixmystreet.assets.add(defaults, {
...
     road: true,
     asset_item: 'road',
     asset_type: 'road',
+    no_asset_msg_id: '#js-not-a-road',
+    always_visible: true,
+    non_interactive: true,
     all_categories: true,
+    actions: {
+        found: function(layer, asset) {
+            fixmystreet.message_controller.road_found(layer, asset.attributes.SITE_CLASS, function(name)
```
dracos, line 61: No biggie, but to be consistent with the rest of this file and the other JS, we put `{` on the same line as the function/if/else statements, and therefore with less indenting.

templates/web/shropshire/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <strong>Not maintained by Shropshire Council</strong>
+    <div id="js-not-a-road" class="hidden js-responsibility-message">
+        <p>The location you have selected doesn't appear to be on <span class="js-roads-asset" data-original="a road">a public road</span> that Shropshire Council maintain.</p>
+        <p>Please select <span class="js-roads-asset" data-original="a road">a road</span> on which to make a report.</p>
+    </div>
+    <div id="js-roads-private" class="js-responsibility-message hidden">
```
dracos, line 7: For this to work on .com, it needs including in the fixmystreet.com equivalent template also. Or you could use `js-not-council-road` instead of `js-roads-private` and then it would use that existing div there, which may be similar enough.

web/cobrands/shropshire/assets.js
```diff
@@ -45,16 +45,35 @@ fixmystreet.assets.add(defaults, {
...
-    always_visible: true,
-    non_interactive: true,
     usrn: {
         attribute: 'USRN',
         field: 'site_code'
     },
     road: true,
     asset_item: 'road',
     asset_type: 'road',
+    no_asset_msg_id: '#js-not-a-road',
```
dracos, line 55: I think (though I'm not entirely sure) you need a no_asset_msgs_class as well - see e.g. equivalent code in Bucks assets.js and roads_message.html (and .com template too) - this is so if you click from a private road to a public road, the message gets correctly removed (otherwise I think it would only remove the "not on a road" message).

dracos (changes requested):
Couple of issues to do with .com and removal of the 'private road' message.

dracos (changes requested):
I assume you'll need to add js-roads-shropshire class to the .com template, think that's all.

dracos (approved):
Looks okay, as long as it's working :)

PR #3649, opened 2021-11-11
[Merton] prevent non-staff users from reopening reports

ludovic-tc:
fixes mysociety/societyworks/issues/2625

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -429,4 +429,11 @@ around 'munge_sendreport_params' => sub {
     $row->areas($original_areas);
 };
 
+sub reopening_disallowed {
+    my ($self, $problem) = @_;
+    my $c = $self->{c};
+    return 1 if $problem->to_body_named("Merton") && $c->user_exists && (!$c->user->from_body || $c->user->from_body->name ne "Merton Council");
+    return $self->next::method($problem);
+}
+
 1;
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -38,4 +38,15 @@ sub open311_config {
 
 sub privacy_policy_url { "https://www.merton.gov.uk/legal/privacy-and-cookies" }
 
+sub reopening_disallowed {
+    my ($self, $problem) = @_;
+    # allow admins to restrict staff from reopening categories using category control
+    return 1 if $self->next::method($problem);
+    # only Merton staff may reopen reports
+    my $c = $self->{c};
+    my $user = $c->user;
+    return 0 if ($c->user_exists && $user->from_body && $user->from_body->cobrand_name eq 'Merton Council');
+    return 1;
+}
+
 1;
--- a/t/cobrand/merton.t
+++ b/t/cobrand/merton.t
@@ -1,0 +1,90 @@
+use Test::MockModule;
+use FixMyStreet::TestMech;
+use HTML::Selector::Element qw(find);
+
+my $mech = FixMyStreet::TestMech->new;
+my $cobrand = Test::MockModule->new('FixMyStreet::Cobrand::Merton');
+
+$cobrand->mock('area_types', sub { [ 'LBO' ] });
+
+my $merton = $mech->create_body_ok(2500, 'Merton Council');
+my @cats = ('Litter', 'Other', 'Potholes', 'Traffic lights');
+for my $contact ( @cats ) {
+    $mech->create_contact_ok(body_id => $merton->id, category => $contact, email => "\L$contact\@merton.example.org");
+}
+
+my $hackney = $mech->create_body_ok(2508, 'Hackney Council');
+for my $contact ( @cats ) {
+    $mech->create_contact_ok(body_id => $hackney->id, category => $contact, email => "\L$contact\@hackney.example.org");
+}
+
+my $superuser = $mech->create_user_ok('superuser@example.com', name => 'Super User', is_superuser => 1);
+my $counciluser = $mech->create_user_ok('counciluser@example.com', name => 'Council User', from_body => $merton);
+my $normaluser = $mech->create_user_ok('normaluser@example.com', name => 'Normal User');
+my $hackneyuser = $mech->create_user_ok('hackneyuser@example.com', name => 'Hackney User', from_body => $hackney);
+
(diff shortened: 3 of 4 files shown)


dracos (changes requested):
This says that reopening is allowed, not disallowed. This would presumably have been spotted if there were a test checking that it did actually disallow reopening on the update form on a report. Also this doesn't prevent reopening of reports on fixmystreet.com, only on the Merton cobrand. So please write a test for showing the reopening field for a fixed report on both Merton and .com that fails, fix this code/make it work for .com as well, and see that your test now passes.

chrismytton (commented):
~~I might be missing something, but can't you just set `updates_allowed` for Merton in the config?~~

```yaml
updates_allowed:
  merton: staff  # Or reporter or whatever
    ...
```

chrismytton:
> I might be missing something, but can't you just set `updates_allowed` for Merton in the config?

I was missing something, this is about reopening reports not updates, ignore me! 🤦‍♂️

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -38,4 +38,13 @@ sub open311_config {
 
 sub privacy_policy_url { "https://www.merton.gov.uk/legal/privacy-and-cookies" }
 
+sub reopening_disallowed {
+    my ($self, $problem, $user) = @_;
+    # allow admins to restrict staff from reopening categories using category control
+    return 1 if $problem->contact && $problem->contact->get_extra_metadata('reopening_disallowed');
+    # only Merton staff may reopen reports
+    return 0 if ($user->from_body && $user->from_body->cobrand_name eq 'Merton Council');
```
dracos, line 46: This will error if $user isn't defined.

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -38,4 +38,13 @@ sub open311_config {
 
 sub privacy_policy_url { "https://www.merton.gov.uk/legal/privacy-and-cookies" }
 
+sub reopening_disallowed {
+    my ($self, $problem, $user) = @_;
```
dracos, line 42: If my main comment is implemented, you don't need to pass $user in/around, as it will be in `$self->{c}->user`.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -553,7 +553,21 @@ has its reopening_disallowed flag set.
...
+
+    # always prefer a cobranded body's own reopening restrictions
+    my $area_cobrand;
+    for my $body (values %{$problem->bodies}) {
+        #last if $area_cobrand = $body->get_cobrand_handler;
+        if ($body->get_cobrand_handler) {
+            $area_cobrand = $body->get_cobrand_handler;
+        }
+    }
+    my $value = $area_cobrand->call_hook('reopening_disallowed', $problem, $user);
```
dracos, line 566: As you can see from the test failures, `$area_cobrand` could be undef at this point and then this will die.

I do see the rationale for doing something like this, but it's not something we've done before in Default.pm; generally if there is different behaviour required on .com as well as a cobrand, it has been put in FixMyStreet.pm. `get_cobrand_handler` requires `council_area_id` which is only used in UK cobrands anyway, so it's probably a bit clearer for any reusers. And then the code in FixMyStreet.pm could use `to_body_named` as some other functions do there:
```
my $c = $self->{c};
return 1 if $problem->to_body_named("Merton") && $c->user_exists && (!$c->user->from_body || $c->user->from_body->name ne "Merton Council");
return super();
```
Or something like that.

dracos (changes requested):
A few suggested moves/simplifications and things to change.

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -38,4 +38,13 @@ sub open311_config {
...
+    # allow admins to restrict staff from reopening categories using category control
+    return 1 if $problem->contact && $problem->contact->get_extra_metadata('reopening_disallowed');
```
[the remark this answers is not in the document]
ludovic-tc: Thanks.

t/cobrand/merton.t
```diff
@@ -0,0 +1,82 @@
...
+my $superuser = $mech->create_user_ok('superuser@example.com', name => 'Super User', is_superuser => 1);
+my $counciluser = $mech->create_user_ok('counciluser@example.com', name => 'Council User', from_body => $body);
+my $normaluser = $mech->create_user_ok('normaluser@example.com', name => 'Normal User');
+
+$normaluser->update({ phone => "+447123456789" });
+
+my ($problem1) = $mech->create_problems_for_body(1, $body->id, 'Title', {
+    postcode => 'SM4 5DX', areas => "2500,", category => 'Potholes', cobrand => 'merton', user => $normaluser
+});
+my ($problem2) = $mech->create_problems_for_body(1, $body->id, 'Title', {
```
dracos, line 24: Problems 2, 3, and 4 aren't used.

t/cobrand/merton.t
```diff
@@ -0,0 +1,82 @@
...
+    postcode => 'SM4 5DX', areas => "2500,", category => 'Litter', cobrand => 'merton',
+    user => $normaluser, extra => {
+        contributed_as => 'another_user',
+        contributed_by => $counciluser->id,
+    }
+});
+
+
+
+$problem1->state('fixed');
```
dracos, line 44: Any reason not to set the state when you create the problem?

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -429,4 +429,11 @@ around 'munge_sendreport_params' => sub {
     $row->areas($original_areas);
 };
 
+sub reopening_disallowed {
+    my ($self, $problem) = @_;
+    my $c = $self->{c};
+    return 1 if $problem->to_body_named("Merton") && $c->user_exists && (!$c->user->from_body || $c->user->from_body->name ne "Merton Council");
+    return super();
```
dracos, line 436: As codecov says, this line isn't covered, and if it were, it would die because super does not exist (eg visiting any non-Merton report on .com). You called `next::method($problem)` fine in the other function.

dracos (changes requested):
One thing from last review, one bug, and some other small suggestions.

I didn't need to see your intermediate commit along the way to fixing the tests :)

dracos (approved):
Looks good :)

PR #3650, opened 2021-11-11
[Shropshire] Fix illegible grey text colour for pagination links

zarino:
Looks like this now:

Screenshot 2021-11-11 at 16 49 58

Makes me wonder whether this bug might also be present on other recent cobrands, where the primary button colour has been changed to somethign dark…


MorayMySoc (commented):
All good, thanks. Merged

PR #3651, opened 2021-11-11
[Buckinghamshire] Fix illegible text colour for pagination links

zarino:
Oh look, here’s [another one](https://github.com/mysociety/fixmystreet/pull/3650)!

Screenshot 2021-11-11 at 17 18 25

Fixes mysociety/societyworks#2709

@lizettal / @ClareA it might be worth scheduling a ticket for the next design sprint, to check through these buttons on every cobrand. I bet there’ll be one or two more that exhibit the same issue.

### DIFF
--- a/web/cobrands/buckinghamshire/base.scss
+++ b/web/cobrands/buckinghamshire/base.scss
@@ -125,6 +125,10 @@ dl dt {
     }
 }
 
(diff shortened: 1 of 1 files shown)


chrismytton (approved):
Thanks @zarino! ✨

PR #3653, opened 2021-11-12
[Shropshire] Override field type to date for Abandoned since field

MorayMySoc:
Fix so 'Abandoned since' field will be a date type field to stop text entry

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Shropshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Shropshire.pm
@@ -52,4 +52,14 @@ sub staff_ignore_form_disable_form {
         && $c->user->belongs_to_body( $self->body->id );
 }
 
+sub open311_contact_meta_override {
+    my ($self, $service, $contact, $meta) = @_;
+    for my $meta_data (@$meta) {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Shropshire.pm
```diff
@@ -52,4 +52,14 @@ sub staff_ignore_form_disable_form {
         && $c->user->belongs_to_body( $self->body->id );
 }
 
+sub open311_contact_meta_override {
+    my ($self, $service, $contact, $meta) = @_;
+    for my $meta_data (@$meta) {
+        if ($$meta_data{'description'} && $$meta_data{'description'} =~ 'Abandoned since') {
```
dracos, line 58: Our convention is `$meta_data->{description}` for clarity (see other `open311_contact_meta_override` examples e.g. Bristol or Greenwich), I don't think `$$` is really used anywhere.

dracos (changes requested):
One convention change, and this could do with a test to show that it does what is required (I think it does, but a test would show it :) ).

dracos (approved):
Looks good

PR #3654, opened 2021-11-12
[Shropshire] Remove text to pick salt bin for new salt bin

MorayMySoc:
Stops user from being asked to pick a salt/grit bin when they are requesting a new salt/grit bin

[skip changelog]


dracos (approved):

MorayMySoc:
Merged into shropshire_reviewed

PR #3655, opened 2021-11-12
[Shropshire] Fix defaults for central_asset_id field and prune fields

MorayMySoc:
'central_asset_id' to populate from correct field in wfs layer

Change asset item from 'grit bin' to 'salt bin' to reflect category name

Removes unused 'feature_id' fields

Removes unused 'central_asset_id' fields

'central_asset_id' on traffic signal asset to populate from correct field in wfs layer.

[skip changelog]

### DIFF
--- a/web/cobrands/shropshire/assets.js
+++ b/web/cobrands/shropshire/assets.js
@@ -18,10 +18,9 @@ var defaults = {
         }
     },
     asset_type: 'spot',
-    asset_id_field: 'central_as',
+    asset_id_field: 'CentralAssetId',
     attributes: {
-        central_asset_id: 'central_as',
-        site_code: 'site_code'
+        central_asset_id: 'CentralAssetId',
     },
     geometryName: 'msGeometry',
     srsName: "EPSG:27700",
@@ -77,11 +76,7 @@ fixmystreet.assets.add(defaults, {
         }
     },
     asset_group: "Streetlights",
-    asset_item: 'streetlight',
-    actions: {
-        asset_found: fixmystreet.assets.named_select_action_found,
-        asset_not_found: fixmystreet.assets.named_select_action_not_found
-    }
+    asset_item: 'streetlight'
 });
 
 fixmystreet.assets.add(defaults, {
@@ -90,13 +85,12 @@ fixmystreet.assets.add(defaults, {
             TYPENAME: "Traffic_Signal_Areas"
         }
     },
-    feature_code: 'FEATURE_ID',
+    asset_id_field: 'ASSET_ID',
+    attributes: {
+        central_asset_id: 'ASSET_ID',
+    },
     asset_group: 'Traffic Signals & Crossings',
-    asset_item: 'traffic signal',
-    actions: {
-        asset_found: fixmystreet.assets.named_select_action_found,
-        asset_not_found: fixmystreet.assets.named_select_action_not_found
-    }
+    asset_item: 'traffic signal'
 });
 
 fixmystreet.assets.add(defaults, {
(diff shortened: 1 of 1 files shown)


web/cobrands/shropshire/assets.js
```diff
@@ -106,7 +106,6 @@ fixmystreet.assets.add(defaults, {
         }
     },
     asset_group: 'Illuminated signs',
-    feature_code: 'CentralAssetId',
     asset_item: 'bollard',
     actions: {
         asset_found: fixmystreet.assets.named_select_action_found,
```
dracos, line 111: If you're not using feature_code, why are you calling the named_select_action_found/not_found code?

web/cobrands/shropshire/assets.js
```diff
@@ -90,7 +88,9 @@ fixmystreet.assets.add(defaults, {
             TYPENAME: "Traffic_Signal_Areas"
         }
     },
-    feature_code: 'FEATURE_ID',
+    attributes: {
+        central_asset_id: 'ASSET_ID',
+    },
     asset_group: 'Traffic Signals & Crossings',
     asset_item: 'traffic signal',
     actions: {
```
dracos, line 96: ditto

web/cobrands/shropshire/assets.js
```diff
@@ -18,10 +18,8 @@ var defaults = {
         }
     },
     asset_type: 'spot',
-    asset_id_field: 'central_as',
```
dracos, line 21: I think you wanted to keep this line, renamed as the attribute was? This is used for knowing if two assets being compared by the code are the same one, for reselecting after layer change or what have you.

web/cobrands/shropshire/assets.js
```diff
@@ -90,7 +88,9 @@ fixmystreet.assets.add(defaults, {
             TYPENAME: "Traffic_Signal_Areas"
         }
     },
-    feature_code: 'FEATURE_ID',
+    attributes: {
+        central_asset_id: 'ASSET_ID',
```
dracos, line 92: This layer *does* have SITE_CODE, so any reason not to use it? (I'm guessing maybe these are not the same site codes as we're expecting, but thought I should check)

dracos (changes requested):
Few small issues, I think.

web/cobrands/shropshire/assets.js
MorayMySoc: I though feature code was for 'construct_selected_asset_message ' and 'asset_item' was for  'named_select_action_found/not_found code'

web/cobrands/shropshire/assets.js
MorayMySoc: Wasn't sure - I thought sticking to the same asset id as provided by the default would be fine, but don't know...do you think I should?

web/cobrands/shropshire/assets.js
dracos: named_select_action_found is to, when an asset is selected, change the message to display the ID of the asset being selected. It calls construct_selected_asset_message, as you can see. asset_item is also used by it (and by other things). If you're not displaying asset IDs when assets are selected, which you don't appear to be, you don't need to call named_select_action_found or similar.

web/cobrands/shropshire/assets.js
dracos: I'd hoped you would have checked my guess about whether they weren't the same site codes or not :) I've done that now, they don't look the same at all, so no, don't use them.

web/cobrands/shropshire/assets.js
```diff
@@ -18,6 +18,7 @@ var defaults = {
         }
     },
     asset_type: 'spot',
+    central_asset_id: 'CentralAssetId',
```
dracos, line 21: the key is asset_id_field...

For completeness, presumably the traffic signal areas layer needs a different asset_id_field to match?

dracos (changes requested):
A typo, I assume, and one suggestion off of it.

dracos (approved):
Woot!

PR #3657, opened 2021-11-15
[Shropshire] Change asset layers to wfs downloaded versions

MorayMySoc:
For LAMP, PLMP, ILUM.

New layers have different keys so switching those out.


dracos (approved):
For clarity I would perhaps have two extensions of defaults, one with ASSET_ID, one with CentralAssetId, but doesn't matter

PR #3658, opened 2021-11-15
Focus style for #key-tools links/buttons

zarino:
Super quick fix, in core, for everyone, but spotted by the team at Shropshire. More noticeable `:focus` style for the `#key-tools` links at the bottom of the map page and report page sidebars.


MorayMySoc (approved):
Just oking this for formality as already in core

PR #3660, opened 2021-11-15
[Shropshire] Set 'Abandoned since field to mandatory

MorayMySoc:
The 'Abandoned since' field requires input or is rejected by confirm.

Request to make 'Registration Mark' mandatory too.


davea (approved):
Codewise this looks fine, though as mentioned on Slack there may be a better solution if Confirm is rejecting empty values for a field it doesn't consider mandatory.

PR #3662, opened 2021-11-16
[Shropshire] Ensure all staff users set a password

davea:
(This has the side effect of requiring users to change their password after ~10 years, but I think that'll be OK.)

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Shropshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Shropshire.pm
@@ -35,6 +35,16 @@ sub disambiguate_location {
     };
 }
 
+sub password_expiry {
+    return if FixMyStreet->test_mode;
+
+    # uncoverable subroutine
+    my $c = shift->{c};
+    return unless $c->user->from_body && $c->user->from_body->name == 'Shropshire Council';
+
+    return 86400 * 365 * 10;
+}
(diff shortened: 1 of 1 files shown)


MorayMySoc (changes requested):
This forces a user to set a password, but it doesn't force them to use it as they can still log in by requesting a link get sent to their email address to log in with. Also, if the user follows the forgotten password link, they can use a compromised email address to log in. I guess if those options were removed, an admin could then send a login email by request and a password could be reset that way?

davea:
I thought about this a bit more and can't think of a sensible way to disable email login for staff that doesn't also disable it for members of the public.

 - Maybe some JS that only reveals the "email me a link to sign in" text when a non-`@shropshire.gov.uk` email is entered?
   - feels quite clunky and surprising, and has accessibility issues
 - Maybe show an error page to users who enter a `@shropshire.gov.uk` address and click the "email me a link to sign in" or "forgotten password" link?
   - Probably would be quite irritating

MorayMySoc:
I suppose the actual content of the emails for 'forgotten password' and 'send link to sign in' could test for Shropshire and just give a policy notice instead of links.

dracos:
I'm going to assume this was decided not to be taken forward

PR #3666, opened 2021-11-17
Add delay before FMS's Open311 `GET service_request_id for a token` call 

### LINKED ISSUE #3663: Wait before `GET service_request_id from a token` is called
## Bug Description
**GET service_request_id from a token** does not

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

ludovic-tc:
fixes #3663

Please check the following:

Thanks for contributing!

### DIFF
--- a/perllib/Open311.pm
+++ b/perllib/Open311.pm
@@ -102,7 +102,25 @@ sub send_service_request {
             if ( my $request_id = $obj->{request}->[0]->{service_request_id} ) {
                 return $request_id unless ref $request_id;
             } elsif ( my $token = $obj->{request}->[0]->{token} ) {
-                return $self->get_service_request_id_from_token( $token );
+
+                my $service_request_id;
+                my $polling_interval = FixMyStreet->config('O311_POLLING_INTERVAL') || 2;
+                my $polling_max_tries = FixMyStreet->config('O311_POLLING_MAX_TRIES') || 5;
+
+                for (1..$polling_max_tries) {
+                    sleep $polling_interval;
+                    $service_request_id = $self->get_service_request_id_from_token( $token );
+                    last if $service_request_id;
+                }
+                if ($service_request_id) {
+                    return $service_request_id;
+                } else {
+                    my $timeout_warning = "Timed out while trying to fetch service_request_id for token $token"
+                        . " and create remote case for FMS problem ID $problem->id.";
+                    $self->error($self->error . $timeout_warning);
+                    warn $timeout_warning unless FixMyStreet->test_mode;
+                    return 0;
+                }
             }
         }
 
--- a/t/open311.t
+++ b/t/open311.t
@@ -126,6 +126,60 @@ subtest 'posting service request with basic_description' => sub {
     is $c->param('description'), $problem->detail, 'description correct';
 };
 
+subtest 'post service request that responds with a token, then fetch SR id using token' => sub {
+    my $extra = {
+        url => 'http://example.com/report/1',
+    };
+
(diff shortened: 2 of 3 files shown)


perllib/Open311.pm
```diff
@@ -102,7 +102,23 @@ sub send_service_request {
             if ( my $request_id = $obj->{request}->[0]->{service_request_id} ) {
                 return $request_id unless ref $request_id;
             } elsif ( my $token = $obj->{request}->[0]->{token} ) {
-                return $self->get_service_request_id_from_token( $token );
+
+                my $service_request_id;
+                my $polling_interval = FixMyStreet->config('O311_POLLING_INTERVAL') || 2;
+                my $polling_timeout = FixMyStreet->config('O311_POLLING_TIMEOUT') || 10;
+
+                while ($polling_timeout) {
```
chrismytton, line 110: Polling timeout will never be 0 if polling interval doesn't divide cleanly into polling timeout. E.g. if polling interval is 3 and polling timeout 10.

Might be better to just retry a fixed number of times, or at a fixed interval, or similar.

perllib/Open311.pm
```diff
@@ -102,7 +102,23 @@ sub send_service_request {
...
+                    sleep $polling_interval;
+                    $polling_timeout -= $polling_interval;
+                    $service_request_id = $self->get_service_request_id_from_token( $token );
+                }
+                if ($service_request_id) {
+                    return $service_request_id;
+                } else {
+                    warn "Timed out while trying to fetch service_request_id for token $token";
+                    warn "and create remote case for FMS problem ID $problem->id.";
```
chrismytton, line 119: I think this information should be added to `$self->error` as well so it's included in the error emails.

t/open311.t
```diff
@@ -126,6 +126,28 @@ subtest 'posting service request with basic_description' => sub {
...
+        $problem,
+        $extra,
+        $problem->category,
+        '<?xml version="1.0" encoding="utf-8"?><service_requests><request><token>369FBB0C-4755-11EC-96C0-F6CCC296753D</token></request></service_requests>',
+        { extended_description => 0 },
+    );
+
+    diag $results->{ res };
+
+    is $results->{ res }, 248, 'got request id';
```
chrismytton, line 144: This won't work because you haven't told the tests what to return for the token response yet.

You'll need to add a `Open311->_inject_response` call to inject a response for the token XML which will contain this request ID.

chrismytton (changes requested):
This looks like it's heading in the right direction! Just needs a few tweaks.

perllib/Open311.pm
ludovic-tc: How about `while ($polling_timeout > 0)`? 
I thought it might save headaches in the future if we could configure the values, but maybe that's premature optimisation!

What I've written is certainly not "the simplest thing that could possibly work", though.

t/open311.t
ludovic-tc: Aye, that's the bit I was working on when I ran out of time.

perllib/Open311.pm
dracos: Also note it doesn't currently stop early if it does get a ID, it keeps retrying for the full number of goes.

chrismytton (approved):
Looks good. As discussed in Slack long-term it would be good to store the token somewhere and have a separate daemon process for fetching the token, but that's something for another ticket.

PR #3669, opened 2021-11-19
[Shropshire] Make Oxfordhshire report archiving script generic to use for Shropshire integration report archiving

MorayMySoc:
This adds the necessary templates for Shropshire to have files closed using automated script

Moves script 'archive-old-enquiries' out of oxfordshire directory as it's a generic script that will run on all cobrands

[skip changelog]

### DIFF
--- a/bin/archive-old-enquiries
+++ b/bin/archive-old-enquiries
@@ -22,7 +22,7 @@ BEGIN {
     use File::Basename qw(dirname);
     use File::Spec;
     my $d = dirname(File::Spec->rel2abs($0));
-    require "$d/../../setenv.pl";
+    require "$d/../setenv.pl";
 }
 
 use FixMyStreet::Script::ArchiveOldEnquiries;
@@ -38,6 +38,7 @@ my ($opts, $usage) = describe_options(
         ['email-cutoff=s',     "Anything before this will be closed with an email sent to the reporter" ],
         ['reports=s',          "csv file with a single column of report ids" ],
     ], required => 1 } ],
+    ['council_name=s', "Name of council for templates", { required => 1 }],
     ['closure-cutoff=s',  "Anything before this will be closed with no email/alert", { required => 1 } ],
     ['closure_file=s',    "path to text file with the message to add to reports" ],
     ['closure_text=s', "text of the message to add to reports" ],
--- a/perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
+++ b/perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
@@ -101,7 +101,9 @@ sub get_closure_message {
         my $file = path($opts->{closure_file});
         chomp(my $message = $file->slurp_utf8);
         return $message;
-
+    } else {
+        my $message = "FixMyStreet is being updated in " . $opts->{council_name} . " to improve how problems get reported.\n\nAs part of this process we are closing all reports made before the update.\n\nAll of your reports will have been received and reviewed by " . $opts->{council_name} . " but, if you believe that this issue has not been resolved, please open a new report on it.\n\nThank you.";
+        return $message;
     }
 }
 
@@ -171,6 +173,7 @@ sub send_email_and_close {
       reports => [@problems],
       report_count => scalar(@problems),
       site_name => $cobrand->moniker,
+      council_name => $opts->{council_name},
       user => $user,
       cobrand => $cobrand,
     );
@@ -179,7 +182,7 @@ sub send_email_and_close {
     printf("    Sending email about %d reports: ", scalar(@problems));
     my $email_error = FixMyStreet::Email::send_cron(
         $problems->result_source->schema,
-        'archive.txt',
+        'archive-old-enquiries.txt',
         \%h,
         {
             To => [ [ $user->email, $user->name ] ],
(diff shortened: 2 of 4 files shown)


templates/email/shropshire/archive.html
```diff
@@ -0,0 +1,55 @@
+[%
```
dracos, line 1: We have six or so of these now, which are quite similar in places, with differences in text. Would be good to consolidate to one main shared template with sub-templates? This one is basically identical to Hounslow/Westminster, looks like.

templates/web/shropshire/archive_on_report.txt
```diff
@@ -0,0 +1,7 @@
+FixMyStreet is being updated in Shropshire to improve how problems get reported.
+
+As part of this process we are closing all reports made more than 3 months before the update.
```
dracos, line 3: It says 3 months here, but just says "before the update" in the archive email, which is correct?

dracos (commented):
Would be good to consolidate these archive templates, I think, and one query about the text.

templates/web/default/archive-old-enquiries.txt
```diff
@@ -0,0 +1,7 @@
+FixMyStreet is being updated in [[Council_name]] to improve how problems get reported.
```
dracos, line 1: We don't have a default web directory, I don't think we want to create one for this. I'd say under `data` would be best place for this, if not just directly in the script itself.

perllib/FixMyStreet/Script/ArchiveOldEnquiries.pm
```diff
@@ -101,7 +103,12 @@ sub get_closure_message {
         my $file = path($opts->{closure_file});
         chomp(my $message = $file->slurp_utf8);
         return $message;
-
+    } else {
+        my $d = dirname(File::Spec->rel2abs($0));
+        my $file = path("$d/../templates/web/default/archive-old-enquiries.txt");
+        chomp(my $message = $file->slurp_utf8);
+        $message =~ s/\[\[Council_name\]\]/$opts->{council_name}/g;
```
dracos, line 110: Seems odd not to use all the standard template handling code for this template! You'd create a `FixMyStreet::Template` object and call process on it, see few other examples in Peterborough or Email.pm

templates/email/default/archive-old-enquiries.txt
```diff
@@ -0,0 +1,35 @@
...
+[% IF council_url %]
+    View report: [% council_url %]/report/[% report.id %]
+[% ELSE %] 
+    View report: [% cobrand.base_url_for_report( report ) %]report/[% report.id %]
+[% END %]
+----
+
+[% END %]
+
+The FixMyStreet team and Shropshire Council
```
dracos, line 34: Still says Shropshire here

templates/email/default/archive-old-enquiries.html
```diff
@@ -26,11 +26,17 @@ <h1 style="[% h1_style %]">Your reports on [% site_name %]</h1>
...
   </p>
   <p style="[% p_style %]">
-    All of your reports will have been received and reviewed by Shropshire, so if
+    All of your reports will have been received and reviewed by [% council_name %], so if
     your report is no longer an issue, you don't need to do anything.
   </p>
   <p style="[% p_style %]">
-    If you believe that the issue has not been resolved you can <a href="https://improvingyourroads.shropshire.gov.uk/">report it again here.</a>
+    If you believe that the issue has not been resolved you can <a href="
+    [% IF council_url %]
```
dracos, line 34: `[%~` here would mean get rid of the whitespace between the opening HTML quote and this opening template tag (and could do same throughout this if statement).

templates/email/default/archive-old-enquiries.txt
```diff
@@ -0,0 +1,35 @@
...
+
+FixMyStreet is being updated in [% council_name %] to improve how problems get reported.
+
+As part of this process we are closing all reports made before the update.
+
+We noticed that you have [% report_count %] old [% nget('report', 'reports', report_count) %] on the system, which we've listed below.
+
+All of your reports will have been received and reviewed by [% council_name %], so if your report is no longer an issue, you don't need to do anything.
+
+[% IF council_url %]
```
dracos, line 13: This is a text email, not HTML. Indenting IF statements will appear indented in the email :)

templates/email/default/archive-old-enquiries.html
```diff
@@ -14,7 +14,7 @@ <h1 style="[% h1_style %]">Your reports on [% site_name %]</h1>
     Hello [% user.name %],
```
dracos, line 14: This is generalised, thanks, but I'd hoped from my previous comment that e.g. the Hounslow/Westminster ones would be removed and any slight differences they had (if any) would be done as included sub-templates. But no matter for now.

bin/archive-old-enquiries
```diff
@@ -38,6 +38,8 @@ my ($opts, $usage) = describe_options(
         ['email-cutoff=s',     "Anything before this will be closed with an email sent to the reporter" ],
         ['reports=s',          "csv file with a single column of report ids" ],
     ], required => 1 } ],
+    ['council_name=s', "Name of council for templates", { required => 1 }],
+    ['council_url=s', "uri for council for templates. Omitting this flag will use cobrand->base_url"],
```
dracos, line 42: Is there ever a time we don't want to use cobrand->base_url/base_url_for_report?

dracos (changes requested):
Afraid still needs a few changes, I think (plus test you've already mentioned).

dracos (approved):

PR #3672, opened 2021-11-23
[Buckinghamshire] Remove custom map bits

chrismytton:
Bucks are looking to switch to OS Maps, so remove the custom map bits so it uses the default map setup. [skip changelog]


dracos (changes requested):
Hooray! Though should also remove web/js/map-wmts-buckinghamshire.js (and then update t/map/tests.t too) :)

dracos (approved):
🍁

chrismytton:
Changed the `default_map_zoom` in cobrand and `max_resolution` in JS, as discussed in Slack, so requesting another review!

dracos:
Looks like zoom change etc has broken the front end test in some way, presumably something is now no longer where it used to be.

dracos (changes requested):
Looks like zoom change etc has broken the front end test in some way, presumably something is now no longer where it used to be.

chrismytton:
I don't fully understand why this test passes on master, but with this change to the maps the road-placement message is triggered, so have to give an answer to that extra step in this test in order for it to pass.

dracos (approved):
"I don't fully understand why this test passes on master" - I guess as the zoom level is now different, the map click (at x/y coordinates) is now clicking in a different place, and before it wasn't actually clicking somewhere that was "Bucks" mocked so was just generic categories? That's all I can think of anyway, so sounds like it might be more accurate now?

PR #3674, opened 2021-11-25
[National Highways] Users can give contact details

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2291 [skip changelog]

### DIFF
--- a/templates/web/base/report/form/user_loggedout_by_email_password.html
+++ b/templates/web/base/report/form/user_loggedout_by_email_password.html
@@ -1,0 +1,10 @@
+<label class="hidden-js hidden-nojs js-new-report-forgotten-shown" for="password_register">[% loc('New password:') %]</label>
+<label for="password_register" class="js-new-report-forgotten-hidden">[% loc('Password (optional)') %]</label>
+[% IF field_errors.password_register %]
+<p class='form-error'>[% field_errors.password_register %]</p>
+[% END %]
+<div class="general-notes js-new-report-forgotten-hidden" id="password_register_hint">
+    <p>[% loc('Providing a password is optional, but doing so will allow you to more easily report problems, leave updates and manage your reports.') %] [% tprintf(loc('Your password should include %d or more characters.'), c.cobrand.password_minimum_length) %]</p>
+</div>
+
+<input class="form-control js-password-validate" type="password" name="password_register" id="password_register" aria-describedby="password_register_hint" value="">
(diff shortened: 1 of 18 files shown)


templates/web/base/report/form/user_loggedout_by_email_password.html
```diff
@@ -0,0 +1,14 @@
+<label class="hidden-js hidden-nojs js-new-report-forgotten-shown" for="password_register">[% loc('New password:') %]</label>
+<label for="password_register" class="js-new-report-forgotten-hidden">[% loc('Password (optional)') %]</label>
+[% IF field_errors.password_register %]
+<p class='form-error'>[% field_errors.password_register %]</p>
+[% END %]
+<div class="general-notes js-new-report-forgotten-hidden" id="password_register_hint">
+  [% IF c.cobrand.moniker == 'bromley' %]
```
chrismytton, line 7: I know this was copied from the other file, but could we potentially have a Bromley-specific version of this template, now that it's a smaller template? Would make the base template a bit cleaner.

templates/web/highwaysengland/report/form/user.html
```diff
@@ -1,6 +1,23 @@
 <!-- report/form/user.html -->
-<div class="js-new-report-user-hidden form-section-preview form-section-preview--next">
-    <button name="report_anonymously" value="yes" class="btn btn--block js-new-report-submit">Report anonymously</button>
-    <small>No personal details will be stored.</small>
+<div class="form-section-preview form-section-preview--next hidden-nojs">
+    <h2 class="form-section-heading form-section-heading--private hidden-nojs">
+        [% loc('Next:') %] [% loc('Tell us about you') %]
+    </h2>
+
+<button type="button" class="btn btn--block hidden-nojs js-reporting-page--next">Continue and provide contact details</button>
```
chrismytton, line 7: Clicking this button works fine for creating reports, but doesn't seem to work when trying to post an update.

I think it's related to the code below. The update form doesn't have the `js-reporting-page` class, so the handler isn't being triggered on the updates page.

https://github.com/mysociety/fixmystreet/blob/2762bd7bbfe8233011d4e9174932d69f2d25fade/web/cobrands/fixmystreet/fixmystreet.js#L1032-L1042

chrismytton (changes requested):

templates/web/highwaysengland/report/form/user.html
dracos: Oh, for some reason I thought there weren't updates on the NH site so didn't check it properly/didn't make sure it had the right bits in. But clearly there are! Will work it out.

templates/web/base/report/form/user_loggedout_by_email_password.html
dracos: Yep, will do

chrismytton (approved):
📛

PR #3678, opened 2021-11-25
Merton reviewed

chrismytton:
This branch is a combination of all the changes that we've done for the Merton cobrand.

Would be good to get a final review of this ahead of go-live.

Relates to https://github.com/mysociety/societyworks/milestone/63

[skip changelog]


chrismytton:
@dracos Ludo and I have just been through the various Merton branches and combined them into this single branch. Would you mind giving it a final sanity check to ensure everything looks sensible. Thanks!

templates/web/merton/footer_extra.html
```diff
@@ -0,0 +1,11 @@
+<div class="societyworks-footer">
+    <div class="container site-footer">
+        <span>
+            Powered by
+            <a class="societyworks-footer__logo societyworks-footer__logo--fms-pro" href="https://www.societyworks.org/services/highways/">FixMyStreet Pro</a>
+        </span>
+        <span>
+            <a href="[%~ cobrand.privacy_policy_url ~%]">Privacy and Cookies</a> | <a href="[%~ cobrand.privacy_policy_url ~%]">Accessibility</a>
```
dracos, line 8: The accessibility link isn't correct, it goes to the privacy policy again. It looked to be okay and then 07e8cbc1f5ca9863e2f234d18293461c22dad185 changed it for some reason?

templates/web/merton/tracking_code.html
```diff
@@ -0,0 +1,7 @@
+<!-- Google Tag Manager -->
```
dracos, line 1: We cannot add this without some form of cookie control/manager to accept/reject these.

templates/web/base/auth/change_password.html
```diff
@@ -38,18 +38,18 @@ <h1>[% title %]</h1>
 [% IF c.user.password %]
     <div class="form-field">
         <label for="current_password">[% loc('Current password:') %]</label>
-        <input class="form-control" type="password" name="current_password" value="[% current_password | html %]">
+        <input id="current_password" class="form-control" type="password" name="current_password" value="[% current_password | html %]">
```
dracos, line 41: The ID fixes in this file should be their own non-Merton commit.

templates/web/merton/footer_extra_js.html
```diff
@@ -0,0 +1,4 @@
+[% PROCESS 'footer_extra_js_base.html' cobrand_js=1 validation=1 tfl=1 roadworks=1 %]
+[% scripts.push(
```
dracos, line 2: The footer_extra_js_base.html includes /cobrands/merton/assets.js because of the cobrand_js=1 parameter, so this addition of the same JS file is unnecessary.

dracos (changes requested):
Looks good, a few small changes is all.

The "Close local reports older than..." commit needs rewording, because it doesn't close them, it is only hiding them on the cobrand. And don't use `@places` in a commit message without the backticks, because GitHub links it to the user with that handle.

In terms of tidying up, I'd say you could squash "Branding and accessibility fixes to comply with style guide", "Email styling", "Fix Local Alerts...", "Prefer Helvetica", and "Additional CSS changes for accessibility" into the first "New cobrand" commit once it's ready, I don't see a need to keep them as separate commits.

web/cobrands/merton/assets.js
```diff
@@ -0,0 +1,44 @@
+(function(){
+
+if (!fixmystreet.maps) {
+    return;
+}
+
+var defaults = {
+    max_resolution: 4.777314267158508,
+    srsName: "EPSG:27700",
+    body: "Merton Borough Council",
```
dracos, line 10: Presumably should be Merton Council

templates/web/merton/tracking_code.html
chrismytton: Apparently this will be handled by the tag manager code, or so we've been told? https://github.com/mysociety/societyworks/issues/2628#issuecomment-934571539

dracos: Oh, fair enough! As long as that works :)

dracos (approved):

ludovic-tc:
also fixes https://github.com/mysociety/societyworks/issues/2616

dracos:
No it doesn't, that is a manual task after go-live to manually close reports older than a year.

PR #3679, opened 2021-11-26
[UK] Fix removal of roadworks message.

dracos:
If the roadworks message had already been shown, but you e.g. change category to something for which the filter returns false, the message would not be removed. Spotted in https://mysocietysupport.freshdesk.com/a/tickets/1558 but nothing to actually do for that ticket.


chrismytton (approved):
🚧

PR #3681, opened 2021-11-29
Replace FMS problem ID with Merton SRQ case reference 

ludovic-tc:
Replaces mention of the numeric FMS problem ID with Merton's 'SRQ' code from their MS Dynamics CRM as the case reference number. 

to fix mysociety/societyworks/issues/2621

Please check the following:

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -807,7 +807,7 @@ sub defect_types {
 #     Note:   this only makes sense when called on a problem that has been sent!
 sub can_display_external_id {
     my $self = shift;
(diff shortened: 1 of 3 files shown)


chrismytton (commented):
@ludovic-tc I've simplified the tests a bit so we're just testing the external_id is displayed, rather than invoking the whole open311 machinery. I've also added a couple of other smoke tests, which I noticed had been added for Shropshire.

@dracos I wrote some of this code, so would you mind reviewing it please?

chrismytton:
Just force-pushed a quick change to make the category in the "External ID is shown on report page" something relevant to Merton, rather than the Shropshire one I copied. Ready for review now!

dracos (approved):
Looks fine.

PR #3682, opened 2021-11-29
Fix email sending with empty attachments array

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2741
[skip changelog] as nothing core uses attachments that wouldn't be working anyway.


chrismytton (approved):
📧

PR #3685, opened 2021-11-29
[Shropshire] parish and part-time lights

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2758 [skip changelog]

image
image
image


davea (changes requested):
There's one little bug where the "Continue" button isn't disabled if the user clicks _near_ a parish light and the snap-to-asset code runs:

image

image


While you're in the area, is it worth putting a `min_resolution` in for the layer too?

image

PR #3686, opened 2021-11-29
Explicit z-indexes to stack mobile menu above map

### LINKED ISSUE #3617: Hamburger button on FMS mobile web has no visible effect after entering location / postcode
**Describe the bug**
After choosing a location, the hamburger button on fixmystreet.com does not display any visible UI elements when it is clicked, other than turning into an 'X' button. 

NB: also affects at least one cobrand (CEC).

**Screenshots**

 

**To Reproduce**
1. Go to the homepage of fixmystreet.com on a mobile device or using browser tools' responsive view. 
2. Click on the 'hamburger' button ( ≡ ). You should see the menu options specified under 'Expected Behaviour' below. Click X to close the menu. 
3. Search for any postcode (e.g. the example postcode).
3. Click on the 'hamburger' button ( ≡ ) again.
4. Now, other than the button turning into an X, nothing else appears on screen. Clicking the X still turns the button back into ≡. 

**Expected behavior**
A menu including (presumably) the following options should appear: 
- Report a problem 
- Sign in
- All reports
- Local alerts
- Help
- Privacy

**Desktop**
 - manifests in browser mobile view tools

   Believed to affect all mobile browsers. Tested on:
 - Device: [e.g. iPhone7
 - OS: iOS 14.6 & 14.7
 - Browser: Safari

zarino:
I am totally at a loss as to when this bug was introduced, or what caused it. The stacking clearly worked fine when the new JavaScript-powered pop-over menu was added 12 months ago (in a01e927), so something must have changed, in core, since then.

I think the absolute positions of everything on the map page have always been relative to the viewport (rather than there being any `position: relative` parents). If so, it can’t be that a positioned parent has been removed, causing the `z-indexes` to no longer be in the right stacking context.

The `z-index: 1` for `.map-fullscreen #map_box` was [set 5 years ago](https://github.com/mysociety/fixmystreet/blame/master/web/cobrands/sass/_base.scss#L2171) in a5ef113, with a comment: `// stack above positioned elements later on the page (eg: .report-list-filters)`.

The `z-index: 1` for `#main-nav` was [set 12 months ago](https://github.com/mysociety/fixmystreet/blame/master/web/cobrands/sass/_base.scss#L784) in 1784886 (part of the same pop-over menu pull request #3270) and it even has an inline comment saying `// So appears over map`!

Anyway, unless someone else can find what changed in the meantime, I’ve done the most obvious thing necessary, which is to define higher z-indexes for the `#js-menu-open-modal` and `#main-nav`.

Fixes #3617

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
         - Do not email inactive body comment users. #3587
         - Look up organizational domain in DMARC checking. #3603
         - Stop slash in category name breaking csv download #3642
+        - Fix CSS z-index bug that resulted in the main menu being hidden behind the map on small screens #3686
     - Admin improvements:
         - Assignees of reports are now visible in admin reports list and report edit pages.
         - Enable per-category hint customisation.
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -773,6 +773,7 @@ html.mobile.js-nav-open #js-menu-open-modal {
   left: 0;
   right: 0;
   background-color: rgba(0, 0, 0, 0.4);
+  z-index: 2; // So appears over map
 }
 
 #main-nav {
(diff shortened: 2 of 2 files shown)


dracos:
In a01e927 there was no menu shown on a map page :) When multi page reporting was added in 1412ead5e57360e133162ec266d0b2681b999cc1 and it was added, I think it was always broken since then.

dracos (approved):

PR #3687, opened 2021-11-29
Fix permissions setting in admin.

dracos:
The upgrade to jQuery 3 broke the fetching of val() on the roles
multiple select input, as the code assumed this returned null if
no role was selected, whereas now it returns an empty array.

I've checked all the other uses of val() and can't see any others on a select multiple.

Fixes https://mysocietysupport.freshdesk.com/a/tickets/1634
[skip changelog] (as upgrade was before last release)


chrismytton (approved):
✅

PR #3695, opened 2021-12-03
Version 3.1.1

dracos:
This is a branch off of the 3.1 tag, including most of the bugfixes and other relevant commits (most importantly, the Mozilla::CA upgrade) from master (at least according to the changelog, so let's hope that is correct and up to date).
Everything included is already on master and should already have been checked at some other point.
This is not for merging, but for tagging as a release assuming tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Releases
 
 * Unreleased
+    - Bugfixes:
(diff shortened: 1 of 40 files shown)


dracos:
This has been done: https://github.com/mysociety/fixmystreet/releases/tag/v3.1.1

PR #3702, opened 2021-12-08
[Merton] save service to extra data to hide spurious service field

ludovic-tc:
Hide 'service' input in inspector toolbar which is an artefact of sending service field via Open311 plus endpoint settings. 

To do this, simply save the service field as an additional, albeit redundant extra data field. 

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -62,10 +62,6 @@ sub reopening_disallowed {
 sub open311_extra_data_include {
     my ($self, $row, $h) = @_;
 
-    my $open311_only = [
-        { name => 'service', value => $row->service },
-    ];
-
     # Reports made via FMS.com or the app probably won't have a USRN
     # value because we don't access the USRN layer on those
     # frontends. Instead we'll look up the closest asset from the WFS
@@ -76,7 +72,18 @@ sub open311_extra_data_include {
         }
     }
 
-    return $open311_only;
+    return [];
+}
+
+sub report_new_munge_before_insert {
+    my ($self, $report) = @_;
+
+    # Save the service attribute into extra data as well as in the
+    # problem to avoid having the field appear as blank and required
+    # in the inspector toolbar for users with 'inspect' permissions.
+    if (!$report->get_extra_field_value('service')) {
+        $report->update_extra_field({ name => 'service', value => $report->service });
+    }
 }
 
 sub lookup_site_code_config { {
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -78,6 +78,11 @@ sub open311_extra_data_include {
 sub report_new_munge_before_insert {
     my ($self, $report) = @_;
 
+    # Workaround for anonymous reports not having a service associated with them.
+    if (!$report->service) {
+        $report->service('unknown');
+    }
+
     # Save the service attribute into extra data as well as in the
     # problem to avoid having the field appear as blank and required
     # in the inspector toolbar for users with 'inspect' permissions.
--- a/t/cobrand/merton.t
+++ b/t/cobrand/merton.t
@@ -191,6 +191,26 @@ FixMyStreet::override_config {
         ok $report, "Found the report";
         is $report->get_extra_field_value("service"), 'desktop', 'origin service recorded in extra data too';
     };
+
+    subtest 'anonymous reports have service "unknown"' => sub {
+        $mech->get_ok('/around');
+        $mech->submit_form_ok( { with_fields => { pc => 'SM4 5DX', } }, "submit location" );
+        $mech->follow_link_ok( { text_regex => qr/skip this step/i, }, "follow 'skip this step' link" );
+        $mech->submit_form_ok(
+            {
+                button => 'report_anonymously',
+                with_fields => {
+                    title => 'Test Report 3',
+                    detail => 'Test report details.',
+                    category => 'Litter',
+                }
+            },
+            "submit report anonymously"
+        );
+        my $report = FixMyStreet::DB->resultset("Problem")->find({ title => 'Test Report 3'});
+        ok $report, "Found the report";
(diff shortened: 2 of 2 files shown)


t/cobrand/merton.t
```diff
@@ -159,4 +160,41 @@ sub test_visit_problem {
...
+        ALLOWED_COBRANDS => 'merton',
+    }, sub {
+        $sender->send($new_report, {
+            easting => 1,
+            northing => 2,
+            url => 'http://example.org/',
+        });
+    };
+
+    $new_report->update;
```
chrismytton, line 194: This persists things in the tests, but not in the actual code, so the service attribute still isn't being persisted. This line needs to go in `perllib/FixMyStreet/Cobrand/Merton.pm` and the tests need a `$new_report->discard_changes` line here to ensure we're retrieving the latest row from the DB, rather than whatever is in memory.

t/cobrand/merton.t
```diff
@@ -159,4 +160,41 @@ sub test_visit_problem {
...
+        cobrand => 'merton', user => $normaluser, state => 'open', service => 'desktop'
+    });
+
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => 'merton',
+    }, sub {
+        $sender->send($new_report, {
+            easting => 1,
+            northing => 2,
+            url => 'http://example.org/',
```
chrismytton, line 190: Tests seem to pass without these three lines, so don't think they're needed.

chrismytton (changes requested):
Doesn't seem to actually work 😬 

I'll pick this up and get it deployed to staging.

chrismytton:
Have moved things around a bit so that the USRN and service attributes are added in `report_new_munge_before_insert`, which means they're available immediately after the report is saved, rather than having to wait until the report is sent over open311.

There's also an issue with anonymous reports, they don't currently get a service attribute because they're not taken into account in this bit of code:

https://github.com/mysociety/fixmystreet/blob/02b09c4505958c9c8d17f78a8390647a18755a98/perllib/FixMyStreet/App/Controller/Report/New.pm#L1099-L1106

So for now I'm just defaulting those to "unknown" for Merton reports, just so there's something in there.

@dracos would you mind checking this over please?

dracos (commented):
Looks okay; only issue is preferably the USRN lookup would only be at sending time rather than during creation flow. You could leave service there, but I think it should have been okay in the extra data include as well, that happens pretty quickly after creation in the normal case. If it wasn't working could it be because service was being added twice - once by the update_extra_field and once by the setting of open311_only which gets pushed onto extra? Not sure what would then get pushed - two values? No need for the open311_only if the update_extra_field code is there instead, I mean, that might fix it a different way.

Yes, service isn't spotted for anonymous reports, is it - would need two submit buttons there like normal report form has, to do the same thing.

chrismytton:
@dracos The service attribute was being sent over open311, as expected, but it wasn't being persisted into the extra data for the problem, which is what this PR was trying to address. I believe it wasn't being persisted because `open311_extra_data_include` doesn't persist the extra data that's added in there.

I've moved the USRN bits back to `open311_extra_data_include` and left the service bits in `report_new_munge_before_insert`, which will hopefully have the desired effect.

dracos:
"I believe it wasn't being persisted because open311_extra_data_include doesn't persist the extra data that's added in there." - it does do that, though? That's why open311_extra_data_include has two ways of doing things - you return data you want including on extra only for the sending of open311 (hence the variable being called open311_only here and elsewhere), but if you actually edit the row's extra in the function, that is persisted.

Must be a test showing this, can't find one offhand, so edited bexley.t to print out get_extra_fields before/after calling send() in the `NSGRef and correct email config` test and e.g.
```
# Subtest: NSGRef and correct email config
$VAR1 = [];
    ok 1 - Correct latitude
    ok 2
    ok 3
$VAR1 = [
          {
            'description' => 'NSG Ref',
            'value' => 'Road ID',
            'name' => 'NSGRef'
          }
        ];
    ok 4 - checking for 1 email(s) in the queue
    ok 5
    ok 6
    ok 7 - Found text body
    ok 8
    ok 9 - cleared email queue
    1..9
ok 21 - NSGRef and correct email config
```

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -62,21 +62,33 @@ sub reopening_disallowed {
...
-    ];
-
     # Reports made via FMS.com or the app probably won't have a USRN
     # value because we don't access the USRN layer on those
     # frontends. Instead we'll look up the closest asset from the WFS
     # service at the point we're sending the report over Open311.
     if (!$row->get_extra_field_value('usrn')) {
         if (my $usrn = $self->lookup_site_code($row, 'usrn')) {
             $row->update_extra_field({ name => 'usrn', value => $usrn });
+            $row->update;
```
dracos, line 72: I guess it doesn't matter, as we haven't done anything else at this point, but this line should not be necessary, see my other comment. Writing a test would be ideal, doesn't look like this function is tested at all.

dracos (commented):
Is fine, but I don't think the update row should be needed if we're happy to take that out (and a test would be great).

chrismytton:
Erk, think there's something up with my local setup then, when I modify that Bexley test I get this:

```
# Subtest: NSGRef and correct email config
$VAR1 = [];
    ok 1 - Correct latitude
    ok 2
    ok 3
$VAR1 = [];
    ok 4 - checking for 1 email(s) in the queue
    ok 5
    ok 6
    ok 7 - Found text body
    ok 8
    ok 9 - cleared email queue
    1..9
ok 21 - NSGRef and correct email config
```

Will see if I can figure out what's going on.

chrismytton:
Ah I hadn't called `$report->discard_changes`, it's actually working as expected! Will add a test for the Merton bits and remove the `$row->update` line.

chrismytton:
@dracos Have remove the unnecessary `$row->update` line and added a test for the USRN bits. Would you mind having another look please?

dracos (approved):
Looks good, thanks :)

chrismytton (approved):

chrismytton:
(Approving because I had initially requested changes!)

PR #3704, opened 2021-12-10
[Merton] drop FMS contact form from privacy page

ludovic-tc:
- Removed FMS contact form _link_ from privacy policy page.
- Set Merton cobrand to only allow contact form for abuse reports.

fixes mysociety/societyworks/issues/2629

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -102,4 +102,6 @@ sub lookup_site_code_config { {
 
 sub cut_off_date { '2020-12-06' } # 1 yr prior to FMS Pro go-live
 
+sub abuse_reports_only { 1 }
+
 1;
--- a/templates/web/fixmystreet-uk-councils/about/_sidebar.html
+++ b/templates/web/fixmystreet-uk-councils/about/_sidebar.html
@@ -1,0 +1,14 @@
+<div class="sticky-sidebar">
+    <aside>
+        <ul class="plain-list">
+            <li>[% INCLUDE link h='/faq' t=loc('Frequently Asked Questions') %]</li>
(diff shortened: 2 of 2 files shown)


templates/web/fixmystreet-uk-councils/about/_sidebar.html
```diff
@@ -0,0 +1,16 @@
+<div class="sticky-sidebar">
+    <aside>
+        <ul class="plain-list">
+            <li>[% INCLUDE link h='/faq' t=loc('Frequently Asked Questions') %]</li>
+            <li>[% INCLUDE link h='/about/posters' t='FixMyStreet Goodies' %]</li>
+            <li>[% INCLUDE link h='/about/house-rules' t='House Rules' %]</li>
```
chrismytton, line 6: We don't want these two lines in council cobrands. I think it's best to copy `templates/web/base/about/_sidebar.html`, which was the template that was being used, and remove the contact link. It looks like you've copied `templates/web/fixmystreet.com/about/_sidebar.html`, which is just for fixmystreet.com.

chrismytton (changes requested):

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -102,4 +102,8 @@ sub lookup_site_code_config { {
 
 sub cut_off_date { '2020-12-06' } # 1 yr prior to FMS Pro go-live
 
+sub abuse_reports_only { 1 }
+
+sub privacy_policy_url { '/about/privacy' }
```
chrismytton, line 107: Might as well move this change into `Default.pm` so that the fix applies to all cobrands.

chrismytton (changes requested):
This now seems to have the commit from https://github.com/mysociety/fixmystreet/pull/3706 in it? These are two independent changes, they don't have any dependency on each other, so I think it makes sense to keep the commits separate.

chrismytton (approved):

PR #3706, opened 2021-12-10
Fix link deactivation for privacy policy link on privacy policy page.

### LINKED ISSUE #3705: Fix privacy policy link deactivation on privacy policy page for most cobrands.
On most FixMyStreet cobrands (including fixmystreet.com), the privacy policy  link does not deactivate when you are on the actual privacy policy page (e.g. https://www.fixmystreet.com/about/privacy). 

Deactivating the link appropriately makes it clearer to the user that they are on the actual privacy policy page already and matches the presentation of the other sidebar links. 

The bug occurs because these cobrands inherit the `privacy_policy_url` method inherited from `FixMyStreet::Cobrand::Default`. This returns `/privacy`, which redirects to the canonical page URL `/about/privacy`. Because the link does not then match the page URL, the link deactivation code does not work.

This bug affects all cobrands that do not override `privacy_policy_url()`. 

Screenshot 2021-12-10 at 15 18 33
Screenshot 2021-12-10 at 15 18 43

ludovic-tc:
fixes #3705

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Default.pm
+++ b/perllib/FixMyStreet/Cobrand/Default.pm
@@ -1360,7 +1360,7 @@ The URL of the privacy policy to use on the report and update submissions forms.
 
 =cut
 
-sub privacy_policy_url { '/privacy' }
+sub privacy_policy_url { '/about/privacy' }
 
 =item emergency_message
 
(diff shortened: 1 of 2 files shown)


chrismytton (approved):
🔗

PR #3708, opened 2021-12-13
[Merton] Run before_insert hook on .com

chrismytton:
This ensures that reports have a service and that the service is persisted to extra data.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -78,6 +78,8 @@ sub open311_extra_data_include {
 sub report_new_munge_before_insert {
     my ($self, $report) = @_;
 
+    return unless $report->to_body_named('Merton');
+
     # Workaround for anonymous reports not having a service associated with them.
     if (!$report->service) {
         $report->service('unknown');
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
We don't want to stick a service extra data field on every single report - this will cause issues for other sites where e.g. open311-adapter starts rejecting reports with an unknown attribute being provided. The Bucks code checks the report is going to Bucks, for example, so something similar would need to be done here. (And should have a test.)

chrismytton:
> We don't want to stick a service extra data field on every single report

Oops, yes should have spotted that. I've added a guard clause like Bucks in e1af643da25fc910daa85dd603a3b6967c28509e.

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -78,6 +78,8 @@ sub open311_extra_data_include {
 sub report_new_munge_before_insert {
     my ($self, $report) = @_;
 
+    return unless $self->{c}->stash->{report}->to_body_named('Merton');
```
dracos, line 81: I realise this is the same as Bucks, but I don't understand why Bucks isn't using `$report` that has been passed in!

dracos (approved):
Is fine (though a test would be good after), but one comment which I think applies to the Bucks code you're reusing here as well.

dracos (changes requested):
I see lots of test failures I'm afraid, though.

dracos (approved):
Oh, dur, that was on just the first commit - assuming they pass with the fix... :)

perllib/FixMyStreet/Cobrand/Merton.pm
chrismytton: Oh yes, that would be a bit simpler wouldn't it!

PR #3709, opened 2021-12-13
Highlight pin on sidebar focus as well as hover.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2786


zarino (approved):
Works! 🎉 Solution was a lot simpler than I expected. Well done!

PR #3712, opened 2021-12-14
Default make_css to web/cobrands rather than web.

dracos:
This would fix an issue where if you have, say, a multi-gigabyte photo cache at web/photos, it won't get searched for SCSS files.


chrismytton (approved):
🎨

PR #3716, opened 2021-12-14
[Northampton] Increase default zoom

MorayMySoc:
Northampton to check a new default zoom of 9

https://mysocietysupport.freshdesk.com/a/tickets/1507

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Map/Northamptonshire.pm
+++ b/perllib/FixMyStreet/Map/Northamptonshire.pm
@@ -3,8 +3,6 @@ use base 'FixMyStreet::Map::WMSBase';
 
 use strict;
 
-sub default_zoom { 8; }
-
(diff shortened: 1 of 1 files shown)


dracos (approved):
👍

dracos (changes requested):
Just to record conversation elsewhere - this doesn't turn out to be used for default zoom, sorry for not realising before, but need to edit Map/Northamptonshire.pm instead. This would actually I think re-fix the tests (hopefully!) as they then wouldn't depend upon this changed value.

dracos (approved):
I think we can also remove the "sub default_zoom { 8; }" line that I don't think is used anywhere by this map type at all (looks like it's used by the base display_map but WMXBase overrides that and no longer uses it).

PR #3717, opened 2021-12-14
[Merton] Display logo at top of map page

zarino:
Fixes mysociety/societyworks#2791. [skip changelog]

I also upgraded to a crisper version of the Merton logo (as found on the merton.gov.uk homepage) while I was at it.

Screenshot 2021-12-14 at 16-57-01 Viewing a location


dracos (changes requested):
There's still no logo on a mobile /around page, I think? Having both "banners" does take up more space. But presumably it should at least be consistent, then can change it all later if need be.

zarino:
> There's still no logo on a mobile /around page, I think? Having both "banners" does take up more space. But presumably it should at least be consistent, then can change it all later if need be.

Yes, I noticed this, but felt it was probably better _not_ to have a double-height header on the mobile `/around` page, because it’d eat too much vertical space. I guess the alternative is to replace the "FixMyStreet" wording with the Merton logo, but it’ll either need to be a dark version of the logo, or we’ll need to make the menu bar dark. 🤔

PR #3719, opened 2021-12-16
[Oxfordshire] Fix vertical scroll bar being shown.

dracos:
[skip changelog]


zarino (approved):
👍 Solves the problem, from what I can tell. Haven’t spotted any side-effects. As you say, bit suspicious as to why we set `overflow: auto` on `#main-nav` in the first place. 🤔

PR #3722, opened 2021-12-16
[Merton] add Google cookies info table to privacy page

ludovic-tc:
### fixes mysociety/societyworks#2628

[skip changelog]

### DIFF
--- a/templates/web/fixmystreet-uk-councils/about/privacy.html
+++ b/templates/web/fixmystreet-uk-councils/about/privacy.html
@@ -489,4 +489,6 @@ can use.
     </tr>
 </table>
 
+[% TRY %][% INCLUDE 'about/googlecookies.html' %][% CATCH file %][% END %]
+
 [% INCLUDE 'footer.html' pagefooter = 'yes' %]
--- a/web/cobrands/merton/base.scss
+++ b/web/cobrands/merton/base.scss
@@ -280,3 +280,11 @@ ol.big-numbers {
 }
 /* end of fixes for around page */
 
+table.googlecookies {
+    /* make all columns equal width */
+    /* otherwise, looks messy in comparison to first cookie table */
+    th, td {
+        width: 1%;
+    }
+}
+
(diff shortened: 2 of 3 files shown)


web/cobrands/merton/base.scss
```diff
@@ -277,3 +277,9 @@ ol.big-numbers {
 }
 /* end of fixes for around page */
 
+table.googlecookies {
+    th, td {
+        width: 1%;
```
davea, line 282: I think I understand what this does but not why/how - could you add a comment explaining it please?

templates/web/fixmystreet-uk-councils/about/googlecookies.html
```diff
@@ -0,0 +1,40 @@
+<h3>Google Analytics Cookies</h3>
```
davea, line 1: Although there's nothing Merton-specific in this particular file, as it's only used by Merton currently I think it might be better to put it in the `merton` cobrand dir, rather than `fixmystreet-uk-councils`. Easy enough to move it back in the future if another council wants to display the same table.

davea (changes requested):
Couple of minor things, nearly there though!

davea (approved):

templates/web/fixmystreet-uk-councils/about/privacy.html
```diff
@@ -476,4 +476,6 @@ <h2 id="cookies">Cookies</h2>
     </tr>
 </table>
 
+[% TRY %][% INCLUDE 'about/googlecookies.html' %][% CATCH file %][% END %]
```
chrismytton, line 479: Don't think you need the `[% END %]` at the end of this line.

davea: It is required, it closes the `TRY`.

chrismytton: Ah so it is, my bad!

ludovic-tc:
Updated text as advised by Merton.

davea (approved):
Looks good. Please squash down to a single commit before merging.

PR #3725, opened 2021-12-20
[Shropshire] Fix missing slash in emails, and improve templates, testing

MorayMySoc:
Fix missing slash in emails, and improve templates add testing to archiving old reports script

Adds slash that was missing between base url and 'report' which broke urls in emails

Changes templates to use cobrand variables throughout and remove passing in council name

Adds tests to check urls and templating

[skip changelog]


davea (approved):
Looks good! Don't forget to click the "checked changes" button in that failing cobrand-templates check.

PR #3727, opened 2021-12-21
Remove custom styling for /around pagination links

zarino:
Seminatically, these are links not buttons, so they should look like links, not buttons.

This also avoids the colour contrast issues that we previously had to work around in commits like 5ef5710e06042b4d81207b5d69905096f8a9d89f and 8c595e720763dbcd19228db9921deafb73dc8042.

While I was at it, I also changed the `<p>` to a `<nav>`, to give assistive devices a hint that the links relate to navigation, and an `aria-label` that explains "Next" and "Previous" relate to pagination.

Here’s an example of how it looks, on Merton cobrand:

Screenshot 2021-12-21 at 17 27 55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Unreleased
     - Front end improvements:
         - Highlight pin on sidebar focus as well as hover.
+        - Map page pagination links now styled as links rather than buttons. #3727
     - Bugfixes:
         - Add ID attributes to change password form inputs.
         - Fix link deactivation for privacy policy link on privacy policy page. #3704
--- a/templates/web/base/pagination.html
+++ b/templates/web/base/pagination.html
@@ -1,5 +1,5 @@
 [% IF pager AND pager.total_entries > 1 %]
-    <p class="pagination" data-page="[% pager.current_page | html %]">
+    <nav aria-label="Pagination" class="pagination" data-page="[% pager.current_page | html %]">
         [% IF pager.previous_page %]
             <a class="prev" href="[% c.uri_with({ $param => pager.previous_page, ajax => undefined }) %][% '#' _ hash IF hash %]">[% loc('Previous') %]</a>
         [% END %]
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Unreleased
     - Front end improvements:
         - Highlight pin on sidebar focus as well as hover.
+        - /around pagination links now styled as links rather than buttons.
```
davea, line 6: Technically this also affects `/reports` and `/my` too.

davea (changes requested):
Something about this has broken the links I'm afraid. Clicking "next" or "previous" doesn't actually move to the next/prev page. *Something* is happening, as the loading spinner does briefly appear over the map, but the pins/sidebar don't change. It looks like the request to the backend hasn't changed the `p` parameter, but I'm afraid I've not figured out the underlying cause.

Should this text be centre-aligned or is the move to the left deliberate?

Before:
image

After:
image

CHANGELOG.md
zarino: Good point! Will amend the commit message and changlog entry.

zarino:
> Should this text be centre-aligned or is the move to the left deliberate?

Damn, somehow managed to commit the styling for `.pagination > span` without adding the `<span>` to the HTML template. Have pushed a fixup commit now.

Screenshot 2021-12-23 at 08 31 03

Still, it’ll never be perfectly centred, because I switched to a flex layout rather than absolute positioning. It seemed more conceptually pure, and potentially more robust against translations with very long words. But now I write those words, I’m not sure it was really worth it. Hmm… will have a think.

zarino:
Yup, couldn’t live with myself. Have reverted back to absolute positioning. Simpler commit now too (once squashed!)

Screenshot 2021-12-23 at 08 38 45

One day, flex, your time will come. One day.

davea (approved):
🎅 👍

PR #3728, opened 2021-12-21
[Stevenage] Remove left-over Stevenage header_extra template

zarino:
Oddly, despite @dracos removing lots of files from `/templates/web/stevenage` in c21f766dee71bdce2e1823fc8af2840ef34c5250, I seemed to still have this one hanging around. We’re not keeping it for a reason, are we?

[skip changelog]

### DIFF
--- a/templates/web/stevenage/header_extra.html
+++ b/templates/web/stevenage/header_extra.html
@@ -1,1 +1,0 @@
-[% INCLUDE 'tracking_code.html' %]


davea (approved):
Looks like it came back in https://github.com/mysociety/fixmystreet/pull/3270 which was committed and then merged either side of https://github.com/mysociety/fixmystreet/commit/c21f766dee71bdce2e1823fc8af2840ef34c5250.

No reason to keep it that I can see!

PR #3729, opened 2021-12-21
[Westminster] Correct "Click map" font size on /around page

zarino:
Somehow a recent commit applied the same massive `font-size: 3.3em` to the map page headings, as to the front page heading. This was particularly unfortunate on the "Click the map…" banner, which is actually a `h1`.

Screenshot 2021-12-21 at 17 54 52

This commit resets `.mappage` headings (including the "Click the map…" banner, and the "Report your problem" heading at the top of the reporting form) to their default font-sizes.

Screenshot 2021-12-21 at 17 53 22

Fixes mysociety/societyworks#2557.

[skip changelog]

### DIFF
--- a/web/cobrands/westminster/base.scss
+++ b/web/cobrands/westminster/base.scss
@@ -8,10 +8,7 @@
 
 body.frontpage {
     background-color: #fff; // instead of $westminster_grey
-}
 
-body.frontpage,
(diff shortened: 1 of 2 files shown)


davea (approved):
# THAT'S BETTER

PR #3730, opened 2022-01-04
[Bromley] Set payment method to DD for GGW renewals

davea:
This fixes an issue where a report is duplicated once per day if the
original GGW report had a problem with payment.

See https://mysocietysupport.freshdesk.com/a/tickets/1704 for an example

Fixes https://github.com/mysociety/societyworks/issues/2845

[skip changelog]


dracos (approved):
I think this is fine to do, but I see line 1456 saying "if it's a renewal of a DD where the initial setup was as a renewal, so perhaps this code is only expecting to find New reports? I don't really know :-/ Probably worth @struan having a look when he's back, plus a think is there any way this can be simplified so it's easier to work out what's going on.

davea:
Thanks for the review. Agree that it'd be good to get @struan's input before deploying this just in case there's something I've missed.

PR #3731, opened 2022-01-04
[Merton] character limit on extra question fields

ludovic-tc:
fixes mysociety/societyworks#2823

Please check the following:

[skip changelog]

### DIFF
--- a/.cypress/cypress/integration/merton.js
+++ b/.cypress/cypress/integration/merton.js
@@ -1,0 +1,31 @@
+it('loads the Merton FMS Pro front page', function() {
+    cy.visit('http://merton.localhost:3001/');
+    cy.contains('Merton Council');
+});
+
+describe('additional question fields maximum length', function() {
+  beforeEach(function() {
+    cy.server();
+    cy.route('/report/new/ajax*').as('report-ajax');
+    cy.visit('http://merton.localhost:3001/report/new?latitude=51.400975&longitude=-0.19655');
+    cy.wait('@report-ajax');
+    cy.pickCategory('Flytipping');
+    cy.nextPageReporting();
+    cy.get('input#form_evidence').as('evidence');
+  });
+  it('does not allow typing >255 chars', function() {
+    cy.get('@evidence').type('a'.repeat(260));
+    cy.get('@evidence').invoke('val').then(function(val){
+        expect(val.length).to.be.at.most(255);
+    });
+    cy.nextPageReporting();
+});
+it('does not allow user to continue if value is >255', function() {
+    cy.get('@evidence').invoke('val', 'a'.repeat(260));
+    cy.get('@evidence').invoke('val').then(function(val){
+        expect(val.length).to.equal(260);
+    });
+    cy.nextPageReporting();
+    cy.contains("Please enter no more than 255 characters.");
+});
+});
--- a/bin/browser-tests
+++ b/bin/browser-tests
@@ -22,6 +22,7 @@ BEGIN {
         hounslow
         isleofwight
         lincolnshire
+        merton
         northamptonshire
         oxfordshire
         peterborough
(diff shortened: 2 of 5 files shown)


ludovic-tc:
- [ ] fix for Merton cobrand
  - [ ] fix for fly-tipping evidence
  - [ ] fix for all additional question fields
- [ ] fix for fixmystreet.com

davea (changes requested):
Works as it should, but please add a test. I don't think we've done custom validation on extra fields before so it'd be good to cover it.

.cypress/cypress/integration/merton.js
```diff
@@ -0,0 +1,13 @@
...
+    cy.contains('Merton Council');
+});
+it('does not allow >255 chars in additional question fields', function() {
+    cy.server();
+    cy.visit('http://merton.localhost:3001/report/new?longitude=-0.199218&latitude=51.405454');
+    cy.get('input[value="Flytipping"]').click();
+    cy.get('button').first().click();
+    var longstring = '';
+    while (longstring.length < 260) { longstring += 'a'; }
+    cy.get('textarea#form_evidence').type(longstring);
```
davea, line 12: This could just be `cy.get('textarea#form_evidence').type("a".repeat(260));` :)

davea (changes requested):
Nearly there, just a couple of small things to fix to get Cypress green.

.cypress/cypress/integration/merton.js
ludovic-tc: Hi Dave, thanks. I've incorporated this while also reusing this code for validation. But makes fewer LOCs and less repetition than before I saw this.

I also spotted that I'd added some redundant code. (The extra question that I got stuck on above, I'd actually already added separately in a way that worked! This is the trouble with putting stuff to one side for weeks.)

.cypress/cypress/integration/merton.js
```diff
@@ -0,0 +1,24 @@
...
+    var longstring260 = 'a'.repeat(260);
+    cy.get('input#form_evidence').as('evidence');
+    cy.get('@evidence').type('a'.repeat(260));
+    cy.get('@evidence').invoke('val').then(function(val){
+        expect(val.length).to.be.at.most(255);
+    });
+    cy.get('@evidence').invoke('val', longstring260);
+    cy.get('@evidence').invoke('val').then(function(val){
+        expect(val.length).to.equal(260);
+    });
```
davea, line 19: I think these lines would make more logical sense in the next test (starting on L21) as they're setting up what's being tested there.

davea (changes requested):
One very tiny tweak is all!

.cypress/cypress/integration/merton.js
ludovic-tc: ✅

.cypress/cypress/integration/merton.js
```diff
@@ -0,0 +1,24 @@
...
+    cy.get('button').first().click();
+    var longstring260 = 'a'.repeat(260);
+    cy.get('input#form_evidence').as('evidence');
+    cy.get('@evidence').type('a'.repeat(260));
+    cy.get('@evidence').invoke('val').then(function(val){
+        expect(val.length).to.be.at.most(255);
+    });
+});
+it('does not allow user to continue if additional question field value is >255', function() {
+    cy.get('@evidence').invoke('val', longstring260);
```
davea, line 18: This is failing because `longstring260` isn't in scope here.

davea (changes requested):
Don't forget to run the tests locally before submitting for review!

.cypress/cypress/integration/merton.js
ludovic-tc: Thanks Dave, sorry, I usually do do this, have added another reminder! 🤦🏼‍♂️

.cypress/cypress/integration/merton.js
```diff
@@ -0,0 +1,28 @@
+it('loads the Merton FMS Pro front page', function() {
+    cy.visit('http://merton.localhost:3001/');
+    cy.contains('Merton Council');
+});
+it('does not allow typing >255 chars in additional question fields', function() {
+    //cy.server();
```
davea, line 6: If this line isn't required can it be removed instead of left commented out?

davea (changes requested):

.cypress/cypress/integration/merton.js
ludovic-tc: 🤦🏼

davea (approved):

davea (approved):

PR #3734, opened 2022-01-05
Include username in inactive email.

dracos:
This has come up in user support (someone had two accounts, one forwarded to the other, they didn't realise) and I see client support (domain name change causing the same issue), so including the email (or phone number) in the body of the inactive email seems like a sensible thing to do at the least.


MorayMySoc (approved):
Looks good to go

PR #3739, opened 2022-01-07
Mark a wait-skipped update as a failure.

dracos:
This should prevent the daemon from constantly retrying these updates. [skip changelog]


davea (approved):
Nice!

PR #3740, opened 2022-01-12
[Bromley] Fix GGW DD subscription amendment.

ludovic-tc:
Fixed amendments to garden waste subscriptions paid for by Direct Debit by adding missing DD due date to hash in call to Integrations::Pay360->one_off_payment.

for https://mysocietysupport.freshdesk.com/a/tickets/1742

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -1270,9 +1270,13 @@ sub waste_munge_enquiry_data {
     $self->_set_user_source;
 }
 sub waste_get_next_dd_day {
-    my $self = shift;
+    my ($self, $payment_type) = @_;
 
-    my $dd_delay = 10; # No days to set up a DD
+    # new DD mandates must have a 10-day wait
+    my $dd_delay = 10;
+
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1270,9 +1270,13 @@ sub waste_munge_enquiry_data {
...
 }
 sub waste_get_next_dd_day {
-    my $self = shift;
+    my ($self, $payment_type) = @_;
+
+    # new DD mandates must have a 10-day wait
+    my $dd_delay = 10;
 
-    my $dd_delay = 10; # No days to set up a DD
+    # ad-hoc payments on an existing mandate only need a 5-day wait
```
dracos, line 1278: I'm sure this is okay, but where does this fact/change come from? I can't see anything in the PR text or elsewhere about it.

dracos (changes requested):
The actual line in Pay360.pm needs to use hyphens rather than slashes, according to the Pay360 documentation on Google Drive. I'd also like to know where the 5 day change came from.

dracos (approved):
Thanks, looks good.

PR #3745, opened 2022-01-13
[Bromley] Assisted collections

dracos:
Add the ability to make the enquiry form staff only, and then add display code to show assisted collection status/link to form to set assisted collection on. To have this working, you would create a new category called "Assisted collection" in the "Waste" group, destination 2149, with the normal waste extra fields "uprn" (hidden), "service_id" (hidden), "fixmystreet_id" (hidden"), plus "Exact_Location" (label "Exact location of containers", textarea), "reason" (label "Reason for request", textarea), and "staff_form" (hidden).

At the open311-adapter end, it would need a default in config - "Assisted Action" 1 (Add) - and some code to set "Assisted Start Date" to current date, "Assisted End Date" to 01/01/2050 and "Review Date" to 2 years from current date. Thought that was better there, than attaching to each report as extra data here. Matching PR at https://github.com/mysociety/open311-adapter/pull/212

Fixes https://github.com/mysociety/societyworks/issues/2476 [skip changelog]

### DIFF
--- a/templates/web/bromley/waste/_more_services_sidebar.html
+++ b/templates/web/bromley/waste/_more_services_sidebar.html
@@ -1,0 +1,26 @@
+[% IF any_report_allowed OR any_request_allowed OR show_garden_subscribe %]
+    <h3>More services</h3>
+    <ul>
+      [% IF any_report_allowed %]
+        <li><a href="[% c.uri_for_action('waste/report', [ property.id ]) %]">Report a missed collection</a></li>
+      [% END %]
+      [% IF any_request_allowed %]
+        <li><a href="[% c.uri_for_action('waste/request', [ property.id ]) %]">Request a new container</a></li>
+      [% END %]
+      [% IF show_garden_subscribe %]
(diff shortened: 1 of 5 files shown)


davea (approved):
♿ 🗑️ 🆗

templates/web/bromley/waste/_more_services_sidebar.html
```diff
@@ -0,0 +1,26 @@
...
+        <li><a href="[% c.uri_for_action('waste/request', [ property.id ]) %]">Request a new container</a></li>
+      [% END %]
+      [% IF show_garden_subscribe %]
+        <li><a href="[% c.uri_for_action('waste/garden', [ property.id ]) %]">Subscribe to Green Garden Waste collection</a></li>
+      [% END %]
+    </ul>
+[% END %]
+
+[% IF c.user.from_body OR c.user.is_superuser %]
+    <h3>Assisted collection</h3>
```
chrismytton, line 17: Might be worth only showing this heading if either `assisted_collection` or `domestic_collection` is true, otherwise you get a heading with nothing under it.

Spotted this on the Bromley staging site:

image

dracos:
Closing until picked up again.

PR #3746, opened 2022-01-14
[Hampshire] Adds cobranding

MorayMySoc:
Initial files for setting up the Hampshire Cobrand

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hampshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Hampshire.pm
@@ -1,0 +1,37 @@
+package FixMyStreet::Cobrand::Hampshire;
+use parent 'FixMyStreet::Cobrand::Whitelabel';
+
+use strict;
+use warnings;
+
+use Moo;
+with 'FixMyStreet::Roles::ConfirmOpen311';
+with 'FixMyStreet::Roles::ConfirmValidation';
+
+sub council_area_id { return 2227; } # https://mapit.mysociety.org/area/2227.html
+sub council_area { return 'Hampshire'; }
+sub council_name { return 'Hampshire County Council'; }
+sub council_url { return 'hampshire'; }
+
+# Created with bin/fixmystreet.com/generate_council_location
+sub disambiguate_location {
+    my $self    = shift;
+    my $string  = shift;
+    return {
+        %{ $self->SUPER::disambiguate_location() },
+        centre => '51.064173482874,-1.28695905016577',
+        span   => '0.677947505634364,1.22793350685304',
+        bounds => [ 50.7059677704377, -1.95727712614646, 51.3839152760721, -0.729343619293414 ],
+    };
+}
+
+sub lookup_site_code_config { {
+    buffer => 200, # metres
+    url => "https://tilma.staging.mysociety.org/mapserver/hampshire",
+    srsname => "urn:ogc:def:crs:EPSG::27700",
+    typename => "Road_Sections",
+    property => "SITE_CODE",
+    accept_feature => sub { 1 }
+} }
+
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Hampshire.pm
```diff
@@ -0,0 +1,37 @@
...
+        %{ $self->SUPER::disambiguate_location() },
+        centre => '51.064173482874,-1.28695905016577',
+        span   => '0.677947505634364,1.22793350685304',
+        bounds => [ 50.7059677704377, -1.95727712614646, 51.3839152760721, -0.729343619293414 ],
+    };
+}
+
+sub lookup_site_code_config { {
+    buffer => 200, # metres
+    url => "https://staging.tilma.mysociety.org/mapserver/hampshire",
```
dracos, line 30: This URL isn't right.

web/cobrands/hampshire/assets.js
```diff
@@ -0,0 +1,75 @@
...
+            SRSNAME: "urn:ogc:def:crs:EPSG::27700"
+        }
+    },
+    geometryName: 'msGeometry',
+    srsName: "EPSG:27700",
+    body: "Hampshire County Council",
+    strategy_class: OpenLayers.Strategy.FixMyStreet
+};
+
+var highways_style = new OpenLayers.Style({
```
dracos, line 26: Is this wanted? I wouldn't say it's something we do by default.

web/cobrands/hampshire/assets.js
```diff
@@ -0,0 +1,75 @@
...
+    }),
+    usrn: {
+        attribute: 'SITE_CODE',
+        field: 'site_code'
+    },
+    road: true,
+    asset_item: 'road',
+    asset_type: 'road',
+    no_asset_msg_id: '#js-not-a-road',
+    always_visible: true,
```
dracos, line 50: Why does this layer have both always_visible and an asset_group? They seem contradictory - and mean it's impossible to report anything not on a road in any category.

web/cobrands/hampshire/assets.js
```diff
@@ -0,0 +1,75 @@
...
+    http_options: {
+        params: {
+            TYPENAME: "salt_bins"
+        }
+    },
+    asset_type: 'spot',
+    asset_id_field: 'Asset_ID',
+    asset_category: ["Salt Bin"],
+    asset_item: 'salt bin'
+});
```
dracos, line 73: Should this be setting any attribute with any data, or is it purely for location accuracy?

dracos (changes requested):
The basics look fine, but the roads layer doesn't appear to work as expected.

web/cobrands/hampshire/assets.js
dracos: Sorry if this wasn't clear - the `asset_id_field: 'Asset_ID'` was fine, that's for keeping the same thing selected if data is reloaded from the server; this was more asking if there should be more code to do anything when it's selected or not (not is fine, it was just a question).

web/cobrands/hampshire/assets.js
```diff
@@ -0,0 +1,63 @@
...
+    strategy_class: OpenLayers.Strategy.FixMyStreet
+};
+
+fixmystreet.assets.add(defaults, {
+    http_options: {
+        params: {
+            TYPENAME: "Road_Sections"
+        }
+    },
+    usrn: {
```
dracos, line 32: You removed styling here, it needs some invisible styling as otherwise you get visible black lines, as you can see?

dracos (approved):
🗺️

PR #3749, opened 2022-01-14
Always assign new staff to a body

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/2202.

With the 'Add user' form on /admin/users, any users added without the 'Staff' checkbox being selected are then not visible for the given body/cobrand. This is because the 'Staff' checkbox assigns a body ID.

This PR fixes this by ensuring a body ID is always saved against any new user; additionally, if someone tries to add a user who already exists but does not have a body, that user will now be silently assigned to a body.

add-staff-user-01


perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -142,31 +142,60 @@ sub add : Local : Args(0) {
...
+        return if %{$c->stash->{field_errors}};
 
-    $c->forward('user_cobrand_extra_fields');
-    $user->insert;
+        # Assign body to user
+        $user->from_body($from_body);
+        $user->update;
 
-    $c->forward( '/admin/log_edit', [ $user->id, 'user', 'add' ] );
+        $c->forward( '/admin/log_edit', [ $user->id, 'user', 'edit' ] );
```
dracos, line 175: This log line is in both the if and else, so could be moved after it outside.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
     - Bugfixes:
         - Add ID attributes to change password form inputs.
         - Fix link deactivation for privacy policy link on privacy policy page. #3704
+    - Admin improvements:
+        - 'Staff' checkbox on `/admin/users` & `/admin/users/add` now permanently checked; submitting the form for a new user guarantees they are assigned to a body.
```
dracos, line 12: This file is generally wrapped to standard length, and tries to fit the change in one line if possible - perhaps something like "Admin add user form now always creates staff" or similar?

t/app/controller/admin/users.t
```diff
@@ -310,6 +310,139 @@ for my $test (
...
+
+            # Check if user appears in list
+            $mech->get_ok('/admin/users');
+            $mech->content_contains(
+                $test_body_user->email,
+                'user should appear in /admin/users list',
+            );
+
+            # Refresh user object by fetching data from DB
+            $test_body_user = $mech->create_user_ok($test_body_user->email);
```
dracos, line 387: You could use `$test_body_user->discard_changes` here.

perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -142,31 +142,60 @@ sub add : Local : Args(0) {
...
-        email_verified => $email && $email_v ? 1 : 0,
-        phone => $phone || undef,
-        phone_verified => $phone && $phone_v ? 1 : 0,
-        from_body => $c->get_param('body') || undef,
-        flagged => $c->get_param('flagged') || 0,
-        # Only superusers can create superusers
-        is_superuser => ( $c->user->is_superuser && $c->get_param('is_superuser') ) || 0,
-    } );
-    $c->stash->{user} = $user;
+    # Previously, an admin could add a staff user without a body (this is
```
dracos, line 145: Some of this comment I think would be better in the commit message - it describes before/after situation which is more suited to the description of the change. The comment here should describe what the code is doing, it doesn't matter what it used to do.

perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -142,31 +142,60 @@ sub add : Local : Args(0) {
...
+    # because the 'Staff' checkbox in the form could remain unchecked). A
+    # staff user without a body does not appear in the /admin/users list for
```
dracos, line 147: Worse, they might not be able to access this user at all because it had no association with the cobrand (unless e.g. admin_user_domain had been set up).

perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -142,31 +142,60 @@ sub add : Local : Args(0) {
...
+    # a given cobrand, which has caused confusion for admin, who may try to
+    # create the user again, only to be informed the user already exists.
+    #
+    # To solve this, a) a 'body' parameter is now guaranteed to be passed
```
dracos, line 151: This isn't strictly true - a superuser could still submit this form without a body (which I think is fine).

templates/web/base/admin/users/index.html
```diff
@@ -98,7 +98,7 @@
 [% END %]
 
 [% IF NOT ( searched || role_selected )%]
-<h2>[% loc('Add user') %]</h2>
+<h2>[% loc('Add staff user') %]</h2>
```
dracos, line 101: If a superuser is using this page, they can still add a normal user (which I think is fine), but then this title wouldn't entirely be accurate. I guess it probably is worth showing different title in that case, just to be clear.

templates/web/base/admin/users/edit.html
```diff
@@ -1,7 +1,7 @@
 [% IF user.id ~%]
     [% title = tprintf(loc('Editing user %d'), user.id ) ~%]
 [% ELSE ~%]
-    [% title = loc('Add user') ~%]
+    [% title = loc('Add staff user') ~%]
```
dracos, line 4: Ditto

dracos (changes requested):
This looks good, nice to see tests and documentation updates :) A few small changes to request is all (please do them as fixup commits so it's easier to re-review, thanks).

Whilst it was nice to see your commit history as you realised something wasn't working and fixed it (with an additional test, great), I as a reviewer didn't need to know that, so you could perhaps have squashed some of these commits together before review (doesn't really matter, though!). Generally it's good if the documentation changes are contained in the same commit that changes the code, easier to see when looking back.

perllib/FixMyStreet/App/Controller/Admin/Users.pm
nephila-nacrea: One logs 'edit', the other 'add', so I think I should keep them as is.

nephila-nacrea: Though I have rewritten the code to be less repetitive.

templates/web/base/admin/users/edit.html
```diff
@@ -1,7 +1,7 @@
 [% IF user.id ~%]
     [% title = tprintf(loc('Editing user %d'), user.id ) ~%]
 [% ELSE ~%]
-    [% title = loc('Add staff user') ~%]
+    [% title = loc('Add ' _ (c.user.is_superuser ? '' : 'staff ') _ 'user') ~%]
```
dracos, line 4: I'm afraid this won't work, because `loc()` needs an entire self-contained string for translation purposes (putting the string in a .po file for people to translate), so it can't deal with concatenations and the like. This would have to be `title = c.user.is_superuser ? loc("Add user") : loc("Add staff user")`

dracos (changes requested):
Fixups all look good, bar one text change that won't work as expected due to how translations work.

Generally (not sure this is written down, could be if it makes sense), I find it easier for PR comments if the PR author uses a thumbs up or a comment to signify something is dealt with (if they want to keep track of it that way) and then the re-reviewer uses the Resolve button when they've confirmed it is successfully dealt with. Just means the re-reviewer doesn't have to go through opening up all the comments to remember what they said :)

templates/web/base/admin/users/edit.html
nephila-nacrea: It looks like I need to do the same on index.html as well.

dracos (approved):
Looks good 🏖️

PR #3750, opened 2022-01-17
[Buckinghamshire] Scripts to setup parish councils

chrismytton:
Adds two scripts for setting up Buckinghamshire parishes as bodies:

1. `bin/buckinghamshire/create-parish-bodies` to pull in all known Buckinghamshire parishes from Mapit and create bodies for them.
2. `bin/buckinghamshire/add-parish-categories` which takes a CSV file with `name` and `email` columns where `name` is the parish body name and creates categories for those parishes which send to the provided email address.

The second script is necessary as Buckinghamshire aren't going to setup categories for all parishes immediately, so we need to be able to control which parishes have the necessary categories.

Fixes https://github.com/mysociety/societyworks/issues/2808

[skip changelog]

### DIFF
--- a/bin/buckinghamshire/add-parish-categories
+++ b/bin/buckinghamshire/add-parish-categories
@@ -1,0 +1,121 @@
+#!/usr/bin/env perl
+#
+# This script adds relevant categories for parishes listed in a CSV file by
+# attempting to match them to the existing parish bodies in FMS.
+#
+# The CSV file needs to have 'name' and 'email' columns with the name of
+# the parish council (which should match the existing body) and the email
+# address that reports should be sent to.
+
+use strict;
+use warnings;
+
+BEGIN {    # set all the paths to the perl code
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use FixMyStreet::DB;
+use Text::CSV;
+
+my $contacts = [
+    { category => 'Hedge problem', extra_metadata => { prefer_if_multiple => 1, group => 'Grass, hedges and weeds' } },
+    { category => 'Fly posting', extra_metadata => { prefer_if_multiple => 1 } },
+    { category => 'Dirty signs', extra_metadata => { prefer_if_multiple => 1 } },
+    {
+        category => 'Grass cutting',
+        extra_metadata => {
+            group => 'Grass, hedges and weeds',
+        },
+        extra_fields => [
+            {
+                code => 'speed_limit',
+                description => 'Is the speed limit on this road 40mph or greater?',
+                datatype => 'singlevaluelist',
+                order => 1,
+                variable => 'true',
+                required => 'true',
+                protected => 'false',
+                values => [
+                    {
+                        key => 'yes',
+                        name => 'Yes',
+                    },
+                    {
+                        key => 'no',
+                        name => 'No',
+                    },
+                    {
+                        key => 'dont_know',
+                        name => "Don't know",
+                    },
+                ],
+            }
+        ],
+    },
+];
+
+my @parishes;
+
+my $csv = Text::CSV->new ({ binary => 1, auto_diag => 1 });
+die "Usage: $0 <csv_file>\n" unless @ARGV;
+open my $fh, "<:encoding(utf8)", $ARGV[0] or die "$ARGV[0]: $!";
+$csv->header($fh);
+
+# Load parishes from supplied CSV file.
+while (my $row = $csv->getline_hr($fh)) {
+    my $name = $row->{name};
+    my $body = FixMyStreet::DB->resultset('Body')->find({ name => $name });
+    if (!$body) {
+        die "Error: Couldn't find body called $name\n";
+    }
+
+    push @parishes, { name => $name, email => $row->{email}, body => $body };
+}
+
+my $db = FixMyStreet::DB->schema->storage;
+$db->txn_do(sub {
+    # Create categories for parishes.
+    foreach my $parish (@parishes) {
+        foreach my $contact (@$contacts) {
+            my $contact_rs = FixMyStreet::DB->resultset('Contact')->search({
+                body_id => $parish->{body}->id,
+                category => $contact->{category},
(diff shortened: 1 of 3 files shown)


bin/buckinghamshire/create-parish-bodies
```diff
@@ -0,0 +1,90 @@
...
+my @areas = values %$mapit_areas;
+
+my $db = FixMyStreet::DB->schema->storage;
+$db->txn_do(sub {
+    foreach my $area (@areas) {
+        my $id = $area->{id};
+        my $name = $name_mappings->{$area->{name}} || "$area->{name} Parish Council";
+
+        my $body_rs = FixMyStreet::DB->resultset("Body")->search({
+            id => $id,
```
dracos, line 74: Whilst we historically set the ID of bodies to the area ID, the code was generalised to not require this some time ago and we shouldn't do it any longer.

bin/buckinghamshire/create-parish-bodies
```diff
@@ -0,0 +1,90 @@
...
+
+# Fetch the list of parishes from MapIt
+my $mapit_areas = mySociety::MaPit::call('area', "$parent_id/children", type => 'CPC');
+my @areas = values %$mapit_areas;
+
+my $db = FixMyStreet::DB->schema->storage;
+$db->txn_do(sub {
+    foreach my $area (@areas) {
+        my $id = $area->{id};
+        my $name = $name_mappings->{$area->{name}} || "$area->{name} Parish Council";
```
dracos, line 71: Do you want to remove the name when found, and then check they've all been found at the end? (I assume they have.)

bin/buckinghamshire/add-parish-categories
```diff
@@ -0,0 +1,77 @@
...
+
+        print "Creating category $category_name for $parish->{name}...";
+
+        my $contact = $contact_rs->create({
+            email => $parish->{email},
+            state => 'confirmed',
+            editor => basename($0),
+            whenedited => \'current_timestamp',
+            note => "created automatically by script",
+        });
```
dracos, line 69: It would be good to have a function on Contact that did this update, so you'd know to supply editor/note and it could add whenedited automatically, used in a few places I reckon.

bin/buckinghamshire/create-parish-bodies
```diff
@@ -0,0 +1,90 @@
...
+
+        if ($body_rs->count > 0) {
+            print "Existing body found for $name, skipping...\n";
+            next;
+        }
+
+        print "Creating body $name ($id)...";
+
+        my $body = $body_rs->create({ send_method => 'Noop' });
+        $body->body_areas->find_or_create({ area_id => $id });
```
dracos, line 86: I guess this won't be going on live before other code is ready, wondered if needed to consider a way to not add this body_area (or add it later on), so that it e.g. could be run on live site without the bodies becoming active as they didn't cover an area, iyswim.

dracos (changes requested):
One issue with body ID, and some more general points for discussion :)

bin/buckinghamshire/create-parish-bodies
chrismytton: Removed this line in 1dfd1c482001296dae538d493f0c9bb85ec4a76a

bin/buckinghamshire/create-parish-bodies
chrismytton: Yep seems like a sensible idea, will help with debugging when it inevitably fails in some interesting way!

Done in d5689a32307c8aa7d4aa4c6436a770f1ff6fb6e7

bin/buckinghamshire/add-parish-categories
chrismytton: Have added a method to the Contact ResultSet in e6304e4d04c17e0487ae28defd901dcd1ee46949 to help with adding a note when creating a contact.

bin/buckinghamshire/create-parish-bodies
chrismytton: Yeah, I guess ideally I'd do this in add-parish-categories when adding the categories to the parish, since we probably want to make it active at that point, maybe? But I don't have easy access to the mapit ID in that script.

If I add the body_areas entry but the parish has no categories, is it still considered to be active?

dracos: Even when you've added the categories, I guess it depends if the other code is all ready for reports to go to them correctly that we discussed (but yes, assuming that, that sounds okay).

If you add the body_areas but have no categories, then when you click the map on FMS and it finds one body with categories and one without, it will put the other body in missing_details_bodies and display the error message about we're missing details for that body and asking them to contact us. You could presumably put in an exception to ignore/not do this in some manner, though would be good to think of a better way....

...ooh if you add the bodies marked as deleted! That might work? Then you can undelete them when ready. Sorry for not thinking of that before (if it works!).

dracos (commented):
One comment about perhaps creating the bodies pre-marked as deleted (think that should work), plus how the name check might need to check the body area for future duplicate names.

chrismytton:
@dracos Marking as deleted seems to work, thanks! Have also scoped the query for existing bodies to check `body_areas.area_id` as well, to avoid potential clashes.

dracos (approved):

chrismytton:
Have pushed a fixup commit that adds `prefer_if_multiple`, category groups and an extra question for grass cutting.

bin/buckinghamshire/add-parish-categories
```diff
@@ -69,6 +97,19 @@ foreach my $parish (@parishes) {
             }
         );
 
+        $new_contact->set_extra_metadata(group => $contact->{group}) if $contact->{group};
+
+        # Only send reports to the parish
+        $new_contact->set_extra_metadata(prefer_if_multiple => 1);
+
+        if ($contact->{extra_fields}) {
+            foreach my $extra (@{$contact->{extra_fields}}) {
```
dracos, line 106: As it's always a new contact, you could probably have `$new_contact->set_extra_fields(@{$contact->{extra_fields}})` here instead of a loop.

dracos (approved):
Looks good, one potential suggestion is all

PR #3751, opened 2022-01-17
Remove aria-hidden from parents of #fms_pan_zoom

zarino:
Since the `#fms_pan_zoom` map controls are keyboard-navigable (and could conceivably be used by even non-sighted users, wanting to move the map view to reveal new markers in the `#map_sidebar` list) we need to make sure the controls aren’t hidden inside an `aria-hidden` parent.

Removing `aria-hidden` from `#map_box`, however, now exposes a whole load of new elements to assistive devices, including (in order):

1. each of the image tiles
2. each of the pin shadows
3. each of the pins
4. map data attribution at the bottom
5. each of the pan/zoom/other controls

In this commit, I’ve at least hidden the image tiles and pins in the non-JavaScript version(s) of the map. But doing the same thing in the OpenLayers JavaScript is beyond my current understanding!

Perhaps a developer can take a look?

TODO:

- [x] Tell OpenLayers to mark the image tiles and pin shadows as `aria-hidden="true"`
- [x] Decide whether to leave the pins accessible, or mark them as `aria-hidden="true"` as well (given there’s no way to skip over/past them, and there’s a list of all reports in `#map_sidebar` anyway).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
     - Bugfixes:
         - Add ID attributes to change password form inputs.
         - Fix link deactivation for privacy policy link on privacy policy page. #3704
+        - Make map pan/zoom controls keyboard-accessible. #3751
     - Admin improvements:
         - Admin 'add user' form now always creates staff users
     - Development improvements:
--- a/templates/web/base/maps/noscript_map.html
+++ b/templates/web/base/maps/noscript_map.html
@@ -1,6 +1,6 @@
 [% SET start = c.config.ADMIN_BASE_URL IF admin -%]
 <div class="noscript">
-    <div id="[% nsm_prefix %]drag">
+    <div id="[% nsm_prefix %]drag" aria-hidden="true">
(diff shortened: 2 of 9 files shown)


dracos:
"In this commit, I’ve at least hidden the image tiles and pins in the non-JavaScript version(s) of the map" - only the tiles I think? pins is the next div on.

zarino:
Thanks for fixing this up @dracos :+1:

I’ve tried it out in Safari with VoiceOver and the map controls are all now accessible with `⌃⎇→` etc. The flow is:

1. header
2. navigation menu
3. "main" element
4. the map copyright statement (shame that comes _first_, but hey ho)
5. each map control in order, starting with "pan up"
6. "Click map to report a problem" banner
7. "Can’t see the map, skip this step"
8. "Get updates" link
9. list filter form
10. pagination
10. each list item link in order

Happy for me to squash into one commit, with a changelog entry, and merge?

dracos (approved):

PR #3752, opened 2022-01-17
Replace “Powered by FixMyStreet Platform” credits on council cobrands with the new SocietyWorks footer

zarino:
Fixes mysociety/societyworks#2652.

I did Bromley first, and then noticed that we probably don’t want to display "Powered by FixMyStreet Pro" at the bottom of a WasteWorks page (?) so I added a fixup commit that hides the SocietyWorks footer on Bromley’s Waste pages.

(I guess, eventually, we’ll want to update the SocietyWorks footer to allow for WasteWorks (and other!) logos, instead of just FixMyStreet Pro.)

[skip changelog]


dracos:
For info - wasteworks has a special "Powered by SocietyWorks" pushed in its main content (either in bromley/footer.html or bromley/report/_updates_disallowed_message.html), so that could be tidied if done in footer properly.

dracos (changes requested):
Couple of questions on Bromley is all

zarino:
@dracos you’re right about the Bromley changes – not sure what I was thinking! I’ve:

* Moved the SocietyWorks footer back into `bromley/footer_extra.html` (which means it no longer appears _after_ the `<script>` elements etc at the end of the `body`)
* Used a `bodyclass` check inside Bromley’s SocietyWorks footer to vary the link and logo based on whether it’s a FixMyStreet or WasteWorks page
* Dropped the existing hacked-in `.bromley-waste-credits` stuff at the end of the WasteWorks page content wrapper (and its associated Sass styles and Veolia image file)

And, not really related to the footer work, but as you suggested:

* Removed `footer_marketing.html` from `bromley/footer_extra.html` to remove the App Store download links

templates/web/bromley/footer_extra.html
```diff
@@ -38,4 +38,17 @@ <h5>Social</h5>
...
     </div><!--footer-container-->
-</footer>
\ No newline at end of file
+</footer>
+
+<div class="societyworks-footer">
+    <div class="container">
+        <p>
+            Powered by
+          [% IF bodyclass == 'waste' %]
```
dracos, line 47: Ah, the problem here is that a waste report page isn't a bodyclass==waste page. Looking at other templates, looks like we have elsewhere used `bodyclass == 'waste' OR (problem AND problem.cobrand_data == 'waste')` which I think would also work here.

dracos (changes requested):
One issue with the Bromley footer logo on waste report pages is all :)

templates/web/bromley/footer_extra.html
zarino: Ah! Thanks for spotting that. It’s a shame we don’t have a simpler "is this part of the waste site or not" flag. Oh well, probably something for another day. I’ve pushed a fixup commit with the change you suggested.

dracos (approved):

PR #3754, opened 2022-01-18
[Shropshire] Remove form input focus outline to match branding guidelines

zarino:
Fixes https://github.com/mysociety/societyworks/issues/2789 (hopefully!)

Hides the browser’s default focus outline on form inputs in the Shropshire cobrand, so that Shropshire’s custom focus (created with a border and box-shadow) shows through, uninterrupted – as requested by their branding guidelines document.

This also means that the multiselect dropdown toggles and the _native_ dropdowns on the `/around` page now match:

Before:

broken

After:

fixed

[skip changelog]


dracos (approved):
🐗

PR #3758, opened 2022-01-19
[Hackney] Remove old noise reporting flow.

dracos:
[skip changelog] Uses static controller to serve a static page at the old location.


davea (approved):
📣 👋

PR #3760, opened 2022-01-20
[Hampshire] base_url returns fixmystreet base_url

MorayMySoc:
Returns FixMyStreet base url instead of cobrand specific url

[skip changelog]

### DIFF
--- a/t/cobrand/hampshire.t
+++ b/t/cobrand/hampshire.t
@@ -1,0 +1,12 @@
+use FixMyStreet::Cobrand::Hampshire;
+use FixMyStreet::TestMech;
+
+FixMyStreet::override_config {
+    ALLOWED_COBRANDS => [ 'fixmystreet' ],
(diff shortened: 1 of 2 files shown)


t/cobrand/hampshire.t
```diff
@@ -0,0 +1,7 @@
+use FixMyStreet::Cobrand::Hampshire;
+use FixMyStreet::TestMech;
+
+my $hampshire = FixMyStreet::Cobrand::Hampshire->new;
+is $hampshire->base_url, 'http://www.example.org', "Hampshire returns fixmystreet base_url";
```
dracos, line 5: Only suggestion is I would put this in an override_config section, just so you know it can't be overridden by anything. But

dracos (commented):
It's only a tiny suggestion, but given you've done the test might as well add the extra bit for it.

dracos (approved):

PR #3762, opened 2022-01-21
[Bristol] Adds bus stops to asset layer

MorayMySoc:
[skip changelog]

Bristol are adding bus stops to their assets layer.

https://github.com/mysociety/societyworks/issues/2774

### DIFF
--- a/web/cobrands/bristol/assets.js
+++ b/web/cobrands/bristol/assets.js
@@ -78,4 +78,10 @@ fixmystreet.assets.add(options, {
     filter_value: 'S180'
 });
 
+fixmystreet.assets.add(options, {
+    asset_group: "Bus Stops",
+    asset_item: 'bus stop',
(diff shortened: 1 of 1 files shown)


web/cobrands/bristol/assets.js
```diff
@@ -78,4 +78,10 @@ fixmystreet.assets.add(options, {
     filter_value: 'S180'
 });
 
+fixmystreet.assets.add(options, {
+    asset_group: "Bus stops",
```
dracos, line 82: The group is called "Bus Stops" in the staging admin

dracos (changes requested):
The group name doesn't look right, but other than that, assuming those are the right values to use, is fine

dracos (approved):

MorayMySoc:
This is merged into Bristol-reviewed

Issue #3304, opened 2021-02-02
Work "publish" into button labels at the end of reporting and commenting forms


zarino:
We want to try to remind people that their reports and comments will be posted publicly. I have a feeling that including the word "publish" in the submission buttons (that are currently labelled just "Submit") would help with this.

Here’s how it might look:

![Screenshot 2021-02-02 at 17 00 31](https://user-images.githubusercontent.com/739624/106635699-0d172180-6579-11eb-880c-f8c9412e3cf9.png)

Some considerations / edge-cases:

* Do we want to try this out on just fixmystreet.com and see whether it helps with user support volumes, before we move it into core? Just thinking, it’s yet another string to be translated in all the cobrands.
* What about reports made in areas we don’t have a `body` for? These aren’t "sent" anywhere. So maybe the button should just say "Publish your report" in those cases?
* What about comments on reports in areas where we don’t send updates to the council. Again, maybe the button should just say "Publish your update" in these cases?

zarino:
Note – it looks like FMS now allows "non public" reports (in only some areas, or for only staff users, maybe? can’t tell) so the button text will _also_ need to consider the value of this checkbox, to avoid a situation where "Private" is ticked but the button still says "Publish"

<img width="448" alt="Screenshot 2022-01-28 at 08 13 44" src="https://user-images.githubusercontent.com/739624/151511273-e5d83e87-bd21-4942-bb5c-1fe4f290d7d1.png">

dracos:
Not doing anything with this at present.

PR #3766, opened 2022-01-28
Work "publish" into button labels at the end of reporting and commenting forms

lucascumsille:
Fixes part of #3304 
This is part of the work needed for changing the submit button text, but I need help from a developer when and how the javascript should run to update the text.


dracos:
Whilst the report form can be filled in client side and so will need some JS (though it should also try and work out what it is server-side too), an update form is on a fixed known report, so everything about what it should say can be server-side, there is no need for any JavaScript in that case, unless I've missed something.

Issue #3756, opened 2022-01-19
Disallow spaces in extra questions [2d]


ludovic-tc:
**Describe the bug**
You can create an extra question for a category with spaces in the name, e.g. 'Road markings' and it will appear to work. However, if you try to set up one of the answers to disable the report and show a message, the question and answers will appear, but the stop message won't work, the Continue button won't get disabled and reporting will sail on to the photos stage. 

**To Reproduce**
* Create an extra question in any category with a name comprised of two words separated by a space. 
* Set one of the answers to disable the report. 
* Create a test report. 

**Expected behavior**
Choosing the appropriate answer should show the stop message and cause the Continue button to be disabled. 

**Platform:**
affects all platforms

**Additional context**
Cf.: https://mysocietysupport.freshdesk.com/a/tickets/1757

PR #3771, opened 2022-02-01
Prevent submission of 'extra data' question codes that contain whitespace

nephila-nacrea:
Prevent submission of 'extra data' question codes that contain whitespace.

Codes are used as part of HTML ids, which require there not be
any whitespace. Codes saved with whitespace therefore
break JS (in particular, question responses that are meant
to disable the form will not work). Codes are now checked for
whitespace when a contact is updated.

Fixes https://github.com/mysociety/fixmystreet/issues/3756.

[skip changelog]

### DIFF
--- a/templates/web/base/admin/extra-metadata-item.html
+++ b/templates/web/base/admin/extra-metadata-item.html
@@ -27,8 +27,8 @@ DEFAULT behaviour = 'question';
 
         <div class="form-group">
             <label for="metadata-[% i %]-code">[% loc('Code') %]</label>
-            <span class="form-hint" id="metadata-[% i %]-code-hint">[% loc('The code used to store this field value in the database.') %]</span>
-            <input class="form-control" name="metadata[[% i %]].code" id="metadata-[% i %]-code" aria-describedby="metadata-[% i %]-code-hint" type=text value="[% meta.code | html %]">
+            <span class="form-hint" id="metadata-[% i %]-code-hint">[% loc('The code used to store this field value in the database.') %] [% loc('Must not contain spaces.') %]</span>
+            <input class="form-control" name="metadata[[% i %]].code" id="metadata-[% i %]-code" pattern="\S+" aria-describedby="metadata-[% i %]-code-hint" type=text value="[% meta.code | html %]">
         </div>
 
         <fieldset>
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -316,6 +316,13 @@ sub update_contact : Private {
     $c->forward('/admin/update_extra_fields', [ $contact ]);
     $c->forward('contact_cobrand_extra_fields', [ $contact, \%errors ]);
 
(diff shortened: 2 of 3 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -776,6 +776,11 @@ sub setup_categories_and_bodies : Private {
         $bodies_to_list{ $contact->body_id } = $contact->body;
 
         my $metas = $contact->get_metadata_for_input;
+
+        for (@$metas) {
+            $_->{code} =~ s/\s/_/g if $_->{code};
```
dracos, line 781: GitHub lost my comment, sigh.
I'm afraid this won't work - you're changing what is output, but then what is read back in in both `check_for_category` and `set_report_extras` does not know to look for this changed code and so won't be reading in the information.

I don't think this is the correct place to try and fix this issue. The possibilities I can see are:
1. Find and fix the JS that is erroring because it can't handle a code with a space in, in a similar way to ed0378056d1d3445a0d7b321da3652aa17f12292
2. In the admin, when someone submits a code with a space in, change it to underscore at that point - so it's stored in the database like that and is consistent everywhere
3. In the admin, error when someone submits a code with a space in and make them change it (and perhaps use a pattern attribute so they can't type a space in the box).

Not sure which is "best" of those - 1 would mean people could still have codes with spaces in, but do we really want that, doesn't feel like something that's ever needed; 2 would work fine, probably won't confuse anyone if it should happen; 3 might be most clear/straightforward? But happy to hear suggestions if you think another is preferable.

dracos (changes requested):

perllib/FixMyStreet/App/Controller/Report/New.pm
nephila-nacrea: I went with a version of 3) in the end. Maybe I should have just gone with an HTML pattern attribute but I felt a deeper check would be 'safer' (e.g. does the pattern attribute work on all browsers? Could it possibly fail/be overridden in some way? Though I may be overthinking it).

templates/web/base/admin/extra-metadata-item.html
```diff
@@ -27,7 +27,7 @@
 
         <div class="form-group">
             <label for="metadata-[% i %]-code">[% loc('Code') %]</label>
-            <span class="form-hint" id="metadata-[% i %]-code-hint">[% loc('The code used to store this field value in the database.') %]</span>
+            <span class="form-hint" id="metadata-[% i %]-code-hint">[% loc('The code used to store this field value in the database. Must not contain spaces.') %]</span>
```
dracos, line 30: This is fine; I wonder if we might as well have it as `[% loc('The code used to store this field value in the database.') %] [% loc('Must not contain spaces.') %]</span>` - so that the current translations aren't disturbed by the string changing, and so the smaller string could potentially be reused elsewhere in future without needing new translation. Both very minor and theoretical, but wouldn't hurt.

dracos (changes requested):
Looks good, works, just a couple of tiny suggestions is all :)

dracos (approved):
Great, works nicely 💮

PR #3773, opened 2022-02-02
[Buckinghamshire] Tweak geocoder result for road on county boundary

davea:
Pushes one particular geocoder result north a few thousandths of a degree so it's within Buckinghamshire's boundary.

For https://github.com/mysociety/societyworks/issues/2873

[skip changelog]


dracos (approved):
👍

PR #3777, opened 2022-02-03
Added invisible padding to Zoom buttons on mobile

lucascumsille:
Created a new file with icons for mobile(under 48em screen width), this adds extra invisible space to the Zoom in and Zoomout buttons.
Width changed from:
36px -> 44px (2 extra pixels above the minimum recommended size for mobile).

Fixes part of mysociety/societyworks/issues/2583

Screenshot 2022-02-03 at 11 45 36


zarino (commented):
Thanks for this @lucascumsille. I hadn’t realised this would require a whole new sprite, but I guess, since the background colour of the buttons, and their rounded corners, are baked into the existing sprite, there’s no way to make the buttons bigger without also making the elements in the sprite bigger too.

If we’re going to do this, I wonder whether it’s _actually_ time to just drop the sprite entirely, and use CSS for the button background-color / border-radius, with the little icons (arrows, plus, minus) being included as SVGs either inlined in the HTML or inlined into the CSS.

@dracos any thoughts?

Another alternative would be to keep the buttons _visually_ the same size as they are right now, but extend an transparent child element out, beyond the edges of the button, to catch stray taps/clicks. eg:

```
#fms_pan_zoom_zoomin {
  &:before {
    content: "";
    position: absolute;
    top: -5px;
    right: -5px;
    bottom: 0;
    left: -5px;
}

Although then I think you have to contend with the `overflow: hidden` on those elements, which is currently used (along with `padding-top` to push the element text out of view), as a quick fix for better accessibility. Hmmm…

dracos:
Always happy to inline little things for performance. How big would the little satellite be as an SVG? ;)
Don't have an opinion on extra space vs bigger button, I guess either have the same effect, bigger button probably bit more obvious?

PR #3780, opened 2022-02-03
Add prefer_if_multiple setting to contacts

chrismytton:
This is for the situation where multiple bodies covering the same area have a contact with the same name. Checking this box indicates that if multiple contacts are found with the same name then the one(s) with this box checked should be preferred.

If no contacts are found with this flag then fall back to the default behaviour of sending to all matching contacts.

If multiple contacts are found with this flag then it will send to all of them.

This is related to the Buckinghamshire parish changes (see #3750), the idea being the parish contacts have this flag on them and therefore only get sent to the parish, rather than to Bucks as well.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -276,7 +276,7 @@ sub update_contact : Private {
     $contact->send_method( $c->get_param('send_method') );
 
     # Set flags in extra to the appropriate values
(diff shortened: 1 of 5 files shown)


dracos (approved):
Nice simple change, looks like it has the desired effect :)
Only thing I can think of that might need checking is how this interacts with the National Highways code, but it looks like it'd be okay, I don't think that would affect this.

chrismytton:
> Only thing I can think of that might need checking is how this interacts with the National Highways code, but it looks like it'd be okay, I don't think that would affect this.

Yeah from the looks of it that's handled by a separate code path with the "single body only" bits, so I don't think it will be affected by this change.

PR #3784, opened 2022-02-04
Added invisible padding to Zoom buttons on mobile

lucascumsille:
I got rid of the sprites, now we are using inline svg for each one of the icons on mobile. For mobile I added some pseudo elements, that add invisible space, to the buttons.

Fixes part of https://github.com/mysociety/societyworks/issues/2583

Screenshot 2022-02-04 at 13 55 18


web/cobrands/sass/_base.scss
```diff
@@ -1927,14 +1971,25 @@ html.js #map .noscript {
...
-      color: #fff;
+      color: #fff !important;
       text-decoration: none;
       opacity: 1;
     }
   }
 
   .map-layer-toggle {
     @include svg-background-image("/cobrands/fixmystreet/images/map-layer-aerial");
+    @media only screen and (max-width:48em) {
```
zarino, line 1982: Absolutely positioning the `:before` element, here, to extend beyond the edge of the parent won’t work, because the parent has `overflow: hidden;` set.

Is this why you did all this `color: transparent` stuff, so that you could remove `overflow: hidden` but still have the text effectively hidden from view?

zarino: I’m starting to wonder whether maybe it might just be easier to make all the buttons physically bigger on smaller screens, rather than adding the invisible padding. At least it’d mean we don’t have to mess around with the `overflow: hidden` stuff 🤔

zarino (commented):
A couple of comments. I’m starting to wonder whether maybe we should just make the buttons bigger on mobile, rather than using the invisible padding? Given the zoom controls are now rendered in CSS (rather than baked into that image sprite) it is at least now _possible_ to change the size of the buttons. And maybe that’d help us avoid complexity with the `overflow: hidden` behaviour on the `.map-links`?

web/cobrands/sass/_base.scss
lucascumsille: @zarino But wouldn't make the buttons bigger would defeat the purpose of having the buttons using more space in the map? which was one Matthew's concern at the very beginning. Let me know what you think.

zarino: Out of interest, I just increased the buttons to 46px width/height in my browser (ie: the current 36px plus 5px padding on all sides) and, yeah, they do look quite plonky. Also turns out that the way we’ve done the aerial view and hide pins icons doesn’t lend itself to changing the size of the buttons either. Sigh.

Maybe the answer here is that there’s a better way for us to be doing the image replacement, than using `overflow: hidden`…

lucascumsille: @zarino I'm confused I thought we weren't using overflow:hidden anymore. You mean find another way instead of the color:transparent?

lucascumsille: @zarino could we get rid of the text inside the divs by using aria-label? I'm guessing there must be a reason why we haven't used it before, but why is that?

zarino: > I'm confused I thought we weren't using overflow:hidden anymore.

You removed `overflow: hidden` from the `*_pan_zoom_*` elements, but not from the `.map-links > *` elements.

As a result, the `-5px` overlap on `.map-links > *:before` has no effect, because the `:before` pseudo-element is cropped by the `overflow: hidden` on its parent.

I looks like you _intended_ to remove the `overflow: hidden` on `.map-links > *` but just forgot, as otherwise there’d be no point in you adding the `color: transparent !important;` on `.map-links > *`.

The downside of removing `overflow: hidden` and just making the text transparent is that the text is _still there_, and is clickable, outside the edges of the button – so if you try to click the map below the "Hide pins" button, for example, there’s a chance you click won’t land on the map, and will instead be captured by the invisible text overhanging below the button.

zarino: > could we get rid of the text inside the divs by using aria-label?

No, because we want those labels ("Aerial map" and "Hide pins") to be visible to _all users_ on hover/focus, so people know what the icons mean.

lucascumsille:
@zarino I just checked your latest commit on IE11, Firefox, and Chrome. I didn't find any issues
thanks for the feedback.

zarino:
@dracos I think this is ready for you to review – and merge, if you’re happy.

dracos (changes requested):
Looks good, only thin is that at desktop width the aerial/pin icons no longer line up with the pan/zoom.

PR #3786, opened 2022-02-07
Added instructions for keyboard users

### LINKED ISSUE #3321: Make map pan/zoomable by keyboard (2 days)
User request. OpenLayers has a keyboard module we don't use, which we could.
You could then pan/zoom the map with arrow/plus/minus; space to place a pin in the middle of the map (or remove it, I guess), and then arrow keys to move placed pin around. Would need instructions somehow.
Ties in a bit with #1360 in that I still think that should be always visible.

lucascumsille:
TODO:

 - [x] Markup and styling for keyboard instructions (@lucascumsille)
 - [x] Enable keyboard navigation in OpenLayers (@dracos)
 - [x] Show/hide instructions at the correct times (@lucascumsille / @dracos)

The idea is that, on page load, the "You can move around with your keyboard" instruction will be shown:

Screenshot 2022-02-16 at 14-00-23 Viewing a location

The instruction stays on screen until either…

**1) You use the keyboard to navigate** – at which point the first instruction is hidden, and the second level of instructions (message at the bottom, plus `.keyboard-pin` and `.keyboard-pulsating-dot`) are shown:

Screenshot 2022-02-16 at 14-01-11 Viewing a location

Or **2) You click the pan/zoom buttons, or you drag the map to pan, or you scroll the map to zoom** – at which point, we figure you’re not interested in keyboard navigation, and we hide the first instruction, without replacing it:

Screenshot 2022-02-16 at 14-01-39 Viewing a location

The instructions will also need to be hidden when you start the reporting flow (either by clicking the map, or pressing the space bar to start a new report).

Early Figma prototype with more explanation: https://www.figma.com/file/qkeLtzegvyzE5uXwYrD9GJ/FixMyStreet?node-id=40%3A570

Fixes #3321


web/cobrands/sass/_base.scss
```diff
@@ -1957,6 +1957,85 @@ html.js #map .noscript {
...
+  height: 0;
+  padding-top: 22px;
+  overflow: hidden;
+  background-color: #ECEBEB;
+  box-shadow: 0px 3px 0px rgba(0, 0, 0, 0.3);
+  background-repeat: no-repeat;
+  background-position: center;
+
+  &.zoom-in {
+    background-image: url("data:image/svg+xml,%3Csvg fill='none' height='14' viewBox='0 0 13 14' width='13' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='%23222'%3E%3Crect height='13' rx='.3' width='3.25' x='4.875' y='.5'/%3E%3Crect height='13' rx='.3' transform='matrix(0 -1 1 0 -8.625 8.625)' width='3.25' y='8.625'/%3E%3C/g%3E%3C/svg%3E");
```
zarino, line 2001: My instinct here would be to either:

1. Use the `svg-background-image()` mixin for these too; or…
2. Embed the SVG markup directly into the HTML template

Either way would be a big improvement over maintaining escaped SVG markup in a `background-image: url()` definition like this.

web/cobrands/sass/_base.scss
```diff
@@ -1957,6 +1957,85 @@ html.js #map .noscript {
...
+  &.left {
+    background-image: url("data:image/svg+xml,%3Csvg fill='none' height='12' viewBox='0 0 17 12' width='17' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m1.72512 6.39261c-.25385-.20019-.25385-.58503 0-.78522l5.76151-4.54354c.03279-.02586.08096-.00251.08096.03926v3.74304h8.20511v2.3077h-8.20511v3.74305c0 .0418-.04817.0651-.08096.0393z' fill='%23222' stroke='%23222' stroke-linecap='round' stroke-width='2'/%3E%3C/svg%3E");
+    background-size: 14px auto;
+  }
+  &.right {
+    background-image: url("data:image/svg+xml,%3Csvg fill='none' height='12' viewBox='0 0 17 12' width='17' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m15.2749 5.60739c.2538.20019.2538.58503 0 .78522l-5.76153 4.54359c-.03279.0258-.08096.0025-.08096-.0393v-3.74305h-8.20513v-2.3077h8.20513v-3.74304c0-.04177.04817-.06512.08096-.03926z' fill='%23222' stroke='%23222' stroke-linecap='round' stroke-width='2'/%3E%3C/svg%3E");
+    background-size: 14px auto;
+  }
+  &.space {
+    @include svg-background-image("/cobrands/fixmystreet/images/space-key");
```
zarino, line 2017: Since this `space-key.svg` just contains the word "space" I wonder whether _this_ element could just use actual text (ie: `<div class="keyboard-key-icon space">space</div>`)?

One benefit is that it’ll make translation easier (rather than us needing a different SVG for every language!)

web/cobrands/sass/_base.scss
```diff
@@ -1957,6 +1957,85 @@ html.js #map .noscript {
...
+    background-size: 38px 15px;
+    width: 54px;
+  }
+}
+
+.keyboard-pin {
```
zarino, line 2023: We could use the existing pin icons (pin, and shadow, separately) for this, rather than making a new image asset. I’ll have a think about the best way to do that… 🤔

web/cobrands/sass/_base.scss
```diff
@@ -1957,6 +1957,85 @@ html.js #map .noscript {
...
+  background-repeat: no-repeat;
+  background-position: center;
+
+  &.zoom-in {
+    background-image: url("data:image/svg+xml,%3Csvg fill='none' height='14' viewBox='0 0 13 14' width='13' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='%23222'%3E%3Crect height='13' rx='.3' width='3.25' x='4.875' y='.5'/%3E%3Crect height='13' rx='.3' transform='matrix(0 -1 1 0 -8.625 8.625)' width='3.25' y='8.625'/%3E%3C/g%3E%3C/svg%3E");
+    background-size: 13px;
+  }
+  &.zoom-out {
+    background-image: url("data:image/svg+xml,%3Csvg fill='none' height='4' viewBox='0 0 15 4' width='15' xmlns='http://www.w3.org/2000/svg'%3E%3Crect fill='%23222' height='14' rx='.3' transform='matrix(0 -1 1 0 -3.25 4.25)' width='3.5' x='.5' y='3.75'/%3E%3C/svg%3E");
+    background-size: auto 4px;
```
zarino, line 2006: Assuming we keep these as background images, it would be _really nice_ if all of the SVGs had the same dimensions (say, 16x16px) so that we didn’t have to mess with different background-sizes for each one. Obviously, the shapes inside the SVG could be wider or narrower, etc, to maintain a consistent thickness / visual weight across the set, but I think an identical "canvas" size for each one, would make everything simpler.

templates/web/base/maps/_map_instructions.html
```diff
@@ -0,0 +1,18 @@
...
+        <div class="keyboard-key-icon right">Right</div>
+    </div>
+    <span>You can move around with your keyboard</span>
+</div>
+
+<div class="keyboard-intructions-wrapper advanced">
+    <div>
+        <div class="keyboard-key-icon space">Space bar</div>
+    </div>
+    <span>Place/Remove pin pressing the space bar</span>
```
zarino, line 15: Will need rewording. Not sure whether it’s clearer to say that the space bar will “drop a pin” or “place a pin” or whatever, or whether it’s better to say it’ll “start a new report”, since that‘s what people are _actually_ wanting to do. Hmmmm 🤔 I will have a think…

web/cobrands/sass/_base.scss
```diff
@@ -1957,6 +1957,85 @@ html.js #map .noscript {
...
+    display: inline-block;
+    margin-left: 20px;
+    font-weight: 600;
+  }
+
+  &.basic {
+    opacity: 0;
+  }
+
+  @media only screen and (max-width:48em) {
```
zarino, line 1983: Oooh, just noticed what you’re doing here. Rather than including stuff in `_base.scss` but hiding it on narrow screens with a media query, we tend to just move the stuff into `_layout.scss` instead, which only runs on wider screens to begin with.

I guess we should think about whether this keyboard prompt really _should_ be hidden on narrow screens, though. I assume the "mobile" map view will still be keyboard navigable. Thinking about it logically, the only time we’d really want to _hide_ the keyboard prompt is when _there is no keyboard_. I wonder whether we can detect that via JavaScript… 🤔

zarino (commented):
Thanks for this @lucascumsille.

I know this is a work in progress, but I‘ve left some comments with thoughts for now. Let’s have a chat about them in the next few days.

web/cobrands/sass/_base.scss
lucascumsille: Pseudo element on the keyboard-pin?

templates/web/base/maps/_map_instructions.html
```diff
@@ -0,0 +1,18 @@
+<div class="keyboard-intructions-wrapper basic">
+    <div>
+        <div class="keyboard-key-icon zoom-in">Zoom in</div>
```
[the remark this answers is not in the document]
zarino: On second thoughts, the text labels should also use the name of the _key_ rather than _what the key will do_, so:

```
<div class="keyboard-key-icon zoom-in">[% loc("plus") %]</div>
```

zarino (changes requested):
Getting there! Before we hand over to me or @dracos to get the JavaScript bits working, there are a few more improvements that came to mind this morning, as I was looking at it with fresh eyes! Let me know if you want to discuss them.

web/cobrands/sass/_layout.scss
```diff
@@ -900,3 +900,133 @@ textarea.form-error {
...
+  background-image: url('/i/pin-green-big.png');
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 108px;
+  height: 90px;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%,-50%);
```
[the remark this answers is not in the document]
lucascumsille: I detailed in the commit message. At the end I left the pulsating dot in the middle of screen, if I was the user I would expect that to be the place where I would put the pin. I placed the pin higher up translate(-50%, -130%).

zarino:
@lucascumsille I‘ve added a few further tweaks. Hope you don’t mind!

One thing I noticed – the message layout really doesn’t handle mid-sized screens very well, eg: if you were crazy enough to use FMS on an iPad in portait mode…

Screen Shot 2022-02-09 at 16 28 46

But really it starts becoming a problem at widths much wider than this – try a window 1100px wide. Seems silly that it always uses just 50% of the available space, rather than trying to keep the message down to one or two lines.

Any thoughts on how best to handle this?

lucascumsille:
> 
@zarino Thanks again and sorry I totally forgot about medium size screens. So I just push what I thought was the best approach:
Screenshot 2022-02-10 at 09 47 05
Screenshot 2022-02-10 at 09 47 32

However I understand the message might be using more space than in the large screens, so a couple of options:

- Should we hide the message on touchscreens? but this might cause issues with users that have a touchscreen and NEED to use a keyboard.
- Other option would be to add some transparency just to white-background, so the user can still what under the message.
Screenshot 2022-02-10 at 09 48 35

Let me know what you think

lucascumsille:
@zarino The changes look really cool, I just checked It and works well. I added a new commit so when the items wrap, they align to the center instead of left.
Let me know what you think.

zarino:
@lucascumsille yeah, I tend to go for left-aligning things, as it’s often easier to read (easier to spot where the lines start). But on a message this short, it probably doesn’t make a difference. If you think centred looks better, we’ll go for that.

I’ve just thought – 1) we should check this works ok on browsers that don’t support flexbox, and 2) we should check it plays nice with left-to-right languages (using FixMyStreet’s `$left`, `$right`, and `flip()` if necessary).

zarino (approved):
I think the markup and styling for this is done. Ready for @dracos to have a go at making the JavaScript bits work.

templates/web/base/maps/_map_instructions.html
```diff
@@ -0,0 +1,20 @@
+<div class="map-instructions" aria-live="polite">
+    <div class="keyboard-intructions-wrapper" id="keyboard-instructions-first">
+        <span>[% loc("You can move around with your keyboard") %]</span>
+        <div class="keyboard-key-group">
+            <div class="keyboard-key-icon zoom-in">[% loc("plus") %]</div>
```
dracos, line 5: Sorry, just noticed - this should be "zoom in" - the text should be an alternative, not a description.

zarino (approved):
I’ve tried this out with @dracos’ javascript additions, and it’s amazing! Spotted a CSS bug in IE9–11, and decided to tidy up all the IE conditional styles while I was fixing it.

dracos:
Tidied up the branch, added the missing map types (so that's all in the first commit now), and added a new commit to refactor the controls given how shared they generally are. So just needs checking that they keyboard stuff and my refactor works in various cobrands.

lucascumsille:
@dracos @zarino 
It looks great and the controls work fine (Tested on Chrome, IE11 and Edge)

I have tested it on FixMyStreet, I noticed a couple of things, but I'm not sure if it has an easy fix:

For a user that can only use the keyboard is a little bit tricky to use:
- After they place the pin with the space bar, they need to press the tab button like 15 times to get to the "continue" button. Is it possible to press "enter" to go to the next step?
- Ideally, instead of having to refresh the page to start over if they have misplaced the pin, pressing the space bar again should allow them to move the pin around until they place the space bar. (It seems similar to what @zarino was commented)

Let me know if you have any thoughts

dracos:
"After they place the pin with the space bar, they need to press the tab button like 15 times to get to the "continue" button. Is it possible to press "enter" to go to the next step?" - quite agree this is annoying, tab would ideally go straight to category. It doesn't seem to be that easy to do, though, I tried to focus on map_sidebar when you hit space, but that didn't seem to work all the time, didn't work out why. And the category list might not have loaded yet. So when you hit space it really needs to have a spinner, wait for categories to load then focus on the category label/question.

"Ideally, instead of having to refresh the page to start over if they have misplaced the pin, pressing the space bar again should allow them to move the pin around until they place the space bar. (It seems similar to what @zarino was commented)" - yes, I'll look at making space a toggle now. Note you don't need to refresh, you can also hit Back to... go back ;-)

lucascumsille:
> "After they place the pin with the space bar, they need to press the tab button like 15 times to get to the "continue" button. Is it possible to press "enter" to go to the next step?" - quite agree this is annoying, tab would ideally go straight to category. It doesn't seem to be that easy to do, though, I tried to focus on map_sidebar when you hit space, but that didn't seem to work all the time, didn't work out why. And the category list might not have loaded yet. So when you hit space it really needs to have a spinner, wait for categories to load then focus on the category label/question.
> I see, yeah I thought there would be a reason why. 
> "Ideally, instead of having to refresh the page to start over if they have misplaced the pin, pressing the space bar again should allow them to move the pin around until they place the space bar. (It seems similar to what @zarino was commented)" - yes, I'll look at making space a toggle now. Note you don't need to refresh, you can also hit Back to... go back ;-)
Thanks for look into that, Matthew. Appreciate it

dracos:
I've pushed a commit that makes space toggle so you can toggle it back on. Not sure if it needs a new help box being shown when you hit space the first time saying "hit space again to reposition" or similar?

zarino:
> Not sure if it needs a new help box being shown when you hit space the first time saying "hit space again to reposition" or similar?

Yeah, feels like another `.keyboard-instructions-wrapper` with content like "Press space again if you need to adjust the location" would help here. Otherwise, I’m not sure people would guess that hitting space a second time would undo the pin placement.

lucascumsille:
Agree

> I've pushed a commit that makes space toggle so you can toggle it back on. Not sure if it needs a new help box being shown when you hit space the first time saying "hit space again to reposition" or similar?

Agree but also, is it just me or before whenever we move around or zoom, we used to get the message "Drop pin on the map to start a new report" now I get the message once I have already pressed the space bar.

lucascumsille:
I just added a commit that displays the message once you have already place the pin with the space bar.
Let me know what you think.

dracos:
Almost perfect - just if you click Back when it's showing "Press space again to adjust the location" then the arrows message reappears but the "Press space again" message doesn't disappear. I've hopefully fixed it :)

dracos (approved):
I added a Cypress test to test the keyboard/instruction display functionality :)

Issue #1360, opened 2016-03-17
"Can't see the map? Skip this step" link is no longer visible


abibroom:
It's there in the page source but it's not showing for me. Tested in Chrome and Firefox.
![fms-nomap](https://cloud.githubusercontent.com/assets/560001/13851557/93872748-ec56-11e5-9cb8-f70656c25eed.png)

dracos:
It is keyboard accessible (appears when tabbing through as similar skip links do on BBC news etc); I wouldn't mind if it were always visible as it used to be, to cope with edge scenarios of map not loading or similar, but don't mind its current behaviour either; check with @zarino if there would be a way to do that or if we decide it's not worth it.

dracos:
We also don't want to encourage skipping unless necessary, I guess.

abimysoc:
Support email today where I think the problem was the user had an out of date browser that didn't load the map properly - this would have helped that person.

PR #3788, opened 2022-02-07
"Can't see the map?" message now is always visible

lucascumsille:
Moved to the top of the page, below main title. I got rid of the yellow background so it doesn't pop up much on the page,
encouraging users that can see the map to actually keep using it, instead of skipping it.

Screenshot 2022-02-07 at 15 12 38

Fixes part of #1360

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Unreleased
     - Front end improvements:
+        - Skip map link on /around is now always visible below big green banner #3788.
         - Highlight pin on sidebar focus as well as hover.
         - Map page pagination links now styled as links rather than buttons. #3727
         - Include username in inactive email.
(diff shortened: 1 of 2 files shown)


zarino (changes requested):
Styling for this looks good, but I don‘t think there’s any need to be moving it around, at run-time, with JavaScript?

Just leaving the `<a id="skip-this-step">` where it is, and giving it `margin: 0 -1em;` would make it appear full-width as required.

lucascumsille:
> Styling for this looks good, but I don‘t think there’s any need to be moving it around, at run-time, with JavaScript?
> 
> Just leaving the `<a id="skip-this-step">` where it is, and giving it `margin: 0 -1em;` would make it appear full-width as required.

I didn't realise that, thanks. I just fixed it.

zarino (approved):
Much better! One final thing, this probably deserves an entry in `CHANGELOG.md`, perhaps something like "Skip map link on /around is now always visible below big green banner #3788" under the "Front end improvements:" heading?

Once you’ve added something to `CHANGELOG.md`, and squashed all your commits into one, I think you’re good to merge to `master` 👍

PR #3789, opened 2022-02-07
Fix 'all' option for report status dropdown

nephila-nacrea:
The status and category dropdowns for report filtering share
much of the same JS code, but we need to include a special case for
the status dropdown; simply selecting all options means no
reports are shown (due to the inclusion of subtractive options such as
'private only'.)

Fixes https://mysocietysupport.freshdesk.com/a/tickets/1765.

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -137,8 +137,14 @@ function isR2L() {
             settings.presets = [];
             settings.presets.push({
                 name: settings.allText,
-                all: true
             });
+
+            if ($select.data('all-options')) {
+                settings.presets[0].options = $select.data('all-options');
+            }
+            else {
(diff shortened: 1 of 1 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -137,8 +137,14 @@ function isR2L() {
             settings.presets = [];
             settings.presets.push({
                 name: settings.allText,
-                all: true
             });
+
+            if ($select.attr('id') === 'statuses') {
```
dracos, line 142: This works fine, but I wonder, for possible futureproofing, if we could use this path if `$select.data('all-options')` exists? Currently that would only catch the same element, but I think it makes sense to use all-options if present in case it's added to something else.

dracos (approved):
Looks good 🐱

PR #3790, opened 2022-02-08
[Peterborough] Prevent problem/request on black bin when large bin request open

dracos:
[skip changelog]
Fixes https://mysocietysupport.freshdesk.com/a/tickets/1814

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -651,7 +651,7 @@ sub bin_services_for_address {
         my $report_service_ids = $container_removal_ids{$container_id};
         my @report_service_ids_open = grep { $open_requests->{$_} } @$report_service_ids;
         my $request_service_ids = $container_request_ids{$container_id};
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -489,6 +489,8 @@ sub look_up_property {
 
     my %premises = map { $_->{uprn} => $_ } @$premises;
 
+    my $attributes = $self->property_attributes($uprn);
```
davea, line 492: My one slight hesitation is that this adds another blocking Bartec API request to the "Your bin days" page, for a total of seven. It does appear to be the fastest, taking ~70ms locally, so perhaps it doesn't matter but thought it worth mentioning.

davea (approved):

perllib/FixMyStreet/Cobrand/Peterborough.pm
dracos: Yeah, I did wonder about that, but I couldn't see another way to know whether to show the report/request links on the bin days page. Guess we could leave the links showing and only if you clicked through would it not have them/greyed them out, but that didn't seem ideal.

PR #3791, opened 2022-02-08
[Peterborough] Don't allow duplicate food caddy requests

davea:
Prevents requests for new food caddies if there are already open service requests.

For https://github.com/mysociety/societyworks/issues/2878

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -711,20 +711,30 @@ sub bin_services_for_address {
     my $skip_bags = $self->{c}->get_param('skip_bags');
 
     @out = () if $bags_only;
-    my $food_containers = $bags_only ? [ 428 ] : $skip_bags ? [ 424, 423 ] : [ 424, 423, 428 ];
 
-    push @out, {
+    my @food_containers;
+    if ($bags_only) {
+        push(@food_containers, 428) unless $open_requests->{428};
+    } else {
+        unless ( $open_requests->{493} ) { # Both food bins
+            push(@food_containers, 424) unless $open_requests->{424}; # Large food caddy
+            push(@food_containers, 423) unless $open_requests->{423}; # Small food caddy
+        }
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -711,18 +711,26 @@ sub bin_services_for_address {
     my $skip_bags = $self->{c}->get_param('skip_bags');
 
     @out = () if $bags_only;
-    my $food_containers = $bags_only ? [ 428 ] : $skip_bags ? [ 424, 423 ] : [ 424, 423, 428 ];
 
-    push @out, {
+    my @food_containers;
+    if ($bags_only) {
+        push @food_containers, 428;
```
dracos, line 717: While we're here, FD 1819 mentions "Already has open 'Food bag request'" so could this check for open 428 too? (and same below).

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -711,18 +711,26 @@ sub bin_services_for_address {
...
+    } else {
+        push(@food_containers, 424) unless $open_requests->{424}; # Large food caddy
```
dracos, line 719: Could this and the next line also check for open 493 as that's a both bins request?

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -711,18 +711,26 @@ sub bin_services_for_address {
...
+        request_containers => \@food_containers,
         request_allowed => 1,
         request_max => 1,
         request_only => 1,
         report_only => 1,
-    };
+    }) if @food_containers;
 
     unless ( $bags_only ) {
         # We want this one to always appear first
```
dracos, line 736: While we're here, might as well also not have the ALL BINS if there's an open 425?

dracos (changes requested):
Code looks fine, a few suggestions of more open requests to also check :)

dracos (approved):
Looks good 🥫

PR #3793, opened 2022-02-09
Category on closure script

dracos:
Useful for if you only want to close reports in a particular category/categories.


chrismytton (approved):
Nice!

PR #3794, opened 2022-02-09
Fix colour contrast issues in map page sidebar

zarino:
- [x] Lighten pagination element background colour (for mysociety/societyworks#2844)
- [x] Sort out "click map" styling (lots of cobrands hacking this their own way, should really be using `.fake-link`, if at all, maybe wording should also change to "use", instead of "see", across the board?)
- [x] Changelog entry

### DIFF
--- a/web/cobrands/oxfordshire/base.scss
+++ b/web/cobrands/oxfordshire/base.scss
@@ -233,11 +233,16 @@ textarea {
 .extra-text {
   padding: 1em;
   margin: 0 -1em;
-  border-bottom: 1px solid $color-oxfordshire-pale-grey-green;
+  border-top: 1px solid $color-oxfordshire-pale-grey-green;
 
   p:last-child {
     margin-bottom: 0;
   }
+}
+
+.report-list-filters-wrapper {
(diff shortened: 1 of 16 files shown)


web/cobrands/oxfordshire/base.scss
```diff
@@ -233,13 +233,18 @@ textarea {
...
   margin: 0 -1em;
-  border-bottom: 1px solid $color-oxfordshire-pale-grey-green;
+  border-top: 1px solid $color-oxfordshire-pale-grey-green;
 
   p:last-child {
     margin-bottom: 0;
   }
 }
 
+.report-list-filterss-wrapper {
```
dracos, line 243: typo in the class name

dracos (changes requested):
One typo, I think, also commit message says you're picking see instead of use :) (Also mix of curly/non-curly quotes in there given you're there anyway ;) )
Looks good, wish we'd done this ages ago!

zarino:
Thanks @dracos I can always trust you to pick me up on missing curly quotes. Serves me right for committing at 6pm!

Happy for me to merge?

dracos (approved):
🐍

PR #3796, opened 2022-02-10
[Peterborough] Replace '==' with 'eq' for memcached property check

nephila-nacrea:
Purely numerical check was not working for Peterborough as its
property IDs are of the form 'postcode:id_no'. So a string
equality check must be used instead.

Fixes https://mysocietysupport.freshdesk.com/a/tickets/1821.

[skip changelog]

### DIFF
--- a/t/app/controller/waste_peterborough.t
+++ b/t/app/controller/waste_peterborough.t
@@ -64,33 +64,7 @@ FixMyStreet::override_config {
         send_reports => 1,
     },
 }, sub {
-    my $b = Test::MockModule->new('Integrations::Bartec');
-    $b->mock('Authenticate', sub {
-        { Token => { TokenString => "TOKEN" } }
-    });
-    $b->mock('Jobs_Get', sub { [
-        { WorkPack => { Name => 'Waste-R1-010821' }, Name => 'Empty Bin 240L Black', ScheduledStart => '2021-08-01T07:00:00' },
-        { WorkPack => { Name => 'Waste-R1-050821' }, Name => 'Empty Bin Recycling 240l', ScheduledStart => '2021-08-05T07:00:00' },
-    ] });
-    my $jobs_fsd_get = [
-        { JobID => 123, PreviousDate => '2021-08-01T11:11:11Z', NextDate => '2021-08-08T11:11:11Z', JobName => 'Empty Bin 240L Black' },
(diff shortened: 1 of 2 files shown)


t/app/controller/waste_peterborough.t
```diff
@@ -502,6 +476,55 @@ FixMyStreet::override_config {
...
+                'Should be forbidden due to overall view limit';
+            $mech->log_in_ok('staff@example.net');
+            $mech->get_ok(
+                '/waste/PE1 3NA:100090215480',
+                'Staff user shouldn\'t be limited'
+            );
+        }
+    };
+
+    $mech->log_in_ok('test@example.net');
```
dracos, line 525: I think it would be preferable to log_out_ok here (or in fact, up a few lines at the end of the SKIP inside which you logged in).

t/app/controller/waste_peterborough.t
```diff
@@ -502,6 +476,55 @@ FixMyStreet::override_config {
...
+                max_requests_per_day   => 3,
+                max_properties_per_day => 1,
+            },
+        },
+    },
+    STAGING_FLAGS => {
+        send_reports => 1,
+    },
+}, sub {
+    my ($b) = shared_bartec_mocks();
```
dracos, line 498: Good solution :)

t/app/controller/waste_peterborough.t
```diff
@@ -502,6 +476,55 @@ FixMyStreet::override_config {
...
+        SKIP: {
+            # MEMCACHED_HOST needs to be set to 'memcached.svc' in
+            # general.yml-example (or whatever config file you are using)
+            skip( "No memcached", 7 )
+                unless Memcached::set( 'waste-prop-test', 1 );
+            Memcached::delete("waste-prop-test");
+            Memcached::delete("waste-req-test");
+            $mech->get_ok('/waste/PE1 3NA:100090215480');
+            $mech->get_ok('/waste/PE1 3NA:100090215480');
+            $mech->get('/waste/PE1 3NA:100090215481');
```
dracos, line 510: If we make this 489 rather than 481, 489 is in the Bartec.pm sample_data and so we won't get some uninitialized warnings when running the tests.

dracos (changes requested):
Looks good, couple of tiny suggestions is all :)

t/app/controller/waste_peterborough.t
nephila-nacrea: Thank you ^.^

dracos (approved):
Looks good 🇬🇶

PR #3798, opened 2022-02-11
[Hackney] Remove reference to non-existent noise.scss

zarino:
4d5c60f1c266d70cdec784c3c99d13e4193e63ca accidentally added this `@import` line, even though `noise.scss` had previously been removed in 2043aa4a00af493690d0debbea7d95e51092e004.

Compiling with the `@import` line in place causes an exception, as the Sass compiler cannot find the file.

@dracos wasn’t sure whether we should just `--fixup` the original commit (even though it’s already been pushed to `master`) or add a new commit. Happy for you to decide!

[skip changelog]


dracos:
I fixup it

PR #3799, opened 2022-02-14
[Peterborough] Some bin fixes

dracos:
No all bins option if a bin request; hide sidebar report links pre 5pm; alwas show (to staff at the mo) request bin link.

[skip changelog]
Fixes rest of https://mysocietysupport.freshdesk.com/a/tickets/1827
Fixes https://mysocietysupport.freshdesk.com/a/tickets/1812
Fixes https://mysocietysupport.freshdesk.com/a/tickets/1818

Hopefully last of the dupe issues.

### DIFF
--- a/templates/web/peterborough/waste/_more_services_sidebar.html
+++ b/templates/web/peterborough/waste/_more_services_sidebar.html
@@ -1,6 +1,6 @@
 <h3>More services</h3>
 <ul>
-  [% IF any_report_allowed %]
+  [% IF any_report_allowed AND NOT on_day_pre_5pm %]
     <li><a href="[% c.uri_for_action('waste/report', [ property.id ]) %]">Report a missed collection</a></li>
   [% END %]
   [% IF any_request_allowed %]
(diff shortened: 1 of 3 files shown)


dracos:
Sorry, just pushed two more things here.

templates/web/peterborough/waste/_more_services_sidebar.html
```diff
@@ -3,7 +3,7 @@ <h3>More services</h3>
   [% IF any_report_allowed AND NOT on_day_pre_5pm %]
     <li><a href="[% c.uri_for_action('waste/report', [ property.id ]) %]">Report a missed collection</a></li>
   [% END %]
-  [% IF any_request_allowed %]
+  [% IF ( NOT waste_features.request_disabled ) AND c.user_exists AND (c.user.is_superuser OR (c.user.from_body AND c.user.from_body.name == "Peterborough City Council")) %]
```
davea, line 6: Wonder if there's a better way to make these things staff-only, now we have a few scattered in the templates. Or at least some greppable comment each time.
Would make it easier to remove staff restriction when the time comes.

davea (approved):
One suggestion, but 👍 overall.

templates/web/peterborough/waste/_more_services_sidebar.html
dracos: This is only the 2nd I think which I copied from the first ;) Take your point about easier to remove, but didn't seem worth it if it was just going to go soon.

davea: Yeah that's fair enough!

PR #3800, opened 2022-02-14
[Merton] Spot duplicate endpoint messages.

dracos:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton.pm
@@ -76,6 +76,19 @@ sub open311_extra_data_include {
     return [];
 }
 
+sub open311_get_update_munging {
+    my ($self, $comment) = @_;
+
+    my $latest = $comment->problem->comments->search({ state => 'confirmed' }, {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -76,6 +76,19 @@ sub open311_extra_data_include {
     return [];
 }
 
+sub open311_get_update_munging {
+    my ($self, $comment) = @_;
+
+    my $latest = $comment->problem->comments->search(undef, {
```
davea, line 82: Does it make sense to search the most recent visible comment rather than all? Can't think why it would matter but the question popped into my head.

```suggestion
    my $latest = $comment->problem->comments->search({ state => 'confirmed' }, {
```

davea (approved):

perllib/FixMyStreet/Cobrand/Merton.pm
dracos: Yeah, makes sense.

PR #3804, opened 2022-02-16
Add an option to bin/fetch to exclude bodies.

dracos:


chrismytton (changes requested):
I'm afraid this isn't working as expected.

The option in `bin/fetch` is still called `body`, but the two .pm scripts are now looking for a `bodies` option. Similarly the `exclude` option isn't being passed through to the scripts at all, and would need to be called `bodies_exclude` at the point it was passed through.

Hopefully easy to fix both of those issues in the `my %params` declaration.

Would also be good to have a test for the `bodies_exclude` parameter, if that's not too tricky.

dracos:
Not sure what I was thinking there! Thanks, fixup pushed with test.

chrismytton (changes requested):
Test looks good!

However I'm seeing errors when I run `bin/fetch`. I think the options passed from `bin/fetch` need some default values and/or the scripts need to check for undef and handle that correctly.

```
$ bin/fetch --body 'Rutland County Council' -v --reports
Can't use an undefined value as an ARRAY reference at perllib/Open311/GetServiceRequests.pm line 34.
```

```
$ bin/fetch --exclude 'Rutland County Council' -v --reports
Can't use an undefined value as an ARRAY reference at /Users/chrismytton/Code/mysociety/fixmystreet/perllib/Open311/GetServiceRequests.pm line 31.

```

PR #3805, opened 2022-02-17
Protect staff name & phone fields in FMS front-end.

ludovic-tc:
Name & phone fields for all staff are now disabled in the reporting and updating pages. 

fixes mysociety/societyworks#2133

Please check the following:

Thanks for contributing!


web/cobrands/fixmystreet/staff.js
```diff
@@ -102,13 +102,17 @@ fixmystreet.staff_set_up = {
         var $phoneInput = $('input[name=phone]');
         var $showNameCheckbox = $('input[name=may_show_name]');
         var $addAlertCheckbox = $('#form_add_alert');
+        var $extraTitle = $('select#form_fms_extra_title');
         if (val === 'myself') {
             $emailInput.val($emailInput.prop('defaultValue')).prop('disabled', true);
             $emailOptionalLabel.addClass('hidden');
-            $nameInput.val($nameInput.prop('defaultValue')).prop('disabled', false);
-            $phoneInput.val($phoneInput.prop('defaultValue')).prop('disabled', false);
+            $nameInput.val($nameInput.prop('defaultValue')).prop('readonly', true);
```
dracos, line 109: This change means that if the staff user switches to "another user", the field remains readonly and cannot be edited. And if you switch to anonymous user and back to yourself, the form is then both readonly and disabled. Is there a reason you switched from disabled to readonly? I can't see one, but may have missed something. It will either have to be switched back to disabled, or the other options here will need to be updated to deal with this change also. The test does not cover switching to another user and checking the fields are then editable.

templates/web/base/report/new/form_user_loggedin.html
```diff
@@ -18,7 +18,7 @@
 [% BLOCK form_as %]
     <label for="form_as">[% loc('Report as') %]</label>
     <select id="form_as" class="form-control js-contribute-as" name="form_as">
-        <option value="myself" [% c.user.from_body AND ( c.user.has_body_permission_to('planned_reports') || c.user.has_body_permission_to('default_to_body') ) ? '' : 'selected'  %]>[% loc('Yourself') %]</option>
+        <option value="myself" [% c.user.from_body AND ( c.user.has_body_permission_to('planned_reports') || c.user.has_body_permission_to('default_to_body') ) ? '' : 'selected'  %]>[% loc('Yourself') %] [% IF c.user.name %]([% c.user.name %])[% END %]</option>
```
dracos, line 21: Does this need an equivalent change in the update form?

dracos (changes requested):
This is a good start, but it e.g. breaks the "another user" option, and the test seems perhaps over-complicated for its needs.

web/cobrands/fixmystreet/staff.js
ludovic-tc: I was getting an error when using disabled because of course that means that those fields are not included in the POST. Report::New seems to require the name field. That is why I went with readonly. 

Thanks for pointing out that this is not finished! 😬

.cypress/cypress/integration/staff.js
```diff
@@ -43,4 +43,52 @@ describe('Staff user tests', function() {
         cy.get('[name=detail]').should('have.value', 'A Graffiti (offensive) problem has been found by Borsetshire County Council');
         cy.cleanUpXHR();
     });
+
+    it('does not let staff update their name, phone or email address whilst reporting or updating', function() {
+        // (Lest CS staff forget to select 'report as another user' and type the reporter's details into their own account.)
+        cy.server();
+        cy.route('**mapserver/peterborough*highways*', 'fixture:peterborough.xml').as('ptboro-roads-layer');
```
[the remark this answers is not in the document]
ludovic-tc: All visits to /report/new with a lat and long in existing tests were to cobrands with a mocked asset layer, so I didn't realise that Borsetshire was set up with a location.

When trying to create a new report with the default Borsetshire staff & location on fixmystreet.com I run into an error caused by missing IOW assets. 

I have handled this by preventing Cypress from bailing out if it finds an uncaught exception.

.cypress/cypress/integration/staff.js
```diff
@@ -43,4 +43,52 @@ describe('Staff user tests', function() {
...
+        // but doesn't always work, so have added {force:true} as well
+        cy.get('input[value="General fly tipping"]').should('be.visible').click({force:true});
+        cy.nextPageReporting();
+
+        // hazardous waste question
+        cy.get('#form_hazardous').select('No');
+        cy.nextPageReporting();
+
+        // photos page
+        cy.get('div[aria-label="Tips for perfect photos"] + button').click();
```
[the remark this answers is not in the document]
ludovic-tc: updated

templates/web/base/report/new/form_user_loggedin.html
ludovic-tc: added

.cypress/cypress/integration/staff.js
```diff
@@ -43,4 +43,52 @@ describe('Staff user tests', function() {
         cy.get('[name=detail]').should('have.value', 'A Graffiti (offensive) problem has been found by Borsetshire County Council');
         cy.cleanUpXHR();
     });
+
+    it('does not let staff update their name, phone or email address whilst reporting or updating', function() {
+        // Lest CS staff forget to select 'report as another user' and type the reporter's details into their own account.
+        cy.server();
+
+        Cypress.on('uncaught:exception', function(err, runnable) {
```
dracos, line 51: "All visits to /report/new with a lat and long in existing tests were to cobrands with a mocked asset layer" - even ignoring the many other test files that don't mock an asset layer (e.g. category_tests.js or regressions.js), the first test in this file does not mock an asset layer and progresses successfully to test a "Report as" box, which is practically the same thing you wish to be testing. You don't need a lat/lon for what you want to test, you could copy that existing test set up (or even perhaps have a shared setup function), that's why I suggested eg "[use] the same method of logging in".

We do not want a "Cypress.on" global exception - we want the tests to fail if there are any errors. At best as per https://docs.cypress.io/api/events/catalog-of-events#To-catch-a-single-uncaught-exception you would want to use `cy.on`, but this is not necessary because you don't need to visit an Isle of Wight location or anything like that. Replacing all the code from `Cypress.on` to the Flytipping selection line with a copy of `cy.route` to `cy.pickCategory` (with the different user) from the test just above this test, the remainder of your test appears to continue to pass successfully, without the need for special exception handling.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1108,6 +1108,8 @@ $.extend(fixmystreet.set_up, {
         hide('.js-new-report-user-hidden');
         show('.js-new-report-user-shown').then(function(){
             focusFirstVisibleInput();
+            // run now to protect staff fields when adding an update:
+            fixmystreet.staff_set_up.contribute_as();
```
dracos, line 1112: `staff_set_up` does not exist at this point if you are not logged in as staff, and will error (not a JS error because the deferred swallows it). This is not the correct place to do this (note no other use of staff in this file), but in staff.js in the `display:report` handler seems a good place to call it.

dracos (changes requested):
Thanks for trying to simplify the test; it does not need to have special exception handling, please see my comment there. The fixes in staff.js look to have been copied/pasted leading to duplicate/incorrect `val()` calls, and the new JS to call the staff set up won't work if you're not logged in as staff.

.cypress/cypress/integration/staff.js
ludovic-tc: Thanks for clarifying. I didn't think the global exception was appropriate – I just wanted to push up some code that didn't have a broken test. I should have made clear I didn't regard this as a proper solution by adding 'WIP'.

web/cobrands/fixmystreet/staff.js
```diff
@@ -123,24 +123,21 @@ fixmystreet.staff_set_up = {
...
+            $nameInput.val('-').prop('disabled', true).prop('readonly', false);
             $phoneInput.val('-').prop('disabled', true);
             $showNameCheckbox.prop('checked', false).prop('disabled', true);
             $addAlertCheckbox.prop('checked', false).prop('disabled', true);
         } else if (val === 'body') {
             $emailInput.val('-').prop('disabled', true);
             $emailOptionalLabel.addClass('hidden');
-            $nameInput.val(txt).prop('disabled', true);
-            $nameInput.val('').prop('readonly', false);
+            $nameInput.val(txt).prop('disabled', false).prop('readonly', true);
```
ludovic-tc, line 140: @dracos I have updated this line to use 'readonly' instead of 'disabled' as the new tests weren't passing otherwise. However, previously `$nameInput` would be set to 'disabled' in this case and was fine. Is this an appropriate / acceptable change?

dracos: I can't replicate this, the new tests pass fine with `disabled` so going to leave this as is.

PR #3806, opened 2022-02-17
[Lincs] Make asset_item_message for road assets an empty string

nephila-nacrea:
Was previously 'null', which was throwing an error in
fixmystreet/assets.js->get_asset_pick_message().

Fixes https://github.com/mysociety/societyworks/issues/2855.

[skip changelog]

### DIFF
--- a/.cypress/cypress/integration/lincolnshire.js
+++ b/.cypress/cypress/integration/lincolnshire.js
@@ -1,0 +1,49 @@
+describe('Lincolnshire cobrand', function(){
+    describe('making a report as a new user', function() {
+        before(function(){
+            cy.server();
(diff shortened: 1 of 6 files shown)


dracos (commented):
I think this is all the fix that is needed, yep! However, it would be good to have a Cypress test of this - Lincolnshire doesn't have a front end test at present, something that visited the front page, entered a postcode, picked a relevant category and checked that an error wasn't thrown (or rather, that without your fix an error was thrown) would be good.

chrismytton (approved):
Awesome! ✨ Good to have a comprehensive test for Lincolnshire's end-to-end reporting workflow.

PR #3807, opened 2022-02-17
[Buckinghamshire] Routing for parish grass cutting categories

chrismytton:
Route grass cutting reports to the parish if the speed limit is 30mph or less, otherwise to the council. This is based on the answer to an extra question set on the parishes grass cutting contact which asks the user what the speed limit of the road is.

I've also made a couple of tweaks to the parish categories script to allow updating existing contacts as I've made some change to the extra question on the grass cutting category and want that to propagate to the existing staging categories.

Fixes https://github.com/mysociety/societyworks/issues/2812

[skip changelog]

### DIFF
--- a/.cypress/cypress/integration/buckinghamshire.js
+++ b/.cypress/cypress/integration/buckinghamshire.js
@@ -54,6 +54,38 @@ describe('buckinghamshire cobrand', function() {
     cy.contains('The road you have selected is on a regular gritting route').should('be.visible');
   });
 
+  describe("Parish grass cutting category speed limit question", function() {
+    var speedGreaterThan30 = '#form_speed_limit_greater_than_30';
+
+    beforeEach(function() {
+      cy.get('#map_box').click(290, 307);
+      cy.wait('@report-ajax');
+      cy.pickCategory('Grass cutting');
+      cy.wait('@around-ajax');
+      cy.nextPageReporting();
+    });
+
+    it('displays the parish name if answer is "no"', function() {
+      cy.get(speedGreaterThan30).select('no');
+      cy.nextPageReporting();
+      cy.nextPageReporting();
+      cy.contains('sent to Adstock Parish Council and also published online').should('be.visible');
+    });
+
+    it('displays the council name if answer is "yes"', function() {
+      cy.get(speedGreaterThan30).select('yes');
+      cy.nextPageReporting();
+      cy.nextPageReporting();
+      cy.contains('sent to Buckinghamshire Council and also published online').should('be.visible');
+    });
+
+    it('displays the council name if answer is "dont_know"', function() {
+      cy.get(speedGreaterThan30).select('dont_know');
+      cy.nextPageReporting();
+      cy.nextPageReporting();
+      cy.contains('sent to Buckinghamshire Council and also published online').should('be.visible');
+    });
+  });
 });
 
 describe('buckinghamshire roads handling', function() {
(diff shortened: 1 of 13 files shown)


dracos (changes requested):
One issue with the script change when updating a category is all, front end stuff looks good :)

bin/buckinghamshire/add-parish-categories
```diff
@@ -85,21 +85,26 @@ $db->txn_do(sub {
...
 
+            my $new_contact;
+
             if ($contact_rs->count > 0) {
-                print "Existing contact found...skipping\n";
-                next;
-            }
+                print "Updating contact $contact->{category} for $parish->{name}...";
+
+                $new_contact = $contact_rs->first;
```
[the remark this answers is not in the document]
chrismytton: Yeah good point, forgot that the note/editor fields would need updating. Luckily doing it the way you suggested actually simplified things a bit, which is nice!

perllib/FixMyStreet/DB/Result/Contact.pm
```diff
@@ -204,4 +204,12 @@ sub category_uneditable {
     return 0;
 }
 
+sub add_note {
```
dracos, line 207: Would be good to have a little t/app/mode/contact.t test testing this I guess.

web/cobrands/buckinghamshire/js.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (!fixmystreet.maps) {
+      return;
+  }
+
+  $(function() {
+    $(document).on('change', '#form_speed_limit_greater_than_30', function(e) {
+      if ($(this).val() === 'no') {
+        // Report for the parish
+        fixmystreet.update_public_councils_text($('#js-councils_text').html(), [fixmystreet.bodies[1]]);
```
dracos, line 11: I don't think we can rely on the ordering of fixmystreet.bodies. I think it will have to grep out Bucks here, and use Bucks in the other location.

dracos (changes requested):
I like the _category_extras_header fix. The JS I think needs fixing (and would be good to have a front-end test of this).

perllib/FixMyStreet/DB/Result/Contact.pm
chrismytton: Added in 177284eb1fb26441aeaac12c0337e2656a2aaccf

web/cobrands/buckinghamshire/js.js
chrismytton: Ah OK, have changed it to look explicitly for the Bucks body and added some tests in 18722f83b84edd58cefb4a3a2b7f93ab7a8dd533.

.cypress/cypress/integration/buckinghamshire.js
```diff
@@ -54,6 +54,43 @@ describe('buckinghamshire cobrand', function() {
...
+      cy.wait('@report-ajax');
+      cy.pickCategory('Grass cutting');
+      cy.wait('@around-ajax');
+      cy.nextPageReporting();
+      cy.get('#form_speed_limit_greater_than_30').select('dont_know');
+      cy.nextPageReporting();
+      cy.nextPageReporting();
+      cy.contains('sent to Buckinghamshire Council and also published online').should('be.visible');
+    });
+  });
```
chrismytton, line 93: Thinking about it I should probably tidy up the duplication in this a bit using `beforeEach`, will do that quickly now.

chrismytton: Done in 4e1c267682d2b8624b24156f7a56be748134f8fa

dracos (changes requested):
One suggestion to remove duplication, and one debugging left in (I think) is all :)

dracos (approved):
Woo :)

PR #3809, opened 2022-02-18
Upgrade Net::DNS.

dracos:
Tricky as the old one has to be uninstalled first... Take a copy of your `local` before testing in case it messes anything up.
Also upgrade Getopt::Long for it to work in Perl 5.18 (trusty in ESM, but should we drop it?).


chrismytton (approved):
Seems to work for me locally!

Could drop support for Perl 5.18, but if supporting it is just a case of updating `Getopt::Long` then guess we might as well continue supporting for a bit?

PR #3811, opened 2022-02-18
Switch geocoder files from http to https

MorayMySoc:
Makes Geocoder links be https rather than http - http causing an issue after an upgrade to server and http only to be problematic going forwards.

 Ticket: **Calls from Geocoder files should be https to prevent unexpected failues** mysociety/societyworks#2898


dracos (approved):
🇬🇸

PR #3814, opened 2022-02-24
Update manual install docs with Mac-specific issues

nephila-nacrea:
[skip changelog]

### DIFF
--- a/docs/install/manual-install.md
+++ b/docs/install/manual-install.md
@@ -47,7 +47,9 @@ A similar list of packages should work for other Debian-based distributions.
 (Please let us know if you would like to contribute such a package list or
 instructions for other distributions.)
 
-#### b. Mac OS X
+<a name="pre-req-pkg-mac"> </a>
+
+#### b. macOS
 
 Install either MacPorts or HomeBrew (you might well have one already), and then
 use the command below to install a few packages that FixMyStreet needs, for
@@ -56,7 +58,7 @@ which it's much simpler to install via a packaging system.
 ##### i. MacPorts
 
 {% highlight bash %}
-$ port install gettext p5-locale-gettext p5-perlmagick jhead postgresql91-server
+$ port install gettext jhead libpng openssl p5-locale-gettext p5-perlmagick postgresql91-server
 {% endhighlight %}
 
 ##### ii. HomeBrew
@@ -65,7 +67,7 @@ $ port install gettext p5-locale-gettext p5-perlmagick jhead postgresql91-server
 find an alternative way to install ImageMagick.
 
 {% highlight bash %}
-$ brew install gettext perlmagick jhead postgresql
+$ brew install gettext jhead libpng openssl@1.1 perlmagick postgresql
 $ brew link gettext --force
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


docs/install/manual-install.md
```diff
@@ -128,11 +128,45 @@ As well as installing dependencies, this script compiles our CSS (using
 `bin/make_css`), installs the database schema (using `bin/update-schema`),
 and compiles any translation `.mo` files (using `commonlib/bin/gettext-makemo`).
 
+#### macOS troubleshooting
```
chrismytton, line 131: We've already got a section with the heading `#### b. Mac OS X`, perhaps we should change that heading to macOS and then move this section you've added as a subsection under there. Something like `##### iii. Troubleshooting Perl module errors` or similar?

docs/install/manual-install.md
```diff
@@ -128,11 +128,45 @@ As well as installing dependencies, this script compiles our CSS (using
...
+
+Some Perl modules may fail to install without certain packages and paths in place.
+
+**Image::PNG::QRCode** requires the `libpng` package.
```
chrismytton, line 135: Might be worth adding this package to the existing `brew install` command on line 68.

docs/install/manual-install.md
```diff
@@ -128,11 +128,45 @@ As well as installing dependencies, this script compiles our CSS (using
...
+
+For any **SSL-related modules**, make sure `openssl` is installed. If using homebrew:
+
+    $ brew install --cask openssl
```
chrismytton, line 139: This command fails with `Error: Cask 'openssl' is unavailable: No Cask with this name exists.` when I run it locally.

I think it's actually the `openssl@1.1` package that's needed, so perhaps that could be added to the existing `brew install` command on line 68?

docs/install/manual-install.md
```diff
@@ -128,11 +128,45 @@ As well as installing dependencies, this script compiles our CSS (using
...
+
+Then you may need to create symlinks to folders Perl modules will recognise:
+
+    $ sudo mkdir /usr/local/lib
+
+    $ sudo ln -s /opt/homebrew/opt/openssl@1.1/lib/libssl.1.1.dylib /usr/local/lib/libssl.dylib
+
+    $ sudo ln -s /opt/homebrew/opt/openssl@1.1/lib/libcrypto.1.1.dylib /usr/local/lib/libcrypto.dylib
```
chrismytton, line 147: Looks like the convention in this file is to use the `{% highlight bash %}` tag for syntax highlighting and have no spaces between lines.

```suggestion
{% highlight bash %}
$ sudo mkdir /usr/local/lib
$ sudo ln -s /opt/homebrew/opt/openssl@1.1/lib/libssl.1.1.dylib /usr/local/lib/libssl.dylib
$ sudo ln -s /opt/homebrew/opt/openssl@1.1/lib/libcrypto.1.1.dylib /usr/local/lib/libcrypto.dylib
{% endhighlight %}
```

docs/install/manual-install.md
```diff
@@ -128,11 +128,45 @@ As well as installing dependencies, this script compiles our CSS (using
...
+
+    export LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib"
+
+    export CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include"
+
+For **Locale::gettext** you may need to set one or more of the following (in addition to installing the `gettext` package):
+
+    export CPATH=/opt/homebrew/include
+
+    export LIBRARY_PATH=/opt/homebrew/lib
```
chrismytton, line 163: Might be safer to use `$HOMEBREW_PREFIX` instead of `/opt/homebrew` for non-standard (or older) homebrew installs.

Also as above this should be wrapped in `{% highlight bash %}` with no spaces between the lines.

chrismytton (changes requested):
Thanks for getting this information written down! And sorry I didn't do it myself when I ran into these issues previously 😬 

Couple of restructuring changes and minor suggestions, but the contents is solid.

docs/install/manual-install.md
nephila-nacrea: I felt it was better to put Perl module troubleshooting in this section as it is specifically dealing with Perl modules (we assume at this point that the reader has installed the required packages in the previous section). But I should move `libpng` there, as you suggest below.

docs/install/manual-install.md
nephila-nacrea: Yeah, I am a bit confused now - `brew install --cask openssl@1.1` fails too (though I have it pasted in my notes, like I ran it before and it worked - I am not really sure what `--cask` does versus installing normally). I will put `brew install openssl@1.1`, unless you know of a reason this is incorrect?

docs/install/manual-install.md
nephila-nacrea: I'll change this to be consistent with the rest of the doc, but I am not sure why this pattern is used, as it does not work in my VSCode preview, nor in GitHub (which I thought supported extended MD syntax) - is it even part of extended MD syntax?

dracos: It's Jekyll highlighting, see e.g. https://gist.github.com/roachhd/e434c1481e74841749d7 point 1. It works on https://fixmystreet.org/install/manual-install/ I think? :)

docs/install/manual-install.md
chrismytton: Yeah, I think `brew install openssl@1.1` should do the trick!

PR #3816, opened 2022-03-01
Added active state to navigation bar

lucascumsille:
The current navigation bar is a bit confusing, it doesn't show easily whereabout on the site the user currently is. This fix adds an active state to the current page.

Screenshot 2022-03-01 at 09 02 53


dracos (approved):

lucascumsille:
Thanks @dracos I'll merge it then.

PR #3820, opened 2022-03-02
[Buckinghamshire] Added fix to wider selection panel

lucascumsille:
This fix was included on layout.scss because it should only affect wider screen. 
Checked current state on mobile I didn't consider it needed any intervention.

This fix also prevents the changes in the position of the menu when selecting more than one option.

Tested on Edge, IE11, Firefox, Chrome, Opera, Safari.

Fixes: mysociety/societyworks#2905
Screenshot 2022-03-02 at 10 43 58

### DIFF
--- a/web/cobrands/buckinghamshire/layout.scss
+++ b/web/cobrands/buckinghamshire/layout.scss
@@ -165,3 +165,14 @@ body, .content {
   font-size: 1em;
   padding: 20px;
 }
+
+// Adjust the size dropdown button and menu for mappage
+.mappage {
+  .multi-select-button {
+    max-width: 80px;
+  }
+  
+  .multi-select-menu {
+    left: -55px;
+  }
+}


web/cobrands/buckinghamshire/layout.scss
```diff
@@ -164,3 +164,15 @@ body, .content {
...
+
+// Adjust the size dropdown button and menu for mappage
+.mappage {
+  .multi-select-button {
+    max-width: 80px;
+  }
+  
+  .multi-select-menu {
+    left: -55px;
+    width: 240px !important;
```
dracos, line 176: I think this line is unnecessary - with the left of the line above, the code automatically calculates a greater width for the box. 240px is too wide with the live category data, as the scrollbar underlaps the edge of the map:
image
But without this line it works it out as c. 215px and looks okay.

dracos (changes requested):
Nice fix for the flapping move, I think one change needed for the width thing.

web/cobrands/buckinghamshire/layout.scss
lucascumsille: You are right, it actually look much better. I just pushed a commit deleting that line. Thanks =)

dracos (approved):
Looks good :) When you squash, could you fix the commit message to have space after the `[...]` bit and capitalise Buckinghamshire, thanks :)

lucascumsille:
> Looks good :) When you squash, could you fix the commit message to have space after the `[...]` bit and capitalise Buckinghamshire, thanks :)

All done =), I fixed the commit message. Thanks for all the help Matthew.

PR #3821, opened 2022-03-02
[Peterborough] Work if multiple schedules present.

dracos:
Rather than take the first found per container, loop through and make sure we find the closest previous/next dates in all schedules to use. Old/new schedules may lack information. [skip changelog]


davea (approved):
✅

PR #3822, opened 2022-03-02
[Oxfordshire] Search all street lighting for dupes

dracos:
If one street lighting subcategory is picked, check all of them for duplicates to display, at a reduced distance. [skip changelog]


chrismytton (approved):
💡

PR #3823, opened 2022-03-03
[Oxfordshire] Remove chevrons from links in updates

MorayMySoc:
Some officers are using chevrons around links so removing chevrons from around fully qualified urls before sanitization, otherwise they are purged from the displayed update.

https://github.com/mysociety/societyworks/issues/2716

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Template.pm
+++ b/perllib/FixMyStreet/Template.pm
@@ -159,8 +159,11 @@ sub html_paragraph_email_factory : FilterFactory('html_para_email') {
 sub sanitize {
     my $text = shift;
 
+    # In case of markdown variant style of <https://www.google.com>
+    $text =~ s/<\s*(https?[^\s>]+)\s*>/$1/g;
+
     $text = $$text if UNIVERSAL::isa($text, 'FixMyStreet::Template::SafeString');
- 
+
     my %allowed_tags = map { $_ => 1 } qw( p ul ol li br b i strong em );
     my $scrubber = HTML::Scrubber->new(
         rules => [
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Template.pm
```diff
@@ -159,8 +159,11 @@ sub html_paragraph_email_factory : FilterFactory('html_para_email') {
 sub sanitize {
     my $text = shift;
 
+    # In case of markdown variant style of <https://www.google.com>
+    $text =~ s/<(\s?)+(https?.(?#>)*?)(\s?)+>/$2/g;
```
dracos, line 163: `(\s?)+` should be `\s*`. I'm not sure what `(?#...)` is commenting here?
I think something like `s/<\s*(https?[^\s>]?)\s*>/$1/g` would be a bit more straightforward; this current one will turn `< https://www and then some <b>bold</b> text` into `https://www and then some <bbold</b> text`

dracos (changes requested):
Think the regex can be simplified a bit, but that's all :)

perllib/FixMyStreet/Template.pm
MorayMySoc: `s/<\s*(https?[^\s>]?.*?)\s*>/$1/g;` Would that be ok or is that negating what you've added as doesn't strip the chevrons as is...

dracos: Oh, sorry, my final `?` should have been a `+` - you don't ever want any > or whitespace after the http, do you, so you don't want to use `.`

PR #3824, opened 2022-03-03
Enable more precise keyboard pin movement.

dracos:
With Ctrl held down, move five times less. And reduce the overall movement. [skip changelog]


lucascumsille:
> With Ctrl held down, move five times less. And reduce the overall movement. [skip changelog]

Great addition @dracos it works well. Just in case for me this works when I press shift+ctrl+arrow. ctrl 

I tested on IE11, safari and Chrome. Working well on all of them.

lucascumsille (approved):

PR #3826, opened 2022-03-03
[Buckinghamshire] Treat parish problems as belonging to the Bucks cobrand

chrismytton:
Because parishes are setup as separate bodies the default behaviour is to direct people to fixmystreet.com to follow the progress of their report.

Instead we want parish problems to be displayed on the cobrand and treated as cobrand problems. This is achieved by adding some new methods to UKCouncils.pm to make it easier to overrride the bodies/areas that are relevant to a particular cobrand, which will hopefully be useful if/when we add parishes elsewhere in the future.

[skip changelog]

### DIFF
--- a/t/app/controller/contact.t
+++ b/t/app/controller/contact.t
@@ -535,7 +535,9 @@ for my $test (
         FixMyStreet::override_config {
             ALLOWED_COBRANDS => [ 'buckinghamshire' ],
         }, sub {
-            $mech->get_ok( '/contact?id=' . $problem_main->id, 'can visit for abuse report' );
+            my $bucks = $mech->create_body_ok(2217, 'Buckinghamshire Council');
+            my ($problem) = $mech->create_problems_for_body(1, $bucks->id, 'Test');
+            $mech->get_ok( '/contact?id=' . $problem->id, 'can visit for abuse report' );
             $mech->submit_form_ok( { with_fields => $test->{fields} } );
             is_deeply $mech->page_errors, $test->{page_errors}, 'page errors';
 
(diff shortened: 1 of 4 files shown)


chrismytton:
Better fix these tests first 😬

t/app/controller/contact.t
```diff
@@ -535,6 +535,8 @@ for my $test (
         FixMyStreet::override_config {
             ALLOWED_COBRANDS => [ 'buckinghamshire' ],
         }, sub {
+            my $bucks = $mech->create_body_ok(2217, 'Buckinghamshire Council');
+            ($problem_main) = $mech->create_problems_for_body(1, $bucks->id, 'Test');
```
dracos, line 539: I'd probably stop reusing problem_main if it's just a problem for this subtest.

dracos (commented):
One tiny test comment. Looks like it was a bit fiddly, but not overly so, phew! Do we need to adjust problems_sql_restrictions similarly as well - should parish reports appear in Bucks local RSS feeds, I guess so?

chrismytton:
> Do we need to adjust problems_sql_restrictions similarly as well - should parish reports appear in Bucks local RSS feeds, I guess so?

Bucks has `sub is_two_tier { 1 }` already set in the cobrand, and `problems_sql_restriction` in UKCouncils.pm skips body checks for two-tier councils (see below). So I _think_ it should work without any additional changes.

https://github.com/mysociety/fixmystreet/blob/0e219477dd6d13240ba0e2c9511a86bcba89bcd6/perllib/FixMyStreet/Cobrand/UKCouncils.pm#L76-L87

t/app/controller/contact.t
chrismytton: Done in https://github.com/mysociety/fixmystreet/pull/3826/commits/29c1f682b45e06854ac6de2375afe4f343a27b9a

dracos:
"So I think it should work without any additional changes." - yes, ha, I remember now I did check that myself and came up with same thing you've just said but then when I came to write the review summary all I remembered was that it existed :)

dracos (approved):
🪂

PR #3827, opened 2022-03-04
[Bexley] Add warning about personal information to new report/update forms

davea:
Adds anti-phishing warning to top of report form, as well as above the update form.

[skip changelog]

For FD-1036.
Fixes https://github.com/mysociety/societyworks/issues/2776.

### DIFF
--- a/t/cobrand/bexley.t
+++ b/t/cobrand/bexley.t
@@ -261,6 +261,12 @@ FixMyStreet::override_config {
         $mech->get_ok('/report/' . $report->id);
         $mech->content_contains('Report ref:&nbsp;' . $report->id);
     };
+
+    subtest 'phishing warning is shown for new reports' => sub {
+        $mech->log_out_ok;
+        $mech->get_ok('/report/new?longitude=0.15356&latitude=51.45556&category=Lamp+post');
+        $mech->content_contains('if asked for personal information, please do not respond');
+    };
 };
 
 subtest 'nearest road returns correct road' => sub {
--- a/t/cobrand/bexley.t
+++ b/t/cobrand/bexley.t
@@ -267,6 +267,11 @@ FixMyStreet::override_config {
         $mech->get_ok('/report/new?longitude=0.15356&latitude=51.45556&category=Lamp+post');
         $mech->content_contains('if asked for personal information, please do not respond');
     };
+
+    subtest 'phishing warning is shown on report pages' => sub {
+        $mech->get_ok('/report/' . $report->id);
+        $mech->content_contains('if asked for personal information, please do not respond');
+    };
 };
 
 subtest 'nearest road returns correct road' => sub {
--- a/templates/web/bexley/report/new/councils_extra_text.html
+++ b/templates/web/bexley/report/new/councils_extra_text.html
@@ -1,0 +1,7 @@
+</p>
+
+<p>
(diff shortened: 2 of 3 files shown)


templates/web/base/report/new/category_extras.html
```diff
@@ -15,12 +15,16 @@
     <div class="extra-category-questions">
         [% UNLESS category_extras_notices.$category %]
           <h2 class="visuallyhidden form-section-heading">[% loc('Extra details') %]</h2>
+          [% TRY %]
+            [% INCLUDE 'report/new/_category_extras_header.html' %]
```
dracos, line 19: The message won't be shown unless a category is picked that has extra questions, and not all of Bexley's do - e.g. "Damage to grass verges", "Vehicles for sale on the public road", Animal fouling -> Park / open space, or Flytipping -> Garage areas.
I guess instead of as code it could be added as a 'notice' on every category, would that work the same? Bit annoying if they want to e.g. update the phone number I guess, but consistently displayed then. Otherwise, not sure - `report/new/form_after_heading.html`?

dracos (changes requested):
Update looks good (and I think is the place it's most needed anyway?), but it won't be displaying for all categories on a new report, I'm afraid.

templates/web/bexley/report/new/_category_extras_header.html
```diff
@@ -0,0 +1,15 @@
+<h2 class="visuallyhidden form-section-heading">[% loc('Extra details') %]</h2>
```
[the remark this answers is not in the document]
davea: Oops, yes copy-paste error :)

templates/web/base/report/new/category_extras.html
davea: I admit I was implementing their wireframes fairly literally, and assumed it would be OK to only appear on categories that already had extra fields...

Might be a bit tedious to put it on every category and manage manually. Maybe added in code in `Bexley::munge_report_new_category_list`?

davea: Though it occurred to me that doing it in the template will restrict the warning to the Bexley cobrand (desirable, I think?), whereas doing it as notices will put it on FMS.com and the app (which I don't think we want?)

dracos: Yeah, good point. Wonder where the best place to put it in a template is, then, if we want it for all categories.

davea: I've asked the client if this would be OK, would make it a much simpler change:

image

t/cobrand/bexley.t
```diff
@@ -261,6 +261,16 @@ FixMyStreet::override_config {
         $mech->get_ok('/report/' . $report->id);
         $mech->content_contains('Report ref:&nbsp;' . $report->id);
     };
+
+    subtest 'phishing warning is shown for new reports' => sub {
+        $mech->log_out_ok;
+        $mech->get_ok('/report/new?longitude=0.15356&latitude=51.45556&category=Lamp+post');
```
chrismytton, line 267: Does this test need a `$mech->content_contains('if asked for personal information, please do not respond');` like the one below?

chrismytton (approved):
One question about the tests, but other than that looks good!

PR #3828, opened 2022-03-04
[Oxfordshire] Show drainage last clean dates

dracos:
And include channels too. 
Also fixes https://github.com/mysociety/societyworks/issues/2698 I think
[skip changelog]


chrismytton (changes requested):
Looks good! Just one typo in the body name.

Issue #3168, opened 2020-08-25
Reports to multiple send methods don't always get sent correctly


davea:
If a report is being sent to multiple send methods and only one succeeds the report is still marked as sent and never retried to the failed destination.

To reproduce:

 - Create two bodies that cover the same location, one which uses Open311 and the other uses Email (e.g. a two-tier arrangement like Oxfordshire County & Cherwell District)
 - Create one contact for each body with the same name (e.g. 'Other')
 - Make a report within those bodies' boundaries and select the 'Other' category, and observe that it's going to be sent to both bodies
 - Making sure the Open311 destination is down/invalid, run `send-reports --verbose --debug`
 - Observe that the report email is sent to the one body but the Open311 send attempt fails
 - Viewing the report in the admin, observe that it's been marked as sent despite one send method failing.
 - Run `send-reports --verbose --debug` again and observe that no attempt is made to send the report via Open311

Expected behaviour:
 - If one send method fails then it should be retried again in the same way as a normal single-destination failure
 - Only a single email should be sent, instead of one for each Open311 attempt

PR #3829, opened 2022-03-07
Issues/public/3168 reports multiple send methods

nephila-nacrea:
A report may fall under the jurisdiction of two (or more) bodies, e.g. Cherwell District Council and Oxfordshire County Council. One body may have a send method of (e.g.) email and the other of (e.g.) Open311; if one of these send methods fails, the report is still marked as having been successfully sent, despite the fact that only one body will have received the report. This PR fixes this so that sending is reattempted until all methods have been successful.

Closes https://github.com/mysociety/fixmystreet/issues/3168.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -744,7 +744,7 @@ sub nearest_address {
 }
 
 sub body {
-    my ( $problem, $link ) = @_;
+    my ( $problem, $link, $successful_send_only ) = @_;
     my $body;
     if ($problem->external_body) {
         if ($problem->cobrand eq 'zurich') {
@@ -754,7 +754,14 @@ sub body {
             $body = FixMyStreet::Template::html_filter($problem->external_body);
         }
     } else {
-        my $bodies = $problem->bodies;
+        my %bodies = %{$problem->bodies};
+
+        # If we only want bodies with successful send methods (in the case of
+        # a problem that has multiple send methods), remove failed
+        if ($successful_send_only) {
+            delete $bodies{$_} for @{ $problem->send_fail_body_ids // [] };
+        }
+
         my @body_names = sort map {
             my $name = $_->name;
             if ($link and FixMyStreet->config('AREA_LINKS_FROM_PROBLEMS')) {
@@ -762,7 +769,7 @@ sub body {
             } else {
                 FixMyStreet::Template::html_filter($name);
             }
-        } values %$bodies;
+        } values %bodies;
         if ( scalar @body_names > 2 ) {
             $body = join( ', ', splice @body_names, 0, -1);
             $body = join( ',' . _(' and '), ($body, $body_names[-1]));
@@ -868,7 +875,8 @@ sub can_display_external_id {
 sub duration_string {
     my $problem = shift;
     my $cobrand = $problem->result_source->schema->cobrand;
-    my $body = $cobrand->call_hook(link_to_council_cobrand => $problem) || $problem->body(1);
+    my $body = $cobrand->call_hook( link_to_council_cobrand => $problem )
+        || $problem->body( 1, 1 );
     my $handler = $cobrand->call_hook(get_body_handler_for_problem => $problem);
     if ( $handler && $handler->call_hook('is_council_with_case_management') ) {
         my $s = sprintf(_('Received by %s moments later'), $body);
@@ -971,6 +979,28 @@ sub has_given_send_fail_body_id {
     my $id   = shift;
 
     return any { $_ == $id } @{ $self->send_fail_body_ids };
+}
+
+sub send_fail_methods {
+    # Deduce from send_fail_body_ids
+    my $self = shift;
+
+    my $bodies = $self->bodies;
+
+    my @send_fail_methods;
+
+    for (@{$self->send_fail_body_ids}) {
+        my $body = $bodies->{$_};
+        push @send_fail_methods, $body->send_method;
+    }
+
+    return \@send_fail_methods;
+}
+
+sub mark_as_sent {
+    my $self = shift;
+    $self->whensent( \'current_timestamp' );
+    $self->send_fail_body_ids( [] );
 }
 
 sub resend {
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -743,8 +743,9 @@ sub nearest_address {
     return $address->{name};
 }
 
+# Does not return bodies whose send methods have failed
 sub body {
-    my ( $problem, $link, $successful_send_only ) = @_;
+    my ( $problem, $link ) = @_;
     my $body;
     if ($problem->external_body) {
         if ($problem->cobrand eq 'zurich') {
@@ -756,11 +757,9 @@ sub body {
     } else {
         my %bodies = %{$problem->bodies};
 
-        # If we only want bodies with successful send methods (in the case of
-        # a problem that has multiple send methods), remove failed
-        if ($successful_send_only) {
-            delete $bodies{$_} for @{ $problem->send_fail_body_ids // [] };
-        }
+        # We only want bodies with successful send methods (in the case of
+        # a problem that has multiple send methods), so remove failed
+        delete $bodies{$_} for @{ $problem->send_fail_body_ids // [] };
 
         my @body_names = sort map {
             my $name = $_->name;
@@ -876,7 +875,7 @@ sub duration_string {
     my $problem = shift;
     my $cobrand = $problem->result_source->schema->cobrand;
     my $body = $cobrand->call_hook( link_to_council_cobrand => $problem )
-        || $problem->body( 1, 1 );
+        || $problem->body(1);
     my $handler = $cobrand->call_hook(get_body_handler_for_problem => $problem);
     if ( $handler && $handler->call_hook('is_council_with_case_management') ) {
         my $s = sprintf(_('Received by %s moments later'), $body);
@@ -945,12 +944,14 @@ sub updates_sent_to_body {
 }
 
 sub add_send_method {
-    my $self = shift;
+    my $self   = shift;
     my $sender = shift;
     $sender =~ s/${\SENDER_REGEX}//;
-    if (my $send_method = $self->send_method_used) {
-        $self->send_method_used("$send_method,$sender");
-    } else {
+
+    if ( my $existing_send_method = $self->send_method_used ) {
+        $self->send_method_used("$existing_send_method,$sender");
+    }
+    else {
         $self->send_method_used($sender);
     }
 }
@@ -960,7 +961,10 @@ sub add_send_fail_body_ids {
     my @new_ids = @_;
 
     $self->send_fail_body_ids(
-        [ uniq( @new_ids, @{ $self->send_fail_body_ids } ) ] );
+        [   sort { $a <=> $b }
+                uniq( @new_ids, @{ $self->send_fail_body_ids } )
+        ]
+    );
 }
 
 sub remove_send_fail_body_ids {
@@ -971,7 +975,7 @@ sub remove_send_fail_body_ids {
 
     delete @existing_ids{@remove_ids};
 
-    $self->send_fail_body_ids( [ keys %existing_ids ] );
+    $self->send_fail_body_ids( [ sort { $a <=> $b } keys %existing_ids ] );
 }
 
 sub has_given_send_fail_body_id {
@@ -981,20 +985,19 @@ sub has_given_send_fail_body_id {
     return any { $_ == $id } @{ $self->send_fail_body_ids };
 }
 
-sub send_fail_methods {
-    # Deduce from send_fail_body_ids
+sub send_fail_bodies {
     my $self = shift;
 
     my $bodies = $self->bodies;
 
-    my @send_fail_methods;
+    my @send_fail_bodies;
 
     for (@{$self->send_fail_body_ids}) {
         my $body = $bodies->{$_};
-        push @send_fail_methods, $body->send_method;
+        push @send_fail_bodies, $body->name;
     }
 
-    return \@send_fail_methods;
+    return \@send_fail_bodies;
 }
 
 sub mark_as_sent {
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -59,9 +59,8 @@ __PACKAGE__->add_columns(
   "created",
   {
     data_type     => "timestamp",
-    default_value => \"current_timestamp",
+    default_value => \"CURRENT_TIMESTAMP",
     is_nullable   => 0,
-    original      => { default_value => \"now()" },
   },
   "confirmed",
   { data_type => "timestamp", is_nullable => 1 },
@@ -78,9 +77,8 @@ __PACKAGE__->add_columns(
   "lastupdate",
   {
     data_type     => "timestamp",
-    default_value => \"current_timestamp",
+    default_value => \"CURRENT_TIMESTAMP",
     is_nullable   => 0,
-    original      => { default_value => \"now()" },
   },
   "whensent",
   { data_type => "timestamp", is_nullable => 1 },
@@ -114,6 +112,12 @@ __PACKAGE__->add_columns(
   { data_type => "text", is_nullable => 1 },
   "defect_type_id",
   { data_type => "integer", is_foreign_key => 1, is_nullable => 1 },
+  "send_fail_body_ids",
+  {
+    data_type     => "integer[]",
+    default_value => \"'{}'::integer[]",
+    is_nullable   => 0,
+  },
 );
 __PACKAGE__->set_primary_key("id");
 __PACKAGE__->has_many(
@@ -170,8 +174,8 @@ __PACKAGE__->has_many(
 );
 
 
-# Created by DBIx::Class::Schema::Loader v0.07035 @ 2019-04-25 12:06:39
-# DO NOT MODIFY THIS OR ANYTHING ABOVE! md5sum:hUXle+TtlkDkxkBrVa/u+g
+# Created by DBIx::Class::Schema::Loader v0.07035 @ 2022-03-10 11:09:46
+# DO NOT MODIFY THIS OR ANYTHING ABOVE! md5sum:W+bK8MqpeFwkTj0Ze/tN0g
 
 # Add fake relationship to stored procedure table
 __PACKAGE__->has_one(
@@ -244,6 +248,7 @@ use Utils;
 use FixMyStreet::Map::FMS;
 use FixMyStreet::Template;
 use FixMyStreet::Template::SafeString;
+use List::Util qw/any uniq/;
 use LWP::Simple qw($ua);
 use RABX;
 use URI;
@@ -254,6 +259,8 @@ my $IM = eval {
     Image::Magick->import;
     1;
 };
+
+use constant SENDER_REGEX => qr/^.*::/;
 
 with 'FixMyStreet::Roles::Abuser',
      'FixMyStreet::Roles::Extra',
@@ -932,7 +939,7 @@ sub updates_sent_to_body {
 sub add_send_method {
     my $self = shift;
     my $sender = shift;
-    ($sender = ref $sender) =~ s/^.*:://;
+    $sender =~ s/${\SENDER_REGEX}//;
     if (my $send_method = $self->send_method_used) {
         $self->send_method_used("$send_method,$sender");
     } else {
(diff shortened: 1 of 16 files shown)


dracos (commented):
In terms of the overall idea, the only concern I can think of (and it's definitely no worse than now! Just wondering if we can fix it as part of this) is the presentation of whensent in the front end. Currently it will say sent to both even if it's only been sent to one; and that remains the case with the changes here. Could we perhaps, instead of storing the failed send types in a new column, we store the *bodies* that have been failed instead? I think that could hopefully be treated in the same way pretty much in terms of the changes you've made, but then would give us additional information we could use to not say it's been sent to somewhere it hasn't.

I've documented making database schema changes better for you, hope that's helpful. I think an ARRAY might be better than a string (I realise we historically have strings elsewhere but we now also have ARRAYs so i think using that if appropriate is fine), I guess you'd probably want non null default empty so there's no annoying "checking empty or null" to worry about.

As well as the `resend` function I mention, I think we would also need to make sure these reports would appear in the /admin/ waiting to be sent table and the send-failure-summary script - so we don't have hidden half-sent reports that keep trying for weeks without anyone noticing.

Hope that's helpful.

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -235,6 +262,8 @@ sub _send {
         }
     }
 
+    $self->report->add_send_fail_type($_) for @send_fail_types;
```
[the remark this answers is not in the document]
nephila-nacrea: There was a reason but I need to remember what it was.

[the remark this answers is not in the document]
nephila-nacrea: If doing it inside the loop and, say, you're looping through send methods (Open311, Email):
- if Open311 fails, it will be added to `send_fail_body_ids` [I changed to that from `send_fail_type`] - no problems there
- but then in the next iteration of the loop, we're looking at Email, and this block will cause sending to skip as we now have a populated send_fail_body_ids array but no 'Email' body in it:
`if ( @{ $report->send_fail_body_ids }
            && !$report->has_given_send_fail_body_id($body_id) )
        {
            $sender->success(1);
            next;
        }`

nephila-nacrea:
> In terms of the overall idea, the only concern I can think of (and it's definitely no worse than now! Just wondering if we can fix it as part of this) is the presentation of whensent in the front end. Currently it will say sent to both even if it's only been sent to one; and that remains the case with the changes here. Could we perhaps, instead of storing the failed send types in a new column, we store the _bodies_ that have been failed instead? I think that could hopefully be treated in the same way pretty much in terms of the changes you've made, but then would give us additional information we could use to not say it's been sent to somewhere it hasn't.
> 
> I've documented making database schema changes better for you, hope that's helpful. I think an ARRAY might be better than a string (I realise we historically have strings elsewhere but we now also have ARRAYs so i think using that if appropriate is fine), I guess you'd probably want non null default empty so there's no annoying "checking empty or null" to worry about.
> 
> As well as the `resend` function I mention, I think we would also need to make sure these reports would appear in the /admin/ waiting to be sent table and the send-failure-summary script - so we don't have hidden half-sent reports that keep trying for weeks without anyone noticing.
> 
> Hope that's helpful.

Regarding the presentation of `whensent`, whereabouts should I be worrying about this?
I also noticed on the 'Waiting to be sent' table under /admin, a report can now have a `whensent` if a send method succeeded but another failed. I wonder if I should include a 'failed send method' in the summary there.

dracos:
"Regarding the presentation of whensent, whereabouts should I be worrying about this?" - the main one I am thinking of is "Sent to %s %s later" in Result/Problem.pm's `duration_string` which is displayed on a report page. This would, for a report to Body A and Body B, as soon as Body A is successfully sent but Body B fails, say "Sent to Body A and Body B 4 minutes later". So it should ignore failed bodies when constructing this string, presumably (doesn't have to do anything about the failed one, really, that would maintain current behaviour).

Your idea on the admin table sounds good.

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -216,13 +216,44 @@ sub _send {
...
+    my @remove_send_fail_body_ids;
+
+    for my $sender_key ( keys %{ $self->reporters } ) {
+        my $sender = $self->reporters->{$sender_key};
+
+        # Skip if no body ID.
+        # (Lack of body at this stage shouldn't actually be possible,
+        # presumably).
+        my $body_id;
+        $body_id = $sender->bodies->[0]->id
```
dracos, line 231: Hmm, I've just thought of an issue. As this line implies, a sender can have multiple bodies (all the bodies for the report associated with a particular SendReport class). Say a report was sent to two bodies, both with email, or both with Open311 - there'd be only one sender covering both bodies, but this would only deal with one of them. (Historically, this is for e.g. if a report was sent to two bodies both by email, it would send one email with both in the To line. It doesn't make as much sense for eg the Open311 sender which has to loop through the bodies again anyway.)

We'd really want separate results for two Open311 senders in this situation, I think (say one succeeded, one failed - we've not yet been in this situation, the common case so far is one Open311, one email), so perhaps the easiest way to deal with this is to have a different reporter for each body, not only one per sender (perhaps with an email exception for the multiple-email case mentioned above).

dracos (changes requested):
The changes all look good, thanks, bar a couple of small things and I'm afraid one bigger thing that will need resolving to do with senders with multiple bodies. Shout to have a chat about it :)

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -251,6 +288,9 @@ sub _post_send {
     if (@errors) {
         $self->report->update_send_failed( join( '|', @errors ) );
     }
+    else {
+        $self->report->send_fail_reason(undef);
```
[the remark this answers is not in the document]
nephila-nacrea: I think my line of thinking was 'if it overwrites an old fail message upon a new failure, it makes sense to delete the message on success', but I can remove this change 👍

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -216,13 +216,44 @@ sub _send {
...
+    my $report = $self->report;
+
+    my @add_send_fail_body_ids;
+    my @remove_send_fail_body_ids;
+
+    for my $sender_key ( keys %{ $self->reporters } ) {
+        my $sender = $self->reporters->{$sender_key};
+
+        # Skip if no body ID.
+        # (Lack of body at this stage shouldn't actually be possible,
```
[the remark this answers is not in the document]
nephila-nacrea: I have tried to fix this; unsure whether my solution is any good!

t/sendreport/multiple-send-methods.t
```diff
@@ -0,0 +1,516 @@
...
+        {   cobrand  => 'fixmystreet',
+            category => 'Other',
+            user     => $user,
+        },
+    );
+
+    # This forces email to be sent to user
+    # (see perllib/FixMyStreet/SendReport/Email.pm->send())
+    # CHECKME Is this a sensible thing to do?
+    FixMyStreet->test_mode(0);
```
nephila-nacrea, line 290: Not sure I should be doing this.

perllib/FixMyStreet/Queue/Item/Report.pm
nephila-nacrea: I believe I have now handled this (including the email exception); not entirely confident about it though.

t/sendreport/multiple-send-methods.t
dracos: The test suite runs with staging set and send_reports defaults to 0, so the test_mode flag is in SendReport::Email in order for most tests to be able to test sending emails. I'm not sure it's vital to test this non-test little bit of code, but I can't see any harm in having it - though if anything, this particular test would be in t/sendreport/email.t I guess, it's not testing multiple send methods at this point.

dracos (changes requested):
A few small comments/bugs, and I'm afraid I don't think the change for senders with multiple bodies works. Let me know if you want to talk it over. Could you try and fix each issue/theme in a separate commit where possible as that makes it easier to review, thanks :)

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -909,18 +924,72 @@ sub updates_sent_to_body {
...
     my $self = shift;
     my $sender = shift;
-    ($sender = ref $sender) =~ s/^.*:://;
-    if (my $send_method = $self->send_method_used) {
-        $self->send_method_used("$send_method,$sender");
-    } else {
+    $sender =~ s/${\SENDER_REGEX}//;
+
+    if ( my $existing_send_method = $self->send_method_used ) {
+        # Make sure we don't duplicate send methods
```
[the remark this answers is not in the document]
nephila-nacrea: IIRC, this is a hangover from a previous attempt where I made a regex to use in a new method, and realised I could just share the regex between the new method and this existing one. But then I changed what I did in the new method. I could probably just remove this.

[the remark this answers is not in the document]
nephila-nacrea: Ugh I misunderstood, you were specifically referring to the check against duplicates. Have removed that now.

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -202,39 +219,105 @@ sub _create_reporters {
...
+
+        if ( FixMyStreet->staging_flag( 'send_reports', 0 ) ) {
+            # If on staging with send_reports set to 0, bypass the body_id
+            # logic below, as we may only have a special Email reporter with
+            # no bodies attached
+            my $res = $sender->send( $report, $self->h );
+            $result *= $res;
+            $report->add_send_method($sender_name) if !$res;
+        }
+        elsif ( $sender_name =~ /${\EMAIL_SENDER_PREFIX}/ ) {
```
dracos, line 255: I think this and the else can be combined - as you say in the comment in the else, the senders there will only have one body, so the code for this block will work the same for them too, unless I've missed something.

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -202,39 +219,105 @@ sub _create_reporters {
...
-        my $res = $sender->send( $self->report, $self->h );
-        $result *= $res;
-        $self->report->add_send_method($sender) if !$res;
         if ( $self->manager ) {
-            if ($sender->unconfirmed_data) {
-                foreach my $e (keys %{ $sender->unconfirmed_data } ) {
-                    foreach my $c (keys %{ $sender->unconfirmed_data->{$e} }) {
-                        $self->manager->unconfirmed_data->{$e}{$c}{count} += $sender->unconfirmed_data->{$e}{$c}{count};
-                        $self->manager->unconfirmed_data->{$e}{$c}{note} = $sender->unconfirmed_data->{$e}{$c}{note};
+            if ( $sender->unconfirmed_data ) {
```
dracos, line 295: This reformatting (automated?) is nothing to do with the fixup or the PR, I'd rather not have it if possible, thanks (in this particular case, I also don't think smaller multiple lines is clearer than slightly longer single lines, especially with `{count}` on its own line).

t/sendreport/two-tier/email-and-open311.t
```diff
@@ -0,0 +1,379 @@
...
+use FixMyStreet::TestMech;
+use Test::Deep;
+use Test::MockModule;
+use Test::MockObject;
+
+my $mech = FixMyStreet::TestMech->new;
+
+my $user = $mech->create_user_ok('user@example.com');
+
+my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council' );
```
dracos, line 12: Given the mapit changes now on master, these will now need cobrand attributes adding.

t/sendreport/two-tier/email-and-open311.t
```diff
@@ -0,0 +1,379 @@
...
+
+my $cobrand = FixMyStreet::Cobrand->get_class_for_moniker('fixmystreet')->new;
+$report->result_source->schema->cobrand($cobrand);
+
+my $mock_email   = Test::MockModule->new('FixMyStreet::SendReport::Email');
+my $mock_open311 = Test::MockModule->new('FixMyStreet::SendReport::Open311');
+my $mock_log     = Test::MockObject->new;
+$mock_log->mock( debug => sub { } );
+my $mock_catalyst = Test::MockObject->new;
+$mock_catalyst->mock( log => sub {$mock_log} );
```
dracos, line 53: Is there a way not to need this mock_catalyst thing (and the constant re-setting of it)? As I understand it, this is needed because you're testing `duration_string` - I don't think this testing code needs to worry about that, can it just check `$problem->body`?

t/sendreport/two-tier/email-and-open311.t
```diff
@@ -0,0 +1,379 @@
...
+
+subtest '1st attempt - email and Open311 both fail' => sub {
```
dracos, line 55: Not asking you to change any of this now, but there seems to be a lot of repetition in these tests, that perhaps could be made clearer with some form of loop and some set up variables, or similar? Maybe it's just me but I find that easier to follow/read than trying to see the slight differences between many copies of similar tests.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -737,7 +744,7 @@ sub nearest_address {
 }
 
 sub body {
-    my ( $problem, $link ) = @_;
+    my ( $problem, $link, $successful_send_only ) = @_;
```
dracos, line 747: Sorry for not mentioning this before, but only noticed as part of checking for this review - I don't think you need this additional argument, because it's never called with it not set? Worth a comment above the `body` function explaining it returns a string of the successfully sent body names, but I don't see a need to provide a variable that's not used in more than one way.

dracos (changes requested):
Nearly there :) The changes to deal with the main problem of the multiple bodies/sender methods looks good - I think there can be a bit of simplification there, and a couple of other bits of suggested simplification elsewhere too, and I think that's it :)

dracos (changes requested):
Nearly there :) The changes to deal with the main problem of the multiple bodies/sender methods looks good - I think there can be a bit of simplification there, and a couple of other bits of suggested simplification elsewhere too, and I think that's it :)

perllib/FixMyStreet/Queue/Item/Report.pm
nephila-nacrea: This doesn't work as an email sender can have multiple bodies, but we then add those to the 'To' line and only send the one email. Using the logic in the `else` means multiple emails are sent.

nephila-nacrea: I mean, there is otherwise a lot of similarity between the blocks of code so maybe there is a more succinct way of writing it, but this keeps the distinction fairly clear.

t/sendreport/two-tier/email-and-open311.t
nephila-nacrea: I am not sure it is worth doing now because it could take me a while 😮‍💨 Since the order of the tests is important I guess I thought this was the clearest way of writing them.

t/sendreport/two-tier/email-both.t
```diff
@@ -0,0 +1,299 @@
...
+my $user = $mech->create_user_ok('user@example.com');
+
+my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council',
+    {}, { cobrand => 'oxfordshire' } );
+my $body_cherwell
+    = $mech->create_body_ok(
+        2419,
+        'Cherwell District Council',
+        {},
+        { cobrand => 'oxfordshire' },
```
dracos, line 19: This should not have a cobrand, because this body is not associated with any cobrand.

t/sendreport/two-tier/email-and-open311.t
```diff
@@ -9,9 +9,15 @@ my $mech = FixMyStreet::TestMech->new;
...
-my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council' );
+my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council',
+    {}, { cobrand => 'oxfordshire' } );
 my $body_cherwell
-    = $mech->create_body_ok( 2419, 'Cherwell District Council' );
+    = $mech->create_body_ok(
+        2419,
+        'Cherwell District Council',
+        {},
+        { cobrand => 'oxfordshire' },
```
dracos, line 19: Ditto

t/sendreport/two-tier/open311-both.t
```diff
@@ -9,9 +9,15 @@ my $mech = FixMyStreet::TestMech->new;
...
-my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council' );
+my $body_oxf = $mech->create_body_ok( 2237, 'Oxfordshire County Council',
+    {}, { cobrand => 'oxfordshire' } );
 my $body_cherwell
-    = $mech->create_body_ok( 2419, 'Cherwell District Council' );
+    = $mech->create_body_ok(
+        2419,
+        'Cherwell District Council',
+        {},
+        { cobrand => 'oxfordshire' },
```
dracos, line 19: And one more :)

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -202,27 +219,78 @@ sub _create_reporters {
...
+
+            my @body_ids = map { $_->id } @{ $sender->bodies };
+            if ($res) {
+                push @add_send_fail_body_ids, @body_ids;
+            }
+            else {
+                $report->add_send_method($sender_name);
+                push @remove_send_fail_body_ids, @body_ids;
+            }
+        }
```
dracos, line 272: Don't generally mind, but could the `else`s etc be on the same line as the braces, as per our coding standards (and the rest of this file), thanks.

perllib/FixMyStreet/Queue/Item/Report.pm
dracos: I meant the opposite of that (which would indeed do as you say) - by "this block" I meant the block I left the comment on. I think you can use the same code from the 'email' part for both it and the else part, ie:

```
if ( FixMyStreet->staging_flag( 'send_reports', 0 ) ) {
    ...
} else {
    my $res = $sender->send( $self->report, $self->h );
    $result *= $res;
    my @body_ids = map { $_->id } @{ $sender->bodies };
    if ($res) {
        push @add_send_fail_body_ids, @body_ids;
    } else {
        $report->add_send_method($sender_name);
        push @remove_send_fail_body_ids, @body_ids;
    }
}
```
There's no need for the else loop, because all the reports that would have gone there only have one body.

dracos (changes requested):
Couple of small things :)

perllib/FixMyStreet/Queue/Item/Report.pm
nephila-nacrea: I've tried to fix this but I may have missed some.

perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -202,27 +225,55 @@ sub _create_reporters {
...
+    my @remove_send_fail_body_ids;
+
+    for my $sender ( @{ $self->reporters } ) {
+        my $sender_name = ref $sender;
+        $self->log( 'Sending using ' . $sender_name );
+
+        if ( FixMyStreet->staging_flag( 'send_reports', 0 ) ) {
+            # If on staging with send_reports set to 0, bypass the body_id
+            # logic below, as we may only have a special Email reporter with
+            # no bodies attached
```
dracos, line 256: Sorry, now I've seen this with the most recent changes, I can also see that I don't think this block is needed either now :) The `else` block will work fine regardless of the `send_reports` setting - in the case there's a special Email reporter as per this comment, @body_ids is empty and it all works as you'd expect.

dracos (changes requested):
One more thing? :) Sorry, I only realised this when I saw the changed code laid out.

templates/web/base/admin/problem_row.html
```diff
@@ -37,6 +37,8 @@
         <td>[% prettify_state(problem.state, 1) %]<br><small>
             [% loc('Created') %]:&nbsp;[% PROCESS format_time time=problem.created %]
             <br>[% loc('When sent') %]:&nbsp;[% PROCESS format_time time=problem.whensent %]
+            [%- send_fail_bodies = problem.send_fail_bodies -%]
+            [%- IF send_fail_bodies %]<br>[% loc('Failed bodies:') %]&nbsp;[% send_fail_bodies.join(', ') %][% END -%]
```
dracos, line 41: This needs a `.size` in the `IF` otherwise it is always true and displays a blank line.

dracos (approved):
🎉  (There is one minor change to the template needed to not always output a blank line, but that doesn't need a re-review when that's changed)

PR #3831, opened 2022-03-07
[Buckinghamshire] Dashboard dropdown for parishes

chrismytton:
This adds an extra "Body" dropdown to the Buckinghamshire dashboard which in addition to Buckinghamshire Council also lists all of the active parishes associated with the Buckinghamshire cobrand. When selected allows viewing the parish bodies stats in the same way as you can with the primary cobrand body.

Fixes https://github.com/mysociety/societyworks/issues/2810

image

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -187,6 +187,12 @@ sub dashboard_export_problems_add_columns {
     shift->_dashboard_export_add_columns(@_);
 }
 
+sub dashboard_extra_bodies {
+    my ($self) = @_;
+
+    return $self->parish_bodies->all;
+}
+
 sub _parish_ids {
     # This is a list of all Parish Councils within Buckinghamshire,
     # taken from https://mapit.mysociety.org/area/2217/covers.json?type=CPC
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -703,14 +709,20 @@ sub area_ids_for_problems {
...
+
 # Show parish problems on the cobrand.
 sub problems_restriction_bodies {
     my ($self) = @_;
 
-    my @parishes = FixMyStreet::DB->resultset('Body')->search(
-        { 'body_areas.area_id' => $self->_parish_ids, },
-        { join => 'body_areas' }
-    )->all;
+    my @parishes = $self->parish_bodies->all;
```
dracos, line 725: Just noticing from what you've put above, but should this be `->active` as well?

dracos (approved):
Looks fine, I had one query about active is all.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
chrismytton: Ah yes, it probably should! I've updated `sub parish_bodies` to always call `->active`.

PR #3833, opened 2022-03-08
[Peterborough] Remove Covid message

MorayMySoc:
* Remove Covid message so emergency message feature works again

This are requested for staging for the time being.
https://mysocietysupport.freshdesk.com/a/tickets/1865

[skip changelog]


dracos (approved):
Looks good, sorry for slightly annoying tests :)

MorayMySoc:
Sorry it's not a fixup - did an amend on the commit message without realising that was committing the staged files. This is a change as stepping back from adding the emergency message as too much further development would be required, but retaining the removing of the permanent emergency message.

dracos (approved):

PR #3835, opened 2022-03-09
[FMS] Email alert dark mode

### LINKED ISSUE #3643: Links in alert emails hard to read
We've had a report from a user that the blue links in alert emails are hard to read.

They provided this screenshot:

image0

It looks like they have dark mode enabled, which could be related?

lucascumsille:
This pull request adds css Dark mode query:
Added a rule that forces the links to keep the same colour in dark and light mode.  

During the testing I also realised that there was a issue with the second column on the email alerts. Where the background colour was being displayed black in dark mode, but the text was still using a really dark colour, causing contrast issues. I added a fix that changes makes the background-colour to be dark, but the colour is now white on dark-mode.

Fixes #3643 

Apple email 15 Dark
appmail15dark-vertical-allowed-1366

Apple email 15
appmail15-vertical-allowed-1366


dracos (changes requested):
Not sure if your screenshot was before your fixup commit, or why that would have made any difference, but this branch doesn't appear to work in Apple Mail at present, I'm afraid: 

image

Can't recall how you're running the dev server, but there will be a way to run it without debugging so you don't have all the debug-toolbar stuff at the bottom, only the HTML of the email itself. (If that had an effect, I'd assume it wouldn't but can't be sure.)

lucascumsille:
That so weird, I just ran it without the debugging. And according to Litmus
This is what I get: 
appmail15dark-vertical-allowed-1366

Just in case I just added new commit that forces the first column to have a dark background, while the text should have a white colour.
appmail15dark-after commit

Let me know if the problem still persist.

templates/email/default/_email_settings.html
```diff
@@ -89,7 +95,9 @@
 
 only_column_style = "padding: ${ column_padding }px; vertical-align: top; background-color: $primary_column_background_color; color: $primary_column_text_color;"
 primary_column_style = "vertical-align: top; width: 50%; background-color: $primary_column_background_color; color: $primary_column_text_color;"
+primary_column_style_dark_mode = "vertical-align: top; width: 50%; background-color: $primary_column_background_color_dark_mode !important; color: $primary_column_text_color;"
```
dracos, line 98: Feels like this should set a white color as well, no point in setting something that won't work in case it is used anywhere (as I think it might be in the heading)

dracos (changes requested):
Just done this test - https://litmus.com/folders/unsorted_emails/emails/5934315/checklist - you can see the paragraphs are fixed but not the non-link part of the heading. I am including the `DOCTYPE` line in the test preview, given that is included in the email, I assume it is that that is making the difference?

templates/email/default/_email_settings.html
lucascumsille: I thought that would be the easier option and I was tempted to go in that direction. My reason to give it the dark background, was from user perspective, it's nice that if I have dark mode enabled, my email is actually is using dark mode, more than us just forcing the email to override the dark mode to look like the light version.
However I understand this is not a massive issue from a UX perspective. 
Let me know what you think. Happy to go with your option.

dracos: Sorry I wasn't clear - happy with the dark background, but you set a dark text colour here too - I was thinking that should change to white, not the background :)

lucascumsille: My bad I thought I had created already the color for the text in the first column. 
I just pushed a commit that adds white color in dark mode.
I just tested it and this time I added the DOCTYPE, that was causing the issue on my previews tests. 
appmail15dark

Let me know if there is anything else, thanks again for the feedback.

lucascumsille: Also I realised that by adding that last fix it also fixes the <h2> element.

templates/email/default/_email_settings.html
```diff
@@ -106,7 +115,8 @@
 list_item_style = "padding-bottom: 20px; margin-bottom: 20px; border-bottom: $list_item_border_bottom;"
 list_item_h2_style = "margin: 0 0 16px 0; font-size: 21px; line-height: 24px;"
 list_item_p_style = "margin: 0 0 16px 0;"
-list_item_date_style = "font-size: 14px; line-height: 20px; margin: 0; color: $color_gunmetal_light;"
+list_item_p_style_dark_mode = "margin: 0 0 16px 0; color: $color_white !important;"
```
dracos, line 118: Is this style needed now you've fixed the primary column text colour? `list_item_p_style` doesn't set a colour, so I'd hope a dark mode version wasn't needed, what do you think?

dracos (changes requested):
Has fixed the main issue, but sorry, just one more question :)

templates/email/default/_email_settings.html
lucascumsille: Not a problem, so technically is not needed, however that was preventing the user name below the problem to keep having the colour: gunmetal_light which has accessibility issues with the dark background colour. 
I wasn't too in love with the user name having the same colour with the problem heading(both white), so what I did in my recent commit ([e18b833](https://github.com/mysociety/fixmystreet/pull/3835/commits/e18b83361e8ad1a9f70147ddd9fe797b0f93c74e)) was to add a class to the user name, to be able to select that element and in dark mode instead of having a white colour, now will have: color_blue_pale. That passes the contrast test and it doesn't have the same colour as the heading. It's a subtle change, but I thought it would be better than creating a new variable (new colour) just for this particular style.

Let me know what you think

dracos: Okay, so you've fixed the username, so I guess my question still stands - is list_item_p_style_dark_mode needed any more? :)

dracos (changes requested):
One typo, I think, and one question :)

dracos (approved):
Woo, looks good :)

lucascumsille:
> Woo, looks good :)

Thanks for all the help, Matthew =) I'll merge it then.

PR #3836, opened 2022-03-09
[Peterborough] More waste fixes

dracos:
[skip changelog] My previous fix now breaks if there is a schedule without a next date (e.g. some brown bins at present). And the waste message wasn't displaying because they have their own header (I thought about factoring it out in some way but it's tiny and whatever).


chrismytton (approved):
📆

PR #3838, opened 2022-03-10
[Peterborough] Only include in service.

dracos:
[skip changelog]
For https://mysocietysupport.freshdesk.com/a/tickets/1904
Example address in PE6 9AJ

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -594,8 +594,8 @@ sub bin_services_for_address {
 
     # TODO parallelize these calls if performance is an issue
     my $jobs = $bartec->Jobs_Get($property->{uprn});
-    my $job_dates = $bartec->Jobs_FeatureScheduleDates_Get($property->{uprn});
     my $schedules = $bartec->Features_Schedules_Get($property->{uprn});
+    my $job_dates = relevant_jobs($bartec, $property->{uprn}, $schedules);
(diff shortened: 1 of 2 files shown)


perllib/Integrations/Bartec.pm
```diff
@@ -251,6 +251,18 @@ sub Jobs_FeatureScheduleDates_Get {
...
 }
 
+sub relevant_jobs {
+    my ($self, $uprn, $schedules) = @_;
+    my $jobs = $self->Jobs_FeatureScheduleDates_Get($uprn);
+    my %schedules = map { $_->{JobName} => $_ } @$schedules;
+    my @jobs = grep {
+        my $name = $_->{JobName};
+        $schedules{$name}->{Feature}->{Status}->{Name} eq 'IN SERVICE'
+        && $schedules{$name}->{Feature}->{FeatureType}->{ID} != 6815;
```
davea, line 261: This Peterborough-specific magic number feels a bit out of place here. Would this method be better in the `Peterborough.pm` cobrand module? Or at least a comment here explaining what it's doing and how it might need refactoring for the next Bartec integration.

davea (changes requested):
One small change please, but otherwise good to go.

perllib/Integrations/Bartec.pm
dracos: I actually had this function in Peterborough.pm originally but moved it to make it easier to not pass $bartec around, probably not a good basis for that decision :) Will move it back.

PR #3839, opened 2022-03-11
[Buckinghamshire] Lookup parishes on .com site for new reports

chrismytton:
When creating a new report we want .com to be able to route reports to parishes in the same way that the Buckinghamshire cobrand can. In order to achieve this we need to include "CPC" area type from mapit when looking for areas to create a new report in.

Fixes https://github.com/mysociety/societyworks/issues/2944

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -419,4 +419,17 @@ sub reopening_disallowed {
     return $self->next::method($problem);
 }
 
+# Make sure CPC areas are included in point lookups for new reports
+# This is so that parish bodies (e.g. in Buckinghamshire) are available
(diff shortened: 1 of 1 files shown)


dracos (approved):
Wondered idly if might be worth adding CPC to UK.pm's `area_types` and so we have it covered everywhere for future use, but looking at the code and seeing e.g. the comment about alert page stuff, this is probably better for now. It's probably something we will want to do, though, I assume.

chrismytton:
> Wondered idly if might be worth adding CPC to UK.pm's `area_types` and so we have it covered everywhere for future use, but looking at the code and seeing e.g. the comment about alert page stuff, this is probably better for now. It's probably something we will want to do, though, I assume.

Yeah I did briefly think about that but as you say, looks like it might need a bit more care and thought.

PR #3843, opened 2022-03-15
[Buckinghamshire] Show the roads layer for grass cutting categories

chrismytton:
This ensures that grass cutting reports are always made on a road.

Initially I tried simply adding the "Grass, hedges and weeds" category to the `asset_group` array, but that wasn't showing the stopper message because the `road_not_found` function checks for `is_single_body`, but there are actually two bodies in this case, Bucks and the parish, but we still want to show the stopper message. So I've added a `criterion` argument to `road_not_found` which the Bucks road layer then uses to always show the stopper message for the grass cutting categories.

[skip changelog]


davea (approved):
🍀 🆗

PR #3844, opened 2022-03-16
Add Cypress method for picking subcategories

chrismytton:
This was originally part of https://github.com/mysociety/fixmystreet/pull/3832, and the code has already been reviewed, but I'm pulling it into a separate PR because I want to use the new pickSubcategory method in https://github.com/mysociety/fixmystreet/pull/3843.

There was already a `cy.pickSubcategory` method, but the name was misleading because it was actually picking a subcategory from within extra data, which isn't how most cobrands work.

This change renames the previous method to pickSubcatExtraInfo to more accurately reflect what it's doing, then the pickSubcategory method has been updated to actually pick a regular subcategory given a category ID and a subcategory label. I've also gone through and updated the obvious places where this pattern was being used to use the new method.

[skip changelog]


davea (approved):
👍

PR #3845, opened 2022-03-16
[Buckinghamshire] Grass cutting routing for .com reports

chrismytton:
Bucks has some special code to ensure grass cutting reports are routed either to Bucks or to the parish. This code needs to be run on .com as well to ensure reports made on there are routed in the same way.

[skip changelog]


davea (approved):

chrismytton:
Turns out this didn't quite work, the working version is e73401abe13246f14ec711844de6ba2960a2c316.

PR #3847, opened 2022-03-16
Added "Can’t use the map? Skip this step." to report problem here

lucascumsille:
This PR fixes the issue where users that are on /around page then click on "Report a problem here" on the navigation bar, and there wasn't a "Can’t use the map? Skip this step." button (same that we have implemented on /report/new) causing accessibility issues.

This has been implemented across all cobrands.

This fixes part of: mysociety/societyworks#2787

Screenshot 2022-03-16 at 11 35 52

Do I need to updated the changelog?

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -89,10 +89,10 @@ sub open311_extra_data_exclude {
 
 sub lookup_site_code_config { {
     buffer => 50, # metres
-    url => "https://tilma.mysociety.org/mapserver/peterborough",
-    srsname => "urn:ogc:def:crs:EPSG::27700",
-    typename => "highways",
-    property => "Usrn",
+    url => 'https://peterborough.assets/7/query?',
+    type => 'arcgis',
+    outFields => 'USRN',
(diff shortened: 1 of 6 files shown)


dracos (changes requested):
As per the associated ticket description, could it be the middle option, not the top, and could it be only visible to screen readers and receives focus when using keyboard navigation, thanks.

lucascumsille:
@dracos  This latest commit would only allow screen reader and keyboard users to have access to skip map button.
Let me know if this solution is alright.
I thought about adding the "screen-reader-only" class, so we can reuse whenever we need something similar.

dracos (changes requested):
Thanks, looking better - but would be good for it appear on focus similar to how `.map-alternatives` does, I guess?

lucascumsille:
> 
Thanks for the advice, Matthew
I just push a --fixup that does the same that `.map-alternatives`, but without all the extra styling.
I also added it to the list of elements that shouldn't be print out when printing the page.

dracos (approved):
Looks good; I've added setting the `url_skip` variable as if you go direct to /report/new rather than go via /around it would not have been set at the point the template was used.

PR #3848, opened 2022-03-16
[Buckinghamshire] Use hard-coded IDs when creating parishes

chrismytton:
Previously this was looking up the list of children for the Unitary authority, but this meant that the parishes created were slightly different to the list of parishes in the "Areas covered" dropdown because the dropdown uses the hard-coded list of parishes in the cobrand.

The impact of this is that "Bierton Parish Council" wasn't associated with an area, because the area ID for the new "Bierton" is different to the one in the `_parish_ids` list in Buckinghamshire.pm.

This changes the create-parish-bodies script to use the same list of parish IDs that's used elsewhere, rather than deriving a list from the area children.

Part of https://github.com/mysociety/societyworks/issues/2938

[skip changelog]

### DIFF
--- a/bin/buckinghamshire/create-parish-bodies
+++ b/bin/buckinghamshire/create-parish-bodies
@@ -14,6 +14,7 @@ BEGIN {    # set all the paths to the perl code
 
 use mySociety::MaPit;
 use FixMyStreet::DB;
+use FixMyStreet::Cobrand::Buckinghamshire;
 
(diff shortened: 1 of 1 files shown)


davea (approved):
This looks fine as it is. We'll have to revisit as part of https://github.com/mysociety/societyworks/issues/1951 (Get FixMyStreet on current MapIt generation) if the children of the unitary don't match up to the hardcoded list that was made for the now-defunct county council.

PR #3850, opened 2022-03-17
[Bromley] Don’t protect title dropdown field for staff

davea:
This was causing issues because setting `disabled` seemed to stop UAs submitting any value for the title field, which is required for Bromley reports.
`readonly` can’t be used with <select> elements, so just drop these lines entirely.

[skip changelog]


dracos (approved):
🎫

PR #3852, opened 2022-03-17
[Buckinghamshire] Remove text about emailing reference number

chrismytton:
When sending a report to a non-open311 endpoint, hide the message about sending a reference number.

## Todo

- [ ] Write test to check that text is/isn't present for email/open311 categories

Fixes https://github.com/mysociety/societyworks/issues/2956

[skip changelog]


chrismytton:
This was done in a better way in https://github.com/mysociety/fixmystreet/pull/3861

PR #3853, opened 2022-03-18
[Peterborough] Switch to their highways layer.

dracos:
They've added this to their ArcGIS so we can use it instead, it's more up to date. [skip changelog]
Should fix https://mysocietysupport.freshdesk.com/a/tickets/1918


davea (approved):
⚽ 🥅

PR #3854, opened 2022-03-18
[Zürich] Don’t allow DMs to provide public update on internal reports

davea:
For https://mysocietysupport.freshdesk.com/a/tickets/1931

[skip changelog]


dracos (approved):
🍡

PR #3857, opened 2022-03-22
[Bexley] Show report ID in confirmation emails.

MorayMySoc:
As requested by the client, this adds the report ID to the questionnaire email sent to a user about a report, and adds a report logged email so the user gets a report ID from that.

https://github.com/mysociety/societyworks/issues/2775

[skip changelog]

### DIFF
--- a/t/cobrand/bexley.t
+++ b/t/cobrand/bexley.t
@@ -5,6 +5,10 @@ use FixMyStreet::TestMech;
 use FixMyStreet::Script::Reports;
 use Catalyst::Test 'FixMyStreet::App';
 
+# disable info logs for this test run
+FixMyStreet::App->log->disable('info');
+END { FixMyStreet::App->log->enable('info'); }
+
 set_fixed_time('2019-10-16T17:00:00Z'); # Out of hours
 
 use_ok 'FixMyStreet::Cobrand::Bexley';
@@ -150,7 +154,9 @@ FixMyStreet::override_config {
             }
 
             if (my $t = $test->{email}) {
-                my $email = $mech->get_email;
+                my @emails = $mech->get_email;
+                # User is getting a report_sent_confirmation_email now and the ordering is random
+                my ($email) = grep { $mech->get_text_body_from_email($_) =~ /Dear London Borough of Bexley/ } @emails;
                 $t = join('@[^@]*', @$t);
                 is $email->header('From'), '"Test User" <do-not-reply@example.org>';
                 like $email->header('To'), qr/^[^@]*$t@[^@]*$/;
@@ -164,7 +170,7 @@ FixMyStreet::override_config {
                 }
                 $mech->clear_emails_ok;
             } else {
-                $mech->email_count_is(0);
+                $mech->email_count_is(1);
             }
         };
     }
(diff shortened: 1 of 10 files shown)


templates/email/bexley/_problem-confirm_extra.html
```diff
@@ -0,0 +1,4 @@
+</p>
+<p style="[% p_style %]">
+  The report's reference number is <strong>[% report.id %]</strong>.
```
davea, line 3: I have a slight concern about including a reference number here, as the report isn't actually confirmed and hasn't been sent at this point. Giving a reference number to the user might lead them to believe they don't need to take any further action, but in fact the one thing they really need to do with this email is click the button.

davea (commented):
Code wise this looks good but have left one comment that we should discuss.

templates/email/bexley/_problem-confirm_extra.html
MorayMySoc: Do you think creating a fragment lower down - that appears after the link - which says something along the lines of "If you confirm your report, the report id will be xxxx if you wish to contact the council directly about it at a later date"

davea: I don't think we should display an ID at all for unconfirmed reports, to be honest. If we want to ensure the user always gets an ID when the report is sent you can use the `sub report_sent_confirmation_email { 'id' }` cobrand method, like we do for e.g. BathNES, which sends an email after the report's been sent.

t/cobrand/bexley.t
```diff
@@ -261,6 +267,40 @@ FixMyStreet::override_config {
         $mech->get_ok('/report/' . $report->id);
         $mech->content_contains('Report ref:&nbsp;' . $report->id);
     };
```
nephila-nacrea, line 269: Just an FYI that rebasing with master produces a conflict around here.

templates/email/bexley/_council_reference.html
```diff
@@ -0,0 +1,3 @@
+<p style="[% p_style %]">The report's reference number is <strong>[% problem.id %]</strong>.
```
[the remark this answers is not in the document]
MorayMySoc: `sub report_sent_confirmation_email { 'id' }` that Dave suggested using then triggers `_dev/email/other-reported` (templates/email/default/other-reported.html) to be sent, which incorporates this fragment, if it exists. The idea is to send an email to users confirming their report has been made - which allows us to put in the id number.

nephila-nacrea (changes requested):
Again, I may be misunderstanding the requirements but it looks like you are missing a few things still:
- For the questionnaire email, a report made on the Bexley site gets the 'ref' text, but a Bexley report made on the FMS site does not
- There don't seem to be changes made for the confirmation email a user receives when a staff member posts an update on the user's report (for either the Bexley site or FMS)
- On the other hand, a user making an update on a report on the Bexley site does get an email with the 'ref' line - but is this part of the requirements? (And if so, a user updating a Bexley report on FMS is *not* getting an email with 'ref')

I hope that makes sense.

t/cobrand/bexley.t
```diff
@@ -5,6 +5,10 @@ use FixMyStreet::TestMech;
 use FixMyStreet::Script::Reports;
 use Catalyst::Test 'FixMyStreet::App';
 
+# disable info logs for this test run
```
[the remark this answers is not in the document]
dracos: It prevents the tests issuing log lines to the visible output such as the `$c->log->info` lines in Report/New.pm

dracos:
(PS Make sure commit messages match what's being changed :) )

MorayMySoc:
Merged, but forgot to push up rebased version for auto-commit

PR #3858, opened 2022-03-22
[Hackney] Alloy environmental services integration

davea:
Some changes/improvements to support the Alloy environmental services integration launch.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hackney.pm
+++ b/perllib/FixMyStreet/Cobrand/Hackney.pm
@@ -306,7 +306,7 @@ sub update_email_shortlisted_user {
     my $c = $self->{c};
     my $cobrand = FixMyStreet::Cobrand::Hackney->new; # $self may be FMS
     return if !$update->problem->to_body_named('Hackney');
-    my $sent_to = $update->problem->get_extra_metadata('sent_to');
+    my $sent_to = $update->problem->get_extra_metadata('sent_to') || [];
     if (@$sent_to) {
         my @to = map { [ $_, $cobrand->council_name ] } @$sent_to;
         $c->send_email('alert-update.txt', {
--- a/perllib/FixMyStreet/Cobrand/Hackney.pm
+++ b/perllib/FixMyStreet/Cobrand/Hackney.pm
@@ -143,6 +143,14 @@ around open311_extra_data_include => sub {
         $contact->email($code) if $code;
     }
 
+    if ($row->geocode) {
+        my $address = $row->geocode->{resourceSets}->[0]->{resources}->[0]->{address};
+        push @$open311_only, (
+            { name => 'closest_address', value => $address->{formattedAddress} }
+        );
+        $h->{closest_address} = '';
+    }
+
     return $open311_only;
 };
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -243,6 +243,13 @@ sub munge_sendreport_params {
     }
 }
 
+sub open311_munge_update_params {
```
dracos, line 299: I was going to say there is a FixMyStreet::Roles::Open311Multi you can include, but I guess that doesn't work unless Hackney had a function already, hmm. Perhaps the role function could be a normal function that calls super instead? Doesn't seem to need to be an "around" way of doing it, I mean.

perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -143,6 +143,14 @@ around open311_extra_data_include => sub {
         $contact->email($code) if $code;
     }
 
+    if ($row->geocode) {
+        my $address = $row->geocode->{resourceSets}->[0]->{resources}->[0]->{address};
+        push @$open311_only, (
+            { name => 'closest_address', value => $address->{formattedAddress} }
+        );
+        $h->{closest_address} = '';
```
dracos, line 151: Ah, clever. (CheshireEast/Oxfordshire/Rutland do not do this, wonder if they should? Or if it's working anyway somehow.)

perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -251,6 +251,41 @@ sub munge_sendreport_params {
...
+    # have manually-created extra fields that should be appended to the
+    # report title when submitted via Open311.
+    # This fetches the field names from the COBRAND_FEATURES config, so
+    # if they change in the future it doesn't require code changes to deploy.
+    my $field_names = $self->feature('environment_extra_fields') || [];
+    my %fields = map { $_ => 1 } @$field_names;
+    if ($contact->email =~ /^Environment/) {
+        for (@{ $row->get_extra_fields }) {
+            if ($fields{$_->{name}}) {
+                $title .= "\n\n" . $_->{description} . "\n" . $_->{value};
```
dracos, line 269: Could do with a test of this.

perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -282,6 +282,16 @@ sub open311_extra_data_include {
           value => $row->category },
     ];
 
+    if ($contact->email eq "Environment-Graffiti") {
+        push @$open311_only,
+            { name => 'graffiti_offensive', value => $row->get_extra_field_value('offensive') || '' };
```
dracos, line 287: Could the extra field values match the names needed by open311 adapter to save this being needed? (Or the names in open311-adapter changed to match the FMS values?) Or am I missing something?

dracos (changes requested):
Only a few small requests/queries again :)

perllib/FixMyStreet/Cobrand/Hackney.pm
davea: Yeah I'm not sure what my thinking was, in hindsight. Think I wanted to preserve the field names from live for some reason but make it obvious what they were for in open311-adapter. Anyway have reverted this and used the `graffiti_`-prefixed field names in FMS.

dracos (approved):

PR #3859, opened 2022-03-24
[Peterborough] Don’t enable service checkboxes that start disabled

davea:
For https://mysocietysupport.freshdesk.com/a/tickets/1913

[skip changelog]


dracos (approved):
🦀

PR #3862, opened 2022-03-25
[Cheshire East] Inspector tool amendments

MorayMySoc:
* Moves 'assignee' information on /admin/reports page into separate column
* Adds dropdown with list of assignees for filtering
* Includes filters 'Unassigned', 'All' and any assignees for the body
* Does not interact with search - search searches all reports unfiltered
* Adds tests to check functionality
* Adds test to check doesn't appear for another cobrand

https://github.com/mysociety/societyworks/issues/2777

[skip changelog]

### DIFF
--- a/templates/web/base/admin/problem_row.html
+++ b/templates/web/base/admin/problem_row.html
@@ -44,6 +44,10 @@
             [%- IF problem.is_closed %]<br>[% prettify_state('closed') %]: [% PROCESS format_time time=problem.lastupdate %][% END -%]
             [%- IF problem.is_open AND problem.lastupdate != problem.whensent %]<br>[% loc('Last&nbsp;update:') %] [% PROCESS format_time time=problem.lastupdate %][% END -%]
         </small></td>
+        [% TRY %]
+            [% INCLUDE admin/_problem_row_assigned.html %]
+        [% CATCH %]
+        [% END %]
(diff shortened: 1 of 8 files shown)


templates/web/base/admin/problem_row.html
```diff
@@ -29,13 +29,15 @@
...
             [%- END -%]
             <br>[% problem.cobrand %]<br>[% problem.cobrand_data | html %]
-            <br>
-            [% IF c.user.has_body_permission_to('assign_report_to_user') %] [%# see report's assigned user %]
-                <span>[% loc('Assigned to') %]: </span>
-                <span>
-                    [%- problem.shortlisted_user.name OR problem.shortlisted_user.username OR loc('unassigned') -%]
-                </span>
-            [%- END -%]
+            [% IF !c.cobrand.moniker == 'cheshireeast' %]
```
davea, line 32: Similar to the Perl module, is there a way we can remove references to an individual cobrand here (and in the other template)? They're so enormous that you couldn't just override the entire template.

davea (changes requested):
Functionally this works great! But a few things to improve/tidy the code a little.

nephila-nacrea (changes requested):

dracos:
Not going ahead

PR #3867, opened 2022-04-01
[Zürich] Update translations

davea:
Took the opportunity to update all translations with new strings too.

For https://github.com/mysociety/societyworks/issues/2979 & FD-1977

[skip changelog]


dracos (changes requested):
Sorry, I fixed/reverted a change to the Show... about... translation so if it's easy to regenerate, that'd be good.

Issue #3868, opened 2022-04-05
Test failing due to timezone issues


MorayMySoc:
**Describe the bug**
The cobrand/hackney.t test is failing on my local server due to timezone issues on my local machine

**To Reproduce**
When I run the test file hackney.t,  'sends branded alert emails' subtest fails as the time is expected to be '13:00' but on my local machine it's reporting as '14:00'. This would seem to have happened after British Summertime clock change.

**Expected behavior**
Tests should pass in any environment/timezone

PR #3869, opened 2022-04-05
Make test for time in email a pattern match for confirming time is in email

MorayMySoc:
As test is not about what the time is, make the time a pattern match to confirm the time is in the email.

Previous test was failing as the timezone on my local server was changing the time by an hour after British Summertime clock changes.

Fixes: Test failing due to timezone issues #3868 

[skip changelog]


struan (approved):
🕐

Issue #3870, opened 2022-04-06
Dropdown for assigning reports to inspectors has repetitions of names


MorayMySoc:
**Describe the bug**
If an inspector has multiple roles assigned that repeat inspector permissions, their name is repeated in the dropdown list of inspectors a report can be assigned to. 

**To Reproduce**
Add two roles that both have inspector permissions in them and assign to a user. Looking at a report on that user's body shows the user name repeated in the dropdown list. 

**Expected behavior**
List should only have any inspector in it once.

**Screenshots**
Screenshot attache
![temp](https://user-images.githubusercontent.com/87022664/161997883-96e53e36-7ac4-42ee-9215-d539ab097fda.png)
d

MorayMySoc:
Client confirmed this removed duplication https://mysocietysupport.freshdesk.com/a/tickets/1982

PR #3871, opened 2022-04-06
staff_with_permission list returns list of users unique by id

MorayMySoc:
staff_with_permission has been collating user names if permissions repeated through assignment to multiple roles.

This filters the list to return unique users.

https://github.com/mysociety/fixmystreet/issues/3870

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Body.pm
+++ b/perllib/FixMyStreet/DB/Result/Body.pm
@@ -295,6 +295,7 @@ sub staff_with_permission {
         { permissions => \"@> ARRAY['$permission']" }
     ], {
         join => [ 'user_body_permissions', { "user_roles" => "role" } ],
+        distinct => 1,
         order_by => { '-asc' => ['name'] },
     });
 }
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/DB/Result/Body.pm
```diff
@@ -290,13 +290,16 @@ granted a specific permission.
 
 sub staff_with_permission {
     my ( $self, $permission ) = @_;
-    return $self->users->search([
+    my @staff = $self->users->search([
         { 'user_body_permissions.permission_type' => $permission },
         { permissions => \"@> ARRAY['$permission']" }
     ], {
         join => [ 'user_body_permissions', { "user_roles" => "role" } ],
         order_by => { '-asc' => ['name'] },
```
struan, line 298: Is there a reason not to use `distinct => 1` here and do the de-duplication in the database?

struan (approved):
one small question about an alternative way to do this, but otherwise, looks good.

perllib/FixMyStreet/DB/Result/Body.pm
MorayMySoc: Nope - far better if I've got that right.

struan (approved):

PR #3872, opened 2022-04-07
Pop up "Browse photos" button when reporting incident

lucascumsille:
The dropzone and link inside the drop zone area were too subtle. This fix should draw more attention to both elements.
Peterborough raised this incident; however, this has been implemented for all cobrands because they all had a similar issue.

Some cobrand needed special treatment because the default solution was accessible (contrast issues).
There is only one cobrand that I couldn't test, which was "highwaysengland."

Desktop view:
Screenshot 2022-04-07 at 13 25 26

Mobile view:
Screenshot 2022-04-07 at 13 27 29

Fixes: mysociety/societyworks#2579

### DIFF
--- a/web/cobrands/bexley/base.scss
+++ b/web/cobrands/bexley/base.scss
@@ -105,3 +105,21 @@ ol.big-numbers > li:before {
 #postcodeForm .form-hint {
 //    color: $white;
 }
+
+.dropzone {
+    border-color: $primary_b;
+    background: mix($primary_b, $white, 15%);
+}
+
+.dz-clickable {
+    .dz-message u {
+        background-color: $primary_b;
+        color: $white;
+        border: 1px solid $primary_b;
+
+        &:hover {
+            background-color: $white ;
+            color: $primary_b;
+        }
+    }
+}
--- a/web/cobrands/sass/_dropzone.scss
+++ b/web/cobrands/sass/_dropzone.scss
@@ -39,12 +39,22 @@ $dropzone-background-color--full: mix($dropzone-border-color--full, $dropzone-ba
   }
 
   .dz-message u {
-    //color: $dropzone-link-color;
-    @extend .btn-primary;
+    padding: 0.75em 1em;
+    text-decoration: none;
     display: inline;
     vertical-align: baseline;
     margin-left: 3px;
     font-weight: 400;
+    border-radius: 4px;
+    background: $primary;
+    color: #fff;
+    border: 1px solid $primary;
+
+    &:hover, &:focus {
+      background: #fff;
+      color: $primary;
+      border-color: $primary;
+    }
   }
 }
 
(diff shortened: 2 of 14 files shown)


lucascumsille:
Hi @ludovic-tc I'm not sure who should review this PR, I saw that you opened the issue, but please let me know if this should be review by someone else.

web/cobrands/bexley/base.scss
```diff
@@ -105,3 +105,21 @@ ol.big-numbers > li:before {
 #postcodeForm .form-hint {
 //    color: $white;
 }
+
+.dropzone {
```
dracos, line 109: Rather than having a large number of council specific overrides like this, which leads to larger CSS as the defaults are also included, it would be better to create any new $dropzone SASS variables for any cobrand that needs them, which would then make all these overrides unnecessary. For example, you could have a `$dropzone-border-colour` variable, defaulting to `$primary`, and any cobrand that needs something different can set `$dropzone-border-colour` in its colours.scss and that's all they have to do, no new actual CSS.

Relatedly, $dropzone-link-colour is now only used to set the dropzone-background-colour--awakened which doesn't feel ideal, could it be used as the button link or background colour, or dropped entirely.

web/cobrands/sass/_dropzone.scss
```diff
@@ -31,10 +35,26 @@ $dropzone-background-color--full: mix($dropzone-border-color--full, $dropzone-ba
...
-    color: $dropzone-link-color;
+    padding: 0.75em 1em;
+    text-decoration: none;
+    display: inline;
+    vertical-align: baseline;
+    margin-left: 3px;
+    font-weight: 400;
+    border-radius: 4px;
+    background: $primary;
+    color: #fff;
```
dracos, line 50: If the background is primary, the color by default should be $primary_text which should always be an inverse that displays on $primary. I see you've used this in a couple of overrides, but I think it should be the default, then any changes made as necessary.

web/cobrands/sass/_dropzone.scss
```diff
@@ -31,10 +35,26 @@ $dropzone-background-color--full: mix($dropzone-border-color--full, $dropzone-ba
...
+    border: 1px solid $primary;
+
+    &:hover, &:focus {
+      background: #fff;
```
dracos, line 54: Ditto, this should be $primary_text.

web/cobrands/bexley/base.scss
dracos: Looking at what's used in the overrides, it would look like almost all could be covered by having additional variables `$dropzone-border-colour`, `$dropzone-button-text` (or repurpose link-colour), `$dropzone-button-background` and `$dropzone-button-border`, I think?

dracos (changes requested):
Looks fine, but I think needs rejigging in how the cobrands set the colour information they need that isn't the default.

lucascumsille:
Thanks @dracos I'll take a look and apply the fixes.

dracos:
Superseded by #3990

PR #3875, opened 2022-04-08
Peabody FMS front end

lucascumsille:
Pull request that adds the styling to Peabody cobrand. This PR includes:

- Logo
- Colours
- Variables
- Mixins
- Basic element such as buttons and links.
- Email styling

It has been tested on the following browsers for Desktop:
- Firefox
- Chrome
- Safari
- Older versions of Safari (5.1)

On Mobile:
- Chrome
- Safari
- Samsung browser

Please let me know if you have any feedback


zarino (changes requested):
@lucascumsille I think you’ve mistakenly edited the `shropshire` cobrand directly, when you probably intended to _copy_ the `shropshire` cobrand into a new `peabody` cobrand, and then _edit that copy_.

We’ll need to extract your changes out into a separate `peabody` cobrand before any of this can be merged. Let me know if you need a hand doing that.

zarino (approved):
@MorayMySoc Ok so the styling here all looks good to me and @lucascumsille.

We’ve added two extra commits on the end that reset the Shropshire styles/templates back to normal, and move the Peabody styles/templates into their own folders, since that’s—I assume—something you’ll be doing anyway. Should hopefully save you a bit of time. If not feel free to disregard!

We’ve added them as extra commits, and not squashed any other commits, so it’s clearer what’s going on. But obviously you’ll want to squash everything down into a single "New cobrand for Peabody"-style commit, as normal, when you’re happy with it all.

lucascumsille:
@MorayMySoc I updated the fonts, but I haven't being able to test them. So whenever the cobrand is ready I'll check if the font are being properly injected.

MorayMySoc (changes requested):
In a fairly arbitrary decision, I've been making the directory 'thamesmead' rather than 'peabody' so if we could go with that, that would be great.

Can you also remove the assets.js file as that is for another cobrand because of the way I asked you to to do this.

On Firefox, I'm getting a link I can't see due to the colours (ie I can highlight it to read the text - it is there) when the 'That location is not covered by...' message comes up.

pleasevisit

MorayMySoc:
> @MorayMySoc I updated the fonts, but I haven't being able to test them. So whenever the cobrand is ready I'll check if the font are being properly injected.

I've pushed up the branch now and that should be fine for you to add the fonts

lucascumsille:
> In a fairly arbitrary decision, I've been making the directory 'thamesmead' rather than 'peabody' so if we could go with that, that would be great.
> 
> Can you also remove the assets.js file as that is for another cobrand because of the way I asked you to to do this.
> 
> On Firefox, I'm getting a link I can't see due to the colours (ie I can highlight it to read the text - it is there) when the 'That location is not covered by...' message comes up.
> 
> pleasevisit

For some reason I couldn't replicate the link issue on my local version, so I ended up replicating it on a live version and stole the classes that I should style. Hopefully now it should be working.

No probs about the folder name, I changed it to 'thamesmead'.

MorayMySoc (changes requested):
I've got this issue on the front page in FF...do you need more detail @lucascumsille ?

temp
.

lucascumsille:
> I've got this issue on the front page in FF...do you need more detail @lucascumsille ?
> 
> temp .

@MorayMySoc Do you know how to deactivate that footer? When I look for it on the code, I can see that some cobrands include footer-marketing.html, but it's not the case for Peabody. I also noticed that it's part of the base code, however not every cobrand has that. Then I thought somewhere in the settings, maybe?

templates/web/thamesmead/around/intro.html
```diff
@@ -0,0 +1,2 @@
+<h1>Report a Problem on a street/road</h1>
+<h2>Report a problem on street/road in Shropshire or track the progress of reported issues</h2>
```
MorayMySoc, line 2: Cobrand name

templates/web/thamesmead/before_wrapper.html
```diff
@@ -0,0 +1,10 @@
+<div class="cookie-warning">
+	<div class="container">
+		<h3>Cookies</h3>
+		<p>This site stores certain information as 'cookies' on your device in order to improve your website experience. We also share information about your use of the site with analytics partners who may combine it with other information that you've provided to them or that they've collected from your use of their services. </p>
+		<a href="#" class="cookieaccept">Yes, that's fine</a>
+		<a class="cookiedeny" href="/about/privacy">No, I dont want these cookies</a>
+	</div>
+</div>
+<script async src=https://www.googletagmanager.com/gtag/js?id=G-1EJP6Q9PMT></script>
+<script src="[% version('/cobrands/shropshire/cookies.js') %]"></script> <!-- this must be below the google gtag script -->
```
MorayMySoc, line 10: Cobrand name

templates/web/thamesmead/header_logo.html
```diff
@@ -0,0 +1,2 @@
+<a href="https://shropshire.gov.uk" id="site-logo">Peabody</a>
```
MorayMySoc, line 1: Cobrand name

templates/web/thamesmead/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
+<div id="js-roads-responsibility" class="box-warning hidden">
+    <strong>Not maintained by Shropshire Council</strong>
```
MorayMySoc, line 2: Cobrand name

templates/web/thamesmead/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
...
+    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-shropshire">
+        <p>The location you have selected doesn't appear to be on <span class="js-roads-asset" data-original="a road">a public road</span> that Shropshire Council maintain.</p>
```
MorayMySoc, line 4: Cobrand name (so Thamesmead won't be Thamesmead council - it will just be Thamesmead)

templates/web/thamesmead/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
...
+        <p>Please select <span class="js-roads-asset" data-original="a road">a road</span> on which to make a report.</p>
+    </div>
+    <div id="js-not-council-road" class="js-responsibility-message hidden js-roads-shropshire">
```
MorayMySoc, line 7: Cobrand name

templates/web/thamesmead/report/new/roads_message.html
```diff
@@ -0,0 +1,10 @@
...
+        <p>The road you have selected is a <strong>Private road</strong> and not maintained by Shropshire Council</p>
```
MorayMySoc, line 8: Cobrand name

web/cobrands/thamesmead/base.scss
```diff
@@ -0,0 +1,347 @@
...
+        flex-wrap: wrap;
+        margin: 0 -0.5em;
+    }
+
+    li {
+        margin: 0 0.5em;
+    }
+}
+
+// As per Shropshire Website style guide
```
MorayMySoc, line 213: Cobrand name

MorayMySoc (changes requested):
Sorry - this is an overhead of my suggesting you use the other cobrand to start, so a few instances where we need to check the names.

The footer issue can be resolved by the way oxfordshire have done it, I think, by using their own footer.html

lucascumsille:
Thanks @MorayMySoc I always the name was a variable that would change with the cobrand name. Noted for the next cobrand I need to style =)

templates/web/thamesmead/header_logo.html
lucascumsille: I have update the link, however instead of using Thamesmead, I kept Peabody, because the link is going to take you to the Peabody website. But let me know if you think it should be Thamesmead.

PR #3876, opened 2022-04-08
Add CORS header for requests from iOS app

davea:
[Recent changes](https://github.com/mysociety/fixmystreet-mobile/pull/298) to the iOS app mean that requests to FMS now include an `Origin: app://localhost` header, so we need to return the corresponding header in the response.

[skip changelog]


struan (approved):
Looks good.

PR #3879, opened 2022-04-12
Use staging tilma for staging FMS sites

davea:
Staging tilma is configured to use V3 of the Street Manager API, this will let us
test it as required.

Leaving as draft because we probably don't need this long term.

For https://github.com/mysociety/societyworks/issues/2892

[skip changelog]


davea:
Closing as live Tilma is now using the V3 API.

PR #3880, opened 2022-04-12
[BathNES] Update WFS to new servers

MorayMySoc:
Request to update assets to new WFS server.

The data appears to be the same, though not sure what LLPG is.

https://mysocietysupport.freshdesk.com/a/tickets/1987

[skip changelog]


struan (approved):
👍

PR #3881, opened 2022-04-13
Improve adding aria-label to multi-select-button.

dracos:
#3863 had a couple of bugs, I'm afraid - 'untriaged' disappeared from the triage filter, it was acting on multi-selects on other pages besides report lists (e.g. admin user form has one that was getting an incorrect aria-label), there was still English in the JS ("Show" and "about") which wouldn't be translated and display in English in other languages. Hopefully what I've done makes sense, let me know if you have any questions.

### DIFF
--- a/templates/web/base/admin/triage/_list-filters.html
+++ b/templates/web/base/admin/triage/_list-filters.html
@@ -21,7 +21,7 @@
 [% END %]
 
         <p class="report-list-filters">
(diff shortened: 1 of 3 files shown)


MorayMySoc (approved):
Seems to work - and from your notes without breaking things

lucascumsille:
Thanks Matthew, sorry I forgot to add translation for the new span elements. This PR works for me, no orphaned labels and the aria-labels are displaying the info we need.

lucascumsille (approved):

PR #3882, opened 2022-04-14
[Buckinghamshire] Fixed font inconsistency in claim form

lucascumsille:
Added minor UI fix that prevents inconsistency in the claims forms.
The current fix uses "Helvetica" just like the rest of the website.

Screenshot 2022-04-14 at 09 25 13

Fixes: mysociety/societyworks#2930


dracos (approved):
Woo

PR #3885, opened 2022-04-19
Added fix that enables buttons from having a full clickable area.

lucascumsille:
This will prevent the ".fancybox-bg-s" element from taking over the clickable area from the button.

This fixes: mysociety/societyworks#2935

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -1647,6 +1647,10 @@ input.final-submit {
     left: 0px;
 }
 
+.fancybox-bg {
+  pointer-events: none;
+}
+
 .problem-back {
   display: block;
   font-size: 1em;


web/cobrands/sass/_base.scss
```diff
@@ -1647,6 +1647,10 @@ input.final-submit {
     left: 0px;
 }
 
+#fancybox-bg-s {
```
dracos, line 1650: Might as well be `.fancybox-bg` to treat all the shadows the same?

dracos (changes requested):
One suggestion on affecting all shadows, not just the bottom one.

web/cobrands/sass/_base.scss
lucascumsille: Thanks Matthew, I just pushed a commit that will apply the the fix in all the ".fancybox-bg" elements instead of just the bottom one.

dracos (approved):

PR #3888, opened 2022-04-20
[Thamesmead] Add Thamesmead cobrand file

MorayMySoc:
Add cobrand file for Peabody/Thamesmead cobrand

https://github.com/mysociety/societyworks/issues/2923

[skip changelog]


davea (changes requested):
Codewise, looks good. Can the base for this PR be a new `thamesmead-reviewed` branch instead of `master` though, please? Also the commit should probably be prefixed `[Thamesmead]` instead of `[Peabody]` to keep things consistent with the code.

PR #3890, opened 2022-04-20
[Kingston] Initial waste work

dracos:
Think you already looked at the first refactor bits, few other small bits there (each own commit), then initial cobrand has your changes in and any changes made (mainly switch to using service task IDs, and wrong bin images), initial templates/styles has Lucas' work in. Then new bin admin fee is main new change. Anyway, would be good to have a check over so far.


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1264,7 +1276,10 @@ sub process_garden_data : Private {
     $data->{bin_count} = $bin_count;
     $data->{new_bins} = $data->{bins_wanted} - $data->{current_bins};
 
-    my $total = $c->cobrand->garden_waste_cost($bin_count);
+    my $cost_pa = $c->cobrand->garden_waste_cost_pa($bin_count);
+    my $cost_now_admin = $c->cobrand->garden_waste_new_bin_admin_fee($data->{new_bins});
+    my $total = $cost_now_admin + $cost_pa;
+
     $c->set_param('payment', $total);
```
struan, line 1283: I think we need to store the total, the admin and the annual fee as the latter two need to be broken out in the Kingston CC parameters.

perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
```diff
@@ -65,12 +72,18 @@ has_page summary => (
...
-        my $new_bins = $data->{bins_wanted} - $data->{current_bins};
+        my $current_bins = $data->{current_bins};
+        my $bin_count = $data->{bins_wanted};
+        my $new_bins = $bin_count - $current_bins;
         my $pro_rata = $c->cobrand->waste_get_pro_rata_cost( $new_bins, $c->stash->{garden_form_data}->{end_date});
-        my $total = $c->cobrand->garden_waste_cost($data->{bins_wanted});
+        my $cost_pa = $c->cobrand->garden_waste_cost_pa($bin_count);
+        my $cost_now_admin = $c->cobrand->garden_waste_new_bin_admin_fee($new_bins);
+        my $total = $cost_pa;
+        $pro_rata += $cost_now_admin;
```
struan, line 82: There's quite a lot of doing the bin cost assignment. I wonder if it might be worth extracting out into a method.

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -0,0 +1,608 @@
...
+            my $garden = 0;
+            my $garden_bins;
+            my $garden_cost = 0;
+            my $garden_due = $self->waste_sub_due($schedules->{end_date});
+            my $garden_overdue = $expired{$_->{Id}};
+            if ($service_name eq 'Garden Waste') {
+                $garden = 1;
+                my $data = Integrations::Echo::force_arrayref($task->{Data}, 'ExtensibleDatum');
+                foreach (@$data) {
+                    next unless $_->{DatatypeName} eq 'LBB - GW Container'; # DatatypeId 5093
```
struan, line 299: I guess we don't know what this should be yet?

struan (approved):
Still feels like there's a lot of common code between the two cobrands (e.g `within_working_days`, `_parse_events` plus a lot of `bin_services_for_address`) but I'm not sure how common it's likely to remain so *shrug*.

Otherwise few minor comments but looks good.

perllib/FixMyStreet/Cobrand/Kingston.pm
dracos: I think it's "SLWP - Containers" (it appears to be that in live but can't find same in staging).

perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
dracos: Yeah, I wondered, but e.g. here it's the pro-rata cost which may (in a case where both pro-rata and admin fee are used) be different from annual.

perllib/FixMyStreet/App/Controller/Waste.pm
dracos: Yep, have done that now in 9d5ff70bdb4e7f410738fe0e7df9e50f892c69da

PR #3893, opened 2022-04-25
[Oxfordshire] Add script for one-off updating of district categories

davea:
For https://github.com/mysociety/societyworks/issues/2877

[skip changelog]

(There's an example data file at `~davea/district_categories.json` on the staging server.)

### DIFF
--- a/bin/oxfordshire/update_district_category_notices
+++ b/bin/oxfordshire/update_district_category_notices
@@ -1,0 +1,135 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+update_district_category_notices
+
+=head1 USAGE
+
+update_district_category_notices path/to/data.json
+
+=head1 DESCRIPTION
+
+This is a one-off script to make some bulk changes to the contacts for the
+district councils in Oxfordshire.
+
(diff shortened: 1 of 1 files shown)


bin/oxfordshire/update_district_category_notices
```diff
@@ -0,0 +1,135 @@
...
+            $note .= "Updated to correct email address. ";
+        }
+
+        if ($cfg->{url}) {
+            my $meta = {
+                code => '_fms_disable_',
+                variable => 'false',
+                protected => 'true',
+                disable_form => 'true',
+                description => "Please report this issue to " . $body->name . " using <a href='" . $cfg->{url} . "'>their website</a>.",
```
dracos, line 114: I think this wording needs to be more active for accessibility - could it all be a link after Please?

dracos (changes requested):
Script looks good, very clear; one suggestion on the link wording is all.

bin/oxfordshire/update_district_category_notices
davea: Good thinking, thanks.

PR #3894, opened 2022-04-25
[Thamesmead] Adds link to Thamesmead's own privacy policy

MorayMySoc:
* Adds cobrand method to return link to Thamesmead's own privacy policy
https://github.com/mysociety/societyworks/issues/2928

[skip changelog]


davea (approved):
Looks fine in isolation, but be aware #3875 has hard-coded `/about/privacy` URLs which will need updating to use this.

PR #3896, opened 2022-04-27
[Thamesmead] Add admin permissions

MorayMySoc:
Adds admin permission based on UK Councils

Adds body based on Highways England

Adds users_restrictions imported from UK Councils

Please check the following:

[skip changelog]


davea (approved):
Made a couple of small commits so I could deploy to staging (which I've now done), looks good 👍

PR #3897, opened 2022-04-28
[Kingston] Credit card flow changes

dracos:
Mostly the changes needed for working CC flow, with couple other small changes. [skip changelog]
Needs a fix for missing payment_method when staff-CNP flow is used (think just needs a default, don't think that has any other effect).
For https://github.com/mysociety/societyworks/issues/3006 (bar finding out if name needed) and https://github.com/mysociety/societyworks/issues/3001

### DIFF
--- a/perllib/Integrations/SCP.pm
+++ b/perllib/Integrations/SCP.pm
@@ -78,6 +78,35 @@ sub pay {
     my ($self, $args) = @_;
 
     my $credentials = $self->credentials($args);
+
+    my @items;
+    my $total = 0;
+    foreach (@{$args->{items}}) {
+        push @items, ixhash(
+            'scpbase:itemSummary' => ixhash(
+                'scpbase:description' => $_->{description},
+                'scpbase:amountInMinorUnits' => $_->{amount},
+                'scpbase:reference' => $_->{reference},
+            ),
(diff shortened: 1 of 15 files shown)


perllib/Integrations/SCP.pm
```diff
@@ -119,13 +119,15 @@ sub pay {
...
-                        },
+                        $args->{vat} ? (
+                            'scpbase:tax' => {
+                                'scpbase:vat' =>ixhash(
+                                    'scpbase:vatCode' => $self->config->{scp_vat_code},
+                                    'scpbase:vatRate' => $self->config->{scp_vat_rate} || 0,
+                                    'scpbase:vatAmountInMinorUnits' => $args->{vat} || 0,
+                                ),
+                            },
+                        ) : (),
```
struan: Might be worth checking if this should be included even if it's zero rated in case they need it for accounting purposes. Mostly I say this as they did include a vat code in the tech spec.

struan (approved):
one small question but looks good otherwise.

perllib/Integrations/SCP.pm
dracos: Ah, that's the Bromley figures, they haven't updated that part at all. But yes, will check.

PR #3898, opened 2022-04-28
[Waste] Do not reuse CSC payment method on modify.

dracos:
If the original payment was made by CSC, this payment method was being copied across to modifications made by normal users. It is already re-set to CSC if CSC is used for the modification so default to credit card in the usual case. Also fix button text for such a payment. Fixes https://github.com/mysociety/societyworks/issues/2865 [skip changelog]


struan (approved):
👍

Issue #3903, opened 2022-05-06
Make report unassignment work fully


nephila-nacrea:
On an 'all reports' page (e.g. https://staging.fixmystreet.com/reports/Lincolnshire) and a single report page (e.g. https://staging.fixmystreet.com/report/14521), there are 'assign to' dropdowns (available to certain admin users).

Admin can use these to assign reports to members of staff; they should also be able to use the 'unassigned' option to unassign reports. However, unassignment only works if the member of staff who is assigned to a report is the same member of staff who raised the report to begin with.

Issue #3646, opened 2021-11-08
Hide "assigned to" bits when there's noone to assign reports to [4hrs]


chrismytton:
Since the changes to the inspector tool in #3535 we're now showing "Assigned to:" in a couple of places, even when it's not relevant.

On the admin report_edit page, underneath the category dropdown:

<img width="216" alt="Screenshot 2021-11-08 at 15 58 57" src="https://user-images.githubusercontent.com/22996/140775579-f34284b7-3646-4147-83d5-8856c168228b.png">

On the report page's inspector sidebar:

![image](https://user-images.githubusercontent.com/22996/140775621-5921e7e3-05e4-4cf6-a042-f455d14c5460.png)

I think these should only be shown if there are users with the relevant permission.

Related Freshdesk ticket: https://mysocietysupport.freshdesk.com/a/tickets/1579

dracos:
Also on list pages, anywhere in the country, I'm seeing it on every report in the sidebar, it's a bit annoying.

jacquelinebylau:
Toggl: https://plan.toggl.com/#timeline/317959/tasks/33843711/embed

PR #3904, opened 2022-05-06
Tidy up behaviour around unassigned reports

nephila-nacrea:
This PR hides 'assigned to' text if there is no assignee for a report.
It also fixes buggy behaviour around unassigning reports.

Fixes https://github.com/mysociety/fixmystreet/issues/3646.
Fixes https://github.com/mysociety/fixmystreet/issues/3903.

Please check the following:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
         - Make calls from Geocoder files to https rather than http
         - Inspector dropdown list only shows name once even if permissions repeated #3870
         - Inspector dropdown list doesn't show anonymised users, removing blank options #3873
+        - Fix report unassignment so it works for users who did not create the report #3903
     - Accessibility improvements:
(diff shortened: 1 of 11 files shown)


dracos (commented):
What's here looks great, does what it says nicely :)
I don't think this fully fixes 3646 though - as it mentions there, the dropdowns are appearing even if there are no users who can be assigned reports. If `body.staff_with_permission('report_inspect')` is empty then it shouldn't output the dropdown at all as superfluous.

dracos (changes requested):
Looks fine, but if you look at the dev toolbar you'll see the query is now being run twice, once for the IF and then again in the loop. Ideally we'd only want to look up this data once - options are either change the templates to look up the data into a variable and then pass that to the subtemplate to use, or change the Body.pm code to cache the result on first lookup.

PR #3905, opened 2022-05-06
[Bristol] Fix broken stylemap options

MorayMySoc:
Fix issue breaking tests with stylemap

### DIFF
--- a/web/cobrands/bristol/assets.js
+++ b/web/cobrands/bristol/assets.js
@@ -19,6 +19,10 @@ var park_style = new OpenLayers.Style({
     fillOpacity: "0.25"
 });
 
+var park_style_map = new OpenLayers.StyleMap({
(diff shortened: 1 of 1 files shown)


web/cobrands/bristol/assets.js
```diff
@@ -19,6 +19,12 @@ var park_style = new OpenLayers.Style({
     fillOpacity: "0.25"
 });
 
+var park_style_map = new OpenLayers.StyleMap({
+    default: park_style,
+    select: park_style,
```
dracos, line 24: You don't need to set select or hover on a layer that can't be selected or hovered :)

dracos (changes requested):
One simplification, otherwise fine assuming the tests now pass :)

dracos (approved):

PR #3909, opened 2022-05-09
[Kingston] waste sacks

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3016 - first commit to enable a way to do this, not sure if there's a nicer way I haven't thought of!

Then lookup for https://github.com/mysociety/societyworks/issues/3018 and then code to hopefully fix https://github.com/mysociety/societyworks/issues/3019 and https://github.com/mysociety/societyworks/issues/3022 and https://github.com/mysociety/societyworks/issues/3020 

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
@@ -4,25 +4,22 @@ use utf8;
 use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Waste';
 
-has_field is_staff => ( type => 'Hidden' );
-
 has_page intro => (
     title => 'Modify your green garden waste subscription',
     template => 'waste/garden/modify_pick.html',
     fields => ['task', 'continue'],
-    update_field_list => sub {
-        my $form = shift;
-        return {
-            is_staff => { default => $form->{c}->stash->{is_staff} }
-        };
-    },
-    next => sub { return $_[0]->{is_staff} ? 'alter_staff' : 'alter'; },
+    next => 'alter',
 );
 
-my %alter_fields = (
+has_page alter => (
     title => 'Modify your green garden waste subscription',
     template => 'waste/garden/modify.html',
-    fields => ['current_bins', 'bins_wanted', 'name', 'continue_review'],
+    fields => ['current_bins', 'bins_wanted', 'name', 'phone', 'email', 'continue_review'],
+    field_ignore_list => sub {
+        my $page = shift;
+        return ['phone', 'email'] unless $page->form->c->stash->{is_staff};
+        return [];
+    },
     update_field_list => sub {
         my $form = shift;
         my $c = $form->c;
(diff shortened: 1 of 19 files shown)


perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -375,6 +390,27 @@ sub bin_services_for_address {
...
 }
 
+# Refuse currently in separate Echo - when not, could do this in main loop
+sub property_has_refuse_sacks {
+    my ($self, $id) = @_;
+    my $cfg = $self->feature('echo_rbk');
+    my $echo = Integrations::Echo->new(%$cfg);
+    my $calls = $echo->call_api($self->{c}, $self->moniker,
+        "look_up_property_rbk:$id",
+        GetServiceUnitsForObject => [ $id ],
```
struan, line 400: Are we sure properties share the same ID across both systems?

perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
```diff
@@ -4,25 +4,27 @@ use utf8;
...
-        };
-    },
-    next => sub { return $_[0]->{is_staff} ? 'alter_staff' : 'alter'; },
+    next => 'alter',
 );
 
-my %alter_fields = (
+has_page alter_all => ( # Dummy so that fields are hidden if not staff below
+    fields => ['current_bins', 'bins_wanted', 'name', 'phone', 'email', 'continue_review'],
+);
```
struan, line 16: Just to check, this is because the form engine shows fields that aren't on a specific page on all pages so we need to have a page that shows optional fields so they aren't on all pages?

struan (approved):
few small comments but mostly queries rather than problems.

perllib/FixMyStreet/Roles/CobrandPay360.pm
```diff
@@ -296,7 +296,7 @@ sub waste_get_current_garden_sub {
 
     my $garden;
     for my $service ( @$services ) {
-        if ( $service->{ServiceId} == $self->garden_waste_service_id ) {
+        if ( $service->{ServiceId} == $self->garden_bin_service_id ) {
```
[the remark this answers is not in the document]
dracos: Not using Pay360 for Kingston, so seemed not as important - it also won't work because it's assuming service ID rather than task ID, for example, wasn't sure how to deal. Should note it though, unless being done differently in your work anyway (which I guess will have to account for it somehow).

perllib/FixMyStreet/App/Form/Waste/Garden/Modify.pm
dracos: Yes, exactly (our code in Wizard.pm marks all fields in pages that are not in `current_page->all_fields` as inactive, to be precise). I couldn't think of an easier solution. I guess it could be an attribute on the page itself and then alter that Wizard.pm code in build_active to read that to make sure they're marked inactive too, that might be bit cleaner?

dracos: Got rid of these silly extra pages and put fields back (with full list) and added a runtime `field_ignore_list` which returns list of fields to not include this time. Simple.pm slightly more complicated code, but all other boilerplate simpler: 7b235e0039187d53b80dadc70c6e3e846c61db95

PR #3911, opened 2022-05-09
[Thamesmead] Changes after testing part 1

MorayMySoc:
1) Confirmation of report refers to it going to a council rather than Peabody
https://3.basecamp.com/4020879/buckets/26382066/todos/4906247418
https://3.basecamp.com/4020879/buckets/26382066/todos/4910559283

2) Content text for page when entering your details, refers to 'Council' and should refer to Peabody
https://3.basecamp.com/4020879/buckets/26382066/todos/4915678385

3) Update home page text to reflect Thamesmead locations
https://3.basecamp.com/4020879/buckets/26382066/todos/4911842219

4) Site going through to FMS.com after confirming a report rather than back to Peabody cobrand
https://3.basecamp.com/4020879/buckets/26382066/todos/4902033447

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Thamesmead.pm
+++ b/perllib/FixMyStreet/Cobrand/Thamesmead.pm
@@ -68,6 +68,15 @@ sub base_url { FixMyStreet::Cobrand::UKCouncils::base_url($_[0]) }
 
 sub default_map_zoom { 6 }
 
+sub enter_postcode_text {
+    my ( $self ) = @_;
+    return _("Enter a road name or postcode adjacent to the area you want to report on or the name of the area");
+}
+
+sub example_places {
+    return [ 'Glendale Way', 'Manorway Green' ];
+}
+
 sub munge_report_new_bodies {
     my ($self, $bodies) = @_;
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -68,6 +68,15 @@ sub base_url { FixMyStreet::Cobrand::UKCouncils::base_url($_[0]) }
 
 sub default_map_zoom { 6 }
 
+sub enter_postcode_text {
+    my ( $self ) = @_;
+    return _("Enter a road name or postcode adjacent to the area you want to report on or the name of the area");
+}
+
+sub example_places {
```
dracos, line 76: You could call the UKCouncils function here and use a cobrand feature for this if that was preferred.

perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -77,6 +77,10 @@ sub example_places {
     return [ 'Glendale Way', 'Manorway Green' ];
 }
 
+sub get_body_handler_for_problem {
```
dracos, line 80: Please log a ticket about how things will need to be changed to have this work if reports can be made on .com so we don't forget.
I think (though not totally sure) the only issue here is if someone looks at a Peabody report on .com, it'll use either Greenwich/Bexley's cobrand for `is_council_with_case_management` (neither of which use that). You could add an exception similar to TfL's to the UK.pm get_body_handler_for_problem, which I guess would fix that for now.

dracos (changes requested):
Couple of things

perllib/FixMyStreet/Cobrand/Thamesmead.pm
MorayMySoc: The UK Councils text is slightly different as cobrand requested text (wasn't sure whether I should use translation trigger for a cobrand though). It seemed easiest to do the example place in the same place while dealing with the query, but happy to move to feature if preferred

dracos: Yes, I was only talking about example_places - doesn't really matter, I guess, might as well leave it.

PR #3912, opened 2022-05-09
[Kingston] Allow editing of renewal bin count.

dracos:
[skip changelog] Fixes https://github.com/mysociety/societyworks/issues/3008


struan (approved):
🔢

PR #3915, opened 2022-05-10
[Thamesmead] Change geocoder to OSM and use OSM mapping methods

MorayMySoc:
1) Southmere drive not in Crayford - OSM doesn't show it as such so changing to OSM removes Crayford
https://3.basecamp.com/4020879/buckets/26382066/todos/4906047321

2) Claridge Way not coming up in search - OSM shows Claridge Way
https://3.basecamp.com/4020879/buckets/26382066/todos/4906653404

3) Holstein and Lensbury way don't show on map and have Crayford in title - OSM removes Crayford. Lensbury way seems to show up but Holstein too small for the map? 
https://3.basecamp.com/4020879/buckets/26382066/todos/4906718571

4) Some locations not found and taken to an area outside of Thamesmead - remove not area search results.
https://3.basecamp.com/4020879/buckets/26382066/todos/4898005285

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Thamesmead.pm
+++ b/perllib/FixMyStreet/Cobrand/Thamesmead.pm
@@ -91,4 +91,31 @@ sub privacy_policy_url {
     'https://www.thamesmeadnow.org.uk/terms-and-conditions/privacy-statement/'
 }
 
+sub get_geocoder { 'OSM' }
+
+sub disambiguate_location {
+    my $self    = shift;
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
geocoder_munge_results passes in one result, not a list (so code is fine, but confusing) - a test of the munging function would be nice.

dracos (approved):
Thanks, looks good

PR #3918, opened 2022-05-11
[Oxfordshire] Add Parish and Districts to All Reports

MorayMySoc:
* Adds Parish and District tabs on all reports for Oxfordshire
* Populates these with relevant Parish or Ward data
* Adds method to body to search for specific children data
* Adds a cobrand feature to allow initialising option

https://github.com/mysociety/societyworks/issues/2658

[skip changelog]


templates/web/base/reports/_other_districts-list.html
```diff
@@ -0,0 +1,58 @@
...
+          [% END %]
+        [% END %]
+      </ul>
+      </section>
+      <p>
+        <input class="btn [% show_district_shown_class %]" type="submit" value="[% loc('View districts') %]" />
+      </p>
+      </form>
+  </section>
+[% END %]
```
chrismytton, line 58: This is a base template but the javascript is only for Oxfordshire. Perhaps this should be in an Oxfordshire-specific template for now (with appropriate try/catch where it's included). Then if/when we want to reuse this on other cobrands in the future we can see which bits can be shared.

chrismytton (changes requested):
Looking good so far!

I've left a couple of comments about rearranging the templates slightly, hopefully makes sense.

I also noticed a JavaScript bug where the menu disappears if you click back to the "View area reports" tab after viewing parish or district reports. Video below shows this:

https://user-images.githubusercontent.com/22996/168588907-e8638627-881e-4018-bd61-fafd7bc60ddb.mov

MorayMySoc:
@chrismytton I've been round the houses on this a bit as the mechanic I used to create this originally got broken and it was recommended that I don't use it even when it was fixed (calling 'covers' from mapit). So I've redone it pretty much. I think the original issue with the tabs not working is that you need to run ./script/setup for the scss to be compiled. I think the tab names need to be slightly different - is it District, Parish and District Ward?

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -180,6 +185,16 @@ sub ward : Path : Args(2) {
             }
             $c->stash->{children} = $children;
         }
+        my $covered = {};
+        $c->cobrand->call_hook('add_parish_wards' => $covered);
+            unless ($covered->{error}) {
+                foreach (values %$covered) {
+                    $_->{url} = $c->uri_for( $c->stash->{body_url}
```
chrismytton, line 192: These lines have four spaces more indentation than needed. Threw me off a bit when I first read it!

t/Mock/MapIt.pm
```diff
@@ -200,6 +200,16 @@ sub dispatch_request {
...
+    sub (GET + /area/*/covers) {
+        my ($self, $area, $type) = @_;
+        if ($area eq '2237') {
+            return $self->output({
+                61327 => {parent_area => 2237, id => 61327, name => "Aston Upthorpe", type => "CPC"},
+                152680 => {parent_area => 2237, id => 152680, name => "Faringdon", type => "DIW"},
+            });
+        }
+    },
+
```
chrismytton, line 212: Is this still needed now we're getting the areas from a JSON file?

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -65,6 +65,11 @@ sub index : Path : Args(0) {
         unless ($children->{error}) {
             $c->stash->{children} = $children;
         }
+        my $covered = {};
+        $c->cobrand->call_hook('add_parish_wards' => $covered);
+        unless ($covered->{error}) {
+            $c->stash->{covered} = $covered;
+        }
```
chrismytton, line 72: I think calling this variable and the stash item (and by extension the template variable) "covered" is a bit confusing. It made sense for Oxfordshire, because the data was coming from a Mapit covers call, but other councils might get the data in a different way.

Perhaps something like `parish_wards` or similar would be a bit clearer?

chrismytton (changes requested):
Running `script/update` fixed the tab issue, thanks!

> I think the tab names need to be slightly different - is it District, Parish and District Ward?

Hmm, not 100% sure. Perhaps something to check with Oxfordshire about when you tell them it's on staging?

Think this is nearly there! I've left a few comments inline, but hopefully nothing major!

MorayMySoc:
I removed some variable from `templates/web/base/reports/body.html` which I think was superfluous too, but may be a Friday foolishness.

templates/web/base/reports/_other_districts-list.html
```diff
@@ -0,0 +1,58 @@
+[% IF covered.size %]
```
chrismytton, line 1: I think this should be `parish_ward.size`? There are a couple of other mentions of `covered` in this file still.

chrismytton (changes requested):
Just one minor issue with the base template having the old variable name in, then I think this is done!

templates/web/base/reports/_other_districts-list.html
MorayMySoc: Realised that I'd broken the main bodies links, so fix for that as well.

davea (approved):
👍

MorayMySoc:
Been feedback on the ticket and I've made some changes regarding that.

davea (approved):

PR #3922, opened 2022-05-12
[Kingston] Admin things, fixes for retry, changes for login

dracos:
Sorry, bit of a bundle of stuff, I can split it up if you want; but three main areas - some payment retry fixes, changes to login/confirmation text/password, and admin improvements (removing unneeded bits and a waste specific admin page). [skip changelog]


struan (approved):
yay for clearer admin page.

PR #3925, opened 2022-05-12
[Waste] Renewal improvements

dracos:
Check all reports, not just the user's, for renewal/bin day page for better showing of renewal button/no need to renew error; fix (presumably?) not doing email/phone validation on garden/renew; and share the Form controller code to get the unique ID double-submit fix from claims (and in general, good to be using the same code!).

Fixes https://github.com/mysociety/societyworks/issues/3038
Connects https://github.com/mysociety/societyworks/issues/2694

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -553,6 +553,10 @@ sub property : Chained('/') : PathPart('waste') : CaptureArgs(1) {
 sub bin_days : Chained('property') : PathPart('') : Args(0) {
     my ($self, $c) = @_;
 
+    # To try and work out whether to show a renewal path or not
+    $c->forward('get_original_sub', ['any']);
+    $c->stash->{current_payment_method} = $c->forward('get_current_payment_method');
+
     my $staff = $c->user_exists && ($c->user->is_superuser || $c->user->from_body);
 
     my $cfg = $c->cobrand->feature('waste_features');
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Form/Waste/Garden/Renew.pm
```diff
@@ -29,12 +29,17 @@ has_page intro => (
 
         my $max_bins = $data->{max_bins};
         my %bin_params = ( default => $data->{bins}, range_end => $max_bins );
+
+        my ($name, $email, $phone) = ('', '', '');
+        if ($c->user_exists) {
+            ($name, $email, $phone) = ($c->user->name, $c->user->email, $c->user->phone);
+        }
```
struan, line 36: This defaults to the staff account details. Is that what we want?

struan (commented):
This all looks good but one question about the name/email fields for staff users.

perllib/FixMyStreet/App/Form/Waste/Garden/Renew.pm
dracos: Ah, probably not, should do whatever the subscribe form does the same I guess.

dracos: So rare, should just make them type it in I think probably, like subscribe.

struan: Yes, that makes sense.

PR #3927, opened 2022-05-13
Permit control over database connection SSL mode

sagepe:
This adds a new configuration variable, `FMS_DB_SSLMODE`, can be used to override the default `sslmode` setting FMS uses when connecting to its PostgreSQL database and changes the default from `allow` to `prefer`.

The previous value for `sslmode` used was hardcoded to `allow` which will at first try an unencrypted connection before trying an encrypted one if the database server enforces SSL. This is an insecure default behaviour and also results in large numbers of FATAL connection errors on database servers that enforce SSL connections.

This nows default to `prefer` if the variable isn't present in the configuration file. This is backwards-compatible with the previous default in that it will allow unsecure connections, but only after first trying an encrypted connection, reversing the previous behaviour.

We recommend setting this to at least `require` and ensuring that your database server is configured to only accept encrypted connections.

Please check the following:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
         - Add an index on problem(external_id) to speed up bin/fetch --updates
         - Upgrade Net::DNS and libwww to deal with IPv6 issues.
         - Add send_state column to updates. #3865
+    - Security
+        - Permit control over database connection `sslmode` via $FMS_DB_SSLMODE
     - Open311 improvements:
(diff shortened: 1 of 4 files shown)


sagepe:
I wasn't sure whether or how to add any relevant tests, so if this is something worth doing some guidance on that would be gratefully received.

chrismytton (approved):
Looks good! 🔐

PR #3928, opened 2022-05-13
[Thamesmead] Further editorial changes

MorayMySoc:
Make text changes based on feedback from client

https://3.basecamp.com/4020879/buckets/26382066/todos/4916971998

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Thamesmead.pm
+++ b/perllib/FixMyStreet/Cobrand/Thamesmead.pm
@@ -66,11 +66,16 @@ sub updates_restriction { FixMyStreet::Cobrand::UKCouncils::updates_restriction(
 
 sub base_url { FixMyStreet::Cobrand::UKCouncils::base_url($_[0]) }
 
+sub contact_email {
+    my $self = shift;
+    return $self->feature('contact_email');
+};
+
 sub default_map_zoom { 6 }
 
 sub enter_postcode_text {
     my ( $self ) = @_;
-    return _("Enter a road name or postcode adjacent to the area you want to report on or the name of the area");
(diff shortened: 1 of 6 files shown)


templates/web/base/contact/blurb.html
```diff
@@ -3,7 +3,10 @@
 </p>
 
 <p>
-[% loc('Please do <strong>not</strong> report problems through this form; messages go to
-the team behind this site, not a council. To report a problem,
-please <a href="/">go to the front page</a> and follow the instructions.') %]
+  [% TRY %]
+    [% INCLUDE 'contact/_blurb.html' %]
```
chrismytton, line 7: I wonder if it might be clearer to simply override the whole `contact/blurb.html` template for Peabody, rather than introducing another partial?

perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -66,11 +66,16 @@ sub updates_restriction { FixMyStreet::Cobrand::UKCouncils::updates_restriction(
 
 sub base_url { FixMyStreet::Cobrand::UKCouncils::base_url($_[0]) }
 
+sub contact_email {
+    my $self = shift;
+    return $self->feature('contact_email');
```
chrismytton, line 71: Could this use `FixMyStreet::Cobrand::UKCouncils:: contact_email`, similar to the methods above?

chrismytton (approved):
Looks good! I've left a couple of comments, but neither of them are blocking.

perllib/FixMyStreet/Cobrand/Thamesmead.pm
MorayMySoc: $self->next::method(); Not working in a quirk I don't understand, so leaving this for now.

PR #3933, opened 2022-05-17
[UK] Changes to support latest MapIt generation

davea:
- Some code changes to use the right area IDs and return the right bodies in various places
 - Script to migrate existing body areas in DB to the correct values

[skip changelog]


perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -170,10 +170,15 @@ sub area_check {
     return 1 if FixMyStreet->staging_flag('skip_checks');
 
     my $councils = $params->{all_areas};
-    my $council_match = defined $councils->{$self->council_area_id};
-    if ($council_match) {
-        return 1;
+
+    # The majority of cobrands only cover a single area, but e.g. Northamptonshire
```
dracos, line 174: This change presumably means the TfL specific version of this function can be dropped.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -243,7 +248,12 @@ sub owns_problem {
     }
     # Want to ignore the TfL body that covers London councils, and HE that is all England
     my %areas = map { %{$_->areas} } grep { $_->name !~ /TfL|National Highways/ } @bodies;
-    return $areas{$self->council_area_id} ? 1 : undef;
+
+    my $council_area_ids = $self->council_area_id;
+    $council_area_ids = [ $council_area_ids ] unless ref $council_area_ids eq 'ARRAY';
+    foreach (@$council_area_ids) {
```
dracos, line 254: Does admin_allow_user further down in this file need updating as well?

perllib/FixMyStreet/Cobrand.pm
```diff
@@ -159,7 +159,14 @@ sub body_handler {
     foreach my $avail ( $class->available_cobrand_classes ) {
         my $cobrand = $class->get_class_for_moniker($avail->{moniker})->new({});
         next unless $cobrand->can('council_area_id');
-        return $cobrand if $areas->{$cobrand->council_area_id};
+
+        # The majority of cobrands only cover a single area, but e.g. Northamptonshire
```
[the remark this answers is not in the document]
dracos: All of this would I think be fixed/simplified by the suggestion we discussed of assigning a cobrand to a body, I think. As it is, if we want to do this step first, I think we'll need some new tests given none of this is caught by anything; I think the fix for now will have to be an exemption in Body's get_cobrand_handler similar to get_body_handler_for_problem, maybe?

dracos (changes requested):
Right sort of changes, certainly, but I'm afraid it looks like it will mean some functions start returning the wrong results. More northamptonshire.t tests with the district body and a Police body would hopefully show up the issues there, but the changes will also have caused problems with the TfL/London council overlaps (which in some way have relied upon the TfL array of council areas not working in FixMyStreet::Cobrand->body_handler).

davea:
Thanks for the review. I've updated the code to implement the body–cobrand link that we discussed. It's caused quite a big diff in the tests, I'm afraid, but I think everything is looking good now.

dracos (changes requested):
This is looking very good, think we're nearly there :)

One issue arising (I think) and a couple of test things - apart from those, I think all I'd like additionally is as mentioned before, a few tests with the Northants set up with police/Highways/W/N to show that your changes now let this work okay; and some form of TfL regression test if possible maybe? But the Northants one mainly.

t/app/model/body.t
```diff
@@ -0,0 +1,11 @@
+use FixMyStreet::TestMech;
+
+my $mech = FixMyStreet::TestMech->new;
+
+my $body = $mech->create_body_ok(163793, 'Buckinghamshire');
+
+# body can have no cobrand
```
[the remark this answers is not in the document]
davea: Oops! Sorted, ta.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -358,33 +358,29 @@ sub report_check_for_errors {
...
 sub get_body_handler_for_problem {
     my ($self, $row) = @_;
 
-    if ($row->to_body_named('TfL')) {
-        return FixMyStreet::Cobrand::TfL->new;
-    }
-    if ($row->to_body_named('Thamesmead')) {
-        return FixMyStreet::Cobrand::Thamesmead->new;
-    }
     # Do not do anything for National Highways here, as we don't want it to
```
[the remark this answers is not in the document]
davea: As far as I can tell this is working correctly. Reports sent to NH are branded with whatever cobrand was used to make them. I will add a test for that behaviour though.

davea:
Thanks - think I've covered everything now... I've added tests for the Northants council/police emails. I believe the TfL behaviour is already covered by [existing tests](https://github.com/mysociety/fixmystreet/blob/mapit-generation-update/t/cobrand/tfl.t#L507-L515), though may have misunderstood!

perllib/FixMyStreet/DB/Result/Body.pm
```diff
@@ -231,6 +237,9 @@ sub cobrand_name {
     # If the current body is TfL then we always want to show TfL as the cobrand name.
     return $self->name if $self->name eq 'TfL' || $self->name eq 'National Highways' || $self->name eq 'Environment Agency';
```
dracos, line 238: Given the new code, is this section needed any more? This confusion shouldn't be able to arise now.

dracos (approved):
One more section of code I think can now be dropped too \o/

That probably tests enough of it. I think the issue at my first review was all the tests (including those TfL ones you mention) passed, but there were issues that meant there should have been tests that failed. So it wasn't just the email branding (which tests the get_body_handler part), but e.g. cobrand_name using get_cobrand_handler would have been giving the wrong results (e.g. the Bexley body could return "TfL", or Northants Police body would return Northants Highways) - this won't be happening now, think that's clear, so it's probably fine.

perllib/FixMyStreet/DB/Result/Body.pm
dracos: @davea Just to note I don't think this was removed.

PR #3913, opened 2022-05-09
[Highway] Added maintenance image to homepage

lucascumsille:
Added maintenance  image to National Highways homepage. I put the image on the right side instead of below the text, so the "Enter a location" is still visible, also to use the empty right space.

Screenshot 2022-05-09 at 15 36 32

Screenshot 2022-05-09 at 15 36 52

Screenshot 2022-05-09 at 15 44 51

Tested on Safari, IE and Chrome.

Please let me know if I misunderstood the ticket.

PR #3935, opened 2022-05-18
[National Highways] Added background image for hero section

lucascumsille:
Hi @zarino 
I didn't realise there was already a mockup on the[ freshdesk](https://mysocietysupport.freshdesk.com/a/tickets/1534) for this [ticket](https://trello.com/c/9sjXSSWD), so I have created a new PR, with a new version of the hero section with a background image.
The only thing extra that I added compared with the mock-up is an overlay to the paragraph below H1, so it's more accessible for users with vision impairment.

The other PR is this one: 
https://github.com/mysociety/fixmystreet/pull/3913

Personally I prefer the version of the newest PR.

Screenshot 2022-05-18 at 11 50 24

Here is the original mockup, uploaded by Clare on the 29th of October
homepage-mockup


zarino:
Yeah, my point to the client was that it’s a shame that the focus of the image (the man’s face) is hidden behind the text – instead, all you see is a sign and an awkward looking hand holding a drill. I wondered whether they had any other shots from the same photoshoot, but the other way round.

(I considered trying to flip the image, but there’s so much text in it, it’d be a nightmare.)

zarino:
As mentioned previously, it seems a shame that the composition of the image means the chap’s face is hidden behind the hero text, and all we see instead is a load of empty space between him and his drill.

I can see why we’ve added the dark box around the hero text, to increase colour contrast with the background image, but it feels like a cludge to me – if contrast is a problem, we just shouldn’t be putting an image back there 🤔 

That said, this _is_ the client’s cobrand, so if they want the homepage as presented here, with those drawbacks, then 🤷 

Perhaps Lizetta / Jackie might have thoughts on what the priority is here?

zarino (commented):
As mentioned previously, it seems a shame that the composition of the image means the chap’s face is hidden behind the hero text, and all we see instead is a load of empty space between him and his drill.

I can see why we’ve added the dark box around the hero text, to increase colour contrast with the background image, but it feels like a cludge to me – if contrast is a problem, we just shouldn’t be putting an image back there 🤔

That said, this is the client’s cobrand, so if they want the homepage as presented here, with those drawbacks, then 🤷

Perhaps Lizetta / Jackie might have thoughts on what the priority is here?

PR #3938, opened 2022-05-19
Kingston waste dd form

struan:
Config and form for creating a direct debit.

[skip changelog]


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -377,10 +377,13 @@ sub populate_dd_details : Private {
     $c->stash->{reference} = $c->cobrand->call_hook( 'waste_dd_payment_ref' => $p ) || 'GGW' . $c->stash->{property}{uprn};
     $c->stash->{lookup} = $reference;
     $c->stash->{payment_date} = $dt;
+    $c->stash->{start_date} = $dt->ymd;
     $c->stash->{day} = $dt->day;
     $c->stash->{month} = $dt->month;
     $c->stash->{month_name} = $dt->month_name;
     $c->stash->{year} = $dt->year;
+
+    $c->stash->{redirect} = $c->cobrand->call_hook( 'garden_waste_dd_redirect_url' => $p ) || '';
```
dracos, line 386: Nothing appears to happen with this so I assume unfinished? Should we take this out of this commit for now, what's the default redirect, does it work without this parameter?

perllib/FixMyStreet/Roles/Bottomline.pm
```diff
@@ -91,4 +91,33 @@ sub waste_payment_type {
...
+
+    return (undef, undef) unless $id;
+    my $origin = FixMyStreet::DB->resultset('Problem')->find($id);
+
+    if ( !$origin ) {
+        $self->log("no matching origin sub for id $id");
+        return (undef, undef);
+    }
+
+    $uprn = $origin->get_extra_field_value('uprn');
```
dracos, line 107: I'm not sure what this section is for - presumably if we've already got the report (from its ID) by here, we don't need to check the UPRN as well?

struan: It's more that we use the UPRN later on in processing when creating any new thing. Although I guess we don't really need to as we can just copy it from the existing one.

perllib/FixMyStreet/App/Controller/Waste.pm
struan: oh, I forgot to commit the redirect param in the form field. I will push a fixup.

By default there is no redirect, you just get a bottomline confirmation page with no link back to Waste so I think we do need it.

PR #3940, opened 2022-05-19
[UK] Use a Contact type for waste/claim reports.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2447
Adds an admin field for picking the type, hopefully lets us get rid of the Contact-specific checking code for Bromley/Bucks/Kingston/Hackney/Peterborough with some generic UKCouncil code instead.
[skip changelog]


davea (approved):
Looks good! Will there be a script to set the `type` field (and remove `waste_only`) automatically on existing contacts or will that just be left to `open311-populate-service-list`?

PR #3941, opened 2022-05-19
[Peterborough] Small improvements to bin day & missed collection messages

davea:
- On the day of collection a more specific message is shown to prevent user confusion. For FD-2046
 - The email link displayed in the "There was a problem with the collection, please email" message now prepopulates the user's address in the subject line. For FD-2040.

[skip changelog]


chrismytton (approved):
Looks good!

PR #3942, opened 2022-05-20
[Kingston] Variety of waste changes

dracos:
The commit I really need reviewing is the last one, "[Waste] Use provided details, even if logged in." as that would affect current existing sites, the rest don't/shouldn't.

The two before that - "Only one garden task type ID" and "Different image lookup" - are due to me misunderstanding how data was stored, thinking there were two service tasks (one for bins, one for sacks), but looks like there is only one with different data contained within. So switching code that was already checking for two to only checking one. Then image lookup is based on service task so that needs updating to work for container (and they don't want an image for sacks at present).

The rest are smaller tweaks, can be split up as follows:

Admin improvements for Kingston:
* Add property admin search (easier to get from admin to front end)
* Add site name (hide "FixMyStreet")

Admin improvements for all:
* Factor out admin intro box (confusing to be shown)

Front end improvements for Kingston:
* Add property admin area check (prevent Sutton postcodes being looked up in Kingston, as both are in Echo)
* Don't re-sort postcode results - they're in a better order as-is (not sure about Bromley, so left that alone)

Front end improvements for all:
* Fix 404 not showing on bad ID
* Pass name to credit card processor (email/address was, just name wasn't being)

Backend for Kingston:
* Update fetching script (refactor Bromley to do same job but not with devolved contacts)

[skip changelog]

### DIFF
--- a/templates/web/base/admin/index.html
+++ b/templates/web/base/admin/index.html
@@ -29,6 +29,8 @@
     </div>
 </form>
 
+[% TRY %][% PROCESS 'admin/_index_extra_search.html' %][% CATCH file %][% END %]
+
 [% IF c.user.is_superuser %]
 <form method="get" action="[% c.uri_for('bodies') %]">
     <label for="search_body">[% loc('Edit body details') %]</label>
--- a/templates/web/base/admin/index.html
+++ b/templates/web/base/admin/index.html
@@ -1,16 +1,7 @@
 [% INCLUDE 'admin/header.html' title=loc('Summary') -%]
 [% PROCESS 'admin/report_blocks.html' %]
 
-<div class="fms-admin-info fms-admin-floated">
-This is the administration interface for [% site_name %]. If you
-need any help or guidance, there is <a href="https://fixmystreet.org/">plenty of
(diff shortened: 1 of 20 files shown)


templates/web/base/admin/index.html
```diff
@@ -1,16 +1,7 @@
...
-This is the administration interface for [% site_name %]. If you
-need any help or guidance, there is <a href="https://fixmystreet.org/">plenty of
-online documentation</a>. The FixMyStreet Platform is
-<a href="https://github.com/mysociety/fixmystreet">actively supported</a> by
-its developers, and has a community of people using or working on the code.
-Please <a href="https://fixmystreet.org/community/">sign up to the mailing list
-or get in touch</a> to let us know about your use of the FixMyStreet Platform,
-and to receive notices of updates.
-</div>
+[% PROCESS 'admin/_index_intro.html' %]
```
davea, line 4: This is a very welcome change!

davea (approved):
Haven't done a deep dive on the Kingston changes, but a quick read through they all seem OK. Have tested that final commit on Peterborough WW and it looks fine.

PR #3943, opened 2022-05-20
[Peterborough] Display help message when ‘new build’ bin request reason selected

davea:
[skip changelog]

For https://mysocietysupport.freshdesk.com/a/tickets/1911


dracos (approved):
Looks fine, but just noticed that if we ever had more than one item hint on a radio/checkboxes, they'd get the same ID? Doesn't look like we do at present (we did have one here for large family that got removed, but left the data hide bits in)

davea:
Yes, good point. Guess in that case we'd rejig it to use classes instead, or something along those lines.

Issue #3945, opened 2022-05-24
Multiple IDs when category has more than one parent


chrismytton:
When a category has multiple parent categories and the category list has been loaded in via JS there will be multiple inputs with the same `id="subcategory_1"` attribute.

This means that when the user tries clicking on the label for the input it _sometimes_ doesn't work.

I think the fix is to include `[% group.id %]` in the id attribute for this input:

https://github.com/mysociety/fixmystreet/blob/0540f7950d0a44a6575e94e2ff8a31cc970820b1/templates/web/base/report/new/subcategories.html#L9

PR #3946, opened 2022-05-24
Ensure subcategories have unique id attributes

chrismytton:
When a subcategory appears under more than one parent category ensure that it gets a unique ID so that the same ID isn't used more than once in the HTML.

Related to https://github.com/mysociety/societyworks/issues/3051

Fixes https://github.com/mysociety/fixmystreet/issues/3945

[skip changelog]


dracos (approved):
Nice they're not used by anything else :) - Doesn't affect a test or anything, but do you want to update the Cypress highways-ajax.json fixture to have these new IDs as well for consistency.

PR #3947, opened 2022-05-24
[Lincolnshire] Change asset_group for bus stops

chrismytton:
Lincolnshire have recently had a category reshuffle. As part of that they've introduced a new group for bus stops and shelters and want the bus stop assets to appear for all subcategories in that group.

Fixes https://github.com/mysociety/societyworks/issues/3052


dracos (approved):
🚌

PR #3948, opened 2022-05-25
[Hackney] Only allow 256 chars in report description

davea:
Technically the limit doesn't need to be applied to email categories, but it seemed simplest/quickest to do it this way.

Fixes https://github.com/mysociety/societyworks/issues/3056

[skip changelog]


dracos (approved):
Ooh, both sets of tests too 👍 🧊

PR #3964, opened 2022-06-06
[Kingston] Add option to hide DD and prevent action scheduled updates

dracos:
[skip changelog]


struan (approved):
Not completely ideal with the single radio button, but certainly works as an immediate solution.

PR #3966, opened 2022-06-07
Kingston bottomline api fixes

struan:
Various small fixes to the bottomline API, plus an extra method to get a mandate by reference.

[skip changelog]


dracos (approved):
Annoying, glad you worked that out.

PR #3979, opened 2022-06-15
[Buckinghamshire] Make it clear when reports are sent to the parish

chrismytton:
Rather than the text during the reporting process saying "These will be sent to Adstock Parish Council or Buckinghamshire Council and also published online" we instead want it to just say the parish name. This is because parish categories have the `prefer_if_multiple` flag, so the reports will only go to those contacts.

## Todo

- [x] Fix Cypress test that's intermittently failing because there are three bodies but the code expects there to be two (Bucks and the parish)

[skip changelog]

### DIFF
--- a/bin/fixmystreet.com/fixture
+++ b/bin/fixmystreet.com/fixture
@@ -122,7 +122,7 @@ if ($opt->test_fixtures) {
         { area_id => 2504, categories => ['Damaged, dirty, or missing bin', 'Signs and bollards', 'Busking and Street performance'], name => 'Westminster City Council', cobrand => 'westminster' },
         { area_id => 2482, categories => ['Lamp Column Damaged', 'Sign Light Not Working'], name => 'Bromley Council', cobrand => 'bromley' },
         { area_id => 164186, categories => ['Shelter Damaged', 'Very Urgent'], name => 'Northamptonshire Highways', cobrand => 'northamptonshire' },
-        { area_id => 163793, categories => ['Flytipping', 'Roads', 'Parks', 'Snow and ice problem/winter salting', 'Grass cutting'], name => 'Buckinghamshire Council', cobrand => 'buckinghamshire' },
+        { area_id => 163793, categories => ['Flytipping', 'Roads', 'Parks', 'Snow and ice problem/winter salting', 'Grass cutting', 'Hedge problem'], name => 'Buckinghamshire Council', cobrand => 'buckinghamshire' },
         { area_id => 53822, categories => [ 'Grass cutting', 'Hedge problem' ], name => 'Adstock Parish Council' }, # Buckinghamshire parish council
         { area_id => 2257, categories => ['Flytipping', 'Graffiti'], name => 'Chiltern District Council' },
         { area_id => 164186, categories => [ 'Graffiti' ], name => 'West Northamptonshire Council' },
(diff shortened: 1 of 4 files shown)


dracos (commented):
Looks like this works fine (assuming all the other parish categories that overlap with Bucks are prefer_if_multiple, which I think they are) - however, unless I'm misunderstanding something, would it not be better to look at `prefer_if_multiple` earlier in the reporting process (e.g. in the set up categories/bodies part of the code, rather than only in contacts_to_bodies) so that bodies for those categories only contain the parish, as we know that it's only going to be sent to there with prefer_if_multiple_set. Then the category_change  client side code won't needed at all because fixmystreet.bodies will be correctly set to only the parish server-side. Let me know if I've misunderstood!

chrismytton:
@dracos Have done as you suggested and pulled out the unneeded contacts earlier in the process, which seems to work nicely!

I'm not totally convinced by my implementation though, the bit that removes any contacts that aren't prefer_if_multiple seems a bit convoluted, so any suggestions for simplifying that welcome!

bin/fixmystreet.com/fixture
```diff
@@ -239,6 +238,13 @@ if ($opt->test_fixtures) {
         $child_cat->update;
     }
 
+    $child_cat = FixMyStreet::DB->resultset("Contact")->find({
+        body => $bodies->{163793},
```
dracos, line 242: Think this should be body_id and is hopefully the cause of the failing Cypress tests

chrismytton: @dracos Other parts of this file use `body => …`, I think the tests were failing in a previous commit because they had the old 2217 identifier for Bucks. Have fixed that issue in the bucks-parishes-reviewed branch and rebased this branch, so should now be passing.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -744,6 +744,24 @@ sub setup_categories_and_bodies : Private {
...
 
+    # If there are multiple contacts with the same category name and one of
+    # them has prefer_if_multiple set then use that one.
+    my %contacts_by_category;
+    foreach (@contacts) {
+        push @{$contacts_by_category{$_->category}}, $_;
+    }
+    @contacts = grep {
+        my $category = $_->category;
+        if (scalar @{$contacts_by_category{$category}} > 1 && any { $_->get_extra_metadata('prefer_if_multiple') } @{$contacts_by_category{$category}}) {
```
dracos, line 755: I don't think you need the size check, which helps (if it's size one and there is a preferred one, it'll match), and other thing I can think of is to pre-fetch the 'preferred' contact IDs and categories. Does this then work the same?

```
my @preferred = grep { $_->get_extra_metadata('prefer_if_multiple') } @contacts;
my %preferred_ids = map { $_->id => 1 } @preferred;
my %preferred_cats = map { $_->category => 1 } @preferred;

@contacts = grep { !$preferred_cats{$_->category} || $preferred_ids{$_->id} } @contacts;
```

dracos (changes requested):
Looks good - one test move suggestion and I think I have come up with a simpler grep (hopefully, I think the logic works!).

perllib/FixMyStreet/App/Controller/Report/New.pm
chrismytton: Ah that's much nicer, and seems to work the same yep!

Have pushed a fixup to my fixup.

dracos (approved):

PR #3975, opened 2022-06-14
[Peterborough] Suppress certain emails for graffiti and fly-tipping

nephila-nacrea:
As requested in https://mysocietysupport.freshdesk.com/a/tickets/1332, this PR:

- stops emails being sent for graffiti reported on public land
- stops emails being sent for fly-tipping, on both public and private land, for incident sizes of
    - 'Single Black Bag'
    - 'Single Item'
- stops emails being sent for graffiti and fly-tipping reports raised by Peterborough staff

Fixes https://github.com/mysociety/societyworks/issues/2529.

Should be deployed before https://github.com/mysociety/fixmystreet/pull/3985.

[skip changelog]

PR #3985, opened 2022-06-17
[Peterborough] Show land type to staff for graffiti & fly-tipping reports

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/2773.

Should be deployed after https://github.com/mysociety/fixmystreet/pull/3975.

Peterborough want an easy way for staff to know whether graffiti or fly-tipping reports are on private or public land. Jason's original suggestion was to change pin colours to reflect land type, and I have done this (blue for public, orange for private), but I have also included text on relevant reports for further clarity.

[skip changelog]

Screenshot 2022-06-22 at 19 34 34
Screenshot 2022-06-22 at 19 35 54
Screenshot 2022-06-22 at 19 36 13


t/cobrand/land_type_for_problem.t
```diff
@@ -0,0 +1,212 @@
+use FixMyStreet::TestMech;
```
davea, line 1: Could this all be part of `peterborough.t`? I know it's got the Bexley stuff in to check `land_type` doesn't get set on non-Peterborough cobrands but that's OK I think.

web/cobrands/sass/_base.scss
```diff
@@ -2244,6 +2244,16 @@ div.assigned-to {
     margin-top: 0.25em;
 }
 
+div.land-type-public {
```
davea, line 2247: These should go into Peterborough's `base.scss` as it only affects that cobrand.

templates/web/base/admin/reports/_edit_main.html
```diff
@@ -30,6 +30,16 @@
 [% IF problem.used_map %][% loc('used map') %][% ELSE %][% loc("didn't use map") %][% END %])</small>
 </li>
 
+[% # Peterborough staff only %]
```
davea, line 33: Because this (and the ones in `information.html` and `_item_small.html`) only apply to Peterborough it would be better to get them into Peterborough-specific templates somehow instead of doing a cobrand check in the base templates.

Generally if the template is fairly small then it's OK to override the entire thing in the cobrand dir - I'd say in this case that applies to `_item_small.html`.

For bigger templates like this one and `information.html` the `[% TRY %][% INCLUDE 'foo.html' %][% CATCH file %][% END %]` pattern allows cobrands to provide a smaller included template that provides custom markup.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -261,6 +261,21 @@ with 'FixMyStreet::Roles::Abuser',
      'FixMyStreet::Roles::Translatable',
      'FixMyStreet::Roles::PhotoSet';
 
+before update => sub {
```
davea, line 264: What are the situations where a problem's location may be updated? I think there are two (may have forgotten others!) - when a report is edited in the admin, and when its pin is dragged in the inspector form. I think rather than having this method on the `Problem` model it would be better to use/create a small cobrand hook that can make changes to a problem before it's saved. I think that would allow you to remove the `land_type_for_problem` method in `Default.pm` too.

davea (changes requested):
Overall I like the approach you've taken here and it's going in the right direction!

There is a lot of code outside of the Peterborough templates/cobrand module though, and I think for a smallish feature like this that only applies to one cobrand we should restrict the changes to that cobrand's files as much as possible.

Leaning on cobrand hooks and template overrides/includes will generally get you most of the way there (though sometimes you do need to stick in a call to a new cobrand hook at the relevant point or break a template into smaller files that can be overridden!) and keeps the core code as generic as possible.

Hope that's OK/makes sense, happy to chat about it of course.

templates/web/base/admin/reports/_edit_main.html
nephila-nacrea: I've ended up adding a single template shared across all three pages; it seemed the most succinct way.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -131,81 +139,94 @@ around 'open311_config' => sub {
...
-            # also do not want email for reports raised by staff.
-            return { method => 'Blackhole' }
-                if _is_small_flytipping_incident($problem)
-                || _is_raised_by_staff($problem);
+    $land_type = 'private' unless $land_type;
 
-            my $contact = $self->SUPER::get_body_sender($body, $problem)->{contact};
-            return { method => 'Email', contact => $contact};
-        }
+    $problem->set_extra_metadata( land_type => $land_type );
```
davea, line 202: This doesn't update the record in the DB because you're not calling `$problem->update`. This means the next time the page is loaded it'll have to do the lookups again.

I'm not certain that calling `$problem->update` is the correct/safe thing to do, however, as there's a chance the `$problem` we've been passed has unpersisted changes made elsewhere. (By that I mean if you look at all the ways this function can be called it might be as a result of a problem being loaded, having some changes applied and displayed to the user e.g. when editing a report in the admin).

I guess you could check `get_dirty_columns` and only update if there aren't any, or perhaps fetching the same row from the DB into a different variable and updating *that* as well might work, e.g.:

```perl
my $problem2 = $self->{c}->model('DB::Problem')->find({ id => $id });
$problem2->set_extra_metadata( land_type => $land_type );
$problem2->update;
```

davea (changes requested):
The changes you've made since the last review look good! I just spotted something I missed the first time (sorry!) which will improve performance a fair bit :)

davea (commented):
Looking really good, nearly there I think. Just one thing to consider is that the pin colour will change even for public users. I think a staff check in the `pin_colour` function would be a good idea.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -199,7 +199,24 @@ sub land_type_for_problem {
...
+    # problem outside this method. So we fetch a fresh version of the problem
+    # from the DB and update the extra_metadata on that.
+    my $problem_clean = FixMyStreet::DB->resultset('Problem')
+        ->find( { id => $problem->id } );
+
+    if ( $problem_clean ) {
+        $problem_clean->set_extra_metadata( land_type => $land_type );
+        $problem_clean->update;
+    }
+    else {
```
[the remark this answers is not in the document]
nephila-nacrea: ^.^

PR #3988, opened 2022-06-21
[Open311] External code removal is not a change.

dracos:
If a report had an external status code, and a new update did not change the report status and blanked out the external status code, then the code ended up looking for any response template at all. Fix this so that it only looks if either the state has changed or the code has changed but is not now blank. For FD-2124. Fixes https://github.com/mysociety/societyworks/issues/3071


chrismytton (approved):
What a subtle bug! At least the fix was fairly straightforward 🙂

PR #3989, opened 2022-06-23
[Bathnes] Accessibility colour improvement

lucascumsille:
This PR fixes issues found by the design team[ here on this Trello ticket](https://trello.com/c/HLfN7G5Q)
There were a couple of accessibility issues:

- Focus state for links too subtle. I made them match the focus state on Bathnes website.
- The contrast of the white text against the light blue of the #front-main element wasn't high enough.
- The #front-main element felt cluttered, so I added some space.
- Reduced the size of the report other issues message on the #front-main element, so the page is easier to scan and clearer which elements are more important than others.
- The colour of the stats on the homepage was changed to navy, so they passed the contrast test.

https://user-images.githubusercontent.com/13790153/175244106-fbb2ad7f-1edc-4e18-ad34-528b7764798a.mov

https://user-images.githubusercontent.com/13790153/175244070-277d1851-d925-474f-b096-ed43091054d4.mov

https://user-images.githubusercontent.com/13790153/175244250-41b07d33-ac0e-49eb-b148-d16a7a0fd791.mov

Let me know if there is any feedback.

[skip changelog]

### DIFF
--- a/web/cobrands/bathnes/base.scss
+++ b/web/cobrands/bathnes/base.scss
@@ -8,6 +8,15 @@
 @import "pattern-lib/navigation";
 @import "pattern-lib/footer";
 
+// Links
+
+a, .fake-link {
+    &:focus {
+        background-color: $bathnes-yellow;
+        outline: 3px solid $bathnes-yellow;
+    }
+}
+
 // Header
 
 #site-header {
--- a/web/cobrands/bathnes/base.scss
+++ b/web/cobrands/bathnes/base.scss
@@ -68,7 +68,22 @@
 
 #front-main #postcodeForm {
     margin-top: 1em;
+
     label {
-        color: #fff;
+        color: $primary_b;
     }
+
+    .form-hint {
+        color: #313131;
+    }
+}
+
+// Stats homepage
+
+ol.big-numbers>li:before {
+    color: $stats-colour;
+}
+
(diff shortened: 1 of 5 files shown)


web/cobrands/bathnes/base.scss
```diff
@@ -64,11 +73,39 @@
         margin: 0;
         font-size: 1.1em;
     }
+
+    #geolocate_link {
+        background-color: $bathnes-white;
+
+        &:hover {
+            background-color: $bathnes-primary;
```
dracos, line 81: Is it okay to have the background change to the same as the bar when hovered? Wondered if it should be slightly different.

dracos (commented):
One comment about the geolocate hover is all :) Do squash "Moved colours to pattern-lib" into the parent commit if you decided to move the variables' location, no need for that to be recorded.

web/cobrands/bathnes/base.scss
lucascumsille: So as long the contrast pass between the text and the background colour, and there is a noticeable difference between the states, it should be fine. For colour blind users works well due the luminosity difference between the hover and the default state.

In terms of style I agree, I fixed up so now the hover state has a light blue version.

Thanks for the feedback @dracos and I squashed the commit you mentioned before =)

PR #3990, opened 2022-06-23
2579 UI Improvements to Dropzone element

lucascumsille:
Replaces: PR #3872 
Fixes: mysociety/societyworks/issues/2579

@dracos I have made the changes proposed on #3872. Instead of having all those overrides now I have created the following variables:

- $dropzone-border-colour
- $dropzone-button-text
- $dropzone-button-background
- $dropzone-button-border

Also the $dropzone-background-color, now depend on $dropzone-button-background instead of $link-color.

This PR also includes the changes needed for specific cobrands, which were experiencing contrast and style issues.

Let me know if there is any further feedback and thanks again for your help.


dracos (changes requested):
Looks good - hope you agree that's much nicer way of setting the cobrand-specific styling :)

Only one issue I can see - when the dropzone form is used in the 'inspector form', it looks like this:
image
Hopefully there's an easy way to deal with that :)

lucascumsille:
> Looks good - hope you agree that's much nicer way of setting the cobrand-specific styling :)
> 
> Only one issue I can see - when the dropzone form is used in the 'inspector form', it looks like this: image Hopefully there's an easy way to deal with that :)

All the way, Matthew. It's much better and smooth if we need to restyle for any reason =)
I'll take a look to the inspector issue and send a fixup.

lucascumsille:
@dracos I made two minor fixes. The White-space property will prevent the button from snapping into two and the display was changed to inline-block, so there is no overlap between "Drag you photo here" and the button itself.

Screenshot 2022-07-07 at 09 45 57

Let me know how it goes =)

dracos (approved):
Great, looks good :)

lucascumsille:
> Great, looks good :)

Thanks Matthew, I'll wait for the checks to finish before merging and deploying.

PR #3991, opened 2022-06-24
Parse incoming email for subject code

dracos:
This PR first refactors handlemail so that the various aspects of it can be tested. And then adds some new functionality to look for a code in the subject and mark report fixed/use a response template like it was an Open311-fetched update.
Fixes https://github.com/mysociety/societyworks/issues/2820. [skip changelog]

### DIFF
--- a/bin/handlemail
+++ b/bin/handlemail
@@ -5,12 +5,9 @@
 #
 # This script should be invoked through the .forward mechanism. It
 # processes bounce messages and replies and deals with them accordingly.
-#
-# Copyright (c) 2016 UK Citizens Online Democracy. All rights reserved.
 
-use strict;
+use v5.14;
 use warnings;
-require 5.8.0;
 
 BEGIN {
     use File::Basename qw(dirname);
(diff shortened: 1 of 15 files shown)


chrismytton (approved):
Nice to have some tests for the handlemail script. And great to see that the Bucks change was actually quite straightforward in the end 🙂

dracos:
@davea Just the final commit e61c55f90941d0e64e0f800bc1a1589e8e1defbc needs reviewing - makes the code a little more future proof by checking the cobrand, which is in order to be able to send a 'not recognised' bounce if they don't include a code or include a bad code.

davea (approved):
Looks good!

dracos:
@davea Hiya, sorry, me again - last two commits to as discussed allow any-cat/chosen-cat templates to co-exist (preferring chosen-cat ones) , and a fixup for the email template finding as turns out that wasn't actually checking category at all. thanks.

davea (approved):
Looks good - though should there be a test in `getservicerequestupdates.t` to ensure the correct template gets chosen in all situations?

PR #3993, opened 2022-06-24
[Buckinghamshire] Mark parish reports as internal referral after sending

chrismytton:
Adds a `post_report_sent` hook to Bucks which checks if any of the receiving bodies are a know Bucks parish and if so marks the report as internal referral.

Fixes https://github.com/mysociety/societyworks/issues/3075

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -48,6 +48,11 @@ sub geocoder_munge_results {
 
 sub on_map_default_status { ('open', 'fixed') }
 
+sub around_nearby_filter {
+    my ($self, $params) = @_;
+    $params->{states}->{'internal referral'} = 1;
+}
+
 sub pin_colour {
     my ( $self, $p, $context ) = @_;
     # updated to match Oxon CC
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -144,6 +144,21 @@ sub report_new_munge_before_insert {
     $report->category('Flytipping (off-road)');
 }
 
+sub post_report_sent {
+    my ($self, $report) = @_;
+
+    my @parishes = $self->parish_bodies->all;
+    my @parish_ids = map { $_->id } @parishes;
+
+    foreach my $body_id (@{$report->bodies_str_ids}) {
+        if (grep { $body_id == $_ } @parish_ids) {
+            # Report is to a Bucks parish, mark as internal referral by default
+            $report->state('internal referral');
+            $report->update;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -388,6 +388,14 @@ sub munge_contacts_to_bodies {
     FixMyStreet::Cobrand::Buckinghamshire::munge_contacts_to_bodies($self, $contacts, $report);
 }
 
+sub post_report_sent {
+    my ($self, $report) = @_;
+
+    # Run Buckinghamshire-specific code
+    my $bucks = FixMyStreet::Cobrand::Buckinghamshire->new;
```
dracos, line 395: Needed to create a new cobrand object here, as calling it directly with the FMS cobrand, it wouldn't have access to _parish_ids.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -144,6 +144,18 @@ sub report_new_munge_before_insert {
     $report->category('Flytipping (off-road)');
 }
 
+sub post_report_sent {
+    my ($self, $report) = @_;
+
+    foreach my $body_id ($report->bodies_str_ids) {
+        if (grep $body_id, $self->_parish_ids) {
```
dracos, line 151: This needs to be `grep { $body_id == $_ } @{$self->_parish_ids}` - this is currently always matching as $body_id is always set.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -144,6 +144,18 @@ sub report_new_munge_before_insert {
     $report->category('Flytipping (off-road)');
 }
 
+sub post_report_sent {
+    my ($self, $report) = @_;
+
+    foreach my $body_id ($report->bodies_str_ids) {
```
dracos, line 150: This also needs `@{...}` around the arrayref, otherwise this loop is always called only once with $body_id set to the ref, not an ID.

dracos (changes requested):

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
dracos: Oh, and we want parish_bodies, not _parish_ids which are the area IDs.

chrismytton:
@dracos Thanks for fixing my mistakes 🙂 This PR is still marked as changes requested, but I think the changes have been done now, is that right?

I ask because I'm looking to merge this into `bucks-parishes-reviewed` so everything is in one place.

dracos:
Yeah, I think I put it on staging but not in bucks-parishes-reviewed 👍

PR #3995, opened 2022-06-27
[Waste] truncate some SCP fields which have limits

struan:
name and address fields are limited to 50 characters in the SCP api so
truncate these before sending.

[skip changelog]


dracos (approved):
Thanks :)

PR #3996, opened 2022-06-27
[Peterborough] suspend food bags temporarily

MorayMySoc:
Remove link to food bags as currently out of stock

[skip changelog]

### DIFF
--- a/templates/web/peterborough/waste/_more_services_sidebar.html
+++ b/templates/web/peterborough/waste/_more_services_sidebar.html
@@ -10,18 +10,20 @@
   <li><a href="https://www.peterborough.gov.uk/residents/rubbish-and-recycling/bin-waste-advice#additional-bins-9-0">Request an additional bin</a></li>
   [% IF show_garden_subscribe %]
     <li><a href="[% c.uri_for_action('waste/garden', [ property.id ]) %]">Subscribe to Green Garden Waste collection</a></li>
-  [% END %]
+  [% END %]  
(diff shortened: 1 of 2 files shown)


templates/web/peterborough/waste/_more_services_sidebar.html
```diff
@@ -11,17 +11,17 @@ <h3>More services</h3>
...
-        <form method="post" action="[% c.uri_for_action('waste/request', [ property.id ]) %]?bags_only=1">
-            <input type="hidden" name="token" value="[% csrf_token %]">
-            <input type="hidden" name="container-428" value="1">
-            <input type="hidden" name="quantity-428" value="1">
-            <input type="hidden" name="process" value="request">
-            <input type="submit" value="Request more food bags" class="waste-service-descriptor waste-service-link">
-          </form>
-    </li>
-  [% END %]
+  <!-- [% UNLESS waste_features.request_disabled %] -->
```
dracos, line 14: I mean, this works ;) (not sure you meant it to with the `UNLESS` matching meaning there'd be an empty `<!-- -->`.
But you can put one `<!--` at the start before the `<li>` and one `-->` at the end after the `</li>`, you don't need to comment out each line individually. And better would be to comment it in [%# because we don't need it to be in the HTML at all, do we?

dracos (changes requested):
Is okay, but I think we can comment it out server side rather than in the HTML.

dracos (approved):
👝

PR #3997, opened 2022-06-27
Trim whitespace on external status codes for response templates

nephila-nacrea:
Please check the following:

Thanks for contributing!


perllib/FixMyStreet/DB/Result/ResponseTemplate.pm
```diff
@@ -63,5 +63,39 @@ __PACKAGE__->has_many(
 
 __PACKAGE__->many_to_many( contacts => 'contact_response_templates', 'contact' );
 
-# You can replace this text with custom code or comments, and it will be preserved on regeneration
+use Utils;
+
+# Trim whitespace for external_status_code
+
+sub insert {
```
dracos, line 70: This would work okay for our uses, I think, but I wonder if it would be slightly better to override `new` and `set_column` instead of `insert` and `update`. `new` would catch new objects being created (either by new or create), whether they're saved or not; and `set_column` would catch any form of the column being changed without you needing to do the "is it in the update call or not" check you're currently doing.

```
sub new {
    my ($class, $attrs) = @_;
    $attrs->{external_status_code} = Utils::trim_text($attrs->{external_status_code})
        if $attrs->{external_status_code};
    my $new = $class->next::method($attrs);
    return $new;
}

sub set_column {
    my ($self, $column, $new_value) = @_;
    if ($column eq 'external_status_code') {
        $new_value = Utils::trim_text($new_value);
    }
    $self->next::method($column, $new_value);
};
```

dracos (commented):
Looks good, I suggest different functions it might be better to override, if I've understood it correctly.

PR #3999, opened 2022-06-30
[Kingston] Include waste data in CSV export.

dracos:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Kingston.pm
+++ b/perllib/FixMyStreet/Cobrand/Kingston.pm
@@ -792,4 +792,58 @@ sub admin_templates_external_status_code_hook {
     return $code;
 }
 
+# Include unconfirmed and hidden reports in CSV export
+sub dashboard_export_include_all_states { 1 }
+
+sub dashboard_export_problems_add_columns {
+    my ($self, $csv) = @_;
+
+    $csv->modify_csv_header( Detail => 'Address' );
+
+    $csv->add_csv_columns(
+        uprn => 'UPRN',
+        user_email => 'User Email',
+        user_phone => 'User Phone',
+        payment_method => 'Payment method',
+        payment_reference => 'Payment reference',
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Reporting.pm
```diff
@@ -123,6 +123,8 @@ sub construct_rs_filter {
         $where{"$table_name.state"} = [ FixMyStreet::DB::Result::Problem->fixed_states() ];
     } elsif ( $self->state ) {
         $where{"$table_name.state"} = $self->state;
+    } elsif ($self->cobrand->call_hook('dashboard_export_include_all_states')) {
+        # Do nothing, want all states
```
struan, line 127: Might be good to have a test that this works

struan (approved):
Two small things but looks good.

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -792,4 +792,48 @@ sub admin_templates_external_status_code_hook {
...
+    $csv->add_csv_columns(
+        uprn => 'UPRN',
+        user_email => 'User Email',
+        user_phone => 'User Phone',
+        payment => 'Payment',
+        pro_rata => 'Pro rata payment',
+        admin_fee => 'Admin fee',
+        container => 'Subscription container',
+        current_bins => 'Bin count declared',
+        quantity => 'Subscription quantity',
```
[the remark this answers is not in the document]
struan: ah yes, whatever we show to the user for the payment reference is what we want I think.

perllib/FixMyStreet/Reporting.pm
dracos: It did not. :) Due to the confirmed date range part of the query, which I've changed to created - probably better for waste anyway, given the DD reports that sit unconfirmed for weeks.

PR #4001, opened 2022-07-04
[Bucks] Add email alternative to templates

MorayMySoc:
* Adds setting to allow email alternative text to be added
* Filters out alerts that need to send the email text from general alerts of added comments
* Update database with new fields
* Create email templates for email template text
* Updates documentation to include template email text

https://github.com/mysociety/societyworks/issues/2817


chrismytton (commented):
Thanks Moray, this is looking really good! Sorry it's taken more so long to review this, I'm not familiar with the updates/alert sending code, so had to do some background research in order to review this properly!

I wonder if this could potentially be simplified by adding an `email_text` column to the comment table, rather than storing it as extra metadata. That way you wouldn't need to mess around with the RABX to get the email text back out. What do you think?

A few other things:

1. I'm not 100% convinced the `template_email_alternative` feature flag is needed. Is this just so other cobrand admins don't get surprised by an extra field on the response templates page?
2. What's the advantage of the `use_template_email` column on the response templates, rather than just checking to see if there is an email template and using that if present?
3. Why is the new `templated_email_alert-update` needed? Looks like you could change the item_text to use the email text in the same place that we're adding the "State changed to:" text (see code below) then you wouldn't need a separate email template.

https://github.com/mysociety/fixmystreet/blob/d00522421caaf7acf248180c8fad92d7358aeabf/perllib/FixMyStreet/Script/Alerts.pm#L117-L129

MorayMySoc:
I've added an email_text column to the comments and removed the extra_data stuff. I've also removed the cobrand specifics. I have left the separate template in as I think that it is a different style of email as we discussed, but happy to reconsider. Also added that these emails should go to the reporter and any subscriber should continue to receive the subscription list - which I think is working. Added a test which makes me think it it anyway!

perllib/Open311/UpdatesBase.pm
```diff
@@ -158,20 +159,24 @@ sub process_update {
...
-            ),
-            confirmed => $request->{comment_time},
-            created => $request->{comment_time},
-        }
+    my ($text, $email_text) = $self->comment_text_for_request(
+                        $request, $p, $state, $old_state,
+                        $external_status_code, $old_external_status_code
+                        );
+    if (!$email_text && $request->{email_text}) {
+        $email_text = $request->{email_text} ? $request->{email_text} : '';
```
chrismytton, line 167: Isn't `$request->{email_text}` always going to be set, based on the if statement?

templates/email/default/templated_email_alert-update.html
```diff
@@ -0,0 +1,31 @@
...
+
+PROCESS '_email_settings.html';
+
+INCLUDE '_email_top.html';
+
+%]
+
+<th style="[% td_style %][% primary_column_style %]" id="primary_column">
+  [% start_padded_box | safe %]
+  <h1 style="[% h1_style %]">New update on <a href="[% problem_url %]">[% title %]</a></h1>
```
chrismytton, line 15: It's my understanding that this email could actually contain multiple updates, which is why the original version of this email says "New updates on". I think it's ok to say new updates when there's only one update, so probably best to follow the original template here.

templates/email/default/templated_email_alert-update.txt
```diff
@@ -0,0 +1,29 @@
+Subject: New [% site_name %] council update on report: '[% report.title %]'
```
chrismytton, line 1: Same as above here, I think we should say "New [% site_name %] council update**s** on report"

chrismytton (changes requested):
Definitely getting closer! I've left a couple of comments inline, as well as the query below about updates to the Perl DB models.

---

This seems to be missing the changes from `db/rerun_dbic_loader.pl` which adds the columns to `perllib/FixMyStreet/DB/Result/Comment.pm` and `perllib/FixMyStreet/DB/Result/ResponseTemplate.pm`. When you run that command it will generate a lot of extraneous fluff, you just need the bits adding the columns, as per the following:

```diff
diff --git a/perllib/FixMyStreet/DB/Result/Comment.pm b/perllib/FixMyStreet/DB/Result/Comment.pm
index b7af66505..61b17ade7 100644
--- a/perllib/FixMyStreet/DB/Result/Comment.pm
+++ b/perllib/FixMyStreet/DB/Result/Comment.pm
@@ -72,6 +72,8 @@ __PACKAGE__->add_columns(
   { data_type => "timestamp", is_nullable => 1 },
   "send_state",
   { data_type => "text", default_value => "unprocessed", is_nullable => 0 },
+  "private_email_text",
+  { data_type => "text", is_nullable => 1 },
 );
 __PACKAGE__->set_primary_key("id");
 __PACKAGE__->has_many(
diff --git a/perllib/FixMyStreet/DB/Result/ResponseTemplate.pm b/perllib/FixMyStreet/DB/Result/ResponseTemplate.pm
index 7d6dfe29f..768013510 100644
--- a/perllib/FixMyStreet/DB/Result/ResponseTemplate.pm
+++ b/perllib/FixMyStreet/DB/Result/ResponseTemplate.pm
@@ -41,6 +41,8 @@ __PACKAGE__->add_columns(
   { data_type => "text", is_nullable => 1 },
   "external_status_code",
   { data_type => "text", is_nullable => 1 },
+  "email_text",
+  { data_type => "text", is_nullable => 1 },
 );
 __PACKAGE__->set_primary_key("id");
 __PACKAGE__->add_unique_constraint("response_templates_body_id_title_key", ["body_id", "title"]);
```

chrismytton:
Oh and I think rebasing this branch onto lastest master should hopefully fix the test failures!

chrismytton (approved):
Looks good to me! 👍 

Let's get this change on staging and see what the client thinks 🙂

MorayMySoc:
@dracos This latest commit to answer feedback on the FD ticket

dracos (changes requested):
A few comments, hopefully nothing major.

I should note that in FD, they have said "If Text is blank and text for email alert is populated, then we don't publish anything online but we do send an email to the customer" - whilst the rest of what they say is correct (for only Text, or both present), I don't believe this part is currently the functionality, and no-one has gone back to point this out, which will need doing. If text is blank, the template won't be used at all, as far as I can tell - if the admin allows blank text to be submitted, presumably that should be fixed.

bin/update-schema
```diff
@@ -215,6 +215,7 @@ else {
 # (assuming schema change files are never half-applied, which should be the case)
 sub get_db_version {
     return 'EMPTY' if ! table_exists('problem');
+    return '0078' if column_exists('response_templates', 'email_text');
```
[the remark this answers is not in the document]
dracos: I've done this as part of staging tidy.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -360,6 +359,31 @@ sub _send_aggregated_alert(%) {
     } );
     $data{unsubscribe_url} = $cobrand->base_url( $data{cobrand_data} ) . '/A/' . $token->token;
 
+# Filter out alerts that have templated email responses for separate sending and send those to the problem reporter
+    my @template_data = grep { $_->{item_private_email_text } && $_->{user_id} == $_->{alert_user_id} } @{ $data{data} };
+    @{ $data{data} } = grep {! $_->{item_private_email_text } || $_->{user_id} != $_->{alert_user_id} } @{ $data{data} };
```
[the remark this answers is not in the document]
MorayMySoc: Relating to this and the one below - yes, I had in my mind that it was a tailored response, as in a specific email. Also, the issue I was concerned about was an email saying your report had the following updates - and then when you look at the report it has a different update.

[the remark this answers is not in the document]
dracos: Makes sense on the separate email - I guess multiple private updates isn't likely to happen, but if there are 2 to be sent I think it'd be okay for them both to be in the same email, I think that still meets your thinking?

perllib/Open311/UpdatesBase.pm
```diff
@@ -287,24 +294,27 @@ sub comment_text_for_request {
...
             -or => $state_params,
         }, $order )->first) {
-            $template = $t->text;
+            $template = $t->text,
+            $template_email_text = $t->email_text,
         }
     }
 
+    $template_email_text = ($template_email_text ? $template_email_text : '');
+    ### NEED TO NOW SET THE TEST TO SET use_template_email ###
```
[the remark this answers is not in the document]
dracos: This comment line is still present, not sure what it's referring to?

perllib/Open311/UpdatesBase.pm
```diff
@@ -287,24 +294,26 @@ sub comment_text_for_request {
...
     }
 
     if ($template) {
         $template =~ s/\{\{description}}/$desc/;
-        return $template;
+        return ($template, $template_email_text);
     }
 
-    return "" if $self->blank_updates_permitted;
+    return ("", "") if $self->blank_updates_permitted;
```
dracos, line 313: Should this, and few lines down, return undef as second argument for consistency? I think that way it would always be NULL set if not present, rather than sometimes null and sometimes the empty string? If I've understood it right.

dracos (changes requested):
Couple of small issues is all.

perllib/Open311/UpdatesBase.pm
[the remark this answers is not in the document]
MorayMySoc: Earlier incarnation I had a checkbox if you wanted to use private template emails but that doesn't exist now.

perllib/Open311/UpdatesBase.pm
MorayMySoc: Also applied a couple of lines up for same reason now

dracos (approved):
Looks good :)

templates/web/base/admin/list_updates.html
```diff
@@ -39,7 +39,12 @@ <h2 id="updates">[% loc('Updates') %]</h2>
         </td>
     </tr>
     <tr>
-        <td colspan=5>[% update.text | html %]</td>
+        <td colspan=5>[% update.text | html %]
+        [% IF update.private_email_text %]
+            <br />---------------------------
+            <br />Template email response: [% update.private_email_text | html %]
```
dracos, line 45: The label needs marking for translation in a `loc()`.

templates/web/base/admin/list_updates.html
```diff
@@ -39,7 +39,12 @@ <h2 id="updates">[% loc('Updates') %]</h2>
         </td>
     </tr>
     <tr>
-        <td colspan=5>[% update.text | html %]</td>
+        <td colspan=5>[% update.text | html %]
+        [% IF update.private_email_text %]
+            <br />---------------------------
```
dracos, line 44: This could be a `<hr>`? :) Then no `<br>` needed.

dracos (approved):

PR #4002, opened 2022-07-05
[Waste] Add syslogging of SCP calls.

dracos:
[skip changelog]

### DIFF
--- a/perllib/Integrations/SCP.pm
+++ b/perllib/Integrations/SCP.pm
@@ -3,6 +3,8 @@ package Integrations::SCP;
 use Moo;
 with 'FixMyStreet::Roles::SOAPIntegration';
 
+use Data::Dumper;
+use Sys::Syslog;
 use DateTime;
 use MIME::Base64;
 use Digest::HMAC;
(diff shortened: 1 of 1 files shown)


perllib/Integrations/SCP.pm
```diff
@@ -38,6 +72,7 @@ sub call {
         }),
         make_soap_structure_with_attr(@params),
     );
+    $self->log($res);
```
struan, line 75: This is quite a verbose thing give it's the entire SOM object, but I guess don't want to try and do too much in case it's something else and we miss the response.

struan (approved):
Looks good, small thought about the response but probably too much debug for now is better

perllib/Integrations/SCP.pm
```diff
@@ -25,11 +27,43 @@ has endpoint => (
...
 
+has log_open => (
+    is => 'ro',
+    lazy => 1,
+    builder => '_syslog_open',
+);
+
+sub _syslog_open {
+    my $self = shift;
+    my $ident = $self->config->{log_ident} or return 0;
```
sagepe, line 38: If I've understood all this correctly this will be set in the config only in places where you want to enable the logging, otherwise it'll short circuit and skip the logging?

sagepe (approved):
Looks fine to me from what I can see - my comment is only really out if interest, but as I understand it you can effectively enable logging by setting `log_ident` for a given integration and it'll log, or leave it out and it won't.

PR #4003, opened 2022-07-05
[Hampshire][TfL] Update tilma links to point to live

MorayMySoc:
* Fixes two issues where calling staging mapserver instead of live

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hampshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Hampshire.pm
@@ -28,7 +28,7 @@ sub disambiguate_location {
 
 sub lookup_site_code_config { {
     buffer => 200, # metres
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
Could you put the cobrands in the commit message as you've done in the PR title?
Also wonder if in tfl/assets.js worth having a `fixmystreet.staging` check like some others do to switch between staging/live? Only if all its layers are present on both, I guess.

MorayMySoc:
> Could you put the cobrands in the commit message as you've done in the PR title? Also wonder if in tfl/assets.js worth having a `fixmystreet.staging` check like some others do to switch between staging/live? Only if all its layers are present on both, I guess.

I've copied two map files from live to staging that weren't there so should be good to have the staging/live check

dracos (approved):

PR #4004, opened 2022-07-07
[Northants] Disable pin snapping for tree layer

nephila-nacrea:
Northamptonshire would like it so users can make a report against a tree, hedge, etc. that does not appear as an asset on the map. So this PR disables pin snapping for trees; a user can select a tree asset if they wish but can also select e.g. the side of a road.

Fixes https://github.com/mysociety/societyworks/issues/2900.

[skip changelog]

### DIFF
--- a/web/cobrands/northamptonshire/assets.js
+++ b/web/cobrands/northamptonshire/assets.js
@@ -56,6 +56,7 @@ var layers = [
   ],
   "layer_name": "Tree",
   "styleid": '5d481376fe2ad8059ce44ef2',
+  "snap_threshold": 0
 },
 {
   "categories": [ "Safety Bollard - Damaged/Missing" ],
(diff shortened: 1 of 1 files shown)


web/cobrands/northamptonshire/assets.js
```diff
@@ -56,6 +56,7 @@ var layers = [
   ],
   "layer_name": "Tree",
   "styleid": '5d481376fe2ad8059ce44ef2',
+  "disable_pin_snapping": "true",
```
dracos, line 59: As discussed, and as per your comment above, this isn't actually to do with the existing `disable_pin_snapping` code which is for disabling a different type of pin snapping. So whilst you could have a different name - `disable_nearest_asset_snapping` perhaps - I think best solution is actually to set `snap_threshold` to 0 directly on this object, as the existing code checks that.

web/cobrands/fixmystreet/assets.js
```diff
@@ -1344,7 +1344,7 @@ fixmystreet.message_controller = (function() {
         },
 
         asset_not_found: function() {
-            if (!this.visibility) {
+            if (this.fixmystreet.ignore_missing_asset || !this.visibility) {
```
dracos, line 1347: I don't think we need this variable - we shouldn't be calling asset_not_found at all if we don't care about it not being found :) Proposed solution elsewhere.

dracos (changes requested):
Good work on tracking this down - I think it could be a tiny bit clearer/similar to existing code if it acted a bit 'higher up' as explained in the line comment. Hope that makes sense :)

davea (approved):

PR #4006, opened 2022-07-08
[Kingston] Fix variable clash.

dracos:
The subscribe form intro page was setting a variable for use in deciding
what step to go to next; the form processing code later on was using the
same variable to decide whether to give someone sacks or not.
[skip changelog]


davea (approved):
✅

PR #4007, opened 2022-07-08
[Open311] Check for identical latest update.

dracos:
This already existed for Merton, but it seemed like a good thing to have in general - could help with e.g. Bucks duplicate updates on occasion (though we've hidden some status codes there as a different solution).


chrismytton (approved):
♊

PR #4008, opened 2022-07-08
[Bromley] task type and renewal messages

dracos:
[skip changelog]


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -649,6 +649,8 @@ sub bin_services_for_address {
                         "$resolution_id,,",
                         $resolution_id,
                     ],
+                }, {
+                    order_by => \'length(me.external_status_code) desc',
```
struan, line 653: might be worth a comment to document how this works/why this is here

struan (approved):
Looks good.

PR #4011, opened 2022-07-15
Scroll responsibility message into view.

dracos:
If the pin was moved and the message shown, it was being scrolled into view by update_pin, but this was not happening if the message was shown after category selection. [skip changelog]


zarino (approved):
Looks good to me! :+1:

PR #4013, opened 2022-07-15
[Buckinghamshire] Use speed limit asset layer to route Grass cutting reports

chrismytton:
This uses the same code that was already in place for this, but the
question is now a hidden field which gets auto-populated by the values
in the asset layer.

[skip changelog]

Fixes https://github.com/mysociety/societyworks/issues/2812

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -384,8 +384,13 @@ sub report_new_munge_before_insert {
 sub munge_contacts_to_bodies {
     my ($self, $contacts, $report) = @_;
 
-    # Make sure Bucks grass cutting reports are routed correctly
-    FixMyStreet::Cobrand::Buckinghamshire::munge_contacts_to_bodies($self, $contacts, $report);
+    my %bodies = map { $_->body->name => 1 } @$contacts;
+
+    if ($bodies{'Buckinghamshire Council'}) {
+        # Make sure Bucks grass cutting reports are routed correctly
+        my $bucks = FixMyStreet::Cobrand::Buckinghamshire->new;
+        $bucks->munge_contacts_to_bodies($contacts, $report);
+    }
 }
 
 sub get_body_handler_for_problem {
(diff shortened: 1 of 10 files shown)


dracos (commented):
A quick idea on how this could perhaps cater for both situations.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -385,7 +385,8 @@ sub munge_contacts_to_bodies {
     my ($self, $contacts, $report) = @_;
 
     # Make sure Bucks grass cutting reports are routed correctly
-    FixMyStreet::Cobrand::Buckinghamshire::munge_contacts_to_bodies($self, $contacts, $report);
+    my $bucks = FixMyStreet::Cobrand::Buckinghamshire->new;
+    $bucks->munge_contacts_to_bodies($contacts, $report);
```
dracos, line 389: Do we need a check here so this only applies to reports within Bucks somehow?

dracos (commented):
Looks good :) One query about .com making queries it probably doesn't want to.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
chrismytton: The Bucks `munge_contacts_to_bodies` method returns early if the report doesn't have an extra field called `speed_limit_greater_than_30`, so my thinking was that would be a rare enough occurrence that it's probably safe to leave it like that. Do you think that's enough?

chrismytton: Oh and it checks that the report category is "Grass cutting". So it will only perform the WFS lookup if the report category is "Grass cutting" _and_ the report has an extra field called `speed_limit_greater_than_30`.

chrismytton: Oh I was looking at the wrong version of the code 🤦 It won't run the query if the category isn't "Grass cutting", but will run on any "Grass cutting" category, so I should probably do something a bit smarter there.

dracos: It doesn't check the field exists, does it? It not existing is the reason it then does the lookup. There are 3 councils with "Grass cutting", others are Peterborough and Bristol. So I think it would be good to have a check somehow - `munge_report_new_category_list` and `munge_report_new_bodies` in the same file both do a name look up on bodies, which I guess you have here either via the $contacts body or the bodies stash.

dracos (approved):
Looks good, assuming tests still pass :)

PR #4014, opened 2022-07-19
[Bucks] Fixes after MapIt generation update

davea:
Restores the correct functionality for Parishes after the MapIt generation update (#3933).

NB when you come to rebase the Bucks parishes work over `master`, you'll need to replace the `2217` area IDs with `163793`.


chrismytton (approved):
Thanks Dave! I absorbed the `owns_problem` change into an earlier commit after rebasing so the history makes sense. I'll merge these changes into the reviewed branch and get them deployed.

PR #4022, opened 2022-07-22
[Bexley] Add next action to updates

MorayMySoc:
For https://github.com/mysociety/societyworks/issues/3059

* Adds parameter for nsgref for open311 updates so the nsgref can be used when sending updates


davea (commented):
Looks like the right direction to me!

davea (approved):

PR #4023, opened 2022-07-22
[Bromley] Don’t munge Open311 update params for waste reports

davea:
Our open311-adapter doesn’t need these params munging in the same way
as the Bromley endpoint.

[skip changelog]


chrismytton (approved):
Looks good! There seems to be a spurious test failure, but I think that's unrelated to this change.

PR #4027, opened 2022-07-26
[TfL] Include bike number in Abandoned Santander Cycle report emails

davea:
For FD-2256

[skip changelog]

### DIFF
--- a/templates/email/tfl/submit.html
+++ b/templates/email/tfl/submit.html
@@ -57,6 +57,9 @@ of a local problem that they believe might require your attention.</p>
     [% IF report.get_extra_field_value('site') %]
       <p style="[% secondary_p_style %]"><strong>Signal site number:</strong> [% report.get_extra_field_value('site') | html %]</p>
     [% END %]
+    [% IF report.category == "Abandoned Santander Cycle" %]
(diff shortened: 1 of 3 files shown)


MorayMySoc (approved):
Looks good!

templates/email/tfl/submit.html
```diff
@@ -57,6 +57,9 @@ <h2 style="[% h2_style %]">[% report.title | html %]</h2>
     [% IF report.get_extra_field_value('site') %]
       <p style="[% secondary_p_style %]"><strong>Signal site number:</strong> [% report.get_extra_field_value('site') | html %]</p>
     [% END %]
+    [% IF report.category == "Abandoned Santander Cycle" %]
+      <p style="[% secondary_p_style %]"><strong>Bike number:</strong> [% report.get_extra_field_value('Question') | html %]</p>
+    [% END %]
```
dracos, line 62: This is fine, but just noting the default submit.html has a section to loop through and output any extra field data here. The annoyance is `hidden_field` doesn't have a description so doesn't have the nice English key name - perhaps we change the admin to allow description on hidden fields as well, and then could use that?

PR #4030, opened 2022-07-28
[Thamesmead] Stats on landing page to reflect Thamesmead only reports

MorayMySoc:
* Removes Thamesmead from best performing council table as not a council
* Removes Thamesmead categories from top five stats

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/UpdateAllReports.pm
+++ b/perllib/FixMyStreet/Script/UpdateAllReports.pm
@@ -224,9 +224,11 @@ sub generate_dashboard {
     my $last_seven_days = $rs->search({
         confirmed => { '>=', $week_ago },
     })->count;
+    my $thamesmead = FixMyStreet::DB->resultset('Body')->search({ name => 'Thamesmead' })->first;
     my @top_five_categories = $rs->search({
         confirmed => { '>=', $week_ago },
         category => { '!=', 'Other' },
+        $thamesmead ? (bodies_str => { '!=', $thamesmead->id }) : (),
     }, {
         select => [ 'category', { count => 'id' } ],
         as => [ 'category', 'count' ],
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -227,6 +227,7 @@ sub generate_dashboard {
     my @top_five_categories = $rs->search({
         confirmed => { '>=', $week_ago },
         category => { '!=', 'Other' },
+        cobrand => { '!=', 'thamesmead' },
```
davea, line 230: This would still include Thamesmead reports made on FMS.com, I think.

perllib/FixMyStreet/Script/UpdateAllReports.pm
```diff
@@ -271,6 +272,7 @@ sub calculate_top_five_bodies {
 
     my $bodies = FixMyStreet::DB->resultset('Body')->search;
     while (my $body = $bodies->next) {
+        next if $body->name eq 'Thamesmead';
```
davea, line 275: Would be better to exclude this from the `search` above, I guess, rather than skipping here.

chrismytton (commented):
Moray and I spoke on Slack about this, decided to split the changes in the cobrand file into a separate PR: https://github.com/mysociety/fixmystreet/pull/4033

Going to leave the other changes for @davea to re-review.

davea (changes requested):
I think there's a bit of tidying up to be done after #4033 was merged, and a test of the new behaviour would be good too.

MorayMySoc:
I added to the test, which fails when if this worked, I don't think it should. I've pushed it up with the error to demonstrate. Was trying to debug, but the file to messy to add to this and have run out of time, for the time being...

PR #4032, opened 2022-07-28
[Bexley] Send report P1 emails even if Symology is down

davea:
For FD-2264

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -166,8 +166,13 @@ sub admin_user_domain { 'bexley.gov.uk' }
 sub open311_post_send {
     my ($self, $row, $h, $sender) = @_;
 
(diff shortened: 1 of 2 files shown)


chrismytton (changes requested):
This doesn't seem to be having the desired effect currently:

- The test still passes without the changes in `perllib/FixMyStreet/Cobrand/Bexley.pm`
- The test fails if I add a `$mech->clear_emails_ok;` at the start of the subtest

I'm not 100% sure what's going on, but it looks like `$row->send_fail_count` is returning `undef` instead of `0`, so I guess perhaps need to discard_changes, though not sure what knock-on effects that will have?

davea:
Oops, thanks for spotting that. I realised it also wouldn't be doing the right thing for subsequent failures, nor when the report was eventually sent. Have added tests for those two situations and think it's behaving as it should now.

chrismytton (approved):
Looks good!

PR #4034, opened 2022-07-29
[Buckinghamshire] Match parish/town councils for all reports page

chrismytton:
On the staging site the Buckinghamshire parish councils were originally
loaded in by simply appending "Parish Council" to the area name in
mapit.

However as part of the Buckinghamshire parish project we got a
more accurate list of parish and town council names to use. [1]

This means we have database entries for e.g. "Aylesbury Parish Council"
(from the original import) and "Aylesbury Town Council" (from the more
recent import). There are many other examples like this.

The unused entries from the original import should all be marked as
deleted, however the code that maps URLs like /reports/Aylesbury
to a body wasn't filtering for active bodies, so the deleted bodies were
being returned.

There was also a separate bug where if there's a third body, in this
case "Aylesbury Vale District Council", the regex wasn't matching for
Parish/Town councils so it would still fail.

[1] `bin/buckinghamshire/create-parish-bodies` has a list of the place names
    that aren't simply "$name Parish Council".

Fixes https://github.com/mysociety/societyworks/issues/2939

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Reports.pm
@@ -254,7 +254,7 @@ sub rss_area_ward : Path('/rss/area') : Args(2) {
         ($c->stash->{area}) = values %$areas;
     } else {
         foreach (keys %$areas) {
-            if (lc($areas->{$_}->{name}) eq lc($area) || $areas->{$_}->{name} =~ /^\Q$area\E (Borough|City|District|County) Council$/i) {
+            if (lc($areas->{$_}->{name}) eq lc($area) || $areas->{$_}->{name} =~ /^\Q$area\E (Borough|City|District|County|Parish|Town) Council$/i) {
                 $c->stash->{area} = $areas->{$_};
             }
         }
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -365,13 +365,13 @@ sub body_find : Private {
     my ($self, $c, $q_body) = @_;
 
     # We must now have a string to check
-    my @bodies = $c->model('DB::Body')->search( { name => { -like => "$q_body%" } } )->all;
+    my @bodies = $c->model('DB::Body')->active->search( { name => { -like => "$q_body%" } } )->all;
```
dracos, line 368: Sadly, this deliberately checks deleted bodies, so that URLs for old now-deleted councils like https://www.fixmystreet.com/reports/Penwith still work. I guess it should try active ones first, and only then try deleted ones if it doesn't find one.

dracos:
@davea Hi, again, just the last commit to check it fixes the issue I raised above, thanks.

davea (approved):

PR #4040, opened 2022-08-03
[TfL] Add bike number column to csv export

MorayMySoc:
Adds a column to the csv export for bike number.

If it's a bike and a bike number has been entered it will have that for data. Otherwise 'n/a'

[skip changelog]


davea (commented):
Code-wise this looks good! But I wonder if an empty string would be better when the bike number isn't present, as the majority of reports won't be in that category so it'll end up adding a bit of noise to the generated CSV.

davea (changes requested):
Would be good to include a test for this in `Dashboard CSV extra columns` in `tfl.t`.

MorayMySoc:
I'll edit the commit message to remove n/a when I do the rebase....

davea (approved):
Looks good!

PR #4041, opened 2022-08-03
[Docs] Mac install tweaks

dracos:
Also the contents of #4021 with a dependency update, and a fix to the install script that never actually even tried to install Image::Magick. [skip changelog]


docs/install/manual-install.md
```diff
@@ -57,28 +57,27 @@ which it's much simpler to install via a packaging system.
 
 ##### i. MacPorts
 
-{% highlight bash %}
-$ port install gettext jhead libpng openssl p5-locale-gettext p5-perlmagick postgresql91-server
-{% endhighlight %}
+You will need to install gettext, jhead, libpng, openssl (perhaps version 1.1),
+p5-locale-gettext, p5-perlmagick, and postgresql13-server.
```
nephila-nacrea, line 61: Why has it changed to postgresql13-server?

bin/install_perl_modules
```diff
@@ -16,11 +16,11 @@ else
     $DIR/vendor/bin/carton install --deployment --without uk --without zurich --without kiitc
 fi
 
-if ! perl -MImage::Magick -e 'exit()' >/dev/null 2>&1
+if ! bin/cron-wrapper perl -MImage::Magick -e 'exit()' >/dev/null 2>&1
```
nephila-nacrea, line 19: What does bin/cron-wrapper do exactly?

bin/install_perl_modules
```diff
@@ -16,11 +16,11 @@ else
...
 then
     read -p "Image::Magick is not installed. Do you want to attempt to install it?" yn
     case $yn in 
-        [Yy]* ) $DIR/vendor/bin/carton install Image::Magick;;
-        * ) echo 'You will need to install it for FixMyStreet to work';;
+        [Yy]* ) $DIR/bin/cpanm -l local Image::Magick;;
```
nephila-nacrea, line 23: And similarly here - why was the previous change not working?

docs/install/manual-install.md
```diff
@@ -132,32 +131,12 @@ and compiles any translation `.mo` files (using `commonlib/bin/gettext-makemo`).
...
-$ sudo mkdir /usr/local/lib
-$ sudo ln -s $HOMEBREW_PREFIX/opt/openssl@1.1/lib/libssl.1.1.dylib /usr/local/lib/libssl.dylib
-$ sudo ln -s $HOMEBREW_PREFIX/opt/openssl@1.1/lib/libcrypto.1.1.dylib /usr/local/lib/libcrypto.dylib
-{% endhighlight %}
-
-You may also need to set one or more of the following env variables:
-
-{% highlight bash %}
-$ export OPENSSL_PREFIX=$HOMEBREW_PREFIX/opt/openssl@1.1
-$ export PKG_CONFIG_PATH="$HOMEBREW_PREFIX/opt/openssl@1.1/lib/pkgconfig"
```
nephila-nacrea, line 151: Are PKG_CONFIG_PATH, LDFLAGS, and CPPFLAGS not needed at all now?

docs/install/manual-install.md
```diff
@@ -132,32 +131,12 @@ and compiles any translation `.mo` files (using `commonlib/bin/gettext-makemo`).
...
 #### macOS troubleshooting
 
-Some Perl modules may fail to install without certain packages and paths in place.
-
-First make sure you have installed any [prerequisite packages](#pre-req-pkg-mac).
-
-For any **SSL-related modules**, you may need to create symlinks to folders Perl modules will recognise:
-
-{% highlight bash %}
-$ sudo mkdir /usr/local/lib
```
nephila-nacrea, line 142: Are these commands definitely not needed? I seem to remember having to run them.

docs/install/manual-install.md
dracos: Postgres 9 is no longer supported, best if we keep pace with what the servers are using, which is currently 13.

bin/install_perl_modules
dracos: It sets up the local::lib for letting perl scripts find modules installed in `local`.

bin/install_perl_modules
dracos: `carton install <package>` is not a valid way to call carton - it does not try and install the package, it ignores that, and is the same as running `carton install`.

docs/install/manual-install.md
dracos: No, on a fresh install on a new Mac, OPENSSL_PREFIX was all that was needed to locate all the SSL related gubbins.

docs/install/manual-install.md
dracos: As above, I hope that OPENSSL_PREFIX should be all that is needed, but I assume there are variations, it's hard to know directly. But I did not need to do anything to /usr/local or use sudo in order to install all the packages.

docs/install/manual-install.md
nephila-nacrea: I hadn't realised that. I had got confused by it saying '91' and then '13' which superficially looks like a downgrade.

nephila-nacrea:
Mostly looks good but I am unsure about the removal of the SSL-symlink portion. It could be a useful thing to try in case things are still failing.

docs/install/manual-install.md
dracos: I guess I just don’t want to encourage something (symlinking things into /usr/local) that involves something that may have unforeseen consequences (what if something else gets confused with them there), unless truly necessary. e.g. brew says: "Linking keg-only OpenSSL means you may end up linking against the insecure,
deprecated system version while using the headers from the Homebrew version." There’s lots of random things round the internet that say sudo things into places or other ‘magic’, I don’t like them much :) I have run these instructions from scratch multiple times on a fresh Mac, so I think I'd want more concrete evidence as to why these steps may be necessary, e.g. with a scenario where a Perl module did not install without this. Like I say, they all do install here with nothing in /usr/local.

nephila-nacrea: Well I can't remember what exactly was the 'fix' for me when I ran these things; it could have been one thing or a combination. If a newcomer has trouble setting things up maybe we'll have to change these docs again.

nephila-nacrea (approved):

PR #4042, opened 2022-08-04
[Bucks] Triage parish fixes

dracos:
[skip changelog] Fixes https://github.com/mysociety/societyworks/issues/3138

This turned out to need multiple steps - firstly fixing the triage list page to show all cobranded reports rather than just the cobrand body's. Next to allow Bucks staff as if they had their staff permissions on all parish bodies (so that viewing a parish report page would show the triage form). Then instead of using the unique-name dropdown, show all contacts, and update the triage server form to deal with contact IDs rather than category name. Don't think this should affect any usage (and may well need extending at some point if same thing applies to e.g. normal admin report edit page).


chrismytton (approved):
Looks good!

PR #4046, opened 2022-08-09
[Bromley] Remove Highways England from alert search

MorayMySoc:
When creating the alert feed url, remove Highways England from the bodies that are selected for covering that area

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UK.pm
+++ b/perllib/FixMyStreet/Cobrand/UK.pm
@@ -207,7 +207,10 @@ sub council_rss_alert_options {
     my ( @options, @reported_to_options );
     if ( $num_councils == 1 or $num_councils == 2 ) {
         my ($council, $ward);
-        my $body = FixMyStreet::DB->resultset('Body')->active->search({ name => { '!=' => 'TfL' } })->for_areas(keys %$all_areas)->first;
+        my $body = FixMyStreet::DB->resultset('Body')->active->search(
+            {
+                name => { -not_in => ['TfL', 'National Highways'] }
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -207,7 +207,11 @@ sub council_rss_alert_options {
     my ( @options, @reported_to_options );
     if ( $num_councils == 1 or $num_councils == 2 ) {
         my ($council, $ward);
-        my $body = FixMyStreet::DB->resultset('Body')->active->search({ name => { '!=' => 'TfL' } })->for_areas(keys %$all_areas)->first;
+        my $body = FixMyStreet::DB->resultset('Body')->active->search(
+            {
+                name => { '!=' => 'TfL' },
+                name => { '!=' => 'National Highways' }
```
dracos, line 213: You can't repeat the same key in a hash, it will just overwrite - so this will have broken the existing TfL handling, I mean (add a test for that in alert as well would be good?). `name => { -not_in => ['TfL', 'National Highways'] }` should be what you want, I think.

dracos (changes requested):
Won't work as is I'm afraid, but hopefully easy fix. Also bit annoying we need a manual list of 'bodies covering areas but we want to ignore', but can't think of any easy way of avoiding it, given that's what this code assumes. Longer term guess it could show all bodies - no reason you couldn't get an alert for all TfL reports I guess? - but this seems right way to go for now.

dracos (approved):
Thanks :)

PR #4047, opened 2022-08-09
Peterborough bulky goods: check availability of collection slots (core behaviour)

nephila-nacrea:
Core behaviour for fetching & displaying viable bulky good collection slots to a user.

Closes https://github.com/mysociety/societyworks/issues/3215.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -27,9 +27,12 @@ has_page about_you => (
 with 'FixMyStreet::App::Form::Waste::AboutYou';
 
 has_page choose_date => (
-    fields => [ 'continue', 'chosen_date' ],
+    fields => [ 'continue', 'chosen_date', 'show_later_dates' ],
     title => 'Choose date for collection',
-    next => 'add_items',
+    template => 'waste/bulky/choose_date.html',
+    next => sub {
+        $_[0]->{continue} ? 'add_items' : 'choose_date';
+    },
 );
 
 has_page add_items => (
@@ -84,10 +87,9 @@ has_field resident => (
 );
 
 has_field chosen_date => (
-    type     => 'Select',
-    widget   => 'RadioGroup',
-    required => 1,
-    label    => 'Available dates',
+    type           => 'Select',
+    widget         => 'RadioGroup',
+    label          => 'Available dates',
     options_method => sub {
         my $self = shift;
         my $c    = $self->form->c;
@@ -96,27 +98,41 @@ has_field chosen_date => (
 
         my @dates = grep {$_} map {
             my $dt = $parser->parse_datetime( $_->{date} );
-            $dt ? {
+            $dt
+                ? {
                 label => $dt->strftime('%d %B'),
                 value => $_->{date},
-            } : undef
-        } @{
+                }
+                : undef
+            } @{
             $c->cobrand->call_hook(
-                find_available_bulky_slots => $c->stash->{property}
+                'find_available_bulky_slots', $c->stash->{property},
+                $c->stash->{start_date},
             )
-        };
-
-        # XXX Hide additional dates with JS(?)
+            };
 
         # XXX Do we want to let the user know there are no available dates
         # much earlier in the journey?
         if ( !@dates ) {
             $self->inactive(1);
             $self->form->field('continue')->inactive(1);
+            $self->form->field('show_later_dates')->inactive(1);
+        }
+        elsif ( !$c->stash->{start_date} ) {
+            $c->stash->{start_date} = $dates[-1]{value};
+        } else {
+            $self->form->field('show_later_dates')->inactive(1);
         }
 
         return \@dates;
     },
+);
+
+has_field show_later_dates => (
+    type => 'Submit',
+    value => 'Show later dates',
+    element_attr => { class => 'govuk-button' },
+    order => 998,
 );
 
 has_field tandc => (
@@ -188,5 +204,12 @@ has_field item2 => item_field(2);
 has_field item3 => item_field(3);
 has_field item4 => item_field(4);
 has_field item5 => item_field(5);
+
+sub validate {
+   my $self = shift;
+   $self->field('chosen_date')->add_error('Available dates field is required')
+       if( $self->field('continue')->input &&
+           !$self->field('chosen_date')->has_value);
+}
 
 1;
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -1,6 +1,7 @@
 package FixMyStreet::App::Form::Waste::Bulky;
 
 use utf8;
+use DateTime::Format::Strptime;
 use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Waste';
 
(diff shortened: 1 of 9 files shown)


davea (commented):
This is definitely going in the right direction :)

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -467,26 +493,111 @@ sub image_for_unit {
...
+        }
+
+        my $total_collection_slots = keys %jobs_per_uprn;
+
+        # Only include if max jobs not already reached
+        push @available_slots => {
+            workpack_id => $workpack->{id},
+            date        => $workpack->{WorkPackDate},
+            }
+            if $total_collection_slots < max_daily_bulky_collection_slots();
```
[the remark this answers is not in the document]
davea: A couple of other thoughts on making the page faster by reducing time spent waiting on API calls:

- Could make the `WorkPacks_Get` and subsequent `Jobs_Get` calls in parallel, like we do in `Integrations::Echo::call_api`
- The `WorkPacks_Get` calls could be reduced to a single call covering the entire date range (as long as it didn't return 1000 results, as then we wouldn't know if we were missing data)

None of this has to be done now, just noting it.

perllib/Integrations/Bartec.pm
```diff
@@ -175,6 +175,46 @@ sub Premises_Get {
...
+        'WorkPacks_Get',
+        token => $self->token,
+        Date  => {
+            MinimumDate => {
+                attr  => { xmlns => "http://www.bartec-systems.com" },
+                value => $date,
+            },
+            MaximumDate => {
+                attr  => { xmlns => "http://www.bartec-systems.com" },
+                value => $date,
```
davea, line 210: Although having the timestamp set to `00:00:00` for both `MinimumDate` and `MaximumDate` does appear to work, do you think it would be worth explicitly setting it to `23:59:59` for the latter?

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -80,19 +81,57 @@ has_field resident => (
...
-    label => 'Available dates',
-    options => [
-        # XXX look these up dynamically
-        { value => '20220825', label => '2022-08-25' },
-        { value => '20220826', label => '2022-08-26' },
-        { value => '20220827', label => '2022-08-27' },
-        { value => '20220828', label => '2022-08-28' },
-    ],
+    label    => 'Available dates',
+    options_method => sub {
```
davea, line 88: This method seems to get called again for subsequent pages in the process (twice in fact when going to the "add items" page), making everything feel very slow :( I wonder if there's a way to improve that by caching the result, or making sure it's only called on the relevant page, or skipped if the form data already has a `chosen_date` value?

perllib/Integrations/Bartec.pm
nephila-nacrea: I'm not so sure about this; mainly because date munging of any kind can be such a faff, and it works as it currently stands.

nephila-nacrea: (As far as I have been able to tell.)

davea: This ought to do it :) 
```
my $start = $date->clone->set(hour => 0, minute => 0, second => 0);
my $end = $date->clone->set(hour => 23, minute => 59, second => 59);
```

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
nephila-nacrea: I now cache the result, though the duplicate calls / extra calls on subsequent pages ought to be addressed ultimately. I think FormHandler reached a point where it was doing my head in :-(

perllib/FixMyStreet/Cobrand/Peterborough.pm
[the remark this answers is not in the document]
nephila-nacrea: Looking into parallelisation would be good. I don't know if I ought to do it as part of this PR or a new one.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -476,26 +511,157 @@ sub image_for_unit {
...
+        = DateTime::Format::Strptime->new( pattern => '%d%m%y' );
+    my $workpack_date_pattern = '%FT%T';
+    my $workpack_date_parser
+        = DateTime::Format::Strptime->new(
+        pattern => $workpack_date_pattern );
+
+    my $last_workpack_date;
+    for my $workpack (@$workpacks) {
+        # Can be an arrayref or a hashref
+        my $action_data = $workpack->{Actions}{Action};
```
MorayMySoc, line 555: It looks like `$workpack->{Actions}` can be an arrayref if there is more than one `{Action}` in which case get the error `Not a HASH reference`

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -76,20 +110,126 @@ has_field resident => (
...
-    ],
+has_field chosen_date_earlier => (
+    _chosen_date_settings(),
+    options_method => sub {
+        my $self = shift;
+        my $form = $self->form;
+        my $c    = $form->c;
+
+        my @dates = _get_dates($c);
+
```
MorayMySoc, line 121: This blocks the page moving forwards while it loads - so the loading happens on the about_you page. It may be a minor thing, but I think it would make sense to move forward to the choose_date_earlier and for loading to take place there. How? Not entirely sure...

perllib/FixMyStreet/Cobrand/Peterborough.pm
MorayMySoc, line 555: It looks like `$workpack->{Actions}` can have an arrayref as a value if there is more than one `{Action}` in which case get the error Not a HASH reference (Edit: which I see you've noted - so I guess it should be my `$action_data = $workpack->{Actions};` and the rest of the code deals with popping it in an arrayref - but it does break [t/app/controller/waste_peterborough.t](https://github.com/mysociety/fixmystreet/pull/4047/files#diff-83fd9cc62337a88c4ff8caeda4a669b3a3717c8b3be8c53100edb9004232e71b))

MorayMySoc (commented):
This is looking really good to me - I've left a few comments on things I thought/noticed, but it seems to be doing the right thing. Dave will need to have a look again though

davea (commented):
Good job getting FormHandler to bend to your will!

There's something a bit odd though - if you select a date from the first set, then click "Show later dates" and select one of those before continuing, the code doesn't know which date you actually wanted as they're stored in two separate variables that both have values (`chosen_date_earlier` and `chosen_date_later`). I think this means when we reach the summary page there's no way to know for sure which date to book? Might have misunderstood.

Hopefully this illustrates it a bit better:


https://user-images.githubusercontent.com/4776/192546239-3bf4427a-f0fc-44ab-aaf7-d0346cb0df3e.mov

One other thing that springs to mind is that although the current approach of dividing all dates into "earlier" and "later" works for the timespans of the spec (i.e. show all dates up to X months in the future) it makes it harder to extend in the future.

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
nephila-nacrea: We could potentially do an asynchronous fetch of data via JS, but that may be outside the scope of the (more-or-less) minimum viable product we are aiming for (it also doesn't really solve the problem of the API calls being verrrry sloooow - I may need to look at the parallelisation work that Dave did for other calls).

perllib/FixMyStreet/Cobrand/Peterborough.pm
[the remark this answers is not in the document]
davea: Have [ticketed](https://github.com/mysociety/societyworks/issues/3258) for a PR later on.

davea (approved):
Looks good!

PR #4051, opened 2022-08-10
Make sure staff permissions removed when anonymized

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2991. Fixes https://github.com/mysociety/societyworks/issues/3027 - carries on from https://github.com/mysociety/fixmystreet/pull/3874, makes sure permissions are removed along with the from_body. Existing entries in db will need updating to match, I think.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Users.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Users.pm
@@ -32,13 +32,11 @@ sub index :Path : Args(0) {
         my $user_rs = FixMyStreet::DB->resultset("User")->search({ id => \@uids });
         if ( $c->get_param('remove-staff') ) {
             foreach my $user ($user_rs->all) {
+                $user->remove_staff;
                 $user->update({
-                    from_body => undef,
                     email_verified => 0,
                     phone_verified => 0,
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -32,13 +32,11 @@ sub index :Path : Args(0) {
         my $user_rs = FixMyStreet::DB->resultset("User")->search({ id => \@uids });
         if ( $c->get_param('remove-staff') ) {
             foreach my $user ($user_rs->all) {
+                $user->remove_staff;
                 $user->update({
-                    from_body => undef,
                     email_verified => 0,
                     phone_verified => 0,
```
MorayMySoc, line 38: So, for my education, why not move phone_verified, email_verified to remove_staff too?

dracos: You can remove staff without unverifiying the user? that's what happens by the middle remove_staff call, when someone removes the body from a user account, but the account remains verified.

MorayMySoc (approved):
All good!

Issue #3330, opened 2021-02-25
Spot anonymous reports before sending questionnaires, auto-replies or other responses by email [4hrs]


sagepe:
**Describe the bug**
Sometimes questionnaires, auto-replies from bodies and other email is generated on anonymously made reports that do not have a valid email address associated.

**Expected behavior**
No emails should be sent for anonymous reports.

dracos:
Reports made as body or anonymous will no longer send questionnaires.

davea:
Moderation emails are incorrectly being sent for anonymous reports.

jacquelinebylau:
Toggl: https://plan.toggl.com/#timeline/317959/tasks/33852437/embed

PR #4052, opened 2022-08-10
Do not send moderation emails to anon accounts.

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/3330. [skip changelog]


davea (approved):
Nice and neat 👍

PR #4054, opened 2022-08-11
Keep hidden area IDs from being removed from a body.

dracos:
[skip changelog] Fixes https://github.com/mysociety/societyworks/issues/2990


chrismytton (approved):
Smaller fix than I anticipated, nice!

PR #4055, opened 2022-08-11
Fix intermittently failing public_asset_ids.t

nephila-nacrea:
Possibly was failing due to original external ID strings (0123,
0234, etc.) also sometimes matching the randomly generated CSRF
token or CSP nonce in the HTML. So external IDs are now prefixed
with the non-base64 char €.

Fixes https://github.com/mysociety/societyworks/issues/3153.

[skip changelog]


dracos (approved):
Thanks :)

PR #4056, opened 2022-08-11
[Northamptonshire] Include report's external ID in CSV export

chrismytton:
This adds an extra "External ID" column to the dashboard CSV export which Northamptonshire have requested so they can map updates to reports in their new system.

Part of https://github.com/mysociety/societyworks/issues/3131

[skip changelog]


MorayMySoc (approved):
All good

PR #4059, opened 2022-08-12
For CSV export, fetch children of all generations.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3156

Area IDs are stored on a report at creation; if the areas in MapIt change, due to boundary changes, those IDs no longer match against the data in the current MapIt generation. So fetch all generations of children so old ward information is included.


davea (approved):
📛 🆕 🆗

PR #4060, opened 2022-08-15
Check asset layer body visibility at creation.

dracos:
Previously always_visible layers were set to visible at setup, but then when the layer was initialized update_layer_visibility was called which would set the visibility to false if the body did not match. However by then a layer data request would already have been made.
[skip changelog]


MorayMySoc:
@dracos Cypress test failures seem to be genuine and not misfires...

dracos:
Yeah, ta, will have to look into them - I assume it's missing a wait that's now necessary or something, or perhaps something worse!

dracos:
@MorayMySoc tests should be okay now

MorayMySoc (approved):
Seems good to me

PR #4063, opened 2022-08-16
[Waste] DD uniqueness, renewal cobrand_data, New dd_date, auto-internal created time

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3169 - This makes the Kingston DD reference code the general code, so replaces Bromley's existing. Leave the old GGW handling for any DDs that come in overlap.
Fixes https://github.com/mysociety/societyworks/issues/3114 - make sure the renewals set the correct cobrand_data
Fixes https://github.com/mysociety/societyworks/issues/3149 - deal with the cron output
Fixes https://github.com/mysociety/societyworks/issues/3171 - fix the update order issue (I think)
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/DDProcessor.pm
+++ b/perllib/FixMyStreet/Roles/DDProcessor.pm
@@ -302,6 +302,7 @@ sub _duplicate_waste_report {
         anonymous => $report->anonymous,
         state => 'unconfirmed',
         non_public => 1,
+        cobrand_data => 'waste',
     });
 
     $extra->{$self->garden_subscription_container_field} ||= $report->get_extra_field_value($self->garden_subscription_container_field);
--- a/perllib/FixMyStreet/Roles/DDProcessor.pm
+++ b/perllib/FixMyStreet/Roles/DDProcessor.pm
@@ -102,6 +102,10 @@ sub waste_reconcile_direct_debits {
                 next unless FixMyStreet::DB::Result::Problem->visible_states()->{$cur->state};
                 my $sub_type = $cur->get_extra_field_value($self->garden_subscription_type_field);
                 if ( $sub_type eq $self->waste_subscription_types->{New} ) {
+                    # already processed - need this because a Renewal report
+                    # may have been switched to a New if the backend
+                    # subscription had expired
+                    next RECORD if $cur->get_extra_metadata('dd_date') && $cur->get_extra_metadata('dd_date') eq $date;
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/Roles/DDProcessor.pm
```diff
@@ -316,14 +321,37 @@ sub _duplicate_waste_report {
...
 
-    (my $uprn = $payer) =~ s/^GGW//;
+    # Old style GGW references
+    if ((my $uprn = $payer) =~ s/^GGW//) {
+        my $len = length($uprn);
+        my $rs = FixMyStreet::DB->resultset('Problem')->search({
+            extra => { like => '%uprn,T5:value,I' . $len . ':'. $uprn . '%' },
+        }, {
+            order_by => [ { -desc => 'created' }, { -desc => 'id' } ],
+        })->to_body( $self->body );
```
davea, line 331: If I'm reading this right it will find the most recent report to Bromley/whoever that has the right `uprn` value in its extra regardless of category etc. I think it might be worth tightening up the search criteria if at all possible just to reduce the chance of some future bug if e.g. an unrelated category added a `uprn` extra or something.

davea (changes requested):
Looks good, one comment though - might not be an issue! Can you please reference the relevant tickets in the commit messages to give future historians a bit more context?

perllib/FixMyStreet/Roles/DDProcessor.pm
dracos: Looks like it always does a further search for category after it's returned. Not sure why it can't do it here, though, you're right

dracos: Never sure about leaving private issue links in public commits, but yeah, should probably do so here, and give better commit messages (looks like I managed 1 out of 4?).

davea: Yeah, it's not super helpful for anyone without access to the issue in question but it doesn't give anything private away (IMO) and most devs who work on the codebase do have access.

dracos: Just to note for future, one issue with it is if you do lots of rebasing of a commit with an issue ref contained, each rebased commit appears on that issue's page as a commit pointing at it, for ever - see e.g. https://github.com/mysociety/societyworks/issues/676 for an extreme case. Would be good if you could associate a PR description with the commits somehow!

PR #4067, opened 2022-08-19
[Lincolnshire] Change default_map_zoom

chrismytton:
Lincolnshire have requested that the map be more zoomed in by default.

Fixes https://github.com/mysociety/societyworks/issues/3173


MorayMySoc (approved):
All good!

MorayMySoc (changes requested):
Ooops...failing cypress test. Might just be a glitch but best to check

chrismytton:
> Ooops...failing cypress test. Might just be a glitch but best to check

Re-ran the tests and they're passing this time, so just a transient Cypress glitch it seems!

MorayMySoc (approved):
Really all good

PR #4069, opened 2022-08-22
Check report body for category on incoming email.

dracos:
If an incoming email concerns a report for a body other than the one we are using to look up a response template, do not match on categories, as they should only apply to reports for that body. This is so e.g. if a category exists for both Bucks and a parish, a reply to a parish report will not use a matching Bucks template. [skip changelog]


davea (approved):
LGTM 👍

PR #4073, opened 2022-08-23
[Northamptonshire] Enable multiple photos over open311

chrismytton:
Fixes https://github.com/mysociety/societyworks/issues/3196

[skip changelog]


dracos (approved):

chrismytton:
Multi-photos is already enabled by the `with 'FixMyStreet::Roles::Open311Alloy';` line in the Northamptonshire cobrand.

The actual issue is that multiple photos aren't being sent at all for updates, which is tracked in https://github.com/mysociety/societyworks/issues/3200.

PR #4076, opened 2022-08-24
[Thamesmead] Integrate Thamesmead/Peabody with FMS/Bexley/Greenwich

MorayMySoc:
* Change Thamesmead body/category filters to be dependent on if point selected is a Thamemead asset
* Add preselected category field to ajax data to select the correct category for Thamesmead asset
* Switch to standard message handling and make layers non interactive so pin is not affected

Fixes https://github.com/mysociety/societyworks/issues/3069
Fixes https://github.com/mysociety/societyworks/issues/3137

[skip changelog]


perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -115,4 +115,60 @@ sub geocoder_munge_results {
 
 }
 
+sub report_new_is_on_thamesmead_area {
```
chrismytton, line 118: I'm not totally sold on the name of this method, but I can't come up with anything better. I think my issue is the name suggests it's going to return a boolean, but it actually returns a category.

I thought about `report_new_is_on_thamesmead_area_and_return_area_type`, but perhaps that's a bit of a mouthful!

Happy to leave this as-is for now, but just flagging in case you had any immediate thoughts on an alternative name.

t/Mock/Tilma.pm
```diff
@@ -10,6 +10,58 @@ has json => (
...
+   xmlns:ms="http://mapserver.gis.umn.edu/mapserver"
+   xmlns:gml="http://www.opengis.net/gml"
+   xmlns:wfs="http://www.opengis.net/wfs"
+   xmlns:ogc="http://www.opengis.net/ogc"
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="http://mapserver.gis.umn.edu/mapserver https://tilma.staging.mysociety.org:80/mapserver/thamesmead?SERVICE=WFS&amp;VERSION=1.1.0&amp;REQUEST=DescribeFeatureType&amp;TYPENAME=hardsurfaces&amp;OUTPUTFORMAT=SFE_XMLSCHEMA  http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
+   <gml:boundedBy>
+      <gml:Null>missing</gml:Null>
+   </gml:boundedBy>
+</wfs:FeatureCollection>';
```
chrismytton, line 63: I'd be tempted to put these two variables inside the `/mapserver/thamesmead` route handler, since they're not used anywhere else and are specific to that route.

chrismytton (approved):
Couple of minor comments, but no blockers so happy to approve this! 👍

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -268,6 +268,7 @@ sub report_form_ajax : Path('ajax') : Args(0) {
         $top_message ? (top_message => $top_message) : (),
         unresponsive => $c->stash->{unresponsive}->{ALL} || '',
         by_category => \%by_category,
+        preselected => $c->stash->{'preselected_categories'} || {},
```
dracos, line 271: Could this behave the same as top message, subcategories, display_names above - not outputting an empty hash if there's nothing, please.

perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -115,4 +115,60 @@ sub geocoder_munge_results {
...
+
+        my $cfg = {
+            url => "https://tilma.mysociety.org/mapserver/thamesmead",
+            srsname => "urn:ogc:def:crs:EPSG::27700",
+            typename => $typename,
+            filter => "<Filter><Contains><PropertyName>Extent</PropertyName><gml:Point><gml:coordinates>$x,$y</gml:coordinates></gml:Point></Contains></Filter>",
+            outputformat => "GML3",
+        };
+
+        my $features = FixMyStreet::Cobrand::UKCouncils->new->_fetch_features($cfg, $x, $y);
```
dracos, line 138: So this is going to make 5 external Tilma calls every time somewhere in Bexley or Greenwich is clicked? That doesn't sound ideal. Can we not make one request with all the typenames at once and see which one we get back?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -448,17 +455,21 @@ sub _fetch_features {
...
     my $response = get($uri) or return;
-
     my $j = JSON->new->utf8->allow_nonref;
     try {
         $j = $j->decode($response);
     } catch {
         # There was either no asset found, or an error with the WFS
         # call - in either case let's just proceed without the USRN.
-        return;
+        # Additionally try for a different map type response
```
dracos, line 464: This is not a good way to do this. If you know that you're going to be requesting XML in fetch_features, it should be passed in as a variable - not trying to decode as JSON and then trying XML if that fails. But a better solution would be to request JSON like all the other uses of this function do, which will need a config line adding to Thamesmead's .map file to say JSON output is allowed.

MorayMySoc: I've worked this through with the current maps which are on Tilma, but there's a PR I'm just waiting for them to confirm which takes four of the five layers to their mapservers. So I guess I'll need to go with passing in a variable... https://github.com/mysociety/fixmystreet/pull/4065

dracos: Ah, and they might not have JSON output then, okay. You should still add a parameter to fetch_features saying you expect/want XML then, I think, clearer than knowing you're asking for XML but trying to parse it as JSON first.

web/cobrands/fixmystreet/assets.js
```diff
@@ -1224,9 +1224,10 @@ fixmystreet.message_controller = (function() {
         }
     }
 
-    // Show the reporting form, unless the road responsibility message is visible.
+    /* Show the reporting form, unless the road responsibility message is visible or we are on a
+    Thamemead asset as Thamesmead assets controlling enabling/disabling */
     function enable_report_form() {
-        if ( $('#js-roads-responsibility').is(':visible') ) {
+        if ( $('#js-roads-responsibility').is(':visible') || fixmystreet.bodies.indexOf('Thamesmead') === 0 ) {
```
dracos, line 1230: What is calling `enable_report_form` that makes this necessary? Seems odd this wouldn't have come up before somewhere. Is it because Thamesmead code isn't using the standard 'message_controller.asset_not_found' asset code - is there a reason for that?

dracos (changes requested):
A few things

t/Mock/Tilma.pm
MorayMySoc: As I was making other changes, I've done this.

perllib/FixMyStreet/App/Controller/Report/New.pm
MorayMySoc: This is done, and the cypress highways test restored to its original state.

perllib/FixMyStreet/Cobrand/Thamesmead.pm
MorayMySoc: I've got this making the one call (thanks for the help - I got the brackets wrong and then thought it an unlikely solution anyway so was trying other stuff). Problem at the moment is that $features may now be an array ref as planters/trees seem to occupy the same/similar locations

perllib/FixMyStreet/Cobrand/UKCouncils.pm
MorayMySoc: This hopefully now as wanted

web/cobrands/fixmystreet/assets.js
MorayMySoc: For some reason I had problems at the time with the message_controller, (which may well have been my fault) so I did it the way it is.

MorayMySoc:
These changes aren't quite finished, I'm afraid, but I'm out of time - hopefully notes make sense on individual areas

perllib/FixMyStreet/Cobrand/Thamesmead.pm
MorayMySoc: It should probably just be return_area_type, thinking about it?

perllib/FixMyStreet/Cobrand/Thamesmead.pm
dracos: I'm going to make the assumption that planters/beds are a subset of trees, looks like it from e.g.
 vs

web/cobrands/fixmystreet/assets.js
dracos: So on mobile currently if you pick a location that doesn't have an asset, it won't let you continue, but with no explanation as to why. The standard way would get the message shown to the user.

dracos:
@chrismytton asking for review of the new commits, which are as follows:
4f4c1dbfb729f8d1c9f87327214309134f419452 - fix the fetching of one area type from tilma, tidy up that code;
119fb2f155d7b1a9b1961979690fd8fcc6a1799d - switch to non-interactive layers. This way they can still be displayed and used for boundary checks, but the pin remains visible and can be placed anywhere within a polygon.
a7b1baf900c0ec3941f9581629dae3d6d1c4ae0e - improve the geocoded - switch to bounds that are actually used, and do not add "Thamesmead" as that is not always present, hopefully the bounds provide good enough guidance there.

chrismytton (approved):

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -447,18 +454,33 @@ sub _fetch_features {
     }
 
     my $uri = $self->_fetch_features_url($cfg);
+    warn $uri;
```
dracos, line 457: debugging left in

PR #4079, opened 2022-08-25
CSV export, instantantiate cobrand class.

dracos:
Until recently, nothing tried to use the class as an instance, only call simple functions, but the Buckinghamshire parish_bodies call caches the fetched bodies on the cobrand. [skip changelog]
Fixes https://mysocietysupport.freshdesk.com/a/tickets/2342


davea (approved):
Oops! Nice find.

PR #4080, opened 2022-08-25
[Waste] Remove hardcoded ‘GGW’ references

davea:
Just a bit of tidying to make it easier for non-GGW modules to use payments code.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -218,20 +218,20 @@ sub pay : Path('pay') : Args(0) {
         amount => $amount,
         reference => $payment->config->{customer_ref},
         description => $p->title,
-        lineId => $c->cobrand->call_hook(waste_cc_payment_line_item_ref => $p) || "GGW$uprn",
+        lineId => $c->cobrand->waste_cc_payment_line_item_ref($p),
     });
     if ($admin_fee) {
         push @items, {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1237,7 +1237,7 @@ sub setup_garden_sub_params : Private {
         $service_id = $c->cobrand->garden_service_id;
     }
     $c->set_param('service_id', $service_id);
-    $c->set_param('client_reference', 'GGW' . $c->stash->{property}->{uprn});
+    $c->set_param('client_reference', $c->cobrand->garden_waste_subscription_client_ref);
```
dracos, line 1240: I think we can drop this line. It is not used, and is therefore confusing if people think it is - no contact has it as a parameter, so open311-adapter generates its own from the report ID (which is good because otherwise I think there would have been uniqueness issues).

dracos, line 1240: I think we can drop this line. It is not used, and is therefore confusing if people think it is - no contact has it as a parameter, so open311-adapter generates its own from the report ID (which is good because otherwise I think there would have been uniqueness issues).

dracos (changes requested):
I think one thing can be dropped is all, unless I've missed something

dracos (approved):

PR #4081, opened 2022-08-25
[Peterborough Bulky Goods] Basic implementation of booking wizard & admin config editor

davea:
This implements the basic wizard for booking a bulky collection, albeit with lots of hardcoded values and behaviour that is not yet correct according to the tech spec.

It's behind a feature flag that must be enabled:

```
COBRAND_FEATURES:
  waste_features:
    peterborough:
      bulky_enabled: 1
```

There is also an config editor for admins that will allow certain aspects of the bulky goods system to be controlled. This was reviewed in https://github.com/mysociety/fixmystreet/pull/4090 (though I've moved the new permission from `Default::available_permissions` to `Peterborough::available_permissions` and put it behind the feature flag). This also needs a feature flag to be active:

```
COBRAND_FEATURES:
  waste_features:
    peterborough:
      admin_config_enabled: 1
```

Per recent discussion I thought having this merged to master will make collaborating on subsequent features easier and reduce the code review burden ahead of go-live.

[skip changelog]

### DIFF
--- a/templates/web/base/waste/bulky/confirmation.html
+++ b/templates/web/base/waste/bulky/confirmation.html
@@ -1,0 +1,20 @@
+[% SET title = form.title ~%]
+[% PROCESS 'waste/header.html' %]
+
+<div class="govuk-panel govuk-panel--confirmation">
+    <h1 class="govuk-panel__title">
+        [% title %]
+    </h1>
+    <div class="govuk-panel__body">
+        <p>Your [% title | lower %];
+          [% IF report.user.email && report.get_extra_metadata('contributed_as') != 'anonymous_user' %]
+            a copy has been sent to your email address, [% report.user.email %].
+          [% END %]
+        </p>
+        <p>
+          [% INCLUDE 'waste/_report_ids.html' %]
+        </p>
+    </div>
+</div>
+
+[% INCLUDE footer.html %]
--- a/templates/web/base/waste/bulky/intro.html
+++ b/templates/web/base/waste/bulky/intro.html
@@ -1,0 +1,9 @@
+<h3>Before you start your booking</h3>
+<ol>
+    <li>Requesting a bulky waste collection usually takes approximately <strong>X minutes</strong></li>
+    <li>You can request up to <strong>5 items per collection</strong></li>
+    <li>In order to help us with your collection, we will ask you to optionally add pictures of the items to be collected and the location</li>
+    <li>Before confirming your booking, check all the info provided is correct</li>
+    <li>You can cancel your booking anytime up until 23:55 the day before the collection is scheduled</li>
+    <li><strong>Cancellations made more than 24 hours</strong> before collection is scheduled are entitled to a refund.</li>
+</ol>
--- a/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
@@ -1,0 +1,116 @@
+package FixMyStreet::App::Controller::Admin::Waste;
+
+use JSON::MaybeXS;
+use Moose;
+use Try::Tiny;
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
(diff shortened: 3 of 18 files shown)


templates/web/base/waste/bulky/confirmation.html
```diff
@@ -0,0 +1,23 @@
...
+        [% title %]
+    </h1>
+    <div class="govuk-panel__body">
+        <p>Your [% title | lower %];
+          [% IF report.user.email && report.get_extra_metadata('contributed_as') != 'anonymous_user' %]
+            a copy has been sent to your email address, [% report.user.email %].
+          [% END %]
+        </p>
+        <p>
+          [% IF report.category == 'Request new container' %]
```
dracos, line 15: This presumably isn't needed at this point.

templates/web/base/waste/bulky/choose_date.html
```diff
@@ -0,0 +1,15 @@
+[% SET title = form.title ~%]
```
dracos, line 1: This template is identical to waste/index.html - can we not use the default?

templates/web/base/waste/bulky/add_items.html
```diff
@@ -0,0 +1,15 @@
+[% SET title = form.title ~%]
```
dracos, line 1: Ditto

templates/web/base/waste/bulky/intro.html
```diff
@@ -0,0 +1,26 @@
+[% SET title = form.title ~%]
```
dracos, line 1: The Claims form introduced an intro_template variable in order to provide a static block above a form on a particular page - perhaps worth reusing/generalising that into the waste template as well, and then that could be used here? Just an idea.

templates/web/base/waste/bulky/residency_check.html
```diff
@@ -0,0 +1,15 @@
+[% SET title = form.title ~%]
```
dracos, line 1: Another identical template

dracos (changes requested):
A few small template changes/improvements is all

templates/web/base/waste/bulky/choose_date.html
davea: The identical template was deliberate to make it easier for someone else to pick up the implementation later on. Perhaps just a one-liner like `[% INCLUDE 'waste/index.html' %]` would be OK?

templates/web/base/waste/bulky/confirmation.html
davea: Ah yes, good spot.

templates/web/base/waste/bulky/choose_date.html
dracos: I guess, but I'd hope that for normal form pages, there wouldn't be a need for a custom HTML page at all, it doesn't look like there generally is.

dracos: Sorry if this wasn't clear, reading it back - I meant that the default template is picked up as a fallback automatically, so I'm not sure specifying templates here in advance makes it easier to pick up later on, because there shouldn't (hopefully/ideally) be anything that needs doing except for very custom pages (like eg the summary page).

dracos:
Also note test failure in what appears to be a relevant test, but couldn't obviously see the cause.

templates/web/base/waste/bulky/choose_date.html
davea: Gotcha. Will update accordingly.

templates/web/base/waste/bulky/intro.html
davea: Oh that's nicer isn't it. Have reused that here as you suggest.

davea:
Looks like the default value for `allow_nonref` changes across versions, so have set it explicitly in 92a84505cead350709caba267c01de8af55b0133 which seems to have cleared up that test failure.

dracos (approved):

PR #4082, opened 2022-08-26
Add role filter to dashboard interface.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2446

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -238,6 +238,24 @@ __PACKAGE__->has_many(
   },
 );
 
+__PACKAGE__->might_have(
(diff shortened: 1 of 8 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -238,6 +238,24 @@ __PACKAGE__->has_many(
   },
 );
 
+__PACKAGE__->might_have(
+  contributed_by => "FixMyStreet::DB::Result::User",
+  sub {
+    my $args = shift;
+    return {
```
dracos, line 245: This could be replaced by something like `$args->{self_alias}.extra->'contributed_by' = $args->{foreign_alias}.id` if column was JSONB. I think.

chrismytton (approved):
🛼

PR #4089, opened 2022-09-01
[UK] Fix crash on /admin/config for cobrands that don’t define report_sent_confirmation_email

davea:
e.g. Peterborough

[skip changelog]


dracos (changes requested):
Good spot - as admin/config is the only usage without $report, could we return something that will be displayed to make it clearer - perhaps "report dependent" or "id if waste"?
A similar line also needs adding to Buckinghamshire.pm looks like.

davea:
That makes sense, ta.

dracos (approved):

PR #4090, opened 2022-09-01
[Bulky Goods] Add JSON editor in admin for configuring WasteWorks parameters

davea:
This is a very basic JSON-in-a-text-area form (there is some validation at least) and will be turned into a fully-featured and more user friendly editor later on.

For https://github.com/mysociety/societyworks/issues/2916

image

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
@@ -1,0 +1,116 @@
+package FixMyStreet::App::Controller::Admin::Waste;
+
+use JSON::MaybeXS;
+use Moose;
+use Try::Tiny;
+use namespace::autoclean;
+
+BEGIN { extends 'Catalyst::Controller'; }
+
+=head1 NAME
+
+FixMyStreet::App::Controller::Admin::Waste - Catalyst Controller
+
+=head1 DESCRIPTION
+
+Admin pages for configuring WasteWorks parameters
+
+=head1 METHODS
+
+=cut
+
+sub index : Path : Args(0) {
+    my ( $self, $c ) = @_;
+
+    my $user = $c->user;
+
+    if ($user->is_superuser) {
+        $c->forward('fetch_wasteworks_bodies');
+    } elsif ( $user->from_body ) {
+        $c->forward('load_wasteworks_body', [ $user->from_body->id ]);
+        $c->res->redirect( $c->uri_for_action( 'admin/waste/edit', $c->stash->{body}->id ) );
+    } else {
+        $c->detach( '/page_error_404_not_found', [] );
+    }
+}
+
+sub edit : Path : Args(1) {
+    my ( $self, $c, $body_id ) = @_;
+
+    foreach (qw(status_message)) {
+        $c->stash->{$_} = $c->flash->{$_} if $c->flash->{$_};
+    }
+
+    $c->forward('load_wasteworks_body', [ $body_id ]);
+    $c->forward('stash_body_config_json');
+    $c->forward('/auth/get_csrf_token');
+
+    if ($c->req->method eq 'POST') {
+        $c->forward('/auth/check_csrf_token');
+
+        my $new_cfg;
+        try {
+            $new_cfg = JSON->new->utf8(1)->decode($c->get_param("body_config"));
+        } catch {
+            $c->stash->{errors} ||= {};
+            my $e = $_;
+            $e =~ s/ at \/.*$//; # trim the filename/lineno
+            $c->stash->{errors}->{body_config} = sprintf(_("Not a valid JSON string: %s"), $e);
+            $c->detach;
+        };
+        if (ref $new_cfg ne 'HASH') {
+            $c->stash->{errors} ||= {};
+            $c->stash->{errors}->{body_config} = _("Config must be a JSON object literal, not array.");
+            $c->detach;
+        }
+        $c->stash->{body}->set_extra_metadata("wasteworks_config", $new_cfg);
+        $c->stash->{body}->update;
+        $c->flash->{status_message} = _("Updated!");
+        $c->res->redirect( $c->uri_for_action( '/admin/waste/edit', $c->stash->{body}->id ) );
+    }
+}
+
+sub fetch_wasteworks_bodies : Private {
+    my ( $self, $c ) = @_;
+
+    my @bodies = $c->model('DB::Body')->search(undef, {
+        columns => [ "id", "name", "extra" ],
+    })->active;
+
+    @bodies = grep {
+        $_->get_cobrand_handler &&
+        $_->get_cobrand_handler->feature('waste_features') &&
+        $_->get_cobrand_handler->feature('waste_features')->{admin_config_enabled}
+    } @bodies;
+    $c->stash->{bodies} = \@bodies;
+}
+
+sub stash_body_config_json : Private {
+    my ($self, $c) = @_;
+
+    if ( my $new_cfg = $c->get_param("body_config") ) {
+        $c->stash->{body_config_json} = $new_cfg;
+    } else {
+        my $cfg = $c->stash->{body}->get_extra_metadata("wasteworks_config", {});
+        $c->stash->{body_config_json} = JSON->new->utf8(1)->pretty->canonical->encode($cfg);
+    }
+}
+
+sub load_wasteworks_body : Private {
+    my ($self, $c, $body_id) = @_;
+
+    unless ( $c->user->has_body_permission_to('wasteworks_config', $body_id) ) {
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -0,0 +1,112 @@
...
+    foreach (qw(status_message)) {
+        $c->stash->{$_} = $c->flash->{$_} if $c->flash->{$_};
+    }
+
+    $c->forward('load_wasteworks_body', [ $body_id ]);
+    $c->forward('stash_body_config_json');
+    $c->forward('/auth/get_csrf_token');
+
+    if ($c->req->method eq 'POST') {
+        $c->forward('/auth/check_csrf_token');
```
nephila-nacrea, line 49: Hasn't this already happened above, or do you always need to do it again for POSTs?

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -0,0 +1,112 @@
...
+
+        my $new_cfg;
+        try {
+            $new_cfg = JSON->new->utf8(1)->allow_nonref->decode($c->get_param("body_config"));
+        } catch {
+            $c->stash->{errors} ||= {};
+            my $e = $_;
+            $e =~ s/ at \/.*$//; # trim the filename/lineno
+            $c->stash->{errors}->{body_config} = sprintf(_("Not a valid JSON string: %s"), $e);
```
nephila-nacrea, line 58: If I recall correctly, the _() method is related to translation/localisation...?

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -0,0 +1,112 @@
...
+}
+
+sub fetch_wasteworks_bodies : Private {
+    my ( $self, $c ) = @_;
+
+    my @bodies = $c->model('DB::Body')->search(undef, {
+        columns => [ "id", "name", "extra" ],
+    })->active;
+
+    @bodies = grep { $_->get_cobrand_handler && $_->get_cobrand_handler->feature('waste_features') && $_->get_cobrand_handler->feature('waste_features')->{admin_config_enabled} } @bodies;
```
nephila-nacrea, line 80: I wonder if it would be better to tidy long lines like these onto multiple lines.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -728,6 +728,9 @@ sub admin_pages {
     if ( $user->has_body_permission_to('emergency_message_edit') ) {
         $pages->{emergencymessage} = [ _('Emergency message'), 12 ];
     }
+    if ( $user->has_body_permission_to('wasteworks_config') ) {
+        $pages->{waste} = [ _('WasteWorks config'), 14];
```
nephila-nacrea, line 732: What does 14 refer to?

t/app/controller/waste_peterborough.t
```diff
@@ -701,6 +703,81 @@ FixMyStreet::override_config {
...
+                admin_config_enabled => 1
+            }
+        }
+    },
+}, sub {
+
+    subtest 'WasteWorks configuration editing' => sub {
+    ok $mech->host('peterborough.fixmystreet.com');
+
+    subtest 'Only cobrands with feature enabled are visible' => sub {
```
nephila-nacrea, line 727: Is this a subtest inside the above subtest? The indentation seems misleading.

t/app/controller/waste_peterborough.t
```diff
@@ -701,6 +703,81 @@ FixMyStreet::override_config {
...
+
+        $body->discard_changes;
+        is $body->get_extra_metadata('wasteworks_config'), undef;
+    };
+
+    subtest 'Submitting valid JSON but not an object shows error' => sub {
+        is $body->get_extra_metadata('wasteworks_config'), undef;
+
+        $mech->submit_form_ok({ with_fields => { body_config => '1234' } });
+        $mech->content_contains("Config must be a JSON object literal, not array.");
```
nephila-nacrea, line 760: What is the definition of an object literal? Also, you have submitted an integer/string, but the error complains about an array.

web/cobrands/sass/_admin.scss
```diff
@@ -323,3 +323,8 @@ $button_bg_col: #a1a1a1;  // also search bar (tables)
     max-height: initial;
   }
 }
+
+textarea.code {
+    font-family: monospace;
+    max-width: initial;
+}
```
nephila-nacrea, line 330: Github has a warning(?) about no newline at the end of the file - is this something we are concerned about?

nephila-nacrea (commented):
Mostly questions rather than spotting actual mistakes 🙂

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -0,0 +1,112 @@
...
+
+    my $user = $c->user;
+
+    if ($user->is_superuser) {
+        $c->forward('fetch_wasteworks_bodies');
+    } elsif ( $user->from_body ) {
+        $c->forward('load_wasteworks_body', [ $user->from_body->id ]);
+        $c->res->redirect( $c->uri_for_action( 'admin/waste/edit', $c->stash->{body}->id ) );
+    } else {
+        $c->detach( '/page_error_404_not_found', [] );
```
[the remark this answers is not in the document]
davea: Nope, not needed!

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
davea: Above we call `get_csrf_token` which sets `csrf_time` and `csrf_token` in the stash. This call to `check_csrf_token` is only needed on POSTs because it compares the value that was sent by the user with the expected value and errors if they don't match.

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
davea: It is indeed. This part of the codebase is technically not Peterborough or even UK specific so it doesn't hurt to get it ready for potential future translation.

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
davea: Oops yes, I was having a bit of a moment and couldn't understand why putting a `\` at the end of each line was a syntax error. But it's not Python, is it :)

perllib/FixMyStreet/Cobrand/Default.pm
davea: The admin nav items are ordered by that value, so this puts it at the end:
image

t/app/controller/waste_peterborough.t
davea: It's one of these `{}`, as opposed to an array `[]` or a string `""` or number `1`. JSON documents have to be arrays or objects (Although I'm glad you spotted this because I'd actually configured the `JSON` object on L53 of `Admin::Waste` incorrectly and it was letting anything through when decoding as long as the syntax was OK) so if we reach this point and it's not an object it must be an array. I've rejigged these tests to describe the correct behaviour.

web/cobrands/sass/_admin.scss
davea: Yep, always good to get rid of warnings! Have fixed.

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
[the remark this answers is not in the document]
dracos: Actually, these can be needed (and so is best to leave them in), because if not present, the arguments provided to the main controller are passed through. For example, a call to /edit/123456 where 123456 does not exist, will end up with a 404 page showing "123456" as the error message, because that is the argument of the main controller.

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+sub edit : Path : Args(1) {
+    my ( $self, $c, $body_id ) = @_;
+
+    foreach (qw(status_message)) {
```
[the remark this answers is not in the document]
dracos: They are synonyms, yes, but I think I prefer to use `for` for old-style C loops, rather than confusing the two types of loop together. The FMS codebase has 13 uses of `for` C-loop style, 17 uses of `for` list-loop style, and 281 uses of `foreach` so foreach would be more consistent with majority usage, at least, but no biggie.

nephila-nacrea (approved):
👍

PR #4091, opened 2022-09-02
[Northamptonshire] Switch assets from Alloy to hosted shapefiles

chrismytton:
Northamptonshire are switching from Alloy to hosted shapefiles for their assets.

This change modifies the JavaScript to (hopefully!) have the same behaviour as before, but gets the assets from shapefiles hosted on tilma, instead of from Alloy.

Part of https://github.com/mysociety/societyworks/issues/3172

[skip changelog]


web/cobrands/fixmystreet/assets.js
```diff
@@ -24,6 +24,8 @@ OpenLayers.Layer.VectorBase = OpenLayers.Class(OpenLayers.Layer.Vector, {
         $(fixmystreet).on('assets:unselected', this.checkSelected.bind(this));
         $(fixmystreet).on('report_new:category_change', this.update_layer_visibility.bind(this));
         $(fixmystreet).on('inspect_form:asset_change', this.update_layer_visibility.bind(this));
+        $(fixmystreet).on('maps:update_pin', this.checkSelected.bind(this));
+        $(fixmystreet).on('report_new:category_change', this.checkSelected.bind(this));
```
dracos, line 28: Whilst there might well be an argument that we should do this, I think for a commit that's just moving like this we should copy Bucks/Peterborough/TfL and create a NorthamptonshireVectorAsset class that adds these two event handlers. I'm assuming there is a reason why they're not in the Base, though perhaps there isn't, but it probably needs more checking than just for this.

web/cobrands/northamptonshire/assets.js
```diff
@@ -209,19 +183,19 @@ var northants_road_defaults = $.extend(true, {}, fixmystreet.alloyv2_defaults, {
         field: 'asset_resource_id'
     },
     getUSRN: function(feature) {
-      return feature.fid;
+      return feature.attributes.asset_id;
```
dracos, line 186: I think you can remove the getUSRN function, and add an `attribute: "asset_id"` in the `usrn:` key just above, now it's a standard way of doing it.

web/cobrands/northamptonshire/assets.js
```diff
@@ -231,29 +205,6 @@ fixmystreet.assets.add(northants_road_defaults, {
     ]
 });
 
-var barrier_style = new OpenLayers.Style({
```
dracos, line 234: Why are we removing the barrier and the footpath/bridleway/etc styling? For the latter, is that data not available in the new layer? But for the former, it's the whole layer, looks like?

web/cobrands/northamptonshire/assets.js
```diff
@@ -55,61 +43,74 @@ var layers = [
...
+    "Damaged Control box",
+    "Signal Failure/Damaged - Toucan/Pelican"
+  ],
+  "item_name": "signal or crossing",
+  "layer_name": "Traffic_Signal_Junction"
 },
+{
+  "categories": [ "Pedestrian Barriers - Damaged / Missing" ],
+  "item_name": "pedestrian barrier",
+  "layer_name": "Pedestrian_Barrier"
```
dracos, line 90: This has lost the asset_type area.

web/cobrands/northamptonshire/assets.js
```diff
@@ -21,32 +15,26 @@ var layers = [
...
 {
   "categories": [ "Grit Bin - damaged/replacement", "Grit Bin - empty/refill" ],
   "item_name": "grit bin",
-  "layer_name": "Grit Bins",
-  "styleid": '5d480942fe2ad809d85a78ad',
+  "layer_name": "GritBins"
 },
 {
   "categories": [ "Highway Bridges - Damaged/Unsafe" ],
-  "asset_type": 'area',
```
dracos, line 35: This could be left in, and then included in your loop if set to be set - is used just in some classes in the main code to not show a 'spot'.

web/cobrands/northamptonshire/assets.js
```diff
@@ -55,61 +43,74 @@ var layers = [
...
-var signal_asset_layer = 'layers_nETCOM_5d483dd7fe2ad809d85a8fab';
-
 // default options for northants assets include
 // a) checking for multiple assets in same location
 // b) preventing submission unless an asset is selected
-var northants_defaults = $.extend(true, {}, fixmystreet.alloyv2_defaults, {
-  class: OpenLayers.Layer.AlloyVectorAsset,
-  protocol_class: OpenLayers.Protocol.AlloyV2,
+var northants_defaults = {
   http_options: {
```
dracos, line 98: As this is changing, and these are all standard WFS layers, we might as well use the slightly nicer code for WFS layers - so instead of http_options object, replace it entirely with wfs_url pointing at the same URL (none of params needed), and then in each assets.add, instead of http_options you only need wfs_feature: typename

dracos (changes requested):
A few small suggestions, and a couple of things seem to have gone missing.

web/cobrands/fixmystreet/assets.js
chrismytton: Yeah I was a bit unsure about this, but was needed to get the Cypress tests passing.

I'll move it into a `NorthamptonshireVectorAsset` as you suggest for now.

web/cobrands/northamptonshire/assets.js
chrismytton: The barriers are now spot assets, so this style is no longer relevant as it's designed for line assets.

As for the footpath/bridleway/etc styling, the data isn't available in the new layer, yep.

web/cobrands/northamptonshire/assets.js
chrismytton: As above, the pedestrian barriers are now spots rather than area assets.

web/cobrands/northamptonshire/assets.js
chrismytton: Oh and the rest of the barrier layer has been moved into the array of layers at the top of the file, so it gets added as part of the loop.;

web/cobrands/northamptonshire/assets.js
chrismytton: The bridges layer is actually a spot layer (and it seems it was always a spot layer coming from alloy). Does it still need this, even though it's a spot layer?

dracos: Nah, no need if it actually is spot (sorry for not actually checking those two, assumed it'd be correct already!).

chrismytton:
Hmm, wfs_url change seems to have broken everything 😭

dracos (approved):

PR #4092, opened 2022-09-02
Peterborough bulky goods: implement form for item selection

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3245.

Items for selection are fetched from the database (admin can provide item list via `/admin/waste/<pbro_body_id>`); user can  select up to 5 items (one item per dropdown), as well as optionally upload one image per item.

This PR does not handle saving booking details to the database - see https://github.com/mysociety/fixmystreet/pull/4127.

Screenshot 2022-10-18 at 11 28 21

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -5,6 +5,8 @@ use DateTime::Format::Strptime;
 use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Waste';
 
+use constant MAX_ITEMS => 5;
+
 has_page intro => (
     title => 'Book bulky goods collection',
     intro => 'bulky/intro.html',
@@ -55,9 +57,10 @@ has_page choose_date_later => (
 );
 
 has_page add_items => (
-    fields => ['continue', 'item1', 'item2', 'item3', 'item4', 'item5'], # XXX build list dynamically
-    title => 'Add items for collection',
-    next => 'location',
+    fields   => [ 'continue', map {"item_$_"} ( 1 .. MAX_ITEMS ) ],
+    template => 'waste/bulky/items.html',
+    title    => 'Add items for collection',
+    next     => 'location',
 );
 
 has_page location => (
@@ -65,7 +68,6 @@ has_page location => (
     fields => ['location', 'location_photo', 'location_photo_fileid', 'continue'],
     next => 'summary',
 );
-
 
 has_page summary => (
     fields => ['submit', 'tandc'],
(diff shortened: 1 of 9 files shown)


templates/web/base/waste/bulky/items.html
```diff
@@ -9,7 +9,7 @@
   [% INCLUDE 'waste/_address_display.html' %]
 [% END %]
 
-<form class="waste" method="post">
+<form class="waste" method="post" enctype="multipart/form-data">
```
nephila-nacrea, line 12: I tried adding this in the form class (as in the example in https://metacpan.org/release/GSHANK/HTML-FormHandler-0.40068/view/lib/HTML/FormHandler/Field/Upload.pm), but it didn't seem to do anything, so if this is the solution I will be simultaneously pleased and annoyed it was this simple all along.

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -110,6 +115,15 @@ has_field resident => (
     ],
 );
 
+has_field tandc => (
```
nephila-nacrea, line 118: Did this get removed in a previous PR? I do not know why it is showing here as a new addition.

templates/web/base/govuk/fields.html
```diff
@@ -334,3 +336,9 @@ <h2 class="govuk-fieldset__heading">
   <input class="govuk-file-upload" id="[% field.id %]" name="[% field.id %]" type="file">
   </div>
 [% END %]
+
+[% BLOCK compound %]
```
nephila-nacrea, line 340: I may not actually need this change given how I render the form elements individually now.

templates/web/base/waste/bulky/items.html
davea: I can't see how they render that as HTML in the example, but note in our HTML the `<form>` element is baked into the template and is independent of the `FormHandler` instance so that attribute won't have any effect.

nephila-nacrea (commented):
I am not sure why, but Item 1 gets a (hashed?) filename displayed below the photo:

Screenshot 2022-10-26 at 23 15 32

Dumping the data I get 
```
'item_1' => {
    'item' => 'Bed',
    'photo_fileid' => '0ad35b4497f229eacae69efb0f1e154a14786c6e.jpeg'
},
'item_1.photo_fileid' => '',
'item_1.photo' => '0ad35b4497f229eacae69efb0f1e154a14786c6e.jpeg',
```
whereas e.g. if I provide a photo for Item 2, it gets an 'item_2.photo' of undef.

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -5,6 +5,8 @@ use DateTime::Format::Strptime;
 use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Waste';
 
+use constant MAX_ITEMS => 5;
```
davea, line 8: Please add an `XXX` comment to indicate that this needs to be un-hardcoded and fetched from config later on.

davea (changes requested):
This could do with some squashing of the commits - e.g. removing the fixup, squashing 805c79eebc59053bb3ff0fab50fb3f8f3addd35e, etc.

PR #4098, opened 2022-09-12
Pboro bulky frontend step 1

lucascumsille:
@davea as we spoke I have added comments for the display logic of each element. Please let me know if there is something you don't understand.
**Important, at the moment the "Bulky waste" section has been hard coded, so if it's merged into master will be displayed in other cobrands.**

This should fix:
https://github.com/mysociety/societyworks/issues/2910
https://github.com/mysociety/societyworks/issues/2903

### Mobile view
It includes a Bulky Waste CTA.

https://user-images.githubusercontent.com/13790153/189643164-e5321b96-2f0f-487c-b1d7-7bd5fce0251b.mov

### Desktop (Between 720px and 1500px)

https://user-images.githubusercontent.com/13790153/189643322-386418eb-5f9b-45ef-83db-185cd0bed769.mov

### Desktop (Above 1500px)

https://user-images.githubusercontent.com/13790153/189643252-45aa8277-2a7d-4097-93bd-53e86fd011e2.mov

Let me know if you have any feedback =)


davea:
Looking good! Once you've done the Sass changes I'll put the conditionals in place for the various different ways data is displayed.

OOI why does the nav jump from the right to the left when the viewport is >1500px wide?

lucascumsille:
@davea Thanks for the feedback. I have moved the changes so they are applied when the body has  the `bulky` class.

I have one question regarding the "add to your calendar" element. The way I have it on `bin_days_bulky` is under the title, but I assume we won't be using that template. Do you have any idea how we can keep it there without messing the other cobrands? Maybe one idea is to move it inside `_service_navigation_bar` template.

davea:
I put some logic in the templates to only show calendar link in the new place, use new button styles etc when using the bulky goods feature. As a result this PR should have no effect at all on the appearance of the page unless the bulky feature is enabled.

PR #4101, opened 2022-09-12
[Cheshire East] Tweaks to abuse form

dracos:
Fixes FD-2000. [skip changelog]


MorayMySoc (approved):
I put a query on the ticket related to scope, but this is fine

PR #4109, opened 2022-09-15
[TfL] Remove contact information from report submission emails

davea:
For FD-2405

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -467,9 +467,8 @@ around 'munge_sendreport_params' => sub {
 
     $self->$orig($row, $h, $params);
 
-    if ($row->category eq "Countdown - not working") {
-        $params->{From} = [ $self->do_not_reply_email, $self->contact_name ];
-    }
+    $params->{From} = [ $self->do_not_reply_email, $self->contact_name ];
+    delete $params->{'Reply-To'} if $params->{'Reply-To'};
 };
 
 sub is_hardcoded_category {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -467,9 +467,7 @@ around 'munge_sendreport_params' => sub {
 
     $self->$orig($row, $h, $params);
 
-    if ($row->category eq "Countdown - not working") {
-        $params->{From} = [ $self->do_not_reply_email, $self->contact_name ];
-    }
+    $params->{From} = [ $self->do_not_reply_email, $self->contact_name ];
```
dracos, line 470: always_use_reply_to is set for all UK cobrands, so I think the Reply-To header would also need removing, as that contains the user's email.

t/cobrand/tfl.t
```diff
@@ -417,12 +412,14 @@ subtest "extra information included in email" => sub {
     $mech->clear_emails_ok;
     FixMyStreet::Script::Reports::send();
     my @email = $mech->get_email;
-    like $email[0]->header('From'), qr/"Joe Bloggs" <fms-report-$id-/;
+    is $email[1]->header('From'), '"TfL Streetcare" <fms-tfl-DO-NOT-REPLY@example.com>';
```
dracos, line 415: This is now same check as line 424, should it still be checking 0, not 1?

dracos (changes requested):
Couple of issues, have left note on FD as to whether this is actually wanted for all, or just the category that it was already working for and just that wasn't working properly.

t/cobrand/tfl.t
davea: Oops! Ta.

davea:
Thanks for spotting those things. Client has confirmed contact data should be removed everywhere.

dracos (approved):
Thanks for checking!

dracos:
Just going through old PRs, worth chasing this one up, so can get it live? Not sure why they didn't ever come back on it, but I assume they still want it

PR #4110, opened 2022-09-15
[Lincolnshire] Pull in user information via Open311 so customers can receive updates

chrismytton:
Lincolnshire have requested that reports pulled in via Open311's
GetServiceRequests are created using the user information stored in
their backend system.

This is so that the reports created in FixMyStreet are associated with
the email address of the original reporter so they can receive updates
on reports.

Open311 adapter change to add the user data is here: https://github.com/mysociety/open311-adapter/pull/228

Part of https://github.com/mysociety/societyworks/issues/3079

[skip changelog]


MorayMySoc (approved):
Looks good to me

davea:
@chris I've added a safety check in 4ebadfcaf65f25f094b9a89edba7c3ace5b58791 to limit when we store contact information on staging sites. Can you give it a look please?

dracos:
The code in this PR doesn't actually cause any logged email to be sent, as far as I can see, or for the pulled in user to be signed up for update alerts.

davea (approved):

PR #4111, opened 2022-09-16
[PCC] Added reusable UI elements for Bulky waste request form

lucascumsille:
This PR includes elements that improves the user journey:

- Added secondary button, which has less emphasis than the primary one. It's good to use it next to the primary one so it 
- Added a secondary button with less emphasis than the primary one. It's good to use it next to the primary one so it doesn't draw as much attention, but it's still fairly obvious is there. You can also use it on its own; when you want to show there is an action available but is not the most important part of the page.
- Added "Next" and "Back" buttons, which help with the page's scannability.
- Added an info message based on the GOVUK warning message. It can be used to provide non-vital information to the user, so if they decide to omit it, there won't be a big consequence. For example: "If you are a landlord, call this number". For instance, when NOT to use it, "You need to dismantle this item before collection". In this case, if the user ignores this message, there will be a consequence, the item won't get picked up.

Fixes: https://github.com/mysociety/societyworks/issues/3231

### DIFF
--- a/web/cobrands/peterborough/base.scss
+++ b/web/cobrands/peterborough/base.scss
@@ -37,6 +37,105 @@ h1, h2 {
     color: $alt-green;
 }
 
+.btn-primary,
+.green-btn,
+.btn--primary, .btn-secondary {
+    border: 0.2em solid $primary;
+    padding: 0.55em 1em;
+    border-radius: 0.25em;
+    text-decoration: none !important;
+    font-weight: bold;
+    line-height: 100%;
+    font-size: 1em;
+
+    &:focus {
+        box-shadow: #4c9aff 0px 0px 0px 3px;
+        outline: none;
+    }
+}
+
+@mixin next-back-button($orientation) {
+    padding-#{$orientation}: 2em;
+    background-size: 9px auto;
+    background-position: center $orientation 1em;
+    background-repeat: no-repeat;
+
+    &.is--disabled, &:disabled {
+        cursor: not-allowed;
+        opacity: 0.5;
+        text-decoration: none;
+        pointer-events: none
+    }
+}
+
+.btn-primary,
+.green-btn,
+.btn--primary {
+    background: $primary;
+
+    &:hover,
+    &:active {
+        background: darken($green, 5%);
+        border-color: darken($green, 5%);
+    }
+
+    &:focus {
+        color: $primary_text;
+        background: $primary;
+    }
+
+    &.is--next {
+        @include svg-background-image('/cobrands/peterborough/images/chevron-next-white');
+        @include next-back-button(right);
+    }
+
+    &.is--back {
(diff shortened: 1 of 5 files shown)


templates/web/peterborough/waste/landlord_info.html
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-warning-text due">
+    <div class="govuk-warning-text__img">
+        <span class="govuk-warning-text__icon" aria-hidden="true">i</span>
+    </div>
+    <div class="govuk-warning-text__content">
+        <span class="govuk-warning-text__assistive">Important information</span>
+        <p class="govuk-!-margin-bottom-0">If you are a Landlord or property manager please contact 01733 747474 to request a quote.</p>
+        <!--
```
nephila-nacrea, line 8: Does this commented section need to be left in?

web/cobrands/peterborough/base.scss
```diff
@@ -37,6 +37,91 @@ h1, h2 {
     color: $alt-green;
 }
 
+.btn-primary,
```
nephila-nacrea, line 40: Are these changes going to affect existing Peterborough / WasteWorks pages, and if so, do we want that?

lucascumsille: I think we should do it, for the following reasons:

- I added a fix for the hover and active state of the primary button. Basically, the text wasn't accessible with the lighter background. So now we have the same effect the PCC council has on their website. Same with the focus effect.
- The bulky waste form is a bit more complex and involves more steps if we compare it with FMS and the current journeys on WW. These new elements will impact the UX journey, like highlighting what we want to highlight and giving more priority to the sections we want to emphasise. Are the 100% essential? No, but they are an improvement.
- I tested this PR on FMS and didn't see style conflict.

templates/web/peterborough/waste/landlord_info.html
lucascumsille: I left it in there just in case we want to re-use it with a button, instead of just text. But we can get rid of it we won't use it.

web/cobrands/peterborough/_colours.scss
```diff
@@ -52,3 +52,18 @@ $search-help-margin-desktop: -2em -1em 0 -1em;
 $search-help-header-font-size-desktop: 1.25em;
 
 $geolocation-link: $green;
+
+// Mixins
```
davea, line 56: This mixin is only used in `base.scss` so can it be moved there?

lucascumsille:
@davea thanks again =) I just made both changes. Let me know if there is anything else you'd like me to change.

davea (approved):
✅

lucascumsille:
Thanks @davea should I merge it? or wait for Victoria?

nephila-nacrea:
> Thanks @davea should I merge it? or wait for Victoria?

Should be OK to go ahead!

PR #4113, opened 2022-09-20
[Waste] More verbose logging output

dracos:
[skip changelog]
Fixes https://github.com/mysociety/societyworks/issues/2695


perllib/FixMyStreet/Roles/DDProcessor.pm
```diff
@@ -181,6 +183,7 @@ sub waste_reconcile_direct_debits {
                         # if we've pulled out more that one record, e.g. because they
                         # failed to make a payment then skip remaining ones.
                         $cur->state('hidden');
+                        $cur->add_to_comments( { text => 'Hiding report as handled elsewhere', user => $user, problem_state => $cur->state } );
```
struan, line 186: I wonder if we set handled to the report id of the report that handles it, we can then include that in the debug and this message to make it a bit easier to see what's going on?

struan (approved):
one small suggestion but looks good.

Issue #1580, opened 2016-11-23
Consider using S3 for photo storage


davea:
The full fms.com photo backups take a long time and use a lot of disk space. Is it worth looking at storing & serving report/update photos in S3 directly?

davea:
Recent changes to the backup system means that photos are now backed up directly from /data/vhost/www.fixmystreet.com/photos on mammoth to S3, instead of being copied between our internal servers and onwards via duplicity.

This ticket is still relevant because it'll make scaling FMS across multiple app servers a lot easier if there's no state on the app server's disk.

dracos:
Note photos is already a mount from another server :)

dracos:
This was done some time ago, there's an issue with old db entries that lack a file extension.

PR #4117, opened 2022-09-23
[FMS] Script to add extension to images without extension in db

MorayMySoc:
* Adds correct extension to images without extension in db
* Assumes that if one image has extension on a record, all do

https://github.com/mysociety/fixmystreet/issues/1580

[skip changelog]


bin/photos-add-extension
```diff
@@ -0,0 +1,78 @@
...
+    '%c',
+    [ 'commit', "Supply argument to commit changes to the database" ],
+    [ 'help', "Print help text" ]
+);
+
+print($usage->text), exit if $opt->help;
+
+my $config =
+    {
+        id => 1305,
```
davea, line 31: Should probably get rid of this :)

MorayMySoc:
@davea - @dracos suggested that checking every file was unnecessary as they will almost all, to a significant degree, be jpeg. So I've simplified this to just add jpeg to any db entries without it.

PR #4120, opened 2022-09-26
[Buckinghamshire] Use speed limit layer to route all parish categories

davea:
For https://mysocietysupport.freshdesk.com/a/tickets/2429

[skip changelog]

### DIFF
--- a/bin/buckinghamshire/add-parish-categories
+++ b/bin/buckinghamshire/add-parish-categories
@@ -20,9 +20,32 @@ BEGIN {    # set all the paths to the perl code
 use FixMyStreet::DB;
 use Text::CSV;
 
+my $speed_limit_q = {
+    code => 'speed_limit_greater_than_30',
+    description => 'Is the speed limit greater than 30mph?',
+    datatype => 'string',
+    order => 1,
+    variable => 'true',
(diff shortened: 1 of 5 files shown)


bin/buckinghamshire/add-parish-categories
```diff
@@ -21,11 +21,44 @@ use FixMyStreet::DB;
...
 
 my $contacts = [
-    { category => 'Hedge problem', extra_metadata => { prefer_if_multiple => 1, group => 'Grass, hedges and weeds' } },
-    { category => 'Dirty signs', extra_metadata => { prefer_if_multiple => 1, group => 'Road and street Signs' } },
+    {
+        category => 'Hedge problem',
+        extra_metadata => {
+            group => 'Grass, hedges and weeds'
+        },
+        extra_fields => [
```
dracos, line 29: Could the question be a factored-out variable, it's the same each time?

dracos (changes requested):
One suggestion, and there are references to Grass cutting in templates/web/buckinghamshire/report/new/_category_extras_header.html and web/cobrands/buckinghamshire/assets.js that need updating too.

PR #4123, opened 2022-09-29
[Bulky] Add ‘Location details’ page to booking wizard

davea:
Missed this previously!

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -34,6 +34,12 @@ has_page choose_date => (
 has_page add_items => (
     fields => ['continue', 'item1', 'item2', 'item3', 'item4', 'item5'], # XXX build list dynamically
     title => 'Add items for collection',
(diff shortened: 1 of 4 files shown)


lucascumsille:
@davea it looks good to me. I agree that we should make that button pop up more and if we can reuse the drag photo element from FMS that should do it. Is that something we do using the FormHandler?

davea:
Yes, I think we'll need to add the appropriate class to the element in the FormHandler field definition. I'll figure that out as part of https://github.com/mysociety/fixmystreet/pull/4092.

davea:
In the end I did it in this PR - see cd827ccf2a93893d6b63a20376f47bbd90446a2c.

lucascumsille:
Looking good @davea nice addition.
I added a commit so the dropzone doesn't use the whole width. Because of the size + background of element I thought it was bit much. If you checkout this branch you won't be able to see it, because the styling is being applied to the dropzone nested inside `body.bulky` and we haven't add that class to body in this PR.

Screenshot 2022-09-30 at 09 00 39

PR #4128, opened 2022-10-05
[SLWP] Fix admin users sharing permissions.

dracos:
I'd added an admin_allow_user function so they could access each other's admin, but any actual permission checking involving a body ID would fail if a Kingston staff tried to look at a Sutton report or vice-versa. Add a body override so that both bodies are always counted if one is present. For https://3.basecamp.com/4020879/buckets/26876928/todos/5400097144. [skip changelog]


davea (approved):
👍

PR #4130, opened 2022-10-07
[Sutton] Only request one container at a time

dracos:
And no reporting before 6pm on bin day. [skip changelog]


struan (approved):
One tiny test suggestion but otherwise looks good.

PR #4131, opened 2022-10-10
[Waste] Re-enable submit button when browser back button is used

davea:
On Safari, at least, it seems that the code on L2 doesn't get called when the back button is used.

[skip changelog]

### DIFF
--- a/web/js/waste.js
+++ b/web/js/waste.js
@@ -1,5 +1,9 @@
+window.addEventListener("pagehide", function() {
+    $('form.waste input[type="submit"]')
+        .prop('disabled', false)
+        .parents('.govuk-form-group').removeClass('loading');
+});
 $(function() {
-    $('form.waste input[type="submit"]').prop('disabled', false);
     $('form.waste').on('submit', function(e) {
         var $btn = $('input[type="submit"]', this);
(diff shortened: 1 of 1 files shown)


web/js/waste.js
```diff
@@ -1,5 +1,10 @@
 $(function() {
     $('form.waste input[type="submit"]').prop('disabled', false);
+    window.addEventListener("pageshow", function() {
```
dracos, line 3: If this were moved to line 1 (outside the $( ready setup, as we're just adding a listener), then presumably line 2 wouldn't be needed, because the event would then fire on the page load that is initially running this JS? Worth a test anyway, so it's only needed once if so.

And then... if we don't have a pageshow, but a pagehide that does this - would that be better? That way we know we've "left" the page in an okay state, so we don't need to do anything when we're coming back to it (however we are).

dracos (commented):
Couple of questions :)

web/js/waste.js
davea: Yes, moving this to line 1 still works. And using `pagehide` is a good idea too 👍

dracos (approved):

PR #4135, opened 2022-10-12
Retrieve children for all body areas

chrismytton:
Updates the code that fetches children for a body to look at all associated areas.

This fixes the issue that Northamptonshire reported where they could only see North Northamptonshire wards in the dashboard dropdown.

Fixes https://github.com/mysociety/societyworks/issues/3262

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Default.pm
+++ b/perllib/FixMyStreet/Cobrand/Default.pm
@@ -814,12 +814,21 @@ cobrand's area_types_children type.
 =cut
 
 sub fetch_area_children {
-    my ($self, $area_id, $all_generations) = @_;
+    my ($self, $area_ids, $all_generations) = @_;
 
-    return FixMyStreet::MapIt::call('area/children', $area_id,
-        type => $self->area_types_children,
-        $all_generations ? (min_generation => 1) : (),
-    );
+    $area_ids = [ $area_ids ] unless ref $area_ids eq 'ARRAY';
+
+    my %all_children;
+
+    foreach my $area_id (@$area_ids) {
+        my $children = FixMyStreet::MapIt::call('area/children', $area_id,
+            type => $self->area_types_children,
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -814,12 +814,21 @@ cobrand's area_types_children type.
...
 
-    return FixMyStreet::MapIt::call('area/children', $area_id,
-        type => $self->area_types_children,
-        $all_generations ? (min_generation => 1) : (),
-    );
+    $area_ids = [ $area_ids ] unless ref $area_ids eq 'ARRAY';
+
+    my %all_children;
+
+    foreach my $area_id (@$area_ids) {
```
dracos, line 823: You can call children with an array ref of IDs (e.g. `call('area/children', $area_ids)`) and that should get you all the results you need with one call.

dracos (changes requested):
Nice, this should also fix the list of wards on the All Reports page for Northants too :) One suggestion as to how to only make the one MapIt call is all, I think.

perllib/FixMyStreet/Cobrand/Default.pm
chrismytton: I tried doing that but it wasn't returning any results. Spoke to Dave about it and it seems Mapit doesn't have code for looking up children for multiple areas. Unless I've missed something?

https://mapit.mysociety.org/areas/164185,164186/children?type=DIW,LBW,MTW,UTE,UTW,LGE,CED,COP&generation=42

(Not sure what's actually handling that route)

I did briefly start looking at changing Mapit, but seeing as there aren't many bodies with two (or more) areas this seemed like the simpler solution.

dracos: Oh, I'm just talking nonsense then, sorry!

dracos (approved):

PR #4140, opened 2022-10-18
[National Highways] Ensure reports don't have an update form

chrismytton:
The HighwaysEngland cobrand module doesn't inherit from UKCouncils, so was falling back to the updates_disallowed method from Default.pm. This means that despite there being configuration designed to prevent updates they were still being permitted.

This change moves the updates_disallowed method from UKCouncils to UK.pm.

I think this should be ok, as the two places where it might matter, FixMyStreet.pm and Thamesmead.pm have their own version of the method.

Thamesmead might be a cause for concern because it calls SUPER, which will now be the new method in UK.pm, rather than Default.pm.

The other two cobrands that might be affected are EastSussex and WestBerkshire, but I can't see that it would make a difference to those.

Fixes https://github.com/mysociety/societyworks/issues/3272

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UK.pm
+++ b/perllib/FixMyStreet/Cobrand/UK.pm
@@ -427,6 +427,52 @@ sub report_new_munge_before_insert {
     }
 }
 
+# Allow cobrands to disallow updates on some things.
+# Note this only ever locks down more than the default.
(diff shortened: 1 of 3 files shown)


dracos (approved):
🆙 📆

PR #4141, opened 2022-10-19
Fix all warnings and offline running

dracos:
@struan First commit for you to check on the DD contact/mandate stuff discussed, and in second commit there's the fix for direct_debit_error not having the property details (already fixed for CC).
@davea In the second commit, there's one bulky goods change for you to just check - to make sure the fileid for location_photo is included in the right page, so it's not included on all pages (and thereby cause undef warnings).
Think all the rest is just to prevent warnings.
[skip changelog]


struan (approved):
DD bits look fine and yay for doing the test warnings 🙏

davea (approved):

PR #4144, opened 2022-10-21
[Bulky] Added Jquery for item screen

lucascumsille:
This PR adds the following functionality:

- The "Delete Item" will delete item + reset the item value in case there was something selected.
- "Add item" button will become enabled when all the visible items have a selected item, but will become disabled when the max number of items has been reached or there is a visible item with nothing selected.
- Both buttons will become invisible if JS is not enabled.
- I used @nephila-nacrea[ commit](https://github.com/mysociety/fixmystreet/pull/4142/commits/62a9f9ee0a083145dc9898c73668832da31af870) for to display message. However I tweak her JS + the markup to make it more accessible and noticeable.

Preview:
https://user-images.githubusercontent.com/13790153/197184457-03e2923d-f278-48f0-90af-08550e2f0745.mov

There is one thing tat it would be nice:
At the moment the photo input field is one element that included the button to upload an image plus the file name or "No file selected", so the clickable area is a bit deceitful, because is larger than it appears. Can we break this element into two? button + message. If it's too much hassle we  can just include a max-width or width auto property, not ideal, but not terrible either.

Fixes https://github.com/mysociety/societyworks/issues/3233
Let me know if there is any feedback.

### DIFF
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -29,12 +29,18 @@
   [% PROCESS form override_fields = [ 'continue', 'saved_data', 'token', 'process', 'unique_id' ] %]
 </form>
 
-[% # XXX If JS enabled, hide unless relevant item selected %]
+[% # Only shown if JS disabled %]
 [% BLOCK extra_text %]
-  [% FOR item_desc IN form.items_extra_text.keys %]
-    [% id = 'item_' _ num _ '.' _ form.format_item_string(item_desc) _ '.extra_text' %]
-    <div id="[% id %]"><b>[% item_desc %]: [% form.items_extra_text.$item_desc %]</b></div>
-  [% END %]
+<div class="govuk-warning-text due bulky-item-message">
+  <div class="govuk-warning-text__img">
+    <span class="govuk-warning-text__icon" aria-hidden="true">i</span>
+  </div>
+  <div class="govuk-warning-text__content">
+    <span class="item-name"></span>
+    <span class="govuk-warning-text__assistive">Important information</span>
+    <p id="item-message" class="govuk-!-margin-bottom-0"></p>
+  </div>
+</div>
 [% END %]
 
 [% INCLUDE footer.html %]
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -15,13 +15,14 @@
     [% # Building names beforehand because override_fields does not seem to like them being built inside its arg list %]
     [% base_field = 'item_' _ num %]
     [% item = base_field _ '.item' %]
+    [% PROCESS form override_fields = [ item ] %]
+    [% PROCESS extra_text num = num %]
     [% photo = base_field _ '.photo' %]
     [% photo_fileid = base_field _ '.photo_fileid' %]
-    [% PROCESS form override_fields = [ item, photo, photo_fileid ] %]
+    [% PROCESS form override_fields = [ photo, photo_fileid ] %]
 
     <button type="button" class="delete-item btn-secondary govuk-!-margin-bottom-3">Delete item</button>
 
-    [% PROCESS extra_text num = num %]
     <hr>
   </div>
   [% END %]
@@ -29,7 +30,6 @@
   [% PROCESS form override_fields = [ 'continue', 'saved_data', 'token', 'process', 'unique_id' ] %]
 </form>
 
-[% # Only shown if JS enabled %]
 [% BLOCK extra_text %]
 <div class="govuk-warning-text due bulky-item-message">
   <div class="govuk-warning-text__img">
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -25,11 +25,11 @@
     <hr>
   </div>
   [% END %]
-  <button type="button" id="add-new-item" class="btn-secondary govuk-!-margin-bottom-3">+ Add item</button>
+  <button type="button" id="add-new-item" class="btn-secondary govuk-!-margin-bottom-3" aria-label="Add item">+ Add item</button>
   [% PROCESS form override_fields = [ 'continue', 'saved_data', 'token', 'process', 'unique_id' ] %]
 </form>
 
-[% # Only shown if JS disabled %]
+[% # Only shown if JS enabled %]
 [% BLOCK extra_text %]
 <div class="govuk-warning-text due bulky-item-message">
   <div class="govuk-warning-text__img">
@@ -38,7 +38,7 @@
   <div class="govuk-warning-text__content">
     <span class="item-name"></span>
     <span class="govuk-warning-text__assistive">Important information</span>
-    <p id="item-message" class="govuk-!-margin-bottom-0"></p>
+    <p class="item-message govuk-!-margin-bottom-0" aria-live="polite"></p>
   </div>
 </div>
 [% END %]
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -9,6 +9,18 @@
   [% INCLUDE 'waste/_address_display.html' %]
 [% END %]
 
+<div class="no-js-message govuk-warning-text due" style="max-width:550px">
+  <div class="govuk-warning-text__img">
+    <span class="govuk-warning-text__icon" aria-hidden="true">i</span>
+  </div>
+  <div class="govuk-warning-text__content">
+    <span class="govuk-warning-text__assistive">Important information</span>
+    <p class="govuk-!-margin-bottom-1"><strong>About your items</strong></p>
+    <p >Before continuing, please read the following file and see if any items have a note or comment that might help our crew to collect them.</p>
+    <a class="btn-primary" href="" download="">Item notes</a>
+  </div>
+</div>
+
 <form id="item-selection-form" class="waste" method="post" enctype="multipart/form-data">
   [% FOR num IN [ 1 .. form.MAX_ITEMS ] %]
   <div class="bulky-item-wrapper">
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -17,6 +17,7 @@
     <span class="govuk-warning-text__assistive">Important information</span>
     <p class="govuk-!-margin-bottom-1"><strong>About your items</strong></p>
     <p >Before continuing, please read the following file and see if any items have a note or comment that might help our crew to collect them.</p>
+    <!-- Needs file -->
     <a class="btn-primary" href="" download="">Item notes</a>
   </div>
 </div>
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -9,19 +9,23 @@
   [% INCLUDE 'waste/_address_display.html' %]
 [% END %]
 
-<form class="waste" method="post" enctype="multipart/form-data">
+<form id="item-selection-form" class="waste" method="post" enctype="multipart/form-data">
   [% FOR num IN [ 1 .. form.MAX_ITEMS ] %]
+  <div class="bulky-item-wrapper">
     [% # Building names beforehand because override_fields does not seem to like them being built inside its arg list %]
     [% base_field = 'item_' _ num %]
     [% item = base_field _ '.item' %]
-
     [% photo = base_field _ '.photo' %]
     [% photo_fileid = base_field _ '.photo_fileid' %]
     [% PROCESS form override_fields = [ item, photo, photo_fileid ] %]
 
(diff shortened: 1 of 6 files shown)


davea (commented):
Couple of comments:

When you come back to the page via the "change answers" button on summary screen, the "add item" button is disabled despite there already being some items present:

image


Also the messages for all the item types are shown instead of just one for the selected item:

Screenshot 2022-10-21 at 13 56 17

lucascumsille:
Thanks for the feedback @davea =)
I just added a fix that should prevent both issues. 
Let me know if there is anything else.

nephila-nacrea (changes requested):
Hi,

Had a play around with things after rebasing this branch on top of the most recent `pboro-bulky-build-item-list`, and things don't seem to work 😞 Possibly because some changes were made in `pboro-bulky-build-item-list` to handle multiple image uploads.

With JS enabled:
    - a) All items appear on the page at once
    - b) There is still an add button despite being max of 5
    - c) It is not clear that you can only add another item after you have selected one from the dropdown
    - d) I think the ‘extra text’ should appear above the ‘upload image’, not below

With JS disabled:
    - a) No extra text showing
    - b) Image upload does not work
    - c) ‘Delete item’ button showing for each item (probably shouldn’t be there?)
    - d) ‘Add item’ button showing (probably shouldn’t be there?)

Again, I am not sure what happened because I assume things were working nicely before the rebase 😞 I wasn't getting any conflicts in the JS files so I had assumed everything would work fine.

lucascumsille:
Thanks for the feedback @nephila-nacrea =)
Not sure why it isn't working for you.  This reminds me of a similar issue I had with Moray and Chris. I'll check with him, but if my memory doesn't betray is because the css and javascript files they were using hadn't been updated. I wonder if is the same issue.

### With JS
- [x] All items appear on the page at once
- [x] There is still an add button despite being max of 5
- [x ] It is not clear that you can only add another item after you have selected one from the dropdown
Because the add item is at the end and not at the beginning seems natural that the first thing the user will do will be to select an item. So I added a Fix for screen readers so when the user adds an item, the focus jumps into the new search autocomplete.
- [ ] I think the 'extra text' should appear above the 'upload image', not below
I agree. Unfortunately, I couldn't change it myself, I tried to move the whole block, but it didn't work. I'm guessing this line is the issue `[% PROCESS form override_fields = [ item, photo, photo_fileid ] %]`. Let me know how I can make the change.

Preview
https://user-images.githubusercontent.com/13790153/199664577-e74b279e-f320-42c3-841f-67e92049266b.mov


### With JS disabled:
- a) No extra text showing
Good point; this is a tricky one. Because if I'm not mistaken, the messages for each item could be different. So one option could be to attach a pdf with the list of items + notes or comments on it.
Other option could be to include * to the items that need to be dismantled, but this is limited because we don't know how much the messages can vary from one to another. In the meantime, I have hidden the message.
Amy other solutions are welcome.

- b) Image upload does not work
It's working for me.
- c) 'Delete item' button showing for each item (probably shouldn't be there?)
Not for me
- d) 'Add item' button showing (probably shouldn't be there?)
Not for me

https://user-images.githubusercontent.com/13790153/199665526-5e190367-d557-4e40-aeea-270127aeacad.mov

Thanks again for the feedback =)

lucascumsille:
@nephila-nacrea I just spoke with Moray. Assuming is the same issue
run `script/setup` to compile all in the FixMyStreet directory.

nephila-nacrea:
Hi,

Running `script/setup` fixed a bunch of things, thanks 👍 

I still noticed a few things, however:
- JS enabled:
    - Add (for example) 3 items, delete the 2nd - it is deleted in the JS, but click through to the summary page and 3 items are there
    - At least some of the time, ‘Add item’ is enabled when there are 5 items; the button only becomes disabled when it is clicked (nothing else happens)

I will push a fix for the position of the extra text 🙂 

- JS disabled:
    - No extra text showing (which you already know)
    - Image upload does not work - i.e. there is no photo displayed on the summary page

With `script/setup` the add & delete buttons no longer show up 👍

lucascumsille:
Thanks for the feedback =) and moving the message.

### JS Enabled

- Playing around I found another issues. When uploading a picture and then deleting the item, the picture would still be there. I just pushed a commit that solves that.
- Add (for example) 3 items, delete the 2nd - it is deleted in the JS, but click through to the summary page and 3 items are there. [Solved I pushed another commit]

> At least some of the time, ‘Add item’ is enabled when there are 5 items; the button only becomes disabled when it is clicked (nothing else happens)

I tried to replicate, but it didn't happened to me. Anyway I can replicate this?

### JS disabled:

- In my summary page I can see there is a file being upload, not the preview, but the name of the file. Personally I don't think is really worth it to spend much time on the JS disable, ad long the file is being uploaded.
- Regarding the message, if there isn't a simpler way I would attach  pdf with all the comments for each single item and ask the user check the file before collection.

Let me know what you think

nephila-nacrea:
> Thanks for the feedback =) and moving the message.
> 
> ### JS Enabled
> * Playing around I found another issues. When uploading a picture and then deleting the item, the picture would still be there. I just pushed a commit that solves that.
> * Add (for example) 3 items, delete the 2nd - it is deleted in the JS, but click through to the summary page and 3 items are there. [Solved I pushed another commit]
> 

This all seems fine now 👍 

> > At least some of the time, ‘Add item’ is enabled when there are 5 items; the button only becomes disabled when it is clicked (nothing else happens)
> 
> I tried to replicate, but it didn't happened to me. Anyway I can replicate this?

Hmm I wasn't able to replicate this again 🤔 So I wouldn't worry about it; if it happens again it is a very minor bug.

> ### JS disabled:
> * In my summary page I can see there is a file being upload, not the preview, but the name of the file. Personally I don't think is really worth it to spend much time on the JS disable, ad long the file is being uploaded.

There is currently a bug where a filename gets displayed for Item 1, but none of the others. I also think images need to show up on the summary page whether JS is enabled or not (like we do for images on reports). But we may be better off moving this fix to another ticket (e.g. a general 'make sure things work for disabled JS' ticket? That could be worked on once most work out of the way.)

> * Regarding the message, if there isn't a simpler way I would attach  pdf with all the comments for each single item and ask the user check the file before collection.
> 
> Let me know what you think

I think this could work; it would be such a small number of users seeing this that it doesn't need to be amazing 😛

nephila-nacrea:
@lucascumsille Have made a ticket (https://github.com/mysociety/societyworks/issues/3328) for capturing miscellaneous fixes like the JS-disabled behaviour.

nephila-nacrea (changes requested):
Hi, I left a reply some days ago but you may not have been notified because it wasn't part of a 'review', so I'll paste it here:

---
> Thanks for the feedback =) and moving the message.
> 
> ### JS Enabled
> * Playing around I found another issues. When uploading a picture and then deleting the item, the picture would still be there. I just pushed a commit that solves that.
> * Add (for example) 3 items, delete the 2nd - it is deleted in the JS, but click through to the summary page and 3 items are there. [Solved I pushed another commit]
> 

This all seems fine now 👍 

> > At least some of the time, ‘Add item’ is enabled when there are 5 items; the button only becomes disabled when it is clicked (nothing else happens)
> 
> I tried to replicate, but it didn't happened to me. Anyway I can replicate this?

Hmm I wasn't able to replicate this again 🤔 So I wouldn't worry about it; if it happens again it is a very minor bug.

> ### JS disabled:
> * In my summary page I can see there is a file being upload, not the preview, but the name of the file. Personally I don't think is really worth it to spend much time on the JS disable, ad long the file is being uploaded.

There is currently a bug where a filename gets displayed for Item 1, but none of the others. I also think images need to show up on the summary page whether JS is enabled or not (like we do for images on reports). But we may be better off moving this fix to another ticket (e.g. a general 'make sure things work for disabled JS' ticket? That could be worked on once most work out of the way.)

> * Regarding the message, if there isn't a simpler way I would attach  pdf with all the comments for each single item and ask the user check the file before collection.
> 
> Let me know what you think

I think this could work; it would be such a small number of users seeing this that it doesn't need to be amazing 😛 

So if you add that PDF change I can approve this PR 👍

lucascumsille:
@nephila-nacrea Sorry the delay I've been busy with other SocietyWorks + Climate tickets.

1. Regarding the images, I would definitely move it to a later ticket. At least for the rest of this sprint I don't think I will be able to solve it. I would probably put this in a summary ticket. What I would ask you if you could tweak the summary form so instead of the file name, we use the image, but again I would probably leave that one for the summary ticket so we work over the file that we will end using.
2. For the summary bug displaying only one file name. Not sure if this branch is using the summary that I built, I don't think so. Either way it would seem that should be in a summary ticket, more than in this one. I wonder why is not bringing all the files names, I wonder if it has to be with the form in the backend.
3. Regarding the file, I added a message at the top and not at the bottom, so users that only want one item to be collected can see the message. We still need to ask the client for the file(they might have it). Either way I would put this into a the ticket "make sure things work for disabled JS' ticket?"

Let me know if there is any feedback =)

templates/web/base/waste/bulky/items.html
```diff
@@ -9,6 +9,18 @@
...
 
+<div class="no-js-message govuk-warning-text due" style="max-width:550px">
+  <div class="govuk-warning-text__img">
+    <span class="govuk-warning-text__icon" aria-hidden="true">i</span>
+  </div>
+  <div class="govuk-warning-text__content">
+    <span class="govuk-warning-text__assistive">Important information</span>
+    <p class="govuk-!-margin-bottom-1"><strong>About your items</strong></p>
+    <p >Before continuing, please read the following file and see if any items have a note or comment that might help our crew to collect them.</p>
+    <a class="btn-primary" href="" download="">Item notes</a>
```
nephila-nacrea, line 20: Could you just add an 'XXX' comment here explaining we will need a file?

Given that the item list is determined by the admin page (http://peterborough.localhost:3000/admin/waste/<id>/bulky_items) the file may need to be generated based on that, rather than provided by the client.

nephila-nacrea:
> @nephila-nacrea Sorry the delay I've been busy with other SocietyWorks + Climate tickets.
> 
> 1. Regarding the images, I would definitely move it to a later ticket. At least for the rest of this sprint I don't think I will be able to solve it. I would probably put this in a summary ticket. What I would ask you if you could tweak the summary form so instead of the file name, we use the image, but again I would probably leave that one for the summary ticket so we work over the file that we will end using.
> 2. For the summary bug displaying only one file name. Not sure if this branch is using the summary that I built, I don't think so. Either way it would seem that should be in a summary ticket, more than in this one. I wonder why is not bringing all the files names, I wonder if it has to be with the form in the backend.
> 3. Regarding the file, I added a message at the top and not at the bottom, so users that only want one item to be collected can see the message. We still need to ask the client for the file(they might have it). Either way I would put this into a the ticket "make sure things work for disabled JS' ticket?"
> 
> Let me know if there is any feedback =)

Which ticket is the one for doing the summary page? I've realised I don't know where it is.

I have a misc/general to-do ticket - https://github.com/mysociety/societyworks/issues/3328 - where I am capturing the JS-disabled behaviour, among other things.

nephila-nacrea (changes requested):

templates/web/base/waste/bulky/items.html
lucascumsille: @nephila-nacrea I don't have access to that url, so if you could generate it would be really helpful. I really don't have more time available this sprint, at the moment I'm jumping from one project to another and I have a heavy deadline for next week with climate.

Here is the ticket for the summary:
https://github.com/mysociety/societyworks/issues/3236
Not sure what has happened with it, if it has already been merged.

I just pushed a commit for the comment =)
Let me know if there is more feedback, but I'll probably won't be able to tackle it till next sprint.

nephila-nacrea: Sorry, I think I caused you confusion. I only really wanted the comment adde; I will approve the change now. The file-generation is something to be handled later 👍

nephila-nacrea (approved):

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -2110,10 +2111,16 @@ setTimeout(function () {
         $("#add-new-item").prop('disabled', true);
     });
 
-      //Erase bulky item
+    //Erase bulky item
+    //https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/blob/master/app/assets/javascripts/autocomplete.js#L40
```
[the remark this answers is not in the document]
dracos: As in, is the functionality of .delete-item purely to deal with this issue, or is it a thing wanted anyway?

dracos:
Finally in as https://github.com/mysociety/fixmystreet/commit/103058636df7f2dce522e40953e0a1432bb4ebc2 - squashed to one, console.log removed, waste JS moved to waste JS file, global variables removed, and some other simplifications.

PR #4150, opened 2022-11-01
[Oxfordshire] Text in rss link for type and use type as check

MorayMySoc:
* When selecting a parish, the url link text says 'Get updates of parish problems'
* When selecting a division, the url link text says 'Get updates of division problems'
* The above updates are reflected on the rss page describing the area using parish or division or ward
* url link text continues to say 'Get update of ward problems' for any others
* Uses type to distinguish between parish/wards/divisions with identical names in showing reports

This is a follow on to the work done for https://github.com/mysociety/societyworks/issues/2658 fixing issues raised in https://mysocietysupport.freshdesk.com/a/tickets/1508 as well as the non-raised issue of duplicate named sub-bodies (ie district ward and parish with the same name) not being handled well

[skip changelog]


templates/web/base/reports/_ward-list.html
```diff
@@ -12,7 +12,8 @@ <h2>[% ward_text %]</h2>
       [% FOR child IN children.values.sort('name') %]
         <li class="item-list__item item-list--wards__item">
           <label class="[% show_wards_shown_class %]">
-            <input type="checkbox" name="ward" value="[% child.name | html %]">
+            [% SET value_type = child.name _ "?type=" _ child.type %]
+            <input type="checkbox" name="ward" value="[% value_type | html %]">
```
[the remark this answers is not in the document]
dracos: Also no changes to the other_districts-list.html file, so I assume this wouldn't work if you selected the ward with the same name.

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -438,13 +445,25 @@ sub ward_check : Private {
     my $qw = $c->cobrand->fetch_area_children($parent_id);
     $c->cobrand->call_hook('add_parish_wards' => $qw);
 
+    if ($c->stash->{ward_code}) {
+        for my $key (keys %$qw) {
+            if (${$qw->{$key}}{type} ne $c->stash->{ward_code}) {
+                delete $qw->{$key};
+            }
+        };
```
dracos, line 453: Instead of this block, how about something like:
```
$qw = values %$qw;
if ($c->stash->{ward_code}) {
    @$qw = grep { $_->{type} ne $c->stash->{ward_code} } @$qw;
}
```
then can use @$qw below rather than `values %$qw`.

dracos (changes requested):
Good start, few things to change or I think could hopefully be simplified.

templates/web/base/reports/_ward-list.html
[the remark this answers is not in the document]
MorayMySoc: The params are not really params - perhaps I should use something else like a colon to distinguish them? They are values passed to ward_check in an array rather than a request passed so would need to be changed more fundamentally.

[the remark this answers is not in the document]
dracos: Only the firsts checked in order to populate ward_code, and then they are all immediately stripped? So I don't understand your point - why could you not pass the value type in a hidden field into ward_code, rather than reading it off the first ward passed in and then stripping it from all of them?

perllib/FixMyStreet/App/Controller/Reports.pm
```diff
@@ -445,25 +445,19 @@ sub ward_check : Private {
...
-                delete $qw->{$key};
-            }
-        };
+        @$qw = grep { $_->{type} eq $c->stash->{ward_code} } @$qw;
     }
 
     my %names = map { $c->cobrand->short_name({ name => $_ }) => 1 } @wards;
     my @areas;
-    foreach my $area (sort { $a->{name} cmp $b->{name} } values %$qw) {
+    foreach my $area (sort { $a->{name} cmp $b->{name} } values @$qw) {
```
dracos, line 455: Don't want "values" here.

templates/web/base/reports/_other_districts-list.html
```diff
@@ -13,7 +13,8 @@ <h2>[% parish_text %]</h2>
         [% IF child.type == 'CPC' %]
         <li class="item-list__item item-list--wards__item">
           <label class="[% show_parish_shown_class %]">
-            <input type="checkbox" name="ward" value="[% child.name | html %]">
+            [% SET value_type = child.name _ "?type=" _ child.type %]
+            <input type="checkbox" name="ward" value="[% value_type | html %]">
```
dracos, line 17: Hmm, why does this base file exist at all, if it's only for Oxfordshire? The only differences appear to be bug fixes in the Oxfordshire version - can the base file be removed entirely?

dracos (changes requested):
One bug, one question, and there's the previous issue about me not understanding why every element needs to have the type added to it - ping me if I've misunderstood!

templates/web/base/reports/_ward-list.html
```diff
@@ -20,7 +20,9 @@ <h2>[% ward_text %]</h2>
       [% END %]
     </ul>
     </section>
+
     <p>
+      <input type="hidden" name="type" value="CED">
```
dracos, line 25: This will break every cobrand for which the children aren't CEDs. Given this only applies to Oxfordshire, a quick cobrand check here for now?

dracos (changes requested):
Looking good, just one base change I think won't work

dracos (approved):
Woo

PR #4151, opened 2022-11-03
[Northamptonshire] New traffic signal categories.

dracos:
Fixes https://mysocietysupport.freshdesk.com/a/tickets/2520 (allows both old and new so they can put it live their end whenever). [skip changelog]


chrismytton (approved):
🚦

PR #4158, opened 2022-11-09
Paginate alerts on user admin page

davea:
Some shared/service user accounts have thousands of alerts and this page was timing out when attempting to render them all.

Prompted by FD-2538

image

### DIFF
--- a/templates/web/base/admin/users/alerts.html
+++ b/templates/web/base/admin/users/alerts.html
@@ -54,7 +54,13 @@
     <td><input type="radio" name="edit_alert[[% alert.id %]]" value="delete"></td>
 </tr>
 [% END %]
+<tr>
+    <td colspan=5></td>
+    <td colspan=3 align="center">
+        <input class="btn" type="submit" value="[% loc('Update') %]">
+    </td>
+</tr>
 </table>
-<input class="btn" type="submit" value="[% loc('Update') %]">
 </form>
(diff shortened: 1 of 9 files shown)


templates/web/base/admin/users/alerts.html
```diff
@@ -54,7 +54,13 @@ <h3>[% loc("User's alerts") %]</h3>
...
+<tr>
+    <td colspan=5></td>
+    <td colspan=3 class="center">
+        <input class="btn" type="submit" value="[% loc('Update') %]">
+    </td>
+</tr>
 </table>
-<input class="btn" type="submit" value="[% loc('Update') %]">
 </form>
+[% INCLUDE 'pagination.html', admin = 1, param = 'p', pager = alerts_pager %]
```
dracos, line 65: What is admin=1 for here? Hmm, I see it's there all the other places, so might as well leave it, but doesn't look like it does anything any more :)

dracos (commented):
Couple of comments, but looks fine :)

templates/web/base/admin/users/alerts.html
davea: Hasn't done anything [since 2016](https://github.com/mysociety/fixmystreet/commit/9df7ab7d80a7f0b76860d93f915e07ec07668d57)! Have removed it everywhere in f44b525adc54aff934d53ce7cb6ed4219a8b2998.

PR #4162, opened 2022-11-14
[Bulky Goods] Add link to Household Recycling Centre to bin days page

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3311.

[skip changelog]

### DIFF
--- a/templates/web/peterborough/waste/_more_services_sidebar.html
+++ b/templates/web/peterborough/waste/_more_services_sidebar.html
@@ -38,6 +38,9 @@
     <li>
         <a href="[% c.uri_for_action('waste/bulky', [ property.id ]) %]">Book bulky goods collection</a>
     </li>
+    <li>
+      <a href="https://www.peterborough.gov.uk/residents/rubbish-and-recycling/household-recycling-centre">Household Recycling Centre</a>
+    </li>
   [% END %]
 </ul>
 


templates/web/peterborough/waste/_more_services_sidebar.html
```diff
@@ -38,6 +38,11 @@ <h3>More services</h3>
     <li>
         <a href="[% c.uri_for_action('waste/bulky', [ property.id ]) %]">Book bulky goods collection</a>
     </li>
+    [% IF c.cobrand.moniker == 'peterborough' %]
```
davea, line 41: This is in the `peterborough` directory so doesn't need the cobrand check :)

davea (changes requested):
One little `IF` to be removed is all!

templates/web/peterborough/waste/_more_services_sidebar.html
nephila-nacrea: 🤦 Don't know how I didn't notice that.

nephila-nacrea: The ticket didn't actually state where the link ought to go so I assume it is OK where it is.

davea: Yeah I think that's the perfect place to put it.

davea (approved):

PR #4163, opened 2022-11-14
[Camden] Front-end

lucascumsille:
Hi @chrismytton 
Sorry, this took longer than expected. But I had some issues with some of the forms input across the website, where the styles weren't consistent, and this made it difficult from an accessibility point of view.

The other reason It took me longer was that the PR changed quite a bit. As you can see, I have added more variables to the file; this should make life easier in keeping consistency across the website and checking most of the boxes when the client gives us a style guide. Most of the newly created variables are to improve accessibility and comply with sizes on smaller screens as well.

Finally, I can see that some things implemented here should be implemented across all cobrands. But due to the time scale, and also it needs further discussion with more developers. I went for the quickest fix, which is just implement them on this cobrand. I'll raise an issue on the FMS repo to discuss those potential fixes.

Let me know if you have any feedback, and before letting the client know, can you let me check the staging website?
Thanks for all your help.

###[ List of issues fixed](https://3.basecamp.com/4020879/buckets/27695641/todos/5470577758#__recording_5518289053)

**Header** 
- [X] On 'Αll reports', the header is slightly higher than on other pages, which makes the menu jump when you go from one page to another.

 
**'Report a problem' page**
- [ ] Page title should have different wording but we might be able to change that on our end?
@jacquelinebylau and @chrismytton I don't think they can change it from their end. We should ask them what are they thinking to put in here.

**Footer**
- [X] Links in footer should follow same styles as in top menu (except font size, which has been correctly applied): not underlined by default, underlined on hover.

**'Help' page/privacy link:** 
- [X]  The link in the right-hand column 'submenu' links now to the Camden website, so it is just a duplication of the link in the footer. Suggestion: Remove the right-hand column all together, keep link in footer.

 
**Typography**
- [X] H1 ok
- [X] H2 should be 32px on large viewports. Currently, it's 24px across all viewports
- [X] Body copy now ok

 
**'Numbers' on Homepage:** 

- [X ] Font size too small on small viewports and too large on large viewports. Should be 36px (small viewports) and 48px (large viewports). This applies to 'How to report' numbers and dashboard figures below.

 
**Buttons**
- [ ] Remove drop shadow from all buttons @jacquelinebylau I couldn't find any with it. Not sure if it's worth it to mention it to them.
- [X] 'Report' button in header (on small viewports only): Hover state: white border missing

 
**Input fields**
- [X] Remove all drop shadow from input fields

'All reports': Filter inputs: 
- [ X] height and style as per style guide
- [X ] no rounded corners

'Local RSS feeds and email alerts'
- [X] height and style as per style guid
- [X] no rounded corners

**Homepage / Report an issue**
- [X] Spacing between input section and instructions below too wide. Should be 35px (small viewports) and 45px large viewports

**All pages**
- [X] Line length should not exceed 770px.

This fixes: https://github.com/mysociety/societyworks/issues/3296


chrismytton (approved):
Thanks @lucascumsille!

PR #4164, opened 2022-11-14
[Camden] Add NSGRef to reports in JS and on server

chrismytton:
The JS asset changed are in the config file under the `asset_layers` key.

Part of https://github.com/mysociety/societyworks/issues/3305

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Camden.pm
+++ b/perllib/FixMyStreet/Cobrand/Camden.pm
@@ -64,4 +64,14 @@ sub open311_extra_data_include {
     return [];
 }
 
+sub open311_config {
+    my ($self, $row, $h, $params) = @_;
+    $params->{multi_photos} = 1;
+}
+
+sub open311_munge_update_params {
+    my ($self, $params, $comment, $body) = @_;
+    $params->{service_request_id_ext} = $comment->problem->id;
+}
+
 1;
--- a/perllib/FixMyStreet/Cobrand/Camden.pm
+++ b/perllib/FixMyStreet/Cobrand/Camden.pm
@@ -34,4 +34,34 @@ sub privacy_policy_url {
 
 sub admin_user_domain { 'camden.gov.uk' }
 
+sub lookup_site_code_config {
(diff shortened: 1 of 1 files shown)


chrismytton:
It has been pointed out to me that this should be using the ✨ New Asset Configuration ✨  rather than being in JS, will change.

perllib/FixMyStreet/Cobrand/Camden.pm
```diff
@@ -34,4 +34,42 @@ sub privacy_policy_url {
...
+    # uncoverable subroutine
+    # uncoverable statement
+    {
+        buffer => 1000, # metres
+        url => "https://tilma.mysociety.org/mapserver/camden",
+        srsname => "urn:ogc:def:crs:EPSG::27700",
+        typename => "Streets",
+        property => $property,
+        accept_feature => sub {
+            # Sometimes the nearest feature has a NULL streetref1 property
```
davea, line 49: These `streetref1` references should be `NSG_REF`, based on L67.

davea (approved):
One tiny thing in the comment, but 👍

perllib/FixMyStreet/Cobrand/Camden.pm
dracos: Also there aren't any null entries, so not sure the whole function is needed :)

dracos:
While you're here, as I've just done this for Brent, you'll want these two functions in the same file:
```
+sub open311_config {
+    my ($self, $row, $h, $params) = @_;
+    $params->{multi_photos} = 1;
+}
+
+sub open311_munge_update_params {
+    my ($self, $params, $comment, $body) = @_;
+    $params->{service_request_id_ext} = $comment->problem->id;
+}
```
First is because Symology open311-adapter accepts multiple photo URLs and puts them all in the description (think that's all it does), and the latter is because sending an update to Symology requires the FMS ID.

chrismytton:
Thanks both! Have added an extra commit with the open311 hooks Matthew mentions.

PR #4165, opened 2022-11-14
[Brent] Add cobrand hook to stop questionnaire changing status

MorayMySoc:
* Add cobrand hook 'prevent_questionnaire_updating_status'
* Add mechanism for stopping status of report updating when cobrand hook present
* Add tests

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Questionnaire.pm
+++ b/perllib/FixMyStreet/App/Controller/Questionnaire.pm
@@ -160,7 +160,7 @@ sub record_state_change : Private {
     $c->stash->{new_state} = $new_state;
 
     # Record state change, if there was one
-    if ( $new_state ) {
+    if ( $new_state && !$c->cobrand->call_hook("prevent_questionnaire_updating_status")) {
         $problem->state( $new_state );
         $problem->lastupdate( \'current_timestamp' );
     } elsif ($problem->state ne $old_state) {
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -58,4 +58,6 @@ sub open311_extra_data_include {
     return $open311_only;
 }
 
+sub prevent_questionnaire_updating_status { 1 };
+
 1;
--- a/perllib/FixMyStreet/Script/Questionnaires.pm
+++ b/perllib/FixMyStreet/Script/Questionnaires.pm
@@ -64,8 +64,8 @@ sub send_questionnaires_period {
             next;
         }
 
-        # Cobranded and non-cobranded messages can share a database. In this case, the conf file 
-        # should specify a vhost to send the reports for each cobrand, so that they don't get sent 
+        # Cobranded and non-cobranded messages can share a database. In this case, the conf file
+        # should specify a vhost to send the reports for each cobrand, so that they don't get sent
         # more than once if there are multiple vhosts running off the same database. The email_host
         # call checks if this is the host that sends mail for this cobrand.
         next unless $cobrand->email_host;
--- a/t/cobrand/brent.t
+++ b/t/cobrand/brent.t
@@ -1,0 +1,69 @@
+use FixMyStreet::TestMech;
+
+my $mech = FixMyStreet::TestMech->new;
+
+# disable info logs for this test run
+FixMyStreet::App->log->disable('info');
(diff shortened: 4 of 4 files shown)


dracos (changes requested):
This doesn't just stop the problem state changing, it also prevents the questionnaire recording their response entirely, so if they say it's been fixed, it'd be recorded (and displayed [1]) that they said it hadn't, and if it was marked as fixed and vice-versa. I think it needs to still let whatever they put be recorded as their response to the questionnaire, but it only doesn't update the problem's state from that.

[1] If new state isn't set, the report is open or closed (but not fixed), and they've answered "No", the questionnaire result is recorded as a steady state, and the front end code then takes that as it must still be open by that.

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -24,4 +24,8 @@ sub disambiguate_location {
     bounds => [ 51.52763684136, -0.335577710963202, 51.6003693511994, -0.191492539132886 ],
 }
 
+sub prevent_questionnaire_updating_status { 1 };
+
+sub send_questionnaires { 1 };
```
dracos, line 29: This is the Default.pm entry, but I guess no harm in repeating.

perllib/FixMyStreet/App/Controller/Questionnaire.pm
```diff
@@ -151,12 +151,23 @@ sub record_state_change : Private {
...
     my $old_state = $c->stash->{old_state};
     my $new_state = '';
-    $new_state = 'fixed - user' if $c->stash->{been_fixed} eq 'Yes' &&
-        FixMyStreet::DB::Result::Problem->open_states()->{$old_state};
-    $new_state = 'fixed - user' if $c->stash->{been_fixed} eq 'Yes' &&
-        FixMyStreet::DB::Result::Problem->closed_states()->{$old_state};
-    $new_state = 'confirmed' if $c->stash->{been_fixed} eq 'No' &&
-        FixMyStreet::DB::Result::Problem->fixed_states()->{$old_state};
+    if ($c->cobrand->call_hook("prevent_questionnaire_updating_status")) {
+        $new_state = '';
```
dracos, line 155: This line isn't needed.

perllib/FixMyStreet/App/Controller/Questionnaire.pm
```diff
@@ -151,12 +151,23 @@ sub record_state_change : Private {
...
+        $new_state = 'fixed - user' if $c->stash->{been_fixed} eq 'Yes' &&
+            FixMyStreet::DB::Result::Problem->open_states()->{$old_state};
+        $new_state = 'fixed - user' if $c->stash->{been_fixed} eq 'Yes' &&
+            FixMyStreet::DB::Result::Problem->closed_states()->{$old_state};
+        $new_state = 'confirmed' if $c->stash->{been_fixed} eq 'No' &&
+            FixMyStreet::DB::Result::Problem->fixed_states()->{$old_state};
+    }
     $c->stash->{new_state} = $new_state;
 
     # Record state change, if there was one
```
dracos, line 173: You still have new_state set to blank, so the actual thing they submitted is not saved in the questionnaire (as it would be in line 190 below), but instead something they did not put is potentially saved.

perllib/FixMyStreet/App/Controller/Questionnaire.pm
```diff
@@ -215,6 +226,7 @@ sub submit_standard : Private {
                 anonymous    => $problem->anonymous,
             }
         );
+        $update->set_extra_metadata('questionnaire_as_comment', 'Reporter says ' . $c->stash->{status_comment} . ' via questionnaire' ) if $c->stash->{status_comment};
```
dracos, line 229: This doesn't work for translation purposes, I'm afraid, you cannot concatenate substrings like this. I'd rather not store a string in the DB, makes it hard to change in future. Do you think this is needed? I thought the existing message worked okay, even if the problem state didn't update. "Still open, via questionnaire" makes sense as a comment, I think, and we don't need to worry about the fixed way round I think? Happy to chat about it.

dracos (changes requested):
Afraid it's still not storing what they submitted in the questionnaire table; as I understand it, what needs changing if the flag is set is where the $problem is updated, but that's it. Happy to chat if helpful.

dracos (approved):
Woo, thanks :)

PR #4166, opened 2022-11-14
[Bulky Goods] Location page improvements

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3310.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -65,8 +65,9 @@ has_page add_items => (
 );
 
 has_page location => (
-    title => 'Location details',
-    fields => ['location', 'location_photo', 'location_photo_fileid', 'continue'],
+    title    => 'Location details',
+    fields   =>
+        [ 'location', 'location_photo', 'location_photo_fileid', 'continue' ],
     next => 'summary',
 );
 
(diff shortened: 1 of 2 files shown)


davea (changes requested):
I think in general we should aim to reduce the need for custom templates for form pages as much as possible. Each new template means more places to keep things in sync - as you saw with the addition of the `bulky/_header.html` file :) It also means more places to look if copy/labels need to be updated in the future.

Instead, keeping the text within the field definition makes things a little more streamlined. I've created a commit with a suggestion for how to get the text onto the page without a new template - 32fea91c790ed9af9aa8c2340ac7eef7e6fd6d39.

What do you think?

nephila-nacrea:
> I think in general we should aim to reduce the need for custom templates for form pages as much as possible. Each new template means more places to keep things in sync - as you saw with the addition of the `bulky/_header.html` file :) It also means more places to look if copy/labels need to be updated in the future.
> 
> Instead, keeping the text within the field definition makes things a little more streamlined. I've created a commit with a suggestion for how to get the text onto the page without a new template - [32fea91](https://github.com/mysociety/fixmystreet/commit/32fea91c790ed9af9aa8c2340ac7eef7e6fd6d39).
> 
> What do you think?

I think that should work; I guess I thought I should be adding extra messaging directly into the template, but it's created more stuff, I suppose!

davea (approved):

PR #4167, opened 2022-11-15
[Bulky Goods] Deep link to bulky workflow after selecting address

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3309.

In addition to the deep link, this PR implements a check for existing pending bulky collections and redirects the user accordingly.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -493,7 +493,28 @@ sub look_up_property {
 
     my %premises = map { $_->{uprn} => $_ } @$premises;
 
+    $premises{$uprn}{has_pending_bulky_collection}
+        = $self->has_pending_bulky_collection( $premises{$uprn} );
+
     return $premises{$uprn};
+}
+
+sub has_pending_bulky_collection {
+    my ( $self, $property ) = @_;
+
+    my @collections = FixMyStreet::DB->resultset('Problem')->search(
+        {   category => 'Bulky collection',
+            extra    => {
+                      like => '%T4:uprn,T5:value,I'
+                    . length( $property->{uprn} ) . ':'
+                    . $property->{uprn} . '%',
+            },
+            state =>
+                { '=', [ FixMyStreet::DB::Result::Problem->open_states ] },
+        },
+    )->all;
+
+    return @collections ? 1 : 0;
 }
 
 sub image_for_unit {
--- a/t/app/controller/waste_peterborough.t
+++ b/t/app/controller/waste_peterborough.t
@@ -740,6 +740,22 @@ FixMyStreet::override_config {
         # behaviour of the system. They are here to remind us to update them as
         # we break them by implementing the correct behaviour :)
 
+        subtest '?type=bulky redirect' => sub {
+            $mech->get_ok('/waste?type=bulky');
+            is $mech->uri, 'http://localhost/waste?type=bulky',
+                'No redirect if no address data';
+            $mech->content_contains( 'What is your address?',
+                'user on address page' );
+
+            $mech->submit_form_ok(
+                { with_fields => { postcode => 'PE1 3NA' } } );
+            $mech->submit_form_ok(
+                { with_fields => { address => 'PE1 3NA:100090215480' } } );
+            is $mech->uri,
+                'http://localhost/waste/PE1%203NA:100090215480/bulky',
+                'Redirected to /bulky if address data';
+        };
+
         $mech->get_ok('/waste/PE1%203NA:100090215480');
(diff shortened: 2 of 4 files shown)


templates/web/base/waste/bin_days.html
```diff
@@ -8,6 +8,19 @@ <h1 class="govuk-heading-xl">Your bin days</h1>
 
 [% INCLUDE 'waste/_address_display.html' %]
 <div class="waste__collections">
+
+[% IF property.has_pending_bulky_collection %]
```
nephila-nacrea, line 12: Not sure how this should look. It looks ugly :-/

t/app/controller/waste_peterborough.t
```diff
@@ -848,6 +864,18 @@ FixMyStreet::override_config {
...
+            $mech->get_ok('/waste?type=bulky');
+            $mech->content_contains( 'What is your address?',
+                'user on address page' );
+            $mech->submit_form_ok(
+                { with_fields => { postcode => 'PE1 3NA' } } );
+            $mech->submit_form_ok(
+                { with_fields => { address => 'PE1 3NA:100090215480' } } );
+            is $mech->uri,
+                'http://localhost/waste/PE1%203NA:100090215480',
+                'Redirected to waste base page';
```
davea, line 877: Would be good to test that the "You have a pending bulky collection." message appears on the page at this point.

templates/web/base/waste/bin_days.html
davea: This should probably be in the `bin_days_bulky.html` template as that's what's used instead of this one.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -495,9 +495,27 @@ sub look_up_property {
...
+
     return $premises{$uprn};
 }
 
+sub has_pending_bulky_collection {
+    my ( $self, $property ) = @_;
+
+    my @collections = FixMyStreet::DB->resultset('Problem')->search(
+        {   category => 'Bulky collection',
+            detail   => { -like => '%' . $property->{address} . '%' },
```
davea, line 510: This works well right now, but I'd be worried about it breaking if we did something like stopped storing the address in the detail field, or changed the way addresses are formatted, etc etc. Looking for a more stable value like the UPRN in the `extra` is a better bet. Something like this:
```suggestion
            extra => { like => '%T7:uprn,T' . length($property->{uprn}) . ':' . $property->{uprn} . '%'},
```

(aside, yes, it's currently a bit horrid to query the `extra` field. There is [a PR](https://github.com/mysociety/fixmystreet/pull/3220) that'll make it much easier)

davea (changes requested):
Looking good so far! A couple of small changes is all :)

davea (changes requested):
Ah yes, I see what you mean now! I've just realised that actually Lucas created the relevant blocks in the template for this situation, they're just not wired up yet. I think if you remove your `IF` block and replace occurrences of `existing_booking` with `property.has_pending_bulky_collection` in `bin_days_bulky.html` that'll get us closer to the desired result. It's not perfect yet but I think fully implementing the logic for that template is beyond this particular PR's scope.

nephila-nacrea:
> Ah yes, I see what you mean now! I've just realised that actually Lucas created the relevant blocks in the template for this situation, they're just not wired up yet. I think if you remove your `IF` block and replace occurrences of `existing_booking` with `property.has_pending_bulky_collection` in `bin_days_bulky.html` that'll get us closer to the desired result. It's not perfect yet but I think fully implementing the logic for that template is beyond this particular PR's scope.

Had wondered about those `existing_booking` sections but had assumed implementing them fully was beyond the scope of this ticket. Have replaced `existing_booking` with `property.has_pending_bulky_collection` like you've suggested, though like you've said, it's not perfect.

t/app/controller/waste_peterborough.t
davea: Rather than removing the test, better to update it with the wording that you expect to appear:

```perl
$mech->content_contains('You need to sign in to see the full details of this property.');
```

davea (changes requested):
Sorry just one tiny thing to improve test coverage.

t/app/controller/waste_peterborough.t
nephila-nacrea: This fails for me, I am assuming because the `NOT c.user_exists` check in the bin days template is false. I am not sure how to get a good bit of text to display without rewriting more of the template so it uses actual values rather than placeholders.

davea: Ah yes I see. I guess you could log out at that point in the test? Might have to log back in again if subsequent tests require it.

davea (approved):
Looks good! 👍

PR #4168, opened 2022-11-15
Restrict report and user flagging to superusers only

neprune:
closes https://github.com/mysociety/societyworks/issues/2580

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -297,9 +297,12 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         my $old_state = $problem->state;
 
         my %columns = (
-            flagged => $c->get_param('flagged') ? 1 : 0,
             non_public => $c->get_param('non_public') ? 1 : 0,
         );
+        # Only superusers can flag / unflag a report
+        if ($c->user->is_superuser) {
+            $columns{flagged} = $c->get_param('flagged') ? 1 : 0;
+        }
         foreach (qw/state anonymous title detail name external_id external_body external_team/) {
             $columns{$_} = $c->get_param($_);
         }
--- a/perllib/FixMyStreet/App/Controller/Admin/Users.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Users.pm
@@ -177,7 +177,8 @@ sub add : Local : Args(0) {
             phone => $phone || undef,
             phone_verified => $phone && $phone_v ? 1 : 0,
             from_body => $from_body,
-            flagged => $c->get_param('flagged') || 0,
+            # Only superusers can flag users, unless for Zurich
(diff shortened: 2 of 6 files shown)


perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -297,7 +297,8 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         my $old_state = $problem->state;
 
         my %columns = (
-            flagged => $c->get_param('flagged') ? 1 : 0,
+            # Only superusers can flag / unflag a report
+            flagged =>  $c->user->is_superuser ? ( $c->get_param('flagged') ? 1 :  0 ) : $problem->flagged,
```
dracos, line 301: We don't need to set all the columns here, only the ones we want to change, so I think it'd be a bit clearer if flagged wasn't set in this default `%columns` creation (given don't need to for non superusers), but then afterwards somewhere you had a `if (superuser) { $columns{flagged} = $c->get_param('flagged') ? 1 :  0; }` similar to what you've done in Users.pm

perllib/FixMyStreet/App/Controller/Admin/Users.pm
```diff
@@ -321,7 +322,10 @@ sub edit : Chained('user') : PathPart('') : Args(0) {
         $user->phone_verified( $phone_v );
         $user->name( $name );
 
-        $user->flagged( $c->get_param('flagged') || 0 );
+        # Only superusers can flag / unflag a user
```
dracos, line 325: This might need to go under the superuser or Zurich bit of code bit further down, see other comment; and I guess the new creation too.

t/cobrand/zurich.t
```diff
@@ -1072,15 +1072,9 @@ $mech->log_out_ok;
 subtest 'users at the top level can be edited' => sub {
     $mech->log_in_ok( $superuser->email );
     $mech->get_ok('/admin/users/' . $superuser->id );
-    $mech->content_contains('name="flagged">');
```
dracos, line 1075: Ah, so now upon looking more closely - Zurich, which as we chatted about is different in a few ways, uses 'flagged' on a user as a reference for a "government-internal" user (see the templates/web/zurich/admin/users/form.html template). So they need to maintain this behaviour, so these tests should pass unchanged.

templates/web/zurich/admin/users/form.html
```diff
@@ -19,12 +19,14 @@
           </select>
         </li>
 
-        <li>
-          <label>
-            [% loc('government-internal') %]
-            <input type="checkbox" id="flagged" name="flagged"[% user.flagged ? ' checked' : '' %]>
-          </label>
-        </li>
+        [% IF c.user.is_superuser %]
```
dracos, line 22: So also this template should remain unchanged (the base changes are fine).

dracos (changes requested):
Zurich will need to keep their current behaviour, sorry about that, and one other tiny suggestion, but the main changes look good :)

dracos (approved):
Looks good, thanks :)

PR #4169, opened 2022-11-16
Async WasteWorks backend calls on bin days page

davea:
This drastically speeds up the response time of "your bin days" page loads by no longer waiting for the results of the `call-wasteworks-backend` script before rendering the page.

Instead, that script calls `fork()` and returns immediately. In that case the FMS shows a loading spinner and sets the `http-equiv=refresh` header to reload after a second. Next time the page loads, if `call-wasteworks-backend` determines the `fork()`ed process has finished then its results are available to the calling code in `ParallelAPI.pm` and things proceed as usual.

Might help a bit with https://github.com/mysociety/sysadmin/issues/1696 as it should reduce the incidence of long-running FMS requests.

Deployment note: Need to set `WASTEWORKS_BACKEND_TMP_DIR` (in `general.yml`) to a directory where the FMS user can write/delete temporary files.

This is very much a proof-of-concept, some obvious flaws/improvements (not an exhaustive list!):

 - ~~Currently breaks Bromley/Sutton/Kingston~~
 - ~~Only works for the "Your bin days" page on Peterborough~~
   - It does work if you enable it for Bromley by changing `0` to `1` [here](https://github.com/mysociety/fixmystreet/blob/a6c60d3ff8e3a551623e733bf2cfb0e3acc3c717/perllib/FixMyStreet/Cobrand/Bromley.pm#L588) but there are some Echo API requests before that point so each refresh is still a bit heavy and might just make things worse.
 - ~~Needs a way for calls to `call_api` to opt-in/-out of the `fork()` behaviour as it might not always be required~~
 - ~~The second time the loading screen is shown, I think a new `fork()`ed process will call the backend _again_ - should probably have a lock file or something~~
 - ~~An entire page refresh looks a bit clunky, could use some JS to reload only what's required (or something fancy like [htmx](https://htmx.org) which is designed for exactly that)~~
 - Some page loads do multiple batches of parallel API calls (e.g. Bromley [here](https://github.com/mysociety/fixmystreet/blob/7eeb42ee2978af70167713a3afc6cbf04f8a630c/perllib/FixMyStreet/Roles/CobrandEcho.pm#L39) and [here](https://github.com/mysociety/fixmystreet/blob/7eeb42ee2978af70167713a3afc6cbf04f8a630c/perllib/FixMyStreet/Cobrand/Bromley.pm#L588)), this can only handle one.

Anyway here it is in action:
 

https://user-images.githubusercontent.com/4776/202300495-1ba1dc94-7204-478c-a017-ad4176d650e2.mov

[skip changelog]


davea:
This has been deployed to staging for Peterborough and seems to work well!

perllib/FixMyStreet/Roles/ParallelAPI.pm
```diff
@@ -35,7 +42,10 @@ sub call_api {
     } else {
         # uncoverable statement
         system(@cmd);
-        $data = Storable::fd_retrieve($tmp);
+        unless ($?) {
+            $data = Storable::retrieve($tmp);
+            unlink $tmp; # don't want to inadvertently cache forever
+        }
     }
     $c->session->{$key} = $data;
```
dracos, line 50: Presumably don't want to do this (or the logging the call time) if there was no data?

perllib/FixMyStreet/Roles/ParallelAPI.pm
```diff
@@ -35,7 +42,10 @@ sub call_api {
     } else {
```
dracos, line 42: I think we can check if $tmp exists here and get it if it does, bypassing calling the external script at all.

bin/fixmystreet.com/call-wasteworks-backend
```diff
@@ -37,10 +38,42 @@ my $cfg = $cobrand->feature($opts->backend);
...
+    if ($pid) {
+        # we're the parent, let the caller know that the data isn't ready yet
+        exit 1;
+    } else {
+        # create the lockfile so we don't fork() the same batch of jobs again
+        open(FH, ">", $lockfile) and close(FH);
+    }
+}
+
 my $calls = decode_json($opts->calls);
```
dracos, line 74: I think you need to trap this script dying and cleaning up the lock file then, just in case something happens.

bin/fixmystreet.com/call-wasteworks-backend
```diff
@@ -37,10 +38,42 @@ my $cfg = $cobrand->feature($opts->backend);
 my $class = $opts->backend eq 'echo' ? 'Integrations::Echo' : 'Integrations::Bartec';
 my $integration = $class->new(%$cfg);
 
+my $lockfile;
+if ($opts->fork) {
+    unless ($opts->out) {
+        die "Must use --out with --fork";
+    }
+
+    $lockfile = $opts->out . ".lock";
```
dracos, line 49: An alternative here would be (I think?) to open the $opts->out file first off (with fork or not), `flock` it exclusively, then use store_fd to save to the already open handle. The front end would then: if file doesn't exist, run this script with `&` to background it immediately; if it does exist, try and get a non-blocking shared lock on the output file - if that fails, it's still being looked up, show loading page; if that succeeds, load in the data. I don't think anything else in this script would then need to change - it wouldn't have to worry about anything at all really, it doesn't need to fork or think about things like that. Does that work? Will come back to this next week.

dracos (commented):
A good start, nice to see it working like this. Some thoughts:
* Unlikely to happen, as the data should be stored in the session first, but we'd probably need to think about how to deal with a POST if the data wasn't present - fall back to the current behaviour and not do this new fork if it's a POST, I guess would be easiest, just to have it working in such a scenario.
* I see from looking at this (and the logs output) that the external script call out takes 2-3 seconds to run just with `--help` on staging! I've cut that quite a lot in new commit by getting rid of the use of cobrands and going direct to the config (calling in all the cobrands pulls in a /lot/ more code to spin up). But even with that improvement, using the script both to kick off the calls and to check for the data being available I think is unnecessary - I think we can bypass calling the script at all if $tmp is there, as that's all it does in that scenario...
* Also think maybe the forking/etc can be simplified with some flocking, if I've understood it right (maybe, hopefully?)
* Having it work for Bromley would then be ideal, I assume - the second batch requires the first batch, but I think, given they're both always called (though in slightly different places as currently laid out), you could probably have a slightly different external script that did the bit of processing needed between them and then returned it all somehow with a bit more work.

bin/fixmystreet.com/call-wasteworks-backend
davea: Is that a case of using an `END` block, or something lower level?

dracos: `END` would work presumably, yeah, but my other comment about not changing this file and using flock instead sounds like a better way to go? Then there wouldn't be a lock file.

davea: Yep probably. Was trying to do the ~simplest changes first in case we needed to press this into production in the same form as it's been tested on staging the last couple of months. Don't think that's the case now though, so will look at flocking stuff.

bin/fixmystreet.com/call-wasteworks-backend
davea: Have implemented this logic in d1aeec07589365399b0f2e323bfc1f1728c00537. Works well locally, will see how it goes on staging.

davea:
Deployed the flock implementation staging and it seems to work well! Thanks for the suggestion, makes it much simpler.

The diff ended up being quite small, so could probably be squashed down to a single commit before merging. Have left my working visible in case that's helpful for review.

perllib/FixMyStreet/Roles/ParallelAPI.pm
```diff
@@ -2,27 +2,35 @@ package FixMyStreet::Roles::ParallelAPI;
...
     my $type = $self->backend_type;
     $key = "$cobrand:$type:$key";
     return $c->session->{$key} if !FixMyStreet->test_mode && $c->session->{$key};
 
-    my $tmp = File::Temp->new;
+    my $calls = encode_json(\@_);
+
+    my $outdir = FixMyStreet->config('WASTEWORKS_BACKEND_TMP_DIR');
+    mkdir($outdir) unless -d $outdir;
+    my $tmp = $outdir . "/" . md5_hex("$key $calls");
```
dracos, line 26: We store the data in the session based solely on the $key, so is there a reason that wouldn't be enough for the tmp file too? Feel I've missed something obvious :)

dracos (commented):
Looks good, glad it worked, though looking back at it I've forgotten how it works again for now ;) One query on the key being used is all.

perllib/FixMyStreet/Roles/ParallelAPI.pm
davea: It felt more correct somehow to have the filename based on the actual API calls being made. I think _technically_ it'd be possible for the same `$key` to be used for different batches of calls (e.g. imagine looking up available slots either side of the 15:00 booking cut off) but appreciate it's unlikely. Don't think doing it this way causes any issues though?

dracos: No, think it's fine.

dracos: (though if that's the case, would that lead to bad data being stored in the session, or does that not matter). Anyway, is fine if you think it's okay

PR #4170, opened 2022-11-17
[Camden] Enable social auth for Microsoft Azure login

chrismytton:
Enable social auth for the Camden cobrand.

At this stage we don't know which fields the name/email are actually in, so I've put in some placeholders and some logging to help us figure out where they are.

Credentials for this are in the config files on the servers.

Part of https://github.com/mysociety/societyworks/issues/3307

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Camden.pm
+++ b/perllib/FixMyStreet/Cobrand/Camden.pm
@@ -92,4 +92,24 @@ sub is_problem_anonymous { 1 }
 
 sub is_comment_anonymous { 1 }
 
+sub social_auth_enabled {
+    my $self = shift;
+
+    return $self->feature('oidc_login') ? 1 : 0;
+}
+
+sub user_from_oidc {
+    my ($self, $payload) = @_;
+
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/Cobrand/Camden.pm
```diff
@@ -92,4 +92,24 @@ sub is_problem_anonymous { 1 }
...
+
+    return $self->feature('oidc_login') ? 1 : 0;
+}
+
+sub user_from_oidc {
+    my ($self, $payload) = @_;
+
+    # TODO: Remove this once we know which fields the name/email are in.
+    use Data::Dumper;
+    $self->{c}->log->debug("Camden OIDC payload: " . Dumper($payload));
```
davea, line 106: Are these lines still required?

chrismytton: For the time being yes, because we don't yet know which fields the email/name will be in. This logging makes it easier to see what's in the payload, so when the email/name are there we can easily tell what fields they're in.

PR #4172, opened 2022-11-18
Move reopening_disallowed to config

chrismytton:
This means that configuring whether reopening is disallowed can be done in config, instead of each cobrand having similar code.

Also should fix Thamesmead, which was still allowing reports to be reopened on .com.

## Note to deployer

The following config will need to be added to FixMyStreet's general.yml before this is deployed:

```yaml
  reopening_allowed:
    camden: staff
    merton: staff
    thamesmead: reporter
    fixmystreet:
      Camden: staff
      Merton: staff
      Thamesmead: reporter
```

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -6,6 +6,7 @@ with 'FixMyStreet::Roles::BoroughEmails';
 
 use constant COUNCIL_ID_BROMLEY => 2482;
 use constant COUNCIL_ID_ISLEOFWIGHT => 2636;
+use FixMyStreet::Cobrand::Default;
 
 sub on_map_default_status { return 'open'; }
 
@@ -309,7 +310,7 @@ sub per_body_config {
     my ($self, $feature, $problem) = @_;
 
     # This is a hash of council name to match, and what to do
-    my $cfg = $self->feature($feature) || {};
+    my $cfg = $self->feature($feature);
 
     my $value;
     my $body;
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -488,11 +488,31 @@ around 'munge_sendreport_params' => sub {
...
+        return 1;
+    } elsif ($cfg eq 'staff') {
+        # Only staff and superusers can reopen
+        my $staff = $c->user_exists && $c->user->from_body && $c->user->from_body->name =~ /$body/;
+        my $superuser = $c->user_exists && $c->user->is_superuser;
+        return 1 unless $staff || $superuser;
+    } elsif ($cfg eq 'reporter') {
+        # Only the original reporter can reopen
+        return 1 unless $c->user_exists && $c->user->id == $problem->user_id;
+    }
```
chrismytton, line 512: This currently doesn't allow staff to reopen reports if the config is set to "reporter". In that case the original reporter _and_ staff should be able to reopen reports.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -458,6 +458,28 @@ sub updates_disallowed {
...
+
+    my $cfg = $self->feature('reopening_allowed') || '';
+    if ($cfg eq 'none') {
+        return 1;
+    } elsif ($cfg eq 'staff') {
+        # Only staff and superusers can reopen
+        my $staff = $c->user_exists && $c->user->from_body && $c->user->from_body->name eq $self->council_name;
+        my $superuser = $c->user_exists && $c->user->is_superuser;
+        return 1 unless $staff || $superuser;
+    }
```
chrismytton, line 477: Would be good to factor the shared bits of this method out to avoid duplication. Currently this version doesn't handle the "reporter" config at all.

chrismytton (commented):
I should probably write some tests specifically for this, rather than relying on the existing Merton ones.

dracos:
Note this PR would work/solve the same request in https://github.com/mysociety/societyworks/issues/3319 presumably

chrismytton:
Going to declare this abandoned for now. Can always revive if there's a need in the future.

PR #4173, opened 2022-11-18
[Bromley] Add cookie banner & analytics to WasteWorks pages

davea:
For FD-2536.

(It seems impossible to run this outside of production as the Cookie Control code checks that the domain matches `bromley.gov.uk`)

[skip changelog]


dracos (approved):
Note Bromley have a CSP in configuration, so that will need to be expanded to include the new third party domains (and hope GTM doesn't pull in more 3rd parties...?).

PR #4177, opened 2022-11-21
[Bulky Goods] Check slot is still available before payment

nephila-nacrea:
Part of https://github.com/mysociety/societyworks/issues/3312.

Since we are not locking collection slots, we need to check that a given slot is still not fully booked before the user moves on to payment.

[skip changelog]


perllib/FixMyStreet/App/Form/Wizard.pm
```diff
@@ -53,6 +53,7 @@ sub next {
         $next = $next->(
             $self->saved_data,
             $self->c->req->params,
+            $self, # Ugh
```
nephila-nacrea, line 56: I don't like this; I feel 'next' should just take the form/$self as the first (and only?) arg - but I didn't want to go around refactoring all existing 'next' subs.

davea: I think it's the perfect time to do that refactor, tbh. There aren't *that* many (I count 10?) and it'd make them all a bit clearer. I agree that `$self` should be the only arg. The existing test coverage should be a good safety net!

davea (changes requested):

perllib/FixMyStreet/App/Form/Wizard.pm
nephila-nacrea: Hmm I've got stuck on this. For some reason, when I just use $self, and then use `$self->c->req->params` in e.g. the 'next' sub for `has_page choose_date_earlier` in the Bulky.pm form, I get `Can't call method "c" on an undefined value`. Like it's not getting a form object at all, despite the original passing in of `$self->c->req->params` working fine? I don't know if I am overlooking something really silly 😞

dracos:
If the code without trying to refactor is running okay, and there's no other issues, but refactoring is hard/taking too much time to get working, I suggest perhaps leaving the refactor to a later time at this point.

davea (approved):
💯 Agree with Matthew - looks good as-is!

dracos:
Now payment branch has been merged, I think this might need a small change to where it's 'activated' as the payment button step is gone.

dracos:
@davea Hmm, in order to get this working on top of payments, have had to introduce a new pre_finish argument that can intercept the call finished and stop it running (so here, we want to redirect to a different page before finished runs), and I couldn't see another place to do it. The only other option was to say the form didn't validate, which is sort of technically correct, but then the code wants to show the same form, not move on somewhere else. Anyway, tests are passing now, so do take a look and check it's okay.

davea (approved):
✨

PR #4183, opened 2022-11-24
[Shropshire] Enable Open311 photo upload

MorayMySoc:
* Add photo uploads to Confirm

Servers branch: shropshire-confirm-photo-upload-live

For FD-2403

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Shropshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Shropshire.pm
@@ -63,4 +63,11 @@ sub open311_contact_meta_override {
     }
 }
 
+sub open311_config {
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
The FD ticket says they always want the photos to be uploaded, but this will only do so if the report is marked non-public (see Open311.pm, the line that checks non_public). I can't see any testing since 1st November and those reports did not send through photos. So I'm very confused. You would need to set always_upload_photos to always do it, I think?

dracos:
Ah open311-adapter fetches the photos, so it still works without.

dracos (approved):

PR #4184, opened 2022-11-24
[Camden] Hide TfL's River Piers category

chrismytton:
Camden don't want this displayed on their cobrand as it's not relevant to their area.

Fixes https://github.com/mysociety/societyworks/issues/3351


MorayMySoc (approved):
Potentially this could be extended to include FMS, but I expect it doesn't matter much if they are only really concerned with the cobrand not having it

chrismytton:
> Potentially this could be extended to include FMS, but I expect it doesn't matter much if they are only really concerned with the cobrand not having it

Ah yes, hadn't thought of that! I think as you say should be ok for now. Easy enough to change if they ask for it in the future.

PR #4185, opened 2022-11-24
[Peterborough] Bulky logged and reminder emails

dracos:
For https://github.com/mysociety/societyworks/issues/2901 (apart from querying if something has been cancelled, awaiting details on that) 
For https://github.com/mysociety/societyworks/issues/2913 (bar text changes and cancellation link - which I think should be the "view bookings" page, most straightforwardly)
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -9,6 +9,7 @@ use DateTime::Format::Strptime;
 use Integrations::Bartec;
 use List::Util qw(any);
 use Sort::Key::Natural qw(natkeysort_inplace);
+use FixMyStreet::Email;
 use FixMyStreet::WorkingDays;
 use Utils;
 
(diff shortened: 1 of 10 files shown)


templates/email/default/waste/bulky-reminder.html
```diff
@@ -0,0 +1,72 @@
...
+
+    [% INCLUDE '_council_reference.html' problem=report %]
+
+  <p style="[% p_style %]">
+    Collection date: [% collection_date %]
+  </p>
+
+  <p style="[% p_style %]">
+    Items to be collected:<br>
+    [% item_list.join('<br>') %]
```
davea, line 38: Not quite sure what's gone on here, but might want a `| safe` at the very least! 
image

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -1361,4 +1361,92 @@ sub bulky_available_feature_types {
...
+            $report->set_extra_metadata(reminder_1 => 1);
+            $report->update;
+        }
+    }
+}
+
+sub _bulky_send_reminder_email {
+    my ($self, $report, $h, $params) = @_;
+
+    my $token = FixMyStreet::App->model('DB::Token')->new({
```
davea, line 1425: Think you need to `use FixMyStreet::App;` up top otherwise this line errors (at least it does here):

```
Can't locate object method "model" via package "FixMyStreet::App" at [...]/fixmystreet/perllib/FixMyStreet/Cobrand/Peterborough.pm line 1428.
```

Though adding that line then results in
```
Can't call method "get" on an undefined value at [...]/fixmystreet/commonlib/perllib/mySociety/Locale.pm line 22.
```
but not sure if that's my setup?

davea (changes requested):
Couple of small bits, but nearly there!

perllib/FixMyStreet/Cobrand/Peterborough.pm
dracos: Ugh, shouldn't have used App, it's not running in the web, it's offline. Guess it is there when tests are running. Should be FixMyStreet::DB.

templates/email/default/waste/bulky-reminder.html
dracos: Ah, it looks like you might have up to ITEM_20? And I'm not filtering out blanks.

davea: Yep, that's exactly what was happening 👍

perllib/FixMyStreet/Cobrand/Peterborough.pm
davea: Aha, in that case you need `use FixMyStreet::Email;`, otherwise:

```
Undefined subroutine &FixMyStreet::Email::send_cron called at [...]/fixmystreet/perllib/FixMyStreet/Cobrand/Peterborough.pm line 1442.
```

davea (changes requested):
Crashes due to a missing `FixMyStreet::Email`, alas. Also don't forget to `chmod +x` the script.

PR #4193, opened 2022-11-28
[Bulky goods] Per-item pricing

dracos:
Still needs a 'nice admin' but can do that separately along with the max-items-per-collcetion field as well.
So fixes most of https://github.com/mysociety/societyworks/issues/2916 (bar per-item maximum I think).
[skip changelog]


davea (approved):
Looks good 👍 

One thing I forgot to mention on #4187, but we should do when making the admin forms - need to have an upper bound of 20 items per collection due to Bartec limitations.

PR #4194, opened 2022-11-29
[Camden] Hide user's name on cobrand

chrismytton:
Camden don't want users to have the option to be able to display their name publicly on the cobrand.

## Todo

- [x] Handle hiding the name for updates

Fixes https://github.com/mysociety/societyworks/issues/3300

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Comment.pm
+++ b/perllib/FixMyStreet/DB/Result/Comment.pm
@@ -283,7 +283,7 @@ sub meta_line {
 
     my $contributed_as = $self->get_extra_metadata('contributed_as') || '';
     my $staff = $self->user->from_body || $self->get_extra_metadata('is_body_user') || $self->get_extra_metadata('is_superuser');
-    my $anon = $self->anonymous || !$self->name;
+    my $anon = $self->anonymous || !$self->name || $cobrand->call_hook('is_comment_anonymous');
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DB/Result/Comment.pm
```diff
@@ -283,7 +283,7 @@ sub meta_line {
 
     my $contributed_as = $self->get_extra_metadata('contributed_as') || '';
     my $staff = $self->user->from_body || $self->get_extra_metadata('is_body_user') || $self->get_extra_metadata('is_superuser');
-    my $anon = $self->anonymous || !$self->name;
+    my $anon = $self->anonymous || !$self->name || $cobrand->call_hook('is_comment_anonymous', $self);
```
MorayMySoc, line 286: I don't think you need to pass anything to the call_hook as it is unless this is something I don't understand - same in below call

MorayMySoc (approved):
All seems to work fine - just the query about passing to the call hook

perllib/FixMyStreet/DB/Result/Comment.pm
chrismytton: Yeah you're right. I was trying to predict the future, but if/when this is needed the problem/comment argument can be added then. Will remove.

PR #4200, opened 2022-12-01
[Bulky Goods] Handle coming back from payment page

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3370 [skip changelog] The other approach would have been to save `saved_data` in the session or somewhere just before redirecting to payment page, and then picking it up if they come back, but assuming most uses aren't going to be going back, that seemed like extra data storage that wasn't needed. I think it's okay reconstructing from the created report, the data should all be there - ~~it needs moving into Peterborough.pm to be the reverse of the waste_munge_bulky_data there, ideally, but wanted to get it up for review first.~~

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -7,6 +7,7 @@ BEGIN { extends 'FixMyStreet::App::Controller::Form' }
 use utf8;
 use Lingua::EN::Inflect qw( NUMWORDS );
 use List::Util qw(any);
+use FixMyStreet::App::Form::Field::JSON;
 use FixMyStreet::App::Form::Waste::UPRN;
 use FixMyStreet::App::Form::Waste::AboutYou;
 use FixMyStreet::App::Form::Waste::Request::Bromley;
@@ -212,6 +213,31 @@ sub pay : Path('pay') : Args(0) {
         $c->stash->{template} = 'waste/cc.html';
         $c->detach;
     }
+}
+
+# redirect from cc processing - bulky goods only at present
+sub pay_cancel : Local : Args(2) {
+    my ($self, $c, $id, $token) = @_;
+
+    my $property_id = $c->get_param('property_id');
+
+    $c->forward('check_payment_redirect_id', [ $id, $token ]);
+
+    $c->forward('/auth/get_csrf_token');
+
+    my $p = $c->stash->{report};
+    my $saved_data = $c->cobrand->waste_reconstruct_bulky_data($p);
+    $saved_data->{name} = $p->name;
+    $saved_data->{email} = $p->user->email;
+    $saved_data->{phone} = $p->user->phone;
+    $saved_data->{resident} = 'Yes';
+
+    my $saved_data_field = FixMyStreet::App::Form::Field::JSON->new(name => 'saved_data');
+    $saved_data = $saved_data_field->deflate_json($saved_data);
+    $c->set_param(saved_data => $saved_data);
+    $c->set_param(goto => 'summary');
+    $c->set_param(process => '');
+    $c->go('bulky', [ $property_id ], []);
 }
 
 # redirect from cc processing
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Roles/SCP.pm
```diff
@@ -27,6 +25,14 @@ sub waste_cc_get_redirect_url {
     $p->set_extra_metadata('redirect_id', $redirect_id);
     $p->update;
 
+    my $backUrl;
+    if ($back eq 'bulky') {
+        # Need to pass through property ID as not sure how to work it out once we're back
+        $backUrl = $c->uri_for('pay_cancel', $p->id, $redirect_id ) . '?property_id=' . $c->stash->{property}{id};
```
davea, line 31: SCP wants the param to be URL encoded otherwise it complains about an invalid request. Guess the space in the postcode is throwing it off.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -214,6 +214,44 @@ sub pay : Path('pay') : Args(0) {
...
+    my @fields = grep { $_->{name} =~ /ITEM_/ } @{$p->get_extra_fields};
+    foreach (@fields) {
+        my ($id) = $_->{name} =~ /ITEM_(\d+)/;
+        $saved_data->{"item_" . ($id+0)} = {
+            item => $_->{value},
+            #photo_fileid => '???',
+        };
+    }
+
+    my $saved_data_field = FixMyStreet::App::Form::Field::JSON->new(name => 'saved_data');
```
davea, line 247: This needs `use`ing at the top or it crashes when you come back to the page.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -214,6 +214,44 @@ sub pay : Path('pay') : Args(0) {
...
+    my $saved_data = {
+        "chosen_date" => $p->get_extra_field_value('DATE'),
+        "location" => $p->get_extra_field_value('CREW NOTES'),
+        #"location_photo_fileid" => '???',
+        "name" => $p->name,
+        "email" => $p->user->email,
+        "phone" => $p->user->phone,
+        "resident" => "Yes",
+    };
+    my @fields = grep { $_->{name} =~ /ITEM_/ } @{$p->get_extra_fields};
```
davea, line 238: Should probably ignore any with an empty `value` otherwise the template lists them all (template also expects `cobrand.bulky_items_maximum`, not sure why that's not working):

image

davea (changes requested):
A few small changes needed.

perllib/FixMyStreet/App/Controller/Waste.pm
dracos: Ah this overlaps with my other PR for fixing the JS photo thing, which will change this so that this will work okay because it won't need the hasher of item_NUM any more.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -222,28 +222,15 @@ sub pay_cancel : Local : Args(2) {
...
-            item => $_->{value},
-            #photo_fileid => '???',
-        };
-    }
+    my $p = $c->stash->{report};
+    my $saved_data = $c->cobrand->waste_reconstruct_bulky_data($p);
+    $saved_data->{name} = $p->name;
+    $saved_data->{email} = $p->user->email;
+    $saved_data->{phone} = $p->user->phone;
+    $saved_data->{resident} = 'Yes';
```
davea, line 233: Just curious why these 4 didn't go into `waste_reconstruct_bulky_data`?

davea (approved):
💰 🔙

perllib/FixMyStreet/App/Controller/Waste.pm
davea: (oh unless it's because there's nothing cobrand-specific about them?)

dracos: Yeah, thought might as well split it like that.

PR #4202, opened 2022-12-02
[Bromley] Adjust codes for adding feature_id

MorayMySoc:
Change to make all light codes add feature_id to the category

https://mysocietysupport.freshdesk.com/a/tickets/2569

[skip changelog]


dracos (approved):
Looks like that should sort it, thanks :)

PR #4203, opened 2022-12-02
[Bulky Goods] Add page for viewing existing booking

davea:
For https://github.com/mysociety/societyworks/issues/3373

(will sort out photo storage in separate PR)

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -26,6 +26,7 @@ use FixMyStreet::App::Form::Waste::Garden::Kingston::Renew;
 use Open311::GetServiceRequestUpdates;
 use Digest::MD5 qw(md5_hex);
 use Memcached;
+use JSON::MaybeXS;
 
 has feature => (
     is => 'ro',
@@ -973,7 +974,7 @@ sub bulky_setup : Chained('property') : PathPart('') : CaptureArgs(0) {
     my ($self, $c) = @_;
 
     if (  !$c->stash->{waste_features}{bulky_enabled}
-        || $c->stash->{property}{has_pending_bulky_collection} )
+        || $c->stash->{property}{pending_bulky_collection} )
     {
         $c->res->redirect( '/waste/' . $c->stash->{property}{id} );
         $c->detach;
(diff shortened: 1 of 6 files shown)


dracos (changes requested):
Looks good (as you say in comment, can probably combine bits with the payment back PR for tidiness). However, is there a reason the view booking page can't be instead of the /report/ID page - do we need two 'views' on the same 'report'? This would also then answer my next problem, which is that the confirmation/reminder emails (and other places?) with links to cancel, I think they should probably go to the relevant view booking page, which I assume would be easier if it was to `/report/ID` as we don't have the postcode to construct the relevant page for this PR. If there's a reason we can't do that (let me know), I think we'll have to store the `property_id` with the report (I think this is straightforward, setting `property_id` as extra data on the category).

davea:
🤦🏻‍♂️I don't know why I didn't think of that. Great idea, will update accordingly.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -1130,6 +1144,25 @@ sub waste_munge_bulky_data {
     # XXX what about photos?
 }
 
+sub waste_reconstruct_bulky_data {
```
dracos, line 1147: This should hopefully disappear with a rebase now.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -514,9 +514,23 @@ sub has_pending_bulky_collection {
             state =>
                 { '=', [ FixMyStreet::DB::Result::Problem->open_states ] },
         },
-    )->all;
+    );
+}
+
+sub bulky_can_view_collection {
```
dracos, line 520: Do we need this function at all now, or is the Report.pm code doing it for us already, it already has all the 'private = reporter or staff only' checks in.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -998,6 +997,11 @@ sub bulky : Chained('bulky_setup') : Args(0) {
     $c->stash->{first_page} = 'intro';
     $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Bulky';
 
+    if ( $c->stash->{property}{pending_bulky_collection} ) {
```
dracos, line 1000: Guess this can go back where it was too, though guess it doesn't matter.

dracos (changes requested):
Looks good, though I think the user checking can be dropped now? Perhaps with a comment around bulky_view explaining it's being called as a report page so someone doesn't add it in future by mistake.

perllib/FixMyStreet/Cobrand/Peterborough.pm
davea: It's still being called from the `bin_days_bulky.html` template so still needed. Don't think it needs to be called from within `bulky_view` though, you're right.

dracos (approved):
Woo, nearly there :) (overall, not this PR ;) )

PR #4204, opened 2022-12-05
[Camden] Add service_code to update params

chrismytton:
This is needed for the Symology code in open311-adapter.

[skip changelog]


MorayMySoc (approved):
Ok - looks fine.

PR #4205, opened 2022-12-05
[Bulky Goods] Store/display photos on collection summary

davea:
[skip changelog]


dracos (approved):
Nice, and works for the view booking page too, huzzah. Do we know what is meant to happen with these photos backend-wise?

PR #4207, opened 2022-12-05
[Bulky Goods] Add per-item maximums.

dracos:
For https://github.com/mysociety/societyworks/issues/3374 [skip changelog]
This adds the admin interface, some text to the items page describing any maximums if present, and prevents submission if items selected are over these maximums. It does not do any client side updating to e.g. disallow an item from being selected once the maximum number have been selected because that seemed like it would be a non-trivial amount of work for not really that much gain, given if someone does do it they get the error immediately and quickly and can resolve it, it'd only be a small improvement and I'm not entirely sure how to have it work with the autocompletes etc.


davea (approved):
Looks good.

Longer term it might be worth changing this to only display the max value when the thing is selected, much like we do for the message value. Can imagine if they added maximums to a lot of item types it'd be annoying to have to scroll past a big info box every time. But, not for now.

PR #4208, opened 2022-12-05
[Sutton] Prevent time of bin collection showing on collection day

MorayMySoc:
Request to suppress time of bin collection on collection day

https://mysocietysupport.freshdesk.com/a/tickets/2577

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandSLWP.pm
+++ b/perllib/FixMyStreet/Roles/CobrandSLWP.pm
@@ -194,6 +194,10 @@ sub waste_on_the_day_criteria {
     # No reports pre-6pm, completed or not
     if ($now->hour < 18) {
         $row->{report_allowed} = 0;
+        if ($self->council_url eq 'sutton' && $row->{last} ) {
+            # Prevent showing collected time until reporting is allowed
+            $row->{last}{completed} = 0;
+        }
     }
 }
 
--- a/t/app/controller/waste_sutton_r.t
+++ b/t/app/controller/waste_sutton_r.t
@@ -92,11 +92,12 @@ FixMyStreet::override_config {
         set_fixed_time('2022-09-09T16:30:00Z');
         $mech->get_ok('/waste/12345');
         $mech->content_like(qr/Friday, 9th September\s+\(this collection has been adjusted from its usual time\)\s+\(In progress\)/);
-        $mech->content_like(qr/Friday, 9th September, at  4:00pm/);
+        $mech->content_lacks(', at  4:00pm');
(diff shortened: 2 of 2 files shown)


chrismytton (approved):

templates/web/base/waste/bin_days.html
```diff
@@ -79,7 +79,9 @@ <h3 class="govuk-heading-l govuk-warning-text__heading">Your subscription is soo
           <dt class="govuk-summary-list__key">Last collection</dt>
           <dd class="govuk-summary-list__value">
               [% date.format(unit.last.date) | replace('~~~', unit.last.ordinal) %]
-              [%~ IF unit.last.completed %], at [% date.format(unit.last.completed, '%l:%M%p') | lower ~%][% END %]
+              [%~ UNLESS (c.cobrand.moniker == 'sutton' AND unit.last.date.ymd == date.format(date.now, "%Y-%m-%d") AND date.format(date.now, "%H") < 18 ) %]
```
dracos, line 82: This is okay, but it feels a bit off having this logic in the template. There already is a `waste_on_the_day_criteria` function that is called for special on-the-day circumstances, and already includes an hour < 18 check for allowing reporting. So clearing row->{last}{completed} at that point would I think be more consistent.

dracos (changes requested):
Works fine, but I think the logic would be better put somewhere else.

templates/web/base/waste/bin_days.html
MorayMySoc: I tried a `next->method` call from `Sutton.pm` which resulted in an error `No next::method 'waste_on_the_day_criteria` - is that because CobrandSWLP.pm is a role and not a class? If so, am I best to replicate the method in Sutton with the change you suggest, or caveat the CobrandSWLP.pm method with a sutton clause?

dracos: CobrandSLWP is for Kingston and Sutton and already has a few small bits that differ between them, I think it's fine to change it in there, yep.

dracos: I might be changing this again soon anyway as I think Kingston has different behaviour, so can always move it to separate calls if that turns out to be the case.

dracos: The row->{last} has just been `delete`d above the code you've added, so you're recreating it with only a completed entry at this point and so the template code that assumes other things exist if `unit.last` exists are throwing that warning (the eval is the template's generated code). You need to only set completed to 0 if row->{last} exists as well.

dracos (approved):
Thanks

PR #4210, opened 2022-12-06
Camden fron end v5

lucascumsille:
Hi @chrismytton =)
I have implemented the latest feedback from Camden. Let me know if you need me to change anything.

Screenshot 2022-12-06 at 09 23 47
Screenshot 2022-12-06 at 09 23 54
Screenshot 2022-12-06 at 09 24 16

Before submitting this pull request. @bekkileaver after our chat here
https://github.com/mysociety/societyworks/issues/3355
I think they client hasn't seen the admin page(@jacquelinebylau  could you confirm this?), so I could revert the special input styles for the admin pages so they look like our default FMS. After your research it seems unnecessary to override our current styles. However in the likely scenario the client ask to keep the consistency between the citizen user and the admin side, can we share your research with them to persuade them not to move forward with it?

Fixes part of:
https://3.basecamp.com/4020879/buckets/27695641/todos/5470577758#__recording_5592648455
Except the alert page feedback, according to this guideline: https://github.com/mysociety/societyworks/issues/3107

### DIFF
--- a/templates/web/camden/main_nav_items.html
+++ b/templates/web/camden/main_nav_items.html
@@ -2,6 +2,8 @@
 
 [%~ IF c.user_exists ~%]
     [%~ INCLUDE navitem uri='/my' label=loc('Your account') ~%]
+[%~ ELSE ~%]
+    [%~ INCLUDE navitem uri='/auth' label=loc('Sign in') ~%]
 [%~ END ~%]
 
 [%~ IF c.user_exists AND c.user.has_body_permission_to('planned_reports') ~%]
--- a/web/cobrands/camden/_mixins.scss
+++ b/web/cobrands/camden/_mixins.scss
@@ -115,10 +115,12 @@
 
 @mixin one-line-input-height {
     height: $input-height-small-screen;
+    line-height: $input-height-small-screen;
 
     @if variable-exists(large-view-point) {
         @media only screen and (min-width: $large-view-point) {
             height: $input-height-large-screen;
+            line-height: $input-height-large-screen;
         }
     }
 }
--- a/web/cobrands/camden/_mixins.scss
+++ b/web/cobrands/camden/_mixins.scss
@@ -45,8 +45,8 @@
     border-radius: $input-border-radius;
     box-shadow: none;
     font-family: $body-font;
-    padding-left: $input-padding-x;
-    padding-right: $input-padding-x;
+    // padding-left: $input-padding-x;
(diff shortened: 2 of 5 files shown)


lucascumsille:
@chrismytton 
In the latest [commit](https://github.com/mysociety/fixmystreet/pull/4210/commits/fd5b4dad6c23b35019ac5946e99d5d2a83b6e0b4) I think I solved the issues of this cobrand not having a sign in button in the navbar. Can you let me know if there is any other place where I should change(I can't think gf any other place).

Fixes: https://mysociety.slack.com/archives/C03DWPYHS4S/p1669977568541399

jacquelinebylau:
@lucascumsille The client has access to the admin pages and I know that they have been looking at it.

lucascumsille:
> @lucascumsille The client has access to the admin pages and I know that they have been looking at it.

Thanks for confirming that, Jackie. Then if they have seen it I'll send this PR, if @bekkileaver wants me to revert the inputs to the default style we can make it in a later commit.

chrismytton:
Thanks @lucascumsille!

I've just noticed that there are still some checkboxes in the admin which aren't quite displaying correctly.

This is on the edit user page:

image

Are you able to fix that as part of this change?

lucascumsille:
> Thanks @lucascumsille!
> 
> I've just noticed that there are still some checkboxes in the admin which aren't quite displaying correctly.
> 
> This is on the edit user page:
> 
> image
> 
> Are you able to fix that as part of this change?

Hey @chrismytton No probs I'll take a look now and try sending a commit within an hour.

dracos:
I thought we were reverting the change to all checkboxes (for now) and only leaving the ones they'd asked for on the user side?

lucascumsille:
> I thought we were reverting the change to all checkboxes (for now) and only leaving the ones they'd asked for on the user side? Given the stars and now this, I want to avoid others cropping up and don't want this being fixed one-at-a-time when without it is working okay.

Perfect I think it's the best approach. I'll revert the changes for the admin pages, no much benefit from it according to Bekki. I'll send a commit with that then.

lucascumsille:
Hi @chrismytton I just pushed a commit to only apply the input styles in the user-citizen interface.
Let me know if you have any comments

Screenshot 2022-12-08 at 07 34 27

Screenshot 2022-12-08 at 07 34 08

chrismytton (approved):
Awesome, thanks @lucascumsille! ✨

PR #4217, opened 2022-12-09
[FMS] Stop cobrand removing stopper of another on mobile

MorayMySoc:
Adds a class to the mobile alert message that an asset has not been found based on the id of the alert message. That way a cobrand will only remove its own messages when triggered through 'not_found'.

Adds a check when enabling the report form too if there are any mobile asset not found messages.

https://github.com/mysociety/societyworks/issues/3269

[skip changelog]


dracos (approved):
Looks good, thanks :)

PR #4218, opened 2022-12-09
[Southwark] Add Confirm set up & assets

davea:


perllib/FixMyStreet/Cobrand/Southwark.pm
```diff
@@ -30,4 +32,168 @@ sub disambiguate_location {
...
+            }
+        }
+    }
+
+    return values %contacts_hash;
+}
+
+sub allow_anonymous_reports { 'button' }
+
+sub anonymous_account {
```
dracos, line 187: Since you wrote this probably, the default code now handles this, so this function shouldn't be needed any more.

perllib/FixMyStreet/Cobrand/Southwark.pm
```diff
@@ -30,4 +32,168 @@ sub disambiguate_location {
...
+        srsname => "urn:ogc:def:crs:EPSG::27700",
+        typename => "Estates",
+        filter => "<Filter><Contains><PropertyName>Geometry</PropertyName><gml:Point><gml:coordinates>$x,$y</gml:coordinates></gml:Point></Contains></Filter>",
+    };
+
+    my $features = $self->_fetch_features($cfg, $x, $y);
+    return $features->[0];
+}
+
+# Southwark do not want certain TfL categories to appear, dependent on
```
dracos, line 131: Thanks for the existing docs, would be good to have docs for the other new functionality being added.

dracos (changes requested):
Looks good, one function can be removed and bit more docs would be good :)

dracos (approved):
Thanks :)

PR #4220, opened 2022-12-12
fix panning action causing zoom due to defunct touch input

### LINKED ISSUE #4132: Map bug when attempting to pan the map while pressing on a report pin
**Describe the bug**

The map can enter a state where:

- Panning touch movements (**not** pinch/zoom movements) can result in the map being zoomed in our out.
- Existing reports can no longer be opened. (Instead, FixMyStreet offers to create a new report when touching a pin.)

**To Reproduce**

1. Search for a location where there are map pins
2. Touch a pin and immediately (while pressing down on the pin) pan the map
3. A preview of the report shows up
4. The map may now (it doesn't always happen) have entered the state described above

**Expected behavior**

The map should continue to work like it did before opening the report preview. Panning movements should pan the map (not zoom it). Additionally, touching existing map pins should show a preview of the report in question (as opposed to offer to create a new report).

**Screenshots**

Here's a video:

https://fixamingata.se/static/fixmystreet-map-problem.webm

Please not that only panning movements (no pinch movements) were performed after the report preview was opened. Despite this, the map is zoomed in an odd way.

 - Device/OS: Emulated Pixel 3a
 - OS: Android, API 32
 - Browser: Chrome
 - Version: 91.0.4472.114

kwokcal:
potential fix for #4132

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -104,7 +104,8 @@ $.extend(fixmystreet.utils, {
     var drag = {
         activate: function() {
             this._drag = new OpenLayers.Control.DragFeatureFMS( fixmystreet.markers, {
-                onComplete: function(feature, e) {
+                //onSuccess ensures that any failure does not get read in as a usable action
+		onSuccess: function(feature, e) {
                     var geom = feature.geometry,
                         lonlat = new OpenLayers.LonLat(geom.x, geom.y);
                     fixmystreet.display.begin_report(lonlat, { noPan: true });


jonkri:
Thank you for this change! We will try it out!

Spontanously, the map behaviour seems a bit better with this change. However, I can still manage to enter a state where attempting to pan the map zooms it in or out.

dracos:
OL2 does not have an onSuccess function, as far as I'm aware, so what this does is remove the onComplete function. This means it is no longer possible to drag the new report pin in order to move the new report somewhere else (you can drag the pin, but it won't update any relevant information, which will still be the old location). So I'm afraid I don't think this change can fix the issue.

PR #4222, opened 2022-12-13
[Brent] Change report prompt and send title as attribute to open311

MorayMySoc:
Report prompt should ask for a location based summary of the issue and then send the title as an attribute so symology endpoint can transfer the attribute into the report location field

https://github.com/mysociety/societyworks/issues/3407229

[skip changelog]

### DIFF
--- a/t/cobrand/brent.t
+++ b/t/cobrand/brent.t
@@ -77,7 +77,7 @@ for my $test (
     };
 };
 
-subtest "UnitID on gully sent across in detail" => sub {
+subtest "Open311 attribute changes" => sub {
     my ($problem) = $mech->create_problems_for_body(1, $brent->id, 'Gully', {
         areas => "2488", category => 'Gully grid missing', cobrand => 'brent',
     });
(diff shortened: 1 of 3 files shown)


t/cobrand/brent.t
```diff
@@ -102,4 +102,26 @@ subtest "UnitID on gully sent across in detail" => sub {
     };
 };
 
+subtest "Report title passed through as attribute" => sub {
```
dracos, line 105: This is fine, though I probably would have only added the `is $c->param('attribute[title]')` line to the test above :)

dracos (approved):
Looks good :)

t/cobrand/brent.t
```diff
@@ -102,4 +102,26 @@ subtest "UnitID on gully sent across in detail" => sub {
...
+        COBRAND_FEATURES => {
+            anonymous_account => {
+                brent => 'anonymous'
+            },
+        },
+    }, sub {
+        FixMyStreet::Script::Reports::send();
+        my $req = Open311->test_req_used;
+        my $c = CGI::Simple->new($req->content);
+        is $c->param('attribute[title]'), 'Gully Test 1 for 1', 'Title passed as attribute for Open311';
```
dracos, line 123: Ah you have used a fixed ID here, that won't work, you'll need to make sure it's the actual $problem->id

dracos (changes requested):

dracos (approved):
That should do it :)

PR #4223, opened 2022-12-13
Allow cookie-less API calls from iOS app

davea:
Works around iOS bug where users login info is forgotten each time they launch the app.

[skip changelog]

For https://github.com/mysociety/fixmystreet-mobile/issues/300

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1014,8 +1014,15 @@ sub process_user : Private {
         my $user = $c->user->obj;
         $report->user( $user );
         $report->name( $user->name );
-        $c->stash->{check_name} = 1;
-        $c->stash->{login_success} = 1;
+
+        # iOS app can't send cookies, so instead sends
+        # user's email/password when making a new report.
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1014,8 +1014,15 @@ sub process_user : Private {
         my $user = $c->user->obj;
         $report->user( $user );
         $report->name( $user->name );
-        $c->stash->{check_name} = 1;
-        $c->stash->{login_success} = 1;
+
+        # iOS app can't send cookies, so instead sends
+        # user's email/password when making a new report.
+        # In that case we can't check the user's name so
```
dracos, line 1020: I see what this is for, but I would like to check the name/anonymous if possible I think. How about if the mobile app sends a '2' if logged in, but remains a 1 if submit_sign_in clicked, and then when it's 2 it doesn't do check_name/login_success then. Does that work okay?

dracos (commented):
Looks good, have asked a question about if it's possible to maintain current behaviour as well. Could do with a test as well.

dracos (approved):

PR #4225, opened 2022-12-13
Peterborough bulky goods: post-staging fixes

nephila-nacrea:
Fixes issues that have been raised by the client themselves or spotted by devs.

These include:

- [ ] Fixing an issue that meant a select element with a single item was showing as empty
- [ ] Preventing a legacy report with no property ID from fetching all postcodes (thus causing a timeout)
- [ ] Collection summary page improvements:
  - [ ] Cancel button actually links to cancellation journey
  - [ ] Show message when collection cancelled
  - [ ] Hide cancellation messaging & button when collection cancelled
  - [ ] DIsplay correct cancellation cutoff date
- [ ] Only allowing admin to view cancellation reports (and adding a link to such on the collection summary page)
- [ ] Displaying the correct pricing on the bin days page

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/My.pm
+++ b/perllib/FixMyStreet/App/Controller/My.pm
@@ -140,9 +140,12 @@ sub get_problems : Private {
         "$table.state" => [ keys %$states ],
     };
 
+    # We do not want to show bulky goods cancellation reports
+    $params->{"$table.category"}{'!='} = 'Bulky cancel';
+
     my $categories = [ $c->get_param_list('filter_category', 1) ];
     if ( @$categories ) {
-        $params->{"$table.category"} = $categories;
+        $params->{"$table.category"}{'='} = $categories;
         $c->stash->{filter_category} = { map { $_ => 1 } @$categories };
     }
 
@@ -191,6 +194,7 @@ sub setup_page_data : Private {
     my $table = $c->action eq 'my/planned' ? 'report' : 'me';
     my @categories = $c->stash->{problems_rs}->search({
         "$table.state" => [ FixMyStreet::DB::Result::Problem->visible_states() ],
+        "$table.category" => { '!=', 'Bulky cancel' },
     }, {
         join => 'contact',
         columns => [ "$table.category", 'contact.extra', 'contact.category' ],
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -1310,21 +1310,48 @@ sub waste_reconstruct_bulky_data {
     return $saved_data;
 }
 
+sub bulky_free_collection_available {
+    my $self = shift;
+    my $c = $self->{c};
+
+    my $cfg = $self->wasteworks_config;
+
+    my $attributes = $c->stash->{property}->{attributes};
+    my $free_collection_available = !$attributes->{'FREE BULKY USED'};
+
+    return $cfg->{free_mode} && $free_collection_available;
+}
+
+# For displaying before user books collection. In the case of individually
+# priced items, we cannot know what the total cost will be, so we return the
+# lowest cost.
+sub bulky_minimum_cost {
+    my $self = shift;
+
+    my $cfg = $self->wasteworks_config;
+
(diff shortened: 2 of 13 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -545,6 +545,23 @@ sub bulky_can_view_collection {
     return $c->user->id == $p->user_id;
 }
 
+sub bulky_can_view_cancellation {
```
dracos, line 548: This function is duplicated above, so needs combining, not overwriting.

nephila-nacrea: I have made some changes so it is no longer a duplicate - Dave & I came to the agreement that non-admin users do not need to view a cancellation report (though if Peterborough decide otherwise, it should be easy enough to change).

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -189,9 +189,19 @@ sub load_problem_or_display_error : Private {
         my $from_email = $c->sessionid && $c->flash->{alert_to_reporter} && $c->flash->{alert_to_reporter} == $problem->id;
 
         my $allowed = 0;
-        $allowed = 1 if $from_email;
-        $allowed = 1 if $c->user_exists && $c->user->id == $problem->user->id;
-        $allowed = 1 if $permissions->{report_inspect} || $permissions->{report_mark_private};
+        if ( $problem->category eq 'Bulky cancel' ) {
```
davea, line 192: Making this change in the core FMS code doesn't feel quite right somehow, but I'm struggling to think of an alternative... Does the user who made the original booking ever actually see the cancellation ID or a link to the report in question at all? If not, then there's presumably no way they'd end up on this page?

nephila-nacrea: Hmm it is actually shown on their 'Your Account' page
Screenshot 2023-01-06 at 09 41 30

but if they click through they get an 'Access Denied' message.
Would it be better to hide the link there, rather than have the check I've added here?

perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -140,12 +140,18 @@ sub get_problems : Private {
...
     my $categories = [ $c->get_param_list('filter_category', 1) ];
     if ( @$categories ) {
-        $params->{"$table.category"} = $categories;
+        $params->{"$table.category"}{'='} = $categories;
         $c->stash->{filter_category} = { map { $_ => 1 } @$categories };
     }
 
+    # Standard users should not be shown bulky goods cancellation reports
+    $params->{"$table.category"}{'!='} = 'Bulky cancel'
+        if !$c->cobrand->admin_allow_user( $c->user );
```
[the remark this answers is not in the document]
nephila-nacrea: That's a good point.
I don't know if at some point we should have a place where e.g. admin can view a list of cancellations specifically, but it's not worth worrying about now.

davea (approved):
👍

web/js/waste.js
```diff
@@ -92,6 +92,8 @@ $(function() {
         var valueAttribute = $this.find('option').filter(':selected').data('extra');
         valueAttribute = valueAttribute ? valueAttribute.message : '';
         if (valueAttribute) {
+            // Make sure utf8 characters display correctly
+            valueAttribute = decodeURIComponent(escape(valueAttribute));
```
dracos, line 96: escape is not valid modern JavaScript - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape - what is this line attempting to do/fix, what was the issue with utf8 characters?

nephila-nacrea: Iirc, on the item selection page, utf8 chars (like ☃, as an extreme case) were showing up as garbled, despite looking fine elsewhere.

dracos: Hmm, the issue needs sorting at source, then, because we shouldn't require a line like this, everything should be encoded correctly already.
image
I can take a look.

nephila-nacrea: Ah, thank you 👍

PR #4228, opened 2022-12-15
[Brent] Add a working bin page with report missed collection

MorayMySoc:
Adds the framework for a working bin page with reporting a missed collection.

Will need services, events and correct bin images adding.

'Request a new container' and 'Subscribe to' are not implemented.

https://github.com/mysociety/societyworks/issues/3403

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -6,6 +6,7 @@ use warnings;
 use Moo;
 with 'FixMyStreet::Roles::Open311Multi';
 with 'FixMyStreet::Roles::CobrandOpenUSRN';
+with 'FixMyStreet::Roles::CobrandEcho';
 
 sub council_area_id { return 2488; }
 sub council_area { return 'Brent'; }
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -104,4 +105,272 @@ sub user_from_oidc {
...
+
+    $self->{c}->stash->{garden_subs} = $self->waste_subscription_types;
+
+    my $result = $self->{api_serviceunits};
+    return [] unless @$result;
+
+    my $events = $self->_parse_events($self->{api_events});
+    $self->{c}->stash->{open_service_requests} = $events->{enquiry};
+
+    # If there is an open Garden subscription (2106) event, assume
```
dracos, line 160: Bromley specific number again, so note

dracos (changes requested):
Looks good; think would be good to note IDs copied from elsewhere to ease when we do get those IDs, and template can be simplified a bit (until it gets complicated again :) ).

dracos (approved):
Looks good :)

PR #4230, opened 2022-12-16
[Brent] Added map pin files

lucascumsille:
This PR includes the following colours for the map pins:
[Brent-map-pins.zip](https://github.com/mysociety/fixmystreet/files/10245297/Brent-map-pins.zip)

Screenshot 2022-12-16 at 11 14 28

I wasn't sure if we needed the 6 colours, so I split the commits into two:

### First commit: Provided by client
- Green
- Yellow
- Red
- Grey 

### Second commit: Not Provided by client
I got them from their style guide
- Orange
- Blue

Let me know if I need to drop the second commit or if any of the other pins colours won't be used.

Also first time handling this type of request, so not sure if I need to something else.


lucascumsille:
@jacquelinebylau I added the map pins. Let me know if I need to change anything.

PR #4238, opened 2022-12-22
[Bulky Goods] Couple of small fixes

davea:
- Wire up 'sign in' button on bin days page for viewing existing bulky booking
 - Ensure item message is displayed even if item name has an `&` in it

[skip changelog]

### DIFF
--- a/templates/web/peterborough/waste/bin_days_bulky.html
+++ b/templates/web/peterborough/waste/bin_days_bulky.html
@@ -182,7 +182,7 @@
           [% END %]
           [% UNLESS c.user_exists %]
             <!-- #07 Should be displayed when: user HASN'T signed in -->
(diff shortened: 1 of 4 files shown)


templates/web/peterborough/waste/bin_days_bulky.html
```diff
@@ -182,7 +182,7 @@ <h3 id="bulky-waste" class="govuk-heading-m waste-service-name">Bulky Waste</h3>
           [% END %]
           [% UNLESS c.user_exists %]
             <!-- #07 Should be displayed when: user HASN'T signed in -->
```
nephila-nacrea, line 184: Should we be removing these `<!--`-style comments?

nephila-nacrea (approved):

PR #4243, opened 2023-01-05
Add reporter/staff update_disallowed options

dracos:
Adds new options to updates_disallowed for reporter/staff to both leave updates (sure this has come up before, but never implemented as far as I can see). Fixes FD-2619 [skip changelog]


MorayMySoc (approved):
Looks good

PR #4246, opened 2023-01-06
[Brent] Change login to match Westminster

MorayMySoc:
Client has requested that we should follow the style and wording that Westminster is using for signing in with the single sign on

[skip changelog]


dracos (approved):
Is fine (and matches Westminster as you say), I think it looks like the general.html and form_user.html templates could be refactored, there's a lot of shared info there, but not for now, perhaps something to ticket for later look.

PR #4256, opened 2023-01-12
Move 3 more asset layers to config, and related

dracos:
As well as d3bfd8c83 moving 3 more layers to config (6d14009c2 shows the replacement config, for checking), this PR also improves Cypress by moving out some config (f1a4b743c624), tries to deal with some flaky tests (7c9d030b) and prevents tests from calling out to tilma (95fd817); and also stops a JS error on the /my page (as part of my trying to deal with flakiness, though not used in end). [skip changelog]


perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -22,6 +22,11 @@ Offline pages Catalyst Controller - service worker handling
 
 sub service_worker : Path("/service-worker.js") {
     my ($self, $c) = @_;
+    if (FixMyStreet->test_mode == 2) {
```
chrismytton, line 25: I think this code might be clearer/more future-proof if there was a `FixMyStreet->browser_testing` (or `->cypress_testing`/something else) method that did these checks.

chrismytton (changes requested):
Looks good! Just a couple of comments on the new `test_mode` that's been introduced.

perllib/FixMyStreet/App/Controller/Offline.pm
dracos: I've changed it to `FixMyStreet->test_mode eq 'cypress'`

bin/browser-tests
```diff
@@ -241,7 +241,7 @@ sub run {
     } else {
         # Child, run the server on port 3001
         require FixMyStreet;
-        FixMyStreet->test_mode(1); # So email doesn't try to send
+        FixMyStreet->test_mode(2); # So email doesn't try to send
```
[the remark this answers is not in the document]
dracos: Thanks, have updated the docs.

PR #4257, opened 2023-01-16
[Bulky] Added front-end fixes for Bulky waste

lucascumsille:
Fixes part of https://github.com/mysociety/societyworks/issues/3425

### Changes included in this PR

- [x] [Pboro] I have erased the styling for the next and back button, because we are not longer using them.
- [x] [Summary page]Tweak the number of items column. At the moment it could be confusing for user to see a number 2 right next to the second item. Some people could actually see it as the number of items been requested.
- [x] [Summary page] Added some extra space between the summary and T&C
- [x] [Summary page] "No image attached for this item" message now is visible for every user.
- [x] [Summary page] [Backend needed] Added a link to each section so users don't have to press back plenty of times if they need to go back to the firsts steps. @davea let me know if adding the links it's possible otherwise I can drop this commit.

Screenshot 2023-01-16 at 09 23 59

### Pending
In the "Location details" page I think we should wrap the instructions using the `govuk-warning-text` with the icon so it doesn't get lost, specially for visually impaired users. It seems like and important message that needs some special attention. I tried changing it, but I realise it's part the `bulky.pm` file.
Screenshot 2023-01-12 at 11 20 40

Let me know if there is any feedback

### DIFF
--- a/templates/web/base/waste/bulky/summary.html
+++ b/templates/web/base/waste/bulky/summary.html
@@ -122,7 +122,7 @@
                   [% IF item.photo %]
                     <img class="img-preview is--small" alt="Preview image successfully attached" src="/photo/temp.[% item.photo %]">
                   [% ELSE %]
-                    <span class="govuk-text__assistive">No image attached for this item</span>
+                    <span>No image attached for this item</span>
                   [% END %]
               </td>
           </tr>
--- a/templates/web/base/waste/bulky/summary.html
+++ b/templates/web/base/waste/bulky/summary.html
@@ -33,7 +33,7 @@
     </div>
   [% END %]
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-5">
       [% IF problem %]
         [% cancellation_report = cobrand.bulky_cancellation_report(problem) %]
         [% IF cancellation_report %]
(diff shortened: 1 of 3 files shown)


lucascumsille:
Regarding inconsistent styling of submit button:
I've been trying to check where the problem is, not sure if the team is referring  to:

- Primary button been different when submitting a picture vs the "continue" button.
- The styling of `govuk-button` is different to the `primary-button`. 
- All the above.

Either case not sure if applying the styling of the primary-button to the govuk-button is an option. It seems the most straightforward. This raises the following question:

- Is it something we should across all cobrands or only Pboro or any cobrand with WW.
- If applying to any cobrands is that something we want to do now?

davea:
@lucascumsille I've added a commit that adds buttons for going back to previous pages in the process.

They're maybe a bit big (especially the "your details" one) - are there any better CSS classes we can use?

image


Don't think we can have a "change address" button because that results in going all the way back to the beginning and throwing away all the user's progress.

lucascumsille:
> @lucascumsille I've added a commit that adds buttons for going back to previous pages in the process.
> 
> They're maybe a bit big (especially the "your details" one) - are there any better CSS classes we can use?
> 
> image
> 
> Don't think we can have a "change address" button because that results in going all the way back to the beginning and throwing away all the user's progress.

haha @davea  I also try the secondary button, but unfortunately the buttons are quite big and they pop up a bit more than the content. I also try to put it at the bottom of each section, but yeah they are quite big.
I would probably try **one** of the following options:

- Using a link instead of a button, but adding an underline so it's easier for users to realise there is a link.
- Link instead of a button, but adding and edit icon + text link. This will make it pop up more, but without been too much.
- Like in bootstrap adding small-button class this probably should need to be implemented across all FMS or WW.

I think all the options above would work well, but some will require adding more code than others. Let me know if you have any preferences.

PR #4259, opened 2023-01-16
Add admin page to view COBRAND_FEATURES config.

dracos:
This adds a superuser-only page to the site listing the `COBRAND_FEATURES` configuration indexed by feature, plus a page per cobrand showing that cobrand's entries in that configuration.

Please check the following:

- [x] Has the code POD documentation been added or updated?

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin.pm
@@ -10,7 +10,7 @@ use FixMyStreet::SMS;
 
 =head1 NAME
 
-FixMyStreet::App::Controller::Admin- Catalyst Controller
+FixMyStreet::App::Controller::Admin - Controller for main admin pages
 
 =head1 DESCRIPTION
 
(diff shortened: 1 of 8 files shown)


davea (changes requested):
This is a neat idea and will make it much easier than digging through YAML! Couple of things though

- [x] The cobrand page (e.g. `admin/config/cobrand_features/oxfordshire`) errors here: `Caught exception in FixMyStreet::App::Controller::Admin->config_page_cobrand_one "Modification of a read-only value attempted at /[...]/fixmystreet/perllib/FixMyStreet/App/Controller/Admin.pm line 140."` - seems the same no matter which cobrand I try
- [ ] The page might flow better if it were a table/tables rather than lots of h3s - it's easy to get lost in the whitespace with no structure
- [x] The `Yes: [list of cobrands]` format is harder to parse when determining if a specific cobrand has a feature, IMO the `cobrand: 1` format is simpler. Compare e.g. `category_groups` vs `multiple_category_groups`:
  - image
  - image
 - The new page doesn't seem to be linked from anywhere?

dracos:
Made the one page more robust, include fallback values on it too, and stopped the special grouping of boolean values. Then changed to use table which I think is easier to read, yep, and added a link from main config page.

PR #4261, opened 2023-01-17
Cascade delete alert_sent and user_planned_reports.

dracos:
If an alert is deleted, automatically drop its alert_sent rows; if a report or a user are deleted, drop any relevant user_planned_reports rows. [skip changelog]


dracos:
Internally note the DB changes of this are already live.

davea (approved):
👍

PR #4266, opened 2023-01-19
Display photo changes in report moderation updates.

neprune:
closes https://github.com/mysociety/societyworks/issues/2585

Screenshot 2023-01-19 at 12 40 56

### DIFF
--- a/perllib/FixMyStreet/DB/Result/ModerationOriginalData.pm
+++ b/perllib/FixMyStreet/DB/Result/ModerationOriginalData.pm
@@ -139,26 +139,30 @@ sub compare_coords {
 sub compare_photo {
     my ($self, $other) = @_;
 
-    my $old = $self->photo || '';
-    my $new = $other->photo || '';
-    return '' if $old eq $new;
+    my $old = $self->photos;
+    my $new = $other->photos;
 
-    $old = [ split /,/, $old ];
-    $new = [ split /,/, $new ];
-
-    my $diff = Algorithm::Diff->new( $old, $new );
     my (@added, @deleted);
-    while ( $diff->Next ) {
-        next if $diff->Same;
-        push @deleted, $diff->Items(1);
-        push @added, $diff->Items(2);
+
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/DB/Result/ModerationOriginalData.pm
```diff
@@ -139,26 +139,35 @@ sub compare_coords {
...
 
-    my $diff = Algorithm::Diff->new( $old, $new );
+    my (%old_photo_ids_set, %new_photo_ids_set);
     my (@added, @deleted);
-    while ( $diff->Next ) {
-        next if $diff->Same;
-        push @deleted, $diff->Items(1);
-        push @added, $diff->Items(2);
+
+    foreach (@$old) {
```
dracos, line 148: All this code is fine, just some perl-isms you might want to know about to use here or in future:

A loop that is setting a hash based upon an array, like here, can be written along with its definition as:
    `my %old_photo_ids_set = map { $_->{id} => 1 } @$old;`
You'll see that used a few places in the code.

perllib/FixMyStreet/DB/Result/ModerationOriginalData.pm
```diff
@@ -139,26 +139,35 @@ sub compare_coords {
...
+        $old_photo_ids_set{$_->{id}} = 1;
+    }
+    foreach (@$new) {
+        $new_photo_ids_set{$_->{id}} = 1;
+        if (! exists $old_photo_ids_set{$_->{id}}) {
```
dracos, line 153: You don't need `exists` because the value could only be undefined or `1` so a standard comparison is fine.
Perl has an `unless` which you can use as the opposite of `if` if that makes sense to do so, and for one-liner if/unless statements you can put the contents first, e.g.:
`push @added, $_ unless $old_photo_ids_set{$_->{id}};`
Which can sometimes read more clearly.

dracos (changes requested):
Couple of small things is all :)

dracos (approved):
Looks good :)

PR #4272, opened 2023-01-30
[Bulky Goods] Munge photo data so it can be picked up by Bartec

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/3410.

Adds code that gathers together the various photos that may be uploaded during a bulky collection,  so that `process_photo_upload_or_cache` in App/Controller/Photo.pm can handle them (put them in an 'upload_fileid' stash field). This is because `save_user_and_report` in App/Controller/Report/New.pm expects 'upload_fileid' in the stash, in order to set 'photo' on the report (the 'photo' field is what is ultimately used to send data through Open311 to Bartec).

[skip changelog]


davea (approved):
🎉

PR #4273, opened 2023-01-31
Small tweaks to admin COBRAND_FEATURES page

davea:
Just a couple of little things that make it easier to navigate the pages that I've found useful.

 - Link to individual cobrand pages from each feature
 - Link back to main page from each feature key on cobrand page
 - Use sticky headers for table on main page

[skip changelog]

### DIFF
--- a/templates/web/base/admin/config_page_cobrand_one.html
+++ b/templates/web/base/admin/config_page_cobrand_one.html
@@ -10,7 +10,7 @@
 
   [% FOR feature IN config %]
     <tr>
-        <td>[% feature.key %]</td>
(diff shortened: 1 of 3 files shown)


templates/web/base/admin/config_page_cobrand_one.html
```diff
@@ -10,7 +10,7 @@
 
   [% FOR feature IN config %]
     <tr>
-        <td>[% feature.key %]</td>
+        <td><a href="/admin/config/cobrand_features#feature-[% feature.key %]">[% feature.key %]</a></td>
```
dracos, line 13: This could also be changed in the fallback section below.

PR #4274, opened 2023-01-31
[FMS] Added 20th anniversary badge in the frontpage header

lucascumsille:
Fixes part of: https://github.com/mysociety/Design-Internal-tasks/issues/8

Screenshot 2023-01-31 at 10 10 50
Screenshot 2023-01-31 at 10 01 47


web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -156,29 +156,69 @@ body.fullwidthpage {
...
             margin-top: 0; // no need for the vertical centering any more
         }
     }
 
+    #mysociety-badge {
+        top: calc((#{$fms-logo-height} + #{$fms-logo-vertical-margin}) * -1);
+        width: $mysociety-badge-width;
+        max-height: calc((#{$fms-logo-height} + #{$fms-logo-vertical-margin}) * 2);
+        svg {
+            width: $mysociety-badge-width;
```
zarino, line 188: You set a `width` for the SVG, but not a `height`, meaning that keeps the original 266px height from the inline `height="266"` attribute. As a result, the hitbox for the SVG’s wrapper link is much taller than the SVG itself, which is a bit weird:

Screenshot 2023-01-31 at 15 59 30

I’d set `height: auto` on the SVG. In fact, I’d probably leave all the sizing to `#mysociety-badge` instead, with something like this:

```scss
#mysociety-badge {
  width: $mysociety-badge-width;

  svg {
    width: 100%;
    height: auto;
  }
}
```

web/cobrands/fixmystreet.com/base.scss
```diff
@@ -49,6 +50,66 @@ svg|g.site-logo__svg {
     @include svg-background-image("/cobrands/fixmystreet/images/site-logo");
 }
 
+body.frontpage {
+  $fms-logo-width: 150px;
+  $fms-logo-height: 35px;
+  $mysociety-badge-width: 90px;
```
zarino, line 56: Maybe it’s just me, but I find this a smidge too large, it feels like it’s overpowering the FMS logo beside it:

Screen Shot 2023-01-31 at 16 09 10

Here’s how it looks reduced from 90px to 75px – a bit more balanced?

Screen Shot 2023-01-31 at 16 09 04

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -156,29 +156,69 @@ body.fullwidthpage {
 }
 
 body.frontpage {
+    $fms-logo-width: 200px;
```
zarino, line 159: Again, this feels very large to me, especially given we’ve reduced the size of the FMS logo at the same time.

web/cobrands/fixmystreet.com/layout.scss
```diff
@@ -156,29 +156,69 @@ body.fullwidthpage {
...
+            // width: 300px; Size without the mySociety badge
+            width: $fms-logo-width;
+            float: right;
+            height: $fms-logo-height;
             margin-top: 0; // no need for the vertical centering any more
         }
     }
 
+    #mysociety-badge {
+        top: calc((#{$fms-logo-height} + #{$fms-logo-vertical-margin}) * -1);
```
zarino, line 184: This all seems massively overcomplicated – do we _have_ to do all of this calculation? And why do we need to set a `max-height`?

zarino (changes requested):
Given we’re about to do a bunch of these, it feels like we should try and get some simple concepts down that we can apply the same everywhere.

I’ve left some comments inline, about sizing and CSS selectors.

I also feel like we could make our lives much easier if we removed the empty space from the top of the graphic itself (so the top of the "banner" touches the top of the SVG/PNG). Then there’d be less need for negative margins / positioning, maybe?

lucascumsille:
Thanks for the feedback @zarino 

- To get rid of the extra space on the top I vertically lowered the dropshadow on the badge so it doesn't use space on the top.
- I reduced the size in both, mobile and desktop.
- Some improvements to make to code less complicated.

Let me know what you think

Screenshot 2023-02-01 at 09 13 54
Screenshot 2023-02-01 at 09 14 10

zarino (commented):
Maybe it’s just me, but I feel like something about the spacing of this corner feels weird:

Screenshot 2023-02-06 at 12 40 22

Maybe it’s the massive vertical padding (well, `margin`) above/below the FMS logo compared with the narrow vertical padding inside the yellow hero section.

Maybe it’s the difference in size between the FMS logo (quite small – or rather, much smaller than it currently is on the live site) and the banner (quite large).

I get that you’ve tried to keep the FMS logo small, so that the nav menu to the right has the same amount of space as it currently has on the banner-less live site. But that does mean the FMS logo looks quite pokey next to this massive banner.

Hmmm… not sure. But something here just doesn’t feel right.

lucascumsille:
> Maybe it’s just me, but I feel like something about the spacing of this corner feels weird:
> 
> Screenshot 2023-02-06 at 12 40 22
> 
> Maybe it’s the massive vertical padding (well, `margin`) above/below the FMS logo compared with the narrow vertical padding inside the yellow hero section.
> 
> Maybe it’s the difference in size between the FMS logo (quite small – or rather, much smaller than it currently is on the live site) and the banner (quite large).
> 
> I get that you’ve tried to keep the FMS logo small, so that the nav menu to the right has the same amount of space as it currently has on the banner-less live site. But that does mean the FMS logo looks quite pokey next to this massive banner.
> 
> Hmmm… not sure. But something here just doesn’t feel right.

I get what you say, in this case I think I will increase the size of the FMS logo, because in the current mockup you can barely notice "mySociety" in the badge.

zarino:
Yeah, I’d deffo make the FMS logo _bigger_ rather than the badge _smaller_.

zarino (approved):
@lucascumsille sorry for the delay in revisiting this. I’ve pushed a number of commits which—if you’re happy with them—you’ll probably just want to squash into a single commit for all of these changes. I’ve:

- Removed the `badge-mysociety-*.png` image files and the `.mysociety-badge__fallback` element and styling, as it doesn’t seem we were actually _attempting_ to display the badge on non-homepage headers (at least, it wasn’t visible).
- Swapped the `badge_mysociety_*.svg` files with versions that have 1) 10px less empty space at the bottom, and 2) a more subtle shadow.
- Slightly bumped up the size of the FMS logo even further.
- Given the nav bar a margin-left, so that the nav items wrap onto two lines when there isn’t space (window widths around 750–850px), rather than overlapping on top of the FMS logo.
- Removed the animation and hover `scale()` effect on the badge and FMS logo (do we really _need_ hover effects on these?) but replicated the yellow focus `outline` on the nav items too, for consistency.

PR #4284, opened 2023-02-08
Set autocomplete=off on subcategories.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3060 some more; by setting autocomplete on the subcategory which I think is a simpler and better fix than page hide (which seemed to work most, but not all, of the time and I couldn't work out why). [skip changelog]


nephila-nacrea (approved):
I am tentatively approving this because it doesn't seem to break any behaviour that I have tested - though I am not able to reproduce the intermittent failures with page hide that you mention in the PR description.

Issue #4176, opened 2022-11-21
Crosshair way of positioning map pins


jonkri:
**Is your feature request related to a problem? Please describe.**

It can be difficult to position map pins using touchscreens, in particular with smaller devices like phones.

**Describe the solution you'd like**

The crosshair way of reporting (used in the Cordova app) is arguably better suited for touchscreens.

**Describe alternatives you've considered**

None.

**Additional context**

See the message [here](https://groups.google.com/a/mysociety.org/g/fixmystreet/c/-E8kZALgvBE/m/sGwCxaFUAgAJ).

dracos:
Hi @jonkri - yes, like my message you link to, I think I agree (Google Maps directions uses a crosshair method too on mobile, I see) - so we'll try and get some time in the near future to look at trying to implement this with some user testing.

jonkri:
@dracos: Hi! 👋 Thank you! 🤞

PR #4285, opened 2023-02-08
Use crosshairs for mobile reporting.

dracos:
Fixes https://github.com/mysociety/fixmystreet/issues/4176
image image image


davea (approved):
Woo, this works really well and makes the mobile experience much nicer.

One small improvement might be to improve the visibility of the crosshair. Currently it can be quite difficult to distinguish the center ╳ from the basemap.

Adding a 2px stroke makes it a bit easier on the eye:
Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-15 at 11 08 42

That said, I wonder if we should redo the crosshair asset entirely as an SVG? As you can see in the screenshot it's a bit blocky on retina screens (despite being an `@2x` asset). @lucascumsille what do you think?

dracos:
Yes, @lucascumsille if you'd like to improve/redesign the crosshair entirely, that'd be good :) And confirm you're happy with the new button or suggest change there.

lucascumsille:
@dracos This is really cool guys.
@davea  Not a problem I'll create a new crosshair svg format. Would you like me to change the colour so it looks more noticeable? I could thicken the strokes otherwise or just leave it as it is if we don't want to highlight it that much.

davea:
I added that 2px stroke just to see how it would look, the end result doesn't necessarily have to look like that! I'll leave it up to you :)

lucascumsille:
@dracos and @davea 
I made two versions:

- Green one I think it's the one the pops up the most, because the map itself already has black strokes and white sections. Not sure if this one will cause issues with clients that want their own colour branding, causing that it might be harder to maintain. 
- A black version. My reason to get rid of the outline was because the white didn't have enough contrast with white/pastel areas of the map.
- While writing this I thought of using the FMS yellow. Probably green would be still my favourite option.

Black version

Green version

Yellow version


[Crosshair.zip](https://github.com/mysociety/fixmystreet/files/10791687/Crosshair.zip)

Internal: Figma file https://www.figma.com/file/QUUQQicJ6dzV05QjzCnjei/FMS-Crosshair?node-id=0%3A1&t=PSqr1cJEK7FtHktA-1

dracos:
Think I might prefer Dave's still, sorry! The x in the middle with space around makes it easier to see the map directly around the location being picked, if you see what I mean. But rather than "tweaks", what would you do to make the clearest possible crosshair for locating a report? There's no reason it has to look like the current one if we can have something much better. A circle with lines seems 'popular' on https://www.iconfinder.com/search?q=crosshair - is that better? Or wider triangles?

lucascumsille:
No probs @dracos I'll take a look and send you another one during the morning =)

lucascumsille:
@dracos this file has the following icons:

Let me know if any of these ones work better, or they need tweaks like thinning the strokes.

Screenshot 2023-03-09 at 09 46 41

[Crosshair.zip](https://github.com/mysociety/fixmystreet/files/10930001/Crosshair.zip)

PR #4292, opened 2023-02-13
[Camden] Ignore live TfL River Piers categories

chrismytton:
On the staging site there's just one "River Piers" category, but on the live site there are quite a few. Luckily they all start with "River Piers", so switching to using `NOT LIKE` is enough to ignore the live categories.

Fixes https://github.com/mysociety/societyworks/issues/3500


MorayMySoc (approved):
Looks good

PR #4295, opened 2023-02-13
[Camden] Get email from preferred_username field

chrismytton:
Azure has been configured so that the users email address appears in the "preferred_username" field, so look for it there.

Fixes https://github.com/mysociety/societyworks/issues/3307


MorayMySoc (approved):
Looks good

PR #4296, opened 2023-02-13
[Brent] Add staff-only add assisted collection form.

neprune:


templates/web/base/waste/bin_days_sidebar.html
```diff
@@ -21,7 +21,7 @@ <h3>Download your collection schedule</h3>
     [% END %]
 
        [% TRY %][% PROCESS waste/_more_services_sidebar.html %][% CATCH file %]
-       [% IF any_report_allowed OR any_request_allowed OR show_garden_subscribe %]
+       [% IF any_report_allowed OR any_request_allowed OR show_garden_subscribe OR (domestic_collection AND (c.user.from_body OR c.user.is_superuser)) %]
```
dracos, line 24: This works fine - I think Brent is/will be the only thing using this default, it looks like all other existing Waste cobrands have created a custom `waste/_more_services_sidebar.html` template. So perhaps for consistency we should do the same here and not have a default, if they're all going to be slightly different.

t/cobrand/brent.t
```diff
@@ -515,6 +520,20 @@ FixMyStreet::override_config {
         $mech->content_like(qr#/waste/12345">Show upcoming#, "contains link to bin page");
         $new_report->delete;
     };
+
+    subtest 'test staff-only assisted collection form' => sub {
+        $mech->log_in_ok($staff_user->email);
+        my $resp = $mech->get_ok('/waste/12345/enquiry?category=Assisted+collection+add&service_id=262');
+        print $resp;
```
dracos, line 527: Debug left in

t/cobrand/brent.t
```diff
@@ -515,6 +520,20 @@ FixMyStreet::override_config {
         $mech->content_like(qr#/waste/12345">Show upcoming#, "contains link to bin page");
         $new_report->delete;
     };
+
+    subtest 'test staff-only assisted collection form' => sub {
+        $mech->log_in_ok($staff_user->email);
+        my $resp = $mech->get_ok('/waste/12345/enquiry?category=Assisted+collection+add&service_id=262');
```
dracos, line 526: Do we want to check the link to this page is shown on the bin days page first?

dracos (changes requested):
Looks good, a few small suggestions :)

dracos (approved):
Sorry, I force pushed brent-reviewed, hence why this PR now seems to have a couple of extra things in it! Changes all look good, thanks :)

PR #4299, opened 2023-02-16
[Brent] Add garden subscription cost

MorayMySoc:
From the technical specification add half price garden bin subscription in Oct/Nov/Dec

Adds cobrand messaging changes to the emails sent out.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -4,6 +4,7 @@ use parent 'FixMyStreet::Cobrand::UKCouncils';
 use strict;
 use warnings;
 use Moo;
+use DateTime;
 with 'FixMyStreet::Roles::Open311Multi';
 with 'FixMyStreet::Roles::CobrandOpenUSRN';
 with 'FixMyStreet::Roles::CobrandEcho';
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -612,7 +613,14 @@ sub garden_waste_cost_pa {
 
     $bin_count ||= 1;
 
-    return $self->feature('payment_gateway')->{ggw_cost} * $bin_count;
+    my $cost = $self->feature('payment_gateway')->{ggw_cost} * $bin_count;
+    my $now = DateTime->now( time_zone => FixMyStreet->local_time_zone );
+
+    if ($now->month =~ /^(10||11||12)$/ ) {
```
dracos, line 619: You only want single `|` here, otherwise the empty string could match (I realise it wouldn't)

t/app/controller/waste_brent_garden.t
```diff
@@ -236,63 +237,130 @@ FixMyStreet::override_config {
...
-        $mech->content_like(qr#/waste/12345">Show upcoming#, "contains link to bin page");
-
-        FixMyStreet::Script::Reports::send();
-        my @emails = $mech->get_email;
-        my $body = $mech->get_text_body_from_email($emails[1]);
-        like $body, qr/Number of bin subscriptions: 1/;
-        like $body, qr/Bins to be delivered: 1/;
-        like $body, qr/Total:.*?50.00/;
-    };
+    for my $test (
```
dracos, line 240: I guess it doesn't hurt to be comprehensive, but it does make the tests take a bit longer :-) I'd probably be happy with checking one in/one out of the 'zone', but no matter.

dracos (changes requested):
One small typo is all really, looks good :)

dracos (approved):
Think you've 'fixup'ed the wrong commit, but sure you can sort that out :)

PR #4303, opened 2023-02-23
[Bulky Goods] Disable time limit for refund, for soft launch

nephila-nacrea:
Allow refund as long as a collection can be cancelled.

For https://3.basecamp.com/4020879/buckets/26662378/todos/5870058641.

[skip changelog]


davea (approved):
🆗

PR #4304, opened 2023-02-24
[Bulky Goods] Sort item list on booking form alphabetically

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/26662378/todos/5821380701.

Removed the optgroup logic because this prevents the list of items being alphabetical under JS (JS users do not see option grouping; option grouping was there for the benefit of non-JS users, but the non-JS experience is already less than ideal so option grouping arguably doesn't add much).

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -216,24 +216,8 @@ has items_master_list => (
 );
 
 sub _build_items_master_list {
-    $_[0]->c->cobrand->call_hook('bulky_items_master_list');
-}
-
-has items_by_category => (
-    is      => 'ro',
-    isa     => 'HashRef',
-    lazy    => 1,
-    builder => '_build_items_by_category',
-);
-
-sub _build_items_by_category {
-    my $self = shift;
-
-    my %hash;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -216,24 +216,8 @@ has items_master_list => (
...
-
-sub _build_items_by_category {
-    my $self = shift;
-
-    my %hash;
-    for my $item ( @{ $self->items_master_list } ) {
-        push @{ $hash{ $item->{category} } }, $item->{name};
-    }
-    return \%hash;
+    [ sort { $a->{name} cmp $b->{name} }
```
davea, line 219: One tiny improvement would be to do a case-insensitive search (I noticed "e-Scooters" was at the bottom of my list):

```suggestion
    [ sort { lc($a->{name}) cmp lc($b->{name}) }
```

davea (approved):
Looks good! I've made one little suggestion, but please do merge after that 👍

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
nephila-nacrea: 👍

PR #4306, opened 2023-02-24
[FMS] Expand options on editing updates

MorayMySoc:


templates/web/base/admin/update_edit.html
```diff
@@ -46,7 +46,7 @@
 [% END -%]
 </li>
 [% IF update.problem_state %]
-<li>[% tprintf(loc('Update changed problem state to %s'), update.problem_state) %]</li>
+<li>[% tprintf(loc('Update changed problem state to: %s'), update.problem_state) %]</li>
```
dracos, line 49: I wouldn't change this text string as part of this, seems okay as is.
The line in the issue was referring to the fact that just because an update has `problem_state`, sadly that doesn't mean this update changed the state, because it could have just been a steady state entry from the last update.

templates/web/base/admin/update_edit.html
```diff
@@ -55,10 +55,42 @@
 </li>
 <li>[% loc('Cobrand:') %] [% update.cobrand %]</li>
 <li>[% loc('Cobrand data:') %] [% update.cobrand_data OR loc('None') %]</li>
+[% IF update.external_id~%]
+  <li>[% loc('External ID:') %] [% update.external_id %] </li>
```
dracos, line 59: To save on translation, note how "External ID" is already in the .po file, but "External ID:" is not. So if you had `[% loc('External ID') %]:` (which matches the other uses) it could share the translation.

templates/web/base/admin/update_edit.html
```diff
@@ -55,10 +55,42 @@
...
+  [%~ IF update.confirmed ~%]
+    [% PROCESS format_time time=update.confirmed %]</li>
+  [% ELSE ~%]
+    Not confirmed
+  [% END %]
 [% IF update.get_extra_metadata('external_status_code') %]
 <li>[% loc('External status code:') %] [% update.get_extra_metadata('external_status_code') | html %]</li>
 [% END %]
+<li>[% loc('Send state:') %] [% update.send_state %]</li>
+[% IF update.send_fail_count > 0 AND update.send_state != 'sent' AND update.send_state != 'skipped' %]
```
dracos, line 72: Note on reports, this send fail information is only shown to superusers, so presumably the same should apply here.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -267,15 +267,45 @@ sub update_edit : Path('update_edit') : Args(1) {
...
+        $update->update();
+        $c->forward( 'log_edit',
+            [ $update->id, 'update', 'send_state' ] );
+
+        $c->stash->{status_message} = _('Update will now be resent.');
+    }
+    elsif ( $c->get_param('mark_sent') ) {
+        $c->forward('/auth/check_csrf_token');
+
+        $update->send_state('sent');
```
dracos, line 283: This should also set `whensent` to match what happens with reports.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -267,15 +267,45 @@ sub update_edit : Path('update_edit') : Args(1) {
...
+        $update->update();
+        $c->forward( 'log_edit',
+            [ $update->id, 'update', 'send_state' ] );
```
dracos, line 286: This could be 'marked sent', again to match reports.

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -267,15 +267,45 @@ sub update_edit : Path('update_edit') : Args(1) {
...
     $c->forward('check_username_for_abuse', [ $update->user ] );
 
-    if ( $c->get_param('submit') ) {
+    if ( $c->get_param('resend') ) {
+        $c->forward('/auth/check_csrf_token');
+
+        $update->send_state('unprocessed');
+        $update->update();
+        $c->forward( 'log_edit',
+            [ $update->id, 'update', 'send_state' ] );
```
dracos, line 276: `resend` to match reports

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -267,15 +267,45 @@ sub update_edit : Path('update_edit') : Args(1) {
 
     $c->forward('check_username_for_abuse', [ $update->user ] );
 
-    if ( $c->get_param('submit') ) {
+    if ( $c->get_param('resend') ) {
+        $c->forward('/auth/check_csrf_token');
+
+        $update->send_state('unprocessed');
```
dracos, line 273: Also unset whensent

perllib/FixMyStreet/App/Controller/Admin.pm
```diff
@@ -267,15 +267,45 @@ sub update_edit : Path('update_edit') : Args(1) {
...
+
+        $c->stash->{status_message} = _('Update has been marked as sent.');
+    }
+    elsif ( $c->get_param('mark_skip') ) {
+        $c->forward('/auth/check_csrf_token');
+
+        $update->send_state('skipped');
+        $update->update();
+        $c->forward( 'log_edit',
+            [ $update->id, 'update', 'send_state' ] );
```
dracos, line 296: And then I guess 'marked skipped' to match

dracos (changes requested):
A few small things, mostly for consistency with what happens with reports.

templates/web/base/admin/update_edit.html
```diff
@@ -55,10 +55,42 @@
...
+  <li>[% loc('Send fail reason:') %] [% update.send_fail_reason %]</li>
+  <li>[% loc('Last send fail:') %] [% PROCESS format_time time=update.send_fail_timestamp %]</li>
+</ul>
+  [% IF update.send_state != 'sent' %]
+    <input type="submit" class="btn" name="mark_sent" value="[% loc('Mark as sent') %]">
+  [% END %]
+  [% IF update.send_state != 'skipped' %]
+    <input type="submit" class="btn" name="mark_skip" value="[% loc('Mark as skipped') %]">
+  [% END %]
+[% ELSIF update.send_state == 'sent' OR update.send_state == 'skipped' %]
```
dracos, line 84: I think it'd be better if whensent is always shown (if it's set), rather than going on the send state - eg I think ones pulled in via Open131 will be processed but have a whensent (and possibly others).

dracos (changes requested):
Changes all look good; sorry, one thing I just thought of.

dracos (approved):
Woo. I think the Cypress failure will be fixed by a rebase on top of current

PR #4307, opened 2023-02-27
Don't remove existing groups when editing a contact with category_groups unset.

neprune:
Partially fixes https://github.com/mysociety/societyworks/issues/3469.


templates/web/base/admin/bodies/contact-form.html
```diff
@@ -21,6 +21,11 @@
         </p>
       </label>
     </p>
+  [% ELSIF contact.extra.group %]
+        [% # Add existing groups as hidden fields so they are not unset when the contact is edited. %]
+        [% FOR group IN contact.extra.group %]
+        <input type="hidden" name="group" value="[% group | html %]">
```
dracos, line 27: Everything is filtered through `html` by default nowadays (since 6c2d3d5a7d84521d34daa2cf7e4be76a54b3b0e0) - things that still have it are left over from before that was the case, so there's no need to use it in new code.

dracos (approved):
Is fine as is, just one note on not needing the html filter by default :)

PR #4308, opened 2023-02-27
Add cobrand hook for validating category name and implement suffix test for National Highways.

neprune:
Partially fixes https://github.com/mysociety/societyworks/issues/3469.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -234,6 +234,14 @@ sub update_contact : Private {
 
     my $category = $self->trim( $c->get_param('category') );
     $errors{category} = _("Please choose a category") unless $category;
+
+    my $body = $c->model('DB::Body')->find({
+        id => $c->stash->{body_id},
+    });
+    my $cobrand = $body->get_cobrand_handler;
+    my $category_validation_error = $cobrand ? $cobrand->call_hook(admin_contact_validate_category => $category) : "";
+    $errors{category} = $category_validation_error if $category_validation_error;
+
     $errors{note} = _('Please enter a message') unless $c->get_param('note') || FixMyStreet->config('STAGING_SITE');
 
     my $contact = $c->model('DB::Contact')->find_or_new(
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -234,6 +234,14 @@ sub update_contact : Private {
 
     my $category = $self->trim( $c->get_param('category') );
     $errors{category} = _("Please choose a category") unless $category;
+
+    my $body = $c->model('DB::Body')->find({
+        id => $c->stash->{body_id},
+    });
+    my $cobrand = $body->get_cobrand_handler;
+    my $category_validation_error = $cobrand ? $cobrand->call_hook(validate_category => $category) : "";
+    $errors{category} = _($category_validation_error) unless !$category_validation_error;
```
dracos, line 243: A translation call like this won't work, because the thing that looks for strings to translate (which scans the files for wrapped strings) won't find the strings themselves. There's no need to call it here, you'd call it where the string was (if it needed translating, which I guess it doesn't here).

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -234,6 +234,14 @@ sub update_contact : Private {
 
     my $category = $self->trim( $c->get_param('category') );
     $errors{category} = _("Please choose a category") unless $category;
+
+    my $body = $c->model('DB::Body')->find({
+        id => $c->stash->{body_id},
+    });
+    my $cobrand = $body->get_cobrand_handler;
+    my $category_validation_error = $cobrand ? $cobrand->call_hook(validate_category => $category) : "";
```
dracos, line 242: I think 'validate_category' is a bit generic for a cobrand hook (in that is it to do with public reporting, this admin, some Open311 sending?) - perhaps something like `admin_contact_validate_category`?

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
dracos: I think `if $category_validation_error` slightly clearer than `unless !$category_validation_error` :)

dracos (changes requested):
This looks good - only a couple of tiny suggestions is all :)

dracos (approved):
Thanks :)

PR #4309, opened 2023-02-28
[Peterborough] Make bin not returned to collection point staff only

MorayMySoc:
Request from client support to make this category only usable by staff and to advise users to phone

https://mysocietysupport.freshdesk.com/a/tickets/2706

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -1563,6 +1563,11 @@ sub waste_munge_problem_data {
 sub waste_munge_problem_form_fields {
     my ($self, $field_list) = @_;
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -1589,7 +1591,8 @@ sub waste_munge_problem_form_fields {
         },
         497 => {
             container_name => "General",
-            label => "The bin wasn’t returned to the collection point",
+            label => "The bin wasn’t returned to the collection point (Please phone 01733 747474 to report this issue)",
```
davea, line 1594: Could this label change depending on `$not_staff`? Think you'd be able to drop the `report_new_munge_before_insert` then too maybe?

davea (approved):

PR #4311, opened 2023-02-28
[Brent] Asset actions for category on either park or road

MorayMySoc:
Request to have an asset category where the report can be made either on a park or a road.

Adds mechanic to check if selected layer is a park or a road and reports 'found - not_found' based on that.

Server config in `brent-3334-parks-and-highways-category`

https://3.basecamp.com/4020879/buckets/28996137/todos/5851140444

[skip changelog]


web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -197,6 +199,57 @@ fixmystreet.assets.brent.not_found = function(layer) {
...
+
+fixmystreet.assets.brent.found_filter = function(layer) {
+    if (fixmystreet.reporting.selectedCategory().category === 'Grass verges / shrub beds - littering') {
+        if (layer.features[0] && layer.features[0].type === 'Parks_and_Open_Spaces') {
+            fixmystreet.assets.brent.parkRoadIsFound.park = true;
+        }
+        if (layer.features[0] && layer.features[0].type === 'Highways') {
+            fixmystreet.assets.brent.parkRoadIsFound.road = true;
+        }
+        if (fixmystreet.assets.brent.parkRoadIsFound.isParkorRoad()) {
```
dracos, line 221: I don't think this `if` is needed, because one of the previous two ifs will have always been true (as this function is called by one of those two layers).

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -197,6 +199,57 @@ fixmystreet.assets.brent.not_found = function(layer) {
...
+        }
+    } else {
+        fixmystreet.message_controller.road_found(layer);
+    }
+};
+
+fixmystreet.assets.brent.not_found_filter = function(layer) {
+    if (fixmystreet.reporting.selectedCategory().category === 'Grass verges / shrub beds - littering') {
+        layer.fixmystreet.no_asset_msg_id = '#js-brent-parks-and-highways';
+        if (layer.features[0] && layer.features[0].type === 'Parks_and_Open_Spaces') {
```
dracos, line 232: Rather than having to check if there are any features, all the type checks could be replaced by `layer.fixmystreet.http_options.params.TYPENAME` which is always present.
(Or you could give the layer a name in config and check layer.name :) )

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -197,6 +199,57 @@ fixmystreet.assets.brent.not_found = function(layer) {
...
+        }
+        if (fixmystreet.assets.brent.parkRoadIsFound.isParkorRoad()) {
+            fixmystreet.message_controller.road_found(layer);
+        } else {
+            fixmystreet.message_controller.road_not_found(layer);
+        }
+    } else {
+        layer.fixmystreet.no_asset_msg_id = '#js-brent-parks-and-highways';
+        fixmystreet.message_controller.road_found(layer);
+        if (layer.features[0]) {
```
dracos, line 246: Again, no need to check features here, can use same check

dracos (changes requested):
A good solution, think some other cobrand code could probably change to be more like this. A few suggestions of improvements, hopefully nothing major :)

web/cobrands/fixmystreet-uk-councils/assets.js
dracos: After discussion on slack, just confirming I meant remove the `if` itself - so that the road_found is always called - not remove the whole if clause :)

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -197,6 +199,56 @@ fixmystreet.assets.brent.not_found = function(layer) {
...
+        if (layer.fixmystreet.http_options.params.TYPENAME === 'Parks_and_Open_Spaces') {
+            brent_parkRoadIsFound.park = true;
+        }
+        if (layer.fixmystreet.http_options.params.TYPENAME === 'Highways') {
+            brent_parkRoadIsFound.road = true;
+        }
+        fixmystreet.message_controller.road_found(layer);
+    } else {
+        fixmystreet.message_controller.road_found(layer);
+    }
```
dracos, line 224: Perhaps, given it's always going to be called:
```suggestion
    }
    fixmystreet.message_controller.road_found(layer);
```

PR #4314, opened 2023-03-01
[Brent] Added styling for .govuk-button--secondary button

lucascumsille:
This PR fixes:
https://github.com/mysociety/societyworks/issues/3516

I couldn't check if "Subscribe to Garden waste collection service" was solved, not if I need to add something to the cobrand in order to be able to access that content. But I think it used the same `.govuk-button--secondary` class than the "Submit a container request". 

Screenshot 2023-03-01 at 09 33 33

Let me know if there is anything I need to amend =)


MorayMySoc (approved):
Looks good! I'll merge it in - thanks

PR #4315, opened 2023-03-01
[Waste] Add renewal disabling flag

dracos:
[skip changelog] For FD-2744.
(And a tweak from Brent branch that might as well be on master now)


MorayMySoc (approved):
This seems to work - my only comment would be that it might be frustrating to have a warning in red that you're due for renewal and no way to renew

dracos:
Oh, I should have hidden the warning as well, where did I miss that?

PR #4318, opened 2023-03-02
[TFL] Added fix that prevents scroll bar to appear on navbar

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3132

They issue of the odd scroll bar appearing on the navbar was a problem only in Chrome, when I tried to replicate the problem on Safari and Firefox I couldn't see it. 

I added a fix that prevents the scroll bar to be displayed.

Screenshot 2023-03-02 at 08 26 15

### DIFF
--- a/web/cobrands/tfl/layout.scss
+++ b/web/cobrands/tfl/layout.scss
@@ -21,6 +21,7 @@ h1 {
 #main-nav {
   float: none;
   margin-top: $mappage-header-height;
+  overflow: hidden;
 
   .ie9 & > * {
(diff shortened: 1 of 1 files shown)


dracos (approved):
The issue is that the nav is shown at 1.125em which is 20.25, with line-height 1.5 = 30.375 + padding of 0.75em => 60.75px total, when it's constrained to 60px. As to why it doesn't always show, even in Chrome, can't work that out though! This solution seems fine.

PR #4320, opened 2023-03-02
[SLWP] Add support for dynamic costs based on date

chrismytton:
This means you can add new costs that will come into effect in the future in the config and it will automatically start using them at the appropriate time.

Fixes https://github.com/mysociety/societyworks/issues/3451

## General notes

- ~~At the moment this is just for the SLWP cobrands, I haven't made this change affect e.g. Bromley.~~ See below, have updated this so it works with all cobrands.
- All of the costs need a `start_date`, so need to make up a start date in the past for the current cost. See example configuration below.
- If the cost in the config is a numerical value rather than an array of values then this will be used as the cost (so will work without updating the config, if needed)

## Note to deployer

Will need to update the server config when deploying this to reflect the new prices.

```yaml
COBRAND_FEATURES:
...
  payment_gateway:
  ...
    kingston:
      ggw_sacks_cost:
        - start_date: 2022-01-01 00:00
          cost: 4100
        - start_date: 2023-03-06 00:00
          cost: 4300
      ggw_cost:
        - start_date: 2022-01-01 00:00
          cost: 8200
        - start_date: 2023-03-06 00:00
          cost: 8600
       ...
    sutton:
      ggw_sacks_cost:
        - start_date: 2022-01-01 00:00
          cost: 7052
        - start_date: 2023-04-01 00:00
          cost: 8052
      ggw_cost:
        - start_date: 2022-01-01 00:00
          cost: 7052
        - start_date: 2023-04-01 00:00
          cost: 8052
      ...
```

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandSLWP.pm
+++ b/perllib/FixMyStreet/Roles/CobrandSLWP.pm
@@ -788,8 +788,7 @@ sub waste_munge_report_data {
 # Same as full cost
 sub waste_get_pro_rata_cost {
     my ($self, $bins, $end) = @_;
-    my $cost = $bins * $self->feature('payment_gateway')->{ggw_cost};
-    return $cost;
+    return $self->garden_waste_cost_pa($bins);
 }
 
 sub waste_display_payment_method {
@@ -801,22 +800,6 @@ sub waste_display_payment_method {
     };
 
     return $display->{$method};
-}
-
-sub garden_waste_sacks_cost_pa {
-    my ($self) = @_;
-    my $cost = $self->feature('payment_gateway')->{ggw_sacks_cost};
-    return $cost;
-}
-
-sub garden_waste_cost_pa {
-    my ($self, $bin_count) = @_;
-
-    $bin_count ||= 1;
-
-    my $per_bin_cost = $self->feature('payment_gateway')->{ggw_cost};
-    my $cost = $per_bin_cost * $bin_count;
-    return $cost;
 }
 
 sub garden_waste_new_bin_admin_fee {
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1196,7 +1196,10 @@ sub garden_setup : Chained('property') : PathPart('') : CaptureArgs(0) {
 
     $c->detach('property_redirect') if $c->stash->{waste_features}->{garden_disabled};
 
-    $c->stash->{per_bin_cost} = $c->cobrand->feature('payment_gateway')->{ggw_cost};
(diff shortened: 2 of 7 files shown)


perllib/FixMyStreet/Roles/CobrandSLWP.pm
```diff
@@ -803,9 +802,30 @@ sub waste_display_payment_method {
     return $display->{$method};
 }
 
+sub _get_cost_from_array {
+    my ($self, $costs) = @_;
+
+    my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
+    my @sorted = sort { $b->{start_date} cmp $a->{start_date} } @$costs;
+    foreach my $cost (@sorted) {
+        my $start_date = DateTime::Format::Strptime->new(
```
dracos, line 811: Also to deal with your "need to provide a date in the past" comment, not that it matters either, but instead of the die you could return the last entry in @sorted assuming you'd want the earliest date.

dracos (changes requested):
Looks good, couple of tiny thoughts. I'd be happy for garden_waste_cost_pa to be in CobrandEcho.pm I think, that would then have it work for Bromley (and indeed Brent), as they're all the same function.

Could we also change the templates so e.g. templates/web/kingston/waste/garden/subscribe_intro.html and templates/web/sutton/waste/garden/subscribe_intro.html use the value from configuration, as they're currently hard-coded values there.

perllib/FixMyStreet/Roles/CobrandSLWP.pm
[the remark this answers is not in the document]
chrismytton: Nice, that's a bit simpler! Done in 99fe74e7b970dbe6c9717fe8aad9fce4a8f203d6.

chrismytton: Tried doing this but still needed a `start_date` for the `sort` comparison, so leaving this as-is for now in the interests of getting this live today 🙂

chrismytton:
> Looks good, couple of tiny thoughts. I'd be happy for garden_waste_cost_pa to be in CobrandEcho.pm I think, that would then have it work for Bromley (and indeed Brent), as they're all the same function.
> 
> Could we also change the templates so e.g. templates/web/kingston/waste/garden/subscribe_intro.html and templates/web/sutton/waste/garden/subscribe_intro.html use the value from configuration, as they're currently hard-coded values there.

Have addressed these in the fixup commits.

Have also switched to using midnight as the example time in the tests based on Lizetta's comment in https://github.com/mysociety/societyworks/issues/3451#issuecomment-1452258277.

dracos (approved):
Woo :)

PR #4322, opened 2023-03-02
[Brent] Add auto template triggering to Echo for Brent

MorayMySoc:
Auto responses for non-waste categories for Echo
need to be stripped down to just the resolution code and not to have commas in.

https://github.com/mysociety/societyworks/issues/3531
(linking to https://3.basecamp.com/4020879/buckets/28996137/todos/5870961914#__subscriptions)

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -132,6 +132,21 @@ sub open311_post_send {
 
 sub prevent_questionnaire_updating_status { 1 };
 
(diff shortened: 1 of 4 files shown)


dracos (approved):
This is fine, thanks. Future work would be moving `admin_templates_external_status_code_hook` to CobrandEcho.pm as they all share it (and having an "Echo" way of having that admin template appear rather than having to remember to add it) - and then having this Brent way be the way of doing the 'normal' external status code everywhere (removing superfluous commas from states in anything stored in the database) so it's more straightforward in future.

PR #4324, opened 2023-03-03
[Bulky Goods] Use Workpacks_Metrics_Get API call instead of Jobs_Get

davea:
The latter is much slower and returns more data than we actually need

[skip changelog]


dracos:
@davea is this still wanted/needed?

davea:
No, I think we can do without this. Technically this way is probably more "correct" but the existing implementation is fast enough.

PR #4328, opened 2023-03-07
[FMS][WW] Progressive web app has different start_urls

MorayMySoc:
Depending on whether a user installs the web app from FMS or from WW, app should have the correct starting url

https://github.com/mysociety/societyworks/issues/3408


templates/web/base/common_header_tags.html
```diff
@@ -31,7 +35,6 @@
     });
 }
 </script>
-
```
dracos, line 34: Not sure why this disappeared :)

templates/web/base/common_header_tags.html
```diff
@@ -1,7 +1,11 @@
 [% SET start = c.config.ADMIN_BASE_URL IF admin %]
 
 <meta http-equiv="content-type" content="text/html; charset=utf-8">
-<link rel="manifest" href="/.well-known/manifest.webmanifest">
+[% IF c.req.uri.path == '/waste' %]
```
dracos, line 4: I think we want this to be on any waste page, not just the 'home' one - so could this check if the path starts with /waste instead

perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -35,10 +35,17 @@ sub fallback : Local {
     my ($self, $c) = @_;
 }
 
-sub manifest: Path("/.well-known/manifest.webmanifest") {
-    my ($self, $c) = @_;
+sub manifest: Path("/.well-known/manifest.webmanifest") : Args(1) {
```
dracos, line 38: I think we want to maintain a recognised file extension (I agree it shouldn't matter, but MDN at https://developer.mozilla.org/en-US/docs/Web/Manifest does say "Browsers generally support manifests with other appropriate extensions" which might imply they don't support things without, though I assume they are fine) - so could this be /.well-known/manifest-fms.webmanifest' and manifest-waste perhaps? Might need two tiny controllers that then call a shared function for the bulk

dracos (changes requested):
Good start, couple of changes needed I think

templates/web/base/common_header_tags.html
```diff
@@ -1,10 +1,10 @@
 [% SET start = c.config.ADMIN_BASE_URL IF admin %]
 
 <meta http-equiv="content-type" content="text/html; charset=utf-8">
-[% IF c.req.uri.path == '/waste' %]
-    <link rel="manifest" href=".well-known/manifest.webmanifest/waste">
+[% IF c.req.uri.path.match('waste') %]
```
dracos, line 4: I'd put `^/` at the start to be on the safe side :)

perllib/FixMyStreet/App/Controller/Offline.pm
```diff
@@ -35,12 +35,27 @@ sub fallback : Local {
     my ($self, $c) = @_;
 }
 
-sub manifest: Path("/.well-known/manifest.webmanifest") : Args(1) {
-    my ($self, $c, $site) = @_;
+sub manifest_waste: Path('/.well-known/manifest-waste.webmanifest') {
+    my ($self, $c) = @_;
+
+    $c->stash->{manifest_application} = 'ww';
```
dracos, line 41: This works fine, but rather than adding more data to the stash, you could pass the data through directly, ie. `$c->forward('send_manifest', [ 'ww' ]);`, and then send_manifest first line would be `my ($self, $c, $app) = @_;`

dracos (changes requested):
Looks good, couple of tiny comments :)

dracos (approved):
🌟

PR #4331, opened 2023-03-08
ParallelAPI copes better with e.g. empty temp file.

dracos:
[skip changelog] Echo was down for a bit yesterday, leading to empty temp files being created. The code would fail to load in the file and then return no data, so either give a 404 (non-Peterborough) or get stuck for ever (Peterborough).


davea (approved):
Oops! Good catch.

PR #4334, opened 2023-03-10
[FMS] Fix issue with role id breaking csv report creation

MorayMySoc:
Raised in https://3.basecamp.com/4020879/buckets/28996137/todos/5921122970

[skip changelog]


dracos (approved):
Thanks!

PR #4339, opened 2023-03-15
[FMS] Add discount checkbox for garden waste

MorayMySoc:
If a cobrand implements the waste feature 'ggw_discount_pay_proportion' a checkbox will be available to allow discounting garden waste to staff.

The ggw_discount_pay_proportion is the percentage that the discounted payment will be - so for a 20% reduction it would be '80'


perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -157,6 +180,14 @@ has_field bins_wanted => (
     },
 );
 
+has_field apply_discount => (
+    type => 'Checkbox',
+    label => 'Customer discount',
+    option_label => FixMyStreet::Template::SafeString->new(
```
dracos, line 186: No harm in being this, just to note that the reason some other option_label use SafeString is because they have HTML that shouldn't be escaped in them, which isn't the case here. Should it perhaps give the % figure discount?

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -22,9 +22,18 @@ sub details_update_fields {
...
+        my $proportion_to_pay = $c->stash->{waste_features}->{ggw_discount_pay_proportion};
+        $c->stash->{cost_pa} = (($cost_pa / 100) / 100) * $proportion_to_pay;
+        $c->stash->{per_bin_cost} = ($c->stash->{per_bin_cost} / 100) * $proportion_to_pay;
+        $form->{c}->stash->{cost_now} = ( ( ($cost_now_admin + $cost_pa) / 100 ) * $proportion_to_pay ) / 100;
+        $form->{c}->stash->{cost_now_admin} = (($cost_now_admin / 100) / 100) * $proportion_to_pay;
+    } else {
+        $c->stash->{cost_pa} = $cost_pa / 100;
+        $form->{c}->stash->{cost_now} = ($cost_now_admin + $cost_pa) / 100;
+        $form->{c}->stash->{cost_now_admin} = $cost_now_admin / 100;
+    }
```
dracos, line 35: It feels like this would be clearer if you kept the existing code (what is now in the else), and then if apply_discount/waste_feature is true, apply the discount to all three then - as it is, you're calculating the first part in both places, if you see what I mean.
(Also could consistently use `$c` everywhere, I assume)

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -22,9 +22,18 @@ sub details_update_fields {
 
     my $cost_pa = $bin_count == 0 ? 0 : $form->{c}->cobrand->garden_waste_cost_pa($bin_count);
     my $cost_now_admin = $form->{c}->cobrand->garden_waste_new_bin_admin_fee($new_bins);
-    $form->{c}->stash->{cost_pa} = $cost_pa / 100;
-    $form->{c}->stash->{cost_now_admin} = $cost_now_admin / 100;
-    $form->{c}->stash->{cost_now} = ($cost_now_admin + $cost_pa) / 100;
+    if ($data->{apply_discount} && $c->stash->{waste_features}->{ggw_discount_pay_proportion}) {
+        my $proportion_to_pay = $c->stash->{waste_features}->{ggw_discount_pay_proportion};
+        $c->stash->{cost_pa} = (($cost_pa / 100) / 100) * $proportion_to_pay;
+        $c->stash->{per_bin_cost} = ($c->stash->{per_bin_cost} / 100) * $proportion_to_pay;
```
dracos, line 28: Should this also change per_new_bin_cost/per_new_bin_first_cost (I realise neither are used for Brent, but in case this then gets used somewhere else)

perllib/FixMyStreet/App/Form/Waste/Garden.pm
dracos: Also, this all needs doing in Garden/Modify and Garden/Renew as well, I assume - perhaps rather than duplicate in these three places, a cobrand hook to alter the prices in one location?
Renew might be trickier because it sets the prices on the intro page before it's shown, hmm.

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -76,6 +92,13 @@ has_page summary => (
         my $new_bins = $bin_count - $current_bins;
         my $cost_pa = $form->{c}->cobrand->garden_waste_cost_pa($bin_count);
         my $cost_now_admin = $form->{c}->cobrand->garden_waste_new_bin_admin_fee($new_bins);
+        if ($data->{apply_discount} && $form->{c}->stash->{waste_features}->{ggw_discount_pay_proportion}) {
+            my $proportion_to_pay = $form->{c}->stash->{waste_features}->{ggw_discount_pay_proportion};
+            $cost_pa = ($cost_pa / 100) * $proportion_to_pay;
+            $form->{c}->stash->{per_bin_cost} = ($form->{c}->stash->{per_bin_cost} / 100) * $proportion_to_pay;
+            $cost_now_admin = ($cost_now_admin / 100) * $proportion_to_pay;
+        }
```
dracos, line 100: Yes, something like this is more what I'd envisage also for the details_update_fields above - but all moved to a hook that could then be used here as well, presumably.

dracos: Similarly for modify/renew as well.

Also note, changes here don't actually affect what is then stored as the payment. Which is calculated again in Controller/Waste.pm in process_garden_data/modification/renew. I think if the payment is now going to depend upon more than just the bin count, we need to factor everywhere it's calculated together in one place/hook so we don't miss something.

dracos (changes requested):
A good start, but I think there's more work needed - if your test had seen what got stored on the report, you'd have seen it didn't apply the discount, plus modify/renewals.

dracos (changes requested):
Thanks, looks like /almost/ everything's covered :-) I think the only issue is on renew/modify if the discount is applied, and they change the number of bins, the JS is still going on the full price being passed in to the template and will show the wrong amount (but be okay when they submit to the summary page). Need to tweak the JS to look at the discount field and adjust the dynamic price shown accordingly. Does mean those two have different method from subscribe (where it works okay due to the intro page first) but think that's okay? Alternative would be to put the discount on its own 'page' on renew/modify if staff, which would then give you the answer for the next main page and can be treated same as subscribe then - which do you prefer?

I wonder if for tidiness at the usage, we could have the `apply_garden_waste_discount` look up the discount rather than having it passed in (it's always the same thing being passed in every time, looks like). Just makes the calling code bit smaller, what do you think? I've applied a fixup for you to take a look at for that.

MorayMySoc:
I'm happy for the function to call for the price as it does make the main code less verbose. I went from a specific function to a more generalised one and had a bit of a whim to make it pure, but it's not for a good reason (and must learn to use map more!)

I think an extra page might just be a bit of unnecessary user overhead if it can be avoided..? I'll have a look at the js and see if I can sort that out if that sounds ok to you.

web/js/waste.js
```diff
@@ -59,11 +74,17 @@ $(function() {
...
+      if ($('[name=apply_discount]')[0] && $('[name=apply_discount]')[0].checked) {
+         var charges = apply_discount(admin_fee, total_per_year, pro_rata_cost);
+         admin_fee = charges[0]; total_per_year = charges[1]; pro_rata_cost = charges[2];
+        }
       $('#cost_per_year').text(total_per_year.toFixed(2));
       $('#cost_now_admin').text(admin_fee.toFixed(2));
       $('#pro_rata_cost').text(pro_rata_cost.toFixed(2));
     }
     $('#modify #bins_wanted, #modify #current_bins').on('change', modify_cost);
+    $('[name=apply_discount]').on('change', modify_cost);
```
dracos, line 87: This needs a #modify prefix otherwise it will run on the renew page as well.

web/js/waste.js
```diff
@@ -35,6 +49,7 @@ $(function() {
     }
     $('#subscribe_details #bins_wanted, #subscribe_details #current_bins').on('change', bin_cost_new);
     $('#renew #bins_wanted, #renew #current_bins').on('change', bin_cost_new);
+    $('[name=apply_discount]').on('change', bin_cost_new);
```
dracos, line 52: Similarly, this needs a #renew prefix, otherwise both handlers will run.

dracos (changes requested):
One issue with the handler running twice is all, I think

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1622,6 +1628,7 @@ sub process_garden_data : Private {
         $data->{bin_count} = 1;
         $data->{new_bins} = 1;
         my $cost_pa = $c->cobrand->garden_waste_sacks_cost_pa();
+        $cost_pa = $c->cobrand->apply_garden_waste_discount($cost_pa) if $data->{apply_discount};
```
dracos, line 1631: This one is missing brackets round LHS so would be set to size of returned array rather than the cost, I think. One to fix before merging.

PR #4340, opened 2023-03-17
[fms.com][Shropshire] Show inline tips on update form if report was for Shropshire

davea:
image

For FD-2699

[skip changelog]


chrismytton (approved):
Thanks for picking this up! 🙏

PR #4342, opened 2023-03-17
[Waste] Ensure confirmation page always has button to get to property page

davea:
For FD-2747

[skip changelog]


davea:
The button does look odd inside the coloured box, not sure what I was thinking. Have moved it underneath, as well as fixed Sutton invisible text.

dracos (approved):

PR #4343, opened 2023-03-17
[Northumberland] Set-up Alloy integration.

neprune:
for https://github.com/mysociety/societyworks/issues/3514

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -4,6 +4,9 @@ use base 'FixMyStreet::Cobrand::UKCouncils';
 use strict;
 use warnings;
 
+use Moo;
+with 'FixMyStreet::Roles::Open311Alloy';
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -37,4 +40,27 @@ sub reopening_disallowed {
...
+        { name => 'description',
+          value => $row->detail },
+        { name => 'category',
+          value => $row->category },
+    ];
+
+    return $open311_only;
+}
+
+sub title_fetched_report {
```
chrismytton, line 60: I think it would be good to prefix this method with `open311_` to match the other hooks that are specific to open311 and to give a hint as to where this hook is called.

chrismytton (changes requested):
Code looks good! Just one suggestion around naming.

chrismytton (approved):
Looks good to me! 👍

PR #4347, opened 2023-03-21
[Shropshire] Prefix Confirm report description with note if map not used

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/3226.

[skip changelog]


perllib/FixMyStreet/Cobrand/Shropshire.pm
```diff
@@ -144,6 +144,35 @@ sub open311_config {
...
+    my $open311_only = $self->$orig($row, $h);
+
+    if (!$row->used_map) {
+        for (@$open311_only) {
+            if ($_->{name} eq 'description') {
+                my $search_string = $row->postcode || 'N/A';
+                my $prefix =<<"HERE";
+NOTE:
+Map was not used; location may not be accurate.
+Search string used: $search_string
```
dracos, line 165: I think I'd prefer if this line was only output if there was a search string.

dracos (changes requested):
Looks good :) Only change would be to have less output if no extra data available.

dracos (approved):

PR #4349, opened 2023-03-22
[FMS] Fix crash on CSV export if category had non-ASCII chars

davea:
The `$_->stat` line in in F::A::C::Dashboard::status was crashing because it was using an escaped/incorrectly encoded version of the filename. This change ensures that the filename never has non-ASCII characters.

Raised in FD-2801.

[skip changelog]


chrismytton (approved):
Are there other ASCII characters that could cause a problem here, like `*|<>` etc?

This certainly fixes the problem at hand, but I wonder if a longer term fix might be to remove all non-alphanumeric characters from the string?

PR #4350, opened 2023-03-23
Brent waste refuse message

dracos:
[skip changelog]
For https://3.basecamp.com/4020879/buckets/28996137/todos/5968609225
(and a fixup for quantity)


MorayMySoc (approved):
That seems to do the trick!

PR #4351, opened 2023-03-27
[Buckinghamshire] Alloy configuration

davea:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -15,6 +15,7 @@ use SUPER;
 use LWP::Simple;
 use URI;
 use Try::Tiny;
+use Utils;
 
 sub council_area_id { return 163793; }
 sub council_area { return 'Buckinghamshire'; }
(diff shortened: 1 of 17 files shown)


perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -110,6 +111,34 @@ sub send_questionnaires {
 
 sub open311_extra_data_exclude { [ 'road-placement' ] }
 
+sub open311_extra_data_include {
+    my ($self, $row, $h) = @_;
+
+    my $open311_only = [
```
dracos, line 117: Worth calling super here (or using around)? Doesn't really matter, unlikely we add a new central Alloy thing is it.

dracos (changes requested):
Nice additions to the cobrand features :) All looks good - would be great to have a test on the new look up site code stuff (think the current test mocks a bit widely, perhaps), and some docs on the new things :)

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
davea: `around` is a good shout 👍

dracos (approved):
Thanks, looks good :) Don't understand it saying some lines in Buckinghamshire.pm aren't covered, they must be surely as I'm reading it. Ah well!

PR #4353, opened 2023-03-28
Block /report/new URLs from search engines

davea:
Users were starting new reports at whatever latlon had appeared in their search results rather than choosing on the map or going via postcode form, meaning lots of reports at identical and incorrect locations.

For .com we should probably combine this with a Varnish rule that redirects users to `/` if they arrive at `/report/new` from Google/Bing/etc.

For FD-2794

[skip changelog]


dracos (approved):
🌑

PR #4362, opened 2023-03-30
Redirect to confirmation message after POST when making report

davea:
Fixes https://github.com/mysociety/societyworks/issues/2229


templates/web/base/tokens/confirm_problem.html
```diff
@@ -2,7 +2,13 @@
 
 <div class="confirmation-header">
 
-  <h1><a href="[% c.cobrand.relative_url_for_report( report ) %][% report.url %]">[% report.title %]</a></h1>
+  <h1><a href="[% c.cobrand.relative_url_for_report( report ) %][% report.url %]">
+    [% IF report.non_public AND NOT can_view_report %]
```
dracos, line 6: `[%~` so as to not have whitespace at the start of the link? (And similar throughout/elsewhere)

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +673,26 @@ sub map :Chained('id') :Args(0) {
     $c->res->body($image->{data});
 }
 
+sub confirmation :Chained('id') :Args(0) {
+    my ($self, $c) = @_;
+
+    my $template = $c->get_param('template') || "";
+
+    if ($template !~ /^(email_sent|confirm_problem)$/) {
+        $c->detach( '/page_error_404_not_found', [] );
```
dracos, line 682: This includes the report's location on the 404 page if you go to /report/ID/confirmation for any ID on any cobrand, would probably be better if it didn't, even if the report isn't non_public.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +673,26 @@ sub map :Chained('id') :Args(0) {
...
+    }
+
+    if ($template eq 'email_sent') {
+        $c->stash->{template} = 'email_sent.html';
+        $c->stash->{email_type} = 'problem';
+    } elsif ($template eq 'confirm_problem') {
+        $c->stash->{template} = 'tokens/confirm_problem.html';
+        $c->stash->{created_report} = $c->get_param('created_report');
+        $c->stash->{report} = $c->stash->{problem};
```
dracos, line 691: This lets me show a confirmation page for any report already in the system for any cobrand, e.g. this BANES report on the Bucks cobrand:
image
Whilst I see your checks for non_public reports, this feels like something confusing/bad could potentially happen at some point (and I don't know all the combinations of `owns_problem` on this page when it's given any report at all that it might not be expecting). More generally, I'd just be worried about the possibility of something being made public by this that wasn't expected.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -174,27 +177,19 @@ sub load_problem_or_display_error : Private {
     }
     elsif ( $problem->state eq 'unconfirmed' ) {
         $c->detach( '/page_error_404_not_found', [ _('Unknown problem ID') ] )
-            unless $c->cobrand->show_unconfirmed_reports ;
+            unless $c->cobrand->show_unconfirmed_reports
+            or $c->action eq 'report/confirmation';
```
dracos, line 181: I'm afraid due to this it appears to leak the details of every unconfirmed report in the system at /report/ID/confirmation?template=confirm_problem for any unconfirmed report.

Presumably this is here for the email_sent page? I think this change needs to be revoked, and email_sent treated as a separate code-path (perhaps email sent could be a basically static page, what would be lose, only the coordinates in "Report another problem here"?)

dracos (changes requested):
Good start, and I can see you've tried to deal with the non_public/unconfirmed issues, but I don't think this is going to work like this due to the leakage (actual and potential) of information. How about this, would this work - redirect to a /report/new/confirmation?token=[token] or /report/new/confirmation/[token] or whatever endpoint where token is generated as, I dunno, `hmac_sha1_hex("$time-$id", secret);` Then that endpoint will validate that hash and allow fetching of that report ID alone for 1 minute/30s since $time (similar to how Auth.pm does this for the CSRF token). And that can then show either template, wouldn't need to worry about state/non_public/etc, as it knows it's the report just created.

perllib/FixMyStreet/App/Controller/Report.pm
davea: There's no longer a report on the stash at this point so shouldn't happen any more.

perllib/FixMyStreet/App/Controller/Report.pm
davea: I've implemented the token system you suggested so this should be OK now.

davea:
Thanks for the suggestion of the token - that's made it much neater! I don't think any/all of the template changes are needed any more but the tests aren't happy if I just back them all out, so once I've got that sorted it'll be ready for another review.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +662,59 @@ sub map :Chained('id') :Args(0) {
...
+    my $token = $c->forward('/report/get_confirmation_token', [ $report ]);
+    unless ( $token eq $token_param ) {
+        $c->detach( '/page_error_404_not_found', [] );
+    }
+
+    # If the token is valid but expired then may as well be helpful and bounce
+    # the user to report page rather than 404.
+    # (NB the report may still be unconfirmed, but end result is the same - a 404)
+    my $cutoff = DateTime->now()->subtract( minutes => 1 )->epoch;
+    my $timestamp = $report->confirmed ? $report->confirmed : $report->created;
```
dracos, line 685: Can be `$report->confirmed || $report->created;`

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +662,59 @@ sub map :Chained('id') :Args(0) {
...
+    my $token_param = $c->get_param('token') || "";
+    my $token = $c->forward('/report/get_confirmation_token', [ $report ]);
+    unless ( $token eq $token_param ) {
+        $c->detach( '/page_error_404_not_found', [] );
+    }
+
+    # If the token is valid but expired then may as well be helpful and bounce
+    # the user to report page rather than 404.
+    # (NB the report may still be unconfirmed, but end result is the same - a 404)
+    my $cutoff = DateTime->now()->subtract( minutes => 1 )->epoch;
```
dracos, line 684: Don't think you need to cast to epoch here or couple of lines below, you should be able to directly do `if ($timestamp < $cutoff)`

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +662,59 @@ sub map :Chained('id') :Args(0) {
...
+    my $timestamp = $report->confirmed ? $report->confirmed : $report->created;
+    if ( $timestamp->epoch < $cutoff ) {
+        return $c->res->redirect($report->url);
+    }
+
+    # We're now confident the user is allowed to view this report so stick it on
+    # the stash and load up the correct template.
+    $c->stash->{problem} = $report;
```
dracos, line 692: What's this line for? Looks like New.pm only ever uses `{report}`. Is it the header navigation template - so has this not been working before? Just curious.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +662,59 @@ sub map :Chained('id') :Args(0) {
...
+        $c->stash->{template} = 'tokens/confirm_problem.html';
+        $c->stash->{created_report} = "loggedin";
+        $c->stash->{report} = $c->stash->{problem};
+    } else {
+        $c->stash->{template} = 'email_sent.html';
+        $c->stash->{email_type} = 'problem';
+    }
+}
+
+sub get_confirmation_token : Private {
```
dracos, line 703: Just wondered if this would be better as a method on Problem.pm - you don't need $c as you can call `FixMyStreet::DB->resultset('Secret')->get`.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1911,9 +1913,15 @@ sub redirect_or_confirm_creation : Private {
             $c->log->info($report->user->id . ' is an inspector - redirecting straight to report page for ' . $report->id);
             $c->res->redirect( $report->url );
         } else {
-            $c->log->info($report->user->id . ' was logged in, showing confirmation page for ' . $report->id);
+            $c->log->info($report->user->id . ' was logged in, redirecting to confirmation page for ' . $report->id);
             $c->stash->{created_report} = 'loggedin';
             $c->stash->{template} = 'tokens/confirm_problem.html';
+            unless ($no_redirect) {
+                my $redirect = $c->uri_for_action( '/report/confirmation', [ $report->id ] );
```
dracos, line 1920: Could factor this + copy below into another method on Problem to return a report's URI (would need to pass in $c this one though, I guess, for `uri_for_action`).

dracos (changes requested):
Sorry, looks good, just some possible refactoring suggestions for perfection ;)

perllib/FixMyStreet/App/Controller/Report.pm
davea: Yep, it's for `navigation/_report.html` and e.g. BANES and the `post_report_report_problem_link` feature.

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -660,6 +660,43 @@ sub map :Chained('id') :Args(0) {
...
+        $c->detach( '/page_error_404_not_found', [] );
+    }
+
+    # If the token is valid but expired then may as well be helpful and bounce
+    # the user to report page rather than 404.
+    # (NB the report may still be unconfirmed, but end result is the same - a 404)
+    my $cutoff = DateTime->now()->subtract( minutes => 1 );
+    my $timestamp = $report->confirmed || $report->created;
+    if ( $timestamp < $cutoff ) {
+        return $c->res->redirect($report->url);
```
dracos, line 684: There's an issue (not caused by the fixup which is fine, I didn't notice the test failure before, but can see them at https://github.com/mysociety/fixmystreet/actions/workflows/coverage.yml?query=branch%3Aredirect-after-report-creation ) with it sometimes, but not always, redirecting to /report/ID rather than confirmation on a report that isn't viewable (eg. two-tier for Hart in report_new.t) and so giving a 404. My only thought is because coverage tests run more slowly than normal, the minute cut-off has been (sometimes, not always) reached and so the thing is trying to redirect but because these are cross-domain reports, that doesn't work. If that's the case, replacing this with the same as the confirm problem template of `c.cobrand.relative_url_for_report( report )` joined to `report.url` should work.

dracos (changes requested):
Sorry for not spotting the test failure before, but hopefully have diagnosed it

perllib/FixMyStreet/App/Controller/Report.pm
davea: Oops, nice catch!

dracos (changes requested):
Sadly the same tests are still failing (only on coverage, so I still think it must be this part somehow? but your fix looks good. And in e.g. report_new.t surely it can't be taking a minute to create the report? But what else could it be) :-/ Guess we might need more debugging into if it's reaching that if statement, what URL it's constructed at that point, might have to push a yucky  `print STDERR "$cutoff $timestamp\n";` and things like that, I can't get it to happen locally at all.

davea:
Looks like it was indeed just taking a long time...! I increased [the token cutoff](https://github.com/mysociety/fixmystreet/pull/4362/commits/ee87a382e936c629b9b1f730538b0ae50f239dae) to 3 minutes and the coverage [tests all passed](https://github.com/mysociety/fixmystreet/actions/runs/7784139717/job/21224094127?pr=4362) that time.

davea:
Phew, OK, think we're there now...

dracos (approved):
Thanks for all that :)

PR #4364, opened 2023-04-03
[Bulky Goods] Remove item list from reminder emails

nephila-nacrea:
As per https://3.basecamp.com/4020879/buckets/26662378/todos/5961885813.

[skip changelog]

### DIFF
--- a/t/app/controller/waste_peterborough.t
+++ b/t/app/controller/waste_peterborough.t
@@ -1312,7 +1312,6 @@ FixMyStreet::override_config {
             if ($days) {
                 my $email = $mech->get_email->as_string;
                 like $email, qr/26 August/;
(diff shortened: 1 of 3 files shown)


t/app/controller/waste_peterborough.t
```diff
@@ -1312,7 +1312,6 @@ FixMyStreet::override_config {
             if ($days) {
                 my $email = $mech->get_email->as_string;
                 like $email, qr/26 August/;
-                like $email, qr/Wardrobe/;
```
davea, line 1315: Might be better as `unlike`, that way we're testing the absence of something rather than the absence of a test for something.

```suggestion
                unlike $email, qr/Wardrobe/;
```

PR #4365, opened 2023-04-03
[Brent][FMS] Add config for changing text on submit ggw payment

MorayMySoc:
Change text on button for Brent payment by adding config override for button text

https://3.basecamp.com/4020879/buckets/28996137/todos/5838639014#__recording_5996609233

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden.pm
@@ -127,7 +127,12 @@ has_page summary => (
         $data->{cost_pa} = $cost_pa / 100;
         $data->{display_total} = $total / 100;
 
-        return {};
+        my $button_text = 'Continue to payment';
+        my $features = $c->cobrand->feature('waste_features');
+        if ($c->stash->{is_staff} && $features->{text_for_waste_payment}) {
+            $button_text = $features->{text_for_waste_payment};
+        }
+        return { submit => { value => $button_text }};
     },
     finished => sub {
         return $_[0]->wizard_finished('process_garden_data');
@@ -247,7 +252,6 @@ has_field continue_review => (
 
 has_field submit => (
     type => 'Submit',
-    value => 'Continue to payment',
     element_attr => { class => 'govuk-button' },
     order => 999,
 );
--- a/t/app/controller/waste_brent_garden.t
+++ b/t/app/controller/waste_brent_garden.t
@@ -139,6 +139,7 @@ FixMyStreet::override_config {
             hmac_id => '1234',
             scpID => '1234',
         } },
+        waste_features => { brent => {text_for_waste_payment => 'Payment processed'} },
         anonymous_account => { brent => 'anonymous.customer' },
     },
 }, sub {
@@ -281,6 +282,7 @@ FixMyStreet::override_config {
             # external redirects make Test::WWW::Mechanize unhappy so clone
             # the mech for the redirect
             my $mech2 = $mech->clone;
+            $mech2->content_contains('Continue to payment', 'Waste features text_for_waste_payment not used for non-staff payment');
             $mech2->submit_form_ok({ with_fields => { tandc => 1 } });
 
             is $mech2->res->previous->code, 302, 'payments issues a redirect';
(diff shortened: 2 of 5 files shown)


perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -110,7 +110,11 @@ has_page summary => (
         $data->{cost_pa} = $cost_pa / 100;
         $data->{display_total} = $total / 100;
 
-        return {};
+        my $button_text = 'Continue to payment';
+        if ($form->{c}->cobrand->feature('waste_features') && $form->{c}->cobrand->feature('waste_features')->{text_for_waste_payment}) {
```
dracos, line 114: I think this doesn't need the first half? feature() will return undef and you can look up a key on that without warning (unless I've missed something).

dracos (changes requested):
I don't think this does what's requested, which is to change the text only for staff who will be getting the PAYE form on the next page? This changes it for everyone, including normal users?

perllib/FixMyStreet/App/Form/Waste/Garden.pm
MorayMySoc: Running test suite getting errors like '[error] Caught exception in FixMyStreet::App::Controller::Waste->form "Can't use an undefined value as a HASH reference at /home/moray/FixMyStreet/fixmystreet/perllib/FixMyStreet/App/Form/Waste/Garden/Renew.pm line 80."' so thought that was the issue

dracos: Fair enough, then!

dracos: Though I don't see why it wouldn't: ` perl -we'use strict; my $d; my $e = $d->{123};` doesn't error.

dracos: Ah, so it's okay if you put it in a variable first, but not if you try and use it directly.

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -251,6 +252,7 @@ has_field continue_review => (
 
 has_field submit => (
     type => 'Submit',
+    value => 'Continue to payment',
```
dracos: Is this needed, your code change doesn't seem to affect this always being set above?

perllib/FixMyStreet/App/Form/Waste/Garden.pm
```diff
@@ -128,8 +128,9 @@ has_page summary => (
         $data->{display_total} = $total / 100;
 
         my $button_text = 'Continue to payment';
-        if ($form->{c}->cobrand->feature('waste_features') && $form->{c}->cobrand->feature('waste_features')->{text_for_waste_payment}) {
-            $button_text = $form->{c}->cobrand->feature('waste_features')->{text_for_waste_payment};
+        my $features = $form->{c}->cobrand->feature('waste_features');
```
dracos, line 131: $c to be consistent with line below? :)

t/app/controller/waste_brent_garden.t
```diff
@@ -281,6 +282,7 @@ FixMyStreet::override_config {
             # external redirects make Test::WWW::Mechanize unhappy so clone
             # the mech for the redirect
             my $mech2 = $mech->clone;
+            $mech2->content_contains('Continue to payment', 'Waste features text_for_waste_payment used');
```
dracos, line 285: Description isn't right here no (but is testing right thing)

perllib/FixMyStreet/App/Form/Waste/Garden.pm
dracos: I don't understand, it was removed then added according to the git log but isn't present in the files view. But I think can be removed!

dracos (changes requested):
Couple of tiny changes only :) I wondered whether instead of a new feature it used whether $c->stash->{staff_payments_allowed} was 'paye', but this way is a bit more flexible, I guess.

dracos (approved):

PR #4369, opened 2023-04-04
[Brent] Add text change for subscribers

MorayMySoc:
Request for change

https://3.basecamp.com/4020879/buckets/28996137/todos/6013100915
[skip changelog]

### DIFF
--- a/templates/web/base/waste/garden/subscribe_confirm.html
+++ b/templates/web/base/waste/garden/subscribe_confirm.html
@@ -22,7 +22,9 @@
 We have sent a confirmation email (you may also receive a payment receipt separately) to [% report.user.email %].
 [% END %]
 </p>
-
+[%- IF c.cobrand.moniker == 'brent' -%]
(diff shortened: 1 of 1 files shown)


templates/web/base/waste/garden/subscribe_confirm.html
```diff
@@ -30,7 +30,7 @@ <h1 class="govuk-panel__title">
     Bins typically arrive within two weeks, but this may vary due to demand.
   [% ELSE %]
     [%- IF c.cobrand.moniker == 'brent' -%]
-    Changes to your subscription will show up within 24 hours.
+    Changes to your subscription will show up within 24 hours. For new subscribers, and existing subscribers who did not renew before the deadline (31st March 2023), it may take up to 14 working days for your first collection to take place.
```
dracos, line 33: This won't be shown to normal users signing up for a new subscription - they'll be getting the `new_subscription` block above as discussed. I think for safety sake probably best to put this as a separate paragraph above this paragraph? Then at least it's always shown.

dracos (changes requested):

PR #4372, opened 2023-04-05
Add some basic POD documentation to some cobrand files

nephila-nacrea:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
@@ -1,14 +1,63 @@
+=head1 NAME
+
+FixMyStreet::Cobrand::CentralBedfordshire - code specific to the Central Bedfordshire cobrand [incomplete]
+
+
+=head1 SYNOPSIS
+
+We integrate with Central Bedfordshire's Symology back end.
+
+=head1 DESCRIPTION
+
+=cut
+
 package FixMyStreet::Cobrand::CentralBedfordshire;
 use parent 'FixMyStreet::Cobrand::Whitelabel';
 
 use strict;
 use warnings;
 
+=head2 Defaults
+
+=over 4
+
+=cut
+
 sub council_area_id { 21070 }
 sub council_area { 'Central Bedfordshire' }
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/CheshireEast.pm
```diff
@@ -8,11 +20,71 @@ use Moo;
...
+
+sub filter_report_description { "" }
+
+=item * Uses the OSM geocoder.
+
+=cut
+
+sub get_geocoder { 'OSM' }
+
+=item * Map type is CheshireEast.
```
dracos, line 64: I'd put this one up one before  the map zoom as they are very linked (the zoom default is different because it's a totally different map type).

perllib/FixMyStreet/Cobrand/Westminster.pm
```diff
@@ -21,35 +39,65 @@ sub disambiguate_location {
...
+
+=item * We do not send questionnaires.
+
+=cut
+
+sub send_questionnaires { 0 }
+
+=item * We do not send alerts to report authors.
+
+=cut
```
dracos, line 76: You don't need cut then pod immediately after.

dracos (changes requested):
Thanks! Couple of tiny suggestions is all. I think I'd also (I was planning to do this on one I had started recently) put "incomplete" in all the line 3s (so the NAME content) so it's clearer on the index generated page that it's not complete documentation of everything (unless it is :) ).

dracos (approved):

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -1,15 +1,64 @@
+=head1 NAME
+
+[incomplete]
```
dracos, line 3: One tiny thing when you're rebasing - I'd put this within the line below in all the files, that's what I meant by in the content - so it'd read `FixMyStreet::Cobrand::CentralBedfordshire - code specific to the Central Bedfordshire cobrand [incomplete]` so that that appears on the index page as well.

PR #4373, opened 2023-04-06
Updates to menu items

sallybracegirdle:
I have updated the names of some of the menu items, removed the link to NoiseWorks and to the parks/estate pages - both of which are very out of date. I have removed the individual case studies sub menu, so that, instead, users can just click directly to the case studies page.  I have removed the features menu - this isn't necessary here.

Please check the following:

Thanks for contributing!


dracos:
Applied the same changes to docs/_layouts/wasteworks.html (which is same but with different sidebar, could probably refactor out in future if makes things easier), and merged as 3ae13306f

PR #4374, opened 2023-04-10
[Bathnes] New style for homepage

lucascumsille:
Bathnes council asked for a new style for their homepage and also to use "Arial" instead of the current one.
Fixes: https://github.com/mysociety/societyworks/issues/3581

Preview:
Screenshot 2023-04-10 at 06 10 56

 [skip changelog ] Have you updated the changelog? If this is not necessary, put square brackets around this: skip changelog

Let me know if there is any feedback

### DIFF
--- a/templates/web/bathnes/around/intro.html
+++ b/templates/web/bathnes/around/intro.html
@@ -1,2 +1,6 @@
-<h1>Report a problem<br />
-Roads, pavements, and parks</h1>
+<h1>Report a problem</h1>
+<ul>
+  <li>Continue here to report a problem on a street, park or green space</li>
+  <li><a href="https://beta.bathnes.gov.uk/report-emergency">Report an emergency</a></li>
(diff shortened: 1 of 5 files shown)


MorayMySoc (approved):
Looks good to me!

lucascumsille:
@MorayMySoc do you have any idea where the inline style for that message with odd alignment comes from.
When I inspect it's an inline style so trace it didn't work that well.

Screenshot 2023-04-11 at 10 49 12

MorayMySoc:
It's the text in the emergency message, which has it's own positioning in it by the look of it - is that the cause?

lucascumsille:
> It's the text in the emergency message, which has it's own positioning in it by the look of it - is that the cause?

It's the emergency message, but I don't get where that inline style comes from and why it is different to the rest of the messages above it.

Issue #4375, opened 2023-04-11
Record iOS/Android in service field when reporting via app store PWA


davea:
`Problem` has a [`service` field](https://github.com/mysociety/fixmystreet/blob/cedc5c7d5ba51c6ca5fc9a61ce7f5d246bd8739c/perllib/FixMyStreet/DB/Result/Problem.pm#L71) which is used to store the type of device used to make a report.

The fixmystreet-mobile app populated this by [sending the value](https://github.com/mysociety/fixmystreet-mobile/blob/f66516c71647590711746524290f6775a270c1ff/www/js/models/report.js#L127) of `device.platform` (either `iOS` or `Android`) when POSTing the report.

The new app, being a wrapped PWA, will currently store `mobile` as the `service` value for all reports regardless of what platform they're made on.

We should figure out a way to preserve the existing behaviour, as it's helpful for stats & debugging purposes.

The frontend JS code is the same on the wrapped PWA and the mobile website, there may be a way to determine what platform we're on there.

Alternatively there are some ways to distinguish requests from the wrapped PWA on the server side:

 - iOS:
   - start url is `/?pwa=ios`
   - User-Agent header is suffixed `iospwa`
 - Android:
   - start url is `/?pwa=android`
   - Initial requests made when starting the app (from its icon, or when e.g. following an emailed report confirmation link) have a Referer header of `android-app://org.mysociety.fixmystreet/`

PR #4376, opened 2023-04-11
Store app platform in service field for reports from wrapped PWA

davea:
For #4375

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1072,10 +1072,20 @@ sub process_report : Private {
         'may_show_name',                         #
         'subcategory',                              #
         'partial',                               #
-        'service',                               #
         'non_public',
       );
     $params{category} = $c->stash->{category};
+
+
+    # If this report is being made by the wrapped PWA then the platform
+    # (iOS/Android) is stored in the session, so use that for the service field.
+    # NB the $c->sessionid check is necessary because we don't want to actually
+    # create a session here if it doesn't already exist (which accessing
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1072,11 +1072,13 @@ sub process_report : Private {
...
         'subcategory',                              #
         'partial',                               #
-        'service',                               #
         'non_public',
       );
     $params{category} = $c->stash->{category};
 
+
+    my $service = ($c->sessionid ? $c->session->{app_platform} : $c->get_param('service')) || '';
+
```
MorayMySoc, line 1081: This would need to be `|| 'desktop'` to pass the cypress test - and presumably it's correct that this is where we need to assert desktop if it's not an app?

MorayMySoc (changes requested):
This seems to work - the only issue is the failing test and I think it's picket up on that we would need to add desktop in this as well.

perllib/FixMyStreet/App/Controller/Report/New.pm
davea: Ooh, yes, good spot. I think I've got the logic wrong. What I was trying to say was "Use `app_platform` from session if it's there, otherwise the `service` param or `''`", but what I've actually written is "If there's a session then use `app_platform` or `''`, if no session then use the `service` param or `''`".

There was no mention of `desktop` before so I don't think that's necessary. I will rejig the logic to be correct and put a explanatory comment in.

MorayMySoc (approved):
Looks good to me!

PR #4387, opened 2023-04-18
[Brent] Add report cut off date

MorayMySoc:
Set cut off date to Brent go-live date so reports in all reports shown from then

[skip changelog]

https://github.com/mysociety/societyworks/issues/3421


dracos (approved):
Fine, though I would change the date in the docs to not be world-ambiguous :)

PR #4388, opened 2023-04-18
[Peterborough] Update bin collection text to match response template

chrismytton:
Client request via FD-2863.


davea (approved):

dracos:
I think it is referencing the wrong ticket, this is https://mysocietysupport.freshdesk.com/a/tickets/2855 which they've just asked to be put live, @chrismytton are you okay to put it live?

PR #4389, opened 2023-04-18
[Brent][WW] Add amend green garden waste

MorayMySoc:
Add a cobrand hook to Waste so Brent can change
the category to an Amend garden waste category
and set extra data fields on that based on the
changes made when modifying a garden waste subscription

Branch with same name on mysociety server too.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1349,7 +1349,11 @@ sub garden_renew : Chained('garden_setup') : Args(0) {
         $c->detach;
     }
 
-    $c->stash->{first_page} = 'intro';
+    if ($c->stash->{waste_features}->{ggw_discount_as_percent} && $c->stash->{is_staff}) {
+        $c->stash->{first_page} = 'discount';
+    } else {
+        $c->stash->{first_page} = 'intro';
+    }
     my $service_id = $c->cobrand->garden_service_id;
     my $max_bins = $c->stash->{quantity_max}->{$service_id};
     $c->stash->{garden_form_data} = {
@@ -1502,7 +1506,6 @@ sub process_garden_modification : Private {
             $c->set_param('pro_rata', $pro_rata);
             $c->set_param('admin_fee', $cost_now_admin);
         }
-        $c->cobrand->call_hook(waste_garden_mod_params => $data);
         $payment_method = $c->stash->{garden_form_data}->{payment_method};
         $payment = $cost_pa;
         $payment = 0 if $payment_method ne 'direct_debit' && $new_bins < 0;
@@ -1511,6 +1514,7 @@ sub process_garden_modification : Private {
     $c->set_param('payment', $payment);
 
     $c->forward('setup_garden_sub_params', [ $data, $c->stash->{garden_subs}->{Amend} ]);
+    $c->cobrand->call_hook(waste_garden_mod_params => $data);
     $c->forward('add_report', [ $data, 1 ]) or return;
 
     if ( FixMyStreet->staging_flag('skip_waste_payment') ) {
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1378,7 +1378,8 @@ sub process_garden_cancellation : Private {
     $data->{payment_method} = $payment_method;
 
     my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
-    $c->set_param('Subscription_End_Date', $now->ymd);
+    my $end_date_field = $c->cobrand->call_hook(alternative_backend_field_names => 'Subscription_End_Date') || 'Subscription_End_Date';
+    $c->set_param($end_date_field, $now->ymd);
 
     my $service = $c->cobrand->garden_current_subscription;
     if (!$c->stash->{slwp_garden_sacks} || $service->{garden_container} == 26 || $service->{garden_container} == 27) {
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1501,7 +1501,7 @@ sub process_garden_modification : Private {
             $c->set_param('pro_rata', $pro_rata);
             $c->set_param('admin_fee', $cost_now_admin);
         }
-
+        $c->cobrand->call_hook(waste_garden_mod_params => $data);
         $payment_method = $c->stash->{garden_form_data}->{payment_method};
         $payment = $cost_pa;
         $payment = 0 if $payment_method ne 'direct_debit' && $new_bins < 0;
--- a/templates/web/base/waste/garden/modify.html
+++ b/templates/web/base/waste/garden/modify.html
@@ -16,19 +16,18 @@
     [% INCLUDE waste/garden/_bin_quantities.html form_page='modify' %]
   <hr class="fieldset-hr">
   <div class="cost-pa__total js-bin-costs"
+  [% IF form.saved_data.apply_discount %]
+    data-per_bin_cost="[% c.cobrand.apply_garden_waste_discount(per_bin_cost) %]"
+    data-per_new_bin_first_cost="[% c.cobrand.apply_garden_waste_discount(per_new_bin_first_cost) %]"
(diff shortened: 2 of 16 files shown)


MorayMySoc:
Richard has asked for this on staging https://3.basecamp.com/4020879/buckets/28996137/todos/6021971134#__recording_6063881779

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1501,7 +1501,7 @@ sub process_garden_modification : Private {
             $c->set_param('pro_rata', $pro_rata);
             $c->set_param('admin_fee', $cost_now_admin);
         }
-
+        $c->cobrand->call_hook(waste_garden_mod_params => $data);
```
dracos, line 1504: I'd put this below after setup_garden_sub_params, I think? Is matching what that is doing (when that applies to both new/modify) - and also needs to override the Container_Type/Quantity set by it (though is setting to the same values).

dracos (commented):
One suggestion to move the call lower down is all, then can stick on staging

web/js/waste.js
```diff
@@ -75,7 +77,7 @@ $(function() {
           $('#new_bin_count').text(0);
       }
       $('#apply_discount_wrapper').hide();
-      if ($('[name=apply_discount]')[0] && $('[name=apply_discount]')[0].checked) {
+      if (discount_selected) {
```
dracos, line 80: Presumably the `$('#modify [name=apply_discount]')` line below is no longer needed if the tick box is no longer present.

templates/web/base/waste/garden/modify.html
```diff
@@ -21,6 +21,7 @@
     data-per_new_bin_cost="[% per_new_bin_cost %]"
     data-pro_rata_bin_cost="[% garden_form_data.pro_rata_bin_cost %]"
     [% IF is_staff && waste_features.ggw_discount_as_percent %]
+        data-discount_applied="[% form.saved_data.apply_discount %]"
```
dracos, line 24: What does the new subscription page do, which also asks for the discount before the page with the numbers/prices on? Looks like it adjusts per_bin_cost so then the already-adjusted value is in the data here, meaning it doesn't need an extra flag. 
This is okay for getting it on staging for them to look at, but after we should probably decide which of two methods is clearer and go with that (think it might be this one, though, this extra flag feels a bit clearer maybe?)

dracos (approved):
One line I think is superfluous now, and one comment about picking between two ways of doing same thing in future

templates/web/base/waste/garden/modify.html
MorayMySoc: Ok, I've put it on staging with this as is, but removed the on-change in js file as suggested below.

MorayMySoc:
The latest fixup for this is for cancelling GGW

dracos (approved):

templates/web/base/waste/garden/modify.html
MorayMySoc: @dracos This is what the new fixup relates to - stop the discount adjusting the new subscriptions

web/js/waste.js
```diff
@@ -86,6 +88,7 @@ $(function() {
       $('#pro_rata_cost').text(pro_rata_cost.toFixed(2));
     }
     $('#modify #bins_wanted, #modify #current_bins').on('change', modify_cost);
+    $(document).ready(modify_cost, bin_cost_new);
```
dracos, line 91: I don't think you can supply multiple functions to read(). Also bit worried that the modify code doesn't expect to be run on the new/renew page, and vice-versa? Also on a new page, don't you need to call it with bin_cost_new(1)?

templates/web/base/waste/garden/subscribe_details.html
```diff
@@ -18,7 +18,16 @@
         data-per_bin_cost="[% per_bin_cost %]"
         data-per_new_bin_first_cost="[% per_new_bin_first_cost %]"
         data-per_new_bin_cost="[% per_new_bin_cost %]"
+        [% IF is_staff && waste_features.ggw_discount_as_percent %]
+          data-discount_applied="[% form.saved_data.apply_discount %]"
```
dracos, line 22: I may be missing something here - but if you didn't have the data-discount_applied line here (and sacks equivalent)  for the new subscription page, you wouldn't need the special precalculated_discount flag in the JS? As that's only used when this line is present (and this line isn't used for anything else).

dracos (changes requested):
Couple of queries to perhaps simplify it (if I've understood correctly), tests a bit odd but doesn't look related, hopefully temporary blip.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1378,7 +1382,8 @@ sub process_garden_cancellation : Private {
     $data->{payment_method} = $payment_method;
 
     my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
-    $c->set_param('Subscription_End_Date', $now->ymd);
+    my $end_date_field = $c->cobrand->call_hook(alternative_backend_field_names => 'Subscription_End_Date') || 'Subscription_End_Date';
+    $c->set_param($end_date_field, $now->ymd);
```
dracos, line 1386: Afraid just a tiny bit too much squashing - I think the change to allow Subscription_End_Date to vary should be its own commit, nothing to do with the rest of it :)

templates/web/base/waste/garden/modify.html
```diff
@@ -16,34 +16,45 @@
...
-        data-discount_percent="[% waste_features.ggw_discount_as_percent %]"
-    [% END %]
+  [% END %]
   >
-  [% IF is_staff && waste_features.ggw_discount_as_percent %]
-    <span hidden class="cost-pa__total-costs">
-      <span id="apply_discount_wrapper" hidden>Discount applied: <span id="apply_discount_amount">[% tprintf( '%.0f', waste_features.ggw_discount_as_percent ) %]%</span></span>
+  [% IF form.saved_data.apply_discount %]
+    <span class="cost-pa__total-costs">
+    Total per year, from [% display_end_date.strftime('%e %B %Y') %]: £<span id="cost_per_year">[% tprintf('%.2f', c.cobrand.apply_garden_waste_discount(cost_pa)) %]</span>
```
dracos, line 33: I believe this discount (and for cost_now_admin and pro_rata) is already taking place in Garden/Modify.pm, in update_field_list under the $data->{apply_discount} part.

templates/web/base/waste/garden/renew.html
```diff
@@ -12,37 +12,48 @@
...
-        data-discount_percent="[% waste_features.ggw_discount_as_percent %]"
-      [% END %]
-    >
-    [% IF is_staff && waste_features.ggw_discount_as_percent %]
-        <span hidden class="cost-pa__total-costs">
-            <span id="apply_discount_wrapper" hidden>Discount applied: <span id="apply_discount_amount">[% tprintf( '%.0f', waste_features.ggw_discount_as_percent ) %]%</span></span>
-        </span>
     [% END %]
+    >
+    [% IF form.saved_data.apply_discount %]
```
dracos, line 26: Similarly here for cost_pa/cost_now_admin/cost_now - unlike the per_bin variables, these ones have already been discounted. Unless I've missed something?

dracos (changes requested):
This looks a lot easier to follow doesn't it :) Couple of queries/suggestions, but hopefully that might then be it.

perllib/FixMyStreet/App/Controller/Waste.pm
dracos: I've separated this out.

templates/web/brent/waste/services.html
```diff
@@ -63,7 +63,7 @@
   </form>
   [% END %]
   [% IF NOT pending_cancellation %]
-      [% IF NOT unit.garden_due AND NOT waste_features.garden_modify_disabled %]
+      [% IF NOT unit.garden_due AND NOT waste_features.garden_modify_disabled AND is_staff %]
```
dracos, line 66: You've put this one back but previously there was another one on line 33 as well, presumably should also be there to match?

MorayMySoc: Seems a bit deliberate but I don't see that the conversation changed from this being staff only

dracos (approved):
🎉 ? :)

PR #4394, opened 2023-04-20
[Brent][WW] Add general enquiry form

MorayMySoc:
Add a general enquiry form for WasteWorks

https://3.basecamp.com/4020879/buckets/28996137/todos/5943751764

Config in society servers 3575-brent-general-enquiry-event-for-staff

[skip changelog]


dracos (approved):
Assume we know what event type ID to use for the event? :)

PR #4396, opened 2023-04-20
[Hackney] Resend report upon email category change.

dracos:
[skip changelog]
Added a few docs in same commit, sorry, only functional change is the new `category_change_force_resend` function.


MorayMySoc (approved):
This looks good to me

PR #4401, opened 2023-04-24
Tweak app-like display CSS.

dracos:
Prevent over-scrolling of the page, user selection on the map, and leave enough gap at the bottom on the full screen map.

I have tested the first two on my phone, but the last one I'm not sure how to emulate properly (as it's only in the native app wrapper that it's fullscreen enough for it to matter, IYSWIM).

[skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -2166,6 +2166,11 @@ html.js #map .noscript {
     }
 }
 
+// Hide the "Start new report here" button when map filters are open
+#mapForm.mobile-filters-active .map-mobile-report-button {
+  display: none;
+}
+
 .map-crosshairs {
     display: none;
     position: absolute;
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -1812,6 +1812,10 @@ input.final-submit {
   overflow: hidden;
   position: relative;
 
+  // Don't want people 'selecting' the map tiles etc and it looking a bit odd
+  user-select: none;
+  -webkit-user-select: none;
+
   #map {
     width: 100%;
     height: 100%;
@@ -2134,6 +2138,7 @@ html.js #map .noscript {
   bottom: 0;
   margin: 0;
   margin-bottom: 6em;
+  margin-bottom: calc(env(safe-area-inset-bottom) + 6em);
 
   @include border-radius(0.5em);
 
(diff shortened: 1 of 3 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -34,6 +34,12 @@ $geolocation-link-border: $geolocation-link !default;
 @import "_mixins";
 @import "_report_list";
 
+// Prevent glow/bounce at bottom of page (also sadly prevents pull to refresh
+// at top, but refresh button still there)
+html {
+  overscroll-behavior-y: none;
```
davea, line 40: Hmm, I actually prefer the rubber banding (and funnily enough the [PWA Builder code does disable it](https://github.com/pwa-builder/pwabuilder-ios/blob/1f8ff59ee554137b34bb616869f5e4e62dc2754f/Microsoft.PWABuilder.IOS.Web/Resources/ios-project-src/pwa-shell/WebView.swift#L31) by default and I re-enabled it when I built the app). Most scrollviews on iOS apps do the rubberbanding by default as far as I can tell.

davea (commented):
The `viewport` meta tag needed tweaking (in 099e3712e9d78a6d3f57dda398dfc58269675f5a) to enable `env`, which showed up a few problems with the underlying list of reports showing through and there being a big gap underneath the filter button when the filters were active:

| Simulator Screenshot - iPhone 14 Pro - 2023-04-26 at 12 00 43 | Simulator Screenshot - iPhone 14 Pro - 2023-04-26 at 12 00 14 |
| --- | --- |

I thought maybe it'd be neat to be able to view the map beneath the safe area so twiddled the CSS a bit in 1077bd9beec7de1318b41d1787b823b6cf957406 to make it so:

| Simulator Screenshot - iPhone 14 Pro - 2023-04-26 at 12 35 22 | Simulator Screenshot - iPhone 14 Pro - 2023-04-26 at 12 35 26 |
| --- | --- |


(also hid the "start new report" button when filters are active in 11f274e117f0840549abb1e696b4fec16c84d493)

web/cobrands/sass/_base.scss
dracos: I think I noticed this when making a new report, and e.g. scrolling down the category list and being annoyed by it bouncing at the end, so took longer to 'settle' and then press the continue button - I don't find apps generally bounce for that kind of thing, but I admit I don't use a lot of apps :) And scrolling the front page, seemed odd to continue past the footer. Dunno.

PR #4402, opened 2023-04-24
Add way to use digest in static filename.

dracos:
A query parameter works fine, and is easier to manage, until you get to the situation where you have your website on three application servers, which get deployed in sequence. After the first server is deployed, the HTML for a page could be served from the first server (with a new hash) but the file is then requested from the third server (with old content) and your load balancer then happily caches the old content with the new hash URL.

In a backwards-compatible manner, we set up a static folder (probably a symlink to a shared folder generated at the start of a deploy) that can contain files with the hash in their filename. If present, they will be used instead of the main file with a hash in the query string.

All the other changes bar the change to the server code are to make all CSS paths relative to the root, so that they work from either location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ FixMyBarangay.po
 /web/photo
 # And theme upload
 /web/theme
+# And copied static files with content hash in the name
+/web/static
 
 # Installation
 /local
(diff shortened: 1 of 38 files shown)


sagepe (approved):

dracos:
Merged as 990f651fc (added a change to a test in case you run tests with static assets generated as then the path didn't match).

PR #4403, opened 2023-04-25
[Zürich] Stop showing optional label on photo field when not optional.

neprune:
fixes https://github.com/mysociety/societyworks/issues/3072

Additionally:
* Move category selection before photo upload for Zurich so the label is changed before the reporter is prompted.
* Make the photo field label visually appear as a label. Currently this appears as normal text which looks especially weird for Zürich after the above change. This applies to all cobrands.

I have skipped setting up a cypress test as it was taking me too long.

# Demo

Setup: category 'Graffiti' requires a photo, the others don't.

## Before changes

Zurich Before

## After

https://user-images.githubusercontent.com/9289297/235450684-cf0452c6-bb82-47aa-a3f0-1ea3bd0ca46b.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
         - Add label to questionnaire textarea. #3944
         - Offline report drafting. #4290
         - Set width=device-width in viewport meta tag. #4384
+        - Fix photo field label not visually appearing as a label.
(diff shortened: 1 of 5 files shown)


chrismytton (approved):
This works really nicely! 👍

PR #4405, opened 2023-04-26
[UK] Move some roadworks setup to server config.

dracos:
[skip changelog]
This would move the config for bath/peterborough to a cobrand feature in the config, being a list of the categories as per the existing JS, plus can then add Brent who want it for `[ 'Faded white lines', 'Faded yellow lines', 'Pavement damage', 'Pothole or other road damage' ]`

For FD-2892


chrismytton (approved):
🚧 👍

PR #4408, opened 2023-04-28
Update asset layer max_resolution documentation

chrismytton:
Add a note about how to look up the current resolution, which can then be used for setting the max_resolution.

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -989,6 +989,8 @@ max_resolution - either a number, or a hash mapping cobrand to number. This
     provides the maximum resolution at which an asset layer will be displayed
     (and when first shown, the map may zoom in to this level). The hash is for
     the case where a cobrand shows a different map with different resolutions.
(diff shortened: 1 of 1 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -989,6 +989,8 @@ max_resolution - either a number, or a hash mapping cobrand to number. This
     provides the maximum resolution at which an asset layer will be displayed
     (and when first shown, the map may zoom in to this level). The hash is for
     the case where a cobrand shows a different map with different resolutions.
+    Run `fixmystreet.map.resolution` in the browser's js console to see the
```
dracos, line 992: This will work fine, though I guess it should be `fixmystreet.map.getResolution()` (which I think handles edge cases where e.g. the base layer isn't present as per the comment at https://github.com/mysociety/ol2/blob/76657c6857748cd795f1e4c26b1228e71fcceaad/lib/OpenLayers/Map.js#L2262)

dracos (approved):
Could switch to the method call is all :)

PR #4412, opened 2023-05-03
[FMS] Add check for cumbria/nyorks/somerset placeholder for emails

nephila-nacrea:
We redirect to the correct email address as we do for Northants, due to 2023 council changes for Cumbria, N Yorkshire, & Somerset.


dracos (commented):
Couple of small comments, only for info, I've made the changes involved.

PR #4413, opened 2023-05-03
Mark all archive updates as processed.

dracos:
Otherwise the daemon still has to go through and discount all the email ones. [skip changelog]


chrismytton (approved):
📩

PR #4415, opened 2023-05-03
[Offline] Multiple drafts

### LINKED ISSUE #4410: Multiple offline drafts
The offline functionality currently only supports a single draft report. To match the functionality of the retired mobile app we should allow the user to create/manage multiple drafts.

davea:
Thought the individual commits might be easier to review, but the whole lot can probably be squashed down to one commit before merging.

[skip changelog]

Fixes #4410

### DIFF
--- a/web/cobrands/fixmystreet/offline_report.js
+++ b/web/cobrands/fixmystreet/offline_report.js
@@ -1,3 +1,5 @@
+// jshint esversion: 8
+
 fixmystreet.offlineReporting = (function() {
     function updateDraftSavedTimestamp(ts) {
         if (ts) {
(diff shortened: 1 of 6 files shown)


web/cobrands/fixmystreet/offline_report.js
```diff
@@ -200,32 +301,92 @@ fixmystreet.offlineReporting = (function() {
...
+
     return {
         offlineFormSetup: function() {
             dropzoneSetup();
-            $(".js-delete-drafts").on("click", function(e) {
+
+            $(document).on('click', '.js-continue-draft', function() {
+                var id = parseInt(this.parentNode.parentNode.getAttribute('data-id'), 10);
+                loadDraft(id).then(function(d) {
+                    location.href = "/report/new?restoreDraft=" + id + "&latitude=" + d.latitude + "&longitude=" + d.longitude;
```
dracos, line 355: This happens at present, but just noticed now in testing - if you haven't saved a location for a report, then this button redirects to the front page, and even if you provide a location your draft text is lost (as restoreDraft not there any more). Worth opening a new ticket for rather than fixing here, I think, eg the reporting moving pin bug is more urgent.

dracos (commented):
One tiny thing I think was added than inside code was removed so not needed, and I noticed a no-location issue (but that's already there not due to this PR) is all :)

PR #4418, opened 2023-05-04
[UK] Add page about installing web app.

dracos:
For #3605, to have something up with text and some screenshots, if not having screenshots/logos superimposed in the right bits etc.

Example for Brent here - they'd all be the same but with different council names/phone screenshots.

brent localhost_3000_about_web-app
[skip changelog]


templates/web/fixmystreet-uk-councils/about/web-app.html
```diff
@@ -0,0 +1,72 @@
...
+    <li>Select ‘add to home screen’
+        <img alt="" class="web-app-install__step" src="/cobrands/fixmystreet-uk-councils/frontshots/iphone-a2hs.jpeg">
+    </li>
+    <li>Select ‘ok’/’add’
+        <img alt="" class="web-app-install__step" src="/cobrands/fixmystreet-uk-councils/frontshots/iphone-add.png">
+    </li>
+    <li>Use [% site_name %] as an app</li>
+</ol>
+
+<p>The PWA’s design makes it feel like you’re using a ‘normal’ app, with its mobile friendly features, but you’re benefiting from using a web-based service that doesn’t require you to keep updating it.</p>
```
sallybracegirdle, line 48: I realise we also say this below, so I'm wondering if we should delete this bit to make it less repetitive?

sallybracegirdle:
I think this looks excellent, Matthew!!! Thank you!!!

templates/web/fixmystreet-uk-councils/about/web-app.html
dracos: I wondered if you'd done that on purpose, so someone who was only looking at the "iPhone" section would still read it as they'd probably stop reading the page once they got to "Android"?

dracos: Could make it part of the "Step 5" text perhaps?

sallybracegirdle: I think that's what I was probably thinking at the time, but it looks odd to repeat it. Unless we move it to be above the iOS and Android section, as a second paragraph in the 'How to install...' section?

dracos:
Have updated screenshots so all use red arrow for consistency, added 2x retina screenshot images for Dave, and tidied up few other bits.

dracos:
This was merged fine, I think might have been during GitHub's recent blip.

PR #4424, opened 2023-05-09
Improve speed of front page ‘recent problems’ query

davea:


perllib/FixMyStreet/DB/ResultSet/Problem.pm
```diff
@@ -171,18 +174,18 @@ sub _recent {
         $attrs->{bind} = [ $lat, $lon, $dist ];
         $attrs->{join} = 'nearby';
         $probs = [ mySociety::Locale::in_gb_locale {
-            $rs->search( $query, $attrs )->all;
+            sort { $b->confirmed <=> $a->confirmed } $rs->search( $query, $attrs )->all;
```
dracos, line 177: I think this will show warnings on Zurich where public reports don't need a confirmed to be set, so might need to default to `($b->confirmed || 0) <=> ...`. Or not bother with sorting in code at all :)

dracos (changes requested):
Odd test failure to investigate, and possible defaults to set - though as I say there, perhaps db sort is enough, and don't bother then sorting by confirmed. If you had:
```
id | created | confirmed
1 | 2023-05-10T08:00 | 2023-05-10T14:00
2 | 2023-05-10T09:00 | 2023-05-10T09:00
3 | 2023-05-10T10:00 | 2023-05-10T11:30
4 | 2023-05-10T11:00 | 2023-05-10T11:00
5 | 2023-05-10T12:00 | 2023-05-10T12:00
6 | 2023-05-10T13:00 | 2023-05-10T13:00

Before PR: 1, 6, 5, 3, 2
With PR: 6, 5, 3, 4, 2
With PR but no code sorting: 6, 5, 4, 3, 2
Not sure it matters if they're displaying as the last rather than the middle :shrug: :)

dracos (approved):

PR #4427, opened 2023-05-10
Fix display of moderated extra.

dracos:
[skip changelog] Spotted by Tracy on a Bucks report with a Confirm reference (which should be ignored anyway, so added).


davea (approved):
👍

PR #4428, opened 2023-05-11
FMS Admin Dashboard form

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3663

This PR includes two commits:

The first one should* fix the non-wrapping behaviour that the dashboard currently has. I think due to the number of fields in the form it makes sense that they wrap up instead of just cutting each other.

Screenshot 2023-05-11 at 07 28 00

_*I couldn't find the original screen that was shown in the issue listed above, so I used one screen that seemed similar(/dashboard?group_by=month), but I might need to enable something in order to see the original screen and properly test it. In this case I only added extra fields in the browser to test what happens with extra fields, like the ones shown in the original issue._

The second one is just for aesthetics purpose and happy to drop it if it only adds more complexity to the code. I noticed the the fields had different height, making the form look a bit odd, especially when is only one line. 

Screenshot 2023-05-11 at 07 27 35

[skip changelog]

### DIFF
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -227,6 +227,7 @@
     // Force field elements onto a single line.
     @media (min-width: 48em) {
         @include flex-container();
+        flex-wrap: wrap;
     }
 
     // No border-top when visually preceded by .dashboard-header
@@ -245,12 +246,13 @@
 
         @media (min-width: 48em) {
             float: $left;
-            max-width: 25%;
         }
     }
 
     .no-label {
         margin-top: 1.25em + 1.5em + 0.5em; // label line-height + margin-top + margin-bottom
+        text-align: right;
+        flex-grow: 1;
     }
 
     select {
@@ -282,6 +284,10 @@
         border-color: #ccc;
         color: #999;
         box-shadow: none;
+    }
+
+    .form-control {
+        height: 2.2em;
     }
 
     .btn {


dracos (commented):
Thanks, this looks better :) (You don't need to put `[FMS]` on commits, this is all in the fixmystreet repo, I've squashed/reworded the commits, hope that makes sense.)
I've pushed one more suggestion so that the button moves to the right if it's on a second row, what do you think?

lucascumsille:
> Thanks, this looks better :) (You don't need to put `[FMS]` on commits, this is all in the fixmystreet repo, I've squashed/reworded the commits, hope that makes sense.)
> I've pushed one more suggestion so that the button moves to the right if it's on a second row, what do you think?

Thanks Matthew =) much appreciated.
Regarding your commit, I like it because it makes that button stand out a bit more. From UI perspective, the bottom doesn't always align with the filter at the top, but I don't think that's such a big deal. See image:
Screenshot 2023-05-16 at 12 49 57

Happy to go ahead with this.

PD: I tried to fix the problem above it by adding `justify-content: space-between` to the `.filter` element, but I noticed that there is an invisible table as first child of `.filter`. I could fix that, but it will make the code more complex, for such a tiny benefit.

Let me know what you think.

dracos:
> I tried to fix the problem above it by adding justify-content: space-between to the .filter element, but I noticed that there is an invisible table as first child of .filter. I could fix that, but it will make the code more complex, for such a tiny benefit.

Also, I don't think (if I've understood correctly) that will work because the start/end date are wrapped in one fieldset as the flex child, so the spacing there will always be independent. Fine with it as is :)

PR #4430, opened 2023-05-11
[Bexley] Include user email in dashboard CSV export

davea:
For FD-2623

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -341,4 +341,28 @@ sub report_form_extras {
 
 sub report_sent_confirmation_email { 'id' }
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -341,4 +341,28 @@ sub report_form_extras {
...
+    $csv->add_csv_columns(
+        user_email => 'User Email',
+    );
+
+    $csv->objects_attrs({
+        '+columns' => ['user.email'],
+        join => 'user',
+    });
+
+    return if $csv->dbi; # Already covered
```
dracos, line 356: You also need to add bexley to the `user_details` hash in CSVExport.pm so that the column is fetched.

PR #4432, opened 2023-05-12
[Brent] Fix form issue with grass verges layer.

dracos:
Asset layer functions are called in sequence. Depending on asset layer ordering, if a point covered by an asset in the second-called layer is clicked, first the other layer’s not_found function will be called, adding the new map page to the form to show its error on mobile. When found is then called for the other layer, that layer would remove its own map page (if it had created it) but not the one created by a different layer. So we need to look up the other layer and make sure it is treated as being found as well.

Also fix the tests so that ones that were mistakingly passing pass for the right reason (by checking their next page and adding the missing road fixture).

[skip changelog]


MorayMySoc (approved):
With the addition of the ``&& f.http_options`` in ``web/cobrands/fixmystreet-uk-councils/assets.js`` we discussed this seems to solve the issue so good to go after that.

PR #4433, opened 2023-05-12
[FMS.com & BANES] Zoom further in when using geolocation on /around

davea:
Thought this might be worth trying to see if it makes the mobile experience a bit better. 

[skip changelog]

For FD-2979

### DIFF
--- a/perllib/FixMyStreet/Cobrand/BathNES.pm
+++ b/perllib/FixMyStreet/Cobrand/BathNES.pm
@@ -99,7 +99,15 @@ sub pin_colour {
 
 sub send_questionnaires { 0 }
 
-sub default_map_zoom { 3 }
+sub default_map_zoom {
+    my $self = shift;
+
+    # If we're displaying the map at the user's GPS location we
+    # want to start a bit more zoomed in than if they'd entered
+    # a postcode/address.
+    return 3 unless $self->{c}; # no c for batch job calling static_map
+    return $self->{c}->get_param("geolocate") ? 5 : 3;
+}
 
 sub available_permissions {
     my $self = shift;
--- a/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
+++ b/perllib/FixMyStreet/Cobrand/FixMyStreet.pm
@@ -262,6 +262,17 @@ sub extra_contact_validation {
     return %errors;
 }
 
+sub default_map_zoom {
+    my $self = shift;
+
+    # If we're displaying the map at the user's GPS location we
+    # want to start a bit more zoomed in than if they'd entered
+    # a postcode/address.
+    return unless $self->{c}; # no c for batch job calling static_map
+    return $self->{c}->get_param("geolocate") ? 4 : undef;
+}
+
+
 =head2 council_dashboard_hook
 
 This is for council-specific dashboard pages, which can only be seen by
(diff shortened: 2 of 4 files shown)


davea:
Having played with this a bit more on real devices I think it's an improvement, what do you think?

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -215,6 +215,16 @@ sub extra_contact_validation {
     return %errors;
 }
 
+sub default_map_zoom {
+    my $self = shift;
+
+    # If we're displaying the map at the user's GPS location we
+    # want to start a bit more zoomed in than if they'd entered
+    # a postcode/address.
+    return $self->{c}->get_param("geolocate") ? 4 : undef;
```
dracos, line 224: As above, wonder if this could be "+1 to default" (so 4 or 3) rather than always 4 (ignoring whether the 2/3 distinction is perfect or not). Just 4 in very rural areas can be a bit blank. But is probably fine?

dracos: On a separate point, these maps are generated at the command line for emails in the `static_map` function, so $self->{c} will not always be present. Tests don't show this because they involve ImageMagick so that function is skipped.

dracos (changes requested):
One issue with test-skipped function (perhaps need to improve that for something like this, GH tests could presumably install imagemagick?) and some chatty discussion about the zoom levels.

perllib/FixMyStreet/Cobrand/BathNES.pm
```diff
@@ -86,7 +86,14 @@ sub pin_colour {
...
 sub send_questionnaires { 0 }
 
-sub default_map_zoom { 3 }
+sub default_map_zoom {
+    my $self = shift;
+
+    # If we're displaying the map at the user's GPS location we
+    # want to start a bit more zoomed in than if they'd entered
+    # a postcode/address.
+    return $self->{c}->get_param("geolocate") ? 5 : 3;
```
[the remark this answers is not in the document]
davea: It's a difficult balance to get it right for everyone/everywhere, but I think zooming _out_ after getting your bearings is nicer than zooming _in_ to find where you are.

For the use case of wanting to make a report at my current location, to my eye this is a reasonable view after tapping the geolocation button:
Simulator Screenshot - iPhone 14 Pro - 2023-05-15 at 11 10 46

I can easily correspond what I can see around me to my position on the map. Yes, that's not a location with a load of nearby landmarks, but I can quickly be confident that the pothole I spotted just over there *waves arms*  should go just there *drags crosshair* on the map.

perllib/FixMyStreet/Cobrand/BathNES.pm
```diff
@@ -92,6 +92,7 @@ sub default_map_zoom {
     # If we're displaying the map at the user's GPS location we
     # want to start a bit more zoomed in than if they'd entered
     # a postcode/address.
+    return unless $self->{c}; # no c for batch job calling static_map
```
dracos, line 95: This presumably should return 3 to match the old behaviour

davea: 🤦 thanks!

dracos (approved):
Woo

PR #4443, opened 2023-05-17
RABX migration fixes

dracos:
No actual code change (we changed the DB directly for Zurich booleans and open311_protect) but update test to match what code will be having, and random speed up noticed. [skip changelog]


davea (approved):
Woo, shaves ~20% off the total test run time for me 👍

PR #4444, opened 2023-05-17
Switch to OSM for reverse geocoding

davea:


davea:
@dracos not quite finished yet (need to ensure all cobrands' use of `resourceSets` can handle OSM results) but does this look like the right sort of thing?

dracos (commented):
Yeah, looking good :)

t/cobrand/peterborough.t
```diff
@@ -392,6 +384,7 @@ subtest 'Dashboard CSV extra columns' => sub {
     }, sub {
         $mech->get_ok('/dashboard?export=1');
     };
+    diag $mech->content;
```
dracos, line 387: diag left in

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -90,13 +90,26 @@ around open311_extra_data_include => sub {
...
-            { name => 'postcode', value => $address->{postalCode} },
-            { name => 'house_no', value => $number },
-            { name => 'street', value => $street }
-        );
+        my ($number, $street, $postcode);
+
+
+        if ($row->geocode->{resourceSets}) { # Bing results
+            my $address = $row->geocode->{resourceSets}->[0]->{resources}->[0]->{address};
+            ($number, $street) = $address->{addressLine} =~ /\s*(\d*)\s*(.*)/;
```
dracos, line 98: The setting of postcode from address->postalCode has been dropped.

dracos (changes requested):
A few small issues, plus one large issue concerning tests making external calls.

Given the amount of code checking display_name/resourceSets, I'd later suggest we factor that into Problem.pm so it can return one canonical set of data for a geocode row perhaps, would simplify all the changes here, but probably not for now given the issue.

dracos (approved):
Woo :)

PR #4454, opened 2023-05-22
[Brent] Show collection time bands from Echo on WasteWorks

davea:
For FD-2999

image

[skip changelog]


perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -561,6 +566,20 @@ sub bin_services_for_address {
     return \@out;
 }
 
+sub _timeband_for_servicetask {
+    my $servicetask = shift;
```
dracos, line 570: Rather than passing in service task and looking up the schedule here (which may fail as some tasks have multiple schedules), you could use/pass in the `$schedules->{next}` which has already parsed the schedules and saved a relevant one in `->{schedule}` (at least for the purposes of looking up TimeBand I would hope, I added it to look up Allocation for PDF).

davea: Ooh neat, I'll have a look. Ta!

dracos (approved):

PR #4455, opened 2023-05-22
Use email field type when SMS authentication not in use

davea:
This makes logging in easier on mobiles as the email-specific soft keyboard is presented.


dracos (approved):
Looks fine :) Is this an opportunity to consolidate these templates? Looks like "add a block for staff", "no create account", and "no sign in with password" flags would handle all the variants currently in use. Probably not worth it.

PR #4456, opened 2023-05-22
Allow cobrands to customise nearby lookup distances for duplicates

davea:
For UK councils this is picked up from cobrand feature config.

For FD-2997.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -679,10 +679,14 @@ sub _nearby_json :Private {
     $c->stash->{page} = 'report';
 
     # distance in metres
-    my $dist = $c->get_param('distance') || '';
+    my $dist;
+    if (my $mode = $c->get_param('mode')) {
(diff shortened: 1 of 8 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -679,10 +679,14 @@ sub _nearby_json :Private {
     $c->stash->{page} = 'report';
 
     # distance in metres
-    my $dist = $c->get_param('distance') || '';
+    my $dist;
+    if (my $mode = $c->get_param('mode')) {
```
dracos, line 683: Given there are two different endpoints for the two different modes, can we set 'mode' in the controllers that then call `_nearby_json` rather than a query parameter?

davea: Suppose we could, but there's currently nothing inspector- or suggestions-specific about the behaviour of the two different paths. I think I'd prefer to leave them as-is with the intention of each call made explicit rather than adding some subtle difference in behaviour of two very similarly named things that'd trip somebody up in the future (most likely me!)

PR #4460, opened 2023-05-24
[FMS] Update House Rules with name change and staff well-being text

MorayMySoc:
Changes references from House Rules to Conditions of Use.

Adds text to be clear about abusive messages to FMS. This is being updated in terms of good practice - not in relation to any abuse.

https://github.com/mysociety/societyworks/issues/3713


dracos (changes requested):
No changelog needed, and one text change (once decided) is all, I think.

PR #4461, opened 2023-05-26
Auto-spot a default favicon.ico.

dracos:
If web/cobrands/<cobrand>/favicon.ico exists, automatically include that in the HTML header.

Also adds favicon for BANES from FD-2954. [skip changelog]


MorayMySoc (approved):
Looks good - take it the 'type="image/x-icon"' option is not important

PR #4464, opened 2023-05-26
[Bristol] Allow reports in two parks outside Bristol's boundaries

chrismytton:
Bristol are responsible for reports in two parks that lie partially outside the Bristol boundary. This change uses the existing parks asset layer to determine if the user is making a report in one of those parks, and if so it treats it as if it's in Bristol.

Fixes https://github.com/mysociety/societyworks/issues/3651

## ⚠️ Notes for reviewer/deployer

The **Bristol body needs to be modified to cover South Gloucestershire and North Somerset** in order for this code to work correctly. I've already done this on staging.

image

---


dracos (commented):
Doesn't matter hugely, but if you mocked `_fetch_features` (either returning [] or an item in a list) that'd at least check the `_park_for_point` code was e.g. returning the data in the right way (as it is, it's not testing that `site_code => $park->{"ms:parks"}->{"ms:SITE_CODE"}` is right, iyswim). Other than that suggestion, looks good :)

perllib/FixMyStreet/Cobrand/Bristol.pm
```diff
@@ -162,4 +168,101 @@ sub post_report_sent {
...
+The Bristol body is configured to cover some other areas, but is only responsible for problems in parks in those areas.
+
+Override this method so an error is shown if the user tries to report a problem outside Bristol.
+
+=cut
+
+sub responsible_for_areas {
+    my ($self, $areas) = @_;
+
+    if (scalar(%$areas) == 1 && (values %$areas)[0]->{name} eq $self->council_name) {
```
dracos, line 182: Is there every a possibility for $areas to not only be 1 in length here? I think it will only be one long.
Could go on the ID matching rather than the name, though unlikely to change I guess.

PR #4471, opened 2023-06-04
Make cobrand text overrides for new report fields apply on fixmystreet.com.

neprune:
fixes: https://github.com/mysociety/societyworks/issues/3497

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -351,9 +351,17 @@ sub by_category_ajax_data : Private {
         $body->{councils_text} = $c->render_fragment( 'report/new/councils_text.html');
     }
 
-    my $hints = $lookups->{hints}{$category};
-    $body->{title_hint} = $hints->{title} if $hints->{title};
-    $body->{detail_hint} = $hints->{detail} if $hints->{detail};
+    my $cobrand_overrides = $c->stash->{cobrand_field_overrides_by_body}->{$bodies->[0]->{name}} if @$bodies == 1;
+    my $category_overrides = $lookups->{hints}{$category};
+
+    my $title_label_override = $cobrand_overrides->{title_label};
+    # prefer category specific overrides if present.
+    my $title_hint_override = $category_overrides->{title} || $cobrand_overrides->{title_hint};
+    my $detail_hint_override = $category_overrides->{detail} || $cobrand_overrides->{detail_hint};
+
+    $body->{title_label} = $title_label_override if $title_label_override;
+    $body->{title_hint} = $title_hint_override if $title_hint_override;
+    $body->{detail_hint} = $detail_hint_override if $detail_hint_override;
 
     return $body;
 }
@@ -778,6 +786,7 @@ sub setup_categories_and_bodies : Private {
       ();    # categories for which the reports are not public
     my %category_photo_required =
       (); # whether a category requires a photo to be uploaded.
+    my %cobrand_field_overrides_by_body = ();
     $c->stash->{unresponsive} = {};
 
     my @refused_bodies = grep { ($_->send_method || "") eq 'Refused' } values %bodies;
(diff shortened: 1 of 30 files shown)


t/app/controller/report_new.t
```diff
@@ -1386,4 +1395,83 @@ subtest "extra google analytics code displayed on email confirmation problem cre
...
+    $second_body->unset_extra_metadata('cobrand');
+    $second_body->update;
+
+    FixMyStreet::override_config {
+        ALLOWED_COBRANDS => 'overrides',
+        MAPIT_URL => 'http://mapit.uk/',
+    }, sub {
+        $json_response = $mech->get_ok_json( '/report/new/ajax?w=1&latitude=' . $lat . '&longitude=' . $long );
+    };
+    isnt defined $json_response->{by_category}->{test}->{title_label}, "cobrand label override unepexctedly applied";
```
dracos, line 1472: These lines are actually testing that eg `defined $json_response->{by_category}->{test}->{title_label}` isn't equal to the string `"cobrand label override unepexctedly applied"` which it definitely won't be :) I think you want `is $json_response->{by_category}->{test}->{title_label}, undef, "cobrand label override unexpectedly applied"` (typo in description too)

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -48,6 +48,18 @@ sub geocoder_munge_results {
...
 
+sub new_report_title_field_label {
+    "Summarise the problem"
+}
+
+sub new_report_title_field_hint {
+    "Exact location, including any landmarks"
+}
+
+sub new_report_title_field_placeholder {
```
dracos, line 59: This isn't a title field placeholder, this is the detail field `hint` (if you're calling the title one `hint` which is good, it's not actually an HTML placeholder). So I think every `new_report_title_field_placeholder` needs to be `new_report_detail_field_hint`.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -846,6 +860,25 @@ sub setup_categories_and_bodies : Private {
         push @category_options, $seen{_('Other')} if $seen{_('Other')};
     }
 
+    foreach my $body (values %bodies_to_list) {
+        my $cobrand_for_body = $body->get_cobrand_handler;
+        next unless $cobrand_for_body;
+
+        $cobrand_by_body{$body->name} = $cobrand_for_body->moniker if $cobrand_for_body->moniker;
```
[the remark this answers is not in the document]
dracos: From my other comment (that I'm actually writing after this one), I think you can drop `cobrand_by_body`, and use the body name as the key for `title_field_overrides_by_cobrand`.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -348,9 +348,21 @@ sub by_category_ajax_data : Private {
         $body->{councils_text} = $c->render_fragment( 'report/new/councils_text.html');
     }
 
-    my $hints = $lookups->{hints}{$category};
-    $body->{title_hint} = $hints->{title} if $hints->{title};
-    $body->{detail_hint} = $hints->{detail} if $hints->{detail};
+    # only use cobrand title overrides if all bodies share a cobrand.
```
dracos, line 351: A cobrand can't be on more than one body, see the check around line 500 in Admin/Bodies.pm, so if there's more than one body it can't be all the same cobrand, and I think these lines can be boiled down to `my $cobrand_overrides = $c->stash->{title_field_overrides_by_cobrand}->{$bodies->[0]->{name}} if @$bodies == 1;` (or something like that anyway).

dracos (changes requested):
Looking really good - I've suggested a way to not duplicate all the labels and drop all the override templates; a way to simplify the cobrand/body checking part (I think); a couple of test issues; and one naming issue.

templates/web/base/report/new/_form_labels.html
```diff
@@ -0,0 +1,5 @@
+[%
+SET form_title = loc(c.cobrand.new_report_title_field_label);
```
dracos, line 2: `loc` needs to be where the string itself is, so need for it in this file.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -6,7 +6,7 @@ BEGIN { extends 'Catalyst::Controller'; }
 
 use utf8;
 use Encode;
-use List::Util qw(uniq any);
+use List::Util qw(uniq all any);
```
dracos, line 9: No longer using `all`

dracos (changes requested):
Thanks, looks like everything asked for done, only a couple of tiny fall outs from the changes :)
I think you can squash all the council cobrand commits together as one `[UK]` commit (potentially everything in one commit, given it's clear it's the independent changes due to the 'core' changes, up to you).

PR #4474, opened 2023-06-06
WW Front page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3635

It started as an issue just for Brent, but it was an opportunity to improve all the WW cobrands. 
This PR:

- Uses the form UI from FMS front page in the WW front page.
- Gives context to the users so they understand what they can do on WW.
- This should improve the maintainability of the front page.

[skip changelog]

### DIFF
--- a/templates/web/base/waste/index.html
+++ b/templates/web/base/waste/index.html
@@ -1,18 +1,28 @@
-[% SET title = form.title ~%]
-[% PROCESS 'waste/header.html' %]
-
-  [% PROCESS 'govuk/fields.html' %]
-  [% PROCESS back %]
-  [% PROCESS errors %]
-  [% PROCESS title %]
-  [% IF property %]
-    [% INCLUDE 'waste/_address_display.html' %]
+[% PROCESS 'govuk/fields.html' %]
+[% pre_container_extra = BLOCK %]
+  [% IF form.errors %]
+    <div class="container">
+      [% PROCESS errors %]
+    </div>
   [% END %]
-
-  [% IF form.intro_template %][% PROCESS "waste/${form.intro_template}" %][% END %]
-
-<form class="waste waste--[% form.current_page.name OR form.name %]" method="post">
-  [% PROCESS form %]
-</form>
-
+    <div id="front-main">
+      <div id="front-main-container">
+        <h1>[% loc('Bins, rubbish and recycling') %]</h1>
+        <form method="post" id="postcodeForm">
+          <label for="pc">[% loc('Postcode') %]</label>
+          <div>
+              <input type="text" name="postcode" value="[% pc | html %]" id="pc" size="10" maxlength="200" required aria-describedby="pc-hint">
+              [% INCLUDE 'around/_postcode_submit_button.html' attr='id="sub"' %]
+          </div>
+          <ul>
+            <li>Find your bin collection days</li>
+            <li>Report a missed bin collection</li>
+            <li>Order new or additional bins</li>
+            <li>Subscribe to Garden waste collection service</li>
+          </ul>
+      </form>
+      </div>
+  </div>
+[% END %]
+[% PROCESS 'waste/header.html' %]
 [% INCLUDE footer.html %]
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -30,7 +30,7 @@ has feature => (
 
 has index_template => (
     is => 'ro',
-    default => 'waste/index.html'
+    default => 'waste/form.html'
 );
 
 sub auto : Private {
@@ -83,6 +83,7 @@ sub index : Path : Args(0) {
     $form->process( params => $c->req->body_params );
     if ($form->validated) {
         my $addresses = $form->value->{postcode};
+        $c->stash->{template} = 'waste/form.html';
         $form = address_list_form($addresses);
     }
     $c->stash->{form} = $form;
--- a/templates/web/base/waste/form.html
+++ b/templates/web/base/waste/form.html
@@ -1,0 +1,18 @@
+[% SET title = form.title ~%]
+[% PROCESS 'waste/header.html' %]
+
+  [% PROCESS 'govuk/fields.html' %]
+  [% PROCESS back %]
+  [% PROCESS errors %]
+  [% PROCESS title %]
+  [% IF property %]
+    [% INCLUDE 'waste/_address_display.html' %]
+  [% END %]
+
+  [% IF form.intro_template %][% PROCESS "waste/${form.intro_template}" %][% END %]
+
+<form class="waste waste--[% form.current_page.name OR form.name %]" method="post">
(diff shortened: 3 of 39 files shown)


lucascumsille:
@dracos 
I still need to do some work with this PR, but I'd like to confirm a couple of things before moving ahead:

### Current state:

Screenshot 2023-06-06 at 10 17 24
Screenshot 2023-06-06 at 10 17 33
Screenshot 2023-06-06 at 10 17 54
Screenshot 2023-06-06 at 10 18 01
Screenshot 2023-06-06 at 10 18 07

1. As you can see I have added a list element after the form using the same copy that Brent provided. I thought it give some context to the users. However I'm not sure if they apply to every cobrand, specially the subscription point. (Do you want to leave this list as the default for everyone or should I add a -> IF file exist in a cobrand add it?)
2. I think your initial comment on Freshdesk is a good idea if we can resemble or even use the styling that we already have on the FMS front page and apply to WW would make it more maintainable and less code needed. I tried to do that as much as I could but I got stuck trying to add the same format to the WW form.

For example the FMS form displays the following:
Screenshot 2023-06-06 at 10 36 14

While the WW one displays this:
Screenshot 2023-06-06 at 10 36 00

I tried to change it myself, but in the WW one the form comes from the following line
`[% PROCESS form %]` while in FMS is quite different. 

Can you show me how to change the content of the WW form? or not sure if you even think is a good idea.

Also I have added some commits that are exclusive to some of the cobrands, I erased some of the code that is not necessary anyomore. Let me know if you want me to rebase those commits and just include one.

lucascumsille:
@dracos Regarding Peterborough there hasn't been any changes to it. They use their own template. Do you want me to do something about that?

lucascumsille:
@dracos Thanks for the feedback, Matthew. Sorry the late reply I didn't get a notification.

Probably my preferred option would have been to replicate the same form that we currently have in the homepage and add it to WW. That way It I could erase all the unnecessary code added on the WW cobrands, like Kingston, Brent and Sutton.

I tried to do that today, but I had some troubles making the form work. In the homepage when adding a postcode takes you to the map-page, but in WW should take you to select property. In my case when going for the second step it kept showing the add your postcode.

The commit that I just added goes for the last option:
>What you've done at the moment is enough - providing the container that makes it display more like the front page, and we go with that for now.

But I have included two lines to improve the spacing.

Now I would probably propose the following options:

1. If we are too tight with time and we want to deploy this as soon as possible because of Brent. I would leave it as it is, merge and deploy. 
2.Or we edit the `"PROCESS form"` to share the same mark up as the current homepage. Presumably that way we wouldn't need to alter the current logic of WW: Postcode -> Select address

Let me know what you think

dracos:
"I tried to do that today, but I had some troubles making the form work. In the homepage when adding a postcode takes you to the map-page, but in WW should take you to select property. In my case when going for the second step it kept showing the add your postcode." - sounds like you only needed to change the action of the form, I think? If you get it looking okay, I'm happy to fix the functionality if that's all that's not working :) 
If you've already tried it and have that still somewhere, might as well do that if we think it'll be nicer long term. Happy to take a look at what you've done to see what bits need adapting.

lucascumsille:
> "I tried to do that today, but I had some troubles making the form work. In the homepage when adding a postcode takes you to the map-page, but in WW should take you to select property. In my case when going for the second step it kept showing the add your postcode." - sounds like you only needed to change the action of the form, I think? If you get it looking okay, I'm happy to fix the functionality if that's all that's not working :)
> If you've already tried it and have that still somewhere, might as well do that if we think it'll be nicer long term. Happy to take a look at what you've done to see what bits need adapting.

Thanks Matthew =) I tried copying the exact mark up for the form that WW current has:
`        <form method="post" id="postcodeForm" class="js-geolocate">
          <label for="pc">[% loc('Bins, rubbish and recycling:') %]</label>
          [% INCLUDE 'around/_postcode_form_examples.html' %]
          <div>
              <input type="text" name="pc" value="[% pc | html %]" id="pc" size="10" maxlength="200" required aria-describedby="pc-hint">
              [% INCLUDE 'around/_postcode_submit_button.html' attr='id="sub"' %]
          </div>

          [% IF partial_token %]
              <input type="hidden" name="partial" value="[% partial_token.token %]">
          [% END %]
      </form>`
But it kept throwing this error:

Screenshot 2023-06-27 at 11 16 56

- I think it would be nicer if we can reuse the same classes we currently use on FMS, that way I can then erase all the CSS related to the WW front page, that won't be necessary anymore. At the same type we keep the same consistency and styling between FMS and WW.

dracos:
Change `name="pc"` to `name="postcode"` - the waste form has a different name for the field, that should be all.

You can't use the same examples as the front page, because only postcode works here, so might want to drop that include.
You can also drop the partial IF lines, that's not relevant to waste.

lucascumsille:
> Change `name="pc"` to `name="postcode"` - the waste form has a different name for the field, that should be all.
> 
> You can't use the same examples as the front page, because only postcode works here, so might want to drop that include. You can also drop the partial IF lines, that's not relevant to waste.

Thanks for that, @dracos , I just pushed a commit, now it is not throwing an error, but it's not moving forward to the "select address" section.

lucascumsille:
Amazing =) @dracos  then I'll include a another commit erasing the unnecessary CSS for WW cobrands.

lucascumsille:
@dracos I just added commits that get rid of the unnecessary CSS for:

- Brent
- Kingston
- Sutton
- _waste.scss

Once the form is ready let me know and I'll test the cobrands, especially Kingston I need to make sure that it doesn't impact other pages.

dracos:
Pushed a commit to fix the form :)

lucascumsille:
@dracos I just added a commit to fix how we display the errors. I moved to the top and wrap it in a container. This avoids clashing colours when `#front-main` has a different colour.

Screenshot 2023-06-29 at 12 07 31

templates/web/base/waste/index.html
```diff
@@ -1,9 +1,8 @@
 [% PROCESS 'govuk/fields.html' %]
 [% pre_container_extra = BLOCK %]
-
+    [% PROCESS errors %]
```
dracos, line 3: It probably doesn't make much difference, but rather than have the div class="container" in the error block (and affecting everywhere else the error block is used, and I'm not sure how - looks like it'd be container inside container on a normal form page), could that be put here instead, and leave fields.html alone?

dracos: (And as this is very unlikely to need to appear here, happy to keep the bottom margin the same for its display on other pages also)

lucascumsille: No, that's actually fine and more simple. That was my first approach, but I leaves an empty div in case there is no error(Which it's not terrible or a deal breaker). I tried replicating this:
```
[% IF error %]
    <p class="form-error">[% error %]</p>
[% END %]

// And leave it like:
 
[% IF error %]
    <div class="container">[% PROCESS errors  %]</div>
[% END %]

```
But it didn't work.

PR #4476, opened 2023-06-06
3356 geolocate button

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3356

This PR moves the geolocate button from outside the `#postcodeForm` element and relocate it inside the form. This change applies to all cobrands that don't have a specific file`postcode_form.html` for their cobrands.

The logic behind this change:

- We had some cobrands asking to include that button inside the form. Especially in cases where the `#postcodeForm` had a different background colour it look bit weird. From a UI perspective it was a bit off and from a UX perspective gave the impression that the button didn't belong to the form.
- In terms of maintainability it also allow me to get rid of unnecessary code and managing this button with variables. Currently this button uses `$geolocation-link-border` and `$geolocation-link`. For accessibility purposes and to cover cases where the background colour clashes with the font colour I added an extra variable: `$geolocation-link-background-color`

### `postcode_form.html`

- There are two cobrands that are using `postcode_form.html`. Camden and Oxfordshire the only difference they have with the default one is that they don't have this:
`        <div class="hidden js-continue-draft">
            <p>You have a draft report made while offline:<br>
            <a href="#" class="btn continue-draft-btn">Continue draft report</a>
        </div>`

Is this important? Because potentially I could just erase their file and use the default one.

### Important
This PR covers most cobrands, but there are some that don't use the variables of the Geolocate button:
- Merton
- Brent
- Camden
- Kingston
All of them have a special styling or use "secondary button" styling. So as an idea we could make that the default for all cobrands. I imagine we could implement it including more variables for the geolocate button. Let me know what you think and I'll add another commit.

### Preview of all cobrands:
[Desktop I.zip](https://github.com/mysociety/fixmystreet/files/11665490/Desktop.I.zip)
[Desktop II.zip](https://github.com/mysociety/fixmystreet/files/11665492/Desktop.II.zip)
[Mobile.zip](https://github.com/mysociety/fixmystreet/files/11665496/Mobile.zip)


lucascumsille:
@dracos It's on a draft state because I'd like your input regarding some point above. Let me know if it's clear enough, happy to jump on a call.

dracos:
This looks good :) Replied on the template question. Regarding the secondary button styling - I think this is probably enough as it is? Can't think of a particular easy way to say 'use your secondary button styling' with a variable, and it's only one line to do so in the cobrands you list at least.

lucascumsille:
Thanks for the feedback @dracos I just pushed a commit to get rid of highwaysengland `postcode_form`

I think we would probably benefit from merging this PR before https://github.com/mysociety/fixmystreet/pull/4504
 in the #4504 some cobrands have problems with the contrast of the geolocate button, because is still sitting outside the form. 

Also do you want me to rebase all the cobrands commit into one?

dracos:
Yep, I think everything here can be squashed into in one commit, if it's got a good explanation, perhaps a first line like the first commit "Moved geolocate button front page to inside form" and then you could include much of your explanation from the PR text (the beginning and the logic section, I mean) in the commit message body.

dracos (approved):
Happy to look at it when squashed

lucascumsille:
Thanks @dracos I just rebase everything to one commit. I also included an explanation, but let me know if you want me to add something else to it.

dracos (approved):
Looks good, what a nice lot of custom overrides being removed :)
Can rebase on top of current master before merging :)

PR #4477, opened 2023-06-06
[Glos] Contact and privacy information

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3653.

- We hide the fixmystreet.support@gloucestershire.gov.uk email from
    public view (this includes the error message shown when contact form
    submission fails)
    
- We direct users to GCC's external privacy page

There is also a commit (4e02a79d9) in mysociety-servers -> `glos-confirm-integration` branch to add `contact_email` to the staging config.

[skip changelog]


davea (approved):
👍 Looks good!

PR #4479, opened 2023-06-08
Add bin/repl script

chrismytton:
This is a very simple REPL for making quick database queries via the perl interface, without having to go via bin/cron-wrapper.

It defines top-level convenience methods for all `FixMyStreet::DB::Result` subclasses, so you can do `Problem->search` instead of `FixMyStreet::DB->resultset('Problem')->search`.

Adding a semi-colon to the end of a line suppresses the output, which is useful when you get back a big database result object that you don't want dumped to the terminal.

## Example

    $ bin/repl
    FMS> $p = Problem->find(id => 42);

    FMS> $p->get_extra_metadata
    {}
    FMS> $p->set_extra_metadata(foo => 'bar')
    {'foo' => 'bar'}
    FMS> $p->update;

    FMS> Comment->search({ state => 'confirmed' });

    FMS> User->search({ email => 'user@example.com' });

### DIFF
--- a/bin/repl
+++ b/bin/repl
@@ -1,0 +1,71 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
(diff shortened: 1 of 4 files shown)


dracos (commented):
This looks good to me, what more do we want before merging? :)

PR #4486, opened 2023-06-14
Fix overlapping UI on mobile ‘select road’ map screen

davea:
Fixes https://github.com/mysociety/societyworks/issues/3637

[skip changelog]

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -187,11 +187,12 @@ $.extend(fixmystreet.utils, {
             lonlat = new OpenLayers.LonLat(lonlat.lon, lonlat.lat);
         }
 
+        var draggable = !$("html").hasClass("mobile");
         if (fixmystreet.page == 'new') {
-            /* Already have a pin */
+            fixmystreet.markers.features[0].attributes.draggable = draggable;
             fixmystreet.markers.features[0].move(lonlat);
         } else {
-            var markers = fixmystreet.maps.markers_list( [ [ lonlat.lat, lonlat.lon, fixmystreet.pin_new_report_colour ] ], false );
+            var markers = fixmystreet.maps.markers_list( [ [ lonlat.lat, lonlat.lon, fixmystreet.pin_new_report_colour, undefined, undefined, undefined, draggable ] ], false );
             fixmystreet.bbox_strategy.layer.protocol.abort(fixmystreet.bbox_strategy.response);
             fixmystreet.bbox_strategy.deactivate();
             fixmystreet.markers.removeAllFeatures();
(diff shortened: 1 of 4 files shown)


web/js/map-OpenLayers.js
```diff
@@ -188,10 +188,11 @@ $.extend(fixmystreet.utils, {
         }
 
         if (fixmystreet.page == 'new') {
-            /* Already have a pin */
+            fixmystreet.markers.features[0].attributes.draggable = !$("html").hasClass("mobile");
             fixmystreet.markers.features[0].move(lonlat);
         } else {
-            var markers = fixmystreet.maps.markers_list( [ [ lonlat.lat, lonlat.lon, fixmystreet.pin_new_report_colour ] ], false );
+            var draggable = !$("html").hasClass("mobile");
```
dracos, line 194: Could we set this above line 190, then it's available to both sides of the if?

web/js/map-OpenLayers.js
```diff
@@ -233,7 +234,7 @@ $.extend(fixmystreet.utils, {
             }
             var id = pin[3] === undefined ? pin[3] : +pin[3];
             var marker_size = (id === window.selected_problem_id) ? selected_size : size;
-            var draggable = (id === window.selected_problem_id) ? true : (pin[6] === false ? false : true);
+            var draggable = (pin[6] === undefined && id === window.selected_problem_id) ? true : (pin[6] === false ? false : true);
```
dracos, line 237: I don't think pin[6] can ever be undefined any more, it's always set to a boolean by map.js. So I think this breaks the behaviour where you e.g. go to an /reports all reports page, click a report, that pin becomes draggable, pan the map a bit, markers reload, this function will be called, and it won't be draggable any more.
Was there something being fixed by adding this? I think the behaviour should be okay if left as it was?

dracos (changes requested):
Looks good, one bug in the dragging bit I think and one tiny optimisation.

web/js/map-OpenLayers.js
davea: So I think what was happening is that `id === window.selected_problem_id` was `true` when both were `undefined`, meaning the pin was draggable even if `pin[6]` was `false`. This was an attempt to fix that, but perhaps there's a better approach?

dracos: `id && id === window.selected_problem_id`?

davea: 🤦 seems so obvious I'm not sure why I didn't try that already! Will give it a go, thanks.

dracos (approved):
Woo

PR #4487, opened 2023-06-15
[IsleOfWight] Fixed accessibility issues with colours

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3151

I based the new colours on these pages:
- https://www.iow.gov.uk/
- https://islandroads.com/current-works/
I don't think they use cyan anymore, which is good because the colour contrast wasn't accessible with white colour.

### Preview
Screenshot 2023-06-15 at 06 58 27
Screenshot 2023-06-15 at 06 56 26
Screenshot 2023-06-15 at 06 58 03

[Skip changelog]


lucascumsille:
Related to:
https://github.com/mysociety/societyworks/issues/3268
I'm mentioning this ticket as well, because one of the ideas that came on the design sprint was to turn some of `#front-main` and postcode into variables. Like the background color and the `.form-hint` colour, that is causing issues in Bristol. Because all of these would involve more work(erasing  not longer needed code) + testing all the cobrands. Before moving ahead I'd like to know your thoughts about this. Going ahead with the variables option would mean that this PR would not longer be needed.

dracos:
Hi @lucascumsille - yes, think as we already discussed, I would be happy for this and #4488 to be combined and for the front-main/form-hint colours to be done using variables/simplified (perhaps existing variable can be used as better defaults)  if that would reduce custom CSS needed, make future cobrands easier, etc.

Also happy for this and #4488 to be merged first to fix the immediate issues, then your suggestion done on top with them already included.

lucascumsille:
> Hi @lucascumsille - yes, think as we already discussed, I would be happy for this and #4488 to be combined and for the front-main/form-hint colours to be done using variables/simplified (perhaps existing variable can be used as better defaults) if that would reduce custom CSS needed, make future cobrands easier, etc.
> 
> Also happy for this and #4488 to be merged first to fix the immediate issues, then your suggestion done on top with them already included.

Thanks @dracos  I have the time, so I'll turn them into variables. Now that we are into this topic is it worth it to also make into variables:

### Background colour
- Mobile background property for the `#front-main #postcodeForm`) element.
- While for Desktop: background property of `front-main-container` (I would leave by default a transparent value)
Basically two separate variables, in some cobrand will be the same colours, but not for all of them, that's why I would keep them separate.

### Padding
Additionally, but with this one not sure if I'm overcomplicating, it's related with the above, sometimes we have cobrands like Kingston which, want the  background property of `front-main-container` to have a different colour to the body, but if I change the colour it also means that I need to include some padding. is it worth it to also include a variable for the padding and just have by default 0?

dracos:
Background colour - if that makes sense and again reduces the amount of overrides, then sure :)

Padding - Kingston doesn't have a front page (it's only waste), so not sure that matters :) If it's always that option - if the front-main-container has a different colour than the default, then it needs some padding, then it sounds like you could do that with an `if` statement rather than another variable? e.g. if (background colour is set) { padding: 2em 0; } or something like that, would that work?

lucascumsille:
> Background colour - if that makes sense and again reduces the amount of overrides, then sure :)
> 
> Padding - Kingston doesn't have a front page (it's only waste), so not sure that matters :) If it's always that option - if the front-main-container has a different colour than the default, then it needs some padding, then it sounds like you could do that with an `if` statement rather than another variable? e.g. if (background colour is set) { padding: 2em 0; } or something like that, would that work?

Clever =) yeah the `if` makes more sense, there shouldn't be much variation between paddings. I'll move ahead and do the above.

PR #4492, opened 2023-06-19
[Camden] Add missing "not on a road" message

chrismytton:
For some reason this template got missed, which was part of the reason the layer preventing reports made off-road wasn't working.

The other part of this fix is commit `265b941e8d5ddd6d131426ddae9ec8b3cb6a357e` in the servers repo.

For FD-3055

### DIFF
--- a/templates/web/fixmystreet.com/report/new/roads_message.html
+++ b/templates/web/fixmystreet.com/report/new/roads_message.html
@@ -10,7 +10,7 @@
         other side of the border.
         </p>
     </div>
-    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-bucks js-roads-centralbeds js-roads-shropshire">
+    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-bucks js-roads-centralbeds js-roads-shropshire js-roads-camden">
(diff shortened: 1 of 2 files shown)


templates/web/fixmystreet.com/report/new/roads_message.html
```diff
@@ -10,7 +10,7 @@
         other side of the border.
         </p>
     </div>
-    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-bucks js-roads-centralbeds js-roads-shropshire">
+    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-bucks js-roads-centralbeds js-roads-shropshire js-roads-camden">
```
dracos, line 13: Doesn't really matter, but note the reason the other councils here have this class (and matching no_asset_msgs_class set in the config) is because they have multiple 'road' error templates here, all of which need hiding when one is shown (so it doesn't end up showing two), the normal ID based one isn't enough. So Camden doesn't actually need this class or the no_asset_msgs_class entry in the config, as it only has the one road error template that could be shown.

dracos (approved):
The class name isn't needed, I think, but no harm in having it if it's already in the server config :)

PR #4494, opened 2023-06-21
[Echo] Check category on missed bin templates.

dracos:
Otherwise it could match a response template with the same ID from a different backend. [skip changelog]
Fixes FD-3066

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandEcho.pm
+++ b/perllib/FixMyStreet/Roles/CobrandEcho.pm
@@ -193,6 +193,7 @@ sub waste_task_resolutions {
         if ($resolution_id) {
             my $template = FixMyStreet::DB->resultset('ResponseTemplate')->search({
                 'me.body_id' => $self->body->id,
+                'contact.category' => [ 'Report missed collection', undef ],
                 'me.external_status_code' => [
                     "$resolution_id,$task_type_id,$state",
                     "$resolution_id,$task_type_id,",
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -193,6 +193,7 @@ sub waste_task_resolutions {
         if ($resolution_id) {
             my $template = FixMyStreet::DB->resultset('ResponseTemplate')->search({
                 'me.body_id' => $self->body->id,
+                'contact.category' => [ 'Report missed collection', undef ],
```
davea, line 196: Any reason this couldn't filter according to the `type` extra metadata value on the contact? Would be one fewer place to have to update hardcoded string if more Echo categories in future.

davea (approved):
Looks good! One small query about improvement, but I may have overlooked a reason why that's not possible.

perllib/FixMyStreet/Roles/CobrandEcho.pm
dracos: The template here is only to display a 'locked out' message for why you can't report a missed bin, so it's only relevant to that category. Given what the issue is here (and it probably couldn't happen), but didn't want to have same thing in some time where the same resolution code can apply to a new container request and a missed collection report and the admin had been set correctly with different templates for both but it picked up the wrong one.

PR #4495, opened 2023-06-26
[Brent] Street cleaning on red routes

dracos:
[skip changelog] Fixes FD-3072.
Adds the relevant Brent group to the cleaning list, and allows through reports if they're made on a red route.


chrismytton (changes requested):
Code looks good, but would be nice to have a test for the JavaScript bits if possible 🙂

PR #4496, opened 2023-06-27
[Brent] Update pin colours

chrismytton:
Matches the colours currently used by Buckinghamshire Council.

Fixes https://github.com/mysociety/societyworks/issues/3756


dracos (changes requested):
Shouldn't path_to_pin_icons change as well? The ticket says the pins should be identical to Bucks, this changes the colours in use but not the Bucks icons/exact colours.
Also, Brent is one-tier so won't have any reports it doesn't own, so doesn't need the owns_problem part, I think.

dracos (approved):
Looks good

PR #4501, opened 2023-06-29
[fixmystreet.com] Updated Google Analytics.

dracos:
[skip changelog]
Removed all events and whatever the e-commerce stuff was. Happy to add things if we know they're going to be used and how they're being used, but presumably this would all have to be done differently and I don't know that any of it is.


ajparsons:
I'm not qualified to interpret the FMS stuff - the tracking ID is the correct one. 

Only thing seems to be there's a `templates/web/bristol/tracking_code.html` that has the old code, but only seems to affect fixmystreet.com - is this actually used anymore? If not could be dumped as it'll stop working anyway.

PR #4502, opened 2023-06-30
[Peterborough] Small bulky text changes

dracos:
And make location page staff only. And only allow bulky on properties with black bin service.

[skip changelog] Fixes FD-3148 and FD-3150


davea:
Looks like the changes in 4be73cac16246f0658aed369fa94b4f238c1f959 broke the tests

dracos:
Teach me to tidy a line just before pushing, missed a function parameter, should be okay now.

davea (approved):

PR #4504, opened 2023-07-03
Added variables for #front-main

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3770

@dracos I'm leaving it as a draft, because before continue and start tidying up/testing the cobrands. I'd like your thoughts regarding the variables:

### Base.scss
```
$front-main-color: $primary_text !default;
$postcodeform-background: $primary !default;
$form-hint-color: transparentize($front-main-color, 0.2);
```

### Layout
```
$front-main-background: transparent !default;
$front-main-container-background: transparent !default;
$front-main-color: $primary_text !default;
```

As you can see in layout I'm creating two similar variables($front-main-background and $front-main-container-background), but they still can be useful depending on the design of the cobrand:
For example:
`$front-main-background` would cover layouts like this one:
Screenshot 2023-07-03 at 12 20 52
Where:
```
$front-main-background: $whateverColour;
$front-main-container-background: transparent;
```

While $front-main-container-background will cover layouts like this one:
Screenshot 2023-07-03 at 12 23 46
Where:
```
$front-main-background: transparent !default;
$front-main-container-background: $whateverColour;
```

-------------
On another topic. I have added the `$front-main-color:` to handle the text. I've had added to `base.scss` and `layout.scss`, to prevent adding it to both files I thought I could added it to `mixin.scss` instead, but technically that couldn't be correct. This leads me to wonder, do you want me to create a `variables.scss` in the sass folder and move all variables there? Similar to what Bootstrap does. The benefit I can see by doing it is to unify all the variables and make it clearer for anyone styling the cobrands about the available options, before start adding custom code.

Thanks and let me know what you think

[Skip changelog]


dracos:
To base.scss I'd probably add another variable for the h2 colour? That's overridden at least 3 or 4 times.

Will you need a variable for the postcodeForm background in layout as well?

$front-main-color - not sure I understand here, if you have/use the variable in base to set the colour of #front-main, why do you also need it in layout? It'll be used anyway?

lucascumsille:
>just noticed that e.g. https://fms.islandroads.com/alert - I think the .form-hint  you added in e8354534e658d317364b23932d0163965aca67a7 needed to be only inside #front-main  I guess? (It was like that in Bristol fix) (edited) 

@dracos considering the above. Would it be worth it to make both of this boxes (Homepage and Alert page) share the same classes, instead of targeting it them by using ID? These components are similar we could just make them look the same, which reminds of a Camden's feedback that they didn't understand why things like the width of the postcode input was so large in the area page.
I'm thinking something like `postcode-form-box` and leave this class out of the nesting of `#front-main` so it targets both of these boxes.

lucascumsille:
> Will you need a variable for the postcodeForm background in layout as well?

@dracos I didn't think it would be necessary. However Cobrands like Southwark could use a variable like that, so no harm in adding it.

I will add the H2 variable as well.

Regarding the `front-main-color` I wanted something that handles the colour of everything, except the `form-hint` and now H2. However now I realise that will be tricky, because we can have cobrands like Bromley, where on desktop the `#postcodeForm` has one background, but on mobile has another background. At this point I'm not sure if I'm overcomplicating it, but trying to think if a conditional could help tackle this situations. I'll give it a think a comeback with a solution.

dracos:
> Would it be worth it to make both of this boxes (Homepage and Alert page) share the same classes, instead of targeting it them by using ID? These components are similar we could just make them look the same, which reminds of a Camden's feedback that they didn't understand why things like the width of the postcode input was so large in the area page.
I'm thinking something like postcode-form-box and leave this class out of the nesting of #front-main so it targets both of these boxes.

Sounds like it could work, but also don't want this PR creeping up and up as it catches more of these :) But if it's straightforward to do and works, then sure, I don't see why they couldn't both share the same class and act more the same, would presumably be more consistent.

dracos:
> Regarding the front-main-color I wanted something that handles the colour of everything, except the form-hint and now H2. However now I realise that will be tricky, because we can have cobrands like Bromley, where on desktop the #postcodeForm has one background, but on mobile has another background. At this point I'm not sure if I'm overcomplicating it, but trying to think if a conditional could help tackle this situations. I'll give it a think a comeback with a solution.

Yeah, I think having two variables for font colour, one in base, one in layout, even if nearly always the same, is probably clear and saves having a separate variables file then as well?

lucascumsille:
@dracos  to make it easy to visualise I did some testing with Rutland and then created different styling scenarios:

###Scenario 1
Styling `#front-main`
Mobile and Desktop sharing the same styling.
Scenario 1

### Scenario 2
#front-main-container with background
Mobile and Desktop sharing the same styling.
Scenario 2

### Scenario 3
#front-main NO background on Desktop
Mobile and Desktop sharing the same styling.
Scenario 3

### Scenario 4
Very similar to three, but I wanted to show you that we can have different background color on mobile and desktop.
Scenario 4

In order to this I created the following variables:

### Base
```
$front-main-color: $primary_text !default;
$front-main-background: none !default;
$postcodeform-background: $primary !default;
$form-hint-color: transparentize($front-main-color, 0.2);
$front-main-h2-color: transparentize($front-main-color, 0.1) !default;
```

### Layout
```
$front-main-background-desktop: transparent !default;
$front-main-color-desktop: $primary_text !default;
$front-main-container-background-desktop: transparent !default;
$form-hint-color-desktop: transparentize($front-main-color-desktop, 0.2) !default;
$front-main-h2-color-desktop: transparentize($front-main-color-desktop, 0.1) !default; 
```

I know these are a lot of variables however this will help us to cover most of the current of situations for our cobrands, if not all of them. If you are alright with this I can start tidying up each cobrand.
Let me know what you think =)

dracos:
Sounds good to me! If you could also comment the variables, that'd help you when someone comes to do the next one to know which one does what :)

lucascumsille:
> Sounds good to me! If you could also comment the variables, that'd help you when someone comes to do the next one to know which one does what :)

Amazing I'll do that then. Thanks @dracos

lucascumsille:
@dracos I think I finish. I have attached screenshot with all the cobrands. I know some of them have only WW, but I have included on the FMS anyways.

Let me know what you think and if you want me to rebase.

Fixes: https://github.com/mysociety/societyworks/issues/3770

### FMS Preview
[FMS Part 1.zip](https://github.com/mysociety/fixmystreet/files/12015621/FMS.Part.1.zip)
[FMS Part 2.zip](https://github.com/mysociety/fixmystreet/files/12015622/FMS.Part.2.zip)


### WasteWorks Preview
[WW.zip](https://github.com/mysociety/fixmystreet/files/12015637/WW.zip)

templates/web/base/alert/index.html
```diff
@@ -5,14 +5,14 @@ <h1>[% loc('Local RSS feeds and email alerts') %]</h1>
 
 [% PROCESS 'alert/_index_text.html' %]
 
-<form method="get" action="/alert/list" class="form-box js-geolocate">
+<form method="get" action="/alert/list" class=postcode-form-box js-geolocate">
```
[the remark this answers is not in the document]
dracos: Does this want to have an id="postcodeForm" to pick up any special cases using that? Or does that not matter?

templates/web/base/around/postcode_form.html
```diff
@@ -1,46 +1,48 @@
 <div id="front-main">
-    <div id="front-main-container">
-        [% UNLESS possible_location_matches %]
-            [% INCLUDE 'around/intro.html' %]
-        [% END %]
+    <div class="container">
```
dracos, line 2: Why is the container element now wrapped in another container element? It now has #front-main-container having a max-width of $container-max-width, and .container also has max-width $container-max-width. So I'm confused as to why both are necessary or what is happening here.

templates/web/bristol/around/intro.html
```diff
@@ -1,2 +1,2 @@
 <h1>Report a problem</h1>
-<p>Report a problem on a street, car park, park, green space or harbourside area in Bristol or look at and discuss problems that have already been reported.</p>
+<h2 style="font-size:1em">Report a problem on a street, car park, park, green space or harbourside area in Bristol or look at and discuss problems that have already been reported.</h2>
```
dracos, line 2: I'm not sure why this change has been made. Presumably Bristol's current look is what they have asked for, and it seems to remain consistent as a `p` - ideally nothing should visually change as a result of this PR, I assume?

web/cobrands/sass/_layout.scss
```diff
@@ -16,6 +16,18 @@ $container-max-width: 60em !default;
 $primary_link_color: null !default;
 $primary_link_hover_color: null !default;
 
+// None of the form variables control the mobile styling. For that refer to _base.scss
+// $front-main-color-desktop controls the colours of all elemets inside #front-main and .postcode-form-box element.
+$front-main-color-desktop: $primary_text !default;
+// $front-main-background-desktop controls the background of the of the #front-main element. This is a full width element.
+$front-main-background-desktop: transparent !default;
+// $front-main-container-desktop changes the background of h1, h2, but without using the full width.
+$front-main-container-background-desktop: $front-main-background-desktop !default;
```
dracos, line 25: Should this be `transparent` or `none`, not front-main-background-desktop? As is, it'll duplicate what's 'underneath' and the padding check in layout will always pass and add padding even when not necessary (e.g. this is happening on Bexley, .com, etc).

web/cobrands/sass/_layout.scss
```diff
@@ -821,33 +833,38 @@ textarea.form-error {
...
   h2 {
     font-style:normal;
     margin:0;
-    color: inherit;
-  }
-  #postcodeForm {
-    color: inherit;
-    background:none;
-    overflow:hidden;
-    padding-bottom: 0;
```
dracos, line 841: Why was this dropped? e.g. Bexley has a lot more bottom padding now than it used to.

web/cobrands/sass/_layout.scss
```diff
@@ -821,33 +833,38 @@ textarea.form-error {
 
 // big yellow bit full screen width
 #front-main {
-  color: $primary_text;
-  background: $primary;
-  margin: 0;
+  color: $front-main-color-desktop;
+  background: $front-main-background-desktop;
   padding: 1em;
+  margin: 0 auto;
```
dracos, line 839: Why the change here from 0 to 0 auto?

web/cobrands/sass/_base.scss
```diff
@@ -2682,6 +2660,49 @@ label .muted {
...
 }
 
+.postcode-form-box {
+  margin: 0 -1em;
+  padding: 1em;
+  color: $front-main-color;
+  background: $postcodeform-background;
+  font-family: $meta-font;
+  label {
+    margin-top: 0;
```
dracos, line 2670: Do you have an example where this makes a difference?

dracos (changes requested):
Variables all look good and again nice to see customisations not being needed any more. Got a number of queries about the changes, hopefully nothing too bad :)

web/cobrands/sass/_base.scss
```diff
@@ -2400,8 +2410,8 @@ label .muted {
 
 .form-hint {
   display: block;
-  color: #666;
-  margin-bottom: 0.5em;
+  color: $form-hint-color;
+  margin-bottom: 0.5em !important;
```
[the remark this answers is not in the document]
dracos: I think I'd prefer if this went back to not having `!important`, and we added a `#front-main p.form-hint` which at least then shows why it's there and what it's overriding, if that's okay?

web/cobrands/sass/_layout.scss
```diff
@@ -821,33 +833,38 @@ textarea.form-error {
...
+.postcode-form-box {
+  color: $front-main-color-desktop;
+  background:$postcodeform-background-desktop;
+  overflow:hidden;
+  label {
+    margin:0.5em 0;
+  }
+  .form-hint {
+      color: $form-hint-color-desktop;
+      margin-top: -0.25em;
```
dracos, line 864: This isn't matching, same reason as what I've just outlined about base margin. Again, I think a `#front-main p.form-hint` would be easiest/clearest here, probably?

templates/web/base/alert/index.html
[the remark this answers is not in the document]
dracos: Same ID to have the same behaviour I assume (although as you say if there isn't any, doesn't really matter)? They don't ever appear on the same page after all.

web/cobrands/sass/_base.scss
lucascumsille: Yes, top of my mind Camden and Bromley

web/cobrands/sass/_layout.scss
lucascumsille: sorry not sure I get this one. If I change the colour using the variable it changes in the homepage and alert page.

lucascumsille:
@dracos I have added commits that tackle the issues above =)
Let me know if there is any further feedback

web/cobrands/sass/_layout.scss
lucascumsille: lol, sorry. It makes sense I'll add a commit now

templates/web/base/around/postcode_form.html
dracos: The waste template still wraps front-main-container in a `.container`.

web/cobrands/sass/_layout.scss
```diff
@@ -159,10 +174,21 @@ h1 {
 
 // .content Is the white box
 .content {
-  margin: 1em 0 0 0;
```
dracos, line 162: Why have we lost the 1em top margin of global `.content`? e.g. the help page on Bromley is now squashed up against the heading when it wasn't before. And then in frontpage you're putting it back, so I don't think it should be removed here. Think this might have happened because of a change then a not identical change back in the fixup.

dracos (changes requested):
I'm sure these are all consequential changes, but some of them feel a little unrelated at the moment :) Sometimes it's a bit hard to match the fixes up with the original changes, would be good if each fix is in its own commit linked to what it's changing. I've seen one more global issue to resolve back (`.content), and waste still has a double container div - once those two are resolved, I'll give it one more overall check with the new if statements etc (which look good if I understand them right). Thanks for making the other changes.

Bromley looks a bit different now to current, for example - in layout, the top padding is 1em more, do yo need to reduce the front-main top padding or otherwise have it better combine with the changes; in base, the postcode form now has a white background whereas it's green on live, I don't know if this was deliberate.

lucascumsille:
Sorry @dracos I have forgotten about this PR. I have added two commits to tackle:
>Bromley looks a bit different now to current, for example - in layout, the top padding is 1em more, do yo need to reduce the front-main top padding or otherwise have it better combine with the changes; in base, the postcode form now has a white background whereas it's green on live, I don't know if this was deliberate.

web/cobrands/sass/_layout.scss
lucascumsille: @dracos I have included a commit to fix this.

dracos:
I have rebased and merged in some fixup commits (e.g. the Bromley fixes) to where they would want to go. I'm afraid looking through a few cobrands there are changes to the look; this PR was meant to introduce variables and (after discussion) make the alert postcode form be the same as the front page form, it shouldn't have changed anything else visually. But eg. Bucks has a button style that is to have a bottom darker green line (you can see this on their own site). This was removed in commit "Erased box-shadow for button" without explanation, and doesn't seem to have anything to do with  the purpose of this PR. I'm going to drop that commit (and fix the box-shadow:0 to box-shadow:none - perhaps that was the reason for it?). Bucks (and Bexley, and Brent, I'll keep looking through) also has additional top padding spacing on the front page, the same as that I raised for Bromley previously. If we want to have more top space, sure, we could do that, but we don't want to do that mixed in with introducing a whole bunch of refactoring and variables, it will just be confusing. `.form-hint` has lost its default 0.5em margin bottom across the admin.

templates/web/base/around/postcode_form.html
```diff
@@ -1,46 +1,48 @@
...
-                [% INCLUDE 'around/_postcode_submit_button.html' attr='id="sub"' %]
+            [%
+                question = c.cobrand.enter_postcode_text || loc('Enter a nearby street name and area');
+            %]
+
+            <div class="hidden js-continue-draft">
+                <p>You have a draft report made while offline:<br>
+                <a href="#" class="btn continue-draft-btn">Continue draft report</a>
             </div>
+            <form class="postcode-form-box" action="[% c.uri_for('/around') %]" method="get" name="postcodeForm" id="postcodeForm" class="js-geolocate">
```
dracos, line 16: This form has two `class` now.

dracos:
Okay, I think this is ready now? :) I've tidied the template changes (and thereby removed the Oxfordshire override templates); I've compared every cobrand by eye and made a number of fixes. I've added the necessary variables for Gloucestershire. Attached are two diff files - one contains the core changes I made, the other the cobrand changes I made. Happy to explain or discuss any of the changes
[sass-changes.txt](https://github.com/mysociety/fixmystreet/files/12896868/sass-changes.txt)
[cobrand-changes.txt](https://github.com/mysociety/fixmystreet/files/12896869/cobrand-changes.txt)

lucascumsille:
Thanks for all you help on this one @dracos. It was a big one =)
Just in case I checked the cobrands and desktop and mobile and they are looking fine.
I agree let's do the extra spacing on a separate PR, I think in some cobrands are looking a bit tight between the `front-main` and the navigation bar.

PR #4357, opened 2023-03-30
[Buckinghamshire] Send claims to Evo API

chrismytton:
Adds a `bin/buckinghamshire/send-claims-to-evo` script, which looks for unsent claims (i.e. ones that don't have `sent_to_evo` in their extra data), then sends the claim along with any associated files and photos to the EvoClaim API.

In order to test this the following needs to be added to `conf/general.yml`. Replace **app_id** and **api_key** with the values from the staging YAML config on the servers.

```yaml
COBRAND_FEATURES:
  # Other bits...
  dwf_evo_claim:
    buckinghamshire:
      base_url: https://buckinghamshireapi.test.evosuite.net
      app_id: APP_ID
      api_key: API_KEY
      earliest_claim_date: 2023-05-12
```

I've added a cron job to run this script on staging every 5 minutes. See `6ac787a` and `046edc9` in the servers repo for the cron script changes.

Fixes https://github.com/mysociety/societyworks/issues/3361

## Related documentation

[EvoClaim API Functional Spec v0.1](https://docs.google.com/document/d/1ly4KOcs1DJ13clHp0mwQd7VfVQPwXPsx/edit)

PR #4506, opened 2023-07-03
[Buckinghamshire] Turn off claims emails

chrismytton:
Bucks no longer need claims going to email now we're sending claims to EvoClaim (#4357).

This change separates the claim auto-response template code from the claims email sending code and then removes the claim email sending code entirely.

Also updated a couple of comments while I was in the file.

Fixes https://github.com/mysociety/societyworks/issues/3768

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -181,11 +181,11 @@ sub open311_pre_send {
     my ($self, $row, $open311) = @_;
     if ($row->category eq 'Claim') {
         if ($row->get_extra_metadata('fault_fixed') eq 'Yes') {
-            # We want to send to Confirm, but with slightly altered information
+            # We want to send via Open311, but with slightly altered information
             $row->update_extra_field({ name => 'title', value => $row->get_extra_metadata('direction') }); # XXX See doc note
             $row->update_extra_field({ name => 'description', value => $row->get_extra_metadata('describe_cause') });
         } else {
-            # We do not want to send to Confirm, only email
+            # We do not want to send via Open311, only email
             return 'SKIP';
         }
     }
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -194,6 +194,8 @@ sub open311_pre_send {
 sub open311_post_send {
     my ($self, $row, $h) = @_;
 
+    $self->_add_claim_auto_response($row, $h) if $row->category eq 'Claim';
```
dracos, line 197: I /think/ this can be removed entirely, if the Claim template is made auto-response, then the default handling should work? This must be a hangover from before that changed to create-on-creation but confirm-on-submission, I think. But comparing the code here with what create_related_things does, it appears to be identical apart from requiring the Claim template to have 'auto response' ticked.

dracos (changes requested):
I think the special-update code can be dropped entirely in preference to the existing update code, which now acts in the same way as this wants it to, I think.
The test /would/ then have an (unconfirmed) update before the sending, and the sending would then confirm it.

perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
chrismytton: @dracos I've had a dig around and it seems that the tests are expecting the claims response template to be included in the initial email, rather than be sent out separately as an update:

https://github.com/mysociety/fixmystreet/blob/0d8725ce773d08a4145b9f0dc04cfa2e0e3002f0/t/app/controller/claims.t#L221-L224

(The "is a lengthy process" bit comes from the claims response template in the tests)

The `_add_claim_auto_response` code is adding the update to `$h->{update}` which is then picked up by `templates/email/buckinghamshire/other-reported-claim.{html,txt}` to be included in the confirm_report_sent email, and then marking the alert as sent so it's not sent out again.

https://github.com/mysociety/fixmystreet/blob/0d8725ce773d08a4145b9f0dc04cfa2e0e3002f0/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm#L250-L268

See https://github.com/mysociety/fixmystreet/commit/7aa154e565cc6e4eafc5884d5c2b6d4f6e781dc5 for the commit which introduced this change.

So I _think_ this code is still needed, in order to include the update in the initial email and to skip the subsequent alert.

dracos: (I wonder if the code could be combined a bit then, but no need at this point)

dracos (approved):

PR #4509, opened 2023-07-05
[Bulky] Refactor code

dracos:
The two main commits move bulky code to their own files, and don't change anything code-wise; two small commits are little issues spotted along the way, final one is storing previous data on resend that might be useful in future (both here and in other things).


templates/web/base/report/update/moderation_photo_diff.html
```diff
@@ -1,4 +1,4 @@
-[% IF added %]
+[% IF added.size %]
```
mroderick, line 1: I think the fix in this commit is great and could be in it's own pull request. That would make it easier to understand and keep it separate from the refactoring in this PR.

mroderick (commented):
I'll start my contributions with just one comment :)

davea (approved):
Phew, what a diff! Looks good though. Couple of little suggestions if you have time, but 👍 overall.

templates/web/base/report/update/moderation_photo_diff.html
dracos: I can see how in a number of scenarios that would be useful, but I decided on balance it wasn’t worth the extra bureaucracy in this case.

mroderick: Fair enough

PR #4512, opened 2023-07-10
[Bulky] Allow multiple bookings

dracos:
Fixes https://github.com/mysociety/societyworks/issues/3780 [skip changelog]
First adds ID to cancel URLs (so that you'll be able to choose which to cancel), then lists all bookings on bin day page (even though there's only one at this point), and then allow multiple bookings.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -100,8 +100,8 @@ sub item_list : Private {
 sub index : PathPart('') : Chained('setup') : Args(0) {
     my ($self, $c) = @_;
 
-    # TODO Remove when allowing more than one booking
-    if ($c->stash->{pending_bulky_collections}) {
+    my $cfg = $c->cobrand->feature('waste_features');
+    if ($c->stash->{pending_bulky_collections} && !$cfg->{bulky_multiple_bookings}) {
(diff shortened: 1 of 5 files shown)


davea (changes requested):
Though dates of existing bookings are greyed out when making a new booking, it is possible to amend a booking to be on the same date as an existing booking:

image

dracos:
> Though dates of existing bookings are greyed out when making a new booking, it is possible to amend a booking to be on the same date as an existing booking

Weird, I added the code to Bulky.pm but not Bulky/Amend.pm - my guess is I was working on both of these at same time, then when I rebased multiple on top of amending I forgot to add this. Will look to see if they can share more of the code here.

dracos:
Okay, added Bulky/Shared.pm in the Amend branch with no changes other than sharing all the form code it can between Bulky.pm and Bulky/Amend.pm - this then meant any pending dates were now disabled on the amend form as well. Then added new bit of code and test to /not/ disable the date of the report being amended.

davea (approved):
Looks good 👍 


One minor consideration is that if the property you're viewing already has a collection you have to sign in before you can see any details about it at all, including the date. But the multiple bookings process discloses the date of the existing booking by way of including as a grayed-out radio button. Wonder if it's worth hiding the date in that case? Probably not, but thought worth mentioning.

PR #4513, opened 2023-07-10
Increase front page thumbnail right padding on mobile

davea:
Having the report thumbnails butted up against the right hand edge of the page looked a little odd, so this commit mirrors the padding from the left hand side for symmetry.

Raised in FD-3161

[skip changelog]


chrismytton (approved):
👍

PR #4514, opened 2023-07-10
[Central Beds] Configure basic Jadu Open311 sending.

neprune:
Closes https://github.com/mysociety/societyworks/issues/3766
Closes https://github.com/mysociety/societyworks/issues/3767
Closes https://github.com/mysociety/societyworks/issues/3744
Closes https://github.com/mysociety/societyworks/issues/3746.


neprune:
Realised I need to add support for `datetime` and `mulitvaluelist` in admin category editing page - doing this now.

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -87,6 +87,14 @@ sub disambiguate_location {
 
 sub enter_postcode_text { 'Enter a postcode, street name and area, or check an existing report number' }
 
+sub open311_config {
+    my ($self, $row, $h, $params, $contact) = @_;
+    if ($contact->email =~ /Jadu/) {
+        $params->{multi_photos} = 1;
```
dracos, line 93: I wonder why this isn't on for all Central Beds? Looks like the open311-adapter symology end can cope with multiple URLs being provided.

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -123,6 +131,35 @@ sub lookup_site_code_config {
 sub open311_extra_data_include {
     my ($self, $row, $h, $contact) = @_;
 
+    if ($contact->email =~ /Jadu/) {
```
dracos, line 134: Do we want to move `$open311_only` creation up above this, then can add the new rows in this if

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -123,6 +131,35 @@ sub lookup_site_code_config {
...
+            },
+            {
+                name => 'title',
+                value => $row->title
+            },
+            {
+                name => 'description',
+                value => $row->detail
+            },
+            {   # TODO: Populate from address.
```
dracos, line 148: I guess having these at the FMS end might make sense for them to look at in the admin, though there's also the possibility of doing the lookup at the open311-adapter end if that makes more sense.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -602,6 +602,8 @@ $.extend(fixmystreet.set_up, {
         }
         if (data.category_extra) {
             $category_meta.replaceWith( data.category_extra );
+            document.dispatchEvent(new Event('category_extra_loaded'));
```
dracos, line 605: At the end of this function is a `report_new:category_change` (jQuery) event being triggered, which you could use instead of a new one?

web/cobrands/centralbedfordshire/flytipping.js
```diff
@@ -0,0 +1,55 @@
+document.addEventListener('category_extra_loaded', function() {
```
dracos, line 1: Or, a second alternative would be to have permanent listeners set up here on a parent that's always present (document, or the form, etc) e.g. `$(document).on('click', '#form_fly_tip_witnessed')` which wouldn't then need to be run on every category change.

dracos (changes requested):
Looks good, a various number of small comments :)

templates/web/base/admin/extra-metadata-option.html
```diff
@@ -13,10 +13,12 @@
     </div>
     <div class="row">
         <div class="col">
+            [% IF type == 'singlevaluelist' %]
```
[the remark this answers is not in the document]
neprune: Have put the beginnings of support for this in `support-stopper-messages-for-checkboxes`, think it just needs stopper checking to be triggered on checkbox select - will get to after the mainline integration work.

templates/web/base/report/new/category_extras_fields.html
```diff
@@ -8,13 +8,31 @@
...
       [% END -%]
       [% IF meta.variable != 'false' %]
-        [% IF meta.item('values').size %]
+        [% IF meta.datatype == 'multivaluelist' %]
+          [% FOR option IN meta.values %]
+            <div class="label-containing-checkbox">
+              <label style="font-weight:normal;" for="[% cat_prefix %][% meta_name %][% option.key %]">[% option.name %]
+                <input class="at-least-one-group group-error-propagate" type="checkbox" id="[% cat_prefix %][% meta_name %][% option.key %]" name="[% cat_prefix %][% meta_name %]" value="[% option.key %]"
+                [% FOREACH key IN report_meta.$x_meta_name.value %][% IF key == option.key %] checked [% LAST %][% END %][% END %]>
+                [%# Iterating through the whole list of values as unable to check if the array contains the key. %]
```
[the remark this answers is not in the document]
neprune: Nope, that works nicely 🙏

web/vendor/jquery.validate.additional-methods.min.js
```diff
@@ -0,0 +1,4 @@
+/*! jQuery Validation Plugin - v1.19.1 - 6/15/2019
```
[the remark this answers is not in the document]
neprune: My bad... 😄

templates/web/base/report/new/category_extras_fields.html
```diff
@@ -8,13 +8,31 @@
...
       [% TRY %][% INCLUDE 'report/new/_category_extra_field_notice.html' %][% CATCH file %][% END %]
       [% IF field_errors.$x_meta_name %]
       <p class='form-error'>[% field_errors.$x_meta_name %]</p>
       [% END -%]
       [% IF meta.variable != 'false' %]
-        [% IF meta.item('values').size %]
+        [% IF meta.datatype == 'multivaluelist' %]
+          [% FOR option IN meta.values %]
+            <div class="label-containing-checkbox">
+              <label style="font-weight:normal;" for="[% cat_prefix %][% meta_name %][% option.key %]">[% option.name %]
```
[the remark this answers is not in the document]
neprune: Nice, also makes the spacing tighter and looks way better

templates/web/base/report/new/category_extras_fields.html
```diff
@@ -20,7 +20,9 @@
       <p class='form-error'>[% field_errors.$x_meta_name %]</p>
       [% END -%]
       [% IF meta.variable != 'false' %]
-        [% IF meta.datatype == 'multivaluelist' %]
+        [% IF meta.datatype == 'datetime' %]
```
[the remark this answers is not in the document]
neprune: Will get to this after the mainline integration work if that's alright

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
neprune: Yeah, does seem fine - have added

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
neprune: Yeah I've been going back and forth on this but probably doing it in open311-adapter where we also have the mapping to hand probably makes more sense - dropping these attributes now

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1344,7 +1344,8 @@ sub set_report_extras : Private {
 
             my $value;
             if ($field->{datatype} eq 'multivaluelist') {
-                $value = $c->req->params->{$param_prefix . $field->{code}} // [];
+                my @params = $c->get_param_list($param_prefix . $field->{code}, 1);
```
dracos, line 1347: Do we want to allow comma based entry here? I assume we're not expecting that as a form of input, that's more for the around query strings passing in multiple status/categories and the like.
You could also do this as one line as `$value = [ $c->get_param_list($param_prefix . $field->{code}) ];`

dracos (changes requested):
Thanks, all changes look good, only the one query about the arguments to get_param_list.

perllib/FixMyStreet/App/Controller/Report/New.pm
neprune: My bad - not needed

dracos (approved):
🛏️

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
dracos: Doh, sorry, my suggestion here broke the code (and test failure showing this) because it needed the change to detail to happen before setting the open311_only. So guess code should be 1. UnitID block, 2. open311_only set, 3. Jadu check.

templates/web/base/report/new/category_extras_fields.html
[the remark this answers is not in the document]
neprune: On second thoughts, I'm not sure about this, it seems like Shropshire and the fixture cats want just a `date` rather than a `datetime` - perhaps this should be its own open311 field type?

PR #4516, opened 2023-07-11
Move form override defaults into cobrand code from templates.

neprune:
Also fixes a bug introduced in #4471 where the detail hint default text would stop displaying for cobrands without a bespoke value set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
     - Front end improvements:
         - Include requirements for redeeming the link in the email change confirmation mail. #4422
         - Use email field type for username if SMS authentication not enabled. #4455
-        - Text overrides for new report fields can be configured to apply when it is known the report will go to a particular cobrand. #4466
+        - Text overrides for new report fields can be configured to apply when it is known the report will go to a particular cobrand. #4466, #4516
     - Bugfixes:
         - Stop map panning breaking after long press. #4423
         - Fix RSS feed subscription from alert page button.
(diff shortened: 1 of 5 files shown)


t/cobrand/southwark.t
```diff
@@ -71,17 +71,8 @@ FixMyStreet::override_config {
...
-                allow_anonymous => 'true',
-                bodies => ['Southwark Council'],
-            },
-            'Bus Stops and Shelters' => {
-                allow_anonymous => 'true',
-                bodies => ['TfL'],
-            },
-        }, "Southwark 'street' area doesn't have River Piers category";
-
+        is_deeply [sort keys %{$json->{by_category}}], ['Abandoned Bike (Street)', 'Bus Stops and Shelters'],
```
dracos, line 74: Hmm, ideally we wouldn't send the default message with every ajax response, seems a waste of bandwidth if it's not necessary (and as we're storing the original already). I like how it's simplified the templates that use them though. Could we perhaps leave the Default.pm functions as they were, returning nothing, and change report/new/_form_labels.html to, I think, this:
```
SET form_title = c.cobrand.new_report_title_field_label.defined ? c.cobrand.new_report_title_field_label.defined : loc("Summarise the problem");
SET form_title_placeholder = c.cobrand.new_report_title_field_hint.defined ? c.cobrand.new_report_title_field_hint : loc("e.g. ‘10 inch pothole on Example St, near post box’");
SET form_detail_placeholder = c.cobrand.new_report_detail_field_hint.defined ? c.cobrand.new_report_detail_field_hint : loc('e.g. ‘This pothole has been here for two months and…’');
```
That would then have the page defaults in the one place, but the ajax calls would be empty as before.

dracos (changes requested):

dracos (approved):

PR #4517, opened 2023-07-12
Add send state to problems (fixes)

dracos:
The first two commits were reviewed in #4505. The two fixups in here deal with issues found before putting live:
* There was an issue with the marking of rows as processed not being saved; and if that had been working, then it would have marked unconfirmed reports as processed
* The DB migration marked all rows as processed, when we need to update any currently unsent rows as unprocessed.


chrismytton (approved):
Ah yes, I didn't consider the currently unsent reports when I wrote the DB migration, sorry!

Couple of minor comments on the test, but other than that looks good I think!

PR #4518, opened 2023-07-12
[Bulky] Allow retrying of failed CSC payments.

dracos:
[skip changelog] Fixes https://github.com/mysociety/societyworks/issues/3781
The form is shown to staff at /waste, as requested by Peterborough.

### DIFF
--- a/templates/web/base/waste/garden/csc_payment_failed.html
+++ b/templates/web/base/waste/garden/csc_payment_failed.html
@@ -3,19 +3,26 @@
 <h1 class="govuk-heading-xl">Payment Failed</h1>
 
 <div class="govuk-panel govuk-panel--confirmation">
-    <h1 class="govuk-panel__title">
-        [% title %]
-    </h1>
     <div class="govuk-panel__body">
-        <p>[% message %]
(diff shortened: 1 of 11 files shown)


templates/web/base/waste/garden/csc_payment_failed.html
```diff
@@ -16,6 +13,13 @@ <h1 class="govuk-panel__title">
         <p>
         No subscription will be created.
```
davea, line 14: Might be confusing, should this be "No booking has been created"?

davea (changes requested):
Entering the reference number on the address page works, but the "Retry booking" button on the property page seems to start the process from scratch with no details prefilled in the report form:

image

dracos:
> Entering the reference number on the address page works, but the "Retry booking" button on the property page seems to start the process from scratch with no details prefilled in the report form:

Doh, I moved the form to the front page from bin days page as they confirmed that's where they wanted it but entirely forgot about the button on the bin days page, sorry!

davea (approved):
💰 ↩️ 🆗

PR #4519, opened 2023-07-13
[FMS] Adds structure for adding asset which overlaps another area

MorayMySoc:
Orginally Bristol parks needed to have responsibility for two parks which overlap a neighbouring area.

Now Brent is adding an asset layer of Brent responsibility which overlaps into neighbouring areas.

This takes the original unmerged code from the Bristol Parks PR https://github.com/mysociety/fixmystreet/pull/4464 and refactors it to adapt responsible_for_areas and new method munge_overlapping_asset_bodies to use a
cobrand hook so that an overlapping asset can be added by putting the overlapped area in the council_area_id array and if report_is_on_cobrand_asset returns truthy, the report in the neighbours area will be reported to the current cobrand. On false, will make the report in the neighbours area.

Brent example based on this https://github.com/mysociety/fixmystreet/pull/4538

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UKCouncils.pm
+++ b/perllib/FixMyStreet/Cobrand/UKCouncils.pm
@@ -171,15 +171,71 @@ sub area_check {
 
     my $councils = $params->{all_areas};
 
-    # The majority of cobrands only cover a single area, but e.g. Northamptonshire
-    # covers multiple so we need to handle that situation.
-    my $council_area_ids = $self->council_area_id;
-    $council_area_ids = [ $council_area_ids ] unless ref $council_area_ids eq 'ARRAY';
-    foreach (@$council_area_ids) {
-        return 1 if defined $councils->{$_};
-    }
+    return 1 if $self->responsible_for_areas($councils);
 
     return ( 0, $self->area_check_error_message($params, $context) );
+}
+
+=head2 responsible_for_areas
+
+Tests to see if the cobrand is responsible for a location. This is done through
+checking if it has the area set in council_area_id and also checking if the
+coverage is limited only to an overlapping asset by testing the hook
+report_is_on_cobrand_asset
+
+=cut
+
+sub responsible_for_areas {
+    my ($self, $councils) = @_;
+
+    if ($self->can('check_report_is_on_cobrand_asset')) {
+        # This will need changing for two tier councils
+        if (scalar(%$councils) == 1 && (values %$councils)[0]->{id} eq $self->council_area_id->[0]) {
+            return 1;
+        } elsif ($self->check_report_is_on_cobrand_asset) {
+            return 1;
+        } else {
+            return 0;
+        }
+    } else {
+        # The majority of cobrands only cover a single area, but e.g. Northamptonshire
+        # covers multiple so we need to handle that situation.
+        my $council_area_ids = $self->council_area_id;
+        $council_area_ids = [ $council_area_ids ] unless ref $council_area_ids eq 'ARRAY';
+        foreach (@$council_area_ids) {
+            return 1 if defined $councils->{$_};
+        }
+
+        return 0;
+    }
+}
+
+=head2 munge_overlapping_asset_bodies
+
+Some bodies have assets which overlap, or are in, another body's area. For
+those, we add the overlapped area into the council_area_id so the point in the
+other area can be selected. But we then need to filter which categories appear
+- if it's on the overlapping asset, the cobrand body, otherwise whichever body
+actually covers that area.
+
+=cut
+
+sub munge_overlapping_asset_bodies {
+    my ($self, $bodies) = @_;
+
+    return unless $self->can('check_report_is_on_cobrand_asset') && $self->body;
+
+    my %body_names = map { $_->name => 1 } grep { $_->name !~ /TfL|National Highways/} values %$bodies;
+    if (scalar(keys %body_names) == 1 && $body_names{$self->body->name}) {
+        # The cobrand body is the only body, other than possibly NH and/or TfL, so do nothing.
+        return;
+    };
+
+    # If we get here, assume we are outside the cobrand's boundaries so run the
+    # test to see if we are on the cobrand's asset and return the relevant bodies
+    if (!$self->check_report_is_on_cobrand_asset) {
+        %$bodies = map { $_->id => $_ } grep { $_->name ne $self->body->name } values %$bodies;
+    }
 }
 
 sub area_check_error_message {
@@ -381,6 +437,8 @@ sub munge_report_new_bodies {
         my $thamesmead = FixMyStreet::Cobrand::Thamesmead->new({ c => $self->{c} });
         $thamesmead->munge_thamesmead_body($bodies);
     }
+
+    $self->munge_overlapping_asset_bodies($bodies);
 }
 
 sub munge_report_new_contacts {


perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -201,6 +201,8 @@ sub munge_report_new_bodies {
         my $thamesmead = FixMyStreet::Cobrand::Thamesmead->new({ c => $self->{c} });
         $thamesmead->munge_thamesmead_body($bodies);
     }
+
+    $self->{c}->cobrand->call_hook(munge_overlapping_asset_bodies => $bodies) if $self->{c}->cobrand->can('report_is_on_cobrand_asset');
```
chrismytton, line 205: I think the `if $self->{c}->cobrand->can('report_is_on_cobrand_asset')` here (and in `UKCouncils.pm` below) should be moved into the `munge_overlapping_asset_bodies` method, as it's more of an implementation detail of that method, not something the caller should have to worry about.

chrismytton (changes requested):
Looks good, just one minor comment!

chrismytton (approved):

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+coverage is limited only to an overlapping asset by testing the hook
+report_is_on_cobrand_asset
+
+=cut
+
+sub responsible_for_areas {
+    my ($self, $councils) = @_;
+
+    if ($self->{c}->cobrand->can('report_is_on_cobrand_asset')) {
+        if (scalar(%$councils) == 1 && (values %$councils)[0]->{name} eq $self->council_name) {
```
dracos, line 192: This would presumably have to do something different in two-tier areas, not sure worth doing without a concrete need, but perhaps worth a note. Not sure we want to rely on names not changing in MapIt ever here, either, even if unlikely - could it go on the first ID in $self->council_area_id perhaps, then more matches the else case.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+checking if it has the area set in council_area_id and also checking if the
+coverage is limited only to an overlapping asset by testing the hook
+report_is_on_cobrand_asset
+
+=cut
+
+sub responsible_for_areas {
+    my ($self, $councils) = @_;
+
+    if ($self->{c}->cobrand->can('report_is_on_cobrand_asset')) {
```
dracos, line 191: $self->{c}->cobrand is the same as $self.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+        if (scalar(%$councils) == 1 && (values %$councils)[0]->{name} eq $self->council_name) {
+            return 1;
+        } elsif ($self->{c}->cobrand->call_hook('report_is_on_cobrand_asset')) {
```
dracos, line 194: You don't need to use `call_hook` here because you've already tested it exists, you can call it directly.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+
+    if (scalar(values %$bodies) == 3 && (grep { $_->name eq 'National Highways' } values %$bodies
+        && grep { $_->name eq 'TfL' } values %$bodies)) {
+        #The cobrand body and National Highways and TfL are the only bodies, so do nothing.
+        return;
+    }
+
+    # If we get here, assume we are outside the cobrand's boundaries so if there's a test
+    # we run it to see if we are on the cobrand's asset and return the relevant bodies
+    if ($self->call_hook('report_is_on_cobrand_asset' => \%this_body, \%other_bodies)) {
```
dracos, line 248: How can the hook take parameters here, but not when called in responsible_for_areas?
You're calling the function twice here as well, either check+call direct, or use call_hook.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+whichever body actually covers that area.
+
+=cut
+
+sub munge_overlapping_asset_bodies {
+    my ($self, $bodies) = @_;
+
+    return unless $self->can('report_is_on_cobrand_asset');
+
+    my %this_body = map { $_->id => $_ } grep { $_->name eq $self->council_name || $_->name eq $self->body->name } values %$bodies;
```
dracos, line 227: It feels like you want to construct a quicker hash look up of body name to 1 given the number of greps you are going on to make in this function. Much as `munge_report_new_bodies` already did at the start for the same reason.

dracos: Why are we checking two names? If we're checking body name, I don't think we need to check council_name as well.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+    my %other_bodies = map { $_->id => $_ } grep { $_->name ne $self->council_name && $_->name ne $self->body->name } values %$bodies;
+
+    if (scalar(values %$bodies) == 1 && (grep { $self->council_name eq $_->name } values %$bodies || grep { $self->body->name eq $_->name } values %$bodies)) {
```
dracos, line 230: Presumably this is the same as other_bodies being empty?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -171,17 +171,88 @@ sub area_check {
...
+        # The cobrand body is the only body, so do nothing.
+        return;
+    }
+
+    if (scalar(values %$bodies) == 2 && grep { $_->name =~ /TfL|National Highways/} values %$bodies) {
```
dracos, line 235: Could we not ignore TfL / National Highways when constructing other bodies (assuming this won't be running on TfL itself), and then these two checks wouldn't be needed, would be covered by the other_bodies being empty above.

dracos (changes requested):
I'm afraid I don't think this works at present, though the idea is good :) The lack of tests is one factor why this wasn't spotted, I guess; if this can't be done in conjunction with a concrete use (either Bristol or Brent, presumably), then faking a cobrand (Brent?) with the needed functions should be included so we know this is working as expected.

perllib/FixMyStreet/Cobrand/FixMyStreet.pm
```diff
@@ -201,6 +201,8 @@ sub munge_report_new_bodies {
         my $thamesmead = FixMyStreet::Cobrand::Thamesmead->new({ c => $self->{c} });
         $thamesmead->munge_thamesmead_body($bodies);
     }
+
+    $self->{c}->cobrand->call_hook(munge_overlapping_asset_bodies => $bodies);
```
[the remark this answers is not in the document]
dracos: I'm not sure how you can do what you'd want here, without having a manual lookup of cobrands that you'd want to initialize and call /their/ munge_overlapping_asset_bodies on. Could have it in configuration I guess, but for now manual is probably fine?

perllib/FixMyStreet/Cobrand/UKCouncils.pm
dracos: Yes, I didn't mean change bodies at that point, only to ignore them for the purposes of checking.

perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -381,6 +438,9 @@ sub munge_report_new_bodies {
         my $thamesmead = FixMyStreet::Cobrand::Thamesmead->new({ c => $self->{c} });
         $thamesmead->munge_thamesmead_body($bodies);
     }
+
+    $self->{c}->cobrand->call_hook(munge_overlapping_asset_bodies => $bodies);
```
dracos: $self->{c}->cobrand is $self, and you don't need call_hook because you know it exists, it's just above :)

dracos (commented):
One comment (which I did while I was there) and a fixup for you to check is okay, thought was easier with a little rewrite and not bothering with `report_is_on_cobrand_asset` for the moment

dracos:
This was done as part of work for e.g. #4464 and #4538.

PR #4520, opened 2023-07-16
Add support for requiring a phone number for particular categories.

neprune:
Closes https://github.com/mysociety/societyworks/issues/3747

### DIFF
--- a/templates/web/base/report/form/user_loggedout_by_email.html
+++ b/templates/web/base/report/form/user_loggedout_by_email.html
@@ -9,7 +9,7 @@
 
   [% IF type != 'update' %]
     [% IF NOT c.cobrand.call_hook('disable_phone_number_entry') AND NOT c.cobrand.sms_authentication %]
-        <label for="form_phone">[% loc('Phone number (optional)') %]</label>
+        <label for="form_phone">[% loc('Phone number') %]<span id="js-optional-phone"> [% loc('(optional)') %]</span></label>
         [% IF field_errors.phone %]
           <p class='form-error'>[% field_errors.phone %]</p>
         [% END %]
@@ -33,11 +33,9 @@
     [% END %]
     <p class="segmented-control segmented-control--radio">
         <input required type="radio" name="update_method" id="update_method_email"
-             data-show="#js-optional-phone" data-hide="#js-optional-email"
             value="email"[% ' checked' IF update_method == 'email' %]>
         <label class="btn" for="update_method_email">[% loc('Email') %]</label>
         <input type="radio" name="update_method" id="update_method_phone"
-             data-show="#js-optional-email" data-hide="#js-optional-phone"
             value="phone"[% ' checked' IF update_method == 'phone' %]>
         <label class="btn" for="update_method_phone">[% loc('Phone') %]</label>
     </p>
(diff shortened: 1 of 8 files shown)


templates/web/base/report/new/form_user_loggedin.html
```diff
@@ -54,7 +54,7 @@
     [% INCLUDE 'report/_show_name_label.html' %]
 
 [% IF NOT c.user.phone_verified AND NOT c.cobrand.call_hook('disable_phone_number_entry') %]
-    <label for="form_phone">[% loc('Phone number (optional)') %]</label>
+    <label for="form_phone">[% loc('Phone number') %] <span id="phone_label_optional_tag">[% loc('(optional)') %]</span></label>
```
dracos, line 57: This is the form for logged in users; you'll need equivalent in `report/form/user_loggedout_by_email.html` for logged out people. And note there it is more complicated because of possibility of signing up to updates by SMS, so it already has 'js-optional-phone' in use as the ID there, which you might want to use as the ID here as well.

dracos (changes requested):
Looks good, needs a bit more work for the logged out use.

templates/web/base/report/form/user_loggedout_by_email.html
```diff
@@ -47,13 +45,13 @@
         <p class='form-error'>[% field_errors.username_register %]</p>
     [% END %]
 
-    <label for="form_email">[% loc('Your email') %]<span class="hidden-js" id="js-optional-email"> [% loc('(optional)') %]</span></label>
```
dracos, line 50: I think these were hidden to begin with so that people wouldn't potentially get confused seeing both listed as optional before they picked an update method above. You can leave the class on here, and instead of hide/show in the JS, do addClass/removeClass/toggleClass hidden-js, I think?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1283,14 +1293,23 @@ $.extend(fixmystreet.set_up, {
 
   reporting_required_phone_email: function() {
     var fem = $('#form_email');
+    var fem_optional = $('#js-optional-email');
```
dracos, line 1296: Probably good to have this behaviour all here and self contained anyway :)

dracos (changes requested):
I think (though feel free to suggest otherwise) the default being hidden was preferable, but that's all, looks good :)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1283,14 +1293,23 @@ $.extend(fixmystreet.set_up, {
...
+    var fph_optional = $('#js-optional-phone-update-method-triggered');
+
     $('#update_method_email').on('change', function() {
       fem.prop('required', true);
-      fph.prop('required', false);
+      fem_optional.addClass('hidden-js');
+      if (!fixmystreet.selected_category_data.phone_required) {
+          fph.prop('required', false);
+          fph_optional.removeClass('hidden-js');
+      }
```
dracos, line 1306: Do we need to make this required/hidden if phone_required is set at this point - otherwise it won't be required, I think.

dracos (changes requested):
Only the one query is all :)

web/cobrands/fixmystreet/fixmystreet.js
neprune: Argh, I was convinced I hadn't missed an edge case this time 🙃

dracos (approved):
Woo (note as tidying up staging, I force pushed cbd-jadu-reviewed to fix the thing I mentioned due to my review comment, so might need a rebase etc)

PR #4524, opened 2023-07-18
[Oxfordshire] Remove some unneeded unused code.

dracos:
The timezone change was causing the code to cut the window being looked at by an hour during BST.
[skip changelog]


davea (approved):
✂️

PR #4535, opened 2023-07-24
[Central Beds] Address Jadu integration customer feedback.

neprune:
Closes https://github.com/mysociety/societyworks/issues/3792
Closes https://github.com/mysociety/societyworks/issues/3800
Closes https://github.com/mysociety/societyworks/issues/3801
Closes https://github.com/mysociety/societyworks/issues/3802

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
@@ -59,6 +59,30 @@ sub new_report_title_field_hint {
 
 sub report_sent_confirmation_email { 'external_id' }
 
+=item * We don't send report sent emails for Jadu backed contacts.
+
+=cut
+
+sub suppress_report_sent_email {
+    my ($self, $report) = @_;
+    foreach my $contact ($report->contacts) {
+        return 1 if $contact->email =~ /Jadu/;
+    }
+    return 0;
+}
+
+=item * We don't send report author alerts for Jadu backed contacts.
+
+=cut
+
+sub suppress_reporter_alerts {
+    my ($self, $report) = @_;
+    foreach my $contact ($report->contacts) {
+        return 1 if $contact->email =~ /Jadu/;
+    }
+    return 0;
+}
+
 =item * We do not send questionnaires.
 
 =cut
--- a/templates/web/base/report/new/after_photo.html
+++ b/templates/web/base/report/new/after_photo.html
@@ -4,6 +4,6 @@
         <li>[% loc('For best results include a close-up and a wide shot') %]</li>
     </ul>
     <ul class="dont">
-        <li>[% loc('Avoid personal information and vehicle number plates') %]</li>
+        <li id="js-photo-tips-avoid-personal-info">[% loc('Avoid personal information and vehicle number plates') %]</li>
     </ul>
 </div>
--- a/web/cobrands/centralbedfordshire/flytipping.js
+++ b/web/cobrands/centralbedfordshire/flytipping.js
@@ -1,4 +1,6 @@
-$(fixmystreet).on('report_new:category_change', function() {
+(function(){
+
+function setupWitnessedRelatedFields() {
     var witnessed = $('#form_fly_tip_witnessed');
     if (!witnessed) {
         return;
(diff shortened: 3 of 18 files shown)


web/cobrands/fixmystreet-uk-councils/council_category_hide_element_rules.js
```diff
@@ -0,0 +1,8 @@
+hide_element_rules = {
+    'Central Bedfordshire Council': {
+        'Fly Tipping': [
+            "description-tips",
```
davea, line 4: If this was a selector that could be passed straight to `$` on L692 of `web/cobrands/fixmystreet/fixmystreet.js` it'd be more flexible than only allowing element IDs.

templates/web/base/report/new/after_photo.html
```diff
@@ -4,6 +4,6 @@
         <li>[% loc('For best results include a close-up and a wide shot') %]</li>
     </ul>
     <ul class="dont">
-        <li>[% loc('Avoid personal information and vehicle number plates') %]</li>
+        <li id="photo-tips-avoid-personal-info">[% loc('Avoid personal information and vehicle number plates') %]</li>
```
davea, line 7: There's a convention of using a `js-` prefix on IDs/classes that are only used for JS functionality, think that would be good to use here (and the other ones too).

web/cobrands/centralbedfordshire/flytipping.js
```diff
@@ -52,4 +55,20 @@ $(fixmystreet).on('report_new:category_change', function() {
 
     checkAndToggle();
     witnessed.on('change', function() { checkAndToggle(); });
+}
+
+$(fixmystreet).on('report_new:category_change', function() {
+    if (!fixmystreet.bodies.includes('Central Bedfordshire Council')) {
```
davea, line 61: `.includes` [isn't supported](https://caniuse.com/array-includes) in IE11, sadly, so think `.indexOf` is the way to go here.

```suggestion
    if (fixmystreet.bodies.indexOf('Central Bedfordshire Council') === -1) {
```

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -676,6 +678,30 @@ $.extend(fixmystreet.set_up, {
...
+        });
+        fixmystreet.hidden_elements = [];
+
+        // apply hide element rules
+        $.each(fixmystreet.bodies, function(index, body) {
+            if ( typeof hide_element_rules !== 'undefined' && hide_element_rules[body] && hide_element_rules[body][category] ) {
+                var elementIDs = hide_element_rules[body][category];
+                $.each(elementIDs, function(index, elementID) {
+                    var element = $('#' + elementID);
+                    if (!element) {
```
davea, line 693: `element` is a jQuery object which is truthy, `.length` is the way to check if any elements matched:

```suggestion
                    if (!element.length) {
```

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -59,6 +59,30 @@ sub new_report_title_field_hint {
 
 sub report_sent_confirmation_email { 'external_id' }
 
+=item * We don't send report sent emails for Jadu backed contacts.
+
+=cut
+
+sub suppress_report_sent_email {
```
davea, line 66: Is this new method needed? Can't you do the same check in the existing `report_sent_confirmation_email` method and return the appropriate value there?

davea (changes requested):
A few little tweaks on the JS side of things, but also would be good to have some tests for the `suppress_reporter_alerts` functionality and `suppress_report_sent_email` (if the latter is staying).

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
neprune: If `report_sent_confirmation_email` returns a falsey value, a report logged email is sent straight after a report is confirmed (`New.pm`). If truthy, a report logged email is sent after the report is sent (`Queue/Item/Report.pm`) instead, using the given string to extract an ID to include in the email.

Beds want no report logged email sent at all, hence the new method.

davea: Ah, shame we can't have a trinary data type :) Makes sense! 👍

davea (approved):
Woo! Looks good 🎉

PR #4536, opened 2023-07-24
[Buckinghamshire] Evo: Handle locations with no nearby roads

chrismytton:
When looking up the nearest road we're only checking a 200m radius, which means that sometimes there are claim reports that don't get a road because they're further than 200m from a road (e.g. claims for accidents on a footpath). Handle this by returning "Unknown location" if no roads are returned.

Fixes https://github.com/mysociety/societyworks/issues/3803


MorayMySoc (approved):
Looks good

PR #4537, opened 2023-07-25
[Brent] Use OSPlaces API as reverse geocoder

nephila-nacrea:
As discussed in https://mysocietysupport.freshdesk.com/a/tickets/3047.

Brent are switching to the OSPlaces API as a reverse geocoder, and along with this change, want a street name column in their reports CSVs.

Also handles OSPlaces data for open311_extra_data_include() method.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -206,6 +206,28 @@ sub user_from_oidc {
 
 =cut
 
+=head2 dashboard_export_problems_add_columns
+
+Brent have an additional column for street name.
+
+=cut
+
+sub dashboard_export_problems_add_columns {
+    my ($self, $csv) = @_;
+
+    $csv->add_csv_columns(
+        street_name => 'Street Name',
+    );
+
+    $csv->csv_extra_data(sub {
+        my $report = shift;
+        return {
+            street_name =>
+                $report->nearest_address_parts->{street}
+        }
+    });
+}
+
 =head2 open311_config
 
 Sends all photo urls in the Open311 data
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -304,14 +304,22 @@ Same as Symology above, but different attribute name.
         }
     }
 
-    # The title field gets pushed to location fields in Echo/Symology, so include closest address
+=item * The title field gets pushed to location fields in Echo/Symology, so include closest address
+
+We use {closest_address}->summary as this is geocoder-agnostic.
+
+=cut
+
     my $title = $row->title;
-    if ($h->{closest_address}) {
-        if (my $addr = $h->{closest_address}{name} || $h->{closest_address}{display_name} || '') {
-            $addr =~ s/, England, United Kingdom$//;
-            $title .= "; Nearest calculated address = $addr";
-        }
+    if ( $h->{closest_address} ) {
+        my $addr = $h->{closest_address}->summary;
+
+        $addr =~ s/, England//;
+        $addr =~ s/, United Kingdom$//;
+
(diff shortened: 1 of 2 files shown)


t/cobrand/brent.t
```diff
@@ -241,33 +241,83 @@ for my $test (
...
+                . '; Nearest calculated address = Engineers Way, London Borough of Brent, London, Greater London, HA9 0FJ';
+            is $c->param('attribute[title]'), $title,
+                'Report title and location passed as attribute for Open311';
+        };
+
+        $problem->delete;
     };
 
-    $problem->delete;
+    subtest 'OSM geocoder' => sub {
```
MorayMySoc, line 281: Should this name highlight testing OSPlaces change?

MorayMySoc (approved):
Looks good to me - may want to clarify the test title highlighted though>

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -206,6 +206,32 @@ sub user_from_oidc {
...
+    return unless $self->get_reverse_geocoder eq 'OSPlaces';
+
+    $csv->add_csv_columns(
+        street_name => 'Street Name',
+    );
+
+    $csv->csv_extra_data(sub {
+        my $report = shift;
+
+        my $closest_address = $self->find_closest($report);
```
dracos, line 227: We shouldn't call this in a CSV export because it may involve external lookups. See how Hackney calls the report's `nearest_address` function instead to get the data if it's already present. `parts->street` would be what you want here.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -593,13 +593,23 @@ sub find_closest {
...
-        if ($problem) {
-            # cache the geocoder results for use in alerts
-            $problem->geocode( $j );
-            $problem->update;
-        }
+    if ($j) {
+        my $addr = FixMyStreet::Geocode::Address->new($j);
+
+        # Cobrand's geocoder may have changed; only return if it hasn't
+        return $addr if ref($addr) =~ $self->get_reverse_geocoder;
```
dracos, line 600: I don't think we want to make this change, because this unnecessarily stops all previous geocodes working if a cobrand changes its provider. The purpose of the Geocode::Address wrapper objects is that it shouldn't matter which provider is used in general.
If we care about old reports, we could in theory do a manual update of their entries, but I don't think that's necessary here.

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -284,10 +310,22 @@ Same as Symology above, but different attribute name.
 
     # The title field gets pushed to location fields in Echo/Symology, so include closest address
     my $title = $row->title;
-    if ($h->{closest_address}) {
-        if (my $addr = $h->{closest_address}{name} || $h->{closest_address}{display_name} || '') {
-            $addr =~ s/, England, United Kingdom$//;
-            $title .= "; Nearest calculated address = $addr";
+    if ( $h->{closest_address} ) {
```
dracos, line 313: As mentioned in Slack at https://mysociety.slack.com/archives/C01TK8P1K8T/p1689929218707969?thread_ts=1689928394.321709&cid=C01TK8P1K8T, there is a `->summary` call on the Address objects that returns the information needed here, so that would be what to use, and this code won't need to know which geocoder is being used.

dracos (changes requested):
Functionality looks good, but I think we need to make a few changes to prevent external lookups during a CSV export and an easier way to get the information after.

perllib/FixMyStreet/Cobrand/Brent.pm
dracos: Could also add a bit of POD here about this in addition

dracos (approved):
Looks good, thanks for making the changes :)

PR #4538, opened 2023-07-25
[Brent] 3716 brent boundary roads shared with camden

MorayMySoc:
Calls an asset layer that covers Brent and overlaps the boundaries and calculates responsibilities for the picked area.
    
If the area is covered by the Brent asset, but is in another council's area, Brent's Symology categories should be combined with the categories usually present for that point so reports can be made to Brent although not officially their area

[skip changelog]


t/cobrand/brent.t
```diff
@@ -367,6 +376,82 @@ FixMyStreet::override_config {
...
+                    $brent_mock->mock('_fetch_features',
+                        sub { [ {'ms:brent_boundaries' => { 'gml:id' => 'brent_boundaries.1' } }] });
+                    $mech->get_ok("/report/new/ajax?longitude=-0.124514&latitude=51.529432");
+                    is $mech->content_lacks("Potholes"), 1, 'Brent category not present';
+                    is $mech->content_contains("Gully grid missing"), 1, 'Brent Symology category present';
+                    is $mech->content_contains("Sweeping"), 1, 'TfL category present';
+                    is $mech->content_contains("Fly-tipping"), 1, 'Camden category present';
+                }
+            };
+            subtest "can access Camden from Brent on asset layer" => sub {
```
dracos, line 434: This test is identical to half of the test above it, I think.

dracos (changes requested):
Core changes checked in #4519, this looks good but one duplicate test, plus it would be good to have some tests for what happens on the Camden cobrand as well, just to check :)

dracos (commented):
Looks good - couple of queries to go on with, will come back about the remaining in-brent-not-brent situation.

MorayMySoc:
(Sorry - extra fixup for broken tests)

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -135,6 +136,101 @@ sub geocoder_munge_results {
...
+            );
+
+        my %exclude = (
+            'Barnet' => { 'Pavements' => 1, 'Potholes' => 1, 'Street cleaning' => 1 },
+            'Camden' => { 'Pavements' => 1, 'Potholes' => 1 },
+            'Harrow' => { 'Pavements' => 1, 'Potholes' => 1 },
+            );
+
+        @$contacts = grep { !($_->email =~ /^Symology/ && $_->body_id == $self->body->id) } @$contacts;
+        @$contacts = grep { !($exclude{$cobrand}{$_->category} && $_->body_id == $ids{$cobrand})} @$contacts;
```
dracos, line 230: Do you want the reverse here? It's remove Brent "roads" and other "roads" rather than other "non-roads" I think?

t/cobrand/brent.t
```diff
@@ -367,6 +376,125 @@ FixMyStreet::override_config {
...
+    ALLOWED_COBRANDS => [ 'brent', 'fixmystreet', 'tfl', 'camden' ],
+    MAPIT_URL => 'http://mapit.uk/',
+}, sub {
+    my $camden = $mech->create_body_ok(2505, 'Camden Borough Council');
+    my $barnet = $mech->create_body_ok(2489, 'Barnet Borough Council');
+    FixMyStreet::DB->resultset('BodyArea')->find_or_create({
+        area_id => 2488,
+        body_id => $barnet->id,
+    });
+    my $harrow = $mech->create_body_ok(2487, 'Harrow Borough Council');
```
dracos, line 389: Should harrow (and camden) also cover brent area, like barnet?

dracos (changes requested):
Couple of queries, steps along the way there :)

perllib/FixMyStreet/Cobrand/Brent.pm
MorayMySoc: So Brent roads are Echo? If we exclude Brent roads and other roads, do we not have no roads?

dracos: Sorry if I wasn't clear. You're removing Brent Symology (roads stuff), okay, but you're also removing Barnet/Camden/Harrow roads, rather than removing their non-roads.

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -196,7 +196,7 @@ sub munge_overlapping_asset_bodies {
     my ($self, $bodies) = @_;
 
     # in_area will be true if the point is within the administrative area of Brent
-    my $in_area = $self->{c}->stash->{all_areas} && scalar(%{$self->{c}->stash->{all_areas}}) == 1 && (values %{$self->{c}->stash->{all_areas}})[0]->{id} eq $self->council_area_id->[0];
+    my $in_area = grep ($self->council_area_id->[0] == $_, keys %{$self->{c}->stash->{all_areas}});
```
dracos, line 199: Presumably this fix also needs to apply to the other place Dave changed in the parent commit, or something else?

PR #4539, opened 2023-07-27
[Waste] Async Echo API calls for bin days page

davea:
Extends the use of the HTMX partial page refresh to Echo WasteWorks API calls. Turned out to be fairly straightforward in the end, the only real change was to ensure the results of the property lookup aren’t cleared if the request is made by HTMX/page refresh while waiting for second batch of API calls to return.

(That makes me wonder if I’ve missed something somewhere though… But everything seems to work as expected.)

[skip changelog]


dracos (approved):
Seems to work fine :) I've updated the code to put the loading detach in the call_api, and updated the template to put the special loading stuff in its own block at the top rather than in a big ELSE round everything else.

PR #4541, opened 2023-07-31
[Brent] Amend template so '9' represents sacks rather than 9 bins

MorayMySoc:
The BRT - Paid Collection Container Quantity has code 1 matching 1 bin and continues with 2:2, 3:3 etc so the number code corresponds to the number of bins.

We have relied on the code for the number of bins.

However, when we reach 9, that corresponds to sacks rather than a number so if we have 9 bins - it is, in fact, a sack subscription and the template adapts accordingly.

https://mysocietysupport.freshdesk.com/a/tickets/3211

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -531,6 +531,7 @@ sub bin_services_for_address {
 
         my $garden = 0;
         my $garden_bins;
+        my $garden_sacks;
         my $garden_cost = 0;
         my $garden_due = $self->waste_sub_due($schedules->{end_date});
         my $garden_overdue = $expired{$_->{Id}};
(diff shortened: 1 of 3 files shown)


templates/web/brent/waste/services.html
```diff
@@ -46,9 +46,17 @@
     <dl class="govuk-summary-list">
       <div class="govuk-summary-list__row">
         <dt class="govuk-summary-list__key">Subscription</dt>
+        [%# unit.gardent_bins is a key value pair of code-number to number of bins and corresponds 1:1, 2:2 etc,
+        until it gets to 9 when it corresponds to sacks %]
+        [% IF unit.garden_bins == 9 %]
```
dracos, line 51: Whilst this works fine, I think it'd be better to do this check in the Brent.pm cobrand code where garden_bins is read/set - if the Value is 9, set garden_sacks to 1 instead of garden_bins, perhaps, set the $garden_cost to garden_waste_sacks_cost_pa, and then here in the template you can have if garden_sacks is set and the same output.

dracos (changes requested):
Just suggesting moving the check to in the code so template is easier and also anything else that might use the data sees the right data.

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -540,7 +541,14 @@ sub bin_services_for_address {
             foreach (@$data) {
                 if ( $_->{DatatypeName} eq 'BRT - Paid Collection Container Quantity' ) {
                     $garden_bins = $_->{Value};
-                    $garden_cost = $self->garden_waste_cost_pa($garden_bins) / 100;
+                    # $_->{Value} is a code for the number of bins and corresponds 1:1 (bin), 2:2 (bins) etc,
+                    # until it gets to 9 when it corresponds to sacks %]
```
dracos, line 545: Little template end at end of line here

dracos (approved):
Looks good, one comment tweak :)

PR #4542, opened 2023-08-01
Fixed padding for 'item-list--reports__item a'

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3810

The rule padding: flip(0 0 0 1em, 0 1em 0 0) has been replaced so there is always some horizontal padding.
This commit also fixes the left padding for cobrands using a pin icon in the "Recently reported problems" section in the front page.

### Cobrand with Pin
Screenshot 2023-08-01 at 08 31 17

Screenshot 2023-08-01 at 08 31 51

### Cobrand with no Pin
Screenshot 2023-08-01 at 08 32 26
Screenshot 2023-08-01 at 08 32 52

[Skip changelog]

### DIFF
--- a/web/cobrands/bromley/layout.scss
+++ b/web/cobrands/bromley/layout.scss
@@ -195,14 +195,6 @@ body.alertindex form.full-width {
     font-size: 1em;
 }
 
-// We have slightly different content in our lists of issues to what
-// reports_list.scss expects, so we need to tweak the padding back to normal.
-// This stops the spacing being too large.
-.item-list__item--with-pin a {
-  padding: 0;
-  padding-left: 3em;
-}
-
 // Bromley's footer
 .site-footer {
   padding: 120px 0 0;
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -1248,7 +1248,7 @@ input.final-submit {
 
   a {
     @include clearfix;
-    padding: flip(0 0 0 1em, 0 1em 0 0);
+    padding: 0 1em;
 
     @media print {
       padding: 0;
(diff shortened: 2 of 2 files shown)


MorayMySoc (approved):
Looks good to me

dracos (changes requested):
The removed flip padding line is because on a map list page, the photos on the right currently touch the edge of the content column, whilst the text on the left does not (and vice versa in right-to-left language):
Screenshot 2023-08-01 at 12 56 02
So removing this, and the front page override, stops that - if we want to change that everywhere, we of course can, but it's not only a fix for a front page overlap bug. (On a map list page, as opposed to the front page, the reduced spacing is to have more reports available to see, presumably? Not sure we'd want to change that.)

The issue here is only affecting those cobrands using the sass/report_list_pins import (which adds the pins to the lists) - which includes the item-list__item--indented which is where the fix would go. So you'd have a `body.frontpage .item-list__item--indented a { padding-#{$left}: 2.5em; }` (I think?) in _base.scss so that it would maintain its correct left indent when used here.

As a hybrid, if we don't want the photos to touch the edge anymore, you could replace the flip line with `padding: 0 1em;`, remove the body.frontpage line, and that would also fix the overlap (without changing the top/bottom padding).

lucascumsille:
>As a hybrid, if we don't want the photos to touch the edge anymore, you could replace the flip line with padding: 0 1em;, remove the body.frontpage line, and that would also fix the overlap (without changing the top/bottom padding).

Thanks @dracos =) at the end I went for the hybrid option.
I also got rid of some code that was causing Bromley's images to be touching the edge.

Let me know if I need to add anything else to this PR.

PR #4544, opened 2023-08-01
[Bucks] Remove claims from going into open311

MorayMySoc:
Request to only send claims to Evo.

Removes sending to Open311 and expectation in tests of open311 response

https://mysocietysupport.freshdesk.com/a/tickets/3221

[skip changelog]


chrismytton (changes requested):
The code that has been removed here is for posting an update on a claim after it has been created. I believe Bucks still want this update to be left on claims, so we'll need to keep this code.

If we change the **Claims** category's send method to be no-op then none of the open311-related code in the Bucks cobrand will be run for claim reports.

So I think the work required is to move the claim-related code that's still needed _out_ of the open311 methods, so it still gets run when a claim is created.

dracos:
Think it's easier just to leave it as Open311 (even though it isn't) but SKIP all Claim reports and let it handle them like that.

chrismytton (approved):
Feels slightly tape-and-string, but does seems to do the trick!

PR #4545, opened 2023-08-01
Return random unprocessed row to daemon.

dracos:
If there are a number of reports failing to send, leaving it to database order could cause the same reports to be returned over and over, never looking at other waiting reports.

And fix fallback code thing.


davea (approved):
⚡ 🆗

PR #4546, opened 2023-08-01
[Central Beds] Indicate whether reported by staff for Jadu reports.

neprune:
Related: https://github.com/mysociety/open311-adapter/pull/291

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
@@ -170,6 +170,8 @@ sub open311_extra_data_include {
     ];
 
     if ($contact->email =~ /Jadu/) {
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -170,6 +170,8 @@ sub open311_extra_data_include {
     ];
 
     if ($contact->email =~ /Jadu/) {
+        my $reported_by_staff = $row->get_extra_metadata('contributed_as', 'n/a') eq 'another_user' ? 'Yes' : 'No';
```
dracos, line 173: Is fine, but I wonder if they'd want to have staff-as-anonymous-user or -body come through as staff as well? Might be easier to see if `contributed_by` is set and go on that?

dracos (commented):
One suggestion for covering wider staff usage is all :)

dracos (approved):

PR #4551, opened 2023-08-02
[Brent] Update assets for ATAK integration

chrismytton:
- Update housing estates responsibility message
- Added new responsibility messages for allotments and verges/flower beds
- Flip the found/not_found methods as we now want to show stopper when a housing estate (etc) _isn't_ found.
- Fix a bug where attributes weren't being set when selecting the subcategory

Fixes https://github.com/mysociety/societyworks/issues/3807

### open311-adapter pull request and servers repo change

- https://github.com/mysociety/open311-adapter/pull/287
- Commit `69f8f51cc72a96d6dc924748f34fb8c875b47721` in the servers repo for the related asset layer changes.


templates/web/brent/report/new/roads_message.html
```diff
@@ -14,3 +14,11 @@
         <p><strong>Please select a park or highway from the map</strong></p>
         <p>To report this issue please select a verge or shrub bed that is in a Brent maintained park or within the boundaries of roads and pavements the council is responsible for</p>
     </div>
+    <div id="js-brent-allotments" class="hidden js-responsibility-message">
```
dracos, line 17: Don't think I've emailed about this yet, but recently I added the ability to have this message functionality work from config - `no_asset_message`, then we don't need to update this template. (Rest of code still needed/same.) - Brent is only user so far, as an example.

web/cobrands/fixmystreet/assets.js
```diff
@@ -319,6 +319,15 @@ OpenLayers.Layer.VectorNearest = OpenLayers.Class(OpenLayers.Layer.VectorBase, {
             return;
         }
         this.checkFeature(null, fixmystreet.get_lonlat_from_dom());
+
+        // It's possible an asset has been selected before a category (e.g. if
```
dracos, line 323: This block is not necessary, I think. `road: true` asset layers use the `usrn:` dictionary to update key-values based on the nearest feature automatically. Instead of `attributes: location_name: site_name` for the road layers, you want `usrn: - { attribute: 'site_name', field: 'location_name' }` which should do what you're after.

dracos (changes requested):
On this PR's contents - one thing I don't think is needed with a change to how you're doing the config, and another thing you may be able to use a new config variable to save adding new things to the templates. And the tests need updating/adding for the changes.

Regarding the wider configuration -  as with cemeteries, do we need to look at how to allow TfL reports within the areas, assuming e.g. the abandoned santander cycle etc should always be allowed wherever? And currently Grass verges (littering or maintenance issue) can be reported on Highways or Parks - there's not any change being made because of this that I can see, presumably that should be removed as part of this?

We need a server fallback for if a report gets submitted without a location name (assuming we want to always try and get one for Atak categories). And like eg Bucks car park, do we want a server-side error if a relevant category is picked and it's not within the relevant thing?

chrismytton:
> On this PR's contents - one thing I don't think is needed with a change to how you're doing the config, and another thing you may be able to use a new config variable to save adding new things to the templates. And the tests need updating/adding for the changes.

- https://github.com/mysociety/fixmystreet/pull/4551/commits/ea3258be869cdf2438e4fe12f1ec0161d16b6331 and https://github.com/mysociety/fixmystreet/pull/4551/commits/63344f51878001bc2649f247c323d03e0e54597a remove code that can be asset config instead
- Asset configuration updated in commit `1e02234a4` of the servers repo.

> Regarding the wider configuration - as with cemeteries, do we need to look at how to allow TfL reports within the areas, assuming e.g. the abandoned santander cycle etc should always be allowed wherever?

I don't think so, these categories aren't `always_visible: true` so they won't have any effect unless an ATAK group has been selected.

> And currently Grass verges (littering or maintenance issue) can be reported on Highways or Parks - there's not any change being made because of this that I can see, presumably that should be removed as part of this?

Have removed most of the code for that (apart from the red routes bits) in https://github.com/mysociety/fixmystreet/pull/4551/commits/ec0d36a205d7c2e74f465d2941c6192e548923b6.

> We need a server fallback for if a report gets submitted without a location name (assuming we want to always try and get one for Atak categories). And like eg Bucks car park, do we want a server-side error if a relevant category is picked and it's not within the relevant thing?

Added in https://github.com/mysociety/fixmystreet/pull/4551/commits/e95159e0112157d3bbef2aebd3384ad0228e2639 and https://github.com/mysociety/fixmystreet/pull/4551/commits/470a4727f8d96a0d568e54cfac974f79aeb79a1e

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -302,6 +304,15 @@ Same as Symology above, but different attribute name.
                 $row->update_extra_field({ name => 'usrn', description => 'USRN', value => $ref });
             }
         }
+
+=item * Adds location name from WFS service for ATAK reports if missing
+
+=cut
+
+    } elsif ($contact->email =~ /^ATAK/) {
+        if (my $name = $self->lookup_location_name($row)) {
```
dracos, line 313: Comment says only do if missing, but it's always doing the lookup and replacing what's already there

dracos (changes requested):
Looks good, one issue is all :)
Also can any of the cypress fixtures also be removed now.

perllib/FixMyStreet/Cobrand/Brent.pm
chrismytton: Ooops, thanks! Have added a check so to avoid the extra HTTP request if it's not necessary (along with a test) in https://github.com/mysociety/fixmystreet/pull/4551/commits/14d85f56fc20a3d064e51f8bc61ac2ccfcd24e40.

Also removed the fixtures in https://github.com/mysociety/fixmystreet/pull/4551/commits/a2532b3a4c71f7f4ed4b0e047c35f3eaa7a20b1d and fixed the tests in https://github.com/mysociety/fixmystreet/pull/4551/commits/20098d9686c14eb6b16de1b4f501ab891b922772.

dracos (approved):
Woo

PR #4553, opened 2023-08-03
Increased contrast for problem state label

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3831
Fixes part of:  https://github.com/mysociety/societyworks/issues/3812

For the background colour I reused the `#666` hex colour that we use across FMS. Now the contrast colour is 5.74.

Screenshot 2023-08-03 at 08 18 15


dracos:
Merged in 94c605c33

Issue #4543, opened 2023-08-01
Outlook 'safelinks' breaks confirmation link UX.


neprune:
An Outlook 'safelink' visits the confirmation link before the user is directed to it, effectively clicking it twice, doing the actual confirmation the first time.
When the report is already confirmed, visiting a confirmation link redirects to the report without logging the user in, to prevent the scenario of someone sharing their confirmation link rather than their report link by accident, and others gaining access to their account.
This means for public reports, a user using Outlook 'safelinks' is directed to the report page, rather than the post-confirmation page, and doesn't get logged in. For private reports the user sees an access denied page instead.

davea:
Do we know if the automated visit from Outlook identifies itself in any way? User-Agent header or something? If we determine the request isn't from a real user we could skip the logic that invalidates the token.

neprune:
It doesn't appear they send any User-Agent header.

PR #4556, opened 2023-08-03
Make confirm links login within 30s of first use.

neprune:
Fixes https://github.com/mysociety/fixmystreet/issues/4543

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1502,6 +1502,9 @@ sub process_confirmation : Private {
     my ( $self, $c ) = @_;
 
     my $data = $c->stash->{token_data};
+    my $token_redeem_cooldown_seconds = $c->stash->{token_redeem_cooldown_seconds};
+    my $token = $c->stash->{token_object};
+
     $c->stash->{template} = $data->{template} || 'tokens/confirm_problem.html';
 
     unless ($c->stash->{report}) {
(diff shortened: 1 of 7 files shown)


neprune:
Will do the same for update confirmations if this looks ok.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1543,55 +1544,63 @@ sub process_confirmation : Private {
...
+            $problem->user->add_oidc_id($data->{oidc_id}) if $data->{oidc_id};
+            $problem->user->extra({
+                %{ $problem->user->get_extra() },
+                %{ $data->{extra} }
+            }) if $data->{extra};
+            $problem->user->update;
         }
     }
+
+    # Make sure extra oauth state is restored, if applicable
```
dracos, line 1596: I think this can be done only the once? It updates the session server-side, unless I've missed something. (Also assume there might have been a reason to only do this if name/password set :) )

dracos (changes requested):
Nice storing of the first redeemed time on the token. Couple of questions is all

dracos (commented):
Ah, I did think about saying "what about reports by text" but looked at Auth/Phone.pm and saw it was setting stash->{token} already so assumed that it must have been covered there before somehow. But I think that's only the string of the token, not the token object, hence the test failures.

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: This is sill present, this block has moved outside the `if ( $data->{name} || $data->{password} ) {` block which seems unrelated to the changes.

dracos (commented):
Looks good, only issue is the movement of a little block I don't think is needed? Other than that :)

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1543,55 +1546,68 @@ sub process_confirmation : Private {
...
+                %{ $data->{extra} }
+            }) if $data->{extra};
+            $problem->user->update;
+
+            # Make sure extra oauth state is restored, if applicable
+            foreach (qw/logout_redirect_uri change_password_uri/) {
+                if ($data->{$_}) {
+                    $c->session->{oauth} ||= ();
+                    $c->session->{oauth}{$_} = $data->{$_};
+                }
```
neprune, line 1606: Sorry, managed to let the fixup get squashed on a rebase, but only change is this block moving back in to the above if block.

dracos (approved):
Thanks :)

PR #4560, opened 2023-08-08
Fms input accessibility

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3821

- Should include these changes in the changelog?
- Should we take the opportunity to create variables for the GOVUK styling: `border-color`, `color`, `focus-yellow` for the GOVUK input elements. Thinking of cases like Camden.
- The layout has changed quite a bit, the filters use more vertical space, but considering that the filter section doesn't have a sticky property and you can still see the reports when the page loads, I didn't think it would be a deal breaker. But let me know what you think.

### What do you think is the best approach with this file:
`web/vendor/jquery.multi-select.min.js` Right now is not a minified version anymore. Also the file has changed quite a lot, so it might be worth it just to move the content to `FixMyStreet.js` or just rename the file to something like `multi-select.js`. The other alternative, but I wasn't to keen for that one, was to apply the gOVUK styling to the multi-select classes, which potentially add a lot of extra code.

### Desktop

https://github.com/mysociety/fixmystreet/assets/13790153/91e6ae43-3c5c-4abd-bcd2-a58d3fef2e89

Screenshot 2023-08-14 at 08 22 38

### Mobile

https://github.com/mysociety/fixmystreet/assets/13790153/41e92433-f5ad-42c9-92d9-655d11b23813

Screenshot 2023-08-14 at 08 23 07

### DIFF
--- a/templates/web/base/reports/_list-filters.html
+++ b/templates/web/base/reports/_list-filters.html
@@ -33,10 +33,14 @@
     <form method="get" action="">
 [% END %]
 
-        <p class="report-list-filters">
-          [% tprintf(loc('<label for="statuses">Show</label> %s reports <label for="filter_categories">about</label> %s', "The first %s is a dropdown of all/fixed/etc, the second is a dropdown of categories"), mark_safe(select_status), mark_safe(select_category)) %]
+        <div class="report-list-filters">
+          [% tprintf(loc('<label for="statuses">Show</label> %s reports', "Dropdown of all/fixed/etc"), mark_safe(select_status)) %]
+        </div>
+
+        <div class="report-list-filters">
+          [% tprintf(loc('<label for="filter_categories">about</label> %s', "Dropdown of categories"), mark_safe(select_category)) %]
           <input type="submit" name="filter_update" value="[% loc('Go') %]">
-        </p>
+        </div>
 
         [% PROCESS 'reports/_list-filters-sort.html' %]
 
--- a/templates/web/base/header/css.html
+++ b/templates/web/base/header/css.html
@@ -7,5 +7,6 @@
     <link rel="stylesheet" href="[% layout_css %]">
 <![endif]-->
 <link rel="stylesheet" href="[% version('/vendor/OpenLayers/theme/default/style.css') %]">
+<link rel="stylesheet" href="[% version('/vendor/govuk-frontend/forms.css') %]">
 
 [% extra_css | safe %]
--- a/templates/web/fixmystreet.com/header/css.html
+++ b/templates/web/fixmystreet.com/header/css.html
@@ -1,6 +1,8 @@
 [% SET base_css = version('/cobrands/' _ c.cobrand.asset_moniker _ '/base.css') %]
 [% SET layout_css = version('/cobrands/' _ c.cobrand.asset_moniker _ '/layout.css') %]
 [% SET ol_css = version('/vendor/OpenLayers/theme/default/style.css') %]
+[% SET govuk_css = version('/vendor/govuk-frontend/forms.css') %]
+
 
 [% TRY %][% critical = INSERT "header/critical.auto.min.css" %][% CATCH file %][% END %]
 
@@ -17,6 +19,7 @@
 <noscript><link rel="stylesheet" href="[% base_css %]"></noscript>
 <link rel="prefetch" href="[% ol_css %]" as="style">
 <link id="preload_base_css" rel="preload" href="[% base_css %]" as="style">
+<link rel="stylesheet" href="[% govuk_css %]" as="style">
 <script nonce="[% csp_nonce %]">
 /* If browser *does* support preload, use stylesheets when loaded */
 document.getElementById('preload_base_css').onload = function(){this.onload=null;this.rel='stylesheet'};
@@ -28,6 +31,7 @@ document.getElementById('preload_base_css').onload = function(){this.onload=null
 [% ELSE %]
 <link rel="stylesheet" href="[% ol_css %]">
 <link rel="stylesheet" href="[% base_css %]">
+<link rel="stylesheet" href="[% govuk_css %]">
 [% END %]
 <link rel="stylesheet" href="[% layout_css %]" media="screen and (min-width:48em)">
 <!--[if (lt IE 9) & (!IEMobile)]>
--- a/templates/web/fixmystreet.com/header/css.html
+++ b/templates/web/fixmystreet.com/header/css.html
@@ -1,8 +1,6 @@
 [% SET base_css = version('/cobrands/' _ c.cobrand.asset_moniker _ '/base.css') %]
 [% SET layout_css = version('/cobrands/' _ c.cobrand.asset_moniker _ '/layout.css') %]
 [% SET ol_css = version('/vendor/OpenLayers/theme/default/style.css') %]
-[% SET govuk_css = version('/vendor/govuk-frontend/forms.css') %]
-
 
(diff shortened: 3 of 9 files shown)


templates/web/base/header.html
```diff
@@ -17,6 +17,7 @@
         [% END %]
         <meta name="HandHeldFriendly" content="true">
         <meta name="mobileoptimized" content="0">
+        <link rel="stylesheet" href="/vendor/govuk-frontend/forms.css">
```
dracos, line 20: Shouldn't need this here if it's in the other place?  I think you need to add it to the fixmystreet.com/header/css.html override template as well rather than here.

templates/web/base/reports/_list-filters.html
```diff
@@ -34,18 +35,20 @@
 [% END %]
 
         <p class="report-list-filters">
-          [% tprintf(loc('<label for="statuses">Show</label> %s reports <label for="filter_categories">about</label> %s', "The first %s is a dropdown of all/fixed/etc, the second is a dropdown of categories"), mark_safe(select_status), mark_safe(select_category)) %]
+          [% tprintf(loc('<label for="statuses">Show reports</label> %s <label for="filter_categories">About</label> %s', "The first %s is a dropdown of all/fixed/etc, the second is a dropdown of categories"), mark_safe(select_status), mark_safe(select_category)) %]
```
dracos, line 38: We'll have to split this up more than this if the labels are now not within a sentence, no need for it to be solely one translation string any longer. But that will break the translations, but I can't see a way to do it otherwise.

dracos (commented):
I think it's fine on mobile to use more vertical space as the filters show/hide. But I think on desktop we will get people complaining if half the vertical space is the filtering. The "Show reports" is generally only open/closed/fixed unless you're staff, so I don't think that needs a whole row; how does it look if you make it a 2x2 even grid with Show reports top left, About top right, Sort by bottom left, and Show older reports bottom right? (We could have done that last with the current layout in fact, big gap there!)

The uncompressed JS is at https://github.com/mysociety/jquery-multi-select/blob/master/src/jquery.multi-select.js you didn't have to try and work with the compressed version! :) I don't know what you've changed, you say it's "changed quite a lot" so it's hard to know what best to do here without knowing what has changed or how we could best move it forward. I don't want to fork our own code unnecessarily, but we'll need actual diff of what's been changed to know.

templates/web/base/reports/_list-filters.html
lucascumsille: Sorry, not sure what you want me to do here.

lucascumsille:
Thanks for all the feedback Matthew =)

> The uncompressed JS is at https://github.com/mysociety/jquery-multi-select/blob/master/src/jquery.multi-select.js you didn't have to try and work with the compressed version! :) I don't know what you've changed, you say it's "changed quite a lot" so it's hard to know what best to do here without knowing what has changed or how we could best move it forward. I don't want to fork our own code unnecessarily, but we'll need actual diff of what's been changed to know.

Thanks when I sent the PR I didn't know that plugin was made by mySociety. I just used Beutify to make the minified version more readable.

To see the difference, I went through the link to file you sent, but that file seems longer, like 200 lines longer, maybe it's just the formatting. So what I did was to commit a beautified version of the original file in master and then put a second commit on the top of that one. You can see [the difference here ](https://github.com/mysociety/fixmystreet/pull/4560/commits/3bab3766a01c361b671c89564966a126e516aa19)

Let me know if this is good enough to see the differences, otherwise I might ask you for some help with that.

lucascumsille:
@dracos I did the 2x2 grid, my only issue with that approach and why I didn't go that option at first, because with long categories, users will need to do some horizontal scrolling when the menu is open. Not sure how much of a dealbreaker that will be.

Screenshot 2023-08-28 at 08 32 07
Screenshot 2023-08-28 at 09 13 42

templates/web/base/reports/_list-filters.html
dracos: As it's no longer a sentence, each bit can be translated separately, more like standard form labels e.g.
```
<label for="statuses">[% loc('Show reports') %]</label>
[% mark_safe(select_status) %]
<label for="filter_categories">[% loc('About') %]</label>
[% mark_safe(select_category) %]
```

dracos:
@lucascumsille If I change your commit link to https://github.com/mysociety/fixmystreet/pull/4560/commits/3bab3766a01c361b671c89564966a126e516aa19?w=1 to remove whitespace changes, then yes, that's much clearer isn't it? :) It looks like we could update the core code to allow a wrapper div to the label, which would cover much of the change, I think? The classes can be customised at launch as per the docs on https://github.com/mysociety/jquery-multi-select so that should be fine.

Horizontal scrolling already does exist in that situation currently - depends if we think this new way would have more space or less. I think we could use smaller text in general, it doesn't need to be bigger than the main problem list text, I would assume? That would help. I don't think we've had any complaints about it being too small. Or perhaps not 50/50 - the category one could probably be wider than the open/closed/fixed one. Also, I don't think the spacing between the two rows needs to be as large? 

The way it's done there's no way to have it overlap the map annoyingly, and can't see an easy way for it to do so. We could probably split on a '/` easily enough in that particular example.

lucascumsille:
> @lucascumsille If I change your commit link to [3bab376?w=1](https://github.com/mysociety/fixmystreet/commit/3bab3766a01c361b671c89564966a126e516aa19?w=1) to remove whitespace changes, then yes, that's much clearer isn't it? :) It looks like we could update the core code to allow a wrapper div to the label, which would cover much of the change, I think? The classes can be customised at launch as per the docs on https://github.com/mysociety/jquery-multi-select so that should be fine.

>that's much clearer isn't it? :) 
@dracos definitely, much better =)
Sorry to clarify a couple of things:

- Basically  we could keep using the `jquery.multi-select.min.js` original file, but I'll just add the custom classes on `FixMyStreet.js`?

>It looks like we could update the core code to allow a wrapper div to the label, which would cover much of the change, I think?
- In here are we talking about including the label on line 932(`$(multiSelectGroup).append(span, multiSelect, select);`) in the `fixMmystreet.js` file?


> Horizontal scrolling already does exist in that situation currently - depends if we think this new way would have more space or less. I think we could use smaller text in general, it doesn't need to be bigger than the main problem list text, I would assume? That would help. I don't think we've had any complaints about it being too small. Or perhaps not 50/50 - the category one could probably be wider than the open/closed/fixed one. Also, I don't think the spacing between the two rows needs to be as large?

With this one I'll play around with this and see what would be better option.

dracos:
Sorry, I could have been clearer. Two things:
* I meant rather than this bit - https://github.com/mysociety/fixmystreet/commit/3bab3766a01c361b671c89564966a126e516aa19?w=1#diff-e4c184a6c747ed805c439d3ab007bd3b1da6a9d8a88482a41e6d8bb49dfe00f6R118-R125 (and the similar bit a few lines further down), we should alter the core https://github.com/mysociety/jquery-multi-select code to add a new config option for an optional wrapper div, which would then do this change in there;
* Then yes, you could set the custom classes needed in the FMS code.

> In here are we talking about including the label on line 932($(multiSelectGroup).append(span, multiSelect, select);) in the 
 fixmystreet.js file?
No, I hadn't actually looked at `add_wrapper` yet, not sure exactly what that's doing but sure it'll be fine :)

lucascumsille:
Superseded by https://github.com/mysociety/fixmystreet/pull/4728

PR #4562, opened 2023-08-14
[Brent] Change title label in Open311 service req description for ATAK reports.

neprune:
https://github.com/mysociety/societyworks/issues/3835

### DIFF
--- a/perllib/Open311.pm
+++ b/perllib/Open311.pm
@@ -260,7 +260,11 @@ sub _generate_service_request_description {
         $description .= "detail: " . $problem->detail . "\n\n";
         $description .= "url: " . $extra->{url} . "\n\n";
         $description .= "Submitted via FixMyStreet\n";
-        $description = "title: " . $problem->title . "\n\n$description";
+        if ($problem->cobrand eq 'brent') {
+            $description = "location of problem: " . $problem->title . "\n\n$description";
+        } else {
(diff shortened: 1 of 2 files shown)


neprune:
> Working on figuring out why (only) perl 5.26 thinks the ATAK category test report should be treated as a waste category: https://github.com/mysociety/fixmystreet/actions/runs/5856953323/job/15877754878?pr=4562

Looks like the re-run fixed it 🤔

perllib/Open311.pm
```diff
@@ -260,7 +260,11 @@ sub _generate_service_request_description {
         $description .= "detail: " . $problem->detail . "\n\n";
         $description .= "url: " . $extra->{url} . "\n\n";
         $description .= "Submitted via FixMyStreet\n";
-        $description = "title: " . $problem->title . "\n\n$description";
+        if ($problem->cobrand eq 'brent' && $problem->contact->email =~ /ATAK/) {
```
chrismytton, line 263: Brent's `new_report_title_field_label` is "Location of the problem", so I think you can drop the `&& $problem->contact->email =~ /ATAK/` bit as this will be useful for non-ATAK reports as well.

chrismytton (approved):
Just one suggestion about expanding this to cover all backends, but otherwise looks good!

PR #4563, opened 2023-08-15
[Peterborough] Update text for no access bin reports

chrismytton:
Fixes FD-3263


dracos (approved):
Not entirely sure why we have Black Bin and Refuse service names etc but this should work fine :)

PR #4565, opened 2023-08-15
[Central Beds] Send staff reporter name for Jadu reports.

neprune:
https://github.com/mysociety/societyworks/issues/3856

[skip changelog]


dracos (approved):
Making one of the reports in one of the existing tests set `contributed_by` to check this is okay would be good, but looks fine :)

(No need to change, but for interest you can do the get/if check in one line, like this:

```
if (my $contributed_by = $row->get_extra_metadata('contributed_by')) {
 ...
}
```
with $contributed_by then only available inside the if/else blocks.

PR #4566, opened 2023-08-15
Improved width for map area in around page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3784

Decreased the width for the .map_side element. This should make the map to have more space on medium screens portrait mode, like tablets.

### 48em to 63em
Screenshot 2023-08-15 at 14 47 32

### 63em to 79em
Screenshot 2023-08-15 at 14 48 08

### >79em
Screenshot 2023-08-15 at 14 48 27

### 48em to 63em
Screenshot 2023-08-15 at 14 49 04

### 63em to 79em
Screenshot 2023-08-15 at 14 49 51

### >79em
Screenshot 2023-08-15 at 14 50 26

----
`48-63em: 22em (46% … 35%)`
In the issue there is a proposed width for the `.map_sidebar` element of 22em. However at that size we the filter in the around area looks like this:

Screenshot 2023-08-15 at 14 51 49

Considering that is not a massive difference we could comeback to this PR after seeing the outcome of [this other PR](https://github.com/mysociety/fixmystreet/pull/4560), where the layout of the filter might change and therefore we could decrease the `.map_sidebar` width further.

Let me know what you think.

### DIFF
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -2,13 +2,17 @@
 
 $layout_front_stats_color: $primary !default;
 $mappage-header-height: 4em !default; // eg: might want this to equal outer height of #site-header on normal pages
-$mappage-sidebar-width: 29em !default;
-$mappage-sidebar-padding: 1em;
+$screen-size--medium: 63em !default;
+$screen-size--large:  79em !default;
+$mappage-sidebar-width: 26em !default;
 $mappage-notes-width: 15em;
-$mappage-actions-width: 25em !default;
+$mappage-actions-width: 18em !default;
+$mappage-sidebar-width--large:  28em !default;
+$mappage-actions-width--large:  20em !default;
+$mappage-sidebar-width--xlarge: 29em !default;
+$mappage-actions-width--xlarge: 25em !default;
+$mappage-sidebar-padding: 1em;
 $mappage-actions-sidebar-width: $mappage-sidebar-width !default;
-$mappage-sidebar-width--medium: 24em !default;
-$mappage-actions-width--medium: 20em !default;
 $header-top-border-width: 0.25em !default;
 $header-top-border: $header-top-border-width solid $primary !default;
 
@@ -252,6 +256,14 @@ body.mappage.admin {
   height: auto; // stretch from bottom of header to bottom of window
   margin: 0;
 
+  @media only screen and (max-width: $screen-size--large) and (min-width: $screen-size--medium)  {
+    #{$left}: $mappage-sidebar-width--large;
+  }
+
+  @media (min-width: $screen-size--large) {
+    #{$left}: $mappage-sidebar-width--xlarge;
+  }
+
   .with-notes & {
     #{$left}: $mappage-sidebar-width + $mappage-notes-width;
   }
@@ -259,8 +271,12 @@ body.mappage.admin {
   .with-actions & {
     #{$left}: $mappage-actions-sidebar-width + $mappage-actions-width;
 
-    @media (max-width: 79em) {
-      #{$left}: $mappage-sidebar-width--medium  + $mappage-actions-width--medium;
+    @media only screen and (max-width: $screen-size--large) and (min-width: $screen-size--medium)  {
+      #{$left}: $mappage-sidebar-width--large + $mappage-actions-width--large;
+    }
+  
+    @media (min-width: $screen-size--large) {
+      #{$left}: $mappage-sidebar-width--xlarge + $mappage-actions-width--xlarge;
     }
   }
 }
@@ -284,6 +300,14 @@ body.mappage.admin {
     background-image: linear-gradient(flip(90deg, 270deg), transparent 29em, #E9F2FF 29em);
   }
 
+  @media only screen and (max-width: $screen-size--large) and (min-width: $screen-size--medium)  {
+    width: ($mappage-sidebar-width--large - ($mappage-sidebar-padding * 2));
+  }
+
+  @media (min-width: $screen-size--large) {
+    width: ($mappage-sidebar-width--xlarge - ($mappage-sidebar-padding * 2));
+  }
+
   .with-actions & {
     width: $mappage-actions-sidebar-width + $mappage-actions-width;
     max-width: 100%; // Secondary column will squish on screens 768-960px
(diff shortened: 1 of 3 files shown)


lucascumsille:
Thanks for the feedback @dracos 

- [ ] 1 Regarding point 1 How  can I replicate that? I tested on the following page `report/585`, basically I just used a random report.
- [ ] 2
- [ ] 3
- [ ] 4 ·  When trying to login I get this error:
```docker-fixmystreet-1  | [error] Caught exception in FixMyStreet::App::Controller::Auth->email_sign_in "No next::method 'contact_name' found for FixMyStreet::Cobrand::HighwaysEngland at /usr/lib/aarch64-linux-gnu/perl/5.32/mro.pm line 30."
docker-fixmystreet-1  | 172.27.0.1 - - [28/Aug/2023:11:47:19 +0000] "GET /favicon.ico HTTP/1.1" 200 15086 "https://highwaysengland.fixmystreet.localhost:3000/auth" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36"```

- [ ] 5 I think I rather wait to merge and deploy https://github.com/mysociety/fixmystreet/pull/4560 before handle this issue.

<img width="1907" alt="Screenshot 2023-08-28 at 12 47 40" src="https://github.com/mysociety/fixmystreet/assets/13790153/dc79136c-75a1-48f5-9481-9867f7d8e9cd">

**Comment by dracos** [comment]
1. If you're logged in as admin, then on your /report/585 click the admin link top of the blue column to go to /admin/report_edit/585 and see the issue there.
2. Still seems to be an issue here - when not logged in, /report/585 has a gap to the right of the bottom links bar.

4. Have rebased your branch on top of current master, guess we can see if this still happens.
5. Yep, makes sense.

So I think all the sizing stuff here still needs looking at.

**Comment by lucascumsille** [comment]
@dracos Sorry this one has being a long wait. I had to rebase so we get the GOVUK styling for the filters. In the bright side I think the way we are managing the width in this PR for the multi-select-button it's probably better than what we currently have.

A few notes:

- For medium and below medium screen I decreased the sidebar of what was originally proposed, I still think when notes and inspector is displayed there is no much space for the map, but without changing the layout not sure how much we can do about it. (Potentially using a modal for the notes, but we would be adding more complexity for something that might not be so relevant)
- As you can see on the videos I have added a rule for the filters, where if they can't be at least 10rem, then they wrap vertically. From medium to large it means they use more vertical space, so a user might not be able to see the list of reports without scrolling vertically, if this is a concern then we can just add a rule for different screen sizes.
- For the key actions buttons when we are on medium screens, they wrapped in a way that is odd, happy to add a fix if we think is relevant.
Screenshot 2024-08-27 at 09 18 08


Here is a preview that takes in consideration the feedback you provided above:

https://github.com/user-attachments/assets/2d3ebbf7-289e-47a7-b42d-503ac5bea5ec


https://github.com/user-attachments/assets/1b720ab4-51f5-4b72-855d-9314895fecc4


https://github.com/user-attachments/assets/54ae67f1-73ff-4ea9-8938-6b512b4f595b

File: web/cobrands/sass/_report_list.scss [hunk 1]
```diff
@@ -152,32 +152,16 @@ input.bulk-assign {
   padding: 1rem;
 
   .filter-group {
-    display: flex;
-    flex-direction: row;
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(10rem, 1fr));
```
dracos, line 156: This looks a bit funny here (right margin of top row compared to underneath), that doesn't happen on live:
Screenshot 2024-08-29 at 17 57 15
Perhaps the `multi-select-button, .govuk-select` needs a width of 100%? Not sure I've understood it fully.

web/cobrands/sass/_layout.scss
```diff
@@ -759,7 +895,7 @@ textarea.form-error {
 // to make space for the #report-a-problem-sidebar.
 #side-form, #side {
   .with-notes & {
-    width: 27em;
+    width: 50%;
```
dracos, line 898: This breaks pages such as the /admin/report_edit and the Zurich /report/new , both of which use `with-notes`.

dracos:
Your last recording looks a bit strange; the /admin/report_edit/ page for a report that's wide now looks like this:
image

That was my point 1 previously; dunno if my proposed solution above would still be the right way to go, but the blue bar does not need to be as wide on the admin page, its contents are still the right width, so I think that needs doing differently. And the same applies to the Zurich reporting page, which also uses `with-notes`.

web/cobrands/sass/_layout.scss
```diff
@@ -280,8 +331,33 @@ body.mappage.admin {
   .with-notes & {
     width: (($mappage-sidebar-width + $mappage-notes-width) - ($mappage-sidebar-padding * 2));
     // TODO: Should have a bitmap image fallback for old browsers!!
-    background-image: -webkit-linear-gradient(flip(90deg, 270deg), transparent 29em, #E9F2FF 29em);
-    background-image: linear-gradient(flip(90deg, 270deg), transparent 29em, #E9F2FF 29em);
+    background-image: -webkit-linear-gradient(flip(90deg, 270deg), transparent 52%, #E9F2FF 52%);
```
dracos, line 334: I don't think 52% will be the equivalent of the widths given. Can we put the actual same width in, with media queries, so it appears at the right point? Or should we just ditch this and do it another way entirely?

web/cobrands/sass/_layout.scss
```diff
@@ -252,15 +271,47 @@ body.mappage.admin {
...
+    }
   }
 
   .with-actions & {
     #{$left}: $mappage-actions-sidebar-width + $mappage-actions-width;
 
-    @media (max-width: 79em) {
-      #{$left}: $mappage-sidebar-width--medium  + $mappage-actions-width--medium;
+    @media (min-width: $screen-size--medium) {
+      #{$left}: $mappage-sidebar-width--medium + $mappage-actions-width--medium;
```
dracos, line 306: The variable names were/are very confusing here, sorry.

With `.with-actions`, the map_sidebar width was either mappage-actions-sidebar-width + mappage-actions-width, or at <79em **mappage-sidebar-width--medium** + mappage-actions-width--medium. Which I know you've kept here, but you've now also started using that for the general width of the sidebar _without_ notes/actions as well, when it was only previously for with actions. If you did an equivalent to your second video when logged out, I think you'd see the sidebar could then be wider throughout (because there's no blue column), we don't need it to be the same for both.

Your variable names in this PR are better now, as you're varying the sidebar width in general, but I think you need to use your new mappage-actions-sidebar-width variables here and in the equivalent other place rather than what you have here. That then lets you specify different widths for when the actions bar is there and when it isn't.

Let me know if that doesn't make sense!

dracos (changes requested):

lucascumsille:
Replaced by https://github.com/mysociety/fixmystreet/pull/5592

PR #4573, opened 2023-08-17
[Brent] Update road message, and move to config.

dracos:
[skip changelog]


MorayMySoc (approved):
Looks good to me

dracos:
Awaiting Atak going live and then moving on top of that, as will impact on same code

PR #4575, opened 2023-08-18
[Glos] Prevent fetching of reports from Confirm for private categories 

nephila-nacrea:
[skip changelog]
Closes https://github.com/mysociety/societyworks/issues/3859.


chrismytton (changes requested):
Code looks good! But would be good to have a test for this, I think.

You might be able to use https://github.com/mysociety/fixmystreet/commit/6da4f2a979a87e82360ce024a9630f7194684718 as inspiration!

PR #4576, opened 2023-08-21
[K&S Bulky] Eligibility check for properties

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3844.

Kingston and Sutton have different property types that are eligible for bulky waste collection: https://docs.google.com/spreadsheets/d/14grAtRpKKb2ji_KbKZpmNQADpgFNANW_So4BK0uWobc/edit#gid=1790878445.

This PR also refactors the eligibility checks across bulky waste code to make them easier to follow.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandEcho.pm
+++ b/perllib/FixMyStreet/Roles/CobrandEcho.pm
@@ -52,6 +52,7 @@ sub look_up_property {
         address => FixMyStreet::Template::title($result->{Description}),
         latitude => $result->{Coordinates}{GeoPoint}{Latitude},
         longitude => $result->{Coordinates}{GeoPoint}{Longitude},
+        type_id => $result->{PointAddressType}{Id},
     };
 }
 
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -21,8 +21,7 @@ has index_template => (
 sub setup : Chained('/waste/property') : PathPart('bulky') : CaptureArgs(0) {
     my ($self, $c) = @_;
 
-    if (  !$c->cobrand->call_hook('bulky_enabled')
-        || !$c->cobrand->call_hook(bulky_allowed_property => $c->stash->{property})
+    if (  !$c->stash->{property}{show_bulky_waste}
         || $c->stash->{property}{pending_bulky_collection} )
     {
         $c->detach('/waste/property_redirect');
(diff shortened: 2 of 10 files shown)


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -172,4 +172,39 @@ sub garden_cc_check_payment_status {
 sub bulky_collection_time { { hours => 6, minutes => 0 } }
 sub bulky_cancellation_cutoff_time { { hours => 6, minutes => 0 } }
 
+sub bulky_allowed_property {
```
dracos, line 175: This function can be in the shared CobrandSLWP.pm

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -172,4 +172,39 @@ sub garden_cc_check_payment_status {
...
+    my ($self, $property) = @_;
+
+    return $self->bulky_enabled
+        && bulky_permitted_property_types()->{ $property->{type}{Id} };
+}
+
+sub bulky_permitted_property_types {
+    return {
+        1   => 'Detached',
```
dracos, line 184: This list should not be hard-coded in the code. Please change to use a configuration value, which we already do for general list of `address_types` available to be looked up in waste. `bulky_address_types` for example. Then `bulky_allowed_property` can look up if the value exists in the configured list. We don't need to store the name associated with each address type, that's never used.

dracos (changes requested):
I haven't checked the refactor commit, just notes on changes for the address type lookup, hopefully straightforward to change :)

dracos (changes requested):
Thanks, think it's nicer using config like that :) And thanks for the 'show' refactoring, I agree it's clearer now.

As the code is shared, I do not think we need identical tests for Kingston and Sutton. One will do, unless there is some non-configuration aspect upon which they will differ in which case that can then have a separate test. Hopefully there won't be (m)any :)

The only change I'm asking for is that due to your refactoring I think there are two instances in peterborough templates where a (bulky_enabled AND show_bulky_waste) can now become just a show_bulky_waste check.

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -52,6 +52,7 @@ sub look_up_property {
         address => FixMyStreet::Template::title($result->{Description}),
         latitude => $result->{Coordinates}{GeoPoint}{Latitude},
         longitude => $result->{Coordinates}{GeoPoint}{Longitude},
+        type => $result->{PointAddressType},
```
dracos, line 55: Sorry, actually thought of one more thing - could we only keep/use the Id here, as this is cached in the session I'd rather not store anything more than we need to.

dracos (approved):
Thanks, looks good :)

Ideally, please don't rebase/reorder fixups when there's going to be a re-review, because GitHub will by default automatically show me/you the commits/changes since the last review, but if you rebase them to reorder then that help is lost (e.g. in this particular case I had to work out that 516570a and 5503d28 were new fixups you'd added in the middle rather than it telling me). Once it's approved you can then reorder/rebase it all up :)

PR #4578, opened 2023-08-24
Added variables for btn-primary, green-btn and btn--primary

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3827
This should fix the accessibility issue with the white text against green background for the `.green-btn` class . A list of variables has been included, so the button can be customised according to a cobrand's branding guide.

NOTE: I went for a neutral default colour for the background, because some cobrands will use these colours as well, and the FMS yellow would look odd with their branding colour. Which brings the question, with this PR, the primary button has variables now, therefore we could update all cobrands so they use their own colours, instead of the default ones.  Let me know what you think so I can create the relevant ticket.

Screenshot 2023-08-24 at 07 21 37

[skip changelog]

### DIFF
--- a/web/cobrands/bathnes/_colours.scss
+++ b/web/cobrands/bathnes/_colours.scss
@@ -55,3 +55,14 @@ $search-help-header-font-size-desktop: 1.25em;
 $heading-font: arial, sans-serif;
 $body-font: arial, sans-serif;
 $meta-font: arial, sans-serif;
+
+
+// Base on this styling https://beta.bathnes.gov.uk/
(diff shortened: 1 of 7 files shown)


dracos (approved):
Make sense :) I imagine lots of cobrands already override this, so if using these variables could simplify them, that'd be good (though not sure how many do that or just change things more than that to use their own styling).

dracos:
I've merged this - I did wonder if you'd be able to tidy/remove some button cobrand CSS by now using these new variables, but you can presumably look at that whenever.

PR #4581, opened 2023-08-24
Added fix to .form-error width

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3873

Another alternative to accomplish the same result was to use `width: fit-content`, but it's not supported by IE11. Not that it matter according to [GOVUK](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices). Let me know if you prefer me to use that property instead.

### Before
Screenshot 2023-08-24 at 11 08 14

### After
Screenshot 2023-08-24 at 11 07 50

[Skip changelog]

### DIFF
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -768,7 +768,8 @@ textarea{
 /* form errors */
 div.form-error,
 p.form-error {
-  display:block;
+  display: block;
+  width: fit-content;
 }
 
 input.form-error,


dracos (changes requested):
Happy for you to use width: fit-content if that's the modern standard way of doing it - it's still functional fine in IE11, which is all we need.

dracos (approved):
Thanks :)

PR #4583, opened 2023-08-24
Added GOVUK styling to 'Show my name publicly'

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3825
Part of this milestone: https://github.com/mysociety/societyworks/milestone/99

Screenshot 2023-08-24 at 11 52 45

Screenshot 2023-08-24 at 11 52 24

[Skip changelog]


templates/web/base/header.html
```diff
@@ -22,6 +22,7 @@
         [% INCLUDE 'header/css.html' %]
         [% INCLUDE 'common_header_tags.html' %]
         [% TRY %][% PROCESS 'header_extra.html' %][% CATCH file %][% END %]
+        <link rel="stylesheet" href="/vendor/govuk-frontend/forms.css">
```
dracos, line 25: Could this be in header/css.html (two templates), and then all the other includes of it in various header_extra.html etc can be removed.

dracos (changes requested):
This one seems the smallest so thought might as well use this to get the forms.css in core properly and everywhere?

lucascumsille:
@dracos Thanks for the feedback. I just pushed a commit that adds the GOVUK stylesheet to core and then I removed it from all the `header_extra`

dracos:
Thanks. After some more investigation of this, we can import this in base.scss, replacing the radios.css that is already there (which is a subset of this CSS). I've force-pushed this branch with that change, let me know that it's still okay for you.

(Both files are being generated from https://github.com/mysociety/fixmystreet.com/tree/master/css if we ever need to make changes at source.)

lucascumsille:
> Thanks. After some more investigation of this, we can import this in base.scss, replacing the radios.css that is already there (which is a subset of this CSS). I've force-pushed this branch with that change, let me know that it's still okay for you.
> 
> (Both files are being generated from https://github.com/mysociety/fixmystreet.com/tree/master/css if we ever need to make changes at source.)

Thanks =) yeah it's fine with me and I can see it working.

PR #4547, opened 2023-08-02
[SLWP] Initial bulky set up

davea:
A few more bits refactored/added to SLWP role to get bulky feature enabled. This allows bookings to be made and passed to Echo (companion PR: https://github.com/mysociety/open311-adapter/pull/298) with what I think are sensible values.

Missing:
- Reserved slot expiry checking
- Correct payment info being sent

<details>
<summary>
Relevant config for open311-adapter's <code>council-sutton_echo.yml</code>:
</summary>

```yaml
service_whitelist:
  1636: 'Bulky collection'

service_extra_data:
  1636:
    payment: 1
    payment_method: 1
    Payment_Type: 1
    Collection_Date: 1
    Payment_Details_Payment_Reference: 1
    Payment_Details_Payment_Amount: 1
    Payment_Details_Payment_Failure_Reason: 1
    Bulky_Collection_Bulky_Items: 1
    Bulky_Collection_Notes: 1
    GUID: 1
    reservation: 1
    Customer Selected Date Beyond SLA?: 1
    First Date Returned to Customer: 1

service_id_override:
  1636: 413

default_data_event_type:
  1636:
    Payment Type: 1 # Online

waste_services:
  - 1636

```
</details>

For https://github.com/mysociety/societyworks/issues/3758 [skip changelog]

PR #4585, opened 2023-08-30
[SLWP] bulky slot expiry

nephila-nacrea:
Contributes to https://github.com/mysociety/fixmystreet/pull/4547.

- Checks if a reserved collection slot has expired when user submits summary page
- If it has, we
  - cancel the reserved slot
  - attempt to automatically reserve a new slot with the same date before continuing to payment
  - if this fails, we return the user to the slot selection page

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -108,7 +108,8 @@ sub index : PathPart('') : Chained('setup') : Args(0) {
     $c->forward('form');
 
     if ( $c->stash->{form}->current_page->name eq 'intro' ) {
-        $c->cobrand->call_hook(clear_cached_lookups_bulky_slots => $c->stash->{property}{uprn});
+        $c->cobrand->call_hook(
+            clear_cached_lookups_bulky_slots => $c->stash->{property}{id} );
     }
 }
 
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -103,15 +103,17 @@ has_page summary => (
         my $form = shift;
         my $c = $form->c;
 
-        my $slot_still_available
-            = $c->cobrand->call_hook(
-            check_bulky_slot_available => $form->saved_data->{chosen_date} );
+        # Some cobrands may set a new chosen_date on the form
+        my $slot_still_available = $c->cobrand->call_hook(
+            check_bulky_slot_available => $form->saved_data->{chosen_date},
+            form                       => $form,
+        );
 
         return 1 if $slot_still_available;
 
         # Clear date cache so user gets updated selection
         $c->cobrand->call_hook(
-            clear_cached_lookups_bulky_slots => $c->stash->{property}{uprn} );
+            clear_cached_lookups_bulky_slots => $c->stash->{property}{id} );
 
         $c->stash->{flash_message} = 'choose_another_date';
         $form->saved_data->{no_slots} = 1;
@@ -120,7 +122,6 @@ has_page summary => (
     finished => sub {
         return $_[0]->wizard_finished('process_bulky_data');
     },
-
 );
 
 has_page done => (
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -110,10 +110,9 @@ has_page summary => (
         return 1 if $slot_still_available;
 
         # Clear date cache so user gets updated selection
-        my $uprn = $c->stash->{property}{uprn};
-        for (qw/earlier later/) {
-            delete $c->session->{"peterborough:bartec:available_bulky_slots:$_:$uprn"};
-        }
+        $c->cobrand->call_hook(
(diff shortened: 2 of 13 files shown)


perllib/FixMyStreet/Roles/CobrandSLWP.pm
```diff
@@ -204,17 +204,36 @@ sub available_permissions {
...
+    }
+
+    # We also need to cancel the reserved slots in Echo
+    my $guid_key = $self->council_url . ":echo:bulky_event_guid:$id";
+    my $guid = $self->{c}->session->{$guid_key};
+
+    my $cfg = $self->feature('echo');
+    my $echo = Integrations::Echo->new(%$cfg);
+
+    $echo->CancelReservedSlotsForEvent($guid);
```
davea, line 234: The first time this is called `$guid` is empty and the resulting SOAP call returns HTTP 500 - we should probably skip the call in that case.

t/app/controller/waste_kands_bulky.t
```diff
@@ -116,6 +116,7 @@ FixMyStreet::override_config {
     $echo->mock( 'GetServiceUnitsForObject', sub { [] } );
     $echo->mock( 'GetTasks',                 sub { [] } );
     $echo->mock( 'GetEventsForObject',       sub { [] } );
+    $echo->mock( 'CancelReservedSlotsForEvent', sub { [] } );
```
davea, line 119: Related to above, could check that this mock (and the others) is always called with a non-nil guid value.

perllib/FixMyStreet/Roles/CobrandSLWP.pm
davea: (also this ends up getting called just before the payment screen somehow, when it probably doesn't need to be, as the event is about to be created in Echo [which releases](https://docs.google.com/document/d/1oEXsphrObtoXnf_o_EMBbnjmYO9roWL5/edit#bookmark=id.f8joworzs9l) the reserved slots)

davea (changes requested):
Looking good, think there's just a couple of small tweaks to the `CancelReservedSlotsForEvent` behaviour needed.

perllib/FixMyStreet/Roles/CobrandSLWP.pm
nephila-nacrea: Is it an issue that it gets called just before payment? Looks like it's due to the `clear_cached_lookups_property` call under `add_report` in perllib/FixMyStreet/App/Controller/Waste.pm.

davea: Maybe, maybe not. The call seemed to succeed when I tried it at least.

I suppose there is a small chance that the following occurs:

 1. user has a reservation on the final available slot for a date
 2. which is then released by the `CancelReservedSlotsForEvent` call just before they pay
 3. another user gets a reservation on that last slot which was just released
 4. The `PostEvent` for the first user's booking happens (by the report being sent to open311-adapter and submitted to Echo) but the slot was already taken
 5. I don't know what happens next?

In fairness there should only be a few seconds between the events of steps 2 and 4 so the chance of step 3 happening is very very small. Either way, if there's a simple way to skip the call if it happens from `add_report` it is probably worth doing.

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -108,7 +108,13 @@ sub index : PathPart('') : Chained('setup') : Args(0) {
     $c->forward('form');
 
     if ( $c->stash->{form}->current_page->name eq 'intro' ) {
-        $c->cobrand->call_hook(clear_cached_lookups_bulky_slots => $c->stash->{property}{uprn});
+        my $id_for_cache
```
dracos, line 111: Looking at master Peterborough code, there is some other confusion as `clear_cached_lookups_property` is called with both an ID and a URPN, so some clearing of the bulky slots isn't working anyway when called from there rather than directly like here.

I think there is a solution, which is change the setting of available_bulky_slots in Peterborough's `find_available_bulky_slots` to use property->{id} rather than {uprn}. Then everything is using ID for the key, not UPRN, and everything to do with clear_cached_lookups* can only use ID and not worry about the UPRN at all, including here

t/app/controller/waste_kingston.t
```diff
@@ -306,6 +307,7 @@ FixMyStreet::override_config {
         };
     });
     $echo->mock('GetServiceUnitsForObject', \&garden_waste_one_bin);
+    mock_CancelReservedSlotsForEvent($echo);
```
dracos, line 310: I think you only need to mock this the first time, it never changes :)

perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -103,25 +103,29 @@ has_page summary => (
...
+        );
 
         return 1 if $slot_still_available;
 
         # Clear date cache so user gets updated selection
-        my $uprn = $c->stash->{property}{uprn};
-        for (qw/earlier later/) {
-            delete $c->session->{"peterborough:bartec:available_bulky_slots:$_:$uprn"};
-        }
+        my $id_for_cache
```
dracos, line 115: This similarly needs updating.

dracos (changes requested):
Missed one other change to make is all

dracos (approved):

PR #4587, opened 2023-08-31
[SLWP] 3759 kands bulky waste email notifications

MorayMySoc:
Moves the code for reminder emails into general role.

Adds templates for Kingston and Sutton for confirmation email and clauses in reminder template for K&S.

For: https://github.com/mysociety/societyworks/issues/3759

[skip changelog]


templates/email/sutton/other-reported-bulky.html
```diff
@@ -0,0 +1,78 @@
...
+[% END %]
+
+  <p style="[% p_style %]">
+    If you wish to cancel your booking, please visit <a href="[% cobrand.base_url %]/waste/[% report.get_extra_field_value('property_id') | uri %]/bulky_cancel">this link</a>.
+  </p>
+
+  <p style="[% p_style %]">
+    Items must be accessible and clearly visible from the front of your house, placed as close to the pavement as possible (but not on it). Our collections crews will not enter your home or retrieve items from your back garden.
+    If you live in a block of flats, you can place your items (on the morning of collection) in your communal bin store area. If the bin store has an access code you should have supplied that during the booking as we won't be able to collect otherwise.
+    Items must be out for collection by 6am on the collection day.
```
dracos, line 61: Can we get the time here from the cobrand function `bulky_nice_collection_time`? Then kingston/sutton could share the email templates. In fact, the changes with Peterborough seem small enough, could we have them all in the one template with a few IFs?

templates/email/kingston/_bulky_data.html
```diff
@@ -0,0 +1,9 @@
+[%
+
+SET payment = report.get_extra_field_value('Payment_Details_Payment_Amount');
+SET collection_date = cobrand.bulky_nice_collection_date(report);
+SET report_date = cobrand.bulky_nice_collection_date(report.confirmed);
```
dracos, line 5: I think this function should only be used on the report/collection date string, not a DateTime from the database, - use strftime directly otherwise I think.

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -577,6 +577,8 @@ sub waste_munge_bulky_data {
             push @notes, $data->{"item_notes_$_"} || '';
             push @ids, $items{$item};
             push @photos, $data->{"item_photos_$_"} || '';
+            my $two = sprintf("%02d", $_);
+            $data->{"extra_ITEM_$two"} = $data->{"item_$_"};
```
dracos, line 581: Whilst I see the reason for doing this, it's Bartec fields that is calling them ITEM_01 ITEM_02 etc, and I'm not sure we should base our storage around that. I also don't really want to clutter up the extra "fields" data with stuff that isn't for the backend, if that makes sense.
In 0288ed0292ea97ad13677c676c3a5a9ec4102b22 - for the waste_reconstruct_bulky_data function I got the list back from the IDs in Bulky_Collection_Bulky_Items - which is also not ideal because if they change the list of things, that wouldn't be able to display either.

So perhaps thinking about it we need a combination of both; we should store the items along with the report, for reference, but separately from backend data. So I guess I'd store it in the report's extra metadata (not fields) - I see in Waste.pm we store item_photo and location_photo as extra_metadata, perhaps we can store the list of items here as well. Perhaps not for Peterborough as it'd be duplicate data, but perhaps easier to do so? And then your bulky nice item list could use the relevant location (and my bulky reconstruct data also).

dracos (changes requested):

templates/email/kingston/_bulky_data.html
dracos: This doesn't appear to have changed, nor in Sutton.

dracos: Was it meant to be removed now default exists?

templates/email/default/waste/other-reported-bulky.html
```diff
@@ -72,5 +90,3 @@
     <h2 style="[% h2_style %]">[% report.title | html %]</h2>
     [% report.detail | html_para_email(secondary_p_style) %]
 [% END %]
-
-[% INCLUDE '_email_bottom.html' %]
```
dracos, line 76: I think this line is needed?

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -552,4 +552,34 @@ sub check_bulky_slot_available {
...
+    foreach (grep { /^item_\d/ } keys %$data) {
+        my $k = $_;
+        $self->{c}->stash->{report}->set_extra_metadata($k => $data->{$_});
+    }
+}
+
+sub bulky_nice_item_list {
+    my ($self, $report) = @_;
+
+    my @items = grep { /^item_\d/ } keys %{$report->get_extra_metadata};
```
dracos, line 567: Do we want to sort numerically somehow so that they're in the order they were given?

dracos (changes requested):
Looks good, a few small changes on top I think is all

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -552,4 +552,34 @@ sub check_bulky_slot_available {
...
+    my @fields;
+    for my $item (@items) {
+        if (my $value = ${$report->get_extra_metadata}{$item}) {
```
dracos, line 570: Think this should be `$report->get_extra_metadata($item)`.

templates/email/default/waste/_bulky_data.html
```diff
@@ -0,0 +1,10 @@
+[%
+
+SET payment = report.get_extra_field_value('Payment_Details_Payment_Amount');
```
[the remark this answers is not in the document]
MorayMySoc: Are you saying that Peterborough sets the extra field 'payment' too? So ```SET payment = report.get_extra_field_value('payment');``` would cover all cobrands?

dracos (approved):
Looks good, only thing I'd ask when squashing/rebasing is that you put Peterborough's `bulky_nice_item_list` back where it was, so it's not appearing in the diff? :)

dracos:
Did the rebase/squash/move after merging, dropped commit 205ad36bf0182fba373cc37b414fa89221aa7398 as was reverted through later commits.

PR #4588, opened 2023-09-01
[SLWP] Hide 'show_later_dates' option for bulky slot selection

nephila-nacrea:
Functionality for 'show_later_dates' for K&S is a) currently broken and b) is not needed, so we simply deactivate that field.

[skip changelog]


dracos (approved):
Neat way of doing it :)

PR #4590, opened 2023-09-05
[Brent] Skip sending updates for ATAK backend.

davea:
Similarly to 22fd8ee6f0b9d for Central Beds Jadu, we don't want to send updates to ATAK.

[skip changelog]


dracos (approved):
:atom:

PR #4594, opened 2023-09-06
[Brent] Correct fields for Echo

dracos:
For https://github.com/mysociety/societyworks/issues/3883
I wondered if it'd be better to have the checkbox flags at the open311-adapter end, but that also seemed fragile in that it would have to e.g. parse the Notes field being passed through (given no individual data is). This is probably okay?
[skip changelog]


chrismytton (approved):
I think having it in FMS rather than open311-adapter makes sense, as you say parsing the notes field is going to be a bit messy.

PR #4595, opened 2023-09-06
[Brent] [Waste] Initial small items setup.

dracos:
Provides an initial small items form, pretty much identical to bulky apart from wording and other small differences (e.g. sort order, photos). Uses the same controller/role/flags as Brent is having small items instead of bulky, and given timescales seemed quite a bit more straightforward than trying to allow both.

Takes a little bit of @MorayMySoc code for saving the items in the report metadata (though only if set), and little bit of @nephila-nacrea code for calling the clear cached slot.

[skip changelog]


chrismytton (approved):
Looks good! One comment but doesn't necessarily needs actioning.

PR #4597, opened 2023-09-08
[Brent][WW] Small items making a booking #3869

MorayMySoc:
Adds count of how many small WEEE items have been selected  and how many different 'types' of item have been selected.

Adds errors if more that 4 small WEEE or more than 3 types.

Adds text to clarify the limits allowed.

Adds tests for above.

https://github.com/mysociety/societyworks/issues/3869#issuecomment-1700920860

[skip changelog]


dracos:
(just changed base, and rebased on top of current brent-initial-setup so to have smaller diff)

PR #4599, opened 2023-09-08
[Brent][WW] Small items eligibility

MorayMySoc:
Allows any Brent resident with a domestic refuse collection to make a small items collection request.

In process fixes small error in Brent test mock data that is confusing but wasn't causing an issue.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -727,6 +727,7 @@ sub bin_services_for_address {
         $expired{$_->{Id}} = $schedules if $self->waste_sub_overdue( $schedules->{end_date}, weeks => 4 );
 
         next unless $schedules->{next} or $schedules->{last};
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -1345,9 +1344,20 @@ sub bulky_can_refund { 0 }
...
+=cut
+
 sub bulky_allowed_property {
-    my ( $self, $property ) = @_;
-    return $self->bulky_enabled;
+    my ($self, $id) = @_;
+
+    return $self->bulky_enabled
+            && $id
+            && grep {$id eq $_} qw/262/;
```
dracos, line 1360: I guess this was in case there was more than one in future? For now, I assume this could be `return $self->bulky_enabled && $id == 262;` :) (it'll always be set, I assume)

dracos (approved):
One suggestion is all, looks good

PR #4602, opened 2023-09-11
[Westminster] Clearer confirmation message when passing report to TfL

davea:
For FD-2916

[skip changelog]

### DIFF
--- a/t/cobrand/bucks.t
+++ b/t/cobrand/bucks.t
@@ -220,7 +220,7 @@ subtest 'Flytipping not on a road going to HE does not get recategorised' => sub
             'road-placement' => 'off-road',
         }
     }, "submit details");
-    $mech->content_contains('We don&rsquo;t handle this type of problem');
+    $mech->content_contains('From the information you have given, we have passed this report on to:');
     my $report = FixMyStreet::DB->resultset("Problem")->search(undef, { order_by => { -desc => 'id' } })->first;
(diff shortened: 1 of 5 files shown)


dracos (commented):
Looks fine, but any reason we don't apply this to base? Assuming it also applies to any other TfL-category, or eg two-tier if it's a category that you could say either could handle in some cases.

davea:
Good point. No reason other than I'm always marginally reluctant to change base templates :) Will do so now.

MorayMySoc:
Also been requested to add a TfL logo for TfL reports - could only see banners and not sure whether we'd have to discuss with TfL first anyway?

dracos:
I think we just say we can't do that, yeah

t/cobrand/bucks.t
```diff
@@ -220,7 +220,7 @@ subtest 'Flytipping not on a road going to HE does not get recategorised' => sub
             'road-placement' => 'off-road',
         }
     }, "submit details");
-    $mech->content_contains('We don&rsquo;t handle this type of problem');
+    $mech->content_contains('Due to the location we don&rsquo;t handle this particular problem so have passed it on to:');
```
dracos, line 223: Test not updated?

dracos (approved):

chrismytton (approved):

PR #4603, opened 2023-09-11
[Brent][WW] Small items collection - Email notifications

MorayMySoc:
Changes templates to adapt for Brent and to add small items in relevant places instead of bulky waste

Adds tests

https://github.com/mysociety/societyworks/issues/3871

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -1440,4 +1440,7 @@ sub waste_reconstruct_bulky_data {
     return $saved_data;
 }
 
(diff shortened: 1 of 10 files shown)


dracos (changes requested):
One typo. I see all the overlap with K&S, hopefully will be easy to merge together at the time, given looks like all text will be the same. Wonder if we just include the item list in all reminders, then no need for some of the cobrand checks (and could always set `sent_confirm_id_ref` as well I think), and similar with report date in other-reported. Then we could put this on master directly and have the Brent (smaller) bits on top

dracos (approved):

PR #4605, opened 2023-09-11
[Bulky waste] Framework for multiple bookings, different way of cancelling

dracos:
Various aspects brought together to make future merging easier:
* Add a report ID to cancel URLs, and list all bookings (no change/point at present, but as/when multiple bookings are allowed, these will be used)
* Add days_before, for times when cancellation might eg be two days before (though don't think is at present).
* Add a way for cancellation to be done by adding an update rather than a new report - most non-Peterborough will work this way, it looks like.

[skip changelog]


MorayMySoc (approved):
Ok, all seems to make sense

PR #4608, opened 2023-09-11
K&S Bulky: Send email if Echo update fails

nephila-nacrea:
If Echo fails to send us an update after payment has been taken for a bulky collection, we send an email to Kingston or Sutton as appropriate, so the collection can be booked manually their end.

Closes https://github.com/mysociety/societyworks/issues/3842.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandEcho.pm
+++ b/perllib/FixMyStreet/Roles/CobrandEcho.pm
@@ -8,6 +8,7 @@ use Moo::Role;
 use Sort::Key::Natural qw(natkeysort_inplace);
 use UUID::Tiny ':std';
 use FixMyStreet::DateRange;
+use FixMyStreet::DB;
 use FixMyStreet::WorkingDays;
 use Open311::GetServiceRequestUpdates;
 
@@ -751,6 +752,83 @@ sub bulky_nice_item_list {
         },
         @fields,
     ];
+}
+
+sub send_bulky_payment_echo_update_failed {
+    my ( $self, $params ) = @_;
+
+    my $email
+        = ( $self->feature('waste_features') || {} )
+        ->{echo_update_failure_email};
+    return unless $email;
+
+    # 3 hours to allow for Echo downtime
+    my $dtf = FixMyStreet::DB->schema->storage->datetime_parser;
+    my $cutoff_date
+        = $dtf->format_datetime( DateTime->now->subtract( hours => 3 ) );
+
+    my $rs = FixMyStreet::DB->resultset('Problem')->search(
+        {   category => 'Bulky collection',
+            cobrand  => $self->moniker,
+            created  => { '<'  => $cutoff_date },
+            extra    => { '\?' => 'payment_reference' },
+            -not => {
+                extra => {
+                    '\?' => [
+                        'echo_update_failure_email_sent',
+                        'echo_update_sent',
+                    ],
+                },
+            },
+        },
+    );
+
+    while ( my $report = $rs->next ) {
+        # Ignore if there is an update in Comment table with an external_id
+        if ( $report->comments->search( { external_id => { '!=', undef } } )
+            ->count > 0
+        ) {
+            # Set flag so we don't repeatedly check the comments for this
+            # report
+            $report->set_extra_metadata( echo_update_sent => 1 );
+            $report->update;
+
+            next;
+        }
+
+        # Send email
+        my $h = {
+            report  => $report,
+            cobrand => $self,
+        };
+
+        my $result = eval {
+            FixMyStreet::Email::send_cron(
+                FixMyStreet::DB->schema,
+                'waste/bulky_payment_echo_update_failed.txt',
+                $h,
+                { To => $email },
+                undef,    # env_from
+                $params->{nomail},
+                $self,
+                $report->lang,
+            );
+        };
+
+        if ($@) {
+            warn 'Sending for report ' . $report->id . " failed: $@\n";
+        } elsif ($result) {
+            print 'Sending for report ' . $report->id . ": failed\n"
+                if $params->{verbose};
+        } else {
+            $report->set_extra_metadata(
+                echo_update_failure_email_sent => 1 );
+            $report->update;
+
+            print 'Sending for report ' . $report->id . ": succeeded\n"
+                if $params->{verbose};
+        }
+    }
 }
 
 1;
--- a/t/app/controller/waste_kands_bulky.t
+++ b/t/app/controller/waste_kands_bulky.t
@@ -74,6 +74,7 @@ FixMyStreet::override_config {
             kingston => {
                 bulky_enabled => 1,
                 bulky_tandc_link => 'tandc_link',
+                echo_update_failure_email => 'fail@example.com',
             },
         },
         echo => {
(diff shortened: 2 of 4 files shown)


bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+    require "$d/../setenv.pl";
+}
+
+use DateTime;
+use FixMyStreet::Cobrand;
+use FixMyStreet::DB;
+use FixMyStreet::Email;
+use mySociety::Locale;
+
+mySociety::Locale::gettext_domain('FixMyStreet');
```
dracos, line 19: I'm not sure what this is here for, it shouldn't be necessary in any cron scripts? (It isn't in any others that I can see.) At most you might call `CronFns::language()` (that send-alerts/reports do), but other scripts that send emails don't, so I'd be interested to know what issue was leading to this.

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+my $comment_text = 'Echo failure email sent';
+
+my %cobrands = (
+    kingston => FixMyStreet::Cobrand->get_class_for_moniker('kingston')->new,
+    sutton   => FixMyStreet::Cobrand->get_class_for_moniker('sutton')->new,
+);
+
+my $dtf         = FixMyStreet::DB->schema->storage->datetime_parser;
+my $cutoff_date = $dtf->format_datetime(
+    DateTime->now( time_zone => 'Europe/London' )->subtract( hours => 1 ) );
```
dracos, line 30: Echo sometimes has hours of maintenance downtime, so this could lead to issues during those periods. Perhaps 3 hours should cover it?

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+        cobrand  => [qw/kingston sutton/],
+        created  => { '<'  => $cutoff_date },
+        extra    => { '\?' => 'payment_reference' },
+    },
+);
+
+while ( my $report = $rs->next ) {
+    # Ignore if there is an update in Comment table with an external_id, or
+    # failure email has been sent previously
+    my $rs_comment = FixMyStreet::DB->resultset('Comment')->search(
```
dracos, line 43: You can use `$report->comments` here as a shorter way of referring to a report's comments.

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+        },
+        undef,    # env_from
+        undef,    # nomail
+        $cobrand,
+        $report->lang,
+    );
+
+    my $comment = FixMyStreet::DB->resultset('Comment')->new(
+        {   problem => $report,
+            text    => $comment_text,
```
dracos, line 82: I don't think we want to leave a public update on a report (that would get sent to the user). Instead of this, I would set a flag on the report object metadata, as some other scripts do (e.g. send-claims-to-echo sets `sent_to_evo` to 1).

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+        report  => $report,
+        cobrand => $cobrand,
+    };
+
+    FixMyStreet::Email::send_cron(
+        FixMyStreet::DB->schema,
+        'waste/echo_failed_bulky_payment.txt',
+        $h,
+        {   To   => $email,
+            From => [
```
dracos, line 69: You do not need From, send_cron defaults that for you.

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+        }
+    );
+
+    next if $rs_comment->count > 0;
+
+    # Send email
+    my $email
+        = $report->cobrand eq 'kingston'
+        ? 'uk.kingston.collections@veolia.com'
+        : 'neighbourhoodservices@sutton.gov.uk';
```
dracos, line 55: These email addresses should not be present in code, but read from configuration.

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+        extra    => { '\?' => 'payment_reference' },
+    },
+);
+
+while ( my $report = $rs->next ) {
+    # Ignore if there is an update in Comment table with an external_id, or
+    # failure email has been sent previously
+    my $rs_comment = FixMyStreet::DB->resultset('Comment')->search(
+        {   problem_id => $report->id,
+            -or => [ text => $comment_text, external_id => { '!=', undef } ],
```
dracos, line 45: As above, if you have a flag on the report, you can filter out ones already sent in your first query, and not need to check that here.

templates/email/default/waste/echo_failed_bulky_payment.txt
```diff
@@ -0,0 +1,40 @@
+Subject: Echo failed to confirm payment
+
+[%
+USE pounds = format('%.2f');
+PROCESS 'waste/_bulky_data.html';
+~%]
+
+Echo failed to confirm payment for the following bulky waste collection:
```
dracos, line 8: The language is the wrong way round - after successful payment, the user is redirected to us, we note this on the collection report as an update, and we then send that update to Echo to confirm payment has been received. It is if that update to Echo fails that we need to send this email. So this and other bits like that will need rewording. "We failed to update Echo about confirmed payment" etc

templates/email/default/waste/echo_failed_bulky_payment.txt
```diff
@@ -0,0 +1,40 @@
...
+
+    Name: [% report.name %]
+
+    Email: [% report.user.email %]
+
+    Phone: [% report.user.phone %] [% # TODO Where is this? %]
+
+Collection instructions: [% report.get_extra_field_value('Exact_Location') %]
+
+Error: [% # TODO Do we have access to any errors from Echo? %]
```
dracos, line 40: No, if something has gone wrong, we'll only have an internal error I imagine.

bin/email-failed-echo-bulky-payment
```diff
@@ -0,0 +1,87 @@
...
+use FixMyStreet::Cobrand;
+use FixMyStreet::DB;
+use FixMyStreet::Email;
+use mySociety::Locale;
+
+mySociety::Locale::gettext_domain('FixMyStreet');
+
+my $comment_text = 'Echo failure email sent';
+
+my %cobrands = (
```
dracos, line 23: As much as possible, so from here onwards ish, this should be code elsewhere (probably in CobrandEcho.pm?), so that tests can be added for it. (Then add a basic test for it.)

dracos (changes requested):
A good start, I've made a number of suggestions, let me know if any don't make sense.

bin/email-failed-echo-bulky-payment
nephila-nacrea: I was getting an error regarding an undefined $gettext in Locale.pm. `CronFns::language()` seems to work though.

templates/email/default/waste/echo_failed_bulky_payment.txt
nephila-nacrea: Should I be trying to fetch this internal error? Where would it be stored?

bin/email-failed-echo-bulky-payment
nephila-nacrea: Am pushing changes for your other comments in the meantime, but will get to work on this.

templates/email/default/waste/echo_failed_bulky_payment.txt
dracos: No, it will literally be the phrase "Internal error" :)

t/app/controller/waste_kands_bulky.t
```diff
@@ -74,11 +74,12 @@ FixMyStreet::override_config {
             kingston => {
                 bulky_enabled => 1,
                 bulky_tandc_link => 'tandc_link',
+                echo_update_failure_email => 'fail@kingston.gov.uk',
```
dracos, line 77: Use an example.com/net/org email by preference, please.

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -657,4 +658,67 @@ sub bulky_nice_item_list {
...
+        # Ignore if there is an update in Comment table with an external_id
+        next
+            if $report->comments->search( { external_id => { '!=', undef } },
+        )->count > 0;
+
+        # Send email
+        my $email
+            = ( $self->feature('waste_features') || {} )
+            ->{echo_update_failure_email};
+        next unless $email;
```
dracos, line 688: This check can happen pre-loop.

perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -657,4 +658,67 @@ sub bulky_nice_item_list {
...
+            extra    => { '\?' => 'payment_reference' },
+            -not => { extra => { '\?' => 'echo_update_failure_email_sent' } },
+        },
+    );
+
+    while ( my $report = $rs->next ) {
+        # Ignore if there is an update in Comment table with an external_id
+        next
+            if $report->comments->search( { external_id => { '!=', undef } },
+        )->count > 0;
```
dracos, line 682: This is going to check every successful Bulky collection report each time it's run, even if it's looked at it before. We should set a flag here as well so that it doesn't have to do that, I think.

dracos (changes requested):
Moving around all looks good, plus all the changes. Couple of suggestions to prevent duplicate checking, and small tweaks.

PR #4609, opened 2023-09-12
[Bromley] [Waste] Initial bulky goods setup.

neprune:


dracos:
Update: I've added a test for the view property price unchanging commit and moved that first. I brought in @MorayMySoc's fix for pricing because that also throws warnings in the view page (as that calls look_up_property but doesn't work out the domestic/trade and shouldn't need to) so called it with exclude_pricing there too.

PR #4610, opened 2023-09-12
[Rutland] Adds easting and northing to open311 data

MorayMySoc:
Request to add the easting and northing co-ordinates alongside the existing lat/long

https://mysocietysupport.freshdesk.com/a/tickets/3345


MorayMySoc:
Wrong approach - abandoned

PR #4611, opened 2023-09-13
[SLWP] Add text on the about you page

MorayMySoc:
Adds explanation for collection of text on about you page

https://3.basecamp.com/4020879/buckets/33314444/todos/6529451251

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky.pm
@@ -36,6 +36,7 @@ has_page cannot_book => (
 );
 
 has_page about_you => (
+    intro => 'bulky/about_you_preamble.html',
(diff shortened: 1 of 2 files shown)


MorayMySoc:
I can't see how to differentiate between the about_you page on bulky waste v elsewhere...

dracos (changes requested):
To answer your question about bulky, and perhaps easier way of doing it in general, how about adding an `intro` template to the bulky about_you page, then having `templates/web/base/waste/bulky/_about_you_preamble.html` which does a `cobrand.moniker` check.

(I see that Report/Request have cobrand-specific subclasses to have the same outcome but they also change the flow etc more so probably not worth doing that here.)

dracos (approved):
Looks good, could you make sure there's no whitespace changes to waste/form.html given you now haven't changed it at all.

PR #4613, opened 2023-09-14
FMS Help page Improvements

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3833

This is draft PR because: 

- When I started working on this issue, I noticed that the content navigation bar was also being shared by some cobrands, like Oxfordshire. Basically, the mobile behaviour was the same, stacking at the top. When I showed this to Sally, we thought it would be nice to fix this issue for everyone and not only for fms.com
- So far, I have updated some cobrands, but not all of them, because first, I'd like to ask one of the developers if the way we are handling the cobrands and file structure in this PR is the best approach. If it is, then I'll go ahead and change it for all cobrands, plus testing and modifying the variables in '[_page-with-sticky-content-bar.scss](https://github.com/mysociety/fixmystreet/compare/master...3833-help-page#diff-b0f7c7c650e9d1581b7066f9112025f1deff29246e3faebf2e612d61e7d8c763)'
- This last point also raises a couple of questions:
  - Not all cobrands use the sticky sidebar. Do we want to give this styling to every single cobrand or just to the ones that currently use the sticky style?
  - Of course, the style is changing(Check preview), which might raise some issues with some clients(I wouldn't imagine many of them would be very concerned, considering it's the help page).

### Preview so far

Desktop FMS
https://github.com/mysociety/fixmystreet/assets/13790153/949916c9-6ecf-45e1-845a-d66b43423e6b

Mobile FMS
https://github.com/mysociety/fixmystreet/assets/13790153/f529ad5b-45d7-4be0-8e19-3d20d943cc90

Desktop Oxfordshire
https://github.com/mysociety/fixmystreet/assets/13790153/7809533f-ba71-435c-80ee-70409ae21b19

Mobile Oxfordshire
https://github.com/mysociety/fixmystreet/assets/13790153/f70568bb-0fcf-4d65-a504-58f127f20637


lucascumsille:
### To resolve:

- [ ] What should we do with fiksgatami

lucascumsille:
I have added a few commits to resolve [the following](https://mysociety.slack.com/archives/C04BEET1F8X/p1694688038404149?thread_ts=1694687194.530979&cid=C04BEET1F8X):

- I would change/update twothirdswidthpage rather than add page-with-sticky-content-bar? I think every page with that class could be changed to your new one, so no need to only apply it to some (and hopefully that means less changes to things, makes it easier to have just one thing?)
- Most cobrands will have their own FAQ, yes, but hopefully you could get away without having to change most of them and only change the _sidebar as you’ve done (perhaps Oxfordshire could switch to using _sidebar?)
- privacy page does need a footer :slightly_smiling_face:
- More consistency is good, so don’t worry if some do/don’t at present, unless they’re a massive change, let’s be consistent across them all if possible

 I'll rebase after this PR has been approved

lucascumsille:
@MorayMySoc sorry to tag you, but when I added you was a last week and now that the PR is ready for review, not sure if you get a notification.

MorayMySoc (approved):
This seems to work nicely to me

templates/web/base/about/privacy.html
```diff
@@ -43,5 +43,3 @@ <h2>Cookies</h2>
         <td>When browser is closed, or four weeks if &ldquo;Keep me signed in&rdquo; is ticked</td>
     </tr>
 </table>
-
-[% INCLUDE 'footer.html' pagefooter = 'yes' %]
```
dracos, line 47: The privacy page needs a footer

templates/web/hart/about/faq-en-gb.html
```diff
@@ -1,4 +1,6 @@
-[% INCLUDE 'header.html', title = 'Help', bodyclass = 'fullwidthpage' %]
+[% INCLUDE 'header.html', title = 'Help', bodyclass = 'fullwidthpage twothirdswidthpage' %]
```
dracos, line 1: Shouldn't be both fullwidth and twothirdswidth

web/cobrands/sass/_base.scss
```diff
@@ -153,15 +153,26 @@ dl {
...
   padding: 0;
   dt {
-    font-size: 1em;
+    font-size: 1;
     line-height: 1.5em;
     font-weight: bold;
   }
   dd {
-    font-weight: 1em;
+    font-weight: 1;
```
dracos, line 161: Is this meant to be font-size?

lucascumsille:
@dracos thanks for the feedback. I just added three commits to fix the three observations above. Let me know if you are happy with it and I'll rebase and merge =)

dracos:
I rebased and pushed some changes to accomplish the following:
* Some "new" code was added then removed, but was removed with different spacing, which confused the diff;
* Fixed the Northamptonshire sidebar which was missing;
* Changed the aside-top to be 0 instead of 6em (doesn't make any visual difference at top, but sticks better to the top (plus margin)), and decrease the padding from 1.5em to 1em to match;
* Add a `:has(:not(div.sticky-sidebar))` so that there's hopefully no big left gap on pages that lack without a sidebar (e.g. FixaMinGata's /about/accessibility);
* removed bunch of CSS changing the sidebar `top` etc that's no longer needed;
* FiksGataMi sidebar fix;
* moved some sidebars to template tidying.
@lucascumsille Let me know if this all looks okay. Diff attached if useful.
[faq-diff.txt](https://github.com/mysociety/fixmystreet/files/12918416/faq-diff.txt)

dracos (commented):
(see above)

lucascumsille:
Thanks again @dracos It looks good to me. I was checking if anything was missing, it looked fine on desktop, but on mobile I added to commits:

- Glos was missing the mobile button to open the menu.
- Brent mobile button didn't have the right colour.

PR #4614, opened 2023-09-14
[SLWP] Add more detail to notes field

MorayMySoc:
Adds purpose of notes field under note label and makes label more direct

https://3.basecamp.com/4020879/buckets/33314444/todos/6529606114

[skip changelog]


dracos (changes requested):
I wouldn't put it in the template (note you haven't used a &lt;label&gt; for example), the way to add a hint to the fields is to have a `tags => { hint => 'Describe the item...' }` on the field, then it should all work/display for you.

dracos (approved):

PR #4615, opened 2023-09-14
[SLWP] Make location field mandatory.

dracos:
[skip changelog]


MorayMySoc (approved):
Seems good to me. Made a comment, but not essential

PR #4617, opened 2023-09-17
[Bromley] [Waste] Per-item pricing depending on property type.

neprune:
https://github.com/mysociety/societyworks/issues/3911

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
+++ b/perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
@@ -117,12 +117,13 @@ sub bulky_items_extra {
     my $self = shift;
 
     my $per_item = $self->bulky_per_item_costs;
+    my $price_key = $self->bulky_per_item_price_key;
 
     my $json = JSON::MaybeXS->new;
     my %hash;
     for my $item ( @{ $self->bulky_items_master_list } ) {
         $hash{ $item->{name} }{message} = $item->{message} if $item->{message};
-        $hash{ $item->{name} }{price} = $item->{price} if $item->{price} && $per_item;
+        $hash{ $item->{name} }{price} = $item->{$price_key} if $item->{$price_key} && $per_item;
         $hash{ $item->{name} }{max} = $item->{max} if $item->{max};
         $hash{ $item->{name} }{json} = $json->encode($hash{$item->{name}}) if $hash{$item->{name}};
     }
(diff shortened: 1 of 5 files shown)


neprune:
Changing property type identification to be based on refuse services for address is in progress...

perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
```diff
@@ -459,4 +463,14 @@ sub _bulky_send_reminder_email {
 
 sub bulky_send_before_payment { 0 }
 
+sub bulky_per_item_pricing_by_property_type { 0 }
+
+sub bulky_per_item_price_key {
+    my $self = shift;
+    return 'price' if !$self->bulky_per_item_pricing_by_property_type;
```
dracos, line 470: You could remove the need for `bulky_per_item_pricing_by_property_type` by having this check check `bulky_per_item_pricing_property_types` directly and seeing if it has any length or not.

dracos (changes requested):
Looking good, one bug and couple of suggestions is all :)

perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
```diff
@@ -463,11 +463,11 @@ sub _bulky_send_reminder_email {
 
 sub bulky_send_before_payment { 0 }
 
-sub bulky_per_item_pricing_by_property_type { 0 }
+sub bulky_per_item_pricing_property_types { [] }
 
 sub bulky_per_item_price_key {
     my $self = shift;
-    return 'price' if !$self->bulky_per_item_pricing_by_property_type;
+    return 'price' if !$self->bulky_per_item_pricing_property_types;
```
dracos, line 470: This will never match, as empty arrayrefs `[]` are not false-y, you'd need `if !@{$self->bulky_per_item_pricing_property_types};` to dereference the arrayref and then check its size.

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -135,6 +135,8 @@ sub view : Private {
     $c->stash->{template} = 'waste/bulky/summary.html';
 
 
+    $c->cobrand->bulky_set_reconstructed_property_data($p, $c->stash->{property});
```
dracos, line 138: Is this needed because summary is constructing the payment from the items in the summary template? That template doing that feels wrong to me if it's being used to view an existing booking - we shouldn't reconstruct the cost in case the payment items have changed in the interim, we should use the stored `payment` data on the record if it's present, as that was the actual amount.

I think if we do that, hopefully all of this commit isn't really needed because it won't need to know if it were domestic/trade.

dracos (changes requested):
One bug, and one hopefully making it better by changing the way the template is doing the price display, shout if I haven't made sense

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
neprune: Makes sense, my bad. Have still kept the part of the commit that does the domestic/trade identification based on the services rather than via property type. Will make the view collection page show the charged price as part of the payment work.

dracos (approved):
Thanks, all good :)

PR #4619, opened 2023-09-19
[Bromley] Add PC/PCSO to titles on reports

MorayMySoc:
Request to add PC/PCSO on Bexley titles.

https://mysocietysupport.freshdesk.com/a/tickets/3379

[skip changelog]


davea (changes requested):
Code looks fine, but commit message should be Bromley not Bexley :)

PR #4622, opened 2023-09-21
[Brent] Add Brent recycling sacks

MorayMySoc:
New recycling sack service

[skip changelog]


dracos (approved):
I would rename `sack-blue` as the other sacks in the same directory are the "refuse bag" type bags/sacks. So perhaps rename it `bag-blue`, for now anyway.

PR #4623, opened 2023-09-21
[Bexley] Add FMS id/street name area to submit email

MorayMySoc:
Add explicit FMS id for Bexley in notification email that a report has been made

https://mysocietysupport.freshdesk.com/a/tickets/3381

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -122,7 +122,7 @@ sub open311_extra_data_include {
     my ($self, $row, $h, $contact) = @_;
 
     my $open311_only;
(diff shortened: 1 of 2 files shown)


templates/email/default/submit.html
```diff
@@ -57,6 +57,14 @@ <h2 style="[% h2_style %]">[% report.title | html %]</h2>
       [%~ END %][% END %]
     </p>
   [% END %]
+
+  [% IF cobrand.moniker == 'bexley' %]
```
dracos, line 61: This template is never used by Bexley, UKCouncils.pm sets `fixmystreet.com` as an email template path above, so that one will always be used in preference.

dracos (approved):

PR #4625, opened 2023-09-22
[Northumberland] Add shortlisted to stats summary

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/3384

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -88,10 +88,12 @@ sub dashboard_export_problems_add_columns {
     $csv->add_csv_columns(
         staff_user => 'Staff User',
         staff_role => 'Staff Role',
+        assigned_to => 'Assigned To',
     );
 
     my $user_lookup = $self->csv_staff_users;
     my $userroles = $self->csv_staff_roles($user_lookup);
+    my $problems_to_user = $self->csv_active_planned_reports;
 
     $csv->csv_extra_data(sub {
         my $report = shift;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -99,14 +100,19 @@ sub dashboard_export_problems_add_columns {
         my $by = $report->get_extra_metadata('contributed_by');
         my $staff_user = '';
         my $staff_role = '';
+        my $assigned_to = '';
         if ($by) {
             $staff_user = $self->csv_staff_user_lookup($by, $user_lookup);
             $staff_role = join(',', @{$userroles->{$by} || []});
         }
+        if ($report->shortlisted_user) {
```
dracos, line 108: Any change to the CSV export needs to consider the effect on performance. In this case, you are asking for two new DB queries for every row being looked at (first the user planned reports table, then the user table, I can't see any prefetches). As with the `csv_staff_user_lookup` being used just above here, I think if this is a wanted column in the export, you will have to prefetch all the current shortlisted users into a hash, and then here you can then use that stored lookup.
(The other option would be to try and change the main DB query to include the shortlisted user in its results, but that seems more prune to issues.)

dracos (changes requested):
Code works fine, but I think will have to be changed to better deal with large exports.

chrismytton:
The code was only picking up the first problem assigned to a user. Have pushed a fixup that gets all of them.

chrismytton:
Found the `shortlisted_user` method on `Problem`, so have used that instead.

dracos:
@chrismytton As per my first review, you can't use shortlisted_user, that's what this PR used originally.

chrismytton:
@dracos OK, will try and figure out why the other way wasn't working.

PR #4633, opened 2023-09-27
[Bromley] [Waste] Bulky payments.

neprune:
https://github.com/mysociety/societyworks/issues/3891

### DIFF
--- a/templates/web/base/waste/bulky/summary.html
+++ b/templates/web/base/waste/bulky/summary.html
@@ -110,7 +110,11 @@
       END %]
       <dl>
         <dt>Price</dt>
-        [% payment = cobrand.bulky_total_cost(data) %]
+        [% IF problem %]
+            [% payment = problem.get_extra_field_value('payment') %]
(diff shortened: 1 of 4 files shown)


templates/web/base/waste/bulky/summary.html
```diff
@@ -110,7 +110,11 @@ <h3 class="summary-section--heading">Items to be collected</h3>
       END %]
       <dl>
         <dt>Price</dt>
-        [% payment = cobrand.bulky_total_cost(data) %]
+        [% IF problem.get_extra_field_value('payment') %]
```
dracos, line 113: I think I'd rather always call this if `problem` is set at all (as then payment always should be present), so as not to fall back onto something that might theoretically be an incorrect value. Unless there's quite possibly a reason to do this I've missed.

dracos (changes requested):
One suggestion is all, looks good :)

dracos (approved):

PR #4636, opened 2023-09-28
[Bromley] [Waste] Send bulky fields through to Echo.

neprune:
https://github.com/mysociety/societyworks/issues/3907


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1055,16 +1055,30 @@ sub waste_munge_bulky_data {
     my %items = map { $_->{name} => $_->{bartec_id} } @items_list;
```
dracos, line 1056: I think `collection_date` (not in this PR I realise), might need to be `Collection_Date`, think other Echo code assumes it's that (as does the test).

dracos (commented):
Code looks fine :) noticed something adjacent in passing

PR #4637, opened 2023-09-28
[Waste] Improve page loading display.

dracos:
On page load after clicking back, the form values can be filled in with their previous values by the browser, but this happens later, so switch to the pageshow event to update the autocomplete/messages and check the hidden select - as the autocomplete will have picked up an empty select at the time it was created. [skip changelog]

### DIFF
--- a/web/js/waste.js
+++ b/web/js/waste.js
@@ -69,15 +69,12 @@ $(function() {
 
 // Bulky waste
 
-$(function() {
-
-    var numItemsVisible = $('.bulky-item-wrapper:visible').length;
-    var maxNumItems = $('.bulky-item-wrapper').length;
-    var itemSelectionCounter = 0;
-    var firstItem = $('.bulky-item-wrapper').first();
+(function(){
+    var maxNumItems;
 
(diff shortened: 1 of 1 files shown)


MorayMySoc (changes requested):
So if this relates to https://3.basecamp.com/4020879/buckets/33314444/todos/6535974890 I get on this branch the same as on the kands-bulky-setup in that if I go back from the 'Location' page and I've added a image to my bulky waste, the image has gone.

dracos:
Sorry, thought I'd noted that, but clearly didn't!

The situation appears to be:
* Firefox - throws up a POST warning or two when you click Back, but if you click through that it does remember the dropdowns and the images;
* Safari - remembers the dropdowns (but has them open) and the images;
* Chrome - remembers the dropdowns (but has them open), but does not remember the images.

It looks like Chrome doesn't remember changed (by JS) hidden fields at all when clicking Back, which I guess is fair enough, and I can't think of an easy workaround. This is better than it was, so going to see if it'll do.

PR #4638, opened 2023-09-28
Kands bulky send confirmation email after payment

MorayMySoc:
Kingston confirmation emails for bulky waste collection only send when payment is received so stop the email sending on report completion and add it into the payment callback.

[skip changelog]

### DIFF
--- a/t/app/controller/waste_kands_bulky.t
+++ b/t/app/controller/waste_kands_bulky.t
@@ -374,6 +374,7 @@ FixMyStreet::override_config {
             subtest 'Summary submission' => \&test_summary_submission;
         };
 
+        my $catch_email;
         subtest 'Payment page' => sub {
             my ( $token, $new_report, $report_id ) = get_report_from_redirect( $sent_params->{returnUrl} );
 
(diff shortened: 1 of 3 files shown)


t/app/controller/waste_kands_bulky.t
```diff
@@ -474,16 +479,17 @@ FixMyStreet::override_config {
...
         my $id = $report->id;
         subtest 'Email confirmation of booking' => sub {
+            $mech->clear_emails_ok;
             FixMyStreet::Script::Reports::send();
             my @emails = $mech->get_email;
-            is $emails[1]->header('Subject'), "Bulky waste collection service - reference RBK-$id";
-            my $confirmation_email_txt = $mech->get_text_body_from_email($emails[1]);
-            my $confirmation_email_html = $mech->get_html_body_from_email($emails[1]);
-            like $confirmation_email_txt, qr/Date booking made: 30 August/, 'Includes booking date';
+            is scalar @emails, 0, "No confirmation email sent as mail sent at payment success for Kingston";
```
dracos, line 488: I think this check is in the wrong place, I'm afraid, we need to check that when it sends before payment that an email isn't sent, we know it won't be sent now as the previous test where catch_email was set sent it then.

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -91,4 +91,21 @@ Kingston starts collections at 6:30am, and lets you cancel up until then.
...
 sub bulky_cancellation_cutoff_time { { hours => 6, minutes => 30 } }
 
+=head2 suppress_report_sent_email
+
+For Bulky Waste reports, we want to send the email after payment has been confirmed, so we
+suppress the email here.
+
+=cut
+
+sub suppress_report_sent_email {
```
dracos, line 101: This function isn't being used anywhere, so the email is still being sent at the normal time.

dracos, line 101: Needs to be for Sutton as well so will move to the role.

PR #4641, opened 2023-09-29
[Brent] Brent recycling sacks pdfs

MorayMySoc:
Uses domestic waste service and blue bag service to calculate pdf to use.

Configuration in mysociety_servers under same branch.

[skip changelog]


dracos:
For reference, removed the 267 refs, simplified the double regexes, and removed the template addition that wasn't in use.

PR #4643, opened 2023-10-03
Bromley bulky refunds

neprune:
https://github.com/mysociety/societyworks/issues/3904


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -993,21 +994,72 @@ sub report_form_extras {
...
+    my ($self, $collection) = @_;
+    my $d = $self->collection_date($collection);
+    my $t = $self->_bulky_cancellation_cutoff_date($d);
+    return $t->subtract( days => 1 ) <= DateTime->now;
+}
+
+sub bulky_refund_amount {
+    my ($self, $collection) = @_;
+    my $charged = $collection->get_extra_field_value('payment');
+    return ($charged - $self->bulky_minimum_charge) if $self->bulky_refund_would_be_partial($collection);
```
dracos, line 1059: I realise once minimum charge is in place this shouldn't be able to happen, but do we want to make sure this function can never return a negative amount, just in case? :)

dracos (commented):
One suggestion is all :)

PR #4647, opened 2023-10-04
[SLWP] Include day and year in date

MorayMySoc:
Request to expand date in emails for Kingston and Sutton

https://3.basecamp.com/4020879/buckets/33314444/todos/6613694708

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
+++ b/perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
@@ -376,6 +376,7 @@ sub _check_within_bulky_refund_window {
 
 sub bulky_nice_collection_date {
     my ($self, $report_or_date) = @_;
+
     my $dt = do {
         if (ref $report_or_date eq 'FixMyStreet::DB::Result::Problem') {
             $self->collection_date($report_or_date);
@@ -383,7 +384,8 @@ sub bulky_nice_collection_date {
             $self->_bulky_date_to_dt($report_or_date);
         }
     };
-    return $dt->strftime('%d %B');
+
+    return $dt->strftime('%A %d %B %Y');
 }
 
 sub bulky_nice_collection_time {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
```diff
@@ -375,15 +375,20 @@ sub _check_within_bulky_refund_window {
...
     my $dt = do {
         if (ref $report_or_date eq 'FixMyStreet::DB::Result::Problem') {
             $self->collection_date($report_or_date);
         } else {
             $self->_bulky_date_to_dt($report_or_date);
         }
     };
-    return $dt->strftime('%d %B');
+    if ($full_date) {
+        return $dt->strftime('%A %d %B %G');
```
dracos, line 388: Unless you really have some hankering for confusing year numbers when the ISO week number is in a different calendar year, I'd go with `%Y` ;)

templates/email/default/waste/_bulky_data.html
```diff
@@ -2,11 +2,21 @@
     [% SET sent_confirm_id_ref = report.id %]
 [% END %]
 
+[% IF cobrand.moniker == 'sutton' || cobrand.moniker == 'kingston' %]
```
dracos, line 5: Do we see a downside to including the day and year all the time? It might be nice to have the day, and I guess the year can't hurt, especially if it was a few weeks off maybe? I'd even perhaps be persuaded to have it the default everywhere, not just in the email - that'd be the summary page, the bin days page, and the confirmation page I guess, hmm. What do you think?

dracos (changes requested):

templates/email/default/waste/_bulky_data.html
MorayMySoc: I suppose the only very minor risk is that it might force a new line on a web page, though not really that much longer and don't see that should really be an issue. I'd certainly appreciate the day being there if I was reading the date.

dracos (approved):
🗓️

PR #4649, opened 2023-10-05
[SLWP] Amend intro text for pricing and cancellation

MorayMySoc:
Changes for https://3.basecamp.com/4020879/buckets/33314444/todos/6618280655

### DIFF
--- a/templates/web/base/waste/bulky/intro.html
+++ b/templates/web/base/waste/bulky/intro.html
@@ -2,6 +2,26 @@
 <ol>
     <li>Requesting a bulky waste collection usually takes approximately <strong>10 minutes</strong></li>
     <li>You can request up to <strong>[% c.cobrand.bulky_items_maximum | numwords %] items per collection</strong></li>
+    [% IF cobrand.moniker == 'kingston' || cobrand.moniker == 'sutton' %]
+    [%
+        USE pounds = format('%.2f');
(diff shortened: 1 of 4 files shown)


templates/web/base/waste/bulky/intro.html
```diff
@@ -2,6 +2,27 @@ <h3>Before you start your booking</h3>
 <ol>
     <li>Requesting a bulky waste collection usually takes approximately <strong>10 minutes</strong></li>
     <li>You can request up to <strong>[% c.cobrand.bulky_items_maximum | numwords %] items per collection</strong></li>
+    [% IF cobrand.moniker == 'kingston' || cobrand.moniker == 'sutton' %]
+    <li>The price you pay depends how many items you would like collected:
+        <ul class="bulky-intro-pricing">
+            [% IF cobrand.moniker == 'sutton' %]
+            <li class="bulky-intro-pricing">
+                Up to 4 items = £35.05
```
dracos, line 10: Can we read all these numbers from the database configuration so we don't have to update them? :)
(That would also fix that it's 3 items 35.05, 6 58.73)

dracos (changes requested):
Location looks fine, would like it to generate from the data though, thanks.

dracos (approved):

PR #4654, opened 2023-10-09
[Hackney] Add favicon

chrismytton:
Add favicon for the Hackney cobrand.

For https://mysocietysupport.freshdesk.com/a/tickets/2970


dracos (changes requested):
If instead of this you put https://hackney.gov.uk/icons-a94e822ed66a743ac1649752651b4d73/favicon.ico at web/cobrands/hackney/favicon.ico then that has more sizes as far as I can see and no need to change a template.

dracos (approved):
Thanks :)

PR #4655, opened 2023-10-09
[SLWP] Kands collection page changes

MorayMySoc:
Adds js for updating the amount of selected items and if price change is imminent.

Changes text for introduction box.

Fixes bug where total not being updated on page load.

Fixes final part of https://github.com/mysociety/societyworks/issues/3937
[skip changelog]

### DIFF
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -17,7 +17,16 @@
   <div class="govuk-warning-text__content">
     <span class="govuk-warning-text__assistive">Information</span>
     <p class="govuk-!-margin-bottom-1"><strong>Note</strong></p>
-    <p>If an item is not available for selection on this page, it is not eligible for collection.</p>
+    [% IF c.cobrand.moniker == 'sutton' || c.cobrand.moniker == 'kingston' %]
+      <p>Select the items that you need us to collect using the drop-down menu below.
+      You can book the collection of up to [% c.cobrand.wasteworks_config.items_per_collection_max | numwords %] items.</p>
+      <p>You can search for an item in the list by typing the first few letters.</p>
+      <p>You can also
+        add an optional note against each item which can help ensure we pick up the right items.
+      </p>
+    [% ELSE %]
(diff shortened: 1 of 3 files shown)


templates/web/base/waste/bulky/items.html
```diff
@@ -9,15 +9,33 @@
...
-
+[% IF cobrand.moniker == 'sutton' || cobrand.moniker == 'kingston' %]
+[% SET cfg = c.cobrand.wasteworks_config %]
+<div
+  data-base-price="[% cfg.base_price %]"
+  data-band1-price="[% cfg.band1_price %]"
+  data-band1-max="[% cfg.band1_max %]"
+  data-items-per-collection-max="[% cfg.items_per_collection_max %]"
+  id="band_pricing_data">
+</div>
```
dracos, line 20: This data is already included (though in a slightly different format) in the `$('#js-bulky-total').data('pricing')` so it might make sense to use the same source?

web/js/waste.js
```diff
@@ -125,6 +125,35 @@ $(function() {
         }
     }
 
+    function display_band_pricing() {
+        var band_pricing = $('#band_pricing_data');
+
+        if (!band_pricing) {
+            return;
+        } else {
```
dracos, line 133: Don't need all this in an else, if you've returned above, can all outdent one iyswim

web/js/waste.js
```diff
@@ -125,6 +125,35 @@ $(function() {
...
+            var base_price = band_pricing.data('base-price') / 100;
+            var base_max = band_pricing.data('items-per-collection-max');
+            var band1_max = band_pricing.data('band1-max');
+            var message = '';
+            var current_selection = $('.bulky-item-wrapper:visible').length;
+            if (current_selection && current_selection < band1_max) {
+                message = current_selection + ' ' + (current_selection === 1 ? 'item' : 'items') + ' selected - you can add up to ' +
```
dracos, line 140: The first half of these are all the same - construct that bit of message first?

dracos (changes requested):
Suggesting using the same source of data in the JS, and bit of JS simplifying maybe :)

PR #4656, opened 2023-10-09
[Waste] [Bromley]  Enable a minimum cost for bulky per-item charging.

neprune:
https://github.com/mysociety/societyworks/issues/3932


dracos (approved):
This all looks good :) Couple of things to note in terms of interaction with other ongoing work:
* 041309ad6b in Brent (and originally in K&S also) rejigs waste.js to improve its handling (includes adding the updateTotal I see you spotted here and Moray also spotted on K&S branch), and I think your pricing change should then be okay on top of that once that's merged which might be soon if Brent going live - you could try and include that in your branch here as well, or just wait;
* K&S is adding a JS updating help text to the items.html page as well, but above not below the price (of course), so I think those two should not need to worry about each other but might be some git conflicts, and perhaps we should try and somehow have them combine in some way in future.

PR #4657, opened 2023-10-09
[Brent] Small items other electrical note

dracos:
Uses most of code from K&S notes commit (might be a bit of work done to merge both as have rejigged slightly)
For https://3.basecamp.com/4020879/buckets/33769634/todos/6624115227
[skip changelog]


perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -59,6 +59,21 @@ sub item_list : Private {
...
+    my $notes_field = {
+        type => 'TextArea',
+        label => 'Item note (optional)',
+        maxlength => 100,
+        tags => { hint => 'Describe the item to help our crew pick up the right thing' },
+    };
+    if ($c->cobrand->moniker eq 'brent') {
+        $notes_field = {
+            type => 'Text',
+            label => 'If other small electrical, please specify',
```
MorayMySoc, line 72: I think this only happens on 'other' so not a big deal but could say please give a description of the 'other' item

web/js/waste.js
```diff
@@ -199,4 +199,17 @@ $(function() {
...
+            var $this = $(this);
+            var $notes = $this.closest('.bulky-item-wrapper').find('[id^="form-item_notes_"]');
+            if ($this.val() == 'Small electricals: Other item under 30x30x30 cm') {
+                $notes.show();
+            } else {
+                $notes.hide();
+            }
+        });
+    }
+
```
MorayMySoc, line 214: This isn't triggered on a page back/refresh

MorayMySoc (approved):
Couple of small points, but think the important stuff works

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
dracos: I worded it like that in case the JS doesn't load/work, and so it'd be displayed all the time, and couldn't face having the text update :)

web/js/waste.js
dracos: I think changing to call the function on every one at load, and then same on change works:
```
         if (fixmystreet.cobrand == 'brent') {
-            $('[id^="form-item_notes_"]').hide();
-            $('.govuk-select[name^="item_"]').change(function(e) {
+            function update_small_items_other_notes() {
                 var $this = $(this);
                 var $notes = $this.closest('.bulky-item-wrapper').find('[id^="form-item_notes_"]');
                 if ($this.val() == 'Small electricals: Other item under 30x30x30 cm') {
@@ -171,7 +170,9 @@ $(function() {
                 } else {
                     $notes.hide();
                 }
-            });
+            }
+            $('.govuk-select[name^="item_"]').change(update_small_items_other_notes);
+            $('.govuk-select[name^="item_"]').each(update_small_items_other_notes);
         }
```

PR #4658, opened 2023-10-09
[SLWP] Add reference to bulky waste to index page

MorayMySoc:
https://3.basecamp.com/4020879/buckets/33314444/todos/6627663646

[skip changelog]


dracos (approved):
Fine, but do we want to check if they want this or "Book a bulky waste collection"? (It's "Report a missed bin collection" above)

PR #4659, opened 2023-10-10
[SLWP] Make comments for bulky create/cancel from council

MorayMySoc:
Changes comments made to the comment user of the council rather than the user taking the create/cancel action

https://3.basecamp.com/4020879/buckets/33314444/todos/6619156888

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -310,10 +310,11 @@ sub confirm_subscription : Private {
     if ($already_confirmed) {
         my $payment = $p->get_extra_field_value('payment');
         $payment = sprintf( '%.2f', $payment / 100 );
-        $p->add_to_comments({
+        my $comment = $p->add_to_comments({
             text => "Payment confirmed, reference $reference, amount £$payment",
-            user => $p->user,
+            user => $c->cobrand->body->comment_user || $p->user,
         });
+        $p->cancel_update_alert($comment->id);
     }
 }
 
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -313,7 +313,7 @@ sub add_cancellation_report : Private {
     if ($c->cobrand->bulky_cancel_by_update) {
         $collection_report->add_to_comments({
             text => 'Booking cancelled by customer',
-            user => $collection_report->user,
+            user => $c->cobrand->body->comment_user || $collection_report->user,
             extra => { bulky_cancellation => 1 },
         });
     } else {
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -1358,8 +1358,8 @@ Takes a comment id as a mandatory argument and will suppress
 alerts being sent out for that comment when
 FixMyStreet::Script::Alerts::send_updates is run.
 
-Optionally adding user_id argument will restrict the suppression to only
-updates provided by the original problem reporter.
+Optionally adding user_id argument will restrict the suppression
+to only alerts for that user ID.
 
 =cut
 
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -1352,4 +1352,35 @@ has comment_count => (
     },
 );
 
+=item cancel_update_alert
+
+Takes a comment id as a mandatory argument and will suppress
+alerts being sent out for that comment when
+FixMyStreet::Script::Alerts::send_updates is run.
+
+Optionally adding user_id argument will restrict the suppression to only
+updates provided by the original problem reporter.
+
+=cut
+
+sub cancel_update_alert {
+    my ($self, $comment_id, $user_id) = @_;
+
+    my $cfg = {
+        alert_type => 'new_updates',
+        parameter  => $self->id,
+        confirmed  => 1,
+    };
+    $cfg->{user_id} = $user_id if $user_id;
+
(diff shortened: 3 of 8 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1352,4 +1352,29 @@ has comment_count => (
...
 
+=item cancel_update_alert
+
+Takes a comment id as single argument and adds to the AlertSent
+table so an alert is not sent when FixMyStreet::Script::Alerts::send_updates
+is run
+
+=cut
+
+sub cancel_update_alert {
```
dracos, line 1363: In this commit, could the parts of the code that could call this function be changed to call this function? :) Admin/Triage.pm, Buckinghamshire.pm, ArchiveOldEnquiries.pm. (I'd say also UpdatesBase.pm except that has an extra check to only find alerts by the report user, so perhaps not.) There's also a typo in the commit message.

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -311,11 +311,12 @@ sub add_cancellation_report : Private {
     $c->cobrand->call_hook( "waste_munge_bulky_cancellation_data", \%data );
 
     if ($c->cobrand->bulky_cancel_by_update) {
-        $collection_report->add_to_comments({
+        my $comment = $collection_report->add_to_comments({
             text => 'Booking cancelled by customer',
             user => $collection_report->user,
             extra => { bulky_cancellation => 1 },
         });
+        $collection_report->cancel_update_alert($comment->id);
```
dracos, line 319: Whilst we want to cancel the payment confirmed one (because they'll have got/get a "report logged" email for the same event when it was sent to Echo), I don't think we want to cancel this one, because we want the user to get an email about the cancellation. (It's possible we'll get an update from Echo and then send two emails, but if that happens in testing we can then get that down to one, would rather start with too many.)

t/app/controller/waste_kands_bulky.t
```diff
@@ -418,6 +419,8 @@ FixMyStreet::override_config {
             is $new_report->get_extra_metadata('scpReference'), '12345', 'correct scp reference on report';
 
             $mech->clear_emails_ok;
+            FixMyStreet::Script::Alerts::send_updates();
+            is $mech->email_count_is(0), 1, 'Update cancelled as would go to body user not report user';
```
dracos, line 423: The update /would/ go to the report user, it's cancelled because we don't want to send it to them in this case.

dracos (changes requested):
Think we only want one of the cancellations, and change other code to use your nice new factored function :)

t/app/controller/waste_kands_bulky.t
```diff
@@ -529,7 +532,8 @@ FixMyStreet::override_config {
             # to Echo
             $mech->content_lacks('/waste/12345/bulky/cancel');
             $mech->content_lacks('Cancel this booking');
-
+            FixMyStreet::Script::Alerts::send_updates();
+            is $mech->email_count_is(0), 1, 'Update cancelled as would go to body user not report user';
```
MorayMySoc, line 536: So I thought this test was checking that the cancel update wasn't being sent, but the test still works even removing the suppress the cancel update - removing it as pointless.

t/app/controller/waste_kands_bulky.t
MorayMySoc: Actually this test is unaffected by whether the update is cancelled or not - when I implemented I was pretty sure it failed when I didn't have the cancel_alert implemented. Not sure where I've got confused

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -310,10 +310,11 @@ sub confirm_subscription : Private {
     if ($already_confirmed) {
         my $payment = $p->get_extra_field_value('payment');
         $payment = sprintf( '%.2f', $payment / 100 );
-        $p->add_to_comments({
+        my $comment = $p->add_to_comments({
             text => "Payment confirmed, reference $reference, amount £$payment",
             user => $p->user,
         });
+        $p->cancel_update_alert($comment->id);
```
MorayMySoc, line 317: I've left this in, but can't provoke the tests to send an update whether it's here or not - as mentioned in other comment, I was pretty sure I was able to get the tests to pass or fail dependent on this being here, but am a bit stumped.

perllib/FixMyStreet/DB/Result/Problem.pm
MorayMySoc: Done a bonus fixup that I think accounts for UpdatesBase, but may be overcomplicating things

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1354,20 +1354,26 @@ has comment_count => (
...
 
-Takes a comment id as single argument and adds to the AlertSent
-table so an alert is not sent when FixMyStreet::Script::Alerts::send_updates
-is run
+Takes a comment id as a mandatory argument and will suppress
+alerts being sent out for that comment when
+FixMyStreet::Script::Alerts::send_updates is run.
+
+Optionally adding user_id argument will restrict the suppression to only
+updates provided by the original problem reporter.
```
dracos: not updates by that user, alerts for that user.

t/app/controller/waste_kands_bulky.t
dracos: You're running this too early - the update isn't created until the call to pay_complete a few further lines down.
And then you haven't actually changed which user is creating the comment to the body user, so the alert wouldn't be sent as is same user as problem.

dracos (changes requested):
Have pushed fixup commits, hope that makes sense, ask if not!

PR #4663, opened 2023-10-11
[fms.com][Gloucestershire] Show correct message for logged-out users trying to leave an update

davea:
If a logged-in user can only leave updates on their reports if they’re closed, don’t incorrectly invite them to log in on open reports.

[skip changelog]


dracos (changes requested):
Commit message should end "open reports" I think? I understood the logic in the end, but would be good to have a test checking it's showing the right message :)

davea:
Thanks, have added a test. Will reword commit before merging.

dracos (approved):

PR #4665, opened 2023-10-11
[Glos, Oxfordshire] Stop job/defect popup disappearing immediately on mobile

davea:
Arguably these two files should be factored together somehow, but I didn't think it worth it as we'll be getting rid of the Gloucestershire one soon anyway.

For https://3.basecamp.com/4020879/buckets/32187059/todos/6618416579

[skip changelog]


nephila-nacrea (approved):
Cool - so the issue was related to the presence of the crosshairs?

PR #4666, opened 2023-10-12
[Bromley] Crystal palace triaging

MorayMySoc:
Adds tests for Crystal Park Triaging - map layer adds stopper to redirect user to different organisation.

Adds Bromley functions for road fount/not_found

Configuration in same branch name on mysociety-servers

https://github.com/mysociety/societyworks/issues/3885
[skip changelog]


web/cobrands/fixmystreet/assets.js
```diff
@@ -1681,7 +1681,7 @@ fixmystreet.message_controller = (function() {
             } else if (fixmystreet.assets.selectedFeature()) {
                 fixmystreet.body_overrides.allow_send(layer.fixmystreet.body);
                 responsibility_off(layer, 'road');
-            } else if ( (criterion && criterion()) || is_only_body(layer.fixmystreet.body) ) {
+            } else if ( ( typeof criterion === 'function' && criterion()) || is_only_body(layer.fixmystreet.body) ) {
```
dracos, line 1684: I see what you're doing with the workaround, but I think it actually means it doesn't work, so perhaps we should leave it as is and deal with when it errors when given something else (which is because you're passing `fixmystreet.message_controller.road_not_found` to `road_found` which passes in selected_feature as the second argument). Brent has its own function for this scenario which then uses a function that always returns 1, and I think you need the same here - because if you had TfL present in your test, I think this wouldn't pass because is_only_body would be false (because of TfL being there) so the responsibility would not switch on. And you always want it to switch on, if I've understood it right.

dracos (changes requested):

dracos (approved):
Thanks, perhaps we can think of a tidier way to do it at some point!

MorayMySoc:
Merged but forgot to push rebase

PR #4667, opened 2023-10-12
[Bromley] [Waste] Enable reporting missed bulky collections.

neprune:
https://github.com/mysociety/societyworks/issues/3908


dracos (approved):
Looks good, glad it did turn out to be this straightforward :) If you'd like, for Brent I ended up wrapping the bulky_check_missed_collection in a feature flag check (because they haven't got missed collections ready yet), I assume that might not be the case here but if we wanted consistent behaviour of that feature flag, could have the same perhaps, don't mind.

PR #4669, opened 2023-10-12
Switch to api.openstreetmap.org API host

danieldegroot2:
Use `api.openstreetmap.org/api/` -and HTTPS- instead of `www.openstreetmap.org/api/*`.

(Is: https://github.com/openstreetmap/operations/issues/951)

cc: @dracos


dracos:
Thanks - looks like only Norway/Sweden could possibly be using this code.

PR #4675, opened 2023-10-17
[Bristol] Show Bristol categories on admin

MorayMySoc:
Bristol categories were being filtered out in the dropdown on a report's admin page for changing category.

Broaden the check on munge_overlapping_bodies to check for more that the first area as report admin uses a larger pool of areas taken from the mapit areas for the report location

https://github.com/mysociety/societyworks/issues/3942

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bristol.pm
+++ b/perllib/FixMyStreet/Cobrand/Bristol.pm
@@ -188,15 +188,17 @@ sub munge_overlapping_asset_bodies {
     my ($self, $bodies) = @_;
 
     my $all_areas = $self->{c}->stash->{all_areas};
-    my $in_area = scalar(%$all_areas) == 1 && (values %$all_areas)[0]->{id} eq $self->council_area_id->[0];
-    return if $in_area;
 
-    # If we get here, assume we are outside Bristol boundaries.
-    if ($self->check_report_is_on_cobrand_asset) {
-        # The report is in a park that Bristol is responsible for, so only show Bristol categories.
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bristol.pm
```diff
@@ -188,15 +188,23 @@ sub munge_overlapping_asset_bodies {
...
 
     my $all_areas = $self->{c}->stash->{all_areas};
-    my $in_area = scalar(%$all_areas) == 1 && (values %$all_areas)[0]->{id} eq $self->council_area_id->[0];
-    return if $in_area;
 
-    # If we get here, assume we are outside Bristol boundaries.
-    if ($self->check_report_is_on_cobrand_asset) {
-        # The report is in a park that Bristol is responsible for, so only show Bristol categories.
+    if (grep ($self->council_area_id->[0] == $_, keys %$all_areas)) {
+        # We are in an area covered by Bristol so only show Bristol categories and any multi-area cateories.
```
dracos, line 193: We're not in an area covered by Bristol here, we are within the area of Bristol here. So I think if that's the case you can simply `return`, and not change anything at this point - unlike Brent/Camden, it's one way, Bristol covers two surrounding areas, but nothing covers Bristol.

dracos (changes requested):
I think it's (in this case!) simpler, phew :)

dracos (commented):
I added a test which showed that there was a second location where the same check was made (in UKCouncils.pm) that also needed updating.

PR #4677, opened 2023-10-18
[Bromley] Don't show amend numbers

MorayMySoc:
Bromley not offering amending bulky collection so don't have line in summary saying how much more can be added

https://3.basecamp.com/4020879/buckets/33906758/todos/6657459835

[skip changelog]


neprune (approved):
Thanks

PR #4682, opened 2023-10-18
Bromley bulky text changes

neprune:


neprune:
> One optional comment :) I think I've matched these up against the basecamp tickets to check, am I right in that there are two things left, the summary-page-for-existing-booking showing an "add up to" line it shouldn't, and the text changes to the confirmation page?

The "add up to" thing has been done by Moray: https://github.com/mysociety/fixmystreet/pull/4677 and also the text changes in the confirmation email https://github.com/mysociety/fixmystreet/pull/4678/files. I think alongside this PR, that should cover everything.

dracos:
> also the text changes in the confirmation email

Not the confirmation email, but the confirmation web page, couldn't see that anywhere, sorry if I missed it.

> The "add up to" thing has been done by Moray: https://github.com/mysociety/fixmystreet/pull/4677 

Ah yes; that fixes it in terms of we never want to show it for Bromley, I was meaning also that if `problem` is set, we also don't want to show that line because you're looking at your booking then, not making one.

neprune:
@dracos I think the confirmation page changes they want is covered by https://github.com/mysociety/fixmystreet/pull/4682/commits/c75d030f1ef1893c53768af1976912e1fced4167 - unless I’m missing something.

dracos:
> I think the confirmation page changes they want is covered by https://github.com/mysociety/fixmystreet/commit/c75d030f1ef1893c53768af1976912e1fced4167 - unless I’m missing something.

Nope, that is it, isn't it, sorry!

PR #4683, opened 2023-10-19
[Peterborough] Add messages to problem page

MorayMySoc:
Green and black bins are no longer being replaced as Peterborough working on new system.

Adds messages instructing what to do in case of broken green or black bin

https://mysocietysupport.freshdesk.com/a/tickets/3459

[skip changelog]


dracos (approved):
Neat way of doing it :)

dracos:
From FD ticket, they no longer need this.

PR #4680, opened 2023-10-18
Bulky phone bug

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/3914.

[skip changelog]

PR #4684, opened 2023-10-20
Bulky phone bugs

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/3914.

Was originally https://github.com/mysociety/fixmystreet/pull/4680 but changes to the underlying kands-bulky-setup branch produced a lot of conflicts when I attempted to rebase.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1550,6 +1550,7 @@ sub add_report : Private {
     }
 
     $report->set_extra_metadata(property_address => $c->stash->{property}{address});
+    $report->set_extra_metadata(phone => $c->stash->{phone});
     $c->cobrand->call_hook('save_item_names_to_report' => $data);
     $report->update;
 
--- a/perllib/FixMyStreet/Cobrand/Peterborough/Bulky.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough/Bulky.pm
@@ -334,7 +334,7 @@ sub waste_reconstruct_bulky_data {
 
     $saved_data->{name} = $p->name;
     $saved_data->{email} = $p->user->email;
-    $saved_data->{phone} = $p->user->phone;
+    $saved_data->{phone} = $p->get_extra_metadata('phone') || $p->user->phone;
     $saved_data->{resident} = 'Yes';
 
     return $saved_data;
--- a/perllib/FixMyStreet/Roles/CobrandSLWP.pm
+++ b/perllib/FixMyStreet/Roles/CobrandSLWP.pm
@@ -1306,7 +1306,7 @@ sub waste_reconstruct_bulky_data {
 
     $saved_data->{name} = $p->name;
     $saved_data->{email} = $p->user->email;
-    $saved_data->{phone} = $p->user->phone;
+    $saved_data->{phone} = $p->get_extra_metadata('phone') || $p->user->phone;
 
     return $saved_data;
 }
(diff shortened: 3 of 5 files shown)


perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1008,6 +1009,8 @@ sub process_user : Private {
             $user->name($name) unless $user->name;
             $c->stash->{no_reporter_alert} = 1;
         } elsif ($c->stash->{contributing_as_another_user}) {
+            # TODO Is this block reachable?
```
dracos, line 1012: Yes, in the circumstance given by the comment below - if neither username nor phone are provided.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1038,7 +1041,10 @@ sub process_user : Private {
 
     my $parsed = FixMyStreet::SMS->parse_username($params{username});
     my $type = $parsed->{type} || 'email';
-    $type = 'email' unless $c->cobrand->sms_authentication || $c->stash->{contributing_as_another_user};
+    $type = 'email'
+        unless $c->cobrand->sms_authentication
+        || $c->stash->{contributing_as_another_user}
+        || $c->stash->{ignore_logged_in_user};
```
dracos, line 1047: Aha, I think this shows the underlying issue; what you raised as point 2 in the ticket only happens if the staff user does not have the "contribute as another user" permission (which it makes sense to demand they have in order to leave reports as other users, I think). I think if you give your staff user in your new test that permission, then the existing code seems to work fine as-is without the extra `ignore_logged_in_user` changes. A staff without that permission can't leave only a phone, but can leave a different email (because any normal user can fill in the form with an email).

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1723,6 +1729,10 @@ sub save_user_and_report : Private {
     } elsif ($c->stash->{contributing_as_anonymous_user}) {
         $report->set_extra_metadata( contributed_as => 'anonymous_user' );
         $report->confirm();
+    } elsif ($c->stash->{ignore_logged_in_user}) {
+        # Waste report on behalf of customer
+        $report->user->update_or_insert;
```
dracos, line 1734: We don't want to e.g. update a user account's name if the staff have given it a different name.

dracos (changes requested):
The fix for your point 1 (number not being saved) looks good :) After some investigation, I think the issue for your point 2 is down to the "contribute as another user" permission on the staff user not being present, and I think we can assume it always will be for staff doing this, so the changes here might not be needed (and guess this explains why it's never come up as an issue).

dracos:
kands-bulky-setup has changed again, but rebasing should be doable; you don't want to rebase the whole thing, as in eg `git rebase origin/kands-bulky-setup` or similar, because that will look for the shared base between where you are now and where you're rebasing to, and try and reapply all the commits on your branch since then, which will be repeats of all the commits now in the new kands-bulky-setup branch. (See end for an example.)

I've rebased `bulky-phone-bug-rebase` with the following command: `git rebase d01774f5ba --onto origin/kands-bulky-setup` (where d01774f5ba was the last commit in your branch before you added new code for this topic) and there was only one conflict in the test file because both branches added tests, so kept them both.

Example
-------

`A - B - C` - master
`A - D - E - F` - kands-bulky-setup before you started work
`A - D - E - F - G - H` - bulky-phone-bug-rebase with two commits
`A - B - C - D' - E' - F'` - I rebase kands-bulky-setup on top of current master

Now if you were on bulky-phone-bug-rebase and did `git rebase origin/kands-bulky-setup` git would go "What commits are present on bulky-phone-bug-rebase that aren't on kands-bulky-setup?" It doesn't know anything clever. And so the answer would be `D - E - F - G - H`. It would then switch to `F'` (kands-bulky-setup) and try and apply D then E then F (even though rebased versions of those are already all present), and you'd probably get a lot of conflicts.

Instead you want do do `git rebase F --onto origin/kands-bulky-setup` - this would then apply G and H only on to F' and you'd end up with `A - B - C - D' - E' - F' - G' - H'`.

PR #4685, opened 2023-10-23
[SLWP] Adds cheque payment for K&S

MorayMySoc:
Adds the option to accept cheque payments and put in the cheque reference number avoiding the credit card payment flow.
    
https://github.com/mysociety/societyworks/issues/3952

[Skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -311,8 +311,14 @@ sub confirm_subscription : Private {
     if ($already_confirmed) {
         my $payment = $p->get_extra_field_value('payment');
         $payment = sprintf( '%.2f', $payment / 100 );
+        my $reference_text = 'reference ';
+        if (!$reference) {
+            $reference_text .= $p->get_extra_metadata('chequeReference') . ' (phone/cheque)';
+        } else {
+            $reference_text .= $reference;
+        }
         my $comment = $p->add_to_comments({
-            text => "Payment confirmed, reference $reference, amount £$payment",
+            text => "Payment confirmed, $reference_text, amount £$payment",
             user => $c->cobrand->body->comment_user || $p->user,
         });
         $p->cancel_update_alert($comment->id);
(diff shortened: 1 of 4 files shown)


MorayMySoc:
Wanted to check this is broadly what's expected. I think there may be an issue in that, if you go to pay by CC changing mind and back button then going forward with cheque doesn't work. At least summary page doesn't show - also text will probably need to change from 'Continue to payment' to something like 'Make booking' I guess.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -311,8 +311,14 @@ sub confirm_subscription : Private {
     if ($already_confirmed) {
         my $payment = $p->get_extra_field_value('payment');
         $payment = sprintf( '%.2f', $payment / 100 );
+        my $reference_text;
+        if (!$reference) {
+            $reference_text = 'paid by phone/cheque';
```
dracos, line 316: Could this be "reference $chequeReference (phone/cheque)" (looking up the ref) so that it matches up at the `open311-adapter` end.

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
```diff
@@ -70,6 +70,15 @@ has_page summary => (
             tandc => { option_label => $label }
         };
     },
+    field_ignore_list => sub {
+        my $page = shift;
+        my $c = $page->form->c;
+
+        if (!($c->cobrand->moniker eq 'sutton' || $c->cobrand->moniker eq 'kingston') || !$c->stash->{is_staff}) {
+            $page->form->{data}->{payment_method} = 'Credit card';
```
dracos, line 78: What does this line do? `payment_method` would be set to `credit_card` I would assume, anyway, so I'm hoping it's not needed.

dracos (changes requested):
Looks good, couple of small things

dracos (approved):
Ta!

PR #4686, opened 2023-10-23
Make sure disable flag for old app is string "1".

dracos:
The old app has a === "1" comparison in it to check for a question answer disabling, so make sure we output a string.
[skip changelog]


davea (approved):
Whoops! Can't remember why we didn't serialise it as `true` at the time, oh well.

PR #4689, opened 2023-10-23
[SLWP] Send confirmation email after confirmation

MorayMySoc:
So staff making a paye payment on behalf of a user will still initiate a confirmation of booking

https://github.com/mysociety/societyworks/issues/3957

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -255,11 +255,6 @@ sub pay_complete : Path('pay_complete') : Args(2) {
         $c->stash->{reference} = $ref;
         $c->stash->{action} = $p->title eq 'Garden Subscription - Amend' ? 'add_containers' : 'new_subscription';
         $c->forward( 'confirm_subscription', [ $ref ] );
-        if ($c->cobrand->suppress_report_sent_email($p)) {
-            # Send bulky confirmation email after payment confirmation (see
-            # the suppress_report_sent_email in CobrandSLWP.pm)
-            $p->send_logged_email({ %{$c->stash} }, 0, $c->cobrand);
-        }
     } else {
         $c->stash->{template} = 'waste/pay_error.html';
         $c->detach;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -287,6 +282,12 @@ sub confirm_subscription : Private {
 
     return unless $p->state eq 'unconfirmed' || $already_confirmed;
 
+    if (($p->category eq 'Bulky collection' || $p->category eq 'Small items collection') && $c->cobrand->suppress_report_sent_email($p)) {
```
dracos, line 285: Is there a reason to limit this to bulky? Could conceivably see this happening with garden, and when it was in its old location it didn't have this check. A test for it being sent for staff would be good too.

dracos (changes requested):

t/app/controller/waste_kands_bulky.t
```diff
@@ -695,6 +657,138 @@ FixMyStreet::override_config {
...
+                    'item_photo_1' => [ $sample_file, undef, Content_Type => 'image/jpeg' ],
+                    'item_2' => 'Bicycle',
+                    'item_3' => 'Bath',
+                },
+            },
+        );
+        $mech->submit_form_ok({ with_fields => { location => 'in the middle of the drive' } });
+        $mech->submit_form_ok({ with_fields => { tandc => 1 } });
+        $mech->submit_form_ok({ with_fields => { payenet_code => '54321' } });
+        FixMyStreet::Script::Alerts::send_updates();
```
dracos, line 785: Are we running alerts here to check no other email sent? The one email being sent/checked for is being sent by the payenet_code submission the line above.

dracos (approved):
Think this is fine, one query about calling the alert script (I guess if a comment user was set for Sutton, an alert email would be being sent, thinking about it?).

t/app/controller/waste_kands_bulky.t
MorayMySoc: I didn't twig that was being sent anyway. As Sutton doesn't currently have a comment user, should I remove that - or should I leave it in anyway?

dracos: Probably clearer to remove it I guess.

PR #4692, opened 2023-10-24
[Brent] Small item missed collection tweaks

dracos:
Couple of pre-go-live tweaks from testing [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandEcho.pm
+++ b/perllib/FixMyStreet/Roles/CobrandEcho.pm
@@ -614,12 +614,13 @@ sub bulky_check_missed_collection {
 
     my $cfg = $self->feature('echo');
     my $service_id = $cfg->{bulky_service_id} or return;
+    my $service_id_missed = $cfg->{bulky_service_id_missed};
     my $event_type_id = $cfg->{bulky_event_type_id} or return;
     my $event = $events->{enquiry}{$event_type_id};
     return unless $event;
     my $row = {
         service_name => 'Bulky waste',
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -614,12 +614,13 @@ sub bulky_check_missed_collection {
...
     my $cfg = $self->feature('echo');
     my $service_id = $cfg->{bulky_service_id} or return;
+    my $service_id_missed = $cfg->{bulky_service_id_missed};
     my $event_type_id = $cfg->{bulky_event_type_id} or return;
     my $event = $events->{enquiry}{$event_type_id};
     return unless $event;
     my $row = {
         service_name => 'Bulky waste',
-        service_id => $service_id,
+        service_id => $service_id_missed || $service_id,
```
MorayMySoc, line 623: I'm not clear on this - if ```$service_id_missed``` is set from config it's always going to override ```$service_id```. So some cobrands will just be using the service_id and some will have a specific service_id_missed?

dracos: `$service_id` is still used below when it's looking for `$missed_events`, so the override is only here, for the reporting of missed collections - and for Brent we want to use a different service ID for that, so that `open311-adapter` can spot that, tick some different boxes, and then it sets it back to the original service ID (in its config it has `service_mapping: 787: 274`.

MorayMySoc (approved):
Looks good

PR #4696, opened 2023-10-27
K&S bulky reminder emails: fixes

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/3953.

- [ ] Fix bug with cancellation link / redirect when user not logged in
- [ ] Remove 'You can still cancel your booking, but a refund is no longer available.'

[skip changelog]


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -587,6 +587,16 @@ sub csc_payment_failed : Path('csc_payment_failed') : Args(0) {
 sub property : Chained('/') : PathPart('waste') : CaptureArgs(1) {
     my ($self, $c, $id) = @_;
 
+    # Some pages chained off /waste/property require user to be logged in, so
+    # perform /auth redirect before we start looking up the property.
+    # These pages include:
+    # /waste/<property_id>/bulky/cancel/<report_id>
+    if ( $c->req->path =~ /\/bulky\/cancel/
```
dracos, line 594: Think this is okay place to put it! Would add to the comment that it's the asynchronous property lookup that makes this necessary as the redirect to auth then doesn't work after that.
And then this doesn't catch the small items path, so might be better to have `$c->action eq 'waste/bulky/cancel' || $c->action eq 'waste/bulky/cancel_small'` (doesn't need to worry about the URLs then either) (I think those are the right actions,but needs checking)

dracos (changes requested):
Looks good, suggestion on using `action` and adding the small items path

MorayMySoc (approved):
Looks good to me

PR #4700, opened 2023-11-01
[FMS] Add parent categories to dashboard categories view

MorayMySoc:
Add a parent categories column to show if a category is a subcategory.

Arbitrary decision to choose the first parent category from the groups method of a contact.

https://github.com/mysociety/societyworks/issues/1458

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -127,7 +127,10 @@ sub index : Path : Args(0) {
         $c->forward('/report/stash_category_groups', [ $c->stash->{contacts} ]);
 
         # See if we've had anything from the body dropdowns
-        $c->stash->{category} = $c->get_param('category');
+        $c->stash->{category} = [ $c->get_param_list('category') ];
+        my %display_categories = map { $_ => 1 } @{$c->stash->{category}};
+        $c->stash->{display_categories} = \%display_categories;
+
         $c->stash->{ward} = [ $c->get_param_list('ward') ];
 
         if ($c->user_exists) {
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -126,10 +126,24 @@ sub index : Path : Args(0) {
         $c->stash->{contacts} = [ $c->stash->{contacts}->all ];
         $c->forward('/report/stash_category_groups', [ $c->stash->{contacts} ]);
 
+        my %group_names = map { $_->{name} => $_->{categories} } @{$c->stash->{category_groups}};
         # See if we've had anything from the body dropdowns
         $c->stash->{category} = [ $c->get_param_list('category') ];
+        my @remove_from_display;
+
+        foreach (@{$c->stash->{category}}) {
+            next unless /^group-(.*)/;
+            for my $contact (@{$group_names{$1}}) {
+                push @{ $c->stash->{category} }, $contact->category;
+                push @remove_from_display, $contact->category;
+            }
+        }
+
         my %display_categories = map { $_ => 1 } @{$c->stash->{category}};
+        delete $display_categories{$_} for (@remove_from_display);
         $c->stash->{display_categories} = \%display_categories;
+
+        @{$c->stash->{category}} = grep { $_ !~ /^group-/} @{$c->stash->{category}};
 
         $c->stash->{ward} = [ $c->get_param_list('ward') ];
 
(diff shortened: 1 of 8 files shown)


MorayMySoc:
I'll need to fix and add tests, but I wanted to check that we were agreed is what's meant to be achieved

MorayMySoc:
Ok, ultimately copied the wards stuff once I understood what was going on - but as far as I see it needs to have the labels selectable in some way (either by click or input check) to select the whole group. Gone in a bit of a long loop on this, so that's actually probably what the real work of this ticket was!

MorayMySoc:
Changes the dropdown for categories in the dashboard stats to be multi-select checkboxes instead of only allowing selecting one category.

Adds group labels to group categories in the 'categories' dropdown menu of the dashboard with an all-*group-name* checkbox to select all of subcategories for that group.

https://github.com/mysociety/societyworks/issues/1458

templates/web/base/report/new/_category_select.html
```diff
@@ -1,5 +1,8 @@
 [%~ FOREACH group IN category_groups ~%]
-    [% IF group.name %]<optgroup label="[% group.name %]">[% END %]
+    [% IF group.name %]<optgroup label="[% group.name %]">
+        [% SET pseudo_category = "group-" _ group.name %]
+        <option value="group-[% group.name %]" [% ' selected' IF display_categories.$pseudo_category %]>All [% group.name %]</option>
```
dracos, line 4: This template is included by the admin report edit category dropdown, and the old-app category select, but you only want this change to apply to the dashboard usage - pass in an `include_group_option` in the `INCLUDE` and use that?

perllib/FixMyStreet/Reporting.pm
```diff
@@ -18,7 +18,7 @@ has on_updates => ( is => 'lazy', default => sub { $_[0]->type eq 'updates' } );
 
 has body => ( is => 'ro', isa => Maybe[InstanceOf['FixMyStreet::DB::Result::Body']] );
 has wards => ( is => 'ro', isa => ArrayRef[Int], default => sub { [] } );
-has category => ( is => 'ro', isa => Maybe[Str] );
+has category => ( is => 'ro', isa => Maybe[ArrayRef[Maybe[Str]]], default => sub { [] } );
```
dracos, line 21: Feels like we could get rid of these Maybes, the code changes in this file could be easier, e.g. if we coerced the category to always be an ArrayRef[Str], then you wouldn't need to check $self->category is set anywhere, it always would be an arrayref.

perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -126,8 +126,32 @@ sub index : Path : Args(0) {
         $c->stash->{contacts} = [ $c->stash->{contacts}->all ];
         $c->forward('/report/stash_category_groups', [ $c->stash->{contacts} ]);
 
+        my %group_names = map { $_->{name} => 1 } @{$c->stash->{category_groups}};
+
         # See if we've had anything from the body dropdowns
-        $c->stash->{category} = $c->get_param('category');
+        $c->stash->{category} = [ $c->get_param_list('category') ];
+        my @remove_from_display;
+        if (grep { $_ =~ /^group-/ } @{ $c->stash->{category} }) {
```
dracos, line 134: Unless I've missed an oddity in what this is trying to do, would something like this construction have the same effect as what you're after?

```
my %group_names = map { $_->{name} => $_->{categories} } @{$c->stash->{category_groups}};
foreach (@{$c->stash->{category}}) {
    next unless /^group-(.*)/;
    for my $contact (@{$group_names{$1}}) {
        push @{ $c->stash->{category} }, $contact->category;
        push @remove_from_display, $contact->category;
    }
}
```

dracos (changes requested):
Looking better, one issue with the template change, and hopeful couple of simplifications :)

perllib/FixMyStreet/Script/CSVExport.pm
```diff
@@ -108,6 +108,7 @@ sub process_body {
         body => $body,
         user => $body->comment_user,
         dbi => 1, # Flag we are doing it via DBI
+        category => [],
```
dracos, line 111: I don't think this is needed, because of your `default`?

dracos (changes requested):
Looking good, just one thing more for me checking I've understood it right :)

dracos (approved):
Thanks! All seems to work as expected :) I actually just spotted in TfL you also don't need to check the first bit of that condition now as you know it will be an arrayref, but it's fine as is as well.

MorayMySoc:
Sorry to force a re-review but figured I may as well get that done too - but wanted to check I haven't misunderstood and got it wrong either.

dracos (approved):
No worries, yes, that's what I meant - think `->category` is guaranteed to be an arrayref now, which makes things bit neater :)

PR #4701, opened 2023-11-01
Do not re-enable report form for non-NH categories

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/3936.

The `regenerate_category()` function in web/cobrands/highways/assets.js was indiscriminately (re-)enabling the report form after category selection. So if user selected an NH road, selected the 'Somewhere else' option from the popup, then selected a category that should disable the form, the form would not be disabled.

[skip changelog]


davea (approved):
Good sleuthing! Looks good 👍

PR #4702, opened 2023-11-02
[SLWP] Bulky can only be offered if service initiated

MorayMySoc:
Request that bulky not be offered if the service hasn't been initialised which may be the case on new builds

https://3.basecamp.com/4020879/buckets/33314444/todos/6713679040


davea (approved):
👍

PR #4704, opened 2023-11-03
[Thamesmead][Bexley][Greenwich] Allow council reporting on Thamesmead cobrand

MorayMySoc:
If a report is made on the Thamesmead cobrand but not on a Thamesmead asset the contacts for the appropriate council will be displayed and a report can be made for that council.

https://github.com/mysociety/societyworks/issues/3571

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Thamesmead.pm
+++ b/perllib/FixMyStreet/Cobrand/Thamesmead.pm
@@ -43,7 +43,8 @@ sub updates_restriction { FixMyStreet::Cobrand::UKCouncils::updates_restriction(
 sub site_key { FixMyStreet::Cobrand::UKCouncils::site_key($_[0], $_[1]) }
 sub all_reports_single_body { FixMyStreet::Cobrand::UKCouncils::all_reports_single_body($_[0], $_[1]) }
 sub suggest_duplicates { FixMyStreet::Cobrand::UKCouncils::suggest_duplicates($_[0]) }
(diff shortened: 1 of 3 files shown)


davea (approved):

perllib/FixMyStreet/Cobrand/Thamesmead.pm
```diff
@@ -65,9 +66,17 @@ sub example_places {
 sub munge_report_new_bodies {
     my ($self, $bodies) = @_;
 
-    %$bodies = map { $_->id => $_ } grep { $_->name eq 'Thamesmead' } values %$bodies;
+    for (FixMyStreet::Cobrand::Bexley->new->body, FixMyStreet::Cobrand::Greenwich->new->body) {
+        if ($_ && $_->{id}) {
+            $bodies->{$_->id} = $_;
+        };
+    }
```
dracos, line 73: This for loop is doing nothing, because `$_->{id}` never exists (it's `$_->id` as in the next line). So %bodies is left alone (which I guess is probably the desired outcome, as checks like this loop would already have happened in fetching the bodies they'd have to exist, I guess TfL are the only potential to consider, but presumably good to have them too? If it's working :) ).

MorayMySoc: https://github.com/mysociety/fixmystreet/pull/4733

PR #4706, opened 2023-11-06
[Brent] Allow different OIDC config for WasteWorks host

davea:
Allows Brent's `oidc_login` `COBRAND_FEATURE` to have a different set of values for whatever hostname is used for the request.

For example, to use a different config for `brent.wasteworks.example.org`:

```yaml
COBRAND_FEATURES:
  oidc_login:
    brent:
      client_id: 'fms_client_id'
      secret: 'fms_secret'
      display_name: 'Brent FMS Login'
      auth_uri: 'https://oidc.example.org/oauth2/v2.0/authorize'
      token_uri: 'https://oidc.example.org/oauth2/v2.0/token'
      hosts:
        brent.wasteworks.example.org:
          client_id: 'ww_client_id'
          secret: 'ww_secret'
          display_name: 'Brent WasteWorks Login'
          auth_uri: 'https://oidc.example.org/oauth2/v2.0/authorize'
          token_uri: 'https://oidc.example.org/oauth2/v2.0/token'

```

For FD-3395

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth/Social.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth/Social.pm
@@ -162,10 +162,16 @@ sub twitter_callback: Path('/auth/Twitter') : Args(0) {
     $c->forward('oauth_success', [ 'twitter', $info->{id}, $info->{name} ]);
 }
 
+sub oidc_config : Private {
+    my ($self, $c) = @_;
+
+    return $c->cobrand->call_hook('oidc_config') || $c->cobrand->feature('oidc_login');
+}
+
 sub oidc : Private {
     my ($self, $c) = @_;
 
-    my $config = $c->cobrand->feature('oidc_login');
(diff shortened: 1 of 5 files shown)


dracos (approved):
Looks good - I think you could have the code that's in `Brent.pm` in `Auth/Social.pm`, working (potentially) for anyone - might come in handy in future and wouldn't affect any existing behaviour?

davea:
That's a good idea - there's nothing Brent-specific about this change is there. However I'm not sure if the `auth/general.html` template could access the config then? Maybe `oidc_config` could go in the `Default` cobrand?

dracos:
> However I'm not sure if the auth/general.html template could access the config then? Maybe oidc_config could go in the Default cobrand?

I think I'd probably have the function in Auth/Social.pm set a stash `oidc_config` then the template doesn't have to fetch anything/worry about it.

PR #4710, opened 2023-11-09
[Brent] Add resolution codes for missed bulky waste

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/3499

[skip changelog]


perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -626,7 +626,7 @@ sub bulky_check_missed_collection {
     foreach my $state_id (keys %$blocked_codes) {
         next unless $event->{state} eq $state_id;
         foreach (keys %{$blocked_codes->{$state_id}}) {
-            if ($event->{resolution} eq $_ || $_ eq 'all') {
+            if ($event->{resolution} eq $_) {
```
dracos, line 629: Why has this been removed? `all` is a fallback for if it exists

dracos (changes requested):
These text lines are displayed to the public? On their bin day page by the small items collection. So I don't think we can use them like this.

PR #4714, opened 2023-11-09
[Open311] Lock report row whilst processing update.

dracos:
This seemed the smallest change that might work [skip changelog]


neprune (approved):

neprune (approved):
Nice. Will let you know if I think of a better name...

PR #4715, opened 2023-11-10
[Southwark] Only display dead animal location question when not council tenant.

neprune:
[skip changelog] [FD-3522](https://mysocietysupport.freshdesk.com/a/tickets/3522)

Works in combination with set-up for a required "Are you a council tenant?" question and an optional "Location of dead animal" question with the stopper and redirect - this set-up is configured on staging https://staging.fixmystreet.com/report/new?longitude=-0.101650&latitude=51.504550#extra.

An issue with this approach is that if the user erroneously selects 'no' for council tenant and 'other' for location, and then corrects their answer for council tenant, they will still be blocked and the location field they need to change to unblock will disappear.

### DIFF
--- a/web/cobrands/fixmystreet.com/base.scss
+++ b/web/cobrands/fixmystreet.com/base.scss
@@ -509,3 +509,17 @@ body.twothirdswidthpage .content {
         margin-bottom: 0;
     }
 }
+
+/* New report conditional questions */
+
+@supports( #{"selector(:has(*))"} ) {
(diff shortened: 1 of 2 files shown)


dracos (commented):
The code looks fine, I think the staging set up might be wrong - if I read it right, it should be disabling the form on front/back/commercial rather than the other way round? Also, we shouldn't use "here" as link text (e.g. for if links are summarised), perhaps something like "Please [make a dead animal removal request here]".

PR #4716, opened 2023-11-13
Pre-compute dashboard problem CSV export

dracos:
This tries to act in an entirely backwards-compatible manner.

It adds a `bin/fixmystreet.com/all-csv-export` script that uses DBI to extract all data for CSV export, along with any per-cobrand per-row extra data required. Once those files exist, they are used by the csv-export and Dashboard controller (and thus data returned will stop at the time those files were generated) to filter the output to the relevant rows with no database access required. This speeds up CSV export an awful lot.

All the cobrands are updated to cope with being given either a Problem object or a hashref,

[skip changelog]


MorayMySoc (approved):
Ok, think this is all good

PR #4721, opened 2023-11-15
[Bromley] [Waste] No refunds for unpaid bulky bookings.

neprune:
[skip changelog] Stops a refund email being sent when a user cancels a booking that hasn't actually been paid for.


dracos (commented):
Would be good to have a test of it :)

dracos:
Put a small test in for the front end.

PR #4724, opened 2023-11-17
[Waste] Add collection details to cancellation page.

dracos:
Requested by K&S for theirs, but I think given you could have multiple collections, this is a nice thing to have everywhere. [skip changelog]


MorayMySoc (approved):
This looks good to me

PR #4728, opened 2023-11-20
GOVUK input for filters in around page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3821
Fixes: https://github.com/mysociety/societyworks/issues/4214
Fixes: https://github.com/mysociety/societyworks/issues/4215

IMPORTANT: This update breaks Camden, therefore I think we should tackle [this issue](https://github.com/mysociety/societyworks/issues/3874) before merging this PR. The Alternative I could add a commit here fixing the issue that Camden currently has, which is the `line-height` property, and leave the tidy up ticket in his original order.

### Mobile
Screenshot 2024-05-01 at 07 23 55

### Desktop
Screenshot 2024-05-01 at 07 22 45

---------------

For [this commit ](https://github.com/mysociety/fixmystreet/pull/4728/commits/6a69f8eb942701f31ff75690e6ce4206c97028e4) I replaced `multi-select.min.js` with the one created on [this PR](https://github.com/mysociety/jquery-multi-select/pull/42) on the jQuery-multi-select repo.

- [x] Clean unnecessary code, specially on `_report_list.scss`
- [x] Override the default font size at source when compiling the GOVUK CSS so then it would apply everywhere the same(from 19px to 16px)
- [x] Decide best layout horizontal or vertical stacking
- [x] Test on mobile
- [x] Check Camden
- [x] Update labels so we comply with WCAG 2.53 and fix [issue 4214](https://github.com/mysociety/societyworks/issues/4214)


lucascumsille:
@dracos in order to fix https://github.com/mysociety/societyworks/issues/4214
I was thinking to use `aria-labelledby` attribute, we would require to add an `id` for `label` and the fake label `span`(Option 1). I would probably make those changes in the plugin itself. However that would only cover the `label` and `.multi-select-button`, the creation of the fake label, it's something we do in FixMyStreet(I think), which makes me wonder if we could just undo that, basically not creating the new `span` and use the label, without hiding it and add `aria-label` so we can explain better what the select element does(This would be Option 2).

### Option 1 
```
 <label class="hidden-js" id="filter_category_label">About - Select a category</label>
  <span id="filter_category_text">About</span>
  
  <span class="multi-select-button" role="button" aria-haspopup="true" tabindex="0" aria-labelledby="filter_category_label filter_category_text">Everything</span>
```

### Option 2
```
  <label id="filter_category_label" for="filter_category_button">About</label>
  <span class="multi-select-button" id="filter_category_button" role="button" aria-haspopup="true" tabindex="0" aria-labelledby="filter_category_label">Everything</span>
```

dracos:
If we're changing the structure to look more like normal fields, it doesn't have to be "About" any more, that's not needed, it's no longer displayed like a readable sentence so doesn't need to be one. It could be "Category". Does that help? Then there's no need for things to be different anywhere. Or have I misunderstood?

dracos:
(The history of this FMS changing label to span is https://github.com/mysociety/fixmystreet/pull/3881 and https://github.com/mysociety/fixmystreet/pull/3863 before that.)

lucascumsille:
> If we're changing the structure to look more like normal fields, it doesn't have to be "About" any more, that's not needed, it's no longer displayed like a readable sentence so doesn't need to be one. It could be "Category". Does that help? Then there's no need for things to be different anywhere. Or have I misunderstood?

To clarify, In this case we would have `label` + `.multi-select-button` However in terms of markup should we  still include the `for` attribute to the `label` for accessibility purpose. I agree with category it should suffice.

lucascumsille:
@dracos I see good to know, I brought the `for` because I didn't see it in this PR, but I can see it on the live website, so I'll check what is breaking that.
In the mean time I have adjusted the `span` so now it says "Category"

lucascumsille:
@dracos `for` attribute was being removed from the label because of this commit(https://github.com/mysociety/jquery-multi-select/commit/ed762f7582a73702a29208f197a065cfbaff37f0) on the jQuery-multi-select plugin. I have added a piece of [code here ](https://github.com/mysociety/fixmystreet/pull/4728/commits/cccf1ac705ec5e2e677ae051d0163fc08ebf1519)to bring it back.

lucascumsille:
@dracos Here is a preview of the current state. I have updated the screenshot above as well.

Screenshot 2024-05-01 at 07 22 45
Screenshot 2024-05-01 at 07 23 55

dracos:
> for attribute was being removed from the label because of this commit(https://github.com/mysociety/jquery-multi-select/commit/ed762f7582a73702a29208f197a065cfbaff37f0) on the jQuery-multi-select plugin. I have added a piece of [code here ](https://github.com/mysociety/fixmystreet/pull/4728/commits/cccf1ac705ec5e2e677ae051d0163fc08ebf1519)to bring it back.

I was vrery confused, because the `for` is currently present on the live site, and the live jquery-multi-select includes the commit you're referring to. But then I found my commit https://github.com/mysociety/fixmystreet/commit/235931eb12008f0cb5511587f234694281cfd176 where I altered the FMS jquery-multi-select but didn't put this change in the jquery-multi-select plugin :-/ And when this PR updated the plugin for the new stuff there, that disappeared. So that would explain that! I guess your change is a better way to do it if it has the same effect and doesn't alter the plugin.

lucascumsille:
I added [this commit](https://github.com/mysociety/fixmystreet/pull/4728/commits/b1c2dbb7f43f91a2ead5ffb50d50861c735386f8) that should help us with this [issue](https://github.com/mysociety/societyworks/issues/4215). I added it here because I though the layout of the current site will change with this PR, therefore it made more sense to include it here, but let me know if you think it shouldn't be here.

dracos:
@lucascumsille  I don't think that's what that issue is about - that issue is saying that if you /change/ the filter, it changes automatically without you having to click a Submit, rather than what is happening on page load. The solution is we either add a submit button (meh, guess is easier not to have one, but could do), or add a message saying filters apply automatically (not what /was/ applied).

lucascumsille:
@dracos I added a [fixup](https://github.com/mysociety/fixmystreet/pull/4728/commits/95b0ca2cff4f026f756d205df8745a13a74d7cc9) for the message commit. At the beginning I thought it would be nice to let the user know what filters have been applied, but yeah it does overcomplicate the ticket.
Let me know if you are happy with this approach.

templates/web/base/reports/_list-filters.html
```diff
@@ -27,26 +28,39 @@
...
 [% IF use_form_wrapper %]
     <form method="get" action="">
 [% END %]
 
-        <p class="report-list-filters">
-          [% tprintf(loc('<label for="statuses">Show</label> %s reports <label for="filter_categories">about</label> %s', "The first %s is a dropdown of all/fixed/etc, the second is a dropdown of categories"), mark_safe(select_status), mark_safe(select_category)) %]
-          <input type="submit" name="filter_update" value="[% loc('Go') %]">
-        </p>
+        <fieldset>
+            <div class="multi-select-filter-group">
```
dracos, line 38: I think this can just be filter-group - there's no difference?

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1086,7 +1093,7 @@ $.extend(fixmystreet.set_up, {
     $('.js-multiple').make_multi();
 
     function update_label(id, str) {
-        $(id).prev('label').addClass('hidden-js').after(function(){ return $('<span>' + this.innerHTML + '</span>'); });
+        $(id).prev('label').addClass('hidden-js').attr('for', id.slice(1)).after(function(){ return $('<span>' + this.innerHTML + '</span>'); });
```
dracos: This needs to be prevAll because the filter announcement gets in the way otherwise

dracos (changes requested):
Looks good, have added fixup for prevAll and to actually have the aria-label on the select match the visible text (which I think was the issue being complained about in 4214). A number of hopefully small issues to deal with

templates/web/base/reports/_list-filters.html
```diff
@@ -27,26 +28,39 @@
...
+                <div class="report-list-filters">
+                    [% tprintf(loc('<label for="statuses">Report status</label> %s', "Dropdown of all/fixed/etc"), mark_safe(select_status)) %]
```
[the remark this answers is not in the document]
dracos: Have pushed fixup for this

web/cobrands/sass/_multiselect.scss
```diff
@@ -158,3 +148,124 @@
...
+font-size: 1.1875rem;
+line-height: 1.3157894737;
+position: relative;
+
+    .govuk-small & {
+        font-size: 1rem;
+        line-height: $govuk-small-input-visible-size;
+    }
+
+    &:has(> input[type="checkbox"]:checked),
```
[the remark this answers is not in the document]
lucascumsille: @dracos I have added this commit that should solve the issue.
https://github.com/mysociety/fixmystreet/pull/4728/commits/0807af665bb667427607e3da9ce4114f11e5da53

lucascumsille:
@dracos Thanks for the feedback I have added a fixup for all the issues above with a thumbs up, I think that should be everything. When we are ready I'll tidy up Camden, which most likely will need some amendments with this PR.

templates/web/base/reports/_list-filters.html
```diff
@@ -27,26 +28,39 @@
...
+                    <input type="submit" name="filter_update" value="[% loc('Go') %]">
```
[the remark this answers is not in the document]
dracos: There are still two go buttons, one under category, one under show older reports; would be good to only have one at the end (so after the show older tickbox, outside the page==around check, I guess?)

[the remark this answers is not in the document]
lucascumsille: @dracos I added [this commit ](https://github.com/mysociety/fixmystreet/pull/4728/commits/948c6ce17c8a07f862aa6da41af660279d7396a4), I did a minor tweak to the JS, because the markup couldn't target the button anymore, for being outside of `.report-list-filters`

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1100,6 +1100,35 @@ $.extend(fixmystreet.set_up, {
...
+            } else if (input.attr('type') === 'radio') {
+              label.addClass('govuk-multi-select__label--radio');
+            }
+      
+            if (input.prop('checked')) {
+              label.addClass('govuk-multi-select__label--checked');
+            }
+      
+            input.on('change', function() {
+              if (this.checked) {
```
dracos, line 1122: This is fine, but you can also have this as one line with:
`label.toggleClass('govuk-multi-select__label--checked', this.checked);`

templates/web/base/reports/_list-filter-status.html
```diff
@@ -1,5 +1,6 @@
 [% SET show_all_states = c.cobrand.filter_show_all_states OR (c.user_exists AND (c.user.is_superuser OR c.user.belongs_to_body(body.id))) %]
 
+<div id="filter-announcement" class="screen-reader-only" role="alert">Information: Filters in use. The current view is customized based on selected filters.</div>
```
dracos, line 3: Should have a loc() around this text string

dracos (commented):
Looks good :-) Will need a bit of checking that the changes to report lists/ multi selects etc haven't affected anything else, I guess.

lucascumsille:
@dracos I found some issues with three cobrands. I couldn't figure out what was causing it, not sure if we are doing something different with them, I couldn't find a template override.

The problem is with "Category" filter, in the markup the `select` element is not present and also not the `multi-select-button`

I tried commenting out this line:
`$('.js-multiple').make_multi();`, but the select element  wouldn't be created.

The Status was is fine

- [ ] highwaysengland
- [ ] isleofwight
- [ ] TFL

lucascumsille:
@dracos I checked all the other cobrands and they are working fine on desktop and mobile.
There were a couple of cobrand that needed some[ tweaks though.](https://github.com/mysociety/fixmystreet/pull/4728/commits/ad64f8e6ac82cb34abf948576828539ad6812d80)

Here is a preview of them:
Screenshot 2024-06-24 at 08 10 23
Screenshot 2024-06-24 at 08 10 13
Screenshot 2024-06-24 at 08 40 59
Screenshot 2024-06-24 at 08 41 17
Screenshot 2024-06-24 at 08 06 51
Screenshot 2024-06-24 at 08 12 13
Screenshot 2024-06-24 at 08 01 44
Screenshot 2024-06-24 at 08 12 02
Screenshot 2024-06-24 at 07 59 27
Screenshot 2024-06-24 at 08 11 49
Screenshot 2024-06-24 at 07 56 28
Screenshot 2024-06-24 at 08 11 38
Screenshot 2024-06-24 at 07 55 42
Screenshot 2024-06-24 at 08 11 28
Screenshot 2024-06-24 at 08 42 50
Screenshot 2024-06-24 at 08 43 02
Screenshot 2024-06-24 at 08 43 11
Screenshot 2024-06-24 at 08 43 20
Screenshot 2024-06-24 at 08 43 35
Screenshot 2024-06-24 at 09 26 29
Screenshot 2024-06-24 at 09 28 04
Screenshot 2024-06-24 at 07 12 12
Screenshot 2024-06-24 at 08 11 17
Screenshot 2024-06-24 at 08 42 31
Bexley
Screenshot 2024-06-24 at 08 11 08
Screenshot 2024-06-24 at 08 42 17
Screenshot 2024-06-24 at 09 28 47
Screenshot 2024-06-24 at 09 28 56
Screenshot 2024-06-24 at 09 29 13
Screenshot 2024-06-24 at 09 29 21
Bathnes
Screenshot 2024-06-24 at 08 42 11
Screenshot 2024-06-24 at 09 28 38
Screenshot 2024-06-24 at 08 42 01
Screenshot 2024-06-24 at 09 28 27

dracos:
Hey, three cobrands - if you haven't set up bodies for them, then it'll just say "Everything" like this:
image
Is that what you mean? That'll be fine in the real thing if so.

lucascumsille:
@dracos ohh perfect
> Hey, three cobrands - if you haven't set up bodies for them, then it'll just say "Everything" like this:

Then we should be ready.

There is one thing that might be odd, I didn't see it till I went to admin, with this deployment we only have styled the multi-select classes as GOVUK ones, which was fine in the user side of things because I couldn't see an scenario where NON multi-select `select` elements would be alongside multi-select elements. The admin side of things will have case that look like this:

Screenshot 2024-06-25 at 10 27 18

We currently have [this ticket to tackled the admin side](https://github.com/mysociety/societyworks/issues/3828) which most likely will be next. However I wouldn't mind to add a commit here that changes the style of `select` elements so they all look like that, no matter which page you are.

dracos:
I don't think that matters really - that was already the case (if slightly more obvious now). What does the dashboard look like? Currently it's eg 
image
That's the only other place I can think of.

lucascumsille:
me neither happy to deploy and then tackle all those cases with the rest of tickets. @dracos, let me know if you are happy with this and I'll rebase and merge. 
Regarding the commits, I'm thinking of getting rid of the --fixups and squash ["Added styling fix for cobrands"](https://github.com/mysociety/fixmystreet/pull/4728/commits/ad64f8e6ac82cb34abf948576828539ad6812d80) into [Added GOVUK styling to multiselect plugin](https://github.com/mysociety/fixmystreet/pull/4728/commits/d014ab2be5288d4c3c52151304e6f8d2125d6a7c).

But let me know if you'd rather have less commits for this PR.

dracos:
@lucascumsille Doing something about the dashboard as I mentioned in my previous message would be the only thing I'd like to be looked at - it looks a bit odd now:
image
Can we for now just shrink them there or something like that?

lucascumsille:
@dracos Sorry I had misread your comment. I added a [commit](https://github.com/mysociety/fixmystreet/pull/4728/commits/8dca6c02d5e4f07d92c7317606b2fbbd98df5254) here that reduces the height of the GOVUK ones only for the dashboard. I did't match it exactly, because I would have to reduce the font-size as well, but this approach should get rid of the oddness.

Screenshot 2024-07-02 at 14 45 59

Let me know if you are happy with it and how you'd like me to handle the rebase. I proposed this a few days ago:
>Regarding the commits, I'm thinking of getting rid of the --fixups and squash ["Added styling fix for cobrands"](https://github.com/mysociety/fixmystreet/pull/4728/commits/ad64f8e6ac82cb34abf948576828539ad6812d80) into [Added GOVUK styling to multiselect plugin](https://github.com/mysociety/fixmystreet/pull/4728/commits/d014ab2be5288d4c3c52151304e6f8d2125d6a7c).

lucascumsille:
Thanks @dracos looks good to me, thanks for adding it to the changelog as well =)

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -132,6 +132,13 @@ function isR2L() {
             settings.noneText = $select.data('none');
         }
 
+        if ($select.attr("id") == 'filter_categories' || $select.attr("id") == 'statuses') {
+            settings.menuItemsHTML = '<div class="govuk-multi-select govuk-multi-select--checkboxes">';
+            settings.menuItemHTML = '<label class="govuk-multi-select__label">';
+            settings.menuFieldsetHTML = '<fieldset class="multi-select-fieldset govuk-fieldset">';
+            settings.menuFieldsetLegendHTML = '<fieldset class="multi-select-fieldset govuk-fieldset__legend govuk-fieldset__legend--s">';
```
dracos, line 139: Not sure how I didn't notice this before, sorry, but this should presumably be a `legend` not a `fieldset`.

lucascumsille:
Thanks @dracos well spotted. I create a PR here:
https://github.com/mysociety/fixmystreet/pull/5069

PR #4730, opened 2023-11-20
[Bucks]  Append some extra field values to description when sending report

davea:
For FD-3519

[skip changelog]


MorayMySoc (approved):
Looks good

dracos:
This was merged as e9edc01204 (which is what our git repo has, not sure why github hadn't updated).

PR #4735, opened 2023-11-22
Daemon locking improvements

dracos:
[skip changelog]


davea (changes requested):
Looks like a nice and neat solution to a thorny problem. I like the change that breaks `open311_update_missing_data` out into its own thing - it was definitely getting a bit heavy in places.

One little copy/paste thing in the SQL migration is all, otherwise 👍

PR #4738, opened 2023-11-23
Dashboard date range fix

dracos:
[skip changelog]
Fixes issue with exact end date matching, and bad dates being supplied to the query.


chrismytton (approved):

dracos:
Just to note start/end_formatted were not cached here, so using them per-loop slowed this down tremendously. Fixed by having DateRange use attributes more, in https://github.com/mysociety/fixmystreet/pull/4741

PR #4740, opened 2023-11-27
[Northumberland] Assign user to report based on updates from Alloy

nephila-nacrea:
Strictly speaking this code is cobrand-agnostic; the Northumberland-specific code is handled in open311-adapter.

For https://github.com/mysociety/societyworks/issues/3978.

[skip changelog]

### DIFF
--- a/bin/fetch
+++ b/bin/fetch
@@ -22,6 +22,11 @@ use Getopt::Long::Descriptive;
 use Open311::GetServiceRequests;
 use Open311::GetServiceRequestUpdates;
 
+use CronFns;
+
+my $site = CronFns::site(FixMyStreet->config('BASE_URL'));
+CronFns::language($site);
+
 my ($opts, $usage) = describe_options(
     '%c %o',
     ['reports', 'fetch reports'],
--- a/perllib/Open311/UpdatesBase.pm
+++ b/perllib/Open311/UpdatesBase.pm
@@ -188,6 +188,36 @@ sub _process_update {
         $request->{comment_time} = $p->whensent + DateTime::Duration->new( seconds => 1 );
     }
 
+    # Assign admin user to report if 'assigned_user_*' fields supplied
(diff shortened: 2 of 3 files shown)


bin/fetch
```diff
@@ -22,6 +22,11 @@ use Getopt::Long::Descriptive;
 use Open311::GetServiceRequests;
 use Open311::GetServiceRequestUpdates;
```
nephila-nacrea, line 24: Not sure if this is needed, but the script was failing on 'Can't call method "get" on an undefined value at /Users/victoria/work/my-society/fixmystreet-local/commonlib/perllib/mySociety/Locale.pm line 22.' after adding my changes 🤔

dracos: Can't hurt! You're saying it didn't do it without your changes? Not sure why you would get this - it means that translations haven't been set up and it's trying to translate something, but nothing you've changed should involve a translation?

dracos: Was it actually fetching an update when giving this error?

perllib/Open311/UpdatesBase.pm
```diff
@@ -188,6 +188,36 @@ sub _process_update {
...
     }
 
+    # Assign admin user to report if 'assigned_user_*' fields supplied
+    if ( $request->{extras} && $request->{extras}{assigned_user_email} ) {
+        my $assigned_user_email = $request->{extras}{assigned_user_email};
+        my $assigned_user_name  = $request->{extras}{assigned_user_name};
+
+        my $assigned_user
+            = FixMyStreet::DB->resultset('User')
+            ->search( { email => $assigned_user_email } )->first;
```
dracos, line 198: Could this use `find()`? That is set up to make sure it only searches verified email addresses.

bin/fetch
nephila-nacrea: Hmm, I'll have another look at this - will try to get the other fixes for this PR done first :-)

dracos (approved):
Thanks :)

PR #4746, opened 2023-11-30
[FMS] Add support for custom duplicate report radius per category

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3889.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -683,14 +683,7 @@ sub _nearby_json :Private {
         # This is for the list template, this is a list on that page.
         $c->stash->{page} = 'report';
 
-        # distance in metres
-        my $dist;
-        if (my $mode = $c->get_param('mode')) {
-            $dist = $c->cobrand->nearby_distances->{$mode};
-        }
-        $dist ||= $c->get_param('distance') || '';
-        $dist = 1000 unless $dist =~ /^\d+$/;
-        $dist = 1000 if $dist > 1000;
+        my $dist = $self->_find_distance($c, $params);
         $params->{distance} = $dist / 1000 unless $params->{distance}; # DB measures in km
 
         my $pin_size = $c->get_param('pin_size') || '';
@@ -725,6 +718,42 @@ sub _nearby_json :Private {
     }
 }
 
+# distance in metres
+#
+# A cobrand may optionally have configured:
+# a) distances for an individual (sub)category
+# b) distances for a group/parent category
+# c) distances by mode
+#
+# NOTE: Distances for category or group only apply for the 'suggestions' mode.
+sub _find_distance {
+    my ($self, $c, $params) = @_;
+
+    my $dist;
+    my $mode = $c->get_param('mode');
+    my $cobrand_distances = $c->cobrand->nearby_distances;
+
+    if ($mode) {
+        if ( $mode eq 'suggestions' && ref $cobrand_distances->{$mode} eq 'HASH' ) {
+            my $category = $params->{categories}[0];
+            my $group    = $params->{group};
+
+            $dist = $category && $cobrand_distances->{$mode}{$category}
+            ? $cobrand_distances->{$mode}{$category}
+            : $group && $cobrand_distances->{$mode}{$group}
+            ? $cobrand_distances->{$mode}{$group}
+            : $cobrand_distances->{$mode}{_fallback};
+        } else {
+            $dist = $cobrand_distances->{$mode};
+        }
+    }
+
+    $dist ||= $c->get_param('distance') || '';
+    $dist = 1000 unless $dist =~ /^\d+$/;
+    $dist = 1000 if $dist > 1000;
+
+    return $dist;
+}
 
 =head2 fetch_permissions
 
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -725,6 +718,38 @@ sub _nearby_json :Private {
...
+#
+# NOTE: Distances for category or group only apply for the 'suggestions' mode.
+sub _find_distance {
+    my ($self, $c, $params) = @_;
+
+    my $dist;
+    my $mode = $c->get_param('mode');
+    my $cobrand_distances = $c->cobrand->nearby_distances;
+
+    if ($mode && $cobrand_distances) {
```
dracos, line 736: $cobrand_distances will always be true

perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -725,6 +718,38 @@ sub _nearby_json :Private {
...
+        my $category = $params->{categories}[0];
+        my $group    = $params->{group};
+
+        $dist
+            = $category && $cobrand_distances->{$category} && $mode eq 'suggestions'
```
dracos, line 741: Having the config like this is okay, though it seems like it'd take up a lot of config file space. Wonder if we could restructure so instead of:
```
nearby_distances:
  oxfordshire:
    suggestions: 800
    'Subcat 1 in Group 1':
      suggestions: 100
    'Group 1':
      suggestions: 400
```

it was structured the other way round, as:
```
nearby_distances:
  oxfordshire:
    suggestions:
      _fallback: 800
      'Subcat 1 in Group 1': 100
      'Group 1': 400
```

(_fallback matching other cobrand features)

t/app/controller/around.t
```diff
@@ -556,14 +556,184 @@ subtest 'check nearby lookup, cobrand custom distances' => sub {
...
+                    . $test->{lon}
+                    . '&filter_category=Pothole'
+                    . ( $test->{mode} ? '&mode=' . $test->{mode} : '' )
+            );
             $mech->contains_or_lacks($test->{contains}, '[51.754926,-1.256179,"yellow",' . $p->id . ',"Open: Around page Test 1 for ' . $body->id . '","small",false]');
         }
     };
 };
 
+subtest 'check nearby lookup, cobrand custom distances per category' => sub {
```
dracos, line 572: Not introduced by you, but these tests are making external requests to tilma (but not failing if internet not present, hence not spotted - I only noticed because I wondered why they were so slow). I've pushed a fix to master if you could rebase on that.

t/app/controller/around.t
```diff
@@ -556,14 +556,184 @@ subtest 'check nearby lookup, cobrand custom distances' => sub {
...
+
+        note 'filter_category = Subcat 1 in Group 1, filter_group = Group 1';
+        for my $test (
+            { lat => 51.7549, lon => -1.256, mode => 'suggestions', filter_category => 'Subcat 1 in Group 1', filter_group => 'Group 1', contains => 1}, # 12m away
+            { lat => 51.752, lon => -1.256, mode => 'suggestions', filter_category => 'Subcat 1 in Group 1', filter_group => 'Group 1', contains => 0}, # 325m away
+            { lat => 51.7485, lon => -1.256, mode => 'suggestions', filter_category => 'Subcat 1 in Group 1', filter_group => 'Group 1', contains => 0}, # 714m away
+            { lat => 51.74, lon => -1.256, mode => 'suggestions', filter_category => 'Subcat 1 in Group 1', filter_group => 'Group 1', contains => 0}, # 1660m away
+        ) {
+            $mech->get_ok( '/around/nearby?latitude='
+                    . $test->{lat}
```
dracos, line 619: Happy to be /bit/ more spaced than previous, but could it perhaps be `'&longitude=' . $test->{lon}` with each variable on one line?

t/app/controller/around.t
```diff
@@ -556,14 +556,184 @@ subtest 'check nearby lookup, cobrand custom distances' => sub {
...
+                    . '&longitude='
+                    . $test->{lon}
+                    . '&mode='
+                    . $test->{mode}
+                    . '&filter_category='
+                    . $test->{filter_category}
+                    . '&filter_group='
+                    . $test->{filter_group} );
+            for my $test_p (
```
dracos, line 628: I'd probably find this more readable as:
```
            for my $test_p (
                { id => $p_1_g_1->id, captured_by_category_filters => 1 },
                { id => $p_2_g_1->id, captured_by_category_filters => 0 },
                { id => $p_g_2->id,   captured_by_category_filters => 0 },
```

dracos (changes requested):
Looks good :) I've suggested restructuring how the config will be held, happy to hear thoughts on that, and a few other small bits and bobs

perllib/FixMyStreet/App/Controller/Report.pm
nephila-nacrea: Would that require restructuring the existing config as e.g. it is for live? I.e.
```
nearby_distances:
    _fallback:
      inspector: 1000
      suggestions: 250
    brent: 
      inspector: 1000
      suggestions: 15
    ...
```
would become
```
nearby_distances:
    _fallback:
      inspector: 1000
      suggestions:
        _fallback: 15
    brent: 
      inspector: 1000
      suggestions:
        _fallback: 15
    ...
```

I'll check if this is the case.

nephila-nacrea: Ah didn't realise you'd already replied 😅

t/app/controller/around.t
nephila-nacrea: Yeah - I often do some auto-formatting / perltidy, but it's a little overeager sometimes.

dracos (approved):
Yes, think config nicer that way round, thanks, looks good :)

PR #4747, opened 2023-11-30
[Northumberland] Add response time to CSV export

davea:
For FD-3351

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -4,6 +4,9 @@ use base 'FixMyStreet::Cobrand::UKCouncils';
 use strict;
 use warnings;
 
+use DateTime::Format::W3CDTF;
+use Utils;
+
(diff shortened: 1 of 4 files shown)


dracos (commented):
My only concern is speed-wise (it might be worth you running `all-csv-export --body 2248` on the full dataset before/after this change to see what sort of difference it makes) - in 'pregeneration' mode, the `$response_time` function will be called once per comment, not once per report, and so repeating itself if there are comments after a fixed one (though the closed timestamp is the latest comment unlike fixed, so repeating that would give different result, don't know if that's desired, perhaps things don't get reopened in Northumberland so doesn't matter). I don't know if we want to add a new hook per 'object' (so same place in Reporting.pm but before `output()` in CSVExport.pm so that it's only called once, once it's compiled all the comment rows. Or this might be fine like this and not matter.
Another alternative would be the pregeneration could ask the database for the epochs directly to save having to compute them here (if that turns out to be a slow bit).

PR #4748, opened 2023-12-01
Update fallback copy.html

sallybracegirdle:
I've made some minor tweaks to the copy and added a link to the new FAQs

Please check the following:

Thanks for contributing!


sallybracegirdle:
Slightly tweaked copy for the 'Start a new offline report' section:

> Fill in the form below to start your report and click ‘save draft’ when you’re done. For peace of mind, any information you provide here will also be saved automatically. When you’re connected to the internet again, come back to finish and submit it. For more information about creating a draft report [see our FAQs](https://www.fixmystreet.com/faq#:~:text=Can%20I%20make%20a%20report%20while%20offline).

dracos:
Just thought - this message is only shown when offline. So you can't read the FAQ (unless we decide we want to cache that offline, which we could do, but that'd be more than just a wording change :) ).

sallybracegirdle:
> Just thought - this message is only shown when offline. So you can't read the FAQ (unless we decide we want to cache that offline, which we could do, but that'd be more than just a wording change :) ).

Oooh, of course 🤣 Oh well, I guess remove the last sentence then.

sallybracegirdle:
@dracos did you need anything else from me for this? The copy doesn't look like it's been updated (when I checked anyway)

dracos:
It's definitely updated and deployed, and the hash is correct in the service worker. The problem with cached offline stuff is that it can be quite tricky to get things to update once they're cached and offline. But it definitely is. I have seen it updated on my phone in web app, but not in the native. I don't know how often the native app updates the service worker, perhaps @davea  does, but anyway, it is there.

davea:
Hasn't updated on mine yet either. I don't think it follows different caching rules to the web app.

Towards the bottom of the offline page there's a line showing when the page was cached:

image

PR #4755, opened 2023-12-06
Heatmap speed improvements

dracos:
The state change should help everywhere, but it's most noticeable on the heatmap [skip changelog]


davea (approved):
Woo! This is much much faster now ⚡

dracos:
After testing this live, I got annoyed by the initial page loading speed (the pins after that were much better from the above), given it was then reloading the same data, and saw some double loading, so I added another commit, 660f74d9bfca31201f72c05e62848969caa3e6e6 which stopped the sidebar from being included at initial load, and made sure the markers were hidden to prevent a double load of (admittedly now faster) data.

PR #4756, opened 2023-12-06
[Shropshire] Add relevant messaging to the no_asset_message

MorayMySoc:
As locking the page from progressing is done using the standard no asset process, update the no_asset_message attribute to contain the same informative messaging as at the continue button.

Also solves issues of informative messaging not appearing on the subcategory screen if you get to subcategory and then choose a parish light which would mean there was no explanation on the page as to why you were blocked from going forward.

Calling "fixmystreet.assets.shropshire.streetlight_asset_message" for the text did not always update so repeated the logic of creating the message. 

https://github.com/mysociety/societyworks/issues/3996

https://github.com/mysociety/societyworks/issues/4014

[skip changelog]


dracos (changes requested):
This does feel like the easiest solution, well done for thinking of it.
Not sure why calling streetlight_asset_message shouldn't work, though, that's a self contained function that only compiles a string, so what happened when it didn't? It'd be better if it could, certainly, to save the repeating.

dracos (approved):
Woo, glad we tracked that down :)

PR #4757, opened 2023-12-06
[Bromley] [Waste] Fix bookings marked payment failed not being cancelled.

neprune:
[skip changelog] [FD-2623](https://mysocietysupport.freshdesk.com/a/tickets/3623)


dracos (changes requested):
I wonder if it'd be easier if, instead of changing this query, in `csc_payment_failed` you immediately closed the report etc there if `bulky_send_before_payment` is set (as it is for Bromley) because presumably you know it can be done at that point. I may have missed the reason not to do this, though!

Issue #1265, opened 2015-10-19
Allow specific users to be nominated to never receive questionnaires


davewhiteland:
Currently, whether or not questionnaires are sent depends on a system-wide setting.

On the Ugandan platform, there are some nominated users working for the FMS-platform-running team who are submitting accumulated reports on behalf of citizen users. These users (i.e., email-address nominated logins) don't benefit from being sent questionnaires because they're actively monitoring the state of the problems they've submitted anyway.

So maybe add a user-specific flag to the user to "never be sent questionnaires".

I thought this might work as a `has_permission_to` setting, but inspecting the code I suspect that's not the case because these reporter-users aren't associated with a body (that is, they're not staff users in the sense of [user-associated-with-a-body](http://fixmystreet.org/glossary/#staff-user)).

dracos:
Since b01df0c7609a1d6c008b943a9936d471cf0e0ce3 `has_permission_to` is independent of the User’s `from_body` (this was done for trusted users), so this could potentially be done that way now. But it would still apply on a per-body basis, though.

davea:
This would also make for a useful feature for public users who want to opt out of receiving all current/future questionnaires. User support message has mentioned it:

> How do I stop the questionnaire emails coming?
I've responded to one or two and your privacy policy suggested I could then opt out but I've seen no option either when submitting incidents or when responding to questionnaire emails?

PR #4758, opened 2023-12-06
[FMS] Allow users to opt out of questionnaires

nephila-nacrea:
Closes https://github.com/mysociety/fixmystreet/issues/1265.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/My.pm
+++ b/perllib/FixMyStreet/App/Controller/My.pm
@@ -382,10 +382,12 @@ sub notify_preference : Local : Args(0) {
 
     my $update_notify = $c->get_param('update_notify');
     my $alert_notify = $c->get_param('alert_notify');
+    my $questionnaire_notify = $c->get_param('questionnaire_notify');
 
     $c->user->set_extra_metadata(
         update_notify => $update_notify,
         alert_notify => $alert_notify,
+        questionnaire_notify => $questionnaire_notify,
     );
     $c->user->update;
     $c->res->redirect('/my');
--- a/perllib/FixMyStreet/Script/Questionnaires.pm
+++ b/perllib/FixMyStreet/Script/Questionnaires.pm
@@ -57,9 +57,12 @@ sub send_questionnaires_period {
         # Cobrands can also override sending per row if they wish
         my $cobrand_send = $cobrand->call_hook('send_questionnaire', $row) // 1;
 
-        if ($row->is_from_abuser || !$row->user->email_verified ||
-            !$cobrand_send || $row->is_closed
-           ) {
+        if (   $row->is_from_abuser
+            || !$row->user->email_verified
+            || !$row->user->questionnaire_notify
+            || !$cobrand_send
+            || $row->is_closed )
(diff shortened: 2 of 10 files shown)


perllib/FixMyStreet/App/Controller/My.pm
```diff
@@ -382,10 +382,12 @@ sub notify_preference : Local : Args(0) {
 
     my $update_notify = $c->get_param('update_notify');
     my $alert_notify = $c->get_param('alert_notify');
+    my $receive_questionnaires = $c->get_param('receive_questionnaires');
```
dracos, line 385: This works fine, but wondered if `questionnaire_notify` would be keeping in sync with the naming of the other similar variables `update_notify` and `alert_notify`?

templates/web/base/my/my.html
```diff
@@ -91,10 +91,23 @@ <h1>[% loc('Your account') %]</h1>
...
+      <p class="segmented-control segmented-control--radio">
+        <input type="radio" name="receive_questionnaires" id="receive_questionnaires_yes" value="1"[% ' checked' IF current_pref %]>
+        <label class="btn" for="receive_questionnaires_yes">[% loc('Yes') %]</label>
+
+        <input type="radio" name="receive_questionnaires" id="receive_questionnaires_no" value="0"[% ' checked' IF !current_pref %]>
+        <label class="btn" for="receive_questionnaires_no">[% loc('No') %]</label>
+      </p>
+    </fieldset>
+</li>
+<p>
```
dracos, line 108: Any reason this has moved outside the final `li`?

perllib/FixMyStreet/Script/Questionnaires.pm
```diff
@@ -43,6 +43,9 @@ sub send_questionnaires_period {
 
     while (my $row = $unsent->next) {
 
+        # User may have opted out of questionnaires
+        next unless $row->user->receive_questionnaires;
```
dracos, line 47: We need to mark the report as not sending questionnaires. Otherwise if in 6 months they turn getting questionnaires back on, they will presumably receive every questionnaire they've ever missed :)
If you move this down to the check around line 63 to add to the set of checks there (cobrand_send, row->is_closed, etc) and add this as an extra check there.

dracos (changes requested):
Looking good, thanks for the privacy page updates, couple of small things, plus one issue to stop people suddenly getting all their missing questionnaires.

templates/web/base/my/my.html
nephila-nacrea: I moved it because being inside the final `li` seemed like a really odd pattern to me (why would the submit button for the entire list be wrapped by one of the list items?) But it works either way so I can undo this bit.

dracos: Thanks; the reason was/is because then the list button displays more as 'part of' the options being updated, whereas outside the list it spans the whole column and feels less joined to the options it is associated with (probably could be fixed with some extra styling but didn't seem worth it).

dracos (approved):
Great :)

PR #4759, opened 2023-12-06
Correct footer and other template issues

MorayMySoc:
A number of templates for Bromley and Peterborough have been testing against 'bodyclass' expecting it to be a single word string. It can be a multiple word string so the test is changed to account for that.

This fixes issue with Bromley not getting analytics data since bulky waste was launched https://mysocietysupport.freshdesk.com/a/tickets/3617

Also fixes footer issues containing wrong SocietyWorks link for WasteWorks

https://github.com/mysociety/societyworks/issues/4012

[skip changelog]


templates/web/peterborough/main_nav_items.html
```diff
@@ -1,4 +1,4 @@
-[%~ IF bodyclass == 'waste' ~%]
+[%~ IF bodyclass.contains('waste') ~%]
```
dracos, line 1: This should be match like the others?

dracos (approved):
Woo

PR #4763, opened 2023-12-11
[Bromley] [Waste] Fix bookings marked payment failed not being cancelled.

neprune:
[skip changelog] [FD-2623](https://mysocietysupport.freshdesk.com/a/tickets/3623)

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -562,6 +562,12 @@ sub csc_payment_failed : Path('csc_payment_failed') : Args(0) {
     $report->update_extra_field({ name => 'payment_reference', value => 'FAILED' });
     $report->update;
 
+    if (($report->category eq 'Bulky collection' || $report->category eq 'Small items collection') && $c->cobrand->bulky_send_before_payment) {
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
Thanks, think this is simpler - now it's here I note one thing that might have to change, which is the "Cancelled at user request" and "Booking cancelled by customer" texts probably should perhaps only be "Cancelled" and "Booking cancelled" in this situation.

dracos (changes requested):
Sorry, I wasn't clear, by "in this situation" I meant have "Booking cancelled" for when this script causes it, but leave it with '.by customer' for when the customer has cancelled it (useful information to have). Looks like test is failing anyway, so hopefully can change while doing that.

dracos:
(Oh, no, looks like test failure is unrelated, from @chrismytton mention yesterday as to something)

neprune:
> Sorry, I wasn't clear, by "in this situation" I meant have "Booking cancelled" for when this script causes it, but leave it with '.by customer' for when the customer has cancelled it (useful information to have). Looks like test is failing anyway, so hopefully can change while doing that.

Sorry, complete brain failure on my part, I think you were clear.

PR #4766, opened 2023-12-12
[FMS][Open311] Send images when comment on private report

MorayMySoc:
If an image is on a comment which was made on a private report Open311 can not download the image as it will not be visible.

In this case, for Alloy, send the photos directly to open311

Open311 end still to do

https://github.com/mysociety/societyworks/issues/3709

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Open311Alloy.pm
+++ b/perllib/FixMyStreet/Roles/Open311Alloy.pm
@@ -12,6 +12,7 @@ sub open311_config {
 
     $params->{multi_photos} = 1;
     $params->{upload_files} = 1;
+    $params->{upload_files_for_updates} = 1;
 }
 
 sub open311_extra_data_include {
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
It looks like open311-adapter Confirm and Alloy are the main ones that do something with a photo on an update. I assume we're not changing Confirm as part of this (not sure what it would do), so we only want this change for Alloy? So I guess I'm saying you need a different flag to check to do this direct attachment of photo in an update, where you know the other end can handle it. Your new `_add_photos_to_upload` isn't checking `upload_files` as `_populate_service_request_uploads` does, so it should probably be doing that, but also have a new flag upload_files_updates or something better, to check, and then can be added to Open311Alloy?

MorayMySoc:
I'm thinking that now, testing for `upload_files_for_updates` is tested before getting to` _add_photos_to_upload` so should be ok

t/sample.jpg
dracos, line 1: Not sure why this has been added? :)

dracos (approved):
Looks good, not sure why the extra image is present though is all.

MorayMySoc:
Not merged as broke master so had to reset master

Issue #3908, opened 2022-05-09
Allow report to be marked as not going to be sent [4hrs]


dracos:
You can mark a report as sent in the admin, but would be good to have a way to remove the bodies_str entirely if it's not going to be sent (or some other way of doing that - that's historically the way a report exists but "not reported", but that does lose the "who would this report have gone to" - if we want a difference between that, might need a different way of marking this, but is rare?)

**Work required**

- [x] Add a way to edit the send state of a report from the admin

dracos:
Ties in with https://github.com/mysociety/fixmystreet/issues/4048

lizettal:
@dracos this old ticket is showing as in progress, but the ticket you linked above is complete so is there anything further that needs doing on this ticket?

dracos:
Yes, we still ideally would like a way to edit the send state of a report from the admin, which is currently not possible.

PR #4767, opened 2023-12-13
[FMS] Add a dropdown to the admin to change send_status

MorayMySoc:
If a report is marked as not sent can choose whether to set it as processed or skipped.

https://github.com/mysociety/fixmystreet/issues/3908

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -330,6 +330,9 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         } else {
             $problem->unset_extra_metadata('closed_updates');
         }
+        if ($c->get_param('send_state') && ($c->get_param('send_state') ne $problem->send_state)) {
+            $problem->send_state($c->get_param('send_state'));
+        };
 
         $c->forward( '/admin/reports/edit_category', [ $problem, $problem->state ne $old_state ] );
         $c->forward('/admin/update_user', [ $problem ]);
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -330,6 +330,11 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         } else {
             $problem->unset_extra_metadata('closed_updates');
         }
+        if ($c->get_param('send_state') && ($c->get_param('send_state') ne $problem->send_state)) {
+            $c->forward('/auth/check_csrf_token');
```
dracos, line 334: This has already been checked above on line 296.

templates/web/base/admin/reports/_edit_main.html
```diff
@@ -64,6 +64,14 @@
     <input type="submit" class="btn" name="mark_sent" value="[% loc('Mark as sent') %]">
   [% END %]
 </li>
+[% IF NOT problem.whensent %]
```
dracos, line 67: We probably want to restrict this to superusers, might be hard to explain otherwise.

perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -330,6 +330,11 @@ sub edit : Path('/admin/report_edit') : Args(1) {
...
+            $problem->send_state($c->get_param('send_state'));
+            $c->forward( '/admin/log_edit', [ $id, 'problem', 'send_state_change'] );
```
dracos, line 336: Do we think this deserves a special log entry? I guess mark as sent/resend do, and it's not really an "edit" is it.

dracos (changes requested):
Couple of small changes is all

MorayMySoc:
I changed an `unlike` to `content_unlike` as well, as think that wasn't doing what I expected.

dracos (approved):

PR #4769, opened 2023-12-14
[Bucks] Add video to front page

dracos:
Tidy up front page templates somewhat, and move Oxfordshire to lite embed to save c. 1MB in bandwidth!
[skip changelog]


chrismytton (approved):
Looks good! 📹 

Was wondering about the other OCC video in the reporting flow, but I see that's lazy-loaded in a fancybox iframe, so guess not much benefit to using lite-yt there.

PR #4778, opened 2024-01-03
[Echo] Try and fetch new expired slots on sending; Convert all images to JPEGs

dracos:
Hopefully resolving two reasons we get stuck/problem reports:
1. Some backends have a maximum filesize for images. This has only ever been hit by PNGs people occasionally upload. Rather than limit this to those backends or PNGs, easier to convert all uploaded images to JPEGs.
2. If a report to Echo fails due to expired slots for some reason (we do check at report creation but if there's another issue sending (e.g. issue 1 above) this could happen), try and fetch new ones immediately for next retry rather than await manual interaction.

### DIFF
--- a/perllib/FixMyStreet/Roles/CobrandEcho.pm
+++ b/perllib/FixMyStreet/Roles/CobrandEcho.pm
@@ -1,6 +1,6 @@
 package FixMyStreet::Roles::CobrandEcho;
 
-use strict;
+use v5.14;
 use warnings;
 use DateTime;
 use DateTime::Format::Strptime;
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Roles/CobrandEcho.pm
```diff
@@ -609,6 +609,57 @@ sub clear_cached_lookups_bulky_slots {
...
+            say "No GUID? Creating new GUID and fetching new reservation and sending" if $verbose;
+        }
+        require UUID::Tiny;
+        $guid = UUID::Tiny::create_uuid_as_string();
+        $row->update_extra_field({ name => 'GUID', value => $guid });
+        $row->state('confirmed');
+        $row->resend;
+    } else {
+        say "Not already sent, fetching new reservation and trying again" if $verbose;
+        $p->send_fail_count(0); # Assuming it's been failing, for an instant retry
```
davea, line 639: Should be `$row` I think.

```suggestion
        $row->send_fail_count(0); # Assuming it's been failing, for an instant retry
```

davea (changes requested):
One typo, otherwise looks good! Is there a risk that if Echo somehow always returned `Selected reservations expired` for every attempt a report would be tried over and over again with little throttling? Guess it's unlikely, however.

perllib/FixMyStreet/Roles/CobrandEcho.pm
dracos: Whoops, copied from live script. Also moved it further down to only do this if slot found to prevent the repetition you raise.

davea (approved):

PR #4780, opened 2024-01-04
[Oxfordshire] Include 'in progress' in available states.

neprune:
[skip changelog]

https://github.com/mysociety/societyworks/issues/4054

Make the 'in progress' state available when editing a response template for Oxfordshire, to allow setting up an auto-response for 'Job raised' alloy updates.

The 'in progress' state was initially dropped to simplify the states list: https://github.com/mysociety/fixmystreetforcouncils/issues/129. 

This approach also makes the state available in:
* Dashboard filtering.
* Reports filtering.
* Staff report state editing.
* Updating a report.

Which seems like a good idea with the state now in use for reports via Alloy.


MorayMySoc (approved):
Ok, looks good to me.

PR #4783, opened 2024-01-08
[TfL] Initialise single sign on

MorayMySoc:
Adding single sign on for TfL staff

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -185,6 +185,12 @@ sub around_nearby_filter {
     delete $params->{states};
 }
 
+sub social_auth_enabled {
+    my $self = shift;
+
+    return $self->feature('oidc_login') ? 1 : 0;
+}
+
 sub state_groups_inspect {
     my $rs = FixMyStreet::DB->resultset("State");
     my @open = grep { $_ !~ /^(planned|investigating|for triage)$/ } FixMyStreet::DB::Result::Problem->open_states;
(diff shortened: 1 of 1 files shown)


chrismytton (changes requested):
You'll need to add a `sub user_from_oidc` method in the cobrand to pull the appropriate fields out of the id_token that's returned.

When I did it for Camden I also added in some (temporary) logging so I could see the fields in the id_token, see https://github.com/mysociety/fixmystreet/pull/4170. Then in https://github.com/mysociety/fixmystreet/pull/4295 I updated it to use the correct fields and removed the temporary logging.

Regarding the server config, I think the `logout_uri` needs to be an Azure URL, which we then call to log the user out in Azure. This is the code that uses the `logout_uri` config:

https://github.com/mysociety/fixmystreet/blob/0eac24d8c4e6bce7d404b13d0f6b18e9a05b6563/perllib/FixMyStreet/App/Controller/Auth/Social.pm#L219-L225

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -185,6 +185,26 @@ sub around_nearby_filter {
...
+sub user_from_oidc {
+    my ($self, $payload) = @_;
+
+    # TODO: Remove this once we know which fields the name/email are in.
+    use Data::Dumper;
+    $self->{c}->log->info("TfL OIDC payload: " . Dumper($payload));
+
+    # Extract the user's name and email address from the payload.
+    my $name = $payload->{first_name};
+    my $email = $payload->{last_name};
```
chrismytton, line 203: Unlikely that the email is in the last_name field, but I guess doesn't really matter if you're coming back to change this anyway! 🙂

chrismytton (approved):
Looks good! 👍

chrismytton:
@MorayMySoc Can this be closed now https://github.com/mysociety/fixmystreet/pull/4796 has been merged?

MorayMySoc:
As above - this will not be merged as was superseded by https://github.com/mysociety/fixmystreet/pull/4796

PR #4786, opened 2024-01-09
[FMS.com] Updated mySociety pennant

lucascumsille:
Fixes part of: https://github.com/mysociety/Design-Internal-tasks/issues/35

Because it's not longer the anniversary we are updating the pennants across all core services, but still keeping a pennant so we continue raising awareness of services being linked to mySociety.

Screenshot 2024-01-09 at 13 05 02


templates/web/fixmystreet.com/header_logo.html
```diff
@@ -1,5 +1,5 @@
 [% IF c.req.uri.path == '/' %]
-    <a href="https://www.mysociety.org/about/anniversary" id="mysociety-badge" aria-label="Visit mySociety 20 years anniversary">
+    <a href="https://www.mysociety.org/" id="mysociety-badge" aria-label="A services from mySociety">
```
davea, line 2: This should be "A service from mySociety":
```suggestion
    <a href="https://www.mysociety.org/" id="mysociety-badge" aria-label="A service from mySociety">
```

davea (changes requested):

templates/web/fixmystreet.com/header_logo.html
lucascumsille: Thanks @davea I just pushed a commit

davea (approved):

PR #4787, opened 2024-01-10
[Brent] Increase Green Garden Waste renewal window

MorayMySoc:
Change Brent's ggw renewal from 28 days to 90 days so users can renew from the beginning of the year.

https://mysocietysupport.freshdesk.com/a/tickets/3693

[skip changelog]


dracos (approved):
Think should be fine!

PR #4788, opened 2024-01-10
[Glos] Handle jobs coming from Confirm

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/4029.

Various changes for pulling Confirm jobs into our system.
- Set pin to 'blue man' for jobs
- Handle lat/lon for jobs (behave differently to enquiries)
- Remove redundant wfs_defect_query code
- Create inactive contacts given 'inactive' keyword
- Prevent 'Sent to GCC <time> later' message from showing on jobs

[skip changelog]


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -941,7 +941,9 @@ sub duration_string {
     my $cobrand = $problem->result_source->schema->cobrand;
     my $body = $cobrand->call_hook( link_to_council_cobrand => $problem )
         || $problem->body(1);
-    return unless $problem->whensent;
+    return
+        unless $problem->whensent
+        && ( $problem->external_id // '' ) !~ /^JOB_/;
```
davea, line 946: I wonder if this would be better to look at `$problem->service` and skip if that's `Open311`? That way we'd never show the duration string for any fetched reports.

davea (approved):
One little suggestion for improvement but overall 👍

perllib/FixMyStreet/DB/Result/Problem.pm
nephila-nacrea: I may need to double-check this, but it looks like this removes the 'Sent to...' string for reports which originated on FMS as well.

PR #4790, opened 2024-01-12
Support different /around map display times for open/closed/fixed reports

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/3965.

Cobrands may now have a report_age() that is a hashref of upper-level states (open, closed, fixed) mapped to time intervals.
If a state is not specified, it will default to 6 months (as specified in FixMyStreet/Cobrand/Default.pm).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
         - WasteWorks PWA can now have a separate name from the FixMyStreet PWA.
         - Users can now opt out of questionnaires.
         - Include reference in report list emails. #4760
+        - Add support for different '/around' map display times for open/closed/fixed reports
     - Bugfixes:
         - Stop map panning breaking after long press. #4423
         - Fix RSS feed subscription from alert page button.
--- a/perllib/FixMyStreet/DB/ResultSet/Nearby.pm
+++ b/perllib/FixMyStreet/DB/ResultSet/Nearby.pm
@@ -23,8 +23,17 @@ sub nearby {
         if $args{ids};
     $params->{'problem.category'} = $args{categories} if $args{categories} && @{$args{categories}};
 
-    $params->{$c->stash->{report_age_field}} = { '>=', \"current_timestamp-'$args{report_age}'::interval" }
-        if $args{report_age};
+    my $report_age = $args{report_age};
+    if ( $report_age && ref $report_age eq 'HASH' ) {
+        push @{ $params->{-and} }, FixMyStreet::DB::ResultSet::Problem->report_age_subquery(
+            state_table      => 'problem',
+            report_age       => $report_age,
+            report_age_field => $c->stash->{report_age_field},
+        );
+    } elsif ($report_age) {
+        $params->{ $c->stash->{report_age_field} }
+            = { '>=', \"current_timestamp-'$report_age'::interval" };
+    }
 
     FixMyStreet::DB::ResultSet::Problem->non_public_if_possible($params, $c, 'problem');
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
         - WasteWorks PWA can now have a separate name from the FixMyStreet PWA.
```
nephila-nacrea, line 10: Question: Should I be adding report_age() changes for Northumberland in this PR, or do that separately?

perllib/FixMyStreet/DB/ResultSet/Nearby.pm
```diff
@@ -23,8 +23,17 @@ sub nearby {
...
-        if $args{report_age};
+    my $report_age = $args{report_age};
+    if ( $report_age && ref $report_age eq 'HASH' ) {
+        $params->{'-or'} = FixMyStreet::DB::ResultSet::Problem->report_age_subquery(
+            state_table      => 'problem',
+            report_age       => $report_age,
+            report_age_field => $c->stash->{report_age_field},
+        );
+    } else {
+        $params->{$c->stash->{report_age_field}} = { '>=', \"current_timestamp-'$report_age'::interval" } if
```
dracos, line 34: Now it's reordered like this, I'd probably put the `if $report_age` here instead of the `else`, so it's `elsif ($report_age)`.

CHANGELOG.md
dracos: Yep, happy to have it here :)

perllib/FixMyStreet/DB/ResultSet/Nearby.pm
```diff
@@ -23,8 +23,17 @@ sub nearby {
         if $args{ids};
     $params->{'problem.category'} = $args{categories} if $args{categories} && @{$args{categories}};
 
-    $params->{$c->stash->{report_age_field}} = { '>=', \"current_timestamp-'$args{report_age}'::interval" }
-        if $args{report_age};
+    my $report_age = $args{report_age};
+    if ( $report_age && ref $report_age eq 'HASH' ) {
+        $params->{'-or'} = FixMyStreet::DB::ResultSet::Problem->report_age_subquery(
```
dracos, line 28: There's an issue here because `non_public_if_possible` also sets `-or` directly. We'll need to have both work in a way that they can work with the other.

dracos (changes requested):
One issue involving the interaction with non public parameter is the main issue, I think, framework looks good :)

CHANGELOG.md
nephila-nacrea: Have realised I am not clear on what timeframes, for what states specifically, Northumberland want in https://mysocietysupport.freshdesk.com/a/tickets/3097. 
Seems to change from 'Can we remove closed reports after 5 working days of us closing?' to 'We could do with a change when the state has been changed to fixed closed etc[...]If pins could be removed if no further actions are required it would be good. Say 10 working days upon completion.'
I.e. what started in my interpretation as { closed => '5 days' } becomes { closed => '10 days', fixed => '10 days' }.

Also it may not be clear to Northumberland that we're only implementing this for upper-level states of open/closed/fixed; Kris mentions things like 'Maybe show open, investigating and action scheduled.' though tbh those are all captured under the wider umbrella of 'open' and if that also includes 'for triage' etc. I doubt that's going to be much of an issue. I can leave a comment for Kris requesting clarification.

dracos: Given that ticket, I wouldn't bother requesting clarification, just introduces delays and possibility of further confusion, I would go with closed/fixed of "14 days" ~ equal 10 working days :)

nephila-nacrea: OK 👍 They can say whether they're happy with it after the fact.

dracos (approved):
You found a nice way to combine those SQL bits :)

This (has to) complicate the SQL somewhat, I don't think it should have any adverse effect, the Around lookup should use the bitmap index on state/lat/lon even with multiple different states, and the Nearby is fetching the nearby results first so can go through them. We can keep an eye on it

PR #4795, opened 2024-01-15
Don't cache waste things in session

dracos:
Currently the code stores a number of waste things in the session for caching/performance reasons. We already store most of these on disk, so instead of deleting those, use them as the cache instead. Also start storing Peterborough postcode lookup on disk to match, and move the form unique ID to its own cookie rather than create sessions purely for them as well.
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Form.pm
+++ b/perllib/FixMyStreet/App/Controller/Form.pm
@@ -10,7 +10,16 @@ sub auto : Private {
     my ( $self, $c ) = @_;
     my $cobrand_check = $c->cobrand->feature( $self->feature );
     $c->detach( '/page_error_404_not_found' ) if !$cobrand_check;
-    $c->session->{form_unique_id} ||= mySociety::AuthToken::random_token();
+    my $form_unique_id = $c->request->cookie('form_unique_id');
+    if ($form_unique_id) {
+        $c->stash->{form_unique_id} = $form_unique_id->value;
+    } else {
+        $c->stash->{form_unique_id} = mySociety::AuthToken::random_token();
+        $c->response->cookies->{form_unique_id} = {
+            value => $c->stash->{form_unique_id},
+            expires => time() + 86400,
+        };
+    }
     return 1;
 }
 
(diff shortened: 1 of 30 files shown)


perllib/FixMyStreet/App/Controller/Form.pm
```diff
@@ -10,7 +10,16 @@ sub auto : Private {
     my ( $self, $c ) = @_;
     my $cobrand_check = $c->cobrand->feature( $self->feature );
     $c->detach( '/page_error_404_not_found' ) if !$cobrand_check;
-    $c->session->{form_unique_id} ||= mySociety::AuthToken::random_token();
+    my $form_unique_id = $c->request->cookie('form_unique_id');
+    if ($form_unique_id) {
+        $c->stash->{form_unique_id} = $form_unique_id->value;
+    } else {
+        $c->stash->{form_unique_id} = mySociety::AuthToken::random_token();
+        $c->response->cookies->{form_unique_id} = {
```
davea, line 18: Should this be documented in the cookies table in `privacy.html`?

perllib/FixMyStreet/Roles/ParallelAPI.pm
```diff
@@ -73,7 +73,7 @@ sub call_api {
     if (FixMyStreet->test_mode || $self->sample_data) {
         $start_call = 0;
         $data = $self->_parallel_api_calls(@_);
-    } elsif (-e $tmp) {
+    } elsif (-e $tmp && time-(stat($tmp))[9] < 3600) {
```
davea, line 76: Might be an idea to extract this magic `3600` somewhere more obvious & document it in case we need to change it later on.

perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -472,41 +475,55 @@ Functions specific to the waste product & Bartec integration.
...
     my $c = $self->{c};
+    my $cfg = $self->feature('bartec');
 
-    my $key = "peterborough:bartec:premises_for_postcode:$pc";
+    my $tmp = $self->_premises_for_postcode_key($pc);
 
-    unless ( $c->session->{$key} ) {
-        my $cfg = $self->feature('bartec');
+    my $data;
+    if ($tmp->exists && time - $tmp->stat->mtime < 3600) {
```
davea, line 496: And this one.

davea (approved):
This works nicely, hopefully will speed things along a bit too. Have made a couple of small observations but looks good overall.

dracos:
I rolled this back as it did not work.

PR #4799, opened 2024-01-17
GOVUK Input styling for Get Alert page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3823

[[Alert page] Added GOVUK input style to 'Get alerts'](https://github.com/mysociety/fixmystreet/commit/fbb9c31ad557b52dbb7dbc079fb9501f16d3b195)

Screenshot 2024-01-17 at 08 53 37

Because of the font-size(19px) that GOVUK uses the layout has change a bit, especially with the specify the KM bit, I thought about overriding the GOVUK just to keep the original size, but I think that would defeat the purpose of using the GOVUK styling. There was another alternative, I thought we could get away by using the [smaller versions ](https://design-system.service.gov.uk/components/radios/small/) for the radio buttons, but they also use 19px and their guide advice to use them only when you need to make the section less prominent, considering that the "Get alerts" section is all about the updates I probably wouldn't want to make it less prominent.

[[Alert page] Added GOVUK styling to subscribe by email](https://github.com/mysociety/fixmystreet/commit/0269880cd2367f50fc954681c36d6a011604d911)

Desktop
Screenshot 2024-01-17 at 12 32 03

Mobile
Screenshot 2024-01-17 at 12 42 59

[Replaced RSS feed icon](https://github.com/mysociety/fixmystreet/commit/2ed237cce9da9f77aeb3d980a07a9b7efb49b424)

This commit is not related to the issue, but because I was already working on this page I decide to include it here just to showcase, if we decide to keep it then I can cherry-pick it and create a new PR for this.

Screenshot 2024-01-17 at 09 14 27

[Skip changelog]


dracos (changes requested):
Looking good to me, could you possibly take a look at a two-tier place though, e.g. GL502PR - the live site indenting looks okay, but it seems a bit odd with the larger buttons, could we make it `flex` nowadays and have no left/right margin but still a gap in the middle?

lucascumsille:
>Looking good to me, could you possibly take a look at a two-tier place though, e.g. GL502PR - the live site indenting looks okay, but it seems a bit odd with the larger buttons, could we make it flex nowadays and have no left/right margin but still a gap in the middle?
@dracos tried that approach on the live site, but I think it will look really odd.

Screenshot 2024-06-18 at 12 12 46
Screenshot 2024-06-18 at 12 15 08

What about making them stack on top of each other, the downside you won't have both options side by side, but it will look less overwhelming, which I think is the case at the moment.

Screenshot 2024-06-18 at 12 17 17

dracos:
> tried that approach on the live site, but I think it will look really odd.

On the `/alert` version - you'd still need to make each half flex-grow or 50% or whatever, then I think that would look okay :-)

I agree on the `/around` version it looks better single stacked - presumably we can automatically do this with CSS, flex wrapping until a certain width and then going half-and-half?

lucascumsille:
@dracos I added a fixup that fixes the layout issue when we are having two columns, so if there is enough space, like in the /alert page then we use two columns, but if we don't have enough space then one single column will be displayed. I went from to flex to Grid, because it can handle scenarios where Children need to have min-width otherwise let me use the whole space available.

I still need to test it on my local, because If I use 'GL502PR' I still get the classic one single column view. How can I enable it?

lucascumsille:
@dracos Thanks for the help with the testing. This would be the preview for the `/around` and `/alert` page:

https://github.com/user-attachments/assets/086a1355-52c5-428a-a5eb-b7e8d49f7f4e

Screenshot 2024-08-27 at 11 23 11

templates/web/base/alert/_list.html
```diff
@@ -6,88 +6,106 @@
...
-        [% tprintf(loc('Reports near %s are sent to different councils, depending on the type of problem.'), name_of_location) %]
-    </p>
-    <p>
-        [% loc('You can choose to subscribe to all problems reported in an area, or reports based on their destination.') %]
-    </p>
-    <div class="alerts__columns">
-        <div>
-    <h3>[% loc('Problems in an area') %]</h3>
-  [% END %]
+    <div class="govuk-radios" data-module="govuk-radios">
```
dracos, line 17: We don't use this, so this wrapper div isn't actually needed (but no real harm leaving it in, I guess).

templates/web/base/alert/_list.html
```diff
@@ -6,88 +6,106 @@
...
+          [% END %]
+            [% loc('(covers roughly 200,000 people)') %]
+          </label>
+        </div>
+
+        <div class="alert-option-wrapper govuk-!-margin-bottom-4">
+          <div id="rss_local_alt" class="govuk-form-group">
+            <label class="govuk-label govuk-!-margin-bottom-2" for="distance">[% loc('Or specify a different radius distance:') %]</label>
+            <div class="govuk-input__wrapper">
+              <input class="govuk-input" name="distance" id="distance" type="text" inputmode="numeric" pattern="[0-9]*">
```
dracos, line 48: I think this has lost some changes you've made on master since starting this e.g. the aria-label?

web/cobrands/sass/_base.scss
```diff
@@ -366,6 +366,11 @@ legend,
   }
 }
 
+// For GOVUK Styling
+.govuk-radios__label {
```
dracos, line 370: I agree this needs to be overridden, it probably should be in _gov.scss, though, in that case?

web/cobrands/sass/_base.scss
```diff
@@ -2975,9 +2980,15 @@ a#geolocate_link {
   }
 }
 
+.cta-box__input-wrapper {
+  display: flex;
+  gap: 0.5rem;
+  flex-wrap: wrap;
```
dracos, line 2986: This means it's always two lines at mobile width, is that what's wanted? Seems to be okay on the one line at all widths to me.

web/cobrands/sass/_layout.scss
```diff
@@ -889,12 +889,19 @@ textarea.form-error {
...
 
 .alerts__columns {
-    @include clearfix;
-    @include flex-container();
-    > div {
-        width: 50%;
-        margin: 0 2em;
-    }
+  display: grid;
+  // This tells the column to have a minimum size(23.75rem, around 380px), if it can't be then allows it to take the whole space available.
```
dracos, line 893: Thanks for the comment :)

web/cobrands/sass/_layout.scss
```diff
@@ -889,12 +889,19 @@ textarea.form-error {
...
+  grid-template-columns: repeat(auto-fit, minmax(23.75rem, 1fr));
+  column-gap: 5rem;
+
+  &::before, &::after {
+    position: absolute;
```
dracos, line 898: I'm not sure what this is for?

dracos (changes requested):
Few small changes needed :)

web/cobrands/sass/_base.scss
lucascumsille: I'm not too bother by having only one line, but the advantage of having two lines is for users with longer emails can see their whole input.

web/cobrands/sass/_layout.scss
lucascumsille: We don't need it anymore, before those two elements were being treated as Childs for the grid parent, and using space accordingly. But now we are not using the `clearfix` mixin we were using before, so those pseudo elements are not being created.

templates/web/base/alert/_list.html
```diff
@@ -45,7 +45,7 @@ <h3>[% loc('Problems in an area') %]</h3>
           <div id="rss_local_alt" class="govuk-form-group">
             <label class="govuk-label govuk-!-margin-bottom-2" for="distance">[% loc('Or specify a different radius distance:') %]</label>
             <div class="govuk-input__wrapper">
-              <input class="govuk-input" name="distance" id="distance" type="text" inputmode="numeric" pattern="[0-9]*">
+              <input class="govuk-input" name="distance" id="distance" type="text" inputmode="numeric" pattern="[0-9]*" aria-label="[% loc('Specify a radius in kilometers') %]">
```
dracos, line 48: kilometres, we're not American :)

web/cobrands/sass/_base.scss
dracos: Yeah, it's fine at actual mobile, I guess it's only at 'nearly desktop' width it looks a little odd, probably not worth doing anything about.

dracos (changes requested):
One typo is all, thanks :)

PR #4803, opened 2024-01-19
[TfL] Truncate whensent field in CSVExport

MorayMySoc:
TfL use the whensent field in their reporting and so truncate it to not include nanoseconds

https://mysocietysupport.freshdesk.com/a/tickets/3722

[skip changelog]


dracos (approved):
🚌

PR #4804, opened 2024-01-23
Added variables to control #postcodeForm color

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4006

This variables will allow to avoid custom override for some cobrands for the label and .form-hint colors. In this commit the `$form-hint` variable will depend by default of the `$postcodeform-color`, usually they both share the same background colour, so it seemed like a good move in order to have both variables passing a contrast test, of course `$postcodeform-color` will need to satisfy the criteria otherwise both label and .form-hint will fail. It's imporant to note that $form-hint-color and `$form-hint-color-desktop` are being transparentize by 20%, therefore if `$postcodeform-color` and `$postcodeform-color-desktop` are passing the contrast test, by a small margin, then the label will most likely fail the test.

`$form-hint-color` and `$form-hint-color-desktop` can still be declared in each cobrands in case clients want to use a custom color for it.

[Mobile.zip](https://github.com/mysociety/fixmystreet/files/14020670/Mobile.zip)

[Desktop.zip](https://github.com/mysociety/fixmystreet/files/14020671/Desktop.zip)

[Skip changelog]

### DIFF
--- a/web/cobrands/bathnes/_colours.scss
+++ b/web/cobrands/bathnes/_colours.scss
@@ -43,6 +43,8 @@ $col_fixed_label: $bathnes-secondary;
 $front-main-color: $primary_b;
 $front-main-color-desktop: $primary_b;
 $postcodeform-background: #fff;
+$postcodeform-color: $primary_b;
+$postcodeform-color-desktop: $primary_b;
 
 $menu-image: 'menu-black';
 $header-top-border: false;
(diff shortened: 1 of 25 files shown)


lizettal:
@neprune do you think you'll get a chance to review this or should I reassign the review to Chris for when he comes onto client support?

PR #4805, opened 2024-01-23
[TfL] Remove FMS log in options for TfL

MorayMySoc:
TfL has requested staff can only log in via single sign on.

Create a call hook and add the call to TfL to
disallow logging in on an email address not supplied for single sign on.

https://github.com/mysociety/societyworks/issues/4100

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth.pm
@@ -109,7 +109,7 @@ sub sign_in : Private {
     $c->logout();
 
     my $parsed = FixMyStreet::SMS->parse_username($username);
-
+    $c->cobrand->call_hook('disable_login_for_email', $parsed->{username}) unless $c->stash->{oauth_need_email};
     $c->forward('throttle_username', [$parsed]);
 
     if ($parsed->{username} && $password && $c->forward('authenticate', [ $parsed->{type}, $parsed->{username}, $password ])) {
@@ -192,6 +192,8 @@ sub email_sign_in : Private {
         $c->stash->{username_error} = $raw_email ? $email_checker->details : 'missing_email';
         return;
     }
+
+    $c->cobrand->call_hook('disable_login_for_email', $good_email) unless $c->get_param('oauth_need_email');
 
     my $password = $c->get_param('password_register');
     if ($password) {
--- a/t/app/controller/auth_social.t
+++ b/t/app/controller/auth_social.t
@@ -15,6 +15,9 @@ my $mech = FixMyStreet::TestMech->new;
 FixMyStreet::App->log->disable('info');
 END { FixMyStreet::App->log->enable('info'); }
 
+#my $tfl_mock = Test::MockModule->new('FixMyStreet::Cobrand::TfL');
+#$tfl_mock->mock('must_have_2fa', sub { 0 });
+
 my $body = $mech->create_body_ok(2504, 'Westminster City Council');
 my $body2 = $mech->create_body_ok(2508, 'Hackney Council');
 my $body3 = $mech->create_body_ok(2488, 'Brent Council', {}, { cobrand => 'brent' });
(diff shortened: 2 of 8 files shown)


perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -109,7 +109,7 @@ sub sign_in : Private {
     $c->logout();
 
     my $parsed = FixMyStreet::SMS->parse_username($username);
-
+    $c->cobrand->call_hook('disable_login_for_email', $username) unless $c->stash->{oauth_need_email};
```
dracos, line 112: Should use $parsed->{username} here.

perllib/FixMyStreet/App/Controller/Auth.pm
```diff
@@ -239,6 +239,7 @@ sub email_sign_in : Private {
 
     $c->stash->{token} = $token_obj->token;
     my $template = $c->stash->{email_template} || 'login.txt';
+    $c->cobrand->call_hook('disable_login_for_email', $good_email) unless $c->get_param('oauth_need_email');
```
dracos, line 242: I think this should be higher, why generate the token etc if going to bail out with this line? Perhaps as soon as good_email is fetched? Unless there's a reason it's here.

dracos (changes requested):

dracos (approved):
Thanks!

t/cobrand/tfl.t
```diff
@@ -264,6 +267,7 @@ subtest "creating a user on TfL creates tfl_password extra_metadata" => sub {
         my $tfl_user = FixMyStreet::DB->resultset("User")->find({ email => $email});
         ok $tfl_user->get_extra_metadata('tfl_password'), "TfL encrypted password created";
         $mech->log_out_ok();
+        undef $tfl_mock;
```
dracos, line 270: Doesn't hurt, but just to note I think the mock should delete itself when it goes out of scope.

t/cobrand/tfl.t
```diff
@@ -310,6 +314,8 @@ subtest "test report creation anonymously by button" => sub {
 };
 
 subtest "test users with tfl email not asked to update their FMS password" => sub {
+    my $tfl_mock = Test::MockModule->new('FixMyStreet::Cobrand::TfL');
```
dracos, line 317: Do we need this subtest at all any more? If they can't log in this method, this can never happen, so no need to mock it to test it.
(Presumably it should test someone logging in via OIDC doesn't get asked to update their password instead, if that doesn't already get done?)

t/cobrand/tfl.t
```diff
@@ -425,7 +472,28 @@ subtest "test report creation and reference number" => sub {
     is $report->name, 'Joe Bloggs';
 };
 
+subtest "test user with tfl email address can't login by updating a report" => sub {
+    $mech->log_out_ok;
+    my $report = FixMyStreet::DB->resultset("Problem")->find({ title => 'Test Report 1'});
+    $mech->get_ok('/report/' . $report->id);
+    $mech->submit_form_ok({
```
dracos, line 479: Not sure why this is two submissions? It's only one form.

t/cobrand/tfl.t
```diff
@@ -425,7 +472,28 @@ subtest "test report creation and reference number" => sub {
...
+        with_fields => {
+            update => 'This is an update',
+        },
+    });
+    $mech->submit_form(
+        form_name => 'updateForm',
+        fields => {
+            username_register => $tfl_staff->email,
+            password_sign_in => $tfl_staff->password,
```
dracos, line 488: Password is blank here

dracos (changes requested):
Few bits :)

t/app/controller/auth_social.t
```diff
@@ -15,6 +15,9 @@ my $mech = FixMyStreet::TestMech->new;
 FixMyStreet::App->log->disable('info');
 END { FixMyStreet::App->log->enable('info'); }
 
+my $tfl_mock = Test::MockModule->new('FixMyStreet::Cobrand::TfL');
```
dracos: I don't understand what this mock is for - the tests pass with it not present?

dracos (commented):
Looks good, one query about why a mock was added is all

PR #4806, opened 2024-01-23
[TfL] Users with TfL email not to be prompted to change password

MorayMySoc:
TfL Staff will be logging in by OIDC only so shouldn't be forced to update their FMS password

https://github.com/mysociety/societyworks/issues/4105

[skip changelog]

### DIFF
--- a/t/cobrand/tfl.t
+++ b/t/cobrand/tfl.t
@@ -31,6 +31,7 @@ FixMyStreet::DB->resultset('BodyArea')->find_or_create({
 });
 my $superuser = $mech->create_user_ok('superuser@example.com', name => 'Super User', is_superuser => 1);
 my $staffuser = $mech->create_user_ok('counciluser@example.com', name => 'Council User', from_body => $body, password => 'password');
+
 $staffuser->user_body_permissions->create({
     body => $body,
     permission_type => 'contribute_as_body',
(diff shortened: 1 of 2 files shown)


MorayMySoc:
Hopefully it's an obvious thing to reviewers, but the log in for the tests isn't working for me - something to do with tfl having an extra data password field - though seems to log in ok later on in another test. I've tried setting extra metadata, but either it's something else or I did that wrong. Pointers appreciated.

t/cobrand/tfl.t
```diff
@@ -296,6 +298,34 @@ subtest "test report creation anonymously by button" => sub {
...
+        "sign in using form"
+    );
+    FixMyStreet->test_mode(1);
+    $tfl_staff->set_extra_metadata('last_password_change', time());
+    is $mech->uri->path, '/my', "logged in to My Account page";
+    $mech->content_lacks('Your password has expired, please create a new one below');
+};
+
+subtest "test users without tfl password asked to update their FMS password" => sub {
+    $tfl_staff->set_extra_metadata('last_password_change', time() - 200000);
```
dracos, line 316: You're not using tfl_staff here so no need to change and reset its last_password_change in this test.

dracos (changes requested):

dracos (approved):
📍

PR #4811, opened 2024-01-25
[FMS] Add grouping to dashboard output

MorayMySoc:
Dashboard should collect categories under groups if they are in a group.

Currently, as we aren't distinguishing between which group a category is in, filters out categories in multiple groups into a multiple category.

Tests added and confirmed can go ahead before https://github.com/mysociety/societyworks/issues/4039 is implemented

https://github.com/mysociety/societyworks/issues/1458

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -9,6 +9,8 @@ use Path::Tiny;
 use Time::Piece;
 use FixMyStreet::DateRange;
 use FixMyStreet::Reporting;
+use List::Util qw(uniq);
+use List::MoreUtils qw(part);
 
 BEGIN { extends 'Catalyst::Controller'; }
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -289,6 +289,29 @@ sub generate_grouped_data : Private {
...
+                    if (!$category_to_group{$category->category}) {
+                        $category_to_group{$category->category} = $group->{name};
+                    } else {
+                        $category_to_group{$category->category} = 'Multiple';
+                    }
+                }
+            }
+        };
+
+        my @multiple_groups = grep { $category_to_group{$_} eq 'Multiple'} @rows;
```
dracos, line 308: And here, could use List::MoreUtils' part to have `my ($non_multiple, $multiple) = part { $category_to_group{$_} eq 'Multiple' } @rows;`

So this section becomes something like:
```
my ($non_multiple, $multiple) = part { $category_to_group{$_} eq 'Multiple' } @rows;
my @multiple = sort(uniq(@$multiple));
@rows = (@$non_multiple, @multiple);
```

templates/web/base/dashboard/index.html
```diff
@@ -159,12 +159,12 @@ <h1>[% loc('Summary statistics') %]</h1>
     [% IF category_to_group.$k_safe AND category_to_group.$k_safe != current_category %]
     [% SET current_category = category_to_group.$k_safe %]
       <tr>
-        <th scope="row">[% category_to_group.$k_safe %]</th>
-        <th scope="row">-</th>
+        <th colspan="(2 or 5)" scope="colgroup">[% category_to_group.$k_safe %]</th>
```
dracos, line 162: Sorry I wasn't clear enough here - by `2 or 5` I meant you had to put 5 if group_by was category+state (because it has 3 extra columns of output in the table), or 2 otherwise - it's the number of columns that are spanned. And by "only need the one entry" I meant then you wouldn't have any of the `<th>-</th>` because this column is spanning them.

dracos (changes requested):
One issue still with the HTML output, but think that's it :)

dracos (approved):
Yep, looks good :)

PR #4812, opened 2024-01-25
[TfL] Start pod documentation

MorayMySoc:
Beginning of pod documentation for TfL Cobrand

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/TfL.pm
+++ b/perllib/FixMyStreet/Cobrand/TfL.pm
@@ -1,3 +1,13 @@
+=head1 NAME
+
+FixMyStreet::Cobrand::TfL - code specific to the Transport for London cobrand
+
+=head1 SYNOPSIS
+
+Transport for London is an FMS integration. It covers multiple London Borough areas
+
+=cut
+
 package FixMyStreet::Cobrand::TfL;
 use parent 'FixMyStreet::Cobrand::Whitelabel';
 
@@ -6,6 +16,15 @@ use warnings;
 use utf8;
 
 use Moo;
+
+=head1 DESCRIPTION
+
+TfL has features related to covering multiple London areas. It doesn't have an external
+integration with a backend, but does have single sign-on with Azure for staff users - they
+are assigned an FMS account and user role when logging in.
+
+=cut
+
 with 'FixMyStreet::Roles::BoroughEmails';
 
 use POSIX qw(strcoll);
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
Thanks for looking at this.
One suggestion - for the oneliners that are sort of configuration, some other cobrand files use a `=over 4` and interleaved `=item` rather than headings for each one, which might make sense for some of these? Can reorder the functions if that makes more sense too (either for short ones, or to group them more thematically if anything could be better elsewhere).

MorayMySoc:
I've tidied up pretty much where I was up to so hopefully tidier, but not complete.

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -86,14 +171,153 @@ sub about_hook {
...
+=cut
+
+sub london_boroughs { (
+    2511, 2489, 2494, 2488, 2482, 2505, 2512, 2481, 2484, 2495,
+    2493, 2508, 2502, 2509, 2487, 2485, 2486, 2483, 2507, 2503,
+    2480, 2490, 2492, 2500, 2510, 2497, 2499, 2491, 2498, 2506,
+    2496, 2501, 2504,
+) }
+
+=item * surrounding_boroughs
```
dracos: is called surrounding_london (as they're not boroughs surrounding)

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -148,6 +389,13 @@ sub problems_sql_restriction {
     return $q;
 }
 
+=item * inactive_reports_filter
+
+Reports are deleted from TfL after 7 years if they are not safety critical or after 3 years if they are
+This is initiated by an annual cron job
```
dracos: It runs more than annually :)

perllib/FixMyStreet/Cobrand/TfL.pm
```diff
@@ -148,6 +389,13 @@ sub problems_sql_restriction {
     return $q;
 }
 
+=item * inactive_reports_filter
+
+Reports are deleted from TfL after 7 years if they are not safety critical or after 3 years if they are
```
dracos: The reverse - safety ones are kept for 7

dracos (commented):
A number of suggestions, but I made them all whilst going through it, so if you'd like to look over and check you agree :)

MorayMySoc:
Ok - thank you! Much more active voice too :-) I've added an extra fix-up that highlights account is created and role is assigned (which was the bit I missed out altogether), which I hope makes sense. But will try to follow this pattern with pod then

PR #4815, opened 2024-01-26
Fms add tilma mock to auth social

MorayMySoc:
Add mySociety::Locale to Mock::Tilma to make sure it's loaded.

Use Mock::Tilma in test

[skip changelog]

### DIFF
--- a/t/app/controller/auth_social.t
+++ b/t/app/controller/auth_social.t
@@ -6,6 +6,7 @@ use JSON::MaybeXS;
 use t::Mock::Facebook;
 use t::Mock::Twitter;
 use t::Mock::OpenIDConnect;
+use t::Mock::Tilma;
 
 use FixMyStreet::TestMech;
 my $mech = FixMyStreet::TestMech->new;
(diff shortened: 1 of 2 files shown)


t/app/controller/auth_social.t
```diff
@@ -45,6 +46,8 @@ foreach ($body->id, $body2->id, $body3->id) {
 my $resolver = Test::MockModule->new('Email::Valid');
 my $social = Test::MockModule->new('FixMyStreet::App::Controller::Auth::Social');
 $social->mock('generate_nonce', sub { 'MyAwesomeRandomValue' });
+my $tilma = t::Mock::Tilma->new;
+LWP::Protocol::PSGI->register($tilma->to_psgi_app, host => 'tilma.mysociety.org');
```
dracos, line 50: As discussed in slack, "with two lines, for live and staging" - the staging line is missing here, but it is being called by Brent.pm because STAGING_SITE is true.

dracos (changes requested):
Tiny thing on the first commit message - it's not really a fallback; it's used directly in the file, so always should have been present, because you don't know if anything else will be importing it or not. (Also missing the '::'.)
Doesn't mock out all the calls being made, as discussed.

t/app/controller/auth_social.t
MorayMySoc: So can I just add another host? By repeating the line but with staging.tilma...?

dracos: Yes

dracos (approved):

PR #4816, opened 2024-01-26
[Bexley][Ww] Adds tests and minor changes for postcode finder

MorayMySoc:
Change postcode finder code so can receive a single hash rather than an array of hashes if only one result.

Remove bullet points from front page that don't apply to Bexley.

Add testing file and testing for the above.

[skip changelog]

### DIFF
--- a/perllib/Integrations/Whitespace.pm
+++ b/perllib/Integrations/Whitespace.pm
@@ -84,7 +84,12 @@ sub GetAddresses {
 
     my $res = $self->call('GetAddresses', getAddressInput => ixhash( Postcode => $postcode ));
 
-    return $res->{Addresses}->{Address};
+    my $addresses = $res->{Addresses}->{Address};
+    # Spec - Results can be returned as an array if more than one result is found.
+    # So presuming one address may not be an array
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -16,6 +16,10 @@ sub bin_addresses_for_postcode {
 
     my $addresses = $self->whitespace->GetAddresses($postcode);
 
+    # Spec - Results can be returned as an array if more than one result is found.
+    # So presuming one address may not be an array
+    $addresses = [ $addresses ] unless ref $addresses eq 'ARRAY';
```
chrismytton, line 21: I wonder if this might be better done in `perllib/Integrations/Whitespace.pm`, similar to the existing `GetStreets` method? That way any future callers of this method can get a consistent return value.

chrismytton (approved):
Looks good! Just one suggestion about possibly moving some code around.

chrismytton (approved):
Looks good!

PR #4817, opened 2024-01-29
More accurate error message for incorrect password during reporting

davea:
The previous error referred to a section of the form that isn't there any more:

image

So now it's clearer:

image

[skip changelog]


davea:
I couldn't come up with a wording that was clear enough for both very different situations, so have gone for having two separate errors.

dracos (approved):
Thanks, nice solution :)

PR #4819, opened 2024-01-29
[Waste] Improve layout in bulky bin day section.

dracos:
Make sure buttons aren't buttressing up against lists, and spaced out, fixing some strange nested `div`s inside `dl`. [skip changelog]
Best viewed with whitespace ignored: https://github.com/mysociety/fixmystreet/pull/4819/files?w=1

Before
image

After
Screenshot 2024-01-29 at 13 23 56

### DIFF
--- a/templates/web/base/waste/bulky/_bin_days_list.html
+++ b/templates/web/base/waste/bulky/_bin_days_list.html
@@ -2,24 +2,24 @@
     [% PROCESS waste/_bulky_not_signed_user.html %]
 [% END %]
 
-<dl class="govuk-summary-list govuk-!-margin-bottom-0">
-
-  [% FOR booking IN pending_bulky_collections %]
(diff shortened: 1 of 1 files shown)


lucascumsille:
It works well for me and we have the same markup/nesting that GOVUK. The only thing that I think we could improve on mobile, is that there is a lot of space between the button and the last line, so it might give the impression at first glance that they are not related.

One solution would be to add to the last `govuk-summary-list` this extra classes `govuk-!-margin-bottom-3 govuk-!-padding-bottom-0`

### Currently
Screenshot 2024-01-29 at 15 52 45

### Mobile
Screenshot 2024-01-29 at 15 58 09

### Desktop
Screenshot 2024-01-29 at 15 58 41

lucascumsille:
@dracos forgot to tag you in the comment above

dracos:
Thanks for looking, will make that change :)

PR #4820, opened 2024-01-29
BEX005 Bexley WW: Service filters & date munging

nephila-nacrea:
Done as part of https://github.com/mysociety/societyworks/issues/4093.

- Filter out services based on SiteServiceValidFrom/-To
- Prettily display next collection dates

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley.pm
@@ -365,5 +365,4 @@ sub dashboard_export_problems_add_columns {
     });
 }
 
-
 1;
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -3,6 +3,9 @@ package FixMyStreet::Cobrand::Bexley::Waste;
 use Moo::Role;
 
 use Integrations::Whitespace;
+use DateTime;
+use DateTime::Format::W3CDTF;
+use FixMyStreet;
 use FixMyStreet::Template;
 use Sort::Key::Natural qw(natkeysort_inplace);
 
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -365,5 +365,8 @@ sub dashboard_export_problems_add_columns {
     });
 }
 
+# Waste
+
+sub bin_day_format { '%A, %-d~~~ %B %Y' }
```
chrismytton, line 370: I think it might make sense to put this in `perllib/FixMyStreet/Cobrand/Bexley/Waste.pm`, with the other WasteWorks related code, what do you think?

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -55,23 +58,71 @@ sub bin_services_for_address {
...
-                date => $_->{NextCollectionDate},
-                ordinal => 'th', # TODO!
-                changed => 0,
-            },
+    my @site_services_filtered;
+
+    my $now_dt = DateTime->now->set_time_zone( FixMyStreet->local_time_zone );
+    for my $service (@$site_services) {
+        # TODO Do we need to filter out services where NextCollectionDate is
+        # in the past?
```
chrismytton, line 66: From what I've seen these dates are either in the future or blank, so don't think we need to worry about dates in the past at this point.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -51,23 +58,71 @@ sub bin_services_for_address {
...
+    my $services = [
+        map {
+            my $next_dt = eval {
+                DateTime::Format::W3CDTF->parse_datetime(
+                    $_->{NextCollectionDate} );
+            };
+            if ($@) {
+                warn $@;
+                next;
+            }
```
chrismytton, line 104: This `next` won't work as expected inside the map call.

You can see the problem by adding an extra service in the tests (inside `_site_collections`). If you create a service 7 by copying service 6 and give it an invalid `NextCollectionDate` then you should get an error about "Exiting subroutine via next".

Think the fix is to ignore services with an invalid `NextCollectionDate` before this map call.

chrismytton (changes requested):
Looking good so far! Couple of minor comments below.

Other thing I noticed is the list of services is showing a blank "Frequency" field, see below.

Bexley aren't expecting this field, so think you can just wrap that bit of the `bin_days.html` template with an `[% IF unit.schedule %]` so it doesn't show if there's no schedule defined for a service.

image

chrismytton:
Oh an Moray has merged https://github.com/mysociety/fixmystreet/pull/4816 now, so you'll need to rebase this onto the `bexley-wasteworks-whitespace-integration` branch and then change the pull request to target that branch.

perllib/FixMyStreet/Cobrand/Bexley.pm
nephila-nacrea: Yep! Think I got muddled because Peterborough have everything in one file.

chrismytton (approved):
Looks good! ✨

PR #4827, opened 2024-02-02
[Bexley WW] Container images

nephila-nacrea:
Adds images for waste containers.

Closes https://github.com/mysociety/societyworks/issues/4067.

[skip changelog]


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -107,6 +107,82 @@ sub bin_services_for_address {
     return \@site_services_filtered;
 }
```
nephila-nacrea, line 109: TODO: Check that these commented-out services are not used, remove those lines.

chrismytton (changes requested):
Looking good so far!

I think since we don't have 2x images it might make sense to just get rid of the `@2x` versions, since they identical to the regular images.

Then if you update the bin days template so it doesn't output the `srcset=` attribute for Bexley the images will be rendered at their "natural" 150px size, which matches what Bexley have on their [existing waste site](https://mybexley.bexley.gov.uk/service/When_is_my_collection_day).

chrismytton (approved):
✨ 🗑️ ✨

PR #4828, opened 2024-02-04
[Bucks] Display last cleaned date for drains in sidebar.

neprune:
[skip changelog] [FD-3731](https://mysocietysupport.freshdesk.com/a/tickets/3731)


dracos (approved):
Simpler :) Would have been good in the PR if you could have linked to the relevant commits in the server configuration, I found them and worked out where the data was coming from.

PR #4829, opened 2024-02-04
[Brent] Auto update garden sub renewal end year.

neprune:
[skip changelog] [FD-3746](https://mysocietysupport.freshdesk.com/a/tickets/3746)


dracos (approved):
Thanks, hopefully will prevent this next year :)

PR #4834, opened 2024-02-06
[Bexley] WasteWorks branding and styles

chrismytton:
Add WasteWorks navigation and stylesheet, along with the "Powered by SocietyWorks" footer.

Fixes https://github.com/mysociety/societyworks/issues/4118

### DIFF
--- a/web/cobrands/bexley/base.scss
+++ b/web/cobrands/bexley/base.scss
@@ -2,6 +2,7 @@
 @import "./_colours";
 @import "../sass/mixins";
 @import "../sass/base";
+@import "../sass/waste";
 
 div.form-error, p.form-error { // TODO Add to more, make a variable.
     background-color: $red;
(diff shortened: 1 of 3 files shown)


web/cobrands/bexley/base.scss
```diff
@@ -3,6 +3,7 @@
 @import "../sass/mixins";
 @import "../sass/base";
 @import "../sass/waste";
+@import "../fixmystreet-uk-councils/societyworks-footer";
```
nephila-nacrea, line 6: Was going to ask, does this not affect the standard Bexley site? But it looks like the live site does not have a footer, when it should...(?)

nephila-nacrea (approved):

web/cobrands/bexley/base.scss
chrismytton: Yeah it also adds the footer to FMS, as it was missing from there.

PR #4836, opened 2024-02-07
[Whitespace] Use force_arrayref function.

dracos:
[skip changelog]


chrismytton (approved):
Thanks! I think I'd seen this method referenced before, but hadn't quite realised this was what it was for 🙂 

(There's also another place it could be used in `GetAddresses`)

nephila-nacrea (commented):
Thanks, this is very useful 🙂

PR #4837, opened 2024-02-07
Bexley WW: Order waste services by earliest upcoming collection

nephila-nacrea:
Order Bexley waste services by earliest upcoming collection (i.e. next collection date).

Closes https://github.com/mysociety/societyworks/issues/4096.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -104,7 +104,20 @@ sub bin_services_for_address {
         };
     }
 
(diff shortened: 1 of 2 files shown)


nephila-nacrea:
From what I can tell so far, we should just be able to sort services at the end of the `bin_services_for_address` method - I have added a `service_sort` method to Bexley/Waste.pm, a similar method could be implemented for other cobrands.

chrismytton (changes requested):
Code looks good! Would be good if we could have a test for this, just so it doesn't accidentally get broken in the future. Sorry to be a pain!

chrismytton (approved):

PR #4843, opened 2024-02-12
[West Northants] Add cookie banner and google analytics.

neprune:
[skip changelog] closes https://github.com/mysociety/societyworks/issues/4065


davea (changes requested):
Can the CSS go into its own file a bit like we do for e.g. Shropshire? ([`_cookies.scss`](https://github.com/mysociety/fixmystreet/blob/master/web/cobrands/shropshire/_cookies.scss), then [included in `base.scss`](https://github.com/mysociety/fixmystreet/blob/master/web/cobrands/shropshire/base.scss#L6)) No need for it to be transferred with every page load.

davea (approved):
🍪

PR #4847, opened 2024-02-14
[Bexley][WW] Pull collections timetable into iCal/.ics file

nephila-nacrea:
Enable users to download a .ics file containing their bin collection timetable for the year.

Closes https://github.com/mysociety/societyworks/issues/4097.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -118,10 +118,11 @@ sub bin_services_for_address {
         }
 
         push @site_services_filtered, {
-            id           => $service->{SiteServiceID},
-            service_id   => $service->{ServiceItemName},
-            service_name => $service->{ServiceItemDescription},
-            next         => {
+            id             => $service->{SiteServiceID},
+            service_id     => $service->{ServiceItemName},
+            service_name   => $service->{ServiceItemDescription},
+            round_schedule => $service->{RoundSchedule},
+            next           => {
                 date    => $service->{NextCollectionDate},
                 ordinal => ordinal( $next_dt->day ),
                 changed => 0,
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -93,10 +93,11 @@ sub bin_services_for_address {
         }
 
         push @site_services_filtered, {
-            id           => $service->{SiteServiceID},
-            service_id   => $service->{ServiceItemName},
-            service_name => $service->{ServiceItemDescription},
-            next         => {
+            id             => $service->{SiteServiceID},
+            service_id     => $service->{ServiceItemName},
+            service_name   => $service->{ServiceItemDescription},
```
MorayMySoc, line 98: I think it might be better to use $self->{c}->stash->{containers} to give this a nice name - would just mean implementing it basically for now, but would mean when it's updated this would have the front facing names

MorayMySoc (changes requested):
All seems to work well for me - only one comment

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
nephila-nacrea: I am not entirely sure what you are asking here - could be because I am unclear what we are doing about `stash->{containers}` (I will need to ask on Slack, as I thought we had a ticket to handle {containers}, but it doesn't seem to exist. Currently it's just empty, marked with a '# TODO' - possibly because we don't have a hardcoded list of services (unlike what I can tell from other cobrands) but have to check dates on the fly).

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -187,4 +232,59 @@ sub ordinal {
     $irregulars{$n % 100} || $irregulars{$n % 10} || 'th';
 }
 
+sub _containers {
+    my ( $self, $property ) = @_;
+
+    my $is_communal
```
nephila-nacrea, line 238: This will apply in https://github.com/mysociety/fixmystreet/pull/4850.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
nephila-nacrea: Understand what you mean now; have added a commit for containers :-)

MorayMySoc: I saw you talking to Chris, an he answered more clearly than I would have managed!

MorayMySoc (approved):
Great - thanks

PR #4848, opened 2024-02-15
[North Northants] Added Frontend

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4115

- The first commit might need to be dropped. I erased those file because I wasn't sure if they are needed.

### Desktop

Screenshot 2024-02-15 at 14 41 09
Screenshot 2024-02-15 at 14 41 01
Screenshot 2024-02-15 at 14 40 55
Screenshot 2024-02-15 at 14 40 47
Screenshot 2024-02-15 at 14 40 38

### Mobile

Screenshot 2024-02-15 at 14 42 03
Screenshot 2024-02-15 at 14 41 45

[skip changelog]


davea (changes requested):
You're right, the first commit should be dropped as we do need those templates (the first one ensures that the correct asset layers/JS are loaded for the map/reporting process, and the second has some custom wording for the front page that we presumably want to keep). Apart from that, the second commit looks good!

lucascumsille:
Thanks @davea I just dropped it should be fine now

PR #4850, opened 2024-02-16
[Bexley][WW] Handle differences in behaviour between parent and child properties

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/4114.

- [x] Child properties do not (usually) have their own services, so fetch for the parent instead
- [] For 'PG-240 | Plastics & glass 240 ltr wheeled bin', use a different image when at parent property
- [] For 'RES-240 | Residual 240 ltr bin', use 'Communal Refuse Bin(s)' when at a parent property

[skip changelog]


chrismytton (approved):
Awesome, great to get the parent/child properties figured out! 🏡

PR #4856, opened 2024-02-22
[Bexley][WW] Allow reporting missed collections

nephila-nacrea:
Adds the necessary checks, templates and hooks to allow reporting missed collections to Whitespace.

Handles the gate checks that need to be done before the report can be submitted:

- Check for existing open reports
- Check that the collection happened in the past 3 working days
- Check the in-cab logs for red tags and service updates

I've also added a superuser-only section at the bottom of the services screen to allow superusers to view the "raw" in-cab logs (red tags and service exceptions), as I think that might be useful in the future.

Closes https://github.com/mysociety/societyworks/issues/4094.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -75,9 +75,23 @@ sub bin_services_for_address {
         if !@{ $site_services // [] }
         && $property->{parent_property};
 
+    # TODO Call these in parallel
+    $property->{missed_collection_reports}
+        = $self->_missed_collection_reports($property);
+    $property->{recent_collections} = $self->_recent_collections($property);
+    my ($property_logs, $street_logs) = $self->_in_cab_logs($property);
+
+    $property->{red_tags} = $property_logs;
+    $property->{service_updates} = $street_logs;
+    my %round_exceptions = map { $_->{round} => 1 } @$property_logs;
+    $property->{round_exceptions} = \%round_exceptions;
+
+    # Set certain things outside of services loop
+    my $containers = $self->_containers($property);
+    my $now_dt = DateTime->now->set_time_zone( FixMyStreet->local_time_zone );
+
     my @site_services_filtered;
 
-    my $now_dt = DateTime->now->set_time_zone( FixMyStreet->local_time_zone );
     for my $service (@$site_services) {
         next if !$service->{NextCollectionDate};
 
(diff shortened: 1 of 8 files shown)


templates/web/bexley/waste/services_extra.html
```diff
@@ -0,0 +1,38 @@
+<div class="govuk-grid-row" id="in-cab-logs">
+  [% IF c.user.is_superuser %]
```
dracos, line 2: Who is this block for? Is it for helping debug any issues?

perllib/Integrations/Whitespace.pm
```diff
@@ -125,11 +124,43 @@ sub GetSiteServiceItemRoundSchedules {
...
+
+    my $res = $self->call(
+        'GetCollectionByUprnAndDate',
+        getCollectionByUprnAndDateInput => ixhash(
+            Uprn                   => $uprn,
+            NextCollectionFromDate => $date_from,
+        ),
+    );
+
+    # TODO Need to force arrayref?
```
dracos, line 162: Will there ever be more than one?

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -113,26 +127,179 @@ sub bin_services_for_address {
...
+        : $property->{uprn};
+
+    my $worksheets = $self->whitespace->GetSiteWorksheets($uprn);
+
+    my %missed_collection_reports;
+    for my $ws (@$worksheets) {
+        if (   $ws->{WorksheetStatusName} eq 'Open'
+            && $ws->{WorksheetSubject} =~ /^Missed/ )
+        {
+            for ( @{  $self->whitespace->GetWorksheetDetailServiceItems(
```
dracos, line 190: Can this be called with more than one ID? Guessing not, but would be fewer calls if it can.

dracos (changes requested):
Nothing really, I've assumed this has been tested against the API data, I haven't had the time to set it up locally I'm afraid. I assume this will all be squashed together given the changes, but looks good overall. Does add a few API calls, but can of course look at performance if needed.
I see the 'last collection' is also added as part of this which I don't think was there before, that might be worth noting in the commit message as well.

templates/web/bexley/waste/services_extra.html
chrismytton: Yeah it's for me/us to do debugging. I've added a comment to this template to make it's purpose clear.

perllib/Integrations/Whitespace.pm
chrismytton: Yes, Whitespace tends to send back an array when there's more than one, but just the single item when there's only one, so probably worth forcing it here, have changed that in the fixup commit.

chrismytton:
> I assume this will all be squashed together given the changes

Yep will squash it all together once it's been reviewed.

> I see the 'last collection' is also added as part of this which I don't think was there before, that might be worth noting in the commit message as well.

Have amended the commit message (but made the requested changes in a separate fixup commit).

perllib/Integrations/Whitespace.pm
dracos: You've changed the Plus function but not this one, if that matters :)

dracos (approved):
Looks good, if you want to do the same to the other GetCollection function as well

PR #4858, opened 2024-02-26
[CBeds] Stop road layer triggering for PROW

MorayMySoc:
Rights of way has to be excluded from the not on a road stopper or reports can't be made.

Map and server side in .com and society-servers with same branch name

https://mysocietysupport.freshdesk.com/a/tickets/3760

[skip changelog]


dracos (changes requested):
The comment presumably needs changing to match?

In the server change, there aren't any "R Byway" entries I can see, if those are the right values to use.

dracos (approved):

PR #4863, opened 2024-03-04
4099 bexley add bin calendar links

MorayMySoc:
If the service is the waste bin use that to work out whether the household is on rotation week one or rotation week two for PDF calendars and present the correct link.

Also assume that if a calendar_link or ggw_calendar_link is stashed then a PDF download link should be shown.

https://github.com/mysociety/societyworks/issues/4099

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -114,7 +114,12 @@ sub bin_services_for_address {
         }
 
         my $containers = $self->_containers($property);
-
+        if (!$self->{c}->stash->{calendar_link} && $service->{RoundSchedule} =~ /(Wk \d+)/) {
+            my ($id) = $service->{RoundSchedule} =~ /(Wk \d+)/;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -114,7 +114,12 @@ sub bin_services_for_address {
         }
 
         my $containers = $self->_containers($property);
-
+        if ($service->{ServiceID} && $service->{ServiceID} == 1) {
```
chrismytton, line 117: I don't think all sites will have a service with `ServiceID` == 1 (e.g. try `/waste/58247/` locally, seems that because it's a communal property there's no residual service).

Not sure what the best approach is here, I guess you could change this to "if the round schedule contains a week then update the calendar link to use that week" and just run it for all services. Or just take the first service that matches the pattern and use the week from that, and skip the check if the calendar link is already set?

chrismytton (changes requested):

chrismytton (approved):

PR #4864, opened 2024-03-05
[FMS] Add Swedish and German language packs to install

MorayMySoc:
Replace lost code from https://github.com/mysociety/fixmystreet/pull/3546

https://github.com/mysociety/fixmystreet/issues/3544


dracos (commented):
It looks like I took them out in 96b529d13d6e16e410d0c536ad2eaa79202151e1 because those packages are Ubuntu only and so it was breaking install on Debian. Not sure if there's an easy workaround.

MorayMySoc:
Is it worth in some way having a packages.ubuntu or something like that?

PR #4867, opened 2024-03-06
[Bexley WasteWorks] Missed collections enhancements

chrismytton:
- Show "Being collected today" when a collection is due today
- Add a "where are your bins located" question to the missed collection form

Part of https://github.com/mysociety/societyworks/issues/4051

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -577,4 +577,18 @@ sub waste_munge_report_data {
     $c->set_param('service_id', $id);
 }
 
+sub waste_munge_report_form_fields {
+    my ($self, $field_list) = @_;
+
+    push @$field_list, "extra_detail" => {
+        type => 'Text',
+        widget => 'Textarea',
+        label => 'Please supply any additional information such as the location of the bin.',
+        maxlength => 1_000,
+        messages => {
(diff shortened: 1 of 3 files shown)


dracos (commented):
Changes all look fine, couple of general questions:
* Should anything be being done with the `extra_detail`?
* Not saying you have to change this, but the other wastes use a `waste_on_the_day_criteria` function to check behaviour about being on the day, e.g. updating the state so it displays as "Date (In progress)" or similar, not sure if you could use that behaviour here or probably doesn't matter.

chrismytton:
> * Should anything be being done with the `extra_detail`?

Yes, that's being handled in https://github.com/mysociety/fixmystreet/pull/4890.

> * Not saying you have to change this, but the other wastes use a `waste_on_the_day_criteria` function to check behaviour about being on the day, e.g. updating the state so it displays as "Date (In progress)" or similar, not sure if you could use that behaviour here or probably doesn't matter.

Looks like the arguments to `waste_on_the_day_criteria` are quite Echo-specific currently. Whitespace doesn't have a reliable way of knowing if the round is complete or what state it's in. I think since Bexley don't want any reporting on the day of the collection it's probably ok as it is now.

dracos (approved):
Looks good. Sorry, I wasn't clear, I wasn't suggesting you use the same arguments as `waste_on_the_day_criteria`, only wondered if the structure of calling a similarly named function would help them behave more in a similar way, but is fine as is!

PR #4869, opened 2024-03-07
[Bromley] Create 'Investigation Required' events for comments on closed Echo-backed reports.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/3267

Example on staging: https://staging.fixmystreet.com/admin/report_edit/23736

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -213,7 +213,7 @@ sub open311_config {
 }
 
 sub open311_extra_data_include {
-    my ($self, $row, $h) = @_;
+    my ($self, $row, $h, $contact) = @_;
 
     my $title = $row->title;
 
@@ -250,7 +250,7 @@ sub open311_extra_data_include {
           value => $row->user->email }
     ];
 
-    if ( $row->category eq 'Garden Subscription' ) {
+    if ( $contact->category eq 'Garden Subscription' ) {
         if ( $row->get_extra_metadata('contributed_as') && $row->get_extra_metadata('contributed_as') eq 'anonymous_user' ) {
             push @$open311_only, { name => 'contributed_as', value => 'anonymous_user' };
         }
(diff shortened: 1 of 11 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -374,16 +391,42 @@ sub open311_contact_meta_override {
...
+can't update closed echo events.
+
 =cut
 
 sub should_skip_sending_update {
     my ($self, $update) = @_;
 
+    my $report = $update->problem;
+    if ($report->is_closed && $self->_has_report_been_sent_to_echo($report)) {
+        $report->set_extra_metadata('open311_category_override' => 'Investigation Required');
```
dracos, line 422: I think in my phase2 work I called this category "Referred to Veolia Streets". The Echo to Bromley one is called "Referred to LB Bromley Streets".

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -261,6 +270,14 @@ sub open311_pre_send {
         my $text = $row->detail . "\n\nPrivate comments: $private_comments";
         $row->detail($text);
     }
+
+    if (my $comment_id = $row->get_extra_metadata('echo_report_reopened_with_comment')) {
+        my $comment = FixMyStreet::DB->resultset('Comment')->find($comment_id);
+        if ($comment && $comment->text) {
+            my $text = "Closed report has a new comment: " . $comment->text . "\n\n" . $row->detail;
+            $row->detail($text);
```
dracos, line 278: I'm not sure if the new event will easily have access to the original event text so this should probably be an addition to the detail, rather than replacing it?

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: I'll rename it as part of my cominbing with my phase2 commit.

dracos (changes requested):
Looks good. I'm just tidying it up now with my phase2 thing and will then try and put on staging. I think for my phase2 I changed the category rather than use a hidden one as you have here, your way might be preferable depending on how it looks, but can see during testing I guess!

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: Dur, it does, at the end!

dracos (approved):

PR #4875, opened 2024-03-13
Added autocomplete attribute to inputs

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4208

- [x] For [Added autocomplete attribute to email types elements](https://github.com/mysociety/fixmystreet/commit/b09ae82509128631e84df4c7e713b4fdc33eb111) I noticed that some elements instead of using `type="email"` use `type="text"` however they are intended for emails. Is there a reason not to use `type="email"`?

- [x] [PENDING] TFL used this element `input id="form_username_register"` as an example of the problem in [issue 4208](https://github.com/mysociety/societyworks/issues/4208). However, I could only find it on the Zurich cobrand. But there is no trace in the default templates, so I still need to change that instance. Do you know where I can find that piece of code `input id="form_username_register`

NOTES:
CyclingUK and Bromley have first and last names in their forms, so they were tackled in this commit:
[Added autocomplete attribute to first and last name](https://github.com/mysociety/fixmystreet/commit/2665a014b61c03b152421139b9db0dcc34c31255)

[Skip changelog]

### DIFF
--- a/templates/web/base/admin/bodies/contact-form.html
+++ b/templates/web/base/admin/bodies/contact-form.html
@@ -40,7 +40,14 @@
         [% loc('An email address or service ID (Open311 or similar).') %]
       [% END %]
     </p>
-    <input type="text" class="form-control" id="destination" aria-describedby="destination-hint" name="email" size="30" value="[% contact.email | html %]" required>
+    <input 
+      [% IF body.can_be_devolved OR  body.send_method == 'Open311' %]
+        type="text"
+      [% ELSE %]
+        type="email"
+        autocomplete="email"
+      [% END %]
+      class="form-control" id="destination" aria-describedby="destination-hint" name="email" size="30" value="[% contact.email | html %]" required>
 
     [% TRY %][% PROCESS 'admin/bodies/_type_field.html' %][% CATCH file %][% END %]
 
--- a/templates/web/base/admin/bodies/contact-form.html
+++ b/templates/web/base/admin/bodies/contact-form.html
@@ -30,24 +30,20 @@
 
     <label for="destination">[% loc('Destination') %]</label>
     <p class="form-hint" id="destination-hint">
+      [% SET field_type = 'text' %]
       [% IF body.can_be_devolved %]
         [% loc('An email address or service ID (Open311 or similar).') %]
       [% ELSIF body.send_method == 'Open311' %]
         [% loc('A service ID (Open311 or similar).') %]
       [% ELSIF body.send_method.match('Email') OR NOT body.send_method %]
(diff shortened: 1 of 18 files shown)


lucascumsille:
@dracos 
TFL used this element input id="form_username_register" as an example of the problem in https://github.com/mysociety/societyworks/issues/4208. However, I could only find it on the Zurich cobrand. But there is no trace in the default templates, so I still need to change that instance. Do you know where I can find that piece of code input id="form_username_register

dracos:
(If some do and some don't, might as well always use type=email I guess. https://design-system.service.gov.uk/patterns/email-addresses/ says should also set spellcheck=false )

In `templates/web/base/report/form/user_loggedout_by_email.html`, it includes `[% PROCESS 'report/form/user_loggedout_email.html' name='username_register' %]` - and that template `templates/web/base/report/form/user_loggedout_email.html` uses the passed-in name to output, which is why you can't find it directly. Note that that field is sometimes used to accept both email or mobile number, as you can see, so you'll have to be careful to only set it when it's only email.

templates/web/base/admin/bodies/contact-form.html
```diff
@@ -40,7 +40,7 @@
         [% loc('An email address or service ID (Open311 or similar).') %]
       [% END %]
     </p>
-    <input type="text" class="form-control" id="destination" aria-describedby="destination-hint" name="email" size="30" value="[% contact.email | html %]" required>
+    <input type="email" class="form-control" id="destination" aria-describedby="destination-hint" name="email" size="30" value="[% contact.email | html %]" required>
```
dracos, line 43: Ah this one is sometimes not an email, look at the label just above.

lucascumsille: I added a conditional to tackle this.

templates/web/base/admin/bodies/contact-form.html
```diff
@@ -40,7 +40,14 @@
         [% loc('An email address or service ID (Open311 or similar).') %]
       [% END %]
     </p>
-    <input type="text" class="form-control" id="destination" aria-describedby="destination-hint" name="email" size="30" value="[% contact.email | html %]" required>
+    <input 
+      [% IF body.can_be_devolved OR  body.send_method == 'Open311' %]
```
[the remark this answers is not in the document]
lucascumsille: @dracos I added a fixup for this one let me know if it covers all the angles =)

dracos (approved):
I assume a rebase will fix the cypress test failure.
All the autocomplete ones can probably be squashed into one commit, I think?

PR #4879, opened 2024-03-14
Improved contrast for #skiplink element

lucascumsille:
Fixes: [#4210](https://github.com/mysociety/societyworks/issues/4210)

This element will now behave like the `.btn` element in terms of styling. This way we don't need to worry about contrast and we can still use the branding of each cobrand.

Screenshot 2024-03-14 at 09 07 53
Screenshot 2024-03-14 at 09 07 57
Screenshot 2024-03-14 at 09 08 10

PR #4880, opened 2024-03-14
Added skip map button

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4211

This has been added after header elements and before any focusable map elements, that way the user doesn't need to go through the map links and compass to be able to go to skip the map.

- [x] I added the link in the `openlayers.html` file. Assuming that is the right place to put should I also include it on the google one?

Screenshot 2024-03-14 at 10 42 24

IMPORTANT: https://github.com/mysociety/fixmystreet/pull/4879 should be merged so the styling of the button is contrast accessible.

### DIFF
--- a/templates/web/base/maps/openlayers.html
+++ b/templates/web/base/maps/openlayers.html
@@ -42,6 +42,7 @@
 <div id="map_box">
     [% pre_map %]
     <div id="map">
+      <a href="#map_sidebar" class="skiplink">Skip map</a>
       [% IF noscript_map_template == 'maps/noscript_map_base_wmx.html' %]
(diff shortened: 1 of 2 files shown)


neprune (approved):

dracos:
I'm afraid that this PR's Cypress tests never passed. Please don't merge things to master if the tests don't all pass, because this could mean something has broken. This also does not include translation of a new base string in English.

lucascumsille:
@dracos Sorry my bad, I didn't notice the Cypress test not passing. I have added a fix for the translation bit [here](https://github.com/mysociety/fixmystreet/pull/4911).
I'll check what you did in Hackney now.

PR #4881, opened 2024-03-14
"Can't use the map" Link always visible.

lucascumsille:
Fixes part of: https://github.com/mysociety/societyworks/issues/4211

This improvement was advised as an "ideal" point during the TFL audit. Having it always visible can help reduce the cognitive load for users with cognitive difficulties by easing decision-making. I tried to make it visible enough, but at the same time, it doesn't use much vertical space. To avoid/reduce potential issues with the cobrands, I used the `$primary` variable instead of the yellow we were using.

Screenshot 2024-03-14 at 11 41 42

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -2669,7 +2669,7 @@ label .muted {
 }
 
 .map-alternatives, .map-alternatives__display {
-  margin: 16px 0;
+  margin: 0;
   font-size: 14px;
   color: #666;
 }
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -392,12 +392,11 @@ body.mappage.admin {
 .map-alternatives {
   display: block;
   color: inherit;
-  //margin: 0 -15px;
+  margin: 0 -1rem;
   padding: 16px;
-  font-size: 18px;
+  font-size: 1rem;
   line-height: 20px;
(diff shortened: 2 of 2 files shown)


dracos:
Can discuss this, perhaps it could be on the TfL cobrand only if desired and it all functions okay, but we can't progress this as is at present, see e.g. https://github.com/mysociety/fixmystreet/commit/ad979117928a3418e6cbbe2dde2e42364758c22c and https://mysocietysupport.freshdesk.com/a/tickets/1907 and https://mysocietysupport.freshdesk.com/a/tickets/1908.

lucascumsille:
@dracos I wonder if it's worth making it an optional feature in case we have more cases like this in the future. Otherwise happy to just make it for TFL.

lucascumsille:
@lizettal and @sallytay, after discussing it with @dracos, we have decided not to move ahead fixing this issue, considering it was marked as a "nice to have" by the client, and the main issue has already been tackled [here](https://github.com/mysociety/fixmystreet/pull/4880). This fix could create some problems with other cobrands(See [Matthew's comment above](https://github.com/mysociety/fixmystreet/pull/4881#issuecomment-2041617337)), especially Bucks, and making a "Nice to have" override just for one cobrand doesn't seem very sensible. 

We are thinking of improving this process in the future, but I don't think we have a deadline at the moment.

sallytay:
@lucascumsille thank you for the update, when I go back to TfL to let them know what we have / haven't done - I'll make sure they know this.

PR #4885, opened 2024-03-15
[Bexley][WW] Refactor iCal code to use GetCollectionByUprnAndDate

nephila-nacrea:
Since newer GetCollectionByUprnAndDatePlus method is not currently supported on Bexley's live system.

Closes https://github.com/mysociety/societyworks/issues/4156.

[skip changelog]


chrismytton (approved):
Yay, glad we got this sorted rather than having to disable it!

PR #4886, opened 2024-03-15
Fix CSV export with category with brackets.

dracos:
If the CSV export was acting on a pre-made CSV, and a category with brackets was explicitly specified in the query, it would not match. [skip changelog] For FD-3976.


MorayMySoc (approved):
All good

PR #4887, opened 2024-03-15
[Peterborough][WW] Add text messaging option for bulky waste

MorayMySoc:
Adds text box to opt into text updates for bulky waste bookings.

Allows multiple govuk_notify templates/keys to be added in array

https://github.com/mysociety/societyworks/issues/3782

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste.pm
+++ b/perllib/FixMyStreet/App/Form/Waste.pm
@@ -59,7 +59,7 @@ sub validate {
     my $email = $self->field('email');
     my $phone = $self->field('phone');
     return 1 unless $email && $phone;
-    my $text_updates = $self->field('extra_bulky_text_updates');
+    my $text_updates = $self->field('extra_bulky_text_reminders');
 
     my $c = $self->c;
     my $cobrand = $c->cobrand->moniker;
--- a/perllib/FixMyStreet/App/Form/Waste.pm
+++ b/perllib/FixMyStreet/App/Form/Waste.pm
@@ -1,5 +1,6 @@
 package FixMyStreet::App::Form::Waste;
 
+use FixMyStreet::SMS;
 use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Wizard';
 
@@ -58,6 +59,7 @@ sub validate {
     my $email = $self->field('email');
     my $phone = $self->field('phone');
     return 1 unless $email && $phone;
+    my $text_updates = $self->field('extra_bulky_text_updates');
 
     my $c = $self->c;
     my $cobrand = $c->cobrand->moniker;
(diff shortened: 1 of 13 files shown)


perllib/FixMyStreet/App/Form/Waste/Bulky.pm
```diff
@@ -34,9 +34,15 @@ has_page cannot_book => (
 
 has_page about_you => (
     intro => 'bulky/about_you_preamble.html',
-    fields => ['name', 'email', 'phone', 'continue'],
+    fields => ['name', 'email', 'phone', 'extra_bulky_text_updates', 'continue'],
     title => 'About you',
     next => 'choose_date_earlier',
+    update_field_list => sub {
+        my $form = shift;
+        if ($form->c->cobrand->moniker eq 'peterborough') {
```
davea, line 42: It would be better to have a `COBRAND_FEATURES` flag for the new feature rather than hardcoding the `peterborough` check everywhere. This would make it easier to roll out to other cobrands later and also means we can turn it on/off for Peterborough with a redeploy rather than code changes.

perllib/FixMyStreet/Roles/CobrandBulkyWaste.pm
```diff
@@ -552,10 +552,14 @@ sub bulky_reminders {
 sub _bulky_send_reminder_email {
     my ($self, $report, $h, $params) = @_;
 
-    return unless $report->user->email;
+    return unless $report->user->email || $report->get_extra_field_value('bulky_text_updates');
```
davea, line 555: This is inside a function called `_bulky_send_reminder_email` so shouldn't have anything to do with sending texts - can you factor this (and the below `call_hook` lines) into a new `_bulky_send_reminder_text` function which gets called independently?

t/app/controller/waste_peterborough_bulky.t
```diff
@@ -64,6 +66,7 @@ create_contact(
     { code => 'payment' },
     { code => 'payment_method' },
     { code => 'property_id' },
+    { code => 'bulky_text_updates' },
```
davea, line 69: This is probably me not paying close enough attention but it took me ages of wondering why the feature wasn't working before I spotted this line and realised I also needed to add a `bulky_text_updates` hidden field to my existing `Bulky collection` category. Always helpful to please include notes for the reviewer when adding code that needs configuration/changes beyond checking out the branch.

perllib/FixMyStreet/App/Form/Waste.pm
```diff
@@ -55,17 +56,25 @@ sub validate {
...
     my $c = $self->c;
     my $cobrand = $c->cobrand->moniker;
     my $is_staff_user = ($c->user_exists && ($c->user->from_body || $c->user->is_superuser));
     my $staff_provide_email = (ref $self) =~ /Garden/ && ($c->cobrand->moniker eq 'kingston' || $c->cobrand->moniker eq 'sutton');
 
     $email->add_error('Please provide an email address')
-        unless $email->is_inactive || $email->value || ($is_staff_user && !$staff_provide_email);
-
+        unless $email->is_inactive || $email->value || ($is_staff_user && !$staff_provide_email)
+        || (FixMyStreet::SMS->parse_username($phone->value)->{may_be_mobile} && $cobrand eq 'peterborough');
```
davea, line 68: This line seems to make the email address optional on the "About you" page if I provide a mobile number, but I can't complete a booking without an email address: 
image

perllib/FixMyStreet/App/Form/Waste/AboutYou.pm
```diff
@@ -62,6 +62,14 @@ has_field email => (
     },
 );
 
+has_field extra_bulky_text_updates => (
```
davea, line 65: The wording here isn't very clear, what about something like:
image

davea (changes requested):
Functionally this is nearly there, but I've made a few comments about ways the code could be improved/tidied up a bit.

perllib/FixMyStreet/App/Form/Waste.pm
MorayMySoc: Fix to this in ba1e0155f9f0afe8ec29ac485f1102d9dad002d3 May need more discussion...

perllib/FixMyStreet/App/Form/Waste/AboutYou.pm
```diff
@@ -62,6 +62,14 @@ has_field email => (
...
+    type => 'Checkbox',
+    label => 'Bulky text updates',
```
davea, line 67: I still think the inclusion of 'updates' here is misleading as the message aren't really updates - they're reminders.

davea (changes requested):

davea (approved):

PR #4888, opened 2024-03-15
[Bexley][WW] Implement postcode lookup using LLPG data

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/4149.

- Includes script for generating SQLite tables from LLPG CSV
- Changes necessitated switching from using Whitespace-specific
AccountSiteID to UPRN
- Check for whether a property has children no longer needed, as
we no longer fetch any parent properties

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -2,6 +2,7 @@ package FixMyStreet::Cobrand::Bexley::Waste;
 
 use Moo::Role;
 
+use BexleyAddresses;
 use Integrations::Whitespace;
 use DateTime;
 use DateTime::Format::W3CDTF;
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -14,50 +15,174 @@ has 'whitespace' => (
...
 
+sub postcode_database_file {
+    FixMyStreet->path_to('../data/bexley-ww-postcodes.sqlite');
+}
+
 sub bin_addresses_for_postcode {
     my ($self, $postcode) = @_;
 
-    my $addresses = $self->whitespace->GetAddresses($postcode);
+    return [] unless -e postcode_database_file();
```
chrismytton, line 25: I think it might be worth putting in a `die` here if there's no database file, don't think much else is going to work without it, so might as well fail loudly if it's missing!

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -14,50 +15,174 @@ has 'whitespace' => (
...
-        }
-    } @$addresses ];
+    my @data =
+        map {
+            my $address_string = _build_address_string($_);
+            {
+                value => $_->{uprn},
+                label => FixMyStreet::Template::title( $address_string )
+            };
+        } sort _sort_addresses @$addresses;
```
chrismytton, line 78: This code looks good, but adds quite a lot of incidental complexity to this file.

I wonder if it would be possible to pull this code (and the related methods below) into a separate `BexleyLLPG` (or whatever) class which encapsulates all the SQLite interactions? Then this `bin_addresses_for_postcode` could simply create a new instance of that class and call a method on it with a postcode and get back the relevant UPRN and address strings.

This is a similar pattern to the `BromleyParks` module, which just exposes a `lookup($search)` method.

chrismytton (changes requested):
This is looking awesome! 🗺️ 

A few suggestions below, though happy to be refuted on any point if I've misunderstood anything!

chrismytton (approved):
Woohoo!

I think some tests are failing because I've pushed some new Bexley waste tests to `master` that are using the old URLs. If you rebase this branch onto master and update the failing tests to use `/waste/10001` type URLs that should hopefully fix those failures.

PR #4892, opened 2024-03-21
[Kingston] Container work

dracos:
For https://github.com/mysociety/societyworks/issues/4199 (newly updated)
[skip changelog]


perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -100,6 +104,32 @@ sub garden_waste_renewal_sacks_cost_pa {
...
+    if (my $cost = $self->_get_cost('request_replace_cost')) {
+        my $cost_more = $self->_get_cost('request_replace_cost_more') || $cost/2;
+        if ($quantity > 1) {
+            $cost += $cost_more * ($quantity-1);
+        }
+        my $names = $self->{c}->stash->{containers};
+        if ($names->{$id} !~ /bag|sack|food/i) {
+            #my $price = sprintf("£%.2f", $cost / 100);
+            #$price =~ s/\.00$//;
+            #my $hint = "$price administration/delivery charge";
```
davea, line 125: Can these go?

davea (approved):

dracos:
Hi @davea,  Parts of this you've already looked at, but has been rejigged due to changes in requirements (see updated issue). I've tried to make this most straightforward to implement by first stripping out all the extra bits that would no longer be needed (so just removals there), then add payments and how-many-live there (you've already reviewed those), then multiple bins (which you had reviewed, but now is for replacements as well as new as the damaged/missing is gone), then lastly the new change capacity (which is similar to the Sutton but done differently due to the extra questions/flow). Shout if you have any questions.

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -101,6 +110,293 @@ sub garden_waste_renewal_sacks_cost_pa {
...
+    foreach (@{$c->stash->{service_data}}) {
+        next unless $_->{next} || $_->{request_only};
+        my $containers = $_->{request_containers};
+        my $maximum = $_->{request_max};
+        foreach my $id (@$containers) {
+            $maxes{$id} = $maximum->{$id};
+        }
+    }
+
+    sub n { my $n = ucfirst NUMWORDS(shift); $n =~ s/Zero/None/; $n; }
```
davea, line 238: I felt like I had to stop and think about my answer when I saw this:

image

Something about the numerals not being shown caught me by surprise. I think this would be a bit more accessible, maybe:

image

templates/web/kingston/waste/biggest_bin_allowed.html
```diff
@@ -0,0 +1,19 @@
...
+[% IF property %]
+    [% INCLUDE 'waste/_address_display.html' %]
+[% END %]
+
+<h1 class="govuk-heading-xl">Black bin size change request</h1>
+
+[% IF form.saved_data.how_many_exchange == '7more' %]
+<p class="govuk-body">
+If you have an exceptionally large household or your household has medical needs that create more waste than normal, you can apply for more capacity, but this will be assessed by our officers.</p>
+<p class="govuk-body"><a href="https://www.kingston.gov.uk/">Large refuse bin application form</a></p>
```
davea, line 12: I guess this URL isn't finalised yet :)

davea (approved):
Looks good!

templates/web/kingston/waste/biggest_bin_allowed.html
dracos: no i should probably put it in config for ease shouldn't I

PR #4895, opened 2024-03-26
[Northumberland] Being able to filter on one category in multiple parent categories

MorayMySoc:
Set up categories in index to include name of their group.

Use this group name to check if a report is in the selected category (through extra->group).

If there are reports that don't identify as being in that category, then still use a 'Multiple' category.

Fixes https://github.com/mysociety/societyworks/issues/4039

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -136,24 +136,54 @@ sub index : Path : Args(0) {
         }
 
         my %group_names = map { $_->{name} => $_->{categories} } @{$c->stash->{category_groups}};
-        # See if we've had anything from the body dropdowns
-        $c->stash->{category} = [ $c->get_param_list('category') ];
-        my @remove_from_display;
 
-        foreach (@{$c->stash->{category}}) {
-            next unless /^group-(.*)/;
-            for my $contact (@{$group_names{$1}}) {
-                push @{ $c->stash->{category} }, $contact->category;
-                push @remove_from_display, $contact->category;
+        # Categories being received are expected to carry group information
+        # from the interface. A category can be:
+        # 'group-Roads' -> A master category 'Roads' which will contain subcategories
+        # 'Potholes-group-Roads' -> A 'Potholes' category under roads
+        # 'Potholes-group-' -> A 'Potholes' category not under another category
+        # We need to reformat any categories that may be manually put into the url
+        # which may be bookmarks from before grouping was added
+
+        my %display_categories;
+        foreach my $display ($c->get_param_list('category')) {
+            if ($display !~ /group-/) {
+                $display = $display . '-group-';
             }
+            $display_categories{$display} = 1;
         }
 
-        my %display_categories = map { $_ => 1 } @{$c->stash->{category}};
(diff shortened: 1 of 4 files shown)


MorayMySoc:
@dracos Was hoping to get this finished but tests failing so need to look into it further. Hopefully 90% complete unless totally wrong approach so putting as a WIP so can get advice if you have time to look at it to get a rough idea.

lizettal:
@dracos as discussed in sprint planning, can you please review this so we can hopefully get the associated ticket completed.

perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -256,18 +274,28 @@ sub generate_grouped_data : Private {
         %grouped = map { $_->category => {} } @{$c->stash->{contacts}};
     }
     my $problems = $c->stash->{objects_rs}->search(undef, {
-        group_by => [ map { ref $_ ? $_->{-as} : $_ } @groups ],
-        select   => [ @groups, { count => 'me.id' } ],
-        as       => [ @groups == 2 ? qw/key1 key2 count/ : qw/key1 count/ ],
+        group_by => [ map { ref $_ ? $_->{-as} : $_ } (@groups, 'me.extra') ],
```
dracos, line 277: I'd fetch only the extra->>'group' specifically here, rather than all of extra - `extra` will be unique for each report pretty much if there's any extra data at all, thereby stopping the group by working at all, I'd assume.

dracos (changes requested):
I first got "Not a HASH reference" from Dashboard.pm line 291 when trying to view the dashboard. I added a `&& ref $col3 eq 'HASH'` there which fixed that (old extra might not be a hash). But I can't then seem to get this to work, I'm afraid?
I have "Street cleaning" in three parent categories, "Parent A", "Parent B", "Parent C". Each has 2 reports in it. I selected the Parent A Street cleaning, hit submit, the URL has `category=Street+cleaning-group-Parent+A` in it, but all three are ticked in the resultant page and all 6 reports are still shown in the summary table, still under "Multiple". Same as selecting "All Parent A". I think the first part is due to the `reverse` which means "Street cleaning" is set as the key, just like it was beforehand; and the category filter still seems to have the same information/display.
Given that, I'm afraid I haven't really looked at the display code further on, but I'm finding it quite hard to understand what it's doing, so it might need talking through the best way to proceed. I don't think we want this to be this complicated if we can help it.

MorayMySoc:
If I add `&& ref $col3 eq 'HASH'` it means it's never true when I run the test file, so not sure if I've misunderstood that.
I've not managed to add the `->>group` to the order as getting an error on that.
The test that is failing seems to be due to some kind of escaping of characters - I've tried to mitigate that with ` [% SET category_built = mark_safe(cat.category) _ '-' _ mark_safe(group.group_id) %]` in index.html and the text looks correct if I log it out, but doesn't seem to match.
But I think that the numbers are matching up - not sure whether I'm slugging my way down a dead end and should look to start from base on this or try and sort out these issues on this code...

perllib/FixMyStreet/App/Controller/Dashboard.pm
MorayMySoc: Thanks for chat - this should be done now

PR #4900, opened 2024-03-28
Some waste/Echo template/code refactoring

dracos:
I did this as a preliminary to Merton work, to see how minimal I could then have that setup. More could always be done, of course, but I think this is a big step to making Echo integrations much easier to set up, removing a lot of identical boilerplate code and replacing with a number of smaller configuration functions. Shouldn't be any functional changes at all. [skip changelog]


templates/web/base/admin/templates/_external.html
```diff
@@ -1,12 +1,41 @@
...
-  <p>
-    [% loc('If you want to use this template to prefill the update field when a report&rsquo;s <strong>external</strong> (e.g. Confirm) status code changes, enter the status code here.') %]
-  </p>
-</div>
-<p>
-  <label for="external_status_code">[% loc('External status code') %]</label>
-  <input type="text" id="external_status_code" name="external_status_code" class="form-control" size="30" value="[% rt.external_status_code | html %]">
-</p>
+
+[% IF c.cobrand.split_echo_external_status_code %]
```
davea, line 5: Ooh, I didn't know this was a thing, very nifty. Might come in handy for Surrey!

davea (approved):

PR #4907, opened 2024-04-05
[Sutton] Waste text changes

MorayMySoc:
Changes requested for text in Sutton WW

Details in commits


templates/web/base/govuk/fields.html
```diff
@@ -195,6 +195,7 @@ <h2 class="govuk-fieldset__heading">
       [% END %]
       [% IF item.tags.divider %]
         <p class="govuk-body"><strong>[% item.tags.divider %]</strong></p>
+        [% INCLUDE 'govuk/divider_intro.html' %]
```
chrismytton, line 198: I have a feeling this will fail if/when this code runs on other cobrands which don't have this template.

Probably worth wrapping it in a try/catch, like we do in other places:

```
[% TRY %][% INCLUDE 'govuk/divider_intro.html' %][% CATCH file %][% END %]
```

templates/web/sutton/waste/request/intro.html
```diff
@@ -1,4 +1,3 @@
 <p>
-Any Sutton specific request container introductory text here.
-Including details/link about needing a 360l for medical purposes.
+The Council has continued to provide waste and recycling containers free for as long as possible, but this is no longer sustainable given the economic pressures that the Council is facing. More information can be found on our <a href=''>web page</a>.
```
chrismytton, line 2: Missing the `href` for this link.

chrismytton (changes requested):

templates/web/sutton/waste/request/intro.html
dracos: Awaiting from Sutton

templates/web/base/govuk/fields.html
dracos: Have changed `divider` to `divider_template` to use as a template include (given added text anyway), then this won't matter as will only be used when that template exists, and made it easier to customise.

dracos (approved):

PR #4916, opened 2024-04-10
[Bromley] Echo open investigate report text change

MorayMySoc:
Commenter name and email to be used to populate the title, name and email for new Event creation

https://github.com/mysociety/societyworks/issues/4232

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -273,7 +273,7 @@ sub open311_pre_send {
     if (my $comment_id = $row->get_extra_metadata('echo_report_reopened_with_comment')) {
         my $comment = FixMyStreet::DB->resultset('Comment')->find($comment_id);
         if ($comment && $comment->text) {
-            my $text = "Closed report has a new comment: " . $comment->text . "\n\n" . $row->detail;
+            my $text = 'Closed report has a new comment: ' . $comment->text . "\r\n" . $comment->user->name . ' ' . $comment->user->email . "\r\n" . $row->detail;
             $row->detail($text);
         }
     }
--- a/t/cobrand/bromley.t
+++ b/t/cobrand/bromley.t
@@ -845,8 +845,9 @@ subtest "comment on a closed echo report result in a resend under 'Referred to V
 
     my $req = Open311->test_req_used;
     my $c = CGI::Simple->new($req->content);
+    my $detail = $report->detail;
     is $c->param('attribute[Event_ID]'), $event_id, 'old event ID included in attributes';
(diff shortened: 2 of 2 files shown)


t/cobrand/bromley.t
```diff
@@ -846,7 +846,7 @@ subtest "comment on a closed echo report result in a resend under 'Referred to V
     my $req = Open311->test_req_used;
     my $c = CGI::Simple->new($req->content);
     is $c->param('attribute[Event_ID]'), $event_id, 'old event ID included in attributes';
-    like $c->param('description'), qr/Closed report has a new comment: comment on closed event/, 'private comments included in description';
+    like $c->param('description'), qr/comment on closed event|Bromley pkg-tcobrandbromleyt-bromley\@example.com|echo report Test 1 for 1 Detail/, 'Comment on closed report included in new report description';
```
dracos, line 849: Is this meant to have regex alternative `|` like this? It'll match if any of them are present, which doesn't sound right, they should presumably be escaped.

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -273,12 +273,19 @@ sub open311_pre_send {
...
+            my $text = _munge_text_for_new_investigation_event($comment, $row->detail);
             $row->detail($text);
         }
     }
 }
 
+sub _munge_text_for_new_investigation_event {
+    my ($comment, $detail) = @_;
+
+    my $text = $comment->text . '|' . $comment->user->name . ' ' . $comment->user->email . '|' . $detail;
```
dracos, line 285: Make sure there are spaces around the `|`

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -273,12 +273,19 @@ sub open311_pre_send {
...
+    return Utils::truncate_text($text, 250, '...');
```
dracos, line 286: Is there a reason to truncate the text here? It feels like we shouldn't truncate here, and if any truncation is needed it would happen at the open311-adapter end that's formatting for sending.

dracos (changes requested):

perllib/FixMyStreet/Cobrand/Bromley.pm
MorayMySoc: No - I guess you will end up with truncated text which encourages you to visit the report, only to find the exact same truncated text.

MorayMySoc:
Changes made to match/pair with https://github.com/mysociety/open311-adapter/pull/335

MorayMySoc:
@dracos This is the partner to the open311 review you just did...

dracos (approved):

PR #4917, opened 2024-04-11
.report-a-problem-btn button in navbar remains visible

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4216

This PR improves the consistency of a component that is repeated across all pages, by keeping `.report-a-problem` button always visible. This fix should help users with low vision, blind or intellectual disabilities.

https://www.w3.org/WAI/WCAG21/Understanding/consistent-navigation

IMPORTANT:
Only the first commit affects the visibility of the button. The second commits is an optional one, that keeps the button's text always as "Report a problem", so it removes all instances "Report another problem here" and "Report a problem here". We can drop it if we think it could controversial with clients and users or it could decrease the UX of site.

[Skip changelog]


dracos:
I'm not sure I agree with this.

https://www.w3.org/WAI/WCAG21/Techniques/general/G61 has as one of its examples: "A Web site has a logo, a title, a search form and a navigation bar at the top of each page; these appear in the same relative order on each page where they are repeated. On one page the search form is missing but the other items are still in the same order." - this would to me be the same situation. On the page you are already reporting, the button is missing, but the other items are still in the same order.

Why do you want a "Report a problem" action on a page where you're already reporting a problem? Surely that is the confusing issue, people might click it and be taken away from the problem they have already started to report.

lucascumsille:
Thanks @dracos I didn't check the "sufficient techniques" section I'll check those ones from now on =), it also makes sense.

@lizettal and @sallytay just in case:

I think we could tell the client that we won't move forward with this point, because the visibility of the "report a problem" button doesn't stop us from meeting [the criteria 3.2.3](https://www.w3.org/WAI/WCAG21/Understanding/consistent-navigation) because of https://www.w3.org/WAI/WCAG21/Techniques/general/G61. Our navigation bar keep a consistent order across pages. We have hidden the "Report a problem" button in some pages to decrease human error when reporting, like accidentally pressing that button in the middle of creating a report.

If I missed something let me know @dracos

PR #4918, opened 2024-04-11
[Bromley][WW] Send multiple photos through on all categories

MorayMySoc:
Adds sending of photos for all categories. Assume all Waste is Echo. Echo set up for receiving photos but think general open311 just receiving urls

https://github.com/mysociety/societyworks/issues/4233

[skip changelog]


chrismytton (approved):

MorayMySoc:
This was actually for merging into https://github.com/mysociety/fixmystreet/pull/4869 which is done

PR #4922, opened 2024-04-15
Added error_color variable and decreased lightness of default error color.

lucascumsille:
Fixes part of:  https://github.com/mysociety/societyworks/issues/4268

Red color `(#FF0000)` wasn't passing contrast test against white text so now it's a bit darker(with this PR the contrast test is 5.45). I have also made it a variable `error_color`, that should help the styling process for cobrands reducing the need for overrides.

Screenshot 2024-04-15 at 15 17 20

The second commit is a tidy up, so now instead of having the overrides for the cobrands that were using their own red, now use the variable instead.

[Skip changelog]

### DIFF
--- a/web/cobrands/bexley/_colours.scss
+++ b/web/cobrands/bexley/_colours.scss
@@ -7,6 +7,7 @@ $mappage-header-height: 128px;
 $white: #fff;
 
 $red: #B10E1E;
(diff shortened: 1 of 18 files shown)


dracos (approved):
Nice little improvement :)

PR #4923, opened 2024-04-16
[Bexley][WW] Fetch updates for missed collection reports

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/4262.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -59,9 +59,10 @@ sub waste_fetch_events {
             ? $missed_collection_properties->{ServicePropertyValue}
             : '';
 
-        # TODO To open311-adapter or some kind of config
+        my $config = $self->feature('whitespace');
         my $new_state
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -15,6 +15,102 @@ has 'whitespace' => (
...
+                        send_state  => 'processed',
+                        text        => $new_state->{text},
+                        user          => $self->body->comment_user,
+                    },
+                );
+                $comment->insert;
+                $report->state( $new_state->{fms_state} );
+                $report->lastupdate( $comment->created );
+                $report->update;
+            } );
```
chrismytton, line 93: I see the existing `FixMyStreet::Roles::CobrandEcho::waste_fetch_events` method uses `Open311::GetServiceRequestUpdates` to process the updates from Echo.

Looking at the code there I think that's mostly so that it can use response templates, which isn't needed here currently because the response text is hard-coded, so I think this approach is ok for now, but mentioning in case this comes up in the future.

chrismytton (approved):
Looks good! 👍

PR #4928, opened 2024-04-21
[Brent] [Waste] Add upcoming renewal email opt-in control for garden sub and renew.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4202

# Preview

renewal control

renewal summary

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1363,6 +1363,7 @@ sub setup_garden_sub_params : Private {
     } else {
         $service_id = $c->cobrand->garden_service_id;
     }
+    $c->set_param('email_renewal_reminders_opt_in', $data->{email_renewal_reminders} eq 'Yes' ? 'Y' : 'N') if $data->{email_renewal_reminders};
     $c->set_param('service_id', $service_id);
     $c->set_param('current_containers', $data->{current_bins});
     $c->set_param('new_containers', $data->{new_bins});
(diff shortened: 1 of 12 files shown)


perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -47,6 +47,8 @@ sub waste_payment_ref_council_code { "RBK" }
 
 sub garden_collection_time { '6:30am' }
 
+sub garden_subscription_email_renew_reminder_opt_in { 0 }
```
dracos, line 50: Don't need this line as it inherits from CobrandSLWP

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1477,6 +1477,11 @@ sub process_garden_renew : Private {
     $c->forward('setup_garden_sub_params', [ $data, $type ]);
     $c->forward('add_report', [ $data, 1 ]) or return;
 
+    if ($data->{email_renewal_reminders}) {
```
dracos, line 1480: Could this (and 1538 similarly) be in `setup_garden_sub_params` so only needs to be in the one place? I know it's also called for cancellation etc but this check would still be there.

dracos (changes requested):
Looks good, only couple of tiny suggestions :)

perllib/FixMyStreet/App/Controller/Waste.pm
neprune: I think `setup_garden_sub_params` is called before we have the report we can attach the extra field to. Not sure there's any good alternative.

dracos (approved):
Thanks :) Will need to update the Brent open311-adapter Echo config on the server for 1159 (garden) to add the extra data field on staging.

PR #4930, opened 2024-04-22
[Buckinghamshire] Remove winter maintainence video

chrismytton:
Fixes FD-4085

### DIFF
--- a/templates/web/buckinghamshire/footer_extra_js.html
+++ b/templates/web/buckinghamshire/footer_extra_js.html
@@ -5,7 +5,4 @@ IF bodyclass.match('mappage');
         version('/cobrands/buckinghamshire/grass_cutting.js')
     );
 END;
-IF bodyclass.match('frontpage');
-    scripts.push( version('/vendor/lite-youtube-embed/lite-yt-embed.js') );
-END;
 %]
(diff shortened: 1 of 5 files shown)


dracos (commented):
Is fine, though I think a (partial) revert of 30507877bca304011d5b28dee2fd297eb2718409 might be better - moving the figures back to the left column to keep it even? (but not removing youtube lite as that's used in the reporting as well)

web/cobrands/sass/_layout.scss
```diff
@@ -861,10 +861,6 @@ textarea.form-error {
   }
 }
 
-#front-howto {
```
dracos, line 864: This isn't the right bit to try and remove, which is about 30 lines further down - but we don't want to remove this base style because eg Oxfordshire are also using it.

dracos (changes requested):
One thing removed by mistake, I think, is all :)

web/cobrands/sass/_layout.scss
```diff
@@ -892,9 +892,6 @@ textarea.form-error {
 #front-recently h2 {
    margin-top:0;
 }
-#front-howto lite-youtube ~ h2 {
```
dracos, line 895: Sorry if my previous comment wasn't clear - when I said "we don't want to remove this base style because eg Oxfordshire are also using it." I meant we don't want to be removing anything from this file.

PR #4935, opened 2024-04-24
Update index.html

sallybracegirdle:
Updated copy to link to the new Open311 API information page, and removed links to broken Open311 website.  [skip changelog]


templates/web/base/open311/index.html
```diff
@@ -22,32 +20,13 @@ <h1>FixMyStreet support for Open311</h1>
...
-Pro</a>, our hosted service which sits seamlessly on your council
-website.</p>
-
-<p>We can integrate it with any council back-end system, but if you use
-Open311, we charge less. That's how much we believe it's the sane way forward
-for council problem-reporting systems.</p>
-[% END %]
+<p>Open311 enables any authority to receive reports from FixMyStreet.com and the app directly into a backend system instead of via email. It will also allow you to close/update reports on FixMyStreet.com or the app when work is completed.
+
+<p>Find out more about connecting to FixMyStreet via Open311</p>
```
dracos, line 28: Should this be a link? :)

dracos (changes requested):
As well as the missing link - this page is shown at /open311 on all installations of FixMyStreet, not just .com. So the language used either needs to make sense for all installations, or .com specific bits need to be put behind a `[% IF c.cobrand.moniker == 'fixmystreet' %]` section (as you can see in part of the removed text here).

sallybracegirdle:
@dracos Ok, I have edited this and it hopefully works better now?

PR #4938, opened 2024-04-26
[Bristol] Disable search bar as doesn't include extra questions

MorayMySoc:
Bristol have Pothole as an extra question on Roads so doesn't show up in search bar

https://mysocietysupport.freshdesk.com/a/tickets/4112

[skip changelog]


dracos (changes requested):
This works fine, but I think easier to create a blank `templates/web/bristol/report/new/category_filter.html` template?

dracos (approved):
Ta :)

PR #4939, opened 2024-04-26
[Waste] Limit image size on Echo backed Bulky Collection reports.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4243


neprune:
@dracos I still need to write tests for some coverage but sharing in case you have an early feedback.

neprune:
@dracos Added some tests now, ready for review.

perllib/FixMyStreet/Roles/EnforcePhotoSizeOpen311PreSend.pm
```diff
@@ -0,0 +1,57 @@
...
+
+Takes the report and the number of images.
+Returns the max number of bytes for each photo on the report.
+0 indicates no max to apply.
+
+=cut
+
+requires 'per_photo_size_limit_for_report_in_bytes';
+
+sub open311_pre_send { }
```
dracos, line 33: I assume this is needed because otherwise the after does not fire if there isn't one in the class? Is fine, just wondering if there's a better way to structure things at a future point.

perllib/FixMyStreet/App/Model/PhotoSet.pm
```diff
@@ -416,4 +416,31 @@ sub redact_image {
...
+# of their original size until they are small enough.
+# Returns the new photoset and a bool indicating if any images were shrunk.
+sub shrink_all_to_size {
+    my ($self, $size_bytes, $resize_percent) = @_;
+
+    my $shrunk = 0;
+    my @images = $self->all_ids;
+    foreach my $i (0.. $#images) {
+        my $blob = $self->get_raw_image($i)->{data};
+        while (length $blob > $size_bytes) {
```
dracos, line 429: Could we multiple $resize_percent by itself each loop, and always shrink from the full size image? Just worried that repeated shrinkings might degrade the quality of the image as opposed to one shrinking (of whatever %).

dracos (changes requested):
A few small suggestions is all :)

perllib/FixMyStreet/Roles/EnforcePhotoSizeOpen311PreSend.pm
neprune: Yeah, `after open311_pre_send` will fail to compile if the method isn't somewhere in the class hierarchy.

dracos (approved):
Thanks :)

PR #4940, opened 2024-04-29
[Peterborough] Allow editing of name field and username field

MorayMySoc:
Discussion that we should maintain the status quo except for Peterborough

https://mysociety.slack.com/archives/C01TK8P1K8T/p1713950574734509?thread_ts=1713948742.228499&cid=C01TK8P1K8T

https://mysocietysupport.freshdesk.com/a/tickets/4086

[skip changelog]


templates/web/base/admin/reports/_edit_waste.html
```diff
@@ -45,12 +45,15 @@
 
 <li><label class="inline-text" for="external_id">[% loc('External ID') %]:</label>
     <input type="text" class="form-control" name="external_id" id="external_id" value="[% problem.external_id | html %]">
-
+[% SET readonly = '' %]
+[% IF !c.cobrand.moniker == 'peterborough' %]
```
chrismytton, line 49: This confused my brain when I first read it! Might be clearer written as `c.cobrand.moniker != 'peterborough'`, perhaps?

Or maybe clearer still would be `[% SET readonly = 'readonly' %]` initially (the default) and then set readonly to blank only if the cobrand is Peterborough (the exception)?

chrismytton (changes requested):
Code looks good in principle, just one question about the conditional in the templates.

chrismytton (approved):
Looks good! 🎉

PR #4943, opened 2024-05-01
Added btn--primary class to "Continue" button in New report page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4246

With this PR "Continue" button will have  same styling as `btn--primary`, to match the branding of each cobrand, the disabled state has also been updated so it has a more evident "disabled" state.

NOTE: To fix the original issue, including `btn-primary` is more of a bonus, but we could also get away with keeping what we currently have and add transparentize effect that passes the contrast colour to the disabled state. However this could be an opportunity to make a standard across all cobrands the use of the mixin `button-variant` for `btn--primary`.  As an extra I think we could also include an extra variable in the mixin to control the `border-radius` property. We would be covering more topics, so happy to open a new issue to update the "button-variant" if you think that would be better. 

### Preview
[PR4943.pdf](https://github.com/user-attachments/files/16929920/PR4943.pdf)

Master Preview(What we currently have)
[Master.pdf](https://github.com/user-attachments/files/16929925/Master.pdf)

### Default:
Screenshot 2024-05-01 at 08 35 42

### Disabled
Screenshot 2024-05-01 at 08 36 28

[Skip changelog]

### DIFF
--- a/templates/web/base/report/new/form_report.html
+++ b/templates/web/base/report/new/form_report.html
@@ -20,7 +20,7 @@
     [% PROCESS "report/new/category_wrapper.html" %]
     [% TRY %][% PROCESS 'report/new/after_category.html' %][% CATCH file %][% END %]
 
-    <button class="btn btn--block btn--final hidden-nojs js-reporting-page--next"
+    <button class="btn btn--block btn--final btn--primary hidden-nojs js-reporting-page--next"
         [% ' disabled' IF disable_form_message %]>[% loc('Continue') %]</button>
 </div>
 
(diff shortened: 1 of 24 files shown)


dracos (commented):
There are a lot of Continue buttons in this template, but you've only updated one of them, was that deliberate?
I guess it looks like the difference was the "Continue" button was steps *within* the process, and then the "Submit" at the end is already btn--primary as the final actual-submission step. But I guess that doesn't really matter/make any difference, if you think it makes more sense for them all to be the same, okay with me.
I don't know how many things would need tidying up, I wouldn't want this to become a giant thing, but as always if it ends up making things simpler for the future, is good.

lucascumsille:
> I guess it looks like the difference was the "Continue" button was steps within the process, and then the "Submit" at the end is already btn--primary as the final actual-submission step. But I guess that doesn't really matter/make any difference, if you think it makes more sense for them all to be the same, okay with me.

My thought was there is only one button visible at a time, so it didn't make much difference. With a "back" button right next to the "continue" one it would have more meaning to have a classic "primary" and "secondary" button approach.

>I don't know how many things would need tidying up, I wouldn't want this to become a giant thing, but as always if it ends up making things simpler for the future, is good.

I think I'm following the same approach we have been following over the last year or so, trying to standardised the branding process with new variables and reduce the need of overrides =)

dracos (changes requested):

dracos (approved):
Thanks :)

PR #4948, opened 2024-05-02
[Merton] Send updates to Crimson

davea:
https://github.com/mysociety/societyworks/issues/4241

For some reason the test fails because of the `-not => { 'me.extra' => { '\?' => 'sent_to_crimson' } },` in `SendWaste::_comments`, despite it appearing to work correctly when run manually. Not sure what's going on there but wanted to get this PR opened in any case.

[skip changelog]


perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -54,4 +55,50 @@ sub _problems {
...
+}
+
+sub _comments {
+    my $self = shift;
+    FixMyStreet::DB->resultset('Comment')->to_body($self->body->id)->search({
+        'problem.cobrand_data' => 'waste',
+        'problem.cobrand' => 'merton',
+        'problem.extra' => { '\?' => 'sent_to_crimson' },
+        'me.external_id' => \'is not null',
+        -not => { 'me.extra' => { '\?' => 'sent_to_crimson' } },
```
dracos, line 98: As extra can be null, this will fail when it is, because the inverse of null is still null. See Bromley.pm for an example, you want something like `-or => [ 'me.extra' => undef, -not => { 'me.extra' => { '\?' => 'sent_to_crimson' } } ],`. Should probably make it have an empty default or try and hide that wrinkle somewhere perhaps.

davea: Aha, thanks. Have updated and added a couple more tests to cover more cases.

PR #4951, opened 2024-05-03
[Merton] Pass Echo event ID to Crimson when sending reports

davea:
For https://github.com/mysociety/societyworks/issues/4277

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/Merton/SendWaste.pm
+++ b/perllib/FixMyStreet/Script/Merton/SendWaste.pm
@@ -5,6 +5,7 @@ use FixMyStreet::DB;
 use FixMyStreet::Queue::Item::Report;
 use FixMyStreet::SendReport::Open311;
 use Open311;
+use Integrations::Echo;
 
 has body => (
     is => 'ro',
@@ -17,6 +18,7 @@ sub send_reports {
     my $problems = $self->_problems;
 
     while (my $row = $problems->next) {
+        $self->set_echo_id($row) or next; # skip if no echo_id
         my $item = FixMyStreet::Queue::Item::Report->new( report => $row );
         FixMyStreet::DB->schema->cobrand($item->cobrand);
         $item->cobrand->set_lang_and_domain($row->lang, 1);
@@ -53,6 +55,27 @@ sub _problems {
         cobrand => 'merton',
         -not => { extra => { '\?' => 'sent_to_crimson' } },
     });
+}
+
+sub set_echo_id {
+    my ($self, $problem) = @_;
+
+    unless ($problem->get_extra_field_value('echo_id')) {
+        my $cobrand = $problem->get_cobrand_logged;
+        my $cfg = $cobrand->feature('echo');
+        my $echo = Integrations::Echo->new(%$cfg);
+        my $event = $echo->GetEvent($problem->external_id);
+        return unless $event && $event->{Id};
+
+        my $row2 = FixMyStreet::DB->resultset('Problem')->search({ id => $problem->id }, { for => \'UPDATE' })->single;
+        $row2->update_extra_field({
+            name => 'echo_id',
+            value => $event->{Id},
+        });
+        $row2->update;
+    }
+
+    return 1;
 }
 
 sub send_comments {
--- a/t/app/controller/waste_merton.t
+++ b/t/app/controller/waste_merton.t
@@ -139,6 +139,7 @@ FixMyStreet::override_config {
     };
     subtest 'Test sending of reports to other endpoint' => sub {
         use_ok 'FixMyStreet::Script::Merton::SendWaste';
+        $e->mock('GetEvent', sub { { Id => 1928374 } });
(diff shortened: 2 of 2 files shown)


perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -17,6 +18,7 @@ sub send_reports {
     my $problems = $self->_problems;
 
     while (my $row = $problems->next) {
+        $self->set_echo_id($row) or next; # skip if no echo_id
```
davea, line 21: If Echo returns an error I figured it'd be better to not send to Crimson, as it might just have been a transient error and the next attempt would work OK.

perllib/FixMyStreet/Cobrand/Merton.pm
```diff
@@ -147,4 +147,15 @@ sub open311_post_send {
         delete $self->{original_external_id};
     }
 }
+
+sub open311_extra_data_include {
+    my ($self, $row, $h) = @_;
+
+    if (my $echo_id = $row->get_extra_field_value('echo_id')) {
```
dracos, line 154: Not sure what this is for - all extra fields are included when sending via Open311 anyway, aren't they?

perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -55,6 +57,27 @@ sub _problems {
...
+        return unless $event && $event->{Id};
+
+        my $row2 = FixMyStreet::DB->resultset('Problem')->search({ id => $problem->id }, { for => \'UPDATE' })->single;
+        $row2->update_extra_field({
+            name => 'echo_id',
+            value => $event->{Id},
+        });
+        $row2->update;
+    }
+
```
dracos, line 77: Guess we want a problem->discard_changes here, so it picks up the data?

perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -55,6 +57,27 @@ sub _problems {
...
+        my $echo = Integrations::Echo->new(%$cfg);
+        my $event = $echo->GetEvent($problem->external_id);
+        return unless $event && $event->{Id};
+
+        my $row2 = FixMyStreet::DB->resultset('Problem')->search({ id => $problem->id }, { for => \'UPDATE' })->single;
+        $row2->update_extra_field({
+            name => 'echo_id',
+            value => $event->{Id},
+        });
+        $row2->update;
```
dracos, line 75: Guess we want a problem->discard_changes here, so it picks up the data?

dracos (commented):
Two (one?) question about the ID - I'd hope extra_data_include wasn't needed if you updated the row to make sure it had the data just added, if I've understood it right.

perllib/FixMyStreet/Script/Merton/SendWaste.pm
davea: Yes that has done the trick. I thought for some reason that it wouldn't be a good idea to change the existing `$problem`, perhaps because of 57c8a7bee25e570215ed55252a972d3bff0316b7?

perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -61,18 +61,18 @@ sub set_echo_id {
...
 
     unless ($problem->get_extra_field_value('echo_id')) {
+        $problem->discard_changes;
         my $cobrand = $problem->get_cobrand_logged;
         my $cfg = $cobrand->feature('echo');
         my $echo = Integrations::Echo->new(%$cfg);
         my $event = $echo->GetEvent($problem->external_id);
         return unless $event && $event->{Id};
 
-        my $row2 = FixMyStreet::DB->resultset('Problem')->search({ id => $problem->id }, { for => \'UPDATE' })->single;
```
dracos: Oh, I was happy with this, because it meant this extra change wouldn't overwrite any simultaneous extra change. I only meant adding a discard_changes after you've done it (so that $problem picks up the change you made with $row).

perllib/FixMyStreet/Script/Merton/SendWaste.pm
dracos: Think it's fine, we're not in a transaction or anything, and it's before any bit of the loop anyway. I'd want it at the end, not the beginning, though?

perllib/FixMyStreet/Script/Merton/SendWaste.pm
dracos: Actually, SendReport::Open311 always calls discard_changes after calling open311_update_missing_data, so I'm not sure discard_changes is needed at all. The tests still pass without it - what was the issue you were looking to fix?

dracos: going to merge without it, and if I've introduced a bug, apologies!

PR #4952, opened 2024-05-07
Improvements to focus order 

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4212

Both of these fixes should improve how users using assistive devices travel across inputs by modifying which focus element we are focusing, making it easier to rectify errors or previous steps.

### List of commits

[Added fix to email input focus behaviour when there is an error](https://github.com/mysociety/fixmystreet/commit/23a43e2081971a375207bc490e7b7a680d4512bb) 
First commit fixes:
SW ref 30: Local RSS feeds and email alerts - part 2
- [x] When the Subscribe by email form is submitted with an error, focus should be sent to the "Email address' input element.

- [x] SW ref 41:  Local RSS feeds and email alerts - part 2
An error message is provided in text, but it is not programmatically associated with the email address input element.

[Added fix for focus behaviour in pages with a .form-section-preview](https://github.com/mysociety/fixmystreet/commit/28c761324c203c0506657cef8be38c5c89b9041d)
SW ref 29: Reported issue - provide an update form - tell us about you (step 2)
- [x] When the second part of the form loads, focus is sent to the Full Name input element. This bypasses the Your Update component, and the opportunity to review the first part of the form.

[Skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1352,7 +1352,7 @@ $.extend(fixmystreet.set_up, {
         $('.js-new-report-user-hidden')[0].scrollIntoView({behavior: "smooth"});
         hide('.js-new-report-user-hidden');
         show('.js-new-report-user-shown').then(function(){
-            focusFirstVisibleInput();
+            $(this).find('.form-section-preview h2').trigger('focus');
         });
     });
 
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -1512,12 +1512,23 @@ $.extend(fixmystreet.set_up, {
     });
     $('body').on('click', '#alert_email_button', function(e) {
         e.preventDefault();
+
+        var emailInput = $(this).closest('.js-alert-list').find('input[type=email]');
+        emailInput[0].required = true;
+
+        if (!$(this).closest('form').validate().form()) {
+            emailInput.focus();
+            return;
+        }
+
         var form = $('<form/>').attr({ method:'post', action:"/alert/subscribe" });
         form.append($('<input name="alert" value="Subscribe me to an email alert" type="hidden" />'));
+
         $(this).closest('.js-alert-list').find('textarea, input[type=email], input[type=text], input[type=hidden], input[type=radio]:checked').each(function() {
             var $v = $(this);
             $('<input/>').attr({ name:$v.attr('name'), value:$v.val(), type:'hidden' }).appendTo(form);
         });
+
         $('body').append(form);
         form.trigger('submit');
     });
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -619,6 +619,10 @@ small.or:after {
     padding-top: 0.6em;
     padding-bottom: 0.5em;
   }
+
+  &:has(.form-error) {
+    @include flex-direction(column);
+  }
 }
 
 .checkbox-group {
--- a/.cypress/cypress/integration/simple_spec.js
+++ b/.cypress/cypress/integration/simple_spec.js
@@ -65,9 +65,10 @@ describe('Clicking the map', function() {
 describe('Leaving updates', function() {
     function leave_update() {
         cy.get('[name=update]').type('Update');
+        // [id=].last() due to #2341
         cy.get('.js-new-report-user-show').last().click();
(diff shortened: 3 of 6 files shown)


dracos (changes requested):
I've tweaked the alert email one, let me know if that works for you (and if it's easy to improve the error display).
The second one I don't think works as is, so needs a bit of a change, let me know if what I've said doesn't make sense.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1512,12 +1512,23 @@ $.extend(fixmystreet.set_up, {
     });
     $('body').on('click', '#alert_email_button', function(e) {
         e.preventDefault();
+    
+        var emailInput = $(this).closest('.js-alert-list').find('input[type=email]');
+        var emailValue = emailInput.val();
+
+        if (jQuery.validator.methods.notEmail.call({ optional: function() { return false; } }, emailValue)) {
+            emailInput.focus();
```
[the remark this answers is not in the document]
dracos: Aside - the error display here is already also present on /auth if you click submit on the empty form.

web/cobrands/sass/_base.scss
```diff
@@ -619,6 +619,10 @@ small.or:after {
     padding-top: 0.6em;
     padding-bottom: 0.5em;
   }
+
+  &:has(.form-error) {
+    @include flex-direction(column);
```
dracos, line 624: This works nicely on /auth, on the alert page it changes it from `[email] [Submit]` to
```
[error]
[email]
[Submit]
```
if you're happy with that, fine with me

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1329,9 +1329,25 @@ $.extend(fixmystreet.set_up, {
...
+            // Ignore logged-in form here, because it should all be pre-filled already!
+            var $inputs = $newReportUserShown.find('#form_sign_in_yes input, #form_sign_in_no input').filter(':visible');
+            $inputs.eq(0).trigger('focus');
+          }
+        }
+      };
+      
+      // Call focusFirstVisibleInput function within the .js-new-report-user-show section
+      $('.js-new-report-user-show').on('click', function() {
+        focusFirstVisibleInput();
```
dracos: This is already being called, by line 1371 below?

dracos: I've added a fixup commit - let me know if that works/makes sense.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1329,9 +1329,25 @@ $.extend(fixmystreet.set_up, {
...
+          // Check if there is an element with class .form-section-preview within the .js-new-report-user-shown section
+          var $formSectionPreview = $newReportUserShown.find('.form-section-preview').first();
+          if ($formSectionPreview.length) {
+            $formSectionPreview.children().first().trigger('focus');
+          } else {
+            // Ignore logged-in form here, because it should all be pre-filled already!
+            var $inputs = $newReportUserShown.find('#form_sign_in_yes input, #form_sign_in_no input').filter(':visible');
+            $inputs.eq(0).trigger('focus');
+          }
+        }
```
dracos: This also lost the else for all the other times this function is called (also fixed in fixup)

dracos (commented):
Have pushed a fixup to fix/simplify the JS - let me know if that's okay for you.

lucascumsille:
Thanks @dracos =)
It's all working on my end. Regarding the rebase would you prefer to keep the two initial main commits or everything in one?

dracos:
Two commits is fine, they're making separate independent changes. If anything, you could say the `&:has(.form-error) { @include flex-direction(column); }` change could be its own commit, as that's a separate thing again, fixing the display of flex errors not just there but elsewhere.

dracos (approved):

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1512,12 +1512,23 @@ $.extend(fixmystreet.set_up, {
     });
     $('body').on('click', '#alert_email_button', function(e) {
         e.preventDefault();
+
+        var emailInput = $(this).closest('.js-alert-list').find('input[type=email]');
+        emailInput[0].required = true;
```
dracos, line 1517: I missed this previously, but if you're logged in, there is no emailInput here (because we don't always show an email box then) and so this line errors. Am fixing now, sorry for missing.

PR #4956, opened 2024-05-10
[Merton][WW] Adds configurations for bulky waste

MorayMySoc:
Allows report missed collection on the day

https://github.com/mysociety/societyworks/issues/4270


dracos:
@MorayMySoc I've taken your idea from https://github.com/mysociety/fixmystreet/pull/4736 and used that instead, as it felt like a nicer way of doing it? I've added a fixup commit - let me know what you think about it.
I've already merged the other commits that were in this PR into merton-ww-bulky, thanks.

MorayMySoc:
@dracos I think this fits better into the flow in the way everything else is working rather than being a different solution to the same problem.

PR #4958, opened 2024-05-14
[Whitespace] Add extra logging to Whitespace call method

chrismytton:
We're seeing some 400 errors from Whitespace, so this should help us pin down which method call is causing those errors.


dracos (changes requested):
One query about logging the response

chrismytton:
This didn't work quite as I intended, and we've got some other logging now, so going to close this.

PR #4959, opened 2024-05-15
Allow h{1,2,3} tags in contact notices/questions

davea:


dracos (changes requested):
Looks fine, but if we're allowing this, we also need to fix the issue that notices appear in the front end wrapped in a needless `<label>` - which I don't think we want to have headings inside (I mean, shouldn't have the ones that are already allowed either!).

dracos:
(I've made a commit now on surrey-cobrand that removes the label)

PR #4960, opened 2024-05-15
[Merton] Allow bulky collections to be amended

davea:
[skip changelog]

For https://github.com/mysociety/societyworks/issues/4283


dracos:
Not sure what happened with the merge commits here! The diff would be e4eeffb74...68d20c121

PR #4961, opened 2024-05-15
[Merton] Add staff-only address change form for garden subscriptions. 

neprune:
[skip changelog] closes https://github.com/mysociety/societyworks/issues/4284

Screenshot 2024-05-15 at 13 34 16

### DIFF
--- a/t/app/controller/waste_merton.t
+++ b/t/app/controller/waste_merton.t
@@ -65,6 +65,13 @@ create_contact({ category => 'Request additional collection', email => 'addition
     { code => 'service_id', required => 1, automated => 'hidden_field' },
     { code => 'fixmystreet_id', required => 1, automated => 'hidden_field' },
 );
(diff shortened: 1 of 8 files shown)


dracos (approved):
Neat way of handling the notice, I'd probably have done an intro- style template but this is nicer :)

t/app/controller/waste_merton.t
```diff
@@ -137,28 +144,54 @@ FixMyStreet::override_config {
         is $cgi->param('attribute[Action]'), '3';
         is $cgi->param('attribute[Reason]'), '2';
     };
+
+    my $echo_report;
```
dracos, line 148: Ah, I see there's a flaky test - the new report being created here, when later on it calls send reports, that can be in any order (though normally is always the same) so might return this one rather than the new one last (to fill in Open311->test_req_used). So I think'll I just tweak this to delete the non-Echo report at the end of its test, then it's not there any more.

PR #4963, opened 2024-05-16
[Merton][WW] Add staff features for requesting new containers

MorayMySoc:
Staff can select any available domestic container when doing a request 

They are not restricted to ordering only one container (they can order 3).

They can leave an optional note when ordering a container.

https://github.com/mysociety/societyworks/issues/4273

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton/Waste.pm
@@ -167,6 +167,16 @@ sub waste_garden_sub_payment_params {
     }
 }
 
+sub staff_override_request_options {
+    my ($self, $rows) = @_;
+    return unless $self->{c}->stash->{is_staff};
+
+    foreach my $row (@$rows) {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Merton/Waste.pm
```diff
@@ -167,6 +168,48 @@ sub waste_garden_sub_payment_params {
...
+    my @containers_on_property;
+
+    foreach my $row (@$request_rows) {
+        push @containers_on_property, @{$row->{request_containers}};
+        $row->{'request_allowed'} = 1;
+        $row->{'request_max'} = 3;
+    }
+
+    my %new_row = ();
+    $new_row{service_name} = 'Other containers';
```
dracos, line 192: These could all be set in the my %new_row line.

dracos (changes requested):
Nice way of doing it with request_only. Few small things is all, I'll do them.
I guess they might ask for containers to be put within the services, we'll see.

PR #4932, opened 2024-04-23
[Bexley] Missed collection can't be reported until round confirmed

MorayMySoc:
Need to check when the cab logs say the round was finished before allowing missed reports on services

Also add delayed message if round not complete after collection date

https://3.basecamp.com/4020879/buckets/35109031/todos/7283503961

[skip changelog]

PR #4965, opened 2024-05-17
[Bexley][WW] Fix cab logs USRN bug

nephila-nacrea:
We were looking up the USRN cab logs using the USRN value from
the first UPRN log returned. But if there are no UPRN logs,
there is no USRN value, and therefore *all* logs are requested.
This causes bugs / timeouts.

This PR makes it so we now look up the USRN via the LLPG database, before we request any logs.

[skip changelog]


nephila-nacrea:
Relevant commit moved to https://github.com/mysociety/fixmystreet/pull/4932.

PR #4969, opened 2024-05-22
[Bexley][WW] Fetch successful collections alongside other in-cab logs

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/35109031/todos/7411178677.

Logs with a reason of 'N/A' were being ignored, but we need them to check whether a collection has already been made today for a given service.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -208,8 +208,10 @@ sub bin_services_for_address {
     $property->{missed_collection_reports}
         = $self->_missed_collection_reports($property);
     $property->{recent_collections} = $self->_recent_collections($property);
-    my ($property_logs, $street_logs) = $self->_in_cab_logs($property);
 
+    my ( $property_logs, $street_logs, $successful_collections )
+        = $self->_in_cab_logs($property);
+    $property->{successful_collections} = $successful_collections;
     $property->{red_tags} = $property_logs;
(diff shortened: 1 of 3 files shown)


chrismytton (changes requested):
I think this is a step in the right direction, but we also need to allow for the fact that there aren't always round completion logs, so the presence of _any_ log should be taken as a sign that the round has been completed.

See [this Basecamp message and the subsequent discussions](https://3.basecamp.com/4020879/buckets/35109031/messages/7294752422#__recording_7343934774) for details about this problem.

The fix I did was https://github.com/mysociety/fixmystreet/pull/4932/commits/6ca9636eb4548fa01e1d08f3eff76093362f2fe2, which gets logs for all UPRNs under the relevant USRN, which meant because the code was checking property and street logs it would allow a report if there was a red tag log against the street (but lock the user out if the red tag was against one of their containers).

Guess we could have the `_in_cab_logs` method just return all logs and then add some other methods that filter down the list of logs to look for ones we're interested in, which I think is:

- Red tags for the current property
- Service updates for the street ("No access" type things, I assume)
- Round completion indicters
  - Any kind of logs within the relevant USRN on the relevant date can be taken as an indictor of round completion, I think.

Sorry!

chrismytton (approved):
Yay, this is a huge improvement!

I think as discussed in Slack it would be good to just use the USRN logs, but that can be done in a separate PR.

In the meantime lets get this change on staging for them 🙂

PR #4975, opened 2024-05-24
[Bexley][WW] Make sure default value is passed for hidden container location field

nephila-nacrea:
Turns out 'Hidden' formhandler fields take 'default', not 'value'.

[skip changelog]


chrismytton (approved):
Ah, sneaky bug!

PR #4979, opened 2024-05-29
[Bexley WW] Simplify in-cab logs code to use USRN logs only

nephila-nacrea:
We are pretty certain that UPRN logs are always a subset of USRN logs, so I have simplified the code so we only fetch USRN logs.

[skip changelog]


chrismytton (approved):
Looks good!

PR #4986, opened 2024-06-04
Prevent hidden missed collection selection if given service is no longer eligible

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4342 .

Touches code shared between all cobrands, so presumably fixes the issue across all cobrands. (At the very least, doesn't break existing behaviour - see screengrab below).

https://github.com/mysociety/fixmystreet/assets/11098355/2fc562a9-944f-44d9-b48b-99045c86a7e3

[skip changelog]


chrismytton (approved):
Looks good!

PR #4987, opened 2024-06-04
Initial notts police setup.

neprune:
Initial cobrand setup.

Also closes https://github.com/mysociety/societyworks/issues/4340.


perllib/FixMyStreet/Cobrand/NottinghamshirePolice.pm
```diff
@@ -0,0 +1,63 @@
...
+
+use Moo;
+
+=head2 Defaults
+
+=over 4
+
+=cut
+
+sub council_area_id { 2236 }
```
davea, line 25: There's a hole for Nottingham City in the County area, so to include that I think you'd need:

```suggestion
sub council_area_id { [ 2236, 2565 ] }
```

davea (changes requested):
Good start, just one little thing about areas covered

perllib/FixMyStreet/Cobrand/NottinghamshirePolice.pm
neprune: Ah wow, I'd always assumed county areas would be a superset but this makes sense.

nephila-nacrea: Ah, is this why the example 'NG2 3DZ' postcode was not returning results for me?

neprune: No that's something that needs to be be set-up in the server configuration under `example_places` - you can see where I've done that on the `notts-police-staging-setup` branch.

dracos:
Do we want the parent of the cobrand to be UKCouncils, is that needed as it's a totally standalone cobrand? I imagine there are plus/minus points either way - e.g. see HighwaysEngland/Thamesmead for the functions that had to be copied from UKCouncils (that code suggests a role) so that's awkward, but the parent has unnecessary stuff - e.g. the problem/user restrictions presumably aren't needed, nor all the national highways/munging etc bits that only really apply to cobrands. Just wanted to check this had been considered.

nephila-nacrea (approved):
Approving this and merging into `notts-police-reviewed`.
Some fixes need to be made as per review on https://github.com/mysociety/fixmystreet/pull/4996, but will be made on top of `notts-police-reviewed`.

PR #4988, opened 2024-06-05
[Bexley] Collected today check should check scheduled collections

chrismytton:
In-cab logs only have a RoundCode, but that can't be matched directly to a collection's RoundSchedule because both weekly and fortnightly collections share a RoundCode in the logs. So we need to do this extra check to ensure we were actually expecting a collection in the past three days.

For https://3.basecamp.com/4020879/buckets/35109031/todos/7468948862

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -272,6 +272,16 @@ sub bin_services_for_address {
             && $now_dt->delta_days($completed_or_attempted_collection_dt)
             ->in_units('days') == 0
             ? 1 : 0;
+        my $scheduled_for_today = $now_dt->ymd eq $next_dt->ymd;
+
+        my $min_dt = $self->_subtract_working_days(WORKING_DAYS_WINDOW);
+        my $last_expected_collection_dt = $property->{recent_collections}{$service->{RoundSchedule}};
+        # If the last expected collection was more than three days ago and
+        # it's not scheduled for collection today, then we don't consider
+        # it to have been collected today.
+        if ($last_expected_collection_dt < $min_dt && !$scheduled_for_today) {
+            $collected_today = 0;
+        }
 
         my $filtered_service = {
             id             => $service->{SiteServiceID},
@@ -284,7 +294,7 @@ sub bin_services_for_address {
                 date              => $service->{NextCollectionDate},
                 ordinal           => ordinal( $next_dt->day ),
                 changed           => 0,
-                is_today          => $now_dt->ymd eq $next_dt->ymd,
+                is_today          => $scheduled_for_today,
                 already_collected => $collected_today,
             },
             assisted_collection => $service->{ServiceName} && $service->{ServiceName} eq 'Assisted Collection' ? 1 : 0,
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -262,6 +262,27 @@ sub bin_services_for_address {
...
+            if (
+                $dt_to_check
+                && (  !$last_dt
+                    || $dt_to_check > $last_dt )
+                )
+            {
+                $last_dt = $dt_to_check;
+            }
+        }
+
```
chrismytton, line 285: Moved this bit up here (unchanged) so that I can use `$last_dt` below.

chrismytton:
@nephila-nacrea I need to write more tests for this I think, but would be good if you could check the changes I've made look sensible.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -577,10 +576,9 @@ sub can_report_missed {
 
     # Needs to be within 3 working days of the last completed round, today
     # not included.
-    # NOTE recent_collections shows the 'ideal'/expected collection date,
-    # not actual. So we need to check cab logs for actual collection date.
-    my $last_expected_collection_dt
-        = $property->{recent_collections}{ $service->{round_schedule} };
+    # NOTE last expected collection is the 'ideal'/expected collection date,
+    # not actual. So we need to check cab logs below for actual collection date.
+    my $last_expected_collection_dt = $service->{last} && $service->{last}{date};
```
nephila-nacrea, line 581: Don't know how to add a comment to a section of code that hasn't been changed, but the bit below with 
```
        # We need to check if the last collection was made over
        # WORKING_DAYS_WINDOW ago because some collections are weekly and some
        # fortnightly, but they share a round code prefix.
```
-- does that need changing?

templates/web/bexley/waste/services_extra.html
```diff
@@ -38,7 +38,48 @@ <h3 class="govuk-heading-m waste-service-name">
   [%# Allow superusers to see red tags and service updates for debugging purposes %]
   [% IF c.user.is_superuser %]
   <details>
-    <summary>Red tags</summary>
+    <summary>🐞 Superuser debugging</summary>
```
nephila-nacrea, line 41: THE ENTOMOLOGIST HAS ENTERED THE CHAT.

nephila-nacrea (changes requested):

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
chrismytton: I don't _think_ so…

The bit below that you mention should now get a date for collections with multiple comma-separated round schedules, because this code is using `$service->{last}` which has done the processing for round schedules like `RES-NOR Fri, RES-NOR Tue`, whereas previously it was using `$property->{recent_collections}` directly.

nephila-nacrea: My brain is probably not clicking it together just right, but I'm still not sure it isn't a redundant check...BUT...worst case, it's just redundant. 👍 Will approve 🙂

nephila-nacrea (approved):

PR #4989, opened 2024-06-06
Added varius stying fixes to .js-post-category-messages element

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4308

Added some spacing and sizing fixes for elements inside `.js-post-category-messages`.

Screenshot 2024-06-05 at 11 52 30

https://github.com/mysociety/fixmystreet/assets/13790153/136a6c8f-6ad4-46ad-b7c9-abd64a786f06

[Skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -3365,6 +3365,29 @@ $site-message-border: 1px solid #525252 !default;
   }
 }
 
+.js-post-category-messages,
+.extra-category-questions {
+  h2 {
+    font-size: 1.3rem;
+  }
+
+  h3 {
+    font-size: 1.2rem;
+  }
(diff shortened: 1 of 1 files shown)


lucascumsille:
@davea I added these changes to every cobrand, so not only for Surrey, thinking that in a future everyone might start adding these tags to the message, but let me know if you prefer me to move it, so it only applies to Surrey

davea:
This change looks good in isolation, but before we can merge it I think there's more work to be done on https://github.com/mysociety/fixmystreet/pull/4959 which might necessitate changes if we remove the `label` element for notices. For now no need to do anything - I'll check this against that `surrey-html-tags-in-notices` branch once I've removed the `label`.

PR #4990, opened 2024-06-06
[TfL] Add admin heads for changing red route categories.

neprune:
[skip changelog]

Placed at top of category list, above add category form and in category edit form.

tfl red route heads up

Could maybe restrict to showing in category edit form for ones that are already red route (something like `body.cobrand._tlrn_categories.grep(contact.category).size`) but thought it better to just include for all.


dracos (changes requested):
Works fine, thanks for looking at this; I think we can use body.name instead of body.cobrand_name?
(Looking at the cobrand_name code, I'm not sure the first half of the function is needed any more, it reads like it dates from a time when the body's area IDs were used to look up a cobrand, whereas now the body has direct metadata storing which cobrand handles it...)

dracos (approved):
Wasn't in time for me to do it again ;)

PR #4995, opened 2024-06-07
[Bexley] Add form for properties with no collections

chrismytton:
Makes some changes to the enquiry form to allow properties with no collections to report a missed collection via email to one of two categories.

- Removes the check for a valid service in the `enquiry` sub in the Waste controller, think this is ok, as nothing breaks if there isn't a service, it just isn't recorded if it's missing.
- Adds support for checkboxes in extra data on enquiry categories
- Changes to support enquiry checkboxes
- Changes to support service no longer being required

For https://3.basecamp.com/4020879/buckets/35109031/todos/7182401179

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1037,7 +1037,7 @@ sub enquiry : Chained('property') : Args(0) {
 
     my $category = $c->get_param('category');
     my $service = $c->get_param('service_id');
-    $c->detach('property_redirect') unless $category && $service && $c->stash->{services}{$service};
+    $c->detach('property_redirect') unless $category;
 
     my ($contact) = grep { $_->category eq $category } @{$c->stash->{contacts}};
     $c->detach('property_redirect') unless $contact;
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1037,7 +1037,7 @@ sub enquiry : Chained('property') : Args(0) {
 
     my $category = $c->get_param('category');
     my $service = $c->get_param('service_id');
-    $c->detach('property_redirect') unless $category && $service && $c->stash->{services}{$service};
+    $c->detach('property_redirect') unless $category;
```
dracos, line 1040: Fine for now, but guess we might want to make this still check `$service` exists for Bromley where it is required.

dracos (approved):
One suggestion to not lose the check where it is needed (not that that should happen), but looks good

perllib/FixMyStreet/App/Controller/Waste.pm
chrismytton: https://github.com/mysociety/societyworks/issues/4382

PR #4996, opened 2024-06-10
[Notts] Added front-end

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4256

### Desktop
Screenshot 2024-06-10 at 12 07 38
Screenshot 2024-06-10 at 12 07 44
Screenshot 2024-06-10 at 12 08 51
Screenshot 2024-06-10 at 12 09 00

### Mobile
Screenshot 2024-06-10 at 12 09 48
Screenshot 2024-06-10 at 12 09 35

### Emails
Screenshot 2024-06-10 at 12 06 30
Screenshot 2024-06-10 at 12 06 20

[skip changelog]


templates/web/nottinghamshirepolice/site-name.html
```diff
@@ -0,0 +1 @@
+Inmediate Justice
```
nephila-nacrea, line 1: Typo here - it needs to be 'Immediate'.

templates/web/nottinghamshirepolice/footer_extra.html
```diff
@@ -0,0 +1,128 @@
+<footer class="main-footer">
+  <div class="container">
+    <div class="main-content">
+      <div class="first-column">
+        <img src="/cobrands/nottinghamshirepolice/images/logo.svg" width="265" height="80" alt="Nottinghamshire PCC logo" load="lazy">
+        <img src="/cobrands/nottinghamshirepolice/images/inmediate-justice-logo.png" width="265" height="100" alt="Inmediate Justice logo" load="lazy">
```
nephila-nacrea, line 6: Typos here too (you'll need to change the name of the .png file) - should be 'immediate' :-)

nephila-nacrea (commented):
Also noticed the button placement looks odd here:
Screenshot 2024-06-17 at 11 35 39

Additionally, may not be an issue but the font size seems bigger than on the default FMS site and other cobrands. I often have the sites zoomed to about 80% but it still seemed really 'big' to me.

lucascumsille:
> Also noticed the button placement looks odd here: Screenshot 2024-06-17 at 11 35 39
> 
> Additionally, may not be an issue but the font size seems bigger than on the default FMS site and other cobrands. I often have the sites zoomed to about 80% but it still seemed really 'big' to me.

Unfortunately both of those issues are due to their branding. In their website they use 20px font-size, which is probably the bigger we have ever used. I think they do it for accessibility purposes and help users with vision impairment. Because of that  I had to increase the max-width of the default container and the sidebar in the map section. We did the same for Camden and other cobrands where the font-size is bigger than 16px.

Regarding the button, is an issue with probably most cobrands that have an outline with some off-set value.

lucascumsille:
@nephila-nacrea let me know when you have merged and I'll fix the typos =)

nephila-nacrea:
@lucascumsille Everything is now on `notts-police-reviewed` so you can make a branch on top of that for your fixes :-)

PR #4997, opened 2024-06-10
[Northumberland] Skip sending updates on reports fetched from Alloy.

neprune:
And remove the code that accidentally set this up on the Northamptonshire cobrand.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -41,6 +41,11 @@ sub disambiguate_location {
 
 sub admin_user_domain { 'northumberland.gov.uk' }
 
+sub is_defect {
+    my ($self, $p) = @_;
+    return $p->service eq 'Open311';
+}
+
 =item * The default map zoom is a bit more zoomed-in
 
 =cut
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -41,6 +41,11 @@ sub disambiguate_location {
 
 sub admin_user_domain { 'northumberland.gov.uk' }
 
+sub is_defect {
+    my ($self, $p) = @_;
+    return $p->user_id == $self->body->comment_user_id;
```
dracos, line 46: This is fine, and matches Northamptonshire, I'm just wondering if there's a reason we don't use `$p->service eq 'Open311'` (which is set for all pulled in reports, as far as I can see).

dracos (commented):
One query is all

perllib/FixMyStreet/Cobrand/Northumberland.pm
neprune: Nice, I hadn't spotted that. Change made.

dracos (approved):

PR #4999, opened 2024-06-10
[Camden] Setup Confirm integration

MorayMySoc:
Adds necessary Confirm roles https://github.com/mysociety/societyworks/issues/4298
Fixes https://github.com/mysociety/societyworks/issues/4504

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Camden.pm
+++ b/perllib/FixMyStreet/Cobrand/Camden.pm
@@ -53,6 +53,12 @@ sub disambiguate_location {
     };
 }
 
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
To note, from discussion in Slack today, this doesn't work because Camden.pm has an `open311_update_missing_data` which is overriding the one from ConfirmOpen311. And `open311_config` can presumably be removed from Camden.pm.

MorayMySoc:
Thanks - changes look good to me!

PR #5001, opened 2024-06-11
[Brent] Fixes intermittent test failure

MorayMySoc:
Deletes report created in one test that is interfering with another test.
[skip changelog]


t/cobrand/brent.t
```diff
@@ -143,6 +143,7 @@ subtest 'role report shows staff problem when staff logged in during problem rep
     $mech->get_ok($mech->get_link_from_email($mech->get_email));
     $mech->get_ok('/dashboard?body=' . $brent->id . '&state=&role=' . $role->id . '&start_date=&end_date=&group_by=category+state&export=1');
     $mech->content_contains('"Spraypaint on wall","Some kind of picture"', 'Report has contributed_by set and so shows in roles report');
+    FixMyStreet::DB->resultset('Problem')->search()->first->delete;
```
dracos, line 146: I guess this is fine as the first test in the file, but to prevent possible future similar occurrences, perhaps make the middle `->search(undef, { order_by => { -desc => 'id' } })` so it'll still work if someone sticks a test before it?

dracos (changes requested):

dracos (approved):

PR #5002, opened 2024-06-11
Update footer_extra.html

sallybracegirdle:
Change of link to FMS Pro page and updated copy.

Please check the following:

Thanks for contributing!

### DIFF
--- a/templates/web/fixmystreet.com/footer_extra.html
+++ b/templates/web/fixmystreet.com/footer_extra.html
@@ -37,9 +37,9 @@
                             %]>[% loc("Help") %]</[% c.req.uri.path == '/faq' ? 'span' : 'a' %]></li>
                     </ul>
                     <ul>
-                        <li><a href="/pro/">FixMyStreet Pro</a></li>
+                        <li><a href="https://www.societyworks.org/services/highways/">FixMyStreet Pro</a></li>
                         <li><[% IF c.req.uri.path == '/about/posters' %]span[% ELSE %]a href="/about/posters"[% END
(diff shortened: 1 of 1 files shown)


templates/web/fixmystreet.com/footer_extra.html
```diff
@@ -37,9 +37,9 @@ <h2 class="mysoc-footer__site-name">FixMyStreet</h2>
                             %]>[% loc("Help") %]</[% c.req.uri.path == '/faq' ? 'span' : 'a' %]></li>
                     </ul>
                     <ul>
-                        <li><a href="/pro/">FixMyStreet Pro</a></li>
+                        <li><a href="https://www.societyworks.org/services/highways/">FixMyStreet Pro</a></li>
                         <li><[% IF c.req.uri.path == '/about/posters' %]span[% ELSE %]a href="/about/posters"[% END
-                            %]>[% loc("FixMyStreet Goodies") %]</[% c.req.uri.path == '/posters' ? 'span' : 'a' %]></li>
+                            %]>[% loc("Community promotional material") %]</[% c.req.uri.path == '/posters' ? 'span' : 'a' %]></li>
```
dracos, line 42: Can we think of anything pithier? Only it makes it take up three lines now, bit odd in comparison to the rest:
image
Or dunno if any other solution.

dracos (changes requested):

templates/web/fixmystreet.com/footer_extra.html
sallybracegirdle: I guess could just say 'Promotional material' which is what the page is called?

PR #5004, opened 2024-06-12
Add order_by helper / fix BST tests

dracos:
Lets you switch any use of `search(undef, { order_by => 'title' })` to `order_by('title')` or `search(undef, { order_by => { -desc => 'id' } })` to `order_by('-id')`.

Also fix some test failures when you have a BST database rather than UTC.
[skip changelog]


davea (approved):
🕐 🎉

PR #5006, opened 2024-06-12
[Surrey] CSV export includes subscribers to the report

MorayMySoc:
Adds a column, "Subscribers" which uses the problem alert method to populate the number for subscribers.

The original reporter is not automatically subscribed to their own report for Surrey so the figure is those that have subscribed after the report has been created.

Also adds initial test file

https://github.com/mysociety/societyworks/issues/4369

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Surrey.pm
+++ b/perllib/FixMyStreet/Cobrand/Surrey.pm
@@ -67,4 +67,32 @@ sub get_town {
     return $town;
 }
 
+=head2 dashboard_export_problems_add_columns
+
+Surrey has an extra column in their stats export showing the number of subscribers to a report.
+They are set up not to subscribe the original reporter to their own report so the alert number
+is the number of users who have subscribed to the report for updates
+
+=cut
+
+sub dashboard_export_problems_add_columns {
+    my ($self, $csv) = @_;
+
+    $csv->add_csv_columns(
+        alerts_count => "Subscribers",
+    );
+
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Surrey.pm
```diff
@@ -46,4 +46,28 @@ sub get_town {
...
+
+    $csv->add_csv_columns(
+        alerts => "Subscribers",
+    );
+
+    $csv->csv_extra_data(sub {
+        my $report = shift;
+
+        return {
+            alerts => $report->alerts || '0',
```
davea, line 68: As mentioned on Slack this should use `Admin::Reports::alerts_for_report` to get the same value as displayed in the admin report edit page.

dracos: Just to add note from Slack discussion - this cannot use that either, because that would still be a per-row database lookup, and we can't have any per-row database lookups in this code.

davea (changes requested):
This is going in the right direction, but there's a bit more to be done as currently the `all-csv-export` script is broken by this change.

You'll need to add the SQL to do the lookup to `CSVExport.pm` and do a `$csv->dbi` check in `Surrey::dashboard_export_problems_add_columns`.  If you look at `$EXTRAS` in `CSVExport.pm` that has a few examples of how this preprocessing is done for other columns.

davea (approved):

PR #5010, opened 2024-06-14
[Kingston] Fix cost when request has collection.

dracos:
[skip changelog] FD-4257


dracos:
Note post-live, changed to make sure `$c->set_param('payment', $cost)` is always set, otherwise it leaks from one report to the next (in test, not spotted as first report is the collection, but if you had one bin delivered and two boxes removed, they'd all inherit the 1800 from the first bin...)

PR #5012, opened 2024-06-17
Load Devel::Dwarn in dev server and tests.

dracos:
This means you can use `::Dwarn $var` in both code and tests without needing to add any other import or code.


chrismytton (approved):
🎉

PR #5013, opened 2024-06-17
[Brent] [Waste] Add garden subscription notice.

neprune:
[skip changelog]

https://mysocietysupport.freshdesk.com/a/tickets/4187

brent subscription text

### DIFF
--- a/templates/web/base/waste/services.html
+++ b/templates/web/base/waste/services.html
@@ -50,6 +50,10 @@
             <input type="submit" value="Modify your [% unit.service_name FILTER lower %] subscription" class="waste-service-descriptor waste-service-link">
          [% END %]
           </form>
+      [% ELSIF c.cobrand.moniker == 'brent' AND NOT is_staff %]
+          <p>
+            If you are looking to add another service, please contact our customer services team on 020 8937 5050 who will be able to do this for you.
+          </p>
       [% END %]
       [% IF c.cobrand.call_hook('waste_garden_allow_cancellation') == 'staff' AND is_staff %]
           <form method="post" action="[% c.uri_for_action('waste/garden_cancel', [ property.id ]) %]">


dracos (changes requested):
If they're already subscribed, there's no way they could view this page, which is the intro page for someone who doesn't have a subscription :)
They didn't say where they wanted this, but from what I read I thought they wanted it on the bin day page - where if you're staff it prints out "Modify your garden subscription" it would have the text there, alongside their garden waste subscription information. 

Given they must already have a sub to view that, the text could be "If you are looking to add another service, please contact our customer services team on 020 8937 5050 who will be able to do this for you."

neprune:
> If they're already subscribed, there's no way they could view this page, which is the intro page for someone who doesn't have a subscription :) They didn't say where they wanted this, but from what I read I thought they wanted it on the bin day page - where if you're staff it prints out "Modify your garden subscription" it would have the text there, alongside their garden waste subscription information.
> 
> Given they must already have a sub to view that, the text could be "If you are looking to add another service, please contact our customer services team on 020 8937 5050 who will be able to do this for you."

Well that makes sense 😅 . Thanks 🙏

neprune:
demo screenshot

templates/web/base/waste/services.html
```diff
@@ -47,6 +47,11 @@
          [% IF c.cobrand.moniker == 'kingston' AND (slwp_garden_sacks AND unit.garden_sacks) %]
              <input type="submit" value="Order more garden sacks" class="waste-service-descriptor waste-service-link">
          [% ELSE %]
+            [% IF c.cobrand.moniker == 'brent' %]
+                <p>
+                  If you already have an active Garden Waste subscription and are looking to add another service, please contact our customer services team on 020 8937 5050 who will be able to do this for you.
```
dracos, line 52: To get here, you must have a subscription, so the first part of this If isn't needed, can just be "If you are looking to add..."

templates/web/base/waste/services.html
```diff
@@ -47,6 +47,11 @@
          [% IF c.cobrand.moniker == 'kingston' AND (slwp_garden_sacks AND unit.garden_sacks) %]
              <input type="submit" value="Order more garden sacks" class="waste-service-descriptor waste-service-link">
          [% ELSE %]
+            [% IF c.cobrand.moniker == 'brent' %]
```
dracos, line 50: This is only reached if `show_garden_modify` is true, which for Brent is only for staff (that's why we want to show this message), so this isn't shown to normal users (see normal user display attached). So it needs to be its own section, presumably can still be inside the `NOT pending_cancellation` section, but outside this section.

image

dracos (changes requested):
Sorry, afraid this still won't appear where you want it to

neprune:
@dracos sorry, not my brightest moment, have changed to appear for non-staff

dracos (approved):
Looks good, thanks :)

PR #5015, opened 2024-06-18
[Waste] Add ability to expire session cache entries.

dracos:
We're not expiring property lookups in the session cache (they get replaced if the same property is looked up again, but otherwise will just stick around) which was causing issues as people built up thousands of cached unneeded properties in the session. Factor out the cache stuff to a plugin that can store the time of setting, then add to the expire script something to remove things older than a day.

[skip changelog]


chrismytton (approved):

dracos:
Changed afterwards to add a check session_data is not empty in Session.pm which was causing an error.

PR #5028, opened 2024-06-24
[Notts Police] Training fixes

davea:
A few little tweaks for things that came up in training.


templates/web/nottinghamshirepolice/contact/_footer.html
```diff
@@ -0,0 +1 @@
+<p style="height: 3em"></p>
```
nephila-nacrea, line 1: Can't remember if I've asked this before, but GitHub marks files (like here) with a red symbol if there's no newline at the end of the file. Is that a problem?
My editor settings add a newline automatically on saving, because newline-at-file-end was the policy at my previous job, but I can't remember the reasons why.

nephila-nacrea (approved):

PR #5029, opened 2024-06-24
Don't display 'make name public' option on private reports.

neprune:
Applies to both new reports and udpates for a private category. Sets the default value to false (i.e. anonymous).

https://github.com/mysociety/societyworks/issues/4377


templates/web/tfl/email_sent.html
```diff
@@ -6,9 +6,17 @@ <h1>[% loc("Nearly done! Now check your email&hellip;") %]</h1>
 
     <p>
     [% IF email_type == 'problem' %]
-        We’ve sent you a confirmation email. Click the link to publish your problem.
+        [% IF non_public %]
+            [% loc("We've sent you a confirmation email. Click the link to submit your problem.") %]
```
dracos, line 10: These don't need marking for translating (though I guess doesn't matter if you do, most cobrand directories are not looked at for finding strings).

templates/web/peterborough/email_sent.html
```diff
@@ -27,9 +27,17 @@ <h1>[% loc("Nearly done! Now check your email&hellip;") %]</h1>
 
     <p>
     [% IF email_type == 'problem' %]
-        [% loc("Click the link in our confirmation email to publish your problem.") %]
+        [% IF non_public %]
```
dracos, line 30: Do we want to factor this bit out to a separate base template, given it is the same here and in base (and cheshireeast is going so doesn't matter, so only tfl would then need a different one)?

dracos (changes requested):
All looks good, only suggested change is to factor out the "Click the link..." texts to their own template.

templates/web/cheshireeast/email_sent.html
```diff
@@ -6,17 +6,9 @@ <h1>[% loc("Nearly done! Now check your email&hellip;") %]</h1>
```
dracos, line 6: Can drop this entire template now :)

templates/web/peterborough/email_sent.html
```diff
@@ -27,17 +27,9 @@ <h1>[% loc("Nearly done! Now check your email&hellip;") %]</h1>
 
     <p>
     [% IF email_type == 'problem' %]
-        [% IF non_public %]
```
dracos: Thanks for this, but I think you could take all of this from the h1 to the "Can't find our email" p into one template - it's identical here (and then not identical for TfL) :)

dracos (changes requested):
You can drop the Cheshire East template now, and think might as well share more of the same bit is all

templates/web/cheshireeast/email_sent.html
neprune: Reverted back to what's on master, but can delete if that's preferred.

dracos (approved):
🐈

PR #5031, opened 2024-06-24
Staff cannot update name reporting when logged out

dracos:
We have had a few cases where staff forget they are not logged in, report on behalf of someone else when logged out, and have their own account name updated when clicking the confirmation link.


MorayMySoc (approved):
Looks good to me

PR #5035, opened 2024-06-27
[Merton] Ask for user details when staff cancelling.

dracos:


MorayMySoc (approved):
This looks good to me

PR #5036, opened 2024-06-27
[Surrey] Various tweaks

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4370 - don't think I've missed anything, but would be good to check. We should probably have a simpler generic way to do a "use external_id over id" option.
Fixes https://github.com/mysociety/societyworks/issues/4409 - this seemed a very simple "fix" so not sure if this will have broken anything!
Fixes https://github.com/mysociety/societyworks/issues/4408 - have added the option, not sure I fully understood Dave's point here but hopefully this is enough for whatever he wanted (could also probably have a generic way of doing this as in use in a few places now)
Fixes https://github.com/mysociety/societyworks/issues/4391 - hopefully this is just the config from the new feature you added.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Surrey.pm
+++ b/perllib/FixMyStreet/Cobrand/Surrey.pm
@@ -44,6 +44,12 @@ sub report_age {
     };
 }
 
+=item * Add display_name as an extra contact field
+
+=cut
+
+sub contact_extra_fields { [ 'display_name' ] }
+
 =head2 get_town
 
 Returns the name of the town from the problem's geocode information, if present.
--- a/perllib/FixMyStreet/Cobrand/Surrey.pm
+++ b/perllib/FixMyStreet/Cobrand/Surrey.pm
@@ -32,6 +32,18 @@ sub disambiguate_location {
 
 sub report_sent_confirmation_email { 'external_id' }
 
+=item * The default map view shows closed/fixed reports for 31 days
+
+=cut
+
+sub report_age {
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/Cobrand/Surrey.pm
```diff
@@ -32,6 +32,17 @@ sub disambiguate_location {
 
 sub report_sent_confirmation_email { 'external_id' }
 
+=item * The default map view shows closed/fixed reports for 31 days
+
```
nephila-nacrea, line 36: https://github.com/mysociety/societyworks/issues/4391 asks for 90 days for open reports; I am assuming this is a default value but where is it defined in our codebase? Default.pm gives 6 months.

nephila-nacrea (changes requested):

templates/email/surrey/_council_reference_alert_update.html
```diff
@@ -0,0 +1 @@
+_council_reference.html
```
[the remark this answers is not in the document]
dracos: If you're reviewing these only in github rather than checked out - these are symlinks to these files, they are not actual files at all.

perllib/FixMyStreet/Cobrand/Surrey.pm
dracos: Doh, misread that entirely, thanks. Pushed a fixup for that

PR #5041, opened 2024-07-02
[Brent] Improve message in referral email.

dracos:
For https://mysocietysupport.freshdesk.com/a/tickets/4308
(Also the last commit on master above this wasn't properly reviewed, as urgent bug fix, but if you had a quick look at that too, thanks.)
[skip changelog]


chrismytton (approved):
This and the previous commit on master all look good to me 🙂

PR #5043, opened 2024-07-04
[Lincs] Switch to OSM for Geocoding

MorayMySoc:
We are getting results from Bing which show the same 3 locations when an address is not recognised.

OSM returns no addresses for the same not recognised addresses. 

https://mysocietysupport.freshdesk.com/a/tickets/4288

[skip changelog]


chrismytton (approved):
📍 👍

PR #5044, opened 2024-07-05
[Notts Police] Change body name to Immediate Justice Team

chrismytton:
And update the relevant templates to match.

I've overridden the whole `report/new/form_public_councils_text.html` template, which means the same message is now shown for reports in public and private categories.

Fixes https://github.com/mysociety/societyworks/issues/4405

## Notes to deployer

- [ ] Update the name of the body to "Immediate Justice Team" just before/after deploying this change

### DIFF
--- a/perllib/FixMyStreet/Cobrand/NottinghamshirePolice.pm
+++ b/perllib/FixMyStreet/Cobrand/NottinghamshirePolice.pm
@@ -30,7 +30,7 @@ sub admin_show_creation_graph { 0 }
 
 sub council_area_id { [ 2236, 2565 ] }
 sub council_area { 'Nottinghamshire'; }
-sub council_name { 'Nottinghamshire Police' }
+sub council_name { 'Immediate Justice Team' }
 
 =item * Any superuser or staff user can access the admin
 
@@ -52,7 +52,7 @@ sub area_check {
     foreach (@{$self->council_area_id}) {
         return 1 if defined $councils->{$_};
     }
-    return ( 0, "That location is not covered by Nottinghamshire Police." );
+    return ( 0, "That location is not covered by the Immediate Justice Team." );
 }
 
 sub disambiguate_location {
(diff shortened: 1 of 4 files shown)


templates/web/nottinghamshirepolice/report/new/councils_text_all.html
```diff
@@ -0,0 +1,18 @@
...
+[% UNLESS non_public_categories.$category;
+
+    tprintf(
+        loc('These will be sent to the <strong>Immediate Justice Team</strong> and also published online for others to see, in accordance with our <a href="%s">privacy policy</a>.'),
+        c.cobrand.privacy_policy_url
+    );
+
+ELSE;
+
+    loc('These will be sent to the <strong>Immediate Justice Team</strong> but not published online.');
```
dracos, line 13: This needs wider thinking - presumably if it's referred to a council, it will be made public, and there are various places (like here) the site will be saying this won't happen, so that will need careful thinking about.

templates/web/nottinghamshirepolice/report/new/councils_text_all.html
```diff
@@ -0,0 +1,18 @@
...
+
+END %]
+
+[% TRY %][% INCLUDE 'report/new/councils_extra_text.html' %][% CATCH file %][% END %]
```
dracos, line 17: Presumably we don't have this template for this cobrand so don't need this, but no harm leaving it in.

dracos (commented):
Looks fine, though raises a wider question

templates/web/nottinghamshirepolice/report/new/councils_text_all.html
dracos: Actually, perhaps this is the only place! In which case, needs changing :)

dracos: Perhaps "These will be sent to the <strong>Immediate Justice Team</strong>" and leave the question to the privacy statement or elsewhere? Or if it can be put here that it might be public if referred to a council (though presumably the police might moderate contents etc, I dunno).
(Doesn't need a `loc()`, not being translated)

dracos (approved):
Looks like good solution :)

PR #5046, opened 2024-07-05
[Central Beds] Change colour of tree asset

MorayMySoc:
Switches tree asset to green when seen on the map
https://mysocietysupport.freshdesk.com/a/tickets/4076

[skip changelog]

### DIFF
--- a/web/cobrands/centralbedfordshire/base.scss
+++ b/web/cobrands/centralbedfordshire/base.scss
@@ -119,4 +119,19 @@ html.js .form-section-preview {
     text-transform: none;
 }
 
+// Tree category has green assets
+#category_meta_message_HousingCommunalTrees {
+    .asset-spot:before {
+        content: "";
+        display: inline-block;
+        width: 10px;
+        height: 10px;
+        border: 2px solid #000;
+        background-color: #55BB00;
+        margin: 0 0.3em 0 0.2em;
+        border-radius: 100%;
+        vertical-align: -2px;
+    }
+}
+
 @import "../fixmystreet-uk-councils/societyworks-footer";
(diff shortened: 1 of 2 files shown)


web/cobrands/centralbedfordshire/base.scss
```diff
@@ -119,4 +119,19 @@ html.js .form-section-preview {
     text-transform: none;
 }
 
+// Tree category has green assets
+#category_meta_message_HousingCommunalTrees {
+    .asset-spot:before {
+        content: "";
```
dracos, line 125: You only need to override the bits you are changing - so the colour presumably.

dracos (changes requested):
Nice way to do it - though it makes these assets a bit bigger than normal as well as green - if we have to alter assets.js anyway, as you do, shall we just stick a new style in there, at the normal size? :)

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -592,6 +592,21 @@ fixmystreet.assets.centralbedfordshire.streetlight_stylemap = new OpenLayers.Sty
     'select': fixmystreet.assets.construct_named_select_style("${lighting_c}")
 });
 
+fixmystreet.assets.centralbedfordshire.style_default_green = new OpenLayers.Style({
```
dracos, line 595: Note the distinction in this file between functions/variables that need to be referred to outside the file (e.g. by the asset layer configuration), so given externally-reference-able names under `fixmystreet.assets...` and ones that don't need to be referred to externally.

dracos (approved):
Looks fine; whilst `fixmystreet.assets.centralbedfordshire.tree_stylemap` is (presumably) referred to in the configuration, the default green style isn't and so could be a local variable, or inline, but is fine as is :)

PR #5048, opened 2024-07-07
[Hylte] Add cobrand

jonkri:
Hylte kommun is a municipality in Sweden. Their cobrand is running on https://hylte.fixamingata.se/. The empty `public_update.html` file is intentional (we are currently not using this section of the inspect form).


jonkri:
Hylte kommun is a municipality in Sweden. Their cobrand is running on https://hylte.fixamingata.se/.

jonkri:
Hi! Is there anything we can do to help move this along? 😊

perllib/FixMyStreet/Cobrand/FixaMinGata.pm
```diff
@@ -87,43 +90,54 @@ sub geocoder_munge_results {
...
     return 0;
 }
 
 sub find_closest {
     my ( $self, $problem ) = @_;
+
     $problem = $problem->{problem} if ref $problem eq 'HASH';
-    return FixMyStreet::Geocode::OSM->closest_road_text( $self, $problem->latitude, $problem->longitude );
+
+    return FixMyStreet::Geocode::OSM::closest_road_text( $self, $problem->latitude, $problem->longitude );
```
dracos, line 140: I don't understand the change here, this makes the tests fail and I think breaks, because it means the OSM class is no longer being passed as the first argument which was a change we made in 6c585965d612fa1ac89a2deeaf80ce3c2d8c8a2d. I've rebased and pushed a fixup so you can check this before merging.

jonkri: Oh, I see! Thanks for noticing this, and the fixup! This can be merged!

Issue #4153, opened 2022-11-05
Don't position important UI elements behind Android banners


jonkri:
**Is your feature request related to a problem? Please describe.**

The in-app install banner on Android, offering to install the PWA, is positioned in the bottom of the app and covers the “Back” and “Continue” buttons.

The translation banner, offering to translate the PWA, has the same issue.

Users may not notice or care about these banners, and may not be ready to accept or dismiss them. Also, users may not be aware that these banners cover these (necessary, in order to create a report) UI elements.

**Describe the solution you'd like**

I'm thinking that we may need to move important enough UI elements (such as the “Back” and “Continue” buttons) from the bottom of the viewport, so that they are visible even if these banners are displayed.

I'm not sure what we would want the UI to look like.

- One potential drawback with putting the buttons at the top of the viewport is that they could be inconvenient to reach (for example when using a phone with one hand).
- One potential drawback with putting the buttons further down is that they then would cover part of the map.

**Describe alternatives you've considered**

I was thinking that we could change how the in-app install banner is displayed using the `beforeinstallprompt` event as described in [https://web.dev/customize-install/](https://web.dev/customize-install/). However, I haven't found a way to do something similar with the translation banner.

Disabling the translation banner would decrease the accessibility of the app.

**Additional context**

FixaMinGata has replaced the Android Cordova app with the [https://fixamingata.se/](https://fixamingata.se/) web app packaged using [PWABuilder](https://www.pwabuilder.com/). The Google Play listing is [here](https://play.google.com/store/apps/details?id=se.sambruk.fixamingata). We are considering releasing the same app on App Store as well.

Implementing crosshair way of reporting (used in the Cordova app) could perhaps be a possibility. (It was mentioned [here](https://groups.google.com/a/mysociety.org/g/fixmystreet/c/-E8kZALgvBE/m/sGwCxaFUAgAJ)).) This could perhaps be done at the same time.

dracos:
Yes, a custom install as you link to could be useful here - we could put the install link into the navigation, or show our own message somehow or similar, that didn't obscure the buttons.

The translate banner - you mean your browser is set to English, so visiting the website it is offering to translate it? Presumably it doesn't normally offer translation for a Swedish website in a Swedish browser? It appears at the top for me, interestingly, over the header. There's a "Translate" option under the three dots, so I guess it is something we could think about disabling potentially.

[...]

PR #5049, opened 2024-07-07
Avoid displaying the translation banner

jonkri:
Here's a suggestion to avoid displaying the translation banner as discussed [here](https://github.com/mysociety/fixmystreet/issues/4153#issuecomment-1365780882). We've been disabling the translation banner for about one and a half years in Sweden and no one has complained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
         - Convert all uploaded images to JPEGs.
         - Redirect after POST when creating reports. #4362
         - Include status change line in report update emails using auto response templates.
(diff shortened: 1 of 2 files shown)


dracos (approved):
I think I'm happy to go with this and see if it has any issues, I don't see why it should.

dracos (approved):

dracos:
I have merged this as ecde96bc08e7cf94469b57af6bbdaf4c5c5c9aca - I put the line a bit higher up in the file, alongside the "don't detect telephone numbers" which seemed to be in vaguely the same sort of area.

PR #4235, opened 2022-12-20
Incremental updates

jonkri:
Please check the following:

- [x] Whether this PR should include changes to any documentation, or the FAQ;
- [x] All cobrand-specific commits start their commit message with the cobrand in square brackets;
- [ ] Is new functionality tested? CodeCov will warn you about the diff coverage, but won’t complain about e.g. new files;
- [x] Will cobrand-specific changes require additional work to ensure consistent behaviour on www.fixmystreet.com? 
- [x] Are the changes tested for accessibility?
- [ ] Have you updated the changelog? If this is not necessary, put square brackets around this: [skip changelog]

Please check the contributing docs, and describe your pull request here.
Screenshots or GIF animations (using e.g. LICEcap) may be helpful.

Please include any issues that are fixed, using "fixes" or "closes" so that
they are auto-closed when the PR is merged.

Thanks for contributing!

PR #5050, opened 2024-07-07
[FixaMinGata] Add various updates

jonkri:
Some minor FixaMinGata-related changes. The empty `public_update.html` file is intentional (we are currently not using this section of the inspect form).
[skip changelog]

Related to #4235.

### DIFF
--- a/templates/web/fixamingata/about/faq-sv.html
+++ b/templates/web/fixamingata/about/faq-sv.html
@@ -37,15 +37,14 @@
     <dt>Vad är inte FixaMinGata till för?</dt>
     <dd>
         FixaMinGata är inte lämpligt för andra typer av problem än de som anges ovan. Du måste kontakta din kommun eller fastighetsägare direkt för problem i stil med:
-        <ul>
+        <ul style="margin-bottom: 0;">
             <li>Brådskande och akuta problem</li>
             <li>Oljud i din närmiljö</li>
-            <li>Brand och rök</li>
             <li>Förslag på nya vägar, farthinder, etc.</li>
             <li>Klaga på dina grannar</li>
             <li>Klaga på din kommun</li>
-            <li>Droger, djurplågeri, stöld, eller andra kriminella aktiviteter.</li>
         </ul>
+        <p style="margin-bottom: 2em;">Vid pågående brott och vid fara för liv, hälsa eller egendom ska du ringa 112.</p>
     </dd>
 
     <dt>Hur använder jag e-tjänsten?</dt>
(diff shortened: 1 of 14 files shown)


jonkri:
Hi! Is there anything we can do to help move this along? 😊

templates/web/fixamingata/about/faq-sv.html
```diff
@@ -1,7 +1,5 @@
 [% INCLUDE 'header.html', title => loc('Frequently Asked Questions'), bodyclass => 'twothirdswidthpage' %]
 
-[% INCLUDE 'about/_sticky-button-content-navbar.html' %]
```
dracos, line 3: This makes the navigation easily navigable on mobile, see www.fixmystreet.com/faq - is there a reason you removed it, or can we leave it in?

templates/web/fixamingata/footer_extra.html
```diff
@@ -34,30 +34,30 @@ <h2 class="mysoc-footer__site-name">FixaMinGata</h2>
             <div class="col-sm-5">
                 <nav class="mysoc-footer__links">
                     <ul>
-                        <li><[% IF c.req.uri.path == '/' %]span[% ELSE %]a href="/"[% END %] class="report-a-problem-btn"
+                        <li><[% IF c.req.uri.path == '/' %]span[% ELSE %]a href="[% base %]/"[% END %] class="report-a-problem-btn"
```
dracos, line 37: Can I ask why these are added? it feels like it might be a problem somewhere else that needs fixing?

web/cobrands/fixamingata/base.scss
```diff
@@ -231,6 +235,18 @@ $grid-breakpoint-sm: $mysoc-footer-breakpoint-sm;
...
+
+.mobile #geolocate_link {
+    margin-top: 15px !important;
+}
+
 #pc-hint {
     display: none;
 }
+
+.sticky-sidebar li:last-child {
```
dracos, line 250: Think this will be pulled in by a rebase from core.

templates/web/fixamingata/about/faq-sv.html
jonkri: Please leave it in, thank you!

templates/web/fixamingata/footer_extra.html
jonkri: I don't remember why that was added, I'm afraid. 🤔 Let's remove it from this pull request!

web/cobrands/fixamingata/base.scss
jonkri: Please remove it, thanks!

dracos (approved):
Okay, I've pushed with the suggested changes - do you want to try it out to check it's all okay before I merge it?

jonkri:
> Okay, I've pushed with the suggested changes - do you want to try it out to check it's all okay before I merge it?

Thanks a lot! 🙂 `margin-bottom: 2em;` on the sticky sidebar would make it look better [next to the content div on the FAQ page](https://beta.fixamingata.se/faq).

jonkri:
How about adding...

```scss
.sticky-sidebar {
    margin-bottom: 2em;
}
```

... to `web/cobrands/fixamingata/layout.scss`?

dracos:
Ah, I wondered why it looked okay on www.fixmystreet.com/faq and it looks like we have no bottom margin on .content but a top margin on the footer instead there. But yes, happy to stick that in to handle it this way for now.

jonkri:
> Ah, I wondered why it looked okay on [www.fixmystreet.com/faq](http://www.fixmystreet.com/faq) and it looks like we have no bottom margin on .content but a top margin on the footer instead there. But yes, happy to stick that in to handle it this way for now.

Aha, great! 💡 I'll add it to my todo list to switch to a footer top margin instead when I change the layout. 🙂

dracos:
There's a test failure because of the removal of the `always_view_body_contribute_details`, so I'll just remove the corresponding test, hope that's okay.

jonkri:
> There's a test failure because of the removal of the `always_view_body_contribute_details`, so I'll just remove the corresponding test, hope that's okay.

Absolutely, I don't think we will use that anymore. Thank you!

PR #5053, opened 2024-07-08
[Bexley] Surface only one waste service if multiple share a name.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4346

This also includes a small change that removes the word 'rotation' before 'schedule'.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -227,11 +227,22 @@ sub bin_services_for_address {
 
     my @site_services_filtered;
 
+    my %seen_containers;
+
     for my $service (@$site_services) {
         next if !$service->{NextCollectionDate};
 
         my $container = $containers->{ $service->{ServiceItemName} };
         next unless $container;
+
+        # There may be duplicate container types; skip if container has been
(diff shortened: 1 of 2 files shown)


neprune:
> Hi @neprune @chrismytton , Have pushed what I think is a simpler fix 🤞 though we may need to revisit this or put it on hold until we get clarification on the question I asked [here](https://3.basecamp.com/4020879/buckets/35109031/todos/7319188833#__recording_7587186843)

Nice, I thought doing the filtering after the original iteration would be more readable but your fixup reads nicely to me, and is obviously more performant.

Not sure if it matters, but if we wanted to be consistent in which service we choose when two clash, we could order the iteration by service code alphabetically, or similar.

nephila-nacrea:
> Not sure if it matters, but if we wanted to be consistent in which service we choose when two clash, we could order the iteration by service code alphabetically, or similar.

Yeah, I have wondered if we ought to be sorting services *before* we do any filtering - I'll see what Bexley come back with because some services are 'identical' containers but actually have different collection dates :-/

nephila-nacrea:
> The rotation schedule text change has been done in #5014, so that doesn't need to be done in this PR, but the rest looks good to me!

Bluh, I will remove!

PR #5054, opened 2024-07-08
[Bexley] Add link to address lookup page to waste navbar.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4406

navbar

### DIFF
--- a/templates/web/bexley/main_nav_items.html
+++ b/templates/web/bexley/main_nav_items.html
@@ -1,4 +1,6 @@
 [%~ IF bodyclass.match('waste') OR (problem AND problem.cobrand_data == 'waste') ~%]
+    [% INCLUDE navitem uri='/waste' label='Look up address' %]
+
     [%~ IF c.user_exists ~%]
         [%~ INCLUDE navitem uri='/my' label=loc('Your account') ~%]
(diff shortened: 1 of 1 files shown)


templates/web/bexley/main_nav_items.html
```diff
@@ -1,4 +1,6 @@
 [%~ IF bodyclass.match('waste') OR (problem AND problem.cobrand_data == 'waste') ~%]
+    [% INCLUDE navitem uri='/waste' label='Lookup address' %]
```
nephila-nacrea, line 2: I think this needs to be 'Look up address' (with space) - or 'Address lookup' (because look up is a verb in the former and a noun in the latter, or something). I can do a quick fix 👍 (Did Bexley ever specify what text they wanted exactly?)

nephila-nacrea (approved):

templates/web/bexley/main_nav_items.html
neprune: Thanks 🙏 . No, I couldn't find any spec for the text they wanted.

nephila-nacrea:
Merged to staging.

PR #5055, opened 2024-07-08
[Bexley] Update text for no known waste collections.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4400

Screenshot:

screenshot

Wireframe spec:

wireframe

### DIFF
--- a/templates/web/bexley/waste/_bin_days_no_collections.html
+++ b/templates/web/bexley/waste/_bin_days_no_collections.html
@@ -1,1 +1,9 @@
-<p>There are currently no collections at this property. Please <a href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?template=no-collections">use this form to report a missed collection</a>.</p>
+<p><b>Sorry, we are unable to find any rubbish and recycling collections associated with this property.</b></p>
+
+<p>If this property is a business or organisation, we are unable to provide collection dates, but you can still tell us about a missed collection.</p>
+
+<p>If you are a residential property, you can report this to the Waste Team to investigate.</p>
+
+<p>Complete the enquiry form below to continue (takes around 2 minutes).</p>
+
+<a class="govuk-button" href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?template=no-collections">Missed collection enquiry form</a>


templates/web/bexley/waste/_bin_days_no_collections.html
```diff
@@ -1 +1,9 @@
-<p>There are currently no collections at this property. Please <a href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?template=no-collections">use this form to report a missed collection</a>.</p>
+<p><b>Sorry, we are unable to find any rubbish and recycling collections associated with this property.</b></p>
+
+<p>If this property is a business or organisation, we are unable to provide collection dates, but you can still tell us about a missed collection.</p>
+
+<p>If you are a residential property, you can report this to the Waste Team to investigate.</p>
+
+<p>Complete the enquiry form below to continue (takes around 2 minutes).</p>
+
+<a class="govuk-button" href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?template=no-collections">Missed collection enquiry form</a>
```
nephila-nacrea, line 9: This will take me to the enquiry page, which will show a couple of radio buttons, but regardless of which one I select, I get redirected back to the bin days page with no explanation.

What does the `template=no-collections` bit do exactly? I can't find 'no-collections' outside of this file and waste_bexley.t.

neprune: Great question. You can probably tell I just assumed the existing link was good and changed the styling. Will take a look...

neprune: So the template query param is telling the waste controller to use `templates/web/bexley/waste/enquiry-no-collections.html`. As to why it just returns at present, still looking...

neprune: Seems to work fine on staging: https://bexley.staging.fixmystreet.com/waste/10023301216. So probably just some missing configuration with our local setups.

dracos: The answers to the radio buttons in that template are category names; if those categories don't exist in your body, then it will redirect back to the bin day page.

nephila-nacrea (approved):

nephila-nacrea:
Deployed on staging.

PR #5058, opened 2024-07-09
[Brent] Add reference and date to notification emails

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/4272

[skip changelog]


dracos (commented):
The other-reported* templates all seem to be very similar mostly, so we should probably think about consolidating them somehow, but doesn't have to be for this :) Chatted about timestamp in other-reported elsewhere.

MorayMySoc:
Sorry - you got to this quicker than I thought and rebased all fixups into the one

dracos (approved):

PR #5062, opened 2024-07-10
Email changes for 'no collections' enquiries

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4394.

This change includes moving to bespoke templates for Bexley waste, and removing Bexley logic from default templates.

[skip changelog]


perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -44,6 +44,12 @@ sub geocode_postcode {
 
 sub disable_resend_button { 1 }
 
+# Bexley uses a separate waste/submit.(html|txt) template for waste reports
```
nephila-nacrea, line 47: I'm not happy with my method names, and am not sure if I've overengineered this (rather than just sticking 'if bexley && waste' checks in Email.pm, but that doesn't seem good either).

dracos: I wouldn't change the code at all in this commit, I think; I would have an IF at the top of the fixmystreet.com/submit.txt and html file specially for Bexley - which matches the `other-reported` way of doing it, and we don't know at this point if we'll need anything else.

perllib/FixMyStreet/SendReport/Email.pm
```diff
@@ -63,14 +63,29 @@ sub get_template {
 
 sub send_from {
     my ( $self, $row ) = @_;
+
+    my $cobrand = $row->get_cobrand_logged;
```
dracos, line 67: I think perhaps an easier way of doing this would be to have a SendReport::Email subclass (there are already a few), perhaps SendReport::Email::Bexley, that would use the do_not_reply/council_name without having to do any checking; you could then set the send method to for this category in the admin.

dracos (changes requested):
A good start - I've suggested a couple of different ways to perhaps achieve the same outcome, let me know your thoughts.

perllib/FixMyStreet/Cobrand/Bexley.pm
nephila-nacrea: OHHHH
There's a bunch of if-else PROCESS statements 🤦‍♀️

nephila-nacrea:
@chrismytton , @jacquelinebylau 
Important note:
When this PR is put on staging and live, the 'Business or organisation missed collection enquiry' and 'Missed collection enquiry' categories will need to have their send methods set to 'Email::Bexley'.

templates/email/bexley/waste/_council_reference.txt
```diff
@@ -0,0 +1 @@
+Your reference number is [% problem.id %].
```
dracos, line 1: Presumably this needs a .html equivalent

dracos (changes requested):
Only one missing template I think

templates/email/bexley/waste/_council_reference.txt
chrismytton: I think the HTML version was added in 14843278c9632b5725d5156840007fa663c53dd6, so was already in place before this pull request.

dracos (approved):

PR #5063, opened 2024-07-10
[Bexley] Use force_arrayref on Whitespace site contracts

chrismytton:
Fixes these errors that are appearing in the logs:

```
Caught exception in FixMyStreet::App::Controller::Waste->process_report_data "Not an ARRAY reference at /data/vhost/www.fixmystreet.com/fixmystreet-2024-07-10T10-44-33/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm line 1015.", referer: /waste/10011843743/report
```

[Basecamp todo](https://3.basecamp.com/4020879/buckets/35109031/todos/7591740937)

### DIFF
--- a/perllib/Integrations/Whitespace.pm
+++ b/perllib/Integrations/Whitespace.pm
@@ -208,7 +208,9 @@ sub GetSiteContracts {
 
     my $res = $self->call('GetSiteContracts', sitecontractInput => ixhash( Uprn => $uprn ));
 
-    return $res->{SiteContracts}->{SiteContract};
+    my $contracts = force_arrayref($res->{SiteContracts}, 'SiteContract');
+
+    return $contracts;
 }
 
 sub GetFullWorksheetDetails {


chrismytton:
@nephila-nacrea I have to admit I don't fully understand the issue, I think it's related to the XML library we're using, so feels like maybe it would be possible to fix it at a lower level so that single item lists are always returned as lists, but I imagine there's a lot more nuance to it than that!

dracos:
There's two separate XML things here.
We use XML::Simple in a few places, and that has configuration arguments as to whether it turns single entries into hashes or keeps them as arrays (either everywhere, or you can specify element names) which we do use.
But SOAP::Lite has its own handling of this - https://github.com/redhotpenguin/perl-soaplite/blob/66be2ded5b45e36763b2501a3d1f25e804fcf331/lib/SOAP/Lite.pm#L2421-L2439 - where you can see it always starts with a hash (or single scalar etc), then if it gets another child turns it into an array. That seems a bit harder to override, hence our force_arrayref.
In this particular case, if it's always the case for every call, it seems like it would make sense to refactor it into the central Whitespace `call` function - pass in the results value, and do it there, and then you'll know it's always necessary - if that makes sense.

PR #5073, opened 2024-07-19
[Brent] Bulky waste missed can be reported if no event by 6pm

MorayMySoc:
If the bulky waste collection hasn't been marked as missed from Echo can be reported as missed/collected after 6pm by the booker.

https://mysocietysupport.freshdesk.com/a/tickets/3500

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -1654,6 +1654,25 @@ sub waste_reconstruct_bulky_data {
     return $saved_data;
 }
 
+=item bulky_open_overdue
+Returns true if the booking is open and after 6pm on the day of the collection.
+=cut
+
+sub bulky_open_overdue {
+    my ($self, $event) = @_;
+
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
Sorry, only mentioned in Slack not here, but looks like might have been forgotten. The ticke says if he day has passed, not 6pm (so midnight the start of the next day).

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -1668,8 +1670,10 @@ sub bulky_open_overdue {
 
 sub _bulky_collection_overdue {
     my $collection_due_date = $_[1]->{date};
-    $collection_due_date->truncate(to => 'day')->set_hour(18);
-    my $today = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
+
+    $collection_due_date->add(days => 1)->truncate(to => 'day')->set_hour(0);
```
dracos, line 1674: truncate to day should set the hour 0, so set_hour shouldn't be needed?

dracos (approved):

PR #5076, opened 2024-07-24
Add toggling via category filter legends.

dracos:
This lets you toggle all the categories in a group on and off at once.


MorayMySoc (changes requested):
This works well on FF until you select 'Everything' and then the legends are no longer active for clicking. A soft refresh, so 'Everything' is still selected will allow the legends to become active again. I'm not sure if it's the same on Chrome/Safari (I was focusing on the 'All reports' page)

dracos:
Thanks, was due to the items being regenerated at the change. Hopefully fixed that now.

MorayMySoc (approved):
Looks good

PR #5077, opened 2024-07-24
[Bexley] Associate Whitespace missed collection report with correct UPRN

chrismytton:
Change to ensure missed collection reports are always made against the UPRN where the service/collection was found.

Fixes https://github.com/mysociety/societyworks/issues/4439


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -193,14 +193,17 @@ sub bin_services_for_address {
     my $self = shift;
     my $property = shift;
 
-    my $site_services = $self->whitespace->GetSiteCollections($property->{uprn});
+    my @site_services;
```
dracos, line 196: Looks like this works fine, but if I'm reading it right - you either have the child services, or if they're empty, the parent services, you never have both. So I think you didn't need to add the UPRN to every result, you only needed to note either the child or the parent UPRN depending on whether the child had services or not. Guess up to you if you want to simplify it or keep it as is.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -435,57 +439,58 @@ sub _remove_service_if_assisted_exists {
...
+                    $ws->{WorksheetStartDate} eq WHITESPACE_UNDEF_DATE ?
+                    '' : $ws->{WorksheetStartDate}
+                ),
+                will_be_completed => (
+                    $ws->{WorksheetEscallatedDate} eq WHITESPACE_UNDEF_DATE ?
+                    '' : $ws->{WorksheetEscallatedDate}
+                ),
+                latest_comment =>
+                    ( $latest_comment ? $latest_comment->text : '' ),
+                uprn => $uprn,
```
dracos, line 489: To just check I understand this - I see this is now looping through both UPRNs if needed, but the UPRN stored here isn't actually used anywhere?

dracos (commented):
I think the logic of the first commit can be simplified if you wish to, looks fine otherwise

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
chrismytton: Good point! Have added a fixup commit to simplify.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
chrismytton: Ah yeah that's not needed, have removed in a fixup.

dracos (approved):

PR #5078, opened 2024-07-24
[Bexley][WW] Couple of minor text changes

chrismytton:
Fixes https://github.com/mysociety/societyworks/issues/4444

Fixes https://github.com/mysociety/societyworks/issues/4450

### DIFF
--- a/t/app/controller/waste_bexley.t
+++ b/t/app/controller/waste_bexley.t
@@ -310,11 +310,11 @@ FixMyStreet::override_config {
             'A blue recycling box collection has already been reported');
         $mech->content_contains('Reported on: N/A');
         $mech->content_contains('Will be completed by: N/A');
-        $mech->content_contains('Comments: N/A');
+        $mech->content_contains('Action: Please leave your waste out, our contractor will return soon');
 
         $mech->content_contains(
             'A green recycling box collection has already been reported');
-        $mech->content_contains('Comments: Preexisting comment');
+        $mech->content_contains('Action: Preexisting comment');
 
         subtest 'service_sort sorts correctly' => sub {
             my $cobrand = FixMyStreet::Cobrand::Bexley->new;
--- a/t/app/controller/waste_bexley.t
+++ b/t/app/controller/waste_bexley.t
@@ -307,13 +307,13 @@ FixMyStreet::override_config {
 
         note 'Missed collection displays';
         $mech->content_contains(
-            'A blue recycling box collection has already been reported');
+            'A blue recycling box collection has already been reported as missed');
         $mech->content_contains('Reported on: N/A');
         $mech->content_contains('Will be completed by: N/A');
         $mech->content_contains('Action: Please leave your waste out, our contractor will return soon');
 
         $mech->content_contains(
-            'A green recycling box collection has already been reported');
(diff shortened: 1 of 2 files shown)


chrismytton:
Oops failing test, will look!

dracos (approved):

chrismytton:
Added an extra commit to take the text in the missed report status section from the config, rather than from the latest update.

Should fix https://github.com/mysociety/societyworks/issues/4444 and https://github.com/mysociety/societyworks/issues/4450

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -463,11 +466,8 @@ sub _missed_collection_reports {
             = $report->get_extra_field_value('service_item_name') // '';
         next if $missed_collection_reports{$service_item_name};
 
-        my $latest_comment
-            = $report->comments->search(
-                {},
-                { order_by => { -desc => 'id' } },
-        )->first;
+        my $status = $ws->{WorksheetStatusName};
```
dracos, line 469: This will always be Open by the if in line 452, that right?

dracos (commented):
Don't quite follow, as the linked ticket talks about adding a response template rather than config, but assuming this is what's required.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
chrismytton: No, that was not right! Clearly wasn't thinking straight yesterday.

chrismytton: Have pushed a fixup commit which uses the fms_state matched to the report->state to determine which text to display.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -446,6 +446,11 @@ sub _missed_collection_reports {
     my $config = $self->feature('whitespace');
     my $state_mapping = $config->{missed_collection_state_mapping};
 
+    # Create a lookup hash mapping fms_state to text
```
dracos, line 449: Can we assume there won't be more than one thing with the same `fms_state` (or that it'll pick one at random if there is)?

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -446,6 +446,11 @@ sub _missed_collection_reports {
...
+    my %fms_state_to_text = map {
```
dracos, line 450: As you don't need the keys, you could have `map { $_->{fms_state} => $_->{text} } values %$state_mapping` here.

dracos (commented):
Looks fine, assuming we don't mind about the possibility of duplicate

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
chrismytton: Good point, there are lots of things for the same fms state, so this isn't going to work.

chrismytton:
Have abandoned the code changes for pulling the status text from config. Just going with an amended text change for now.

PR #5080, opened 2024-07-25
Added $body-font variable to input, select and textarea elements

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4415

UI Improvement: Currently, FMS and cobrands use the system default. Depending on the font the cobrand and system use, this can feel quite disjointed.

### Preview:
Screenshot 2024-07-25 at 08 54 15

[Skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -236,6 +236,7 @@ select,
 input,
 textarea {
   font-size: 99%;
+  font-family: $meta-font;
 }
 
 // To deal with bug from drop-down being wider than holder


dracos (commented):
I'm wondering if this should be `$meta-font`, not `$body-font`. I dunno. If you think this is better everywhere, happy to go with it; I was a bit curious as to whether it's actually better or if it's better to look like other controls the user will be familiar with, but it looks like big UK sites all have custom font in their inputs now anyway.

lucascumsille:
@dracos Good point, to be fair I haven't deal with many cobrands using the `$meta-font`, but I'm on board with that. I added an `if` where basically we use the meta if has been declared.
Another alternative would be to do it at the very top, so if no meta has been declared then: `meta-font: $body-font`

Let me know what you think

dracos (commented):
`$meta-font` is always defined; if it's not set up in a cobrand, then `sass/_base.scss` already does set `$meta-font: Helmet, Freesans, sans-serif !default;` at the top. So you don't need the if. It's used in `small`/lines about reports, and other "meta" places like that, so I guess its whether we think that's also something to use for inputs, or not. I'm not sure.

lucascumsille:
@dracos I pushed a fixup, let's use the `meta-font` is available let's take advantage of it =)

dracos (approved):

PR #5081, opened 2024-07-25
[Merton] Various text tweaks

dracos:
[skip changelog]
5 of the 6 changes requested at https://3.basecamp.com/4020879/buckets/36310026/todolists/7082041259 (except the last).

### DIFF
--- a/templates/web/base/waste/bulky/intro_amend.html
+++ b/templates/web/base/waste/bulky/intro_amend.html
@@ -6,6 +6,8 @@
     <li>In order to help us with your collection, we will ask you to optionally add pictures of the items to be collected and the location</li>
     <li>Before confirming your booking, check all the info provided is correct</li>
     [% UNLESS cobrand.call_hook('bulky_enabled_staff_only') %]
+        [% IF cobrand.moniker == 'peterborough' %]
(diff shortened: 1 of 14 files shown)


templates/web/base/waste/bulky/intro_amend.html
```diff
@@ -6,6 +6,8 @@ <h3>Before you amend your booking</h3>
     <li>In order to help us with your collection, we will ask you to optionally add pictures of the items to be collected and the location</li>
     <li>Before confirming your booking, check all the info provided is correct</li>
     [% UNLESS cobrand.call_hook('bulky_enabled_staff_only') %]
+        [% IF cobrand.moniker == 'peterborough' %]
```
chrismytton, line 9: Assume it's fine to remove this from the other live cobrands?

chrismytton (approved):

templates/web/base/waste/bulky/intro_amend.html
dracos: Yeah, I'm assuming so, as the same text is in intro.html (for new bulky collections) and has this condition there, so I think it was messed. Better to ere on side of not saying people can have refunds anyway.

PR #5083, opened 2024-07-30
Allow send-reports with a specific ID

nephila-nacrea:
[skip changelog]


dracos (approved):
Aha, thanks for spotting this!

PR #5085, opened 2024-07-30
[Waste] Create symbol SVG for waste images.

dracos:
Turns out you can use an external SVG and still style it, by using `use` and what amounts to an SVG sprite file. This moves to using that so all containers are in one SVG with 5 "symbols" inside it. It also fixes a tiny issue with one shadow on a sack and makes sure bin page images are consistent. [skip changelog]


MorayMySoc (changes requested):
Firefox 128.0.3 (64-bit) (left) seems to not be picking up the colours, compared with Chromium. Not sure why the difference

image

dracos:
Guess can revisit in future, maybe, not much point at present.

PR #5086, opened 2024-07-30
[Bexley] Make Whitespace API calls in parallel

chrismytton:
To speed up the loading of a bin days page this changes some of the Whitespace API calls to be made in parallel, and then cached.

Closes https://github.com/mysociety/societyworks/issues/4384

[skip changelog]

## Todo

- [ ] How long are things cached for?
- [x] Should we be cacheing worksheet lookups? Might want those to be up-to-date
- [x] Tests not passing

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -56,6 +56,23 @@ use constant WHITESPACE_UNDEF_DATE => '0001-01-01T00:00:00';
 use constant MISSED_COLLECTION_SERVICE_PROPERTY_ID => 68;
 use constant REQUEST_SERVICE_PROPERTY_ID => 69;
 
+sub fetch_whitespace_data {
+    my ($self, $method, $uprn) = @_;
+
+    my $c = $self->{c};
+
+    my $data = $self->whitespace->call_api(
+        $c,
+        "bexley",
+        "bin_days_page:$uprn",
+        0,
+        GetSiteInfo        => [$uprn],
+        GetSiteCollections => [$uprn],
+    );
+
+    return $data->{"$method $uprn"};
+}
+
 sub waste_fetch_events {
     my ( $self, $params ) = @_;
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -56,6 +56,25 @@ use constant WHITESPACE_UNDEF_DATE => '0001-01-01T00:00:00';
 use constant MISSED_COLLECTION_SERVICE_PROPERTY_ID => 68;
 use constant REQUEST_SERVICE_PROPERTY_ID => 69;
 
+sub fetch_whitespace_data {
+    my ($self, $method, $uprn) = @_;
+
+    my $c = $self->{c};
+
+    # Check if we've already fetched the data for this UPRN
+    if (!$c->stash->{whitespace_data} || $c->stash->{whitespace_data_uprn} ne $uprn) {
```
dracos, line 65: Couple of questions/more info:
1. call_api is already caching returned results (by using waste_cache_get/set), so do you need to be storing anything in the stash at all - can you just call call_api and return the data, letting it do the caching? Or is the stash doing something else here?
2. Rather than have a `whitespace_data_uprn` check, would it be better to have `whitespace_data` be a UPRN-keyed hash? Then you could have if $c->stash->{whitespace_data}{$uprn} doesn't exist, call_api and store it there, and return $c->stash->{whitespace_data}{$uprn}{"$method $uprn"}. Guess it doesn't make much difference.

davea (approved):

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -56,6 +56,26 @@ use constant WHITESPACE_UNDEF_DATE => '0001-01-01T00:00:00';
...
 
+sub fetch_whitespace_data {
+    my ($self, $method, $uprn) = @_;
+
+    my $c = $self->{c};
+
+    my $async = $c->action eq 'waste/bin_days'
+        && $c->req->method eq 'GET';
+
+    my $data = $self->whitespace->call_api(
```
dracos, line 67: Note that this call stores its results in the user's session, so a corresponding clear_cached_lookups_property call needs to be added so that it's cleared in appropriate places (e.g. when the bin day page is visited anew).

PR #5088, opened 2024-07-31
[Merton] Calendar changes

dracos:
Add a help page for people to subscribe, because calendars on the internet are still rubbish.
Plus couple of issues, Merton want to reduce the time the calendar looks ahead, and fix a bug where services lack a task ID.[skip changelog]


davea (changes requested):
image

On narrow mobile screens the link can be too long and obscures the clipboard button. Maybe a line break between them?

Could "the link" in "press and hold the link"/"right-click / Ctrl-click the link," be an actual link reworded as "this link"?

dracos:
"Maybe a line break between them?" - my flexbox was missing a wrap (flexbox is nowrap by default, which is bit annoying). But then I decided no need to show the link at all really, given your second idea, so have simplified it a bit and made it a link as suggested

davea (approved):

dracos:
Just to note I tweaked this to move the library to the normal extra_js because I realised that if you went straight to /calendar (so the async template intercepted) then I wanted it already present for when the page was pulled in.

dracos:
Changed after to move loading wrapper higher so that it included the JS needed to run if the page was pulled in from the async loading page (if someone went straight to /calendar).

PR #5089, opened 2024-07-31
[Merton] Send through all payment details

dracos:
[skip changelog]
If we're confirming payment by an update (SLWP), store the payment as extra data on the update, so that it'll be sent through to the backend. Also store *all* payments, so for amendments the full history is sent through with the amended new booking.


MorayMySoc (approved):
Looks good to me

PR #5091, opened 2024-08-01
Allow click on Everything to unselect everything, and improve legend toggle

dracos:
"Everything" ideally would change to a checkbox (or even three-state to show when some selected), but this will be a workaround for the present.
Legend moves to option instead of checkboxes to improve history behaviour.
[skip changelog]


MorayMySoc (approved):
Works for me in FF and Chromium nicely

PR #5093, opened 2024-08-02
Move admin nav to sidebar

dracos:
Moves the navigation to a sidebar, and makes it prettier


lucascumsille:
@dracos I really like it, it looks more tidy and easier to scan.

As a suggestion for the hover effect what do you think if we include a variable for the colour and background, they would have by default `$button-primary-text` and `$button-primary-bg-top`. Currently not all the cobrands are overriding those variables, however that is being [tackled here](https://github.com/mysociety/fixmystreet/pull/4943/files), but we stopped working on that PR because we ended just making the change for Surrey, but I could add it to this sprint and finish it.

PR #5094, opened 2024-08-02
[Camden] Send location data with title to Confirm

MorayMySoc:
https://3.basecamp.com/4020879/buckets/36530864/todos/7607947511#__recording_7668470194


MorayMySoc:
Going to do this in open311https://github.com/mysociety/open311-adapter/pull/352

PR #5096, opened 2024-08-02
[Surrey] Add HappyOrNot

davea:
For https://github.com/mysociety/societyworks/issues/4420

[skip changelog]

### DIFF
--- a/templates/web/surrey/footer_extra.html
+++ b/templates/web/surrey/footer_extra.html
@@ -36,6 +36,9 @@
             </li>
         </ul>
     </div>
+    <div class="container">
+        <div id="happy-or-not-smiley-digital-embed"></div>
+    </div>
 </footer>
 [% PROCESS '_footer_main_nav.html' %]
 [% PROCESS '_footer_extra_powered_by.html' %]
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
This 3rd-party JS directly uses the `?.` operator, which only exists in Chrome 80+, Safari 13+ etc, not Cypress anyway, so breaks the Cypress tests @MorayMySoc has added in https://github.com/mysociety/fixmystreet/pull/5118 (and any old browser).

The domain could/should be added to the block list for Cypress, but then the second line here will need at least `window.HappyOrNot && ` adding at the start to stop it erroring when the object is missing.

I also don't know why the language is set to Finnish (fi-FI) ;)

dracos:
Test files should not be loading/calling external 3rd-party JS at all, think how many times these tests might be being run, so please add it to the block list as well (which should be fine given the change you have already made to deal with it not loading/erroring).

MorayMySoc:
I think that should be it in 1ab153f3324976b0a34f5cb9ff632b78215951dc ?

dracos:
Ah, sorry, you did say you'd put it direct on the surrey-cobrand branch. Guess would have been good to be on this PR as well :)

PR #5101, opened 2024-08-06
[Echo] Move photo shrinking to update hook.

dracos:
By the time the open311_pre_send hook is reached, the report has had Open311-only extra fields set on it and we are not in a transaction, so we don't want to update the database at this point. The open311_update_missing_data hook is run within a transaction, so is the best place to make any necessary database updates, before the report is sent.

Only concern I can think of is if the photo shrinking is very slow and it's stuck in a transaction for a while, but hopefully that isn't the case?

[skip changelog]


neprune (approved):
Nice, and yeeah, I'm pretty sure the shrinking fast, so there shouldn't be any issues there.

PR #5102, opened 2024-08-06
[Merton] Stop script during Echo downtime.

dracos:


neprune (approved):
LGTM. Wasn't sure if we needed a start time buffer as well in case the script runs moments before downtime, but I think we're alright with cron timing and downtimes starting on the hour (unless we think 5 mins beforehand is worth forgoing).

PR #5105, opened 2024-08-08
Add plain text preview links to /_dev/email

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/4459.


neprune (approved):
Ah, I hadn't even realised it already existed as a query string parameter. LGTM!

PR #5110, opened 2024-08-14
[Oxfordshire] Upgrade/improve lite-youtube-embed.

dracos:
Upgrade to latest release (0.3.3), make it a module, so we can rely on things being present inside, and add a fallback link inside the custom element. Also add a closure-compiler minified version, with a polyfill for fromEntries, removing other added boilerplate. [skip changelog]


MorayMySoc (approved):
Seems good to me

PR #5112, opened 2024-08-14
[Merton] Add fallback check

dracos:
Needed to move a few things into the model, but worked through them all in the end, I think [skip changelog]
Means you can call payment confirmation (for garden or bulky) outside the controller flow, should that be needed elsewhere in future.


neprune (approved):
Nice, LGTM

PR #5113, opened 2024-08-15
Added max-width to govuk-select class

lucascumsille:
**Fixes:**
The multi-select has lost its maximum width at some point so e.g. if you’re on the dashboard for a council and pick a number of different wards, see how it gets bigger and bigger (or on a body admin page if you pick a number of areas to cover, etc)
Screenshot 2024-08-14 at 22 34 13

**Preview**
This PR adds `max-width` to  the class `.govuk-select` 

https://github.com/user-attachments/assets/58239b8c-2fa9-4834-8e90-71f2bbf8c5b8

### DIFF
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -297,8 +297,7 @@
 
 .dashboard-filters {
     .multi-select-button {
-        height: 28px;
-        line-height: 100%;
+        max-width: 10em;
     }
 }
 
--- a/web/cobrands/sass/_dashboard.scss
+++ b/web/cobrands/sass/_dashboard.scss
@@ -299,6 +299,11 @@
     .multi-select-button {
         max-width: 10em;
     }
+
+    .form-control {
+        // To make them have the same height as the .multi-select-button
+        height: 3.3em;
+    }
 }
 
 .dashboard-options-tabs {
--- a/web/cobrands/sass/_govuk.scss
+++ b/web/cobrands/sass/_govuk.scss
@@ -15,9 +15,11 @@
 }
 
 .govuk-select {
+  // This class also affects .multi-select-button
   font-family: $body-font;
   font-size: 0.875rem;
   line-height: 150%;
+  padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);
   @media (min-width: 48em) {
     font-size: 1rem;
   }
--- a/web/cobrands/sass/_report_list.scss
+++ b/web/cobrands/sass/_report_list.scss
@@ -67,7 +67,6 @@ $govuk-input-focus-box-shadow: #fd0 !default;
 
   .multi-select-button {
     box-sizing: border-box;
-    padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);
     vertical-align: top;
   }
 


dracos (changes requested):
_report_lists.scss already has things like:
```
.multi-select-button {
    box-sizing: border-box;
    padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);
    vertical-align: top;
}
```
Can that now be removed/made more generic/used here instead?
Note the `flip` there to handle RTL languages as well.

Why is the `max-width` not 20em, as it was on .multi-select-button before it was removed - is there a reason for it to be 200px, where has that figure come from? Also, is that something we want on all selects, not just multi-select-button ones? I imagine there might be selects in use that are fine to be wider (e.g. the address one without JS is too narrow at 200px).

20px padding for the arrow seems a tiny bit too small:
image

Lastly, I see _waste.scss includes a global `govuk-select` change which changes all govuk-selects, but only in those cobrands that have waste, which seems odd behaviour to have - should this be restricted to waste pages? If not, should it be in the generic CSS for all cobrands in some way? Just looking for an opportunity as always to simplify the CSS, not just add more lines on top :)

lucascumsille:
@dracos  I was planning to do the tidy up and simplification [here](https://github.com/mysociety/societyworks/milestone/99), ideally when we finish with that milestone all the selects and multi-selects should have the govuk styling so I'm sure we can simplify the CSS. In this sprint I'm not sure I have the time to do all the above + testing. So I'll modify this PR, to fix the bug we currently have, I'll create a ticket to tackle your comment.

lucascumsille:
@dracos I ended up moving `padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);` to `.govuk-select` to make it more generic. For testing I created a `input.govuk-select` nested juts inside a container to see if it would cause any issue:

Screenshot 2024-08-15 at 12 02 11

This is the report page
Screenshot 2024-08-15 at 12 02 48

I added one commit `c638247` to make the height of all elements nested inside `.dashboard-filters` similar.  The overrides in the` line-height` and `height that we added a few months ago for the `.dashboard-filters .multi-select-button` elements ` was causing an issue with the new generic `padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);`, so I deleted them. Another solution would have been just to give the `govuk-select` to "Report state", but we still haven't implemented a styling for the `input[date]`, so 'Start date' and 'End date' would have still needed some extra `height` so not much of a solution at the end. I'll add the styling for the `input[date]` one to the milestone above.
 
**Preview c638247**

Screenshot 2024-08-15 at 11 42 18

**Preview without c638247**

Screenshot 2024-08-15 at 11 36 32

Let me know what you think =)

lucascumsille:
By the way the ` max-width: 10em;` prevents any second line jumping, so no matter how many selections are made, they "Look up" button doesn't jump to a second line.

web/cobrands/sass/_dashboard.scss
```diff
@@ -297,8 +297,12 @@
...
 .dashboard-filters {
     .multi-select-button {
-        height: 28px;
-        line-height: 100%;
+        max-width: 10em;
+    }
+
+    .form-control {
+        // To make them have the same height as the .multi-select-button
+        height: 3.3em;
```
dracos, line 305: This is fine - something else bit odd to think about later, as we now have select.form-control { height: 2.2em }, .filters .form-control { height: 2.2em }, .report-list-filters .form-control { height: 1.8em } and now with this .dashboard-filters .form-control { height: 3.3em }. And the filters on the heatmap page and the triage page do not look as nice as the public filters, I don't think they were updated to match the public around page.

dracos (approved):
Fixes the issue, thanks.

As an example of my comment, for later, here's the Bromley heatmap filters at present:
image
(I think the public filters changed the definition of what .report-list-filters was used for, so other users of that class have gone a little wrong.)

PR #5115, opened 2024-08-15
[Brent] Check for 3rd contamination event.

dracos:
For FD-3672
Given info was: "With regards to your contamination events you would need to look for any properties that have had a 3rd, 4th or 5th contamination in the last X months. Use the GetEvents call with EventIDs: 2923, 2924 and 2925."
Am only checking for 3rd, as presumably that means they've had 3, so no need to check 4th/5th.
[skip changelog]


MorayMySoc (approved):
Looks good to me

dracos:
Heard nothing, taking off staging

PR #5116, opened 2024-08-15
Added navigation links to footer

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4486

Adding navigation links to footer. With this PR we would be complying with [WCAG 2.4.5: Multiple Ways](https://www.w3.org/WAI/WCAG22/Understanding/multiple-ways), so users can access webpages using a different method that is not the navigation bar at the top.

### Desktop
Screenshot 2024-08-15 at 14 46 54

### Mobile
Screenshot 2024-08-15 at 14 48 54

[Skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -60,6 +60,11 @@ $geolocation-link: #222 !default;
 $geolocation-link-border: $geolocation-link !default;
 $geolocation-link-background-color: transparent !default;
 
+$footer-main-nav-font-size: 1rem !default;
+$footer-main-nav-padding: 2em 0 !default;
+$footer-main-nav-background: #2E3033 !default;
+$footer-main-nav-color: #fff !default;
+
 @import "_mixins";
 @import "_report_list";
 
@@ -1075,6 +1080,34 @@ footer {
         font-size: 0.75em;
         margin-bottom: 0;
       }
+    }
+  }
+}
+
+.footer-main-nav {
+  padding: $footer-main-nav-padding;
+  background-color: $footer-main-nav-background;
+  .nav-footer {
+    display: flex;
+    column-gap: 2rem;
+    row-gap: 1rem;
+    font-size: $footer-main-nav-font-size;
+    justify-content: center;
+    margin: 0;
+    flex-wrap: wrap;
+  
+    li {
+        list-style: none;
+        margin-bottom: 0;
+    }
+
+    a {
+        color: $footer-main-nav-color;
+  
+        &:focus {
+            background-color: #ffff00;
+            color: #000;
+        }
     }
   }
 }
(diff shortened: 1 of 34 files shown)


davea (changes requested):
I like the idea but think the SocietyWorks footer template is the wrong place to include this.

As a suggestion I've made a new branch `ACC-SCC01-footer-navigation-suggestion` with a commit that creates a new `_footer_main_nav.html` template which re-uses the `main_nav.html` template to achieve the same thing (well, almost the same thing, as I thought having the SW logo as the final element on the page looked better :)). This way we don't need to duplicate the code that outputs the nav items, and additionally allows cobrands to opt-in to the footer nav independently of the SW footer. I've added it to the Bexley template as an example (see Bexley's `footer_extra.html`) but would just be a case of adding an extra `INCLUDE` line wherever we `INCLUDE` the SW footer. I've not done anything with the SCSS, but guess if you agree with this approach would maybe also benefit from being extracted from the SW footer styles. What do you think?

lucascumsille:
@davea I like you idea. I merged your branch into mine and I added a commit to move the styling from the SW bit. I guess my only question would be about which approach to include this we will have. I thought by default we would include this footer for every cobrand that has the powered footer, so we can comply with the WCAG criteria. At the same time is not like this will have much impact, it's more to tick a box. You reckon we should include this by default or we include it only when a cobrand ask to include it?

web/cobrands/sass/_base.scss
```diff
@@ -1087,6 +1092,34 @@ footer {
...
+    font-size: $footer-main-nav-font-size;
+    justify-content: center;
+    margin: 0;
+    flex-wrap: wrap;
+  
+    li {
+        list-style: none;
+        margin-bottom: 0;
+    }
+
```
davea, line 1111: I think this bit needs adding back in:

```suggestion
    .frontpage & {
            // In frontpage users can't interact with 'Report a problem' link
            li:first-child {
                display: none;
            }
        }
```


image

davea (changes requested):
Sorry one small SCSS thing!

I think we should include this by default, as you say it's an accessibility improvement so can't hurt to include it... As for _how_, I think adding the `[% PROCESS '_footer_main_nav.html' %]` line to all the cobrands' `footer_extra.html` is the best solution. Makes it a one-line change to remove if the client requests it.

web/cobrands/sass/_base.scss
lucascumsille: This bit shouldn't be needed anymore because of this: https://github.com/mysociety/fixmystreet/pull/5122

lucascumsille:
@davea I have updated the file for all the cobrand that use `footer_extra.html` I also rebased to master so the PR that I was talking is reflected here.

davea (approved):

PR #5118, opened 2024-08-15
[Surrey] Allow some categories to be reported off road

MorayMySoc:
Request for two categories to be reportable not on highway 

https://3.basecamp.com/4020879/buckets/36546415/todos/7696151989


MorayMySoc:
Will need to update yml on server to take effect

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -1404,6 +1404,25 @@ fixmystreet.assets.shropshire.streetlight_asset_message = function(asset) {
     return out;
 };
 
+/* Surrey */
+
+fixmystreet.assets.surrey = {};
+
+fixmystreet.assets.surrey.road_not_found = function(asset) {
+    var permissive_categories = ['Flooding inside a building', 'Flooding not on the highway but in a park or private land'];
```
dracos, line 1412: Just thinking, could we include this as data on the layer?  (The argument to this function is the layer, not an asset). eg if we store it in `permissive_categories` in the data, it should be lookable here as layer.fixmystreet.permissive_categories.

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -1404,6 +1404,25 @@ fixmystreet.assets.shropshire.streetlight_asset_message = function(asset) {
     return out;
 };
 
+/* Surrey */
+
+fixmystreet.assets.surrey = {};
+
+fixmystreet.assets.surrey.road_not_found = function(asset) {
```
dracos, line 1411: function(layer)

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -1404,6 +1404,25 @@ fixmystreet.assets.shropshire.streetlight_asset_message = function(asset) {
...
+    var permissive_categories = ['Flooding inside a building', 'Flooding not on the highway but in a park or private land'];
+    var currentCategory = fixmystreet.reporting.selectedCategory().category;
+        if (!fixmystreet.reporting_data || currentCategory === '') {
```
dracos, line 1414: funny indent here

dracos (changes requested):
Looks good, and especially thanks for doing a Cypress test - one suggestion as to how to have the categories in the config rather than hard-coded

dracos (approved):
Thanks, looks good :)

PR #5120, opened 2024-08-19
Added tabindex to .js-recentre-map element

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4488

Currently `.js-recentre-map` is the only control element in the map that can't be focused. By adding a `tabindex="0"` we allow keyboard users to be able to use that element.

Preview here:

https://github.com/user-attachments/assets/4efb1a2b-d266-4915-bf83-fcc6c77bf82a

[Skip changelog]

### DIFF
--- a/templates/web/base/maps/openlayers.html
+++ b/templates/web/base/maps/openlayers.html
@@ -73,7 +73,7 @@
     [% ELSE %]
       <a class="map-pins-toggle" rel="nofollow" href="[% c.uri_with( { no_pins => 1 } ) %]"><span class="map-link-label">[% loc('Hide pins') %]</span></a>
     [% END %]
-    <a class="js-recentre-map hidden-nojs">
+    <a class="js-recentre-map hidden-nojs" tabindex="0">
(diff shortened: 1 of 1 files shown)


davea (changes requested):
For me on macOS Safari the "Aerial map" and "Hide pins" buttons aren't focusable:

tabindex


Strangely, adding a `tabindex` to the `a` elements for those buttons doesn't make them focusable...

lucascumsille:
@davea can you double check if the element has the tabindex, I just tested on safari an it's working for me.

https://github.com/user-attachments/assets/eab68000-230b-489c-af5f-c9d10eab3bd2

lizettal:
I think this got resolved on our call earlier but @lucascumsille @davea can you please update accordingly

lucascumsille:
I think it should be ready, but @davea let me know if you think I need to change something

davea:
Yep, resolved. This was because I didn't have the relevant "Press Tab to highlight each item on a web page" option turned on in Safari.

davea (approved):

PR #5128, opened 2024-08-22
[Surrey] Display 'password required'-type msg if empty pword

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4484.

[skip changelog]


dracos (approved):
This works okay, I'm a tiny bit worried in case the 'required' code kicks in but somehow the removal does not and then it would be impossible to submit, but I assume that's pretty unlikely. Such a unlikely scenario in the first place though.

PR #5132, opened 2024-08-23
Include same-day data in pre-generated CSV exports

dracos:
If a request for a CSV export includes the current day, stop the pre-generated export at midnight and switch to a live export for the current day. [skip changelog]
Fixes https://github.com/mysociety/societyworks/issues/4170


davea (approved):
Lovely!

PR #5133, opened 2024-08-23
[Camden] Adds boundary check against neighbours

MorayMySoc:
Camden have provided a responsibility layer around boundary areas - we check against this to see who is responsible for a location that may be outside their defined boundary area and set the body responsible accordingly. Alternatively, it may be inside their boundary area but not be their responsibility.

They do not divide up responsibilities by category type

Camden needs to be set to be covered by, and to cover, all of its boundary areas.

https://github.com/mysociety/societyworks/issues/4399

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Camden.pm
+++ b/perllib/FixMyStreet/Cobrand/Camden.pm
@@ -15,7 +15,17 @@ use parent 'FixMyStreet::Cobrand::Whitelabel';
 use strict;
 use warnings;
 
-sub council_area_id { return [2505, 2488]; }
+sub council_area_id {
+    return [
+        2505, # Camden
+        2488, # Brent
+        2489, # Barnet
+        2504, # Westminster
+        2507, # Islington
+        2509, # Haringey
+        2512, # City of London
+    ];
+}
 sub council_area { return 'Camden'; }
 sub council_name { return 'Camden Council'; }
 sub council_url { return 'camden'; }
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/Camden.pm
```diff
@@ -15,7 +15,14 @@ use parent 'FixMyStreet::Cobrand::Whitelabel';
...
 
-sub council_area_id { return [2505, 2488]; }
+use constant BRENT_MAPIT_ID => 2488;
+use constant BARNET_MAPIT_ID => 2489;
+use constant WESTMINSTER_MAPIT_ID => 2504;
+use constant ISLINGTON_MAPIT_ID => 2507;
+use constant HARINGEY_MAPIT_ID => 2509;
+use constant CITY_MAPIT_ID => 2512;
+
+sub council_area_id { return [2505, BRENT_MAPIT_ID, BARNET_MAPIT_ID, WESTMINSTER_MAPIT_ID, ISLINGTON_MAPIT_ID, HARINGEY_MAPIT_ID, CITY_MAPIT_ID]; }
```
chrismytton, line 25: I wonder if the constants might be a bit of an overkill, given they're only used here?

One simpler alternative might be to break the ids out onto separate lines with comments:

```perl
sub council_area_id {
    return [
        2505,   # Camden
        2488,   # Brent
        2489,   # Barnet
        2504,   # Westminster
        2507,   # Islington
        2509,   # Haringey
        2512    # City of London
    ];
}
```

chrismytton (approved):
One very minor comment about a potentially simpler version of `council_area_id`, but otherwise looks good!

PR #5138, opened 2024-08-27
[Admin] Minor user section improvements

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4433 in first commit.
Rest was either me clicking around, or wondering why the user page was so slow to load in production.


chrismytton (approved):
This is great! ✨

PR #5140, opened 2024-08-27
[Admin] Move add body/category forms to own pages, limit history

dracos:
One very old 'bug', in that we can now view those pages (and then separately deal with whatever is creating 1000s of log entries), and then I was fed up of being confused by two forms on one page.


chrismytton (approved):
Much nicer having these forms on separate pages 👌

PR #5141, opened 2024-08-27
[Admin] problem/update display tweaks

dracos:
Fixes https://mysocietysupport.freshdesk.com/a/tickets/4060 as first bit.
Others are just things I noticed while being around there.


MorayMySoc (approved):
👍

PR #5144, opened 2024-08-29
Added GOVUK Styling to 'View area reports'

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3822

- In the first commit I have tidy up `govuk.scss` and `multiselect`. Working on the initial issue made me realise the classes names weren't ideal for reusability and also a lot of code should have been in the govuk file. This commit is not directly related with the initial issue, so if you prefer I can create another PR for it.
- The second commit adds the govuk styling to "View area reports section" in `/reports` page

Preview:
https://github.com/user-attachments/assets/77d8f216-c3f7-4803-a92a-2f6004ef781c

[skip changelog]


lucascumsille:
@dracos I have tidy up some bits and improve the naming for classes on this branch. As an idea  I'm wondering if it would be a good opportunity to also change the approach so far for [this milestone](https://github.com/mysociety/societyworks/milestone/99) where every ticket creates a PR and then merged to master, instead we could use this one as the main branch and I start creating all the other branches and consequently PRs to be merged here and then when we are ready we can merge it to master. I'm trying to avoid situations like we had in the past(multi-select dashboard or bodies) where we have conflicting styles. Let me know what you think

lucascumsille:
>The problem in the past was that there were bits a PR changed that we didn’t spot, I don’t see how that’s fixed by having one giant long-running PR. If anything, smaller PRs that change one thing, but do so everywhere they’re used, would seem preferable, but like I say, I might have misunderstood

Thanks for clarifying so we'll continue as we were doing

dracos:
Also, just read https://www.tpgi.com/should-form-labels-be-wrapped-or-separate/ so perhaps would be easier to swap to the gov.uk-standard way, if we should be adding the for/id anyway.

lucascumsille:
Interesting! thanks for sharing, then I'll swap to that then so we match the GOVUK markup/styling. Yeah it will be probably easier.

lucascumsille:
Replaced by this PR: https://github.com/mysociety/fixmystreet/pull/5185

PR #5145, opened 2024-08-29
Consolidate primary button CSS classes.

dracos:
Gets rid of `green-btn` and `btn-primary` so all primary buttons across the board are `btn btn--primary`.

Ran `node check.js --base http://{s}.localhost:3000 --path /about/buttons --width 375 --height 600 --out before` (and after) with a small manual `templates/web/base/about/buttons.html` to show the buttons  - may be edge cases but hopefully no more than now/before: [buttons-before.pdf](https://github.com/user-attachments/files/16795587/buttons-before.pdf) [buttons-after.pdf](https://github.com/user-attachments/files/16795588/buttons-after.pdf)

Northumberland might be the only one to think about - its old green-btn and btn--primary were different, i kept the green-btn one but perhaps it shouldn't be that and they should all be the same?

For info, from my notes, live usage before this PR:
*    'btn btn-primary' (triage save, save draft, inspector save changes, bulky check/amend/cancel/retry/book now buttons, Pboro report missed/request new, Report another problem, Bexley garden waste subscribe button)
*    'btn-primary' alone (Subscribe on alert page, 2FA/text code pages, "Hide my name" button, Bucks view parish reports, Cycling Join Today and Cycling report another problem)
*    'green-btn' (Subscribe on problem updates, create account/forgot password, Sign in button, contact Send, moderate Save, Add support, Zurich report submit)
*    'btn btn--primary' (bulky admin items, 'Sign in with password', Creator fixed button, Report form submission)

### DIFF
--- a/templates/web/base/admin/triage/_inspect.html
+++ b/templates/web/base/admin/triage/_inspect.html
@@ -81,7 +81,7 @@
 
         <p>
           <input type="hidden" name="token" value="[% csrf_token %]" />
-          <input class="btn btn-primary" type="submit" value="[% loc('Save changes') %]" data-value-original="[% loc('Save changes') %]" data-value-duplicate="[% loc('Save + close as duplicate') %]" name="save" />
+          <input class="btn btn--primary" type="submit" value="[% loc('Save changes') %]" data-value-original="[% loc('Save changes') %]" data-value-duplicate="[% loc('Save + close as duplicate') %]" name="save" />
         </p>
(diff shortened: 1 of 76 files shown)


lucascumsille:
@dracos It looks good. There was only an issue with the `.btn` element for Kingston where the button wasn't accessible, but that was a problem before this PR.

lucascumsille (approved):

PR #5149, opened 2024-09-02
Remove svg-background-image, twothirdswithpage, fullwidthpage, #postcodeForm specific CSS

dracos:
[skip changelog] Lots of small changes with no noticeable effect but hopefully makes it all a bit simpler.

For the postcode changes, checked front page/alert/waste mobile/desktop before/after with the new check script, couldn't see any changes (or very occasional improvement!).


dracos:
Actually, this changes the desktop layout of the hint somehow that I didn't spot (and they're not different in my check screenshots, so perhaps I didn't run it on the right commit or something? :-/ ), so will look at that.

lucascumsille:
@dracos It looks good I tested using the screenshot tool for the faq and homepage on desktop and mobile. I couldn't find any remnants of the classes and mixins in the code.

lucascumsille (approved):

PR #5152, opened 2024-09-03
[Hart] Homepage updates

chrismytton:
Some changes to the homepage requested by Harts in FD-4500.


MorayMySoc (approved):
Looks good to me

PR #5154, opened 2024-09-03
Detect if a report was made via PWA.

neprune:
[skip changelog]

This took far longer than it needed to 😄.

Appends '(probably PWA)' to the service if we detect the display-mode is minimal-ui. 
Wasn't able to find a more robust way to determine if the frontend is a PWA or not.

Closes https://github.com/mysociety/societyworks/issues/4052


neprune:
Closed in favour of https://github.com/mysociety/fixmystreet/pull/5156

PR #5156, opened 2024-09-05
Detect if a report was made via PWA.

neprune:
[skip changelog]

closes https://github.com/mysociety/societyworks/issues/4052


davea (approved):
Looks good! One minor thing is that we lose the distinction between desktop and mobile PWA - does this matter?

neprune:
> Looks good! One minor thing is that we lose the distinction between desktop and mobile PWA - does this matter?

Ah good point, I don't see any harm in adding it to the service description. Have made it so.

davea (approved):

PR #5157, opened 2024-09-06
Fix dupe yesterday in CSV export

dracos:
And also speed up the export when active_user_planned_reports is used, by switching to a more normal way of doing that.
[skip changelog]


MorayMySoc (approved):
Looks good to me

PR #5159, opened 2024-09-10
[Waste] Remove use of loading wrapper.

dracos:
As the async code can kick in at any point property is loaded, you had to make sure any subsequent results page had the async loading wrapper div, otherwise you ended up with a blank page. I had this happen to me with the new waste calendar help page where I forgot to add it and so got a blank page if someone went straight to /waste/N/calendar. So I thought to prevent that happening again in future, we should have something that didn't rely on a div being added to the right page. Then I thought if I was doing that, did we need HTMX at all - this file which uses a few fns from HTMX and a little bit of usage is about 96% smaller. And then I tweaked it so the ajax call wouldn't return the header/footer (which HTMX was discarding) which makes the transferred data smaller as well. I know HTMX wasn't particularly large, and happy to have it if it's doing something bigger, but thought it made sense to reduce the usage further, especially as making the swap less to worry about [skip changelog]


davea (changes requested):
A couple of small things:

On Peterborough when it loads in it includes the page header again:

image

On Bromley the site message disappears:

bromley

dracos:
Thanks; missed that peterborough had its own waste/header.html that needed updating as well, and changed the header to still include the site message output rather than skip the whole template.

davea (approved):

dracos:
I added an extra RETURN IF ajax_loading to Bromley's footer.html (missed that, didn't matter unless 404 page returned), and the base/header.html (ditto).

PR #5160, opened 2024-09-10
[Bromley] Changes number of containers for requests

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/4522

[skip changelog]


dracos (approved):
(Just for note, e.g. 535 is residential, 536 is communal, doesn't look like they specified any difference, so what you have is fine)

PR #5161, opened 2024-09-10
[Hart] User own privacy url

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/4524

[skip changelog]


chrismytton (changes requested):
This works for the link on the help page, but it appears that the footer (`templates/web/hart/footer_extra.html`) has a hard-coded link to the old privacy policy URL.

Is it possible to make the footer use the `privacy_policy_url`?

chrismytton (approved):
Great, thanks! 🙏

PR #5162, opened 2024-09-11
Dupe suggestions can be disabled on a per-category basis in config

davea:
Setting a `suggestions` value to 0 within the `nearby_distances` cobrand feature disables the dupe suggestion feature for that group/category, eg:

```
COBRAND_FEATURES:
  nearby_distances:
    my_cobrand:
      suggestions:
        _fallback: 100
        Potholes: 0

```

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report.pm
+++ b/perllib/FixMyStreet/App/Controller/Report.pm
@@ -727,38 +727,42 @@ sub _nearby_json :Private {
         # This is for the list template, this is a list on that page.
         $c->stash->{page} = 'report';
 
-        my $dist = $self->_find_distance($c, $params);
-        $params->{distance} = $dist / 1000 unless $params->{distance}; # DB measures in km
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Report.pm
```diff
@@ -727,38 +727,42 @@ sub _nearby_json :Private {
         # This is for the list template, this is a list on that page.
         $c->stash->{page} = 'report';
 
-        my $dist = $self->_find_distance($c, $params);
-        $params->{distance} = $dist / 1000 unless $params->{distance}; # DB measures in km
+        if (my $dist = $self->_find_distance($c, $params)) { # distance of 0 means we can skip lookup entirely
```
dracos, line 730: Is fine, but could be done as an early return if not present, then don't need to indent everything further

dracos (approved):
Looks good, feel free to return early instead, or leave :)

PR #5163, opened 2024-09-12
Add --update-contacts flag to export-import-data

davea:
This allows you to update existing contacts when importing a JSON file.

Handy to e.g. batch update emails, or add parent categories, etc.

[skip changelog]


MorayMySoc (approved):
Looks good to me

PR #5164, opened 2024-09-12
Add reporter/staff/notopen311-open option for updates_allowed

chrismytton:
North Northants already had notopen311-open, but asked to also have the reporter/staff-open option, which didn't exist.

This change combines the notopen311-open and reporter/staff-open options into one.

For FD-4419


MorayMySoc (approved):
Looks good

MorayMySoc:
I'm going to deploy this now as it's being waited on...

PR #5165, opened 2024-09-12
[Bexley] Look up service description

MorayMySoc:
Adds the service description to the template email so the service and container are shown.

https://mysocietysupport.freshdesk.com/a/tickets/4449
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -1091,18 +1091,23 @@ sub waste_munge_report_data {
 
     my $c = $self->{c};
 
-    my $address = $c->stash->{property}->{address};
-    my $service = $c->stash->{services}{$id}{service_name};
+    my $property = $c->stash->{property};
+    my $address = $property->{address};
+    my $service_id = $c->stash->{services}{$id}{service_id};
+    my $service_name = $c->stash->{services}{$id}{service_name};
     my $uprn = $c->stash->{services}{$id}{uprn};
-    $data->{title} = "Report missed $service";
+    my $containers = $self->_containers($property);
+    my $service_description = $containers->{$service_id}->{description};
+    $service_description = 'Various' if $service_description =~ /<li>/;
+    $data->{title} = "$service_name ($service_description)";
     $data->{detail} = "$data->{title}\n\n$address";
     $c->set_param('uprn', $uprn);
     $c->set_param('service_id', $id);
     $c->set_param('location_of_containers', $data->{extra_detail}) if $data->{extra_detail};
-    $c->set_param('service_item_name', $c->stash->{services}{$id}{service_id});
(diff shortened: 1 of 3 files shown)


templates/email/bexley/waste/_sidebar_content.html
```diff
@@ -7,6 +8,7 @@
   ELSIF is_missed_collection;
     title = report.title;
     title = title.remove('Report missed ');
+    title = title _ ' (' _ containers.$title _ ')';
```
dracos, line 11: This feels a bit fragile, to rely upon a title text string lookup. I may be missing something here, but if "Report missed" is being removed from the title, why are we adding it in the first place in `Bexley/Waste.pm`? Could we not add it when we construct the title at report creation, also add the type of container to the title there, and then we don't need to add it here or remove any text. I don't think anything expects the title to be of a particular format.

MorayMySoc: I've not checked whether there is any other objection to not adding 'Report missed' to the service, but does seem to make sense...

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -1094,7 +1094,9 @@ sub waste_munge_report_data {
     my $address = $c->stash->{property}->{address};
     my $service = $c->stash->{services}{$id}{service_name};
     my $uprn = $c->stash->{services}{$id}{uprn};
-    $data->{title} = "Report missed $service";
+    my $service_description = _containers->{$c->stash->{services}{$id}->{service_id}}->{description};
```
dracos, line 1097: I think this needs to be $self->_containers and property needs to be passed in, otherwise it might have the wrong information for communal properties

dracos (changes requested):
One bug, I think, "Various" handling looks good

dracos:
@chrismytton As I made the fix here, could you give this a once over

chrismytton (approved):

templates/email/bexley/waste/_sidebar_content.html
```diff
@@ -1,16 +1,14 @@
 [% # Hacky string munging :-( %]
+[% containers = cobrand.container_desc_for_template %]
```
dracos, line 2: This line no longer needed

templates/email/bexley/waste/_sidebar_content.html
```diff
@@ -1,16 +1,14 @@
...
     title = report.detail;
     title = title.remove(address);
-  ELSIF is_missed_collection;
-    title = report.title;
-    title = title.remove('Report missed ');
   END
 %]
 <h2 style="[% h2_style %]">Bins reported as missed:</h2>
-[% title | html_para_email(secondary_p_style) %]
+[% report.title | html_para_email(secondary_p_style) %]
```
dracos, line 11: This line still needs to be title, otherwise the 'is_no_collections_enquiry" changes are lost.

PR #5168, opened 2024-09-16
Ensure cut_off_date is respected if is_two_tier is set

davea:
[skip changelog]


dracos (approved):
Looks good; guess you /could/ factor it out with the same code at line 68, but is fine as is.

PR #5169, opened 2024-09-17
Full name is aria-required in report creation

chrismytton:
Ensure the full name field is marked as required for screen readers.

Fixes https://github.com/mysociety/societyworks/issues/4492

### DIFF
--- a/templates/web/base/report/form/user_name.html
+++ b/templates/web/base/report/form/user_name.html
@@ -9,5 +9,5 @@
     <p class='form-error'>[% field_errors.name %]</p>
 [% END %]
 <input type="text" class="form-control [% valid_class OR 'validName' %] js-form-name [% extra_class %]"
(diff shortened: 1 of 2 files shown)


davea (approved):

dracos (changes requested):
I think the reason these aren't `required` is because that can cause issues with the multi-page nature of the form (ie. these being required but on a non-visible page if you're signing in rather than giving your name). The JS in fixmystreet.js already adding required and aria-required and removing, as far as I can see, under "set correct required status depending on what we submit", which @chrismytton you added, so I'd be worried this e.g. made the sign in form unsubmittable for screenreaders. I'd like more details from the person complaining, I think, before we add this, as to the circumstance when it didnt' have aria-required?
Or perhaps aria-required doesn't matter so much. But it's not required if they're signing in is my point, iyswim.

dracos (approved):

PR #5174, opened 2024-09-18
Fix admin editing of anonymous reports

davea:
Not submitting the name field would cause a DB error because it would try to set `name` to NULL.
And including the other two seemed like a good idea too.

[skip changelog]


dracos (approved):
Ah great, thanks, I had this on my list to do as it came up as a side effect in FD-4569.
I don't think we would ever want to make the anonymous user not anonymous, and username shouldn't matter if not present, but I agree, might as well pass them all back for now just to match what it was before.

PR #5175, opened 2024-09-18
Capture group from hoisted categories.

dracos:
If a category is in a group on its own, the reporting form hoists it up to the top level, to save a pointless extra step. In doing so, it was no longer getting the associated group to store for submission. Rewrite the input so it is clearer as to what we are receiving. For FD-4534 [skip changelog]


davea (approved):
🪝 🆙

PR #5177, opened 2024-09-19
[Open311] Add --exclude to bin/open311-populate-service-list

davea:
Allows bodies to be excluded from Open311 service list fetching.

This seemed like the simplest approach and matches what we do for `bin/fetch`. Other options I considered were adding a "skip service list fetch" checkbox to the body admin edit page, and adding a cobrand hook to allow cobrand to opt-out of service list fetching.


dracos (approved):
Is fine, though I guess to match fetch, we'd have bodies_exclude in the Populate package and do the exclusion in there, which would then make this testable as well. But is small thing.

PR #5179, opened 2024-09-19
[Echo] Mark update as skipped after fallback email

dracos:
Once we have sent an email concerning a payment confirmation update failure, we should stop trying to send that payment confirmation update. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
@@ -1300,9 +1300,11 @@ sub send_bulky_payment_echo_update_failed {
 
     while ( my $report = $rs->next ) {
         # Ignore if there is an update in Comment table with an external_id
-        if ( $report->comments->search( { external_id => { '!=', undef } } )
-            ->count > 0
-        ) {
+        if ( $report->comments->search({
+            external_id => { '!=', undef },
+            text => { like => 'Payment confirmed%' },
+            # Can be extra => { '\?' => 'fms_extra_payments' }, after that's added
+        })->count) {
             # Set flag so we don't repeatedly check the comments for this
             # report
             $report->set_extra_metadata( echo_update_sent => 1 );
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1340,6 +1342,12 @@ sub send_bulky_payment_echo_update_failed {
                 echo_update_failure_email_sent => 1 );
             $report->update;
 
+            $report->comments->search({
```
nephila-nacrea, line 1345: So:
We don't send a failure email for any report with a 'Payment confirmed' comment (update), if that comment has an external ID (the logic above)
then
here, we update comments that *don't* have external IDs.
But where are comments that don't have external IDs originally created?
Do all comments have no external ID by default, then if the external ID is not populated later, it implies failure?

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1300,9 +1300,11 @@ sub send_bulky_payment_echo_update_failed {
 
     while ( my $report = $rs->next ) {
         # Ignore if there is an update in Comment table with an external_id
-        if ( $report->comments->search( { external_id => { '!=', undef } } )
-            ->count > 0
-        ) {
+        if ( $report->comments->search({
+            external_id => { '!=', undef },
+            text => { like => 'Payment confirmed%' },
+            # Can be extra => { '\?' => 'fms_extra_payments' }, after that's added
```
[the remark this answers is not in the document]
dracos: Yeah, this PR: https://github.com/mysociety/fixmystreet/pull/5089

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
nephila-nacrea: And the place where the external_id is set on the comment in perllib/Open311/PostServiceRequestUpdates.pm->process_update, from what I can see (is that correct?)

git blame is telling me that I worked on this code a year ago; I don't know why this has all gone from my head 😞 

I wonder if you should have 'Payment confirmed' in a variable, since it's checked more than once, but it's not a major thing.

nephila-nacrea (approved):

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
dracos: Yes, exactly, in the normal case of sending an Open311 update to a backend system, that's where the external_id would then be set after the successful send. I don't remember things I've worked on weeks ago, that's why I try (don't always succeed) to make sure it makes sense and documented for when I come back :) I think you were adding the sending of the email itself back then, so didn't need to worry about the updates sending.

PR #5180, opened 2024-09-19
[Greenwich] Update link colour.

dracos:
Update to match their site. The front page colour didn't seem to do anything, and the visited seemed pointless. [skip changelog] Fixes https://github.com/mysociety/societyworks/issues/3435

### DIFF
--- a/web/cobrands/greenwich/base.scss
+++ b/web/cobrands/greenwich/base.scss
@@ -6,16 +6,6 @@
 @import "../sass/base";
 @import "../sass/report_list_pins";
 
-body.frontpage {
-    color: $primary_text;
-    a,
-    .fake-link {
-        &:visited {
-            color: $links_visited;
-        }
-    }
-}
-
 #site-logo {
     background-image: url("/cobrands/greenwich/images/greenwich-logo.svg");
     // background: url("/cobrands/greenwich/images/greenwich-logo.svg") 0 50% no-repeat;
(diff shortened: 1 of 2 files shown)


web/cobrands/greenwich/base.scss
```diff
@@ -137,7 +127,7 @@ form {
                         text-decoration: underline;
                     }
                     &:visited {
-                        color: $links_visited;
+                        color: darken($link-color, 10%);
```
lucascumsille, line 130: I added this bit, it wasn't compiling the CSS because there was a remanent of `links_visited`

lucascumsille (commented):
The other approach would be to remove the visited bit altogether.

web/cobrands/greenwich/base.scss
dracos: I don't think this ever happens at all, in fact - I can't see a fake-link within a #form_sign_in within a fieldset anywhere.

lucascumsille: To be fair is not even needed. I just deleted in another commit

lucascumsille:
@dracos I think this should be ready =)

web/cobrands/greenwich/base.scss
dracos: Sorry, I meant that entire block is never used, the whole thing. I've removed it.

PR #5182, opened 2024-09-20
[Surrey] Exclude Open311-fetched reports from cut_off_date

davea:
For https://github.com/mysociety/societyworks/issues/4512

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Surrey.pm
+++ b/perllib/FixMyStreet/Cobrand/Surrey.pm
@@ -66,6 +66,44 @@ sub enter_postcode_text { 'Enter a nearby UK postcode, or street name and area' 
 sub cut_off_date { '2024-09-16' }
 
 
+=head2 problems_restriction/problems_sql_restriction/problems_on_map_restriction
+
+Reports made on FMS.com before the cut off date are not shown on the Surrey cobrand;
+however if a report is fetched over Open311 it is shown regardless of the cut off date.
+
+=cut
+
+sub problems_restriction {
+    my ($self, $rs) = @_;
+    return $rs if FixMyStreet->staging_flag('skip_checks');
+
+    $rs = $rs->to_body($self->body);
+
+    my $date = $self->cut_off_date;
+    my $table = ref $rs eq 'FixMyStreet::DB::ResultSet::Nearby' ? 'problem' : 'me';
+    return $rs->search([
+        { "$table.created" => { '>=', $date } },
+        { "$table.service" => 'Open311' },
+    ]);
+}
+
+sub problems_sql_restriction {
+    my ($self, $item_table) = @_;
+    my $date = $self->cut_off_date;
+    return " AND ( created >= '$date' OR service = 'Open311' )";
+}
+
+sub problems_on_map_restriction {
+    my ($self, $rs) = @_;
+    my $date = $self->cut_off_date;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Surrey.pm
```diff
@@ -66,6 +66,50 @@ sub enter_postcode_text { 'Enter a nearby UK postcode, or street name and area'
...
+
+sub problems_restriction {
+    my ($self, $rs) = @_;
+    return $rs if FixMyStreet->staging_flag('skip_checks');
+
+    $rs = $rs->to_body($self->body);
+
+    my $date = $self->cut_off_date;
+    my $table = ref $rs eq 'FixMyStreet::DB::ResultSet::Nearby' ? 'problem' : 'me';
+    return $rs->search({
```
dracos, line 84: I think (but do check), this can be either `$rs->search({ -or => { "$table.created" => { '>=', $date }, "$table.service" => 'Open311' } });` or even `$rs->search([ "$table.created" => { '>=', $date }, "$table.service" => 'Open311' ]);`

perllib/FixMyStreet/Cobrand/Surrey.pm
```diff
@@ -66,6 +66,50 @@ sub enter_postcode_text { 'Enter a nearby UK postcode, or street name and area'
...
+    my ($self, $item_table) = @_;
+    my $date = $self->cut_off_date;
+    return " AND ( created >= '$date' OR service = 'Open311' )";
+}
+
+sub problems_on_map_restriction {
+    my ($self, $rs) = @_;
+    my $date = $self->cut_off_date;
+    my $table = ref $rs eq 'FixMyStreet::DB::ResultSet::Nearby' ? 'problem' : 'me';
+    return $rs->search({
```
dracos, line 103: Similar

dracos (changes requested):
Think query code can be simplified, same outcome, is all :)

davea:
Based on my understanding of `EXPLAIN ANALYZE` output I don't think this will undo the improvements that came from adding the index on `created`, but would appreciate your opinion!

dracos (approved):
If you look at the analyze when it's only `created`, you can see a 'nearby' query boils down to:
1. Get the nearby problems using bitmap heap/index scan on problem_radians_latitude_longitude_idx
2. Do a bitmap heap scan using bitmap index scans on both problem_bodies_str_array_idx and problem_created_idx, filtering out state/timestamp based (179 rows)
3. hash together steps 1 and 2, filtering out shown reports
4. sort/limit

With it also checking `service`, this becomes:
1. Same - nearby problems using bitmap heap/index scan
2. Can only use problem_bodies_str_array_idx now, so bitmap heap scan with that, filter out ones that don't match created/service/state/timestamp (46,317 rows)
3. Same
4. Same

Which is equivalent to what it does when there isn't a cut off date at all; still using a bitmap heap scan, so much better than when it was going wrong. Its runtime is comparable, but it is having to do quite a bit more 'work' (looking at the calculated cost, not the actual time). So yes, think it's fine, we might want to have a think of is there's something we can do to improve the general 'no cut off date' (or here) query, could another index on state, or perhaps a combined (state,created) or (created,state) index, or some other index for this common query help, for example?

PR #5185, opened 2024-09-24
Added GOVUK styling to 'View area reports' sections

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3822

### DIFF
--- a/templates/web/oxfordshire/reports/_other_districts-list.html
+++ b/templates/web/oxfordshire/reports/_other_districts-list.html
@@ -7,12 +7,12 @@
     <p class="[% show_parish_hidden_class %]"><a href="?show_parishes=1" class="js-ward-select-multiple">[% loc('Select multiple parishes?') %]</a></p>
     <form method="get" action="">
     <section class="full-width">
-    <ul class="item-list item-list--wards">
+    <ul class="item-list item-list--wards govuk-checkboxes govuk-checkboxes--small">
       [% FOR child IN parish_ward.values.sort('name') %]
       [% IF child.type == 'CPC' %]
-      <li class="item-list__item item-list--wards__item">
-        <label class="[% show_parish_shown_class %]">
-            <input type="checkbox" name="ward" value="[% child.name | html %]">
+      <li class="govuk-checkboxes__item">
+        <input class="[% show_wards_shown_class %] govuk-checkboxes__input" id="[% child.name | lower | replace(' ', '-') %]" type="checkbox" name="ward" value="[% child.name | html %]">
+        <label class="[% show_parish_shown_class %] js-ward-multi govuk-label govuk-checkboxes__label" for="[% child.name | lower | replace(' ', '-') %]">
           [% child.name %]
         </label>
         <a class="[% show_parish_hidden_class %]" href="[% child.url %]">[% child.name %]</a>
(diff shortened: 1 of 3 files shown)


web/cobrands/sass/_base.scss
```diff
@@ -1326,12 +1326,23 @@ input.final-submit {
...
 
-.item-list--wards__item {
+.item-list--wards {
   @extend %item-list__item--link;
-
+  padding: 1em;
   label {
-    margin: 0;
-    padding: 0.5em 1em;
+    margin-top: inherit;
```
dracos, line 1332: Don't think this is doing anything? It's overridden by the margin-top: -2px of the govuk style, unless I've missed something.

templates/web/oxfordshire/reports/_other_districts-list.html
```diff
@@ -67,12 +67,12 @@ <h2>View reports by district</h2>
   <p class="[% show_district_hidden_class %]"><a href="?show_districts=1" class="js-ward-select-multiple">Select multiple districts?</a></p>
   <form method="get" action="">
     <section class="full-width">
-      <ul class="item-list item-list--wards">
+      <ul class="item-list item-list--wards govuk-checkboxes govuk-checkboxes--small">
         [% FOR child IN parish_ward.values.sort('name') ~%]
         [% IF child.type == 'DIS' %]
-        <li class="item-list__item item-list--wards__item">
```
dracos, line 73: `item-list--wards__item` has been removed from the HTML, but is still present in a number of cobrand CSS (only ever copying what reports__item is doing, looks like), so they can all be removed/checked?

dracos (changes requested):
I was a bit confused because the parent `item-list--wards` is extending a child `%item-list__item--link` mixin, which I see works (apart from having to override the padding), but I wonder if it would be clearer if that mixin was changed to a parent one used by `item-list--wards` and `item-list--reports`. I've added a commit that also makes the hover take up the full width, as before, let me know if that looks okay.

lucascumsille:
> I was a bit confused because the parent `item-list--wards` is extending a child `%item-list__item--link` mixin, which I see works (apart from having to override the padding), but I wonder if it would be clearer if that mixin was changed to a parent one used by `item-list--wards` and `item-list--reports`. I've added a commit that also makes the hover take up the full width, as before, let me know if that looks okay.
Good point, I made some amendments for this bit and the other two comments. Let me know if this makes more sense.

web/cobrands/sass/_base.scss
```diff
@@ -1326,17 +1326,25 @@ input.final-submit {
...
+  a {
+    display: block;
+    margin-left: calc(-1em + -34px); // for the padding-left on govuk-checkboxes__item + padding of item-list--wards
+    margin-bottom: 0.5rem;
   }
 }
 
 .item-list--reports__item {
-  @extend %item-list__item--link;
+  @extend %item-list;
```
dracos: We've still got the issue here with this here is a child `<li>` `.item-list--reports__item` extending `%item-list`, and above we have the `.item-list--wards` `<ul>` extending `%item-list`, which feels wrong.

web/cobrands/sass/_base.scss
```diff
@@ -1311,7 +1311,7 @@ input.final-submit {
   margin-bottom: 0.5em;
 }
 
-%item-list__item--link {
+%item-list {
   padding: 0;
```
dracos: This padding does nothing when extended by wards, because `.item-list` already has padding:0; this is a list item needed thing.

dracos (commented):
There was still a child/parent issue - I've pushed a change to hopefully explain better than me trying to explain it, let me know if that makes sense

lucascumsille:
> There was still a child/parent issue - I've pushed a change to hopefully explain better than me trying to explain it, let me know if that makes sense (and should be done if so)

It does thank you =) @dracos  I have rebase everything into one commit.

PR #5187, opened 2024-09-25
[BANES] Prevent staff from assigning a report to a disabled category.

neprune:
[skip changelog] [FD-4530](https://mysocietysupport.freshdesk.com/a/tickets/4530)

### DIFF
--- a/templates/web/base/admin/report-category.html
+++ b/templates/web/base/admin/report-category.html
@@ -1,5 +1,8 @@
 [%~ BLOCK category_option ~%]
-<option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %]>[% cat.category_display | html %]</option>
+[% IF cat.is_disabled AND NOT cobrand.staff_can_assign_reports_to_disabled_categories %]
+    [% SET disable = 1 %]
+[% END %]
+<option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %][% ' disabled' IF disable %]>[% cat.category_display | html %][% ' (' _ loc('disabled')  _ ')' IF disable %]</option>
 [%~ END ~%]
 
 <select class="form-control" name="category" id="category">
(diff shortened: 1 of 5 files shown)


templates/web/base/admin/report-category.html
```diff
@@ -1,5 +1,8 @@
 [%~ BLOCK category_option ~%]
-<option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %]>[% cat.category_display | html %]</option>
+[% IF cat.disable_form_field AND NOT cat.body.get_cobrand_handler.staff_can_assign_reports_to_disabled_categories %]
```
dracos, line 2: Whilst this is the standard way of marking a category as disabled (and I think Bath probably does use it), any extra data can actually do it with its `disable_form` flag, so I think we'd need to have a new method to check if a category gets disabled, here.

I don't think we need to worry about handler, because staff will always be on their cobrand, so cobrand.staff_can_assign... would probably do?

dracos (changes requested):
Looking good, will need a little more looking at the disabled nature

templates/web/base/admin/report-category.html
```diff
@@ -1,5 +1,8 @@
 [%~ BLOCK category_option ~%]
-<option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %]>[% cat.category_display | html %]</option>
+[% IF cat.is_disabled AND NOT cobrand.staff_can_assign_reports_to_disabled_categories %]
+    [% SET disable = 1 %]
+[% END %]
+<option value="[% cat.category | html %]"[% ' selected' IF problem.category == cat.category %][% ' disabled' IF disable %]>[% cat.category_display | html %][% ' (disabled)' IF disable %]</option>
```
dracos, line 5: Could we stick a loc() around the string so it can be translated

dracos (changes requested):
Thanks, changes look good, just tiny couple of tweaks to go

dracos (approved):
Thanks :)

PR #5188, opened 2024-09-25
[Island Roads] Default 'CONFIRM Subject' to be a blank selection

nephila-nacrea:
A 'Please select' option with a blank value is added under every group header in the dropdown; this is selected by default (under the relevant header) to force the admin user to choose a triage category (rather than it defaulting to the first category in the group).

[skip changelog]

Screenshot 2024-09-25 at 14 54 00

### DIFF
--- a/templates/web/base/admin/triage/_inspect.html
+++ b/templates/web/base/admin/triage/_inspect.html
@@ -1,27 +1,61 @@
 [% BLOCK category_list %]
-<select class="form-control" name="[% field_name %]" id="[% field_name %]">
-  [% IF category_options.size %]
-        [%~ IF category_groups.size ~%]
-            [%~ FOREACH group IN category_groups ~%]
-                [% IF group.name %]<optgroup label="[% group.name %]">[% END %]
-                [% group_select = 0 %]
-                [% selected = 0 %]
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
This looks good, but almost all of the two halves of the IF are the same, so can we have a much smaller IF inside? It looks to me you only need to change the inmost option-related bit. IF IoW, print out blank; and the existing selected IF could be dropped entirely I think, as that only applies for IoW.
Also stick a required on the `<select>` (in general, that can be I think) :)

nephila-nacrea:
> This looks good, but almost all of the two halves of the IF are the same, so can we have a much smaller IF inside? It looks to me you only need to change the inmost option-related bit. IF IoW, print out blank; and the existing selected IF could be dropped entirely I think, as that only applies for IoW. Also stick a required on the `<select>` (in general, that can be I think) :)

@dracos Not sure I did it in quite the way you asked but it's compact and seems to work :-) Also added some tests.

dracos (approved):
Thanks, yep, looks good :)

MorayMySoc (approved):
Great - thank you

PR #5192, opened 2024-09-27
[WW][Bulky][Sutton] Bulky update images available logged out

MorayMySoc:
When viewing a link from an alert confirming a bulky waste order has been picked up, make the images available if the user is not logged in to their WasteWorks account.

https://github.com/mysociety/societyworks/issues/4525

[skip changelog]

### DIFF
--- a/t/cobrand/sutton.t
+++ b/t/cobrand/sutton.t
@@ -214,6 +214,7 @@ subtest 'updating of waste reports' => sub {
         user => $normal_user,
         category => 'Garden Subscription',
         cobrand_data => 'waste',
+        non_public => 1,
     });
     $reports[1]->update({ external_id => 'something-else' }); # To test loop
     $report = $reports[0];
@@ -321,6 +322,9 @@ EOF
         $report->update_extra_field({ name => 'Collection_Date', value => '2023-09-26T00:00:00Z' });
         $report->set_extra_metadata( item_1 => 'Armchair' );
         $report->set_extra_metadata( item_2 => 'BBQ' );
+        $report->push_extra_fields( # Add extra fields expected on a Bulky waste report
+            { name => 'Bulky_Collection_Bulky_Items', value => '3::85::83'}, { name => 'Bulky_Collection_Notes', value => 'One::Two::Three' }
+        );
         $report->update({ category => 'Bulky collection', external_id => 'waste-15005-' });
 
         $in = <<EOF;
@@ -378,14 +382,20 @@ EOF
         is $report->state, 'fixed - council', 'A state change';
         my $update = FixMyStreet::DB->resultset("Comment")->order_by('-id')->first;
         is $update->photo, '34c2a90ba9eb225b87ca1bac05fddd0e08ac865f.jpeg';
-
         FixMyStreet::Script::Alerts::send_updates();
-        my $body = $mech->get_text_body_from_email;
+        my $body = $mech->get_email->as_string;
         my $id = $report->id;
         like $body, qr/Reference: LBS-$id/;
         like $body, qr/Armchair/;
         like $body, qr/26 September/;
         like $body, qr/Your collection has now been completed/;
+        $mech->host('sutton.example.org');
+        (my $token) = $body =~ m#http://sutton.example.org(/R/.*?)"#;
+        $mech->get_ok($token);
+        (my $photo_link_thumbnail) = $mech->content =~ m#Photo of this report" src="(/photo.*?1)"#;
+        (my $photo_link_full) = $mech->content =~ m#a href="(/photo.*?1)"#;
+        $mech->get_ok($photo_link_thumbnail, "Successfully call thumbnail image");
+        $mech->get_ok($photo_link_full, "Successfully call full image");
     };
 };
 
--- a/templates/web/base/report/photo.html
+++ b/templates/web/base/report/photo.html
@@ -5,8 +5,13 @@
   [% FOR photo IN object.photos %]
     [% IF c.cobrand.allow_photo_display(object, photo.idx) %]
       <div class="update-img">
-          <a href="[% photo.url_full %]" rel="fancy">
-              <img alt="Photo of this report" src="[% photo.url %]">
+          [% IF alert_to_reporter %]
+          <a href="[% photo.url_temp_full %]" rel="fancy">
+              <img alt="Photo of this report" src="[% photo.url_temp %]">
+          [% ELSE %]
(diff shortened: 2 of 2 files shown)


templates/web/base/report/photo.html
```diff
@@ -5,8 +5,8 @@
   [% FOR photo IN object.photos %]
     [% IF c.cobrand.allow_photo_display(object, photo.idx) %]
       <div class="update-img">
-          <a href="[% photo.url_full %]" rel="fancy">
-              <img alt="Photo of this report" src="[% photo.url %]">
+          <a href="[% photo.url_temp_full %]" rel="fancy">
```
dracos, line 8: We do not want to be outputting hash URLs in the general case, that makes them e.g. much harder/impossible to redact once they're picked up by anything else, and this would apply to every photo on the entire site. As we discussed at https://mysociety.slack.com/archives/C01TK8P1K8T/p1727273888229009?thread_ts=1727273574.154349&cid=C01TK8P1K8T this should only be happening when the alert_to_reporter variable is set and we know it's someone who has come direct from an alert email.

t/cobrand/zurich.t
```diff
@@ -434,9 +434,8 @@ subtest 'Photo publishing' => sub {
     $mech->content_contains('photo/' . $report->id . '.1.fp.jpeg');
     $mech->content_lacks('photo/' . $report->id . '.2.fp.jpeg');
     $mech->get_ok( '/report/' . $report->id );
-    $mech->content_lacks('photo/' . $report->id . '.0.jpeg');
-    $mech->content_contains('photo/' . $report->id . '.1.jpeg');
-    $mech->content_lacks('photo/' . $report->id . '.2.jpeg');
+    my @photo_links = $mech->content =~ m#(Photo of this report" src="/photo/.*?")#g;
```
dracos, line 437: This change won't be necessary.

t/app/controller/report_updates.t
```diff
@@ -1559,9 +1559,7 @@ for my $test (
 
         my $update = $report->comments->first;
         ok $update, 'found update';
-
-        $mech->content_contains("/photo/c/" . $update->id . ".0.jpeg") if $test->{fields}->{photo1};
-
+        $mech->content_contains('Photo of this report" src="/photo/temp') if $test->{fields}->{photo1};
```
dracos, line 1562: Ditto

t/app/controller/report_inspect.t
```diff
@@ -623,7 +623,7 @@ FixMyStreet::override_config {
         my $comment = $report->comments->order_by('-id')->first;
         is $comment->photo, '74e3362283b6ef0c48686fb0e161da4043bbcc97.jpeg', 'photo added to comment';
         $mech->get_ok("/report/$report_id");
-        $mech->content_contains("/photo/c/" . $comment->id . ".0.jpeg");
+        $mech->content_contains("/photo/temp.74e3362283b6ef0c48686fb0e161da4043bbcc97.jpeg");
```
dracos, line 626: Ditto

t/cobrand/sutton.t
```diff
@@ -378,14 +379,26 @@ EOF
         is $report->state, 'fixed - council', 'A state change';
         my $update = FixMyStreet::DB->resultset("Comment")->order_by('-id')->first;
         is $update->photo, '34c2a90ba9eb225b87ca1bac05fddd0e08ac865f.jpeg';
-
+        $report->push_extra_fields( # Add extra fields expected on a Bulky waste report
```
dracos, line 382: I would move this line to 324 when the report is made a bulky collection one, so set up/update is all in the same place.

t/cobrand/sutton.t
```diff
@@ -378,14 +379,26 @@ EOF
...
+            { name => 'Bulky_Collection_Bulky_Items', value => '3::85::83'}, { name => 'Bulky_Collection_Notes', value => 'One::Two::Three' }
+        );
+        $report->update;
         FixMyStreet::Script::Alerts::send_updates();
-        my $body = $mech->get_text_body_from_email;
+        my $body = $mech->get_email->as_string;
```
dracos, line 387: What was this change for? Surely the link should be present in the text part as well?

t/cobrand/sutton.t
```diff
@@ -378,14 +379,26 @@ EOF
...
         my $id = $report->id;
         like $body, qr/Reference: LBS-$id/;
         like $body, qr/Armchair/;
         like $body, qr/26 September/;
         like $body, qr/Your collection has now been completed/;
+        $mech->host('sutton.example.org');
+        (my $token) = $body =~ m#http://sutton.example.org(/R/.*?)"#;
+        $mech->get($token);
```
dracos, line 395: get_ok?

dracos (changes requested):
Test change mostly looks fine, but the actual change needs to only be done when necessary.

t/cobrand/sutton.t
MorayMySoc: It doesn't - finding it hard to track down the template for this, so left this as it is for the moment.

t/cobrand/sutton.t
MorayMySoc: Added to fetching the photos too...not sure if that matters so much

t/cobrand/sutton.t
dracos: The alert template? Is templates/email/default/alert-udate.txt / html, both of which contain the `problem_url` variable which is set in Alerts.pm to an /R/ URL if the alert_user_id is the user_id.

t/cobrand/sutton.t
```diff
@@ -378,14 +382,23 @@ EOF
         is $report->state, 'fixed - council', 'A state change';
         my $update = FixMyStreet::DB->resultset("Comment")->order_by('-id')->first;
         is $update->photo, '34c2a90ba9eb225b87ca1bac05fddd0e08ac865f.jpeg';
-
+        $report->update;
```
dracos, line 385: So you don't need this update now, do you.

t/cobrand/sutton.t
dracos: The reason your regex wasn't finding the URL was because the text version of the email doesn't have a `"` at the end of it. Replacing that with e.g. `sutton.example.org(/R/[^"]*)` would match in the text version presumably.

dracos: Yes, I see now, sutton has its own alert-update templates, and the text one does not include the link, but the HTML one still does by making the image of the update clickable.

dracos (approved):
Thank you :)

PR #5193, opened 2024-09-27
[Waste] Calendar performance improvements

davea:
- Adds `Cache-Control` header to `calendar.ics` responses
 - No longer hits backend APIs for calendar help page

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -568,9 +568,14 @@ sub csc_payment_failed : Path('csc_payment_failed') : Args(0) {
     $c->detach;
 }
 
-sub property : Chained('/') : PathPart('waste') : CaptureArgs(1) {
+sub property_id : Chained('/') : PathPart('waste') : CaptureArgs(1) {
     my ($self, $c, $id) = @_;
+    $c->stash->{property_id} = $id;
+}
 
+sub property : Chained('property_id') : PathPart('') : CaptureArgs(0) {
+    my ($self, $c) = @_;
+    my $id = $c->stash->{property_id};
 
     # Some actions chained off /waste/property require user to be logged in.
     # The redirect to /auth does not work if it follows the asynchronous
@@ -670,9 +675,8 @@ sub bin_day_deny : Private {
     $c->detach('/page_error_403_access_denied', [ $msg ]);
 }
 
-sub calendar : Chained('property') : Args(0) {
+sub calendar : Chained('property_id') : Args(0) {
     my ($self, $c) = @_;
-
     $c->forward('/about/page', ['waste-calendar']);
 }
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -588,6 +588,13 @@ sub property : Chained('/') : PathPart('waste') : CaptureArgs(1) {
         $c->detach;
     }
 
+    # The calendar help page doesn't need anything but the property ID,
```
dracos, line 591: I think a clearer way to do this would be to stop calendar being Chained to property entirely - if it was its own route then you wouldn't need to complicate this with a special case. ie `sub calendar : Path : Args(1) {` or similar, then sticking the ID on the stash.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -679,6 +686,7 @@ sub calendar : Chained('property') : Args(0) {
 sub calendar_ics : Chained('property') : PathPart('calendar.ics') : Args(0) {
     my ($self, $c) = @_;
     $c->res->header(Content_Type => 'text/calendar');
+    $c->res->header(Cache_Control => 'max-age=86400');
```
dracos, line 689: If we wanted easier control over the timing, this could be in the server's httpd.conf, but is also fine here.

dracos (changes requested):
Good spot - think it'd be clearer to stop the help page being chained entirely.

perllib/FixMyStreet/App/Controller/Waste.pm
davea: Thought it made sense here because we also set the TTL within the iCal file a few lines below so keeps it all together.

perllib/FixMyStreet/App/Controller/Waste.pm
davea: I did try this to begin with but couldn't figure out the right incantation - will have another go based off your suggestion 👍

davea: Have moved to `LocalRegex` in a22a4007bb48d38018883786108d9ce7991c77d2, though that seems to be deprecated and the docs suggest using chains instead. But I can't see a way to do that without chaining off `property`...

dracos: Ugh, sorry, I must have had it in my head it was calendar/ID... So this way is fine (we have a number of other LocalRegexes), or your original way - I guess the actual cleanest would be to change `property` to Chain off a new tiny function (property_id?) that fetched the ID and didn't do anything else, and then calendar could Chain off of property_id, leaving rest of property alone.

dracos: Have pushed a change to use Chain for the ID alone, let me know if that works for you

davea: Looks great, thanks. Happy for me to merge?

dracos: Yep!

dracos: Actually, need to deploy, so doing it now.

PR #5194, opened 2024-09-30
Remove svg icon from accessibility tree in Geolocation button

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4558

Added role="img" and aria-hidden="true" attributes to geolocate button. I used this approach instead of the role="presentation", the later removes the semantic of the element, but the content is still available.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role
https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs

[skip changelog]


templates/web/base/around/geolocate_link_icon.html
```diff
@@ -1,2 +1,2 @@
...
-<svg class="geolocate-link__loading" width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><g fill="currentColor" fill-rule="nonzero"><path d="M16 8a8 8 0 11-8-8v2a6 6 0 106 6z"></path><circle cx="8" cy="1" r="1"></circle><circle cx="15" cy="8" r="1"></circle></g></svg>
+<svg class="geolocate-link__icon" width="16" height="16" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" role="img" aria-hidden="true"><path d="M23.744.255a.874.874 0 01.18.974l-9.89 22.252a.874.874 0 01-1.603-.012l-3.568-8.332-8.334-3.57a.874.874 0 01-.01-1.602L22.77.075a.874.874 0 01.972.18h.002z" fill="currentColor" fill-rule="nonzero"/></svg>
+<svg class="geolocate-link__loading" width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" role="img" aria-hidden="true"><g fill="currentColor" fill-rule="nonzero"><path d="M16 8a8 8 0 11-8-8v2a6 6 0 106 6z"></path><circle cx="8" cy="1" r="1"></circle><circle cx="15" cy="8" r="1"></circle></g></svg>
```
dracos, line 2: I agree on the little icon, but the loading, we're going to need something other than hiding it, aren't we? https://stackoverflow.com/questions/38704467/how-to-label-a-loading-animation-for-wai-aria seems to provide a few different options, but I think geolocation.js will need some changes to improve this.

dracos (changes requested):

templates/web/base/around/geolocate_link_icon.html
lucascumsille: mmm good point, however the loading state is so fast that most of the time the screen reader won't have the time to update the user, however while working on this I noticed related issue that we currently have, which is any error message, for example if the user declined the permission popup; the message "You declined..." doesn't get read. To solve this I added `aria-live="polite"` so whenever the message gets updated now it gets announced(I tried it with VoiceOver).

For the scenario when the user gets ask location permissions I noted that VoiceOver jumps straight away to the dialogue pop up, so this is already providing some context about what is happening the user. But I figured it was still worth it to include a message in the button in case some assistive devices behave differently.

I added a message when the location permission is granted, again this seems more a formality or for cases where the loading might be slow, but most of the time the screen reader won't have the time to read the message and we'll jump straight to the map page.

dracos (changes requested):
The "Allow FMS to access your location" replaces the little arrow icon and we lose the spinner as well, which isn't ideal.
I think the aria-live="polite" needs to sit on the `a` because otherwise it's overwritten by anything that changes innerHTML - but you said it worked, so perhaps I've got confused.
Anyway, I'd suggest moving the aria-live to the `<a>`, and not bothering with the new strings, we haven't had any complaints after all and as you say should be obvious from what's happening.

lucascumsille:
Thanks for the feedback, I added a fixup that get rid of the new strings and also move the `aria-live` to the `a` element, I also restore the `geolocate.js`. Regarding how we handle the loading icon, this is what happens with this PR atm:

- User hasn't provide permissions yet, so the dialogue popup will appear giving the user context about what is happening.
- User has already giving permission, so the user will be directed to the map page in a few seconds, most of the time not giving time to the assistive device to announce the success. However this won't cover cases where for X reason the loading time is taking longer, but this is probably very unlikely to happen.
- There is an error with the geolocation and gets announced by the browser.

dracos (approved):
Thanks :)

PR #5196, opened 2024-09-30
[Camden] Practical boundary update

dracos:
For FD-4608 - the CSV ward uses the same function as being fixed here for the all reports area being displayed. 
[skip changelog] (Updates the jquery validation library as well; edit, sorry and another couple of unrelated tiny fixes)


MorayMySoc (approved):
Ok, looks good to me

PR #5201, opened 2024-10-02
[Zurich][FMS] Adds a filter_description url argument

MorayMySoc:
For Zurich, can add a prefill_description to the url arguments which will fill the description of a report.

Can also add a prefill_category, and it will remove the category choice, so category can not be changed - also provides a call hook used by Zurich to allow reports to be made in not active categories.

https://github.com/mysociety/societyworks/issues/4517

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Report/New.pm
+++ b/perllib/FixMyStreet/App/Controller/Report/New.pm
@@ -1228,6 +1228,9 @@ sub process_report : Private {
 
     if ( $report->category ) {
         my @contacts = grep { $_->category eq $report->category } @{$c->stash->{contacts}};
+        if (!@contacts) {
+            @contacts = @{ $c->cobrand->call_hook('report_on_private_contacts' => $report->category) || [] };
+        }
         unless ( @contacts ) {
             $c->stash->{field_errors}->{category} = _('Please choose a category');
             $report->bodies_str( -1 );
@@ -1861,7 +1864,10 @@ sub check_for_category : Private {
     # Just check to see if the filter had an option
     $category ||= $c->get_param('filter_category') || '';
     $c->stash->{category} = $category;
-
+    if ($c->cobrand->moniker eq 'zurich') {
+        $c->stash->{prefill_category} = $c->get_param('prefill_category') if $c->get_param('prefill_category');
+        $c->stash->{report}->detail($c->get_param('prefill_description')) if $c->get_param('prefill_description');
+    }
     # Bit of a copy of set_report_extras, because we need the results here, but
     # don't want to run all of that fn until later as it e.g. alters field
     # errors at that point. Also, the report might already have some answers in
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -1438,4 +1438,13 @@ sub report_on_private_contacts {
     return [ FixMyStreet::DB->resultset('Contact')->not_deleted->search({category => $category})->all ];
 }
 
+sub munge_around_filter_category_list {
+    my $self = shift;
+
+    my $c = $self->{c};
+
+    $c->stash->{prefill_category} = $c->get_param('prefill_category') if $c->get_param('prefill_category');
+    $c->stash->{prefill_description} = $c->get_param('prefill_description') if $c->get_param('prefill_description');
+}
+
 1;
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -691,7 +691,7 @@ sub admin_report_edit {
         }
 
         if (
-            ($state eq 'confirmed') 
+            ($state eq 'confirmed')
             && $new_cat
             && $new_cat ne $problem->category
         ) {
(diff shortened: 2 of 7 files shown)


MorayMySoc:
There are a couple of issues with this in that removing the category list makes the photo box expand as well as move up. I can look into this further or ask a designer. Also when the Rss gets called in a later test there are warnings, which I will need to work out. But thought it best to check this is broadly correct before getting lost on the minutiae.

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -1432,4 +1432,10 @@ sub report_new_munge_before_insert {
     }
 }
 
+sub report_on_private_contacts {
+    my ($self, $category) = @_;
+
+    return [ FixMyStreet::DB->resultset('Contact')->search({category => $category})->all ];
```
dracos, line 1438: I guess this would also return deleted, which I guess might not be wanted, happy to keep the inactive in here, or exclude deleted (there's a `not_deleted` you could call before the search)

dracos (changes requested):
Thanks, changes all look good, just one query left because it was state=inactive before and now is everything

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -1435,7 +1435,7 @@ sub report_new_munge_before_insert {
 sub report_on_private_contacts {
     my ($self, $category) = @_;
 
-    return [ FixMyStreet::DB->resultset('Contact')->search({category => $category})->all ];
+    return [ FixMyStreet::DB->resultset('Contact')->search({state => 'inactive', category => $category})->all ];
```
dracos, line 1438: This works okay, but as I suggested I think `FixMyStreet::DB->resultset('Contact')->not_deleted->search({ category => $category })->all` is using the shared filter then, that other code like this uses? Unless there was a reason not to use that you didn't say.

dracos (changes requested):

perllib/FixMyStreet/Cobrand/Zurich.pm
MorayMySoc: Sort of misunderstood what you said in the previous comment was all.

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -1858,7 +1861,10 @@ sub check_for_category : Private {
     # Just check to see if the filter had an option
     $category ||= $c->get_param('filter_category') || '';
     $c->stash->{category} = $category;
-
+    if ((my $description = $c->get_param('filter_description')) && $c->cobrand->moniker eq 'zurich' && $category) {
```
dracos, line 1864: Sorry, I only just thought of this when coming to this as a whole, but I think this should probably be called 'prefill_description', not 'filter_description' - the category one is called filter_category because it is the variable to filter by category on the `/around` page; this one isn't filtering at all :)
And that led me on to the other annoying thought - should we actually add a new prefill_category option rather than rely on category/filter_category? I say that because currently you have to supply a description for this to kick in, but going back to their quote they don't say that they'll always be supplying one, they might want to only supply a category prefill - and then we can't really use category/filter_category because then if there's normal use of those the category options would disappear.
So, hopefully nothing big, but if we change this if to `if ($c->cobrand->moniker eq 'zurich' && ((my $description = $c->get_param('prefill_description')) || (my $prefill_category = $c->get_param('prefill_category')))) {` and change next couple of lines/test accordingly to cope, I think that's all, no need to change template.

dracos (changes requested):
Woo. Sorry, when looking at this in overall at the end, I then thought that we probably want to change the variable names, hope that makes sense, happy to chat

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: Quite right!

dracos (approved):
Thanks for that, think it's nice and clear now :)

dracos (approved):
Yep, think that's it :-)
You don't need the `| html` in the templates, that's automatic (but doesn't do anything if left in).

templates/web/base/around/postcode_form.html
```diff
@@ -41,11 +41,19 @@
             [% ELSIF c.user_exists AND c.user.categories.size %]
               <input type="hidden" name="filter_category" value="[% c.user.categories_string | html %]">
             [% END %]
+            [% IF c.cobrand.moniker == 'zurich' AND ( c.get_param('prefill_category') OR c.get_param('prefill_description') ) %]
+                <input type="hidden" name="prefill_category" value="[% c.get_param('prefill_category') | html %]">
```
dracos, line 45: Doesn't need the html here, just like you removed below :)

t/cobrand/zurich.t
```diff
@@ -345,6 +346,27 @@ subtest "Auto select private category and input description from url" => sub {
...
 
+subtest "Auto select category and input description from front page" => sub {
+    my $module = Test::MockModule->new('FixMyStreet::Geocode::Zurich');
+    $module->mock(string => sub { { latitude => 47.376056, longitude => 8.525481 } });
+
+    $mech->get_ok('/?prefill_category=Allgemein&prefill_description=Test 5678');
+    $mech->submit_form_ok({ with_fields => {
+        pc  => 'Langstrasse',
+    }});
+    is $mech->uri->path, '/around', 'Redirected to around page';
```
dracos, line 357: It is, but could we just check the prefill desc is within the source of this page, and the category list isn't being shown, just so it isn't lost in future by some accident.

dracos (changes requested):
Looks good, just a couple of tiny things :)

dracos (approved):
Thanks :)

PR #5203, opened 2024-10-03
[Bexley][WW] Only skip parent addresses if they are marked as 'shell' properties

nephila-nacrea:
We were skipping over all parent addresses on the assumption that all were 'shell' addresses that shouldn't be displayed. But the 'shell' status only applies to some parents.

For https://mysocietysupport.freshdesk.com/a/tickets/4432.
Closes https://github.com/mysociety/societyworks/issues/4498.

[skip changelog]

### DIFF
--- a/bin/bexley/make-bexley-ww-postcode-db
+++ b/bin/bexley/make-bexley-ww-postcode-db
@@ -205,11 +205,17 @@ while( my $row = $csv->getline($fh) ) {
         @row_h{@columns_blpu} = @$row;
 
         $uprn_extra_data{ $row_h{UPRN} }{has_parent} = !!$row_h{PARENT_UPRN};
+        $uprn_extra_data{ $row_h{UPRN} }{blpu_class} = $row_h{BLPU_CLASS};
     } elsif ( $record_id == $id_lpi ) {
         @row_h{@columns_lpi} = @$row;
 
         next if ( $row_h{LOGICAL_STATUS} || 0 ) != 1;
         next unless $row_h{POSTCODE};
+
+        # 'P' or 'PP' mean the property is a 'parent shell' and so should not be
+        # offered as a selectable address.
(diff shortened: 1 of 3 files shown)


bin/bexley/make-bexley-ww-postcode-db
```diff
@@ -246,20 +257,10 @@ while( my $row = $csv->getline($fh) ) {
             $row_h{SAO_END_NUMBER},
             $row_h{SAO_END_SUFFIX},
             $row_h{SAO_TEXT},
+
+            $blpu_class,
         );
     }
 }
 
-# It is possible for UPRNs in child_uprns to not occur in the postcodes table
```
nephila-nacrea, line 253: Turns out that the P/PP check fixes this issue as well, so have deleted this portion.

bin/bexley/make-bexley-ww-postcode-db
```diff
@@ -214,22 +204,24 @@ while( my $row = $csv->getline($fh) ) {
...
     } elsif ( $record_id == $id_lpi ) {
         @row_h{@columns_lpi} = @$row;
 
         next if ( $row_h{LOGICAL_STATUS} || 0 ) != 1;
         next unless $row_h{POSTCODE};
 
+        # 'P' or 'PP' mean the property is a 'parent shell' and so should not be
+        # offered as a selectable address.
+        my $blpu_class = $uprn_extra_data{ $row_h{UPRN} }{blpu_class} // '';
+        next if $blpu_class && ( $blpu_class eq 'P' || $blpu_class eq 'PP' );
```
dracos, line 218: Doesn't need the first if I think, given you've made sure it's not undefined in the line before.

dracos (changes requested):
One tiny change, good to get rid of the unneeded table; could you split it into two commits, one that changes to use the parent flag and removes the child_uprns table, and then one that ignores P/PP? :)

dracos (approved):
📮

PR #5206, opened 2024-10-04
[offline] On desktop /my/planned don’t cache reports automatically

davea:
Desktop users are presumably less likely to find themselves offline, so it doesn't make sense to cache all reports in the shortlist automatically. This commit adds a link to the bottom of the report list on the desktop version of the /my/planned page that allows users to kick off the caching process manually.

Behaviour on mobile is unchanged.

image

### DIFF
--- a/templates/web/base/my/planned.html
+++ b/templates/web/base/my/planned.html
@@ -29,6 +29,12 @@
 <div id="js-reports-list">
 [% INCLUDE 'my/_problem-list.html' shortlist = 1 %]
 </div>
+<div class="shadow-wrap hidden-nojs">
+    <ul id="key-tools">
+        <li><a class="offline js-cache-reports" href="#">[% loc("Save to this device for offline use") %]</a></li>
+    </ul>
+</div>
+
 </section>
 
         </div>
--- a/web/cobrands/fixmystreet/offline.js
+++ b/web/cobrands/fixmystreet/offline.js
@@ -419,9 +419,19 @@ if ($('#offline_list').length) {
 } else {
     fixmystreet.offlineBanner.make(false);
 
-    // On /my/planned, when online, cache all shortlisted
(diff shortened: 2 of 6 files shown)


web/cobrands/fixmystreet/offline.js
```diff
@@ -419,9 +419,18 @@ if ($('#offline_list').length) {
...
+    // On /my/planned, when online, cache all shortlisted automatically on mobile.
+    // On desktop, show a button to cache them.
     if (location.pathname === '/my/planned') {
-        fixmystreet.offline.updateCachedReports();
+        if ($("html").hasClass("mobile")) {
+            $(".js-cache-reports").hide();
+            fixmystreet.offline.updateCachedReports();
+        } else {
+            $(".js-cache-reports").on("click", function() {
+                $(".js-cache-reports").hide();
```
dracos, line 430: Think this could have a preventDefault, just to stop # in the URL?

templates/web/base/my/planned.html
```diff
@@ -29,6 +29,12 @@ <h1>[% loc('Your shortlist') %]</h1>
 <div id="js-reports-list">
 [% INCLUDE 'my/_problem-list.html' shortlist = 1 %]
 </div>
+<div class="shadow-wrap">
```
dracos, line 32: Could we have this only appear if JS is working? Sticking a "hidden-nojs" on this div should do it.

web/cobrands/fixmystreet/offline.js
dracos: Also this hides the link but keeps the bottom bar, should we hide the whole div?

dracos (changes requested):
Nice, only a few small suggestions :)

dracos (approved):

PR #5207, opened 2024-10-07
Added Geolocate button before 'Go' button in PostcodeForm

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4561

Currently users using assistive devices might not use the "Use my current location" button because it comes after the "Go" button. This commit includes a geolocate button visible only for screen readers, located right before the "Go" button.

Some comments:

- Because the markup is slightly different for both geolocate buttons I thought of creating a new template just for the screen-reader one. Using the Id attribute for the function in JS wouldn't be the best approach anymore so I changed them to classes instead. But let me know if you think of a better approach.
- The styling however is not ideal because when a user focus on the invisible button, this one becomes visible, but most users don't use the keyboard anyway. 

https://github.com/user-attachments/assets/7558e548-9f0d-4492-98b8-40aaf7fcdf42

- While working on this ticket I thought that a better approach would have been to put the invisible button before the postcode input, that way we avoid scenarios where a user will still add the postcode/address to later find out there is a geolocate button. But this approach would need some tweak because we currently focus straightaway on the postcode input field, so to not loose that I thought that another approach would be: We don't include the invisible button at all, instead we add an `aria-label` in the postcode input with something like -> "Add a postcode or address. Alternatively, there is a geolocate button at the end of this form." Probably not the most elegant, but I think this approach is the most UX friendly one.

Let me know if you can think of a better one or if you rather any of the above.

### Pending

- [ ] Alert page

[skip changelog]

### DIFF
--- a/templates/web/base/around/_postcode_form_geolocation.html
+++ b/templates/web/base/around/_postcode_form_geolocation.html
@@ -1,2 +1,2 @@
-<a href="[% url | html %]" id="geolocate_link" aria-live="polite">[% INCLUDE 'around/geolocate_link_icon.html' %]
+<a href="[% url | html %]" id="geolocate_link" class="js-geolocate-link" aria-live="polite">[% INCLUDE 'around/geolocate_link_icon.html' %]
   [% loc('Use my current location') %]</a>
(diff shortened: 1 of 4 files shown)


dracos (commented):
If we're having two links, class seems fine to me, yep. But I like your idea of an aria-label, that seems to cover letting people know it exists without changing the behaviour (which I understand no-one has actively had an issue with). I can't actually find any other useful examples of anyone doing the same sort of thing (eg. in Google Maps the locate me is a long way after the search box tabwise, and I imagine a lot more people are using that...)

We currently have aria-describedby of pc-hint and "search-help pc-error" if there was a lookup error. Perhaps this could be added to or changed to cover this as well. Perhaps: "Add a postcode or address to search for here, or there is a geolocate button at the end of this form."?

But if you'd prefer this solution you currently have, I don't think it has any real downsides, I'd be okay with it also.

lucascumsille:
Replaced by: https://github.com/mysociety/fixmystreet/pull/5227

PR #5208, opened 2024-10-07
Improve meaningful sequence in report page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4560

Currently users using keyboard might miss that they can provide an update for a report, because the keyboard jumps straight to the key-tools like "Get udpates" which is at the bottom of the page. Also depending on the length of the report all the update related section might be down the bottom and invisible make it even less obvious that a user has elements they can interact with.

Please let me know if the changes in the order might cause some problems. 

As a side note, the TFL audit also suggested to do the same in the /around page, which at first seems sensible, but when you have an area with a lot of reports the amount of keyboard navigation through the whole list of reports to reach to the "key-tools" make it really annoying. As an alternative if we wanted to implement this in the /around page would be to add an invisible link right before the list of reports that takes you to the "key-tools" div.

https://github.com/user-attachments/assets/dd6348fd-8201-4efa-be14-c208bd21b10b

[skip changelog]

### DIFF
--- a/templates/web/base/alert/_updates.html
+++ b/templates/web/base/alert/_updates.html
@@ -1,15 +1,16 @@
 <form action="[% c.uri_for( '/alert/subscribe' ) %]" method="post">
-
     <p><a href="[% c.uri_for( '/rss', problem.id ) %]">
         <img src="/i/feed.png" width="16" height="16" title="[% loc('RSS feed') %]" alt="[% loc('RSS feed of updates to this problem' ) %]" border="0" style="float:right">
     </a>
 
     [% IF NOT c.user_exists OR c.user.alert_updates_by(c.cobrand) != 'phone' %]
-        [% loc('Receive email when updates are left on this problem.') %]
+        <span id="report-updates-data-title">[% loc('Receive email when updates are left on this problem.') %]</span>
     [% ELSE %]
-        [% loc('Receive a text when updates are left on this problem.') %]
-    [% END %]
+        <span id="report-updates-data-title">[% loc('Receive a text when updates are left on this problem.') %]
+          [% END %]</span>
     </p>
+
+    <button type="button" class="close-drawer screen-reader-only" aria-label="Close this dialog"></button>
 
     [% PROCESS 'auth/form_extra.html' %]
 
--- a/templates/web/base/report/display.html
+++ b/templates/web/base/report/display.html
@@ -45,18 +45,19 @@
 [% END %]
 
 [% TRY %][% INCLUDE 'report/_message_manager.html' %][% CATCH file %][% END %]
-[% INCLUDE 'report/display_tools.html' %]
+[% INCLUDE 'report/display_tools_mobile.html' %]
 [% TRY %][% INCLUDE 'report/sharing.html' %][% CATCH file %][% END %]
 [% INCLUDE 'report/updates.html' %]
 
 [% IF problem.duplicate_of %]
-    [% INCLUDE 'report/duplicate-no-updates.html' %]
+[% INCLUDE 'report/duplicate-no-updates.html' %]
 [% ELSIF problem.cobrand_data == 'noise' %]
-    <p>[% loc('This report is now closed to updates from the public.') %]</p>
+<p>[% loc('This report is now closed to updates from the public.') %]</p>
 [% ELSIF NOT shown_form  %]
-    [% PROCESS 'report/update-form-wrapper.html' %]
+[% PROCESS 'report/update-form-wrapper.html' %]
 [% END %]
 
+[% INCLUDE 'report/display_tools.html' %]
   </div>
 
   [% second_column | safe %]
--- a/templates/web/base/report/display.html
+++ b/templates/web/base/report/display.html
@@ -46,8 +46,6 @@
 
 [% TRY %][% INCLUDE 'report/_message_manager.html' %][% CATCH file %][% END %]
 [% INCLUDE 'report/display_tools_mobile.html' %]
-[% TRY %][% INCLUDE 'report/sharing.html' %][% CATCH file %][% END %]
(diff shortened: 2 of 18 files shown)


dracos (changes requested):
This doesn't work on mobile - the positioning in the source is a row of buttons after the report itself, before its updates. Having these right at the bottom under the update form, they'll be lost, I think? I'm not saying their current position is perfect, perhaps we can improve on that, but it does at least seem a sensible place for them to go at present. (And the tabbing order is fine then, because they're not out of any order).

I don't think we should change an around page, no. Perhaps the whole drawer concept could be changed, but I imagine everyone is used to it... An argument could be made that for consistency (WCAG SC 3.2.4), this counts as surrounding navigation and thus should be consistent between the around page and a report page? However, I agree it's a bit odd tabbing on desktop (because it comes after the report description, before the updates, so 'in the middle') - we could have separate tools row on mobile/desktop, ugh, I can't really think of an easier solution, but perhaps we can think of something different entirely.

(I notice when testing a report page that the "Shortlist" button isn't tabbable to at all, why aren't they picking up actual problems like this?)

lizettal:
@dracos could you please look at this on Monday so Lucas can potentially get this finished before he does on leave from Tuesday?

templates/web/base/alert/_updates.html
```diff
@@ -1,16 +1,17 @@
...
-        [% loc('Receive email when updates are left on this problem.') %]
+        <span id="report-updates-data-title">[% loc('Receive email when updates are left on this problem.') %]</span>
     [% ELSE %]
-        [% loc('Receive a text when updates are left on this problem.') %]
-    [% END %]
+        <span id="report-updates-data-title">[% loc('Receive a text when updates are left on this problem.') %]
+          [% END %]</span>
     </p>
 
+    <button type="button" class="close-drawer screen-reader-only" aria-label="Close this dialog"></button>
```
dracos, line 13: Is it better to use aria-label or just have the button have the same text as its actual contents?

templates/web/base/reports/_rss.html
```diff
@@ -1,7 +1,7 @@
 <div id="key-tools-wrapper" class="shadow-wrap">
     <ul id="key-tools">
-        <li class="feed-list-wrapper"><a rel="nofollow" id="key-tool-updates-area" class="feed" href="[% rss_url %]">[%
-            SET monikers = ['bromley','hounslow'];
+        <li class="feed-list-wrappe"><a rel="nofollow" id="key-tool-updates-area" class="feed has-inline-svg" href="[% rss_url %]">
```
dracos, line 3: Lost an r on wrapper

templates/web/base/reports/_list-filters.html
```diff
@@ -27,7 +27,7 @@
   [% END %]
 [% END %]
 
-<div class="report-list-filters-wrapper govuk-fieldset-wrapper govuk-small">
+<div id="report-list-filters" class="report-list-filters-wrapper govuk-fieldset-wrapper govuk-small">
```
dracos, line 30: Don't think this ID is used?

templates/web/base/report/display.html
```diff
@@ -45,18 +45,19 @@
...
-    <p>[% loc('This report is now closed to updates from the public.') %]</p>
+<p>[% loc('This report is now closed to updates from the public.') %]</p>
 [% ELSIF NOT shown_form  %]
-    [% PROCESS 'report/update-form-wrapper.html' %]
+[% PROCESS 'report/update-form-wrapper.html' %]
 [% END %]
 
+[% INCLUDE 'report/display_tools.html' %]
+[% TRY %][% INCLUDE 'report/sharing.html' %][% CATCH file %][% END %]
+[% INCLUDE 'report/updates.html' %]
```
dracos, line 60: The get updates and share things are appearing in the wrong place at mobile width for staff - you click either button and the extra stuff appears too far down, underneath the "Provide an update" button rather than above it. And for non-staff, the buttons aren't working at all, nothing happens when clicked.

templates/web/base/report/display_tools_mobile.html
```diff
@@ -0,0 +1,43 @@
+<div class="shadow-wrap">
```
dracos, line 1: This template seems very similar to the non-mobile one, I can combine them later.

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1293,12 +1323,11 @@ $.extend(fixmystreet.set_up, {
 
     if ($('.mobile').length) {
         // Make sure we end up with one Get updates link
-        if ($('#key-tools a.js-feed').length) {
+        if ($('#key-tools-mobile a.js-feed').length) {
```
dracos, line 1326: This change shouldn't be made - the js-feed is still in key-tools on any page it is used (not a report page), and needs to moved from there to sub_map_links.
And I'm afraid that sub_map_links uses a background image, so you then end up with two RSS feed icons being shown.
image

dracos (changes requested):
Overall seems like the right approach, but some things are not working, I'm afraid.

"Added accessibility improvements to small_drawer" contains a lot more in that commit than that, and other commits appear to have got mixed up as well.
It would be good if we could break this down into:
1. A commit to switch from external to inline SVGs, no other changes (that would also involve e.g. fixing the issue where the Feed link is moved to sub_map_links).
2. A commit to add the close button to the small drawer (and anything else related to that, but nothing more).
3. The skip button on around page if needed.
4. Changes to have a separate key-tools for desktop than mobile (fixing the issue with placement, not working for non-staff, etc)
Let me know if you have any questions

lucascumsille:
Replaced by: https://github.com/mysociety/fixmystreet/pull/5354

PR #5209, opened 2024-10-07
[Bexley] Handle incoming pushed waste updates

MorayMySoc:
https://github.com/mysociety/societyworks/issues/4095
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
@@ -20,22 +20,90 @@ has log_ident => (
     }
 );
 
+has whitespace_config => (
+    is => 'rw',
+    isa => 'HashRef',
+    default => sub { {} }
+);
+
 sub receive_whitespace_event_notification : Path('/waste/whitespace') : Args(0) {
     my ($self, $c) = @_;
 
-    my %headers = $c->req->headers->flatten;
-    $self->log($c->req->method);
-    $self->log(\%headers);
-    $self->log($c->req->parameters);
-    if ($c->req->body) {
-        my $soap = join('', $c->req->body->getlines);
-        $self->log($soap);
-    } else {
-        $self->log('No body');
-    }
+    require SOAP::Lite;
 
+    $c->detach('/waste/echo/soap_error', [ 'Invalid method', 405 ]) unless $c->req->method eq 'POST';
+
+    my $whitespace = $c->cobrand->feature('whitespace');
+    $self->whitespace_config($whitespace);
+    $c->detach('/waste/echo/soap_error', [ 'Missing config', 500 ]) unless $whitespace;
+
+    $c->detach('/waste/echo/soap_error', [ 'Missing body' ]) unless $c->req->body;
+    my $soap = join('', $c->req->body->getlines);
+    $self->log($soap);
+
+    my $body = $c->cobrand->body;
+    $c->detach('soap_error', [ 'Bad jurisdiction' ]) unless $body;
+
+    my $env = SOAP::Deserializer->deserialize($soap);
+    if ($env->valueof('//WorksheetPoke/secret') ne $self->whitespace_config->{missed_collection_secret}) {
+        return $c->detach('/waste/echo/soap_error', [ 'Unauthorized', 401 ]);
+    };
+
+    # Return okay if we're in endpoint test mode
+    my $cobrand_check = $c->cobrand->feature('waste');
+    $c->detach('soap_ok') if $cobrand_check eq 'echo-push-only';
+
+    my %params;
+    $params{external_id} = $env->valueof('//WorksheetPoke/worksheetId');
+    $params{status} = $env->valueof('//WorksheetPoke/status');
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
```diff
@@ -20,22 +22,96 @@ has log_ident => (
...
+    return 1;
+}
+
+sub _check_params {
+    my $hash = shift;
+
+    for my $key (keys %{$hash}) {
+        return 1 unless $hash->{$key};
+    }
+}
```
chrismytton, line 115: I think this will just return `1` if there's _any_ key present, rather than checking that they're _all_ present.

perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
```diff
@@ -20,22 +22,96 @@ has log_ident => (
...
     $c->response->body('OK');
 }
 
+sub update_report {
+    my ($self, $params) = @_;
+
+    my $report = FixMyStreet::DB->resultset("Problem")->search( { external_id => 'Whitespace-' . $params->{external_id} } )->first;
+    return unless $report;
+
+    my $status_hash = $self->whitespace_config->{missed_collection_state_mapping}{$params->{status}};
```
chrismytton, line 71: Might be worth checking the status is actually found in the state_mapping, in case they add new statuses in the future or similar.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -52,101 +52,6 @@ C<0001-01-01T00:00:00> represents an undefined date in Whitespace.
 
 use constant WHITESPACE_UNDEF_DATE => '0001-01-01T00:00:00';
 
-sub waste_fetch_events {
```
chrismytton, line 55: This method is called by the `bin/fixmystreet.com/fetch-waste` script, which is used when there's a need to manually pull events from Whitespace, in case a push is missed or similar.

Would be good if we could keep this method working so we've got the option to pull updates, similar to how the Echo ones work.

chrismytton (changes requested):

perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
MorayMySoc: If the status isn't present the data should now be rejected

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
MorayMySoc: I've broken everything down into separate commits rather than just fixups to try and keep this clear as it was a lot of changes moving back into the files. Let me know if it's not clear.

I have added a fixup at the end as I think that I had overlooked the comment id is probably now actually supplied in the data if I've understood the data correctly.

perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
MorayMySoc: I've added a test which hopefully shows this failing if one key is missing. Otherwise tests pass with all data present, so I think it's ok, but let me know if not.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
dracos: The worksheet reference is our report ID, so it's not a unique update ID. We could concatenate the worksheet ID and the timestamp?

perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
```diff
@@ -20,22 +20,86 @@ has log_ident => (
...
         my $soap = join('', $c->req->body->getlines);
         $self->log($soap);
+        my $env = SOAP::Deserializer->deserialize($soap);
+        if ($env->valueof('//WorksheetPoke/secret') ne $self->whitespace_config->{missed_collection_secret}) {
+            return $c->detach('/waste/echo/soap_error', [ 'Unauthorized', 401 ]);
+        };
+        my %params;
+        $params{external_id} = $env->valueof('//WorksheetPoke/worksheetId');
+        $params{status} = $env->valueof('//WorksheetPoke/status');
+        $params{completed_dt} = $env->valueof('//WorksheetPoke/completedDate');
```
dracos, line 52: This isn't a DateTime, it's just a string, I think the update code assumes in a couple of places it'll be a DateTime. `updated_datetime` is what the Echo code uses.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
dracos: Or just use 'waste' as that is what Echo use for the update_id, to match.

perllib/FixMyStreet/App/Controller/Waste/Whitespace.pm
dracos: It appears the statuses in the push are actually only "Completed" or "Cancelled", not the states you'd get from the worksheet details.

PR #5210, opened 2024-10-08
[Open311] Add 'staff' keyword handling

dracos:
The main point of this PR is to add the ability for an Open311 endpoint to use a 'staff' keyword which will then automatically mark the category as staff only when it's created/updated. That's the last commit.

However, while implementing it it seemed good to tidy up the populate script so that it only updated the contact table once for a category (it was possible for a new Open311 category to update the DB once to create the category, once to add extra fields (with no note), once to set the group, once to set as non public or inactive or waste; updating a category would only do it for changes, but could still leave more than one). And then also I thought it good to make a difference between undeleting a category and having its name/destination change (in the former I think we can drop any devolved send method; and in the latter case keep the state rather than always switch to confirmed, which could eg potentially lose a staff status).

### DIFF
--- a/perllib/FixMyStreet/Test.pm
+++ b/perllib/FixMyStreet/Test.pm
@@ -69,7 +69,7 @@ BEGIN {
         my ($self, $path, $content, $code) = @_;
         my $test_res = HTTP::Response->new();
         $test_res->code($code || 200);
-        $test_res->content($content);
+        $test_res->content(encode_utf8($content));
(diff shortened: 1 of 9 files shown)


MorayMySoc (approved):
Looks good to me - thanks for adding the removal of the send_method on the deleted contacts...

dracos:
Am trying this out on staging first, looks to have not deleted all the contacts at least.

PR #5211, opened 2024-10-08
[P'bro][WW] Limit bulky location details to 250 characters

nephila-nacrea:
Bartec has a 250 char limit on the CREW_NOTES field (the 'location' field our end); reports fail to send if this limit is exceeded.

For https://3.basecamp.com/4020879/buckets/32749602/todos/7873595742 /  fixes https://github.com/mysociety/societyworks/issues/4568 .

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
@@ -53,6 +53,11 @@ has_page location => (
             $fields->{location}{label} = 'Please tell us where you will place the items for collection (include any access codes the crew will need)';
             $fields->{location}{tags}{hint} = 'For example, ‘On the driveway’';
         }
+
+        my $maxlength
+            = $form->c->cobrand->call_hook('bulky_location_max_length');
+        $fields->{location}{maxlength} = $maxlength if $maxlength;
+
         return $fields;
     },
 );
--- a/perllib/FixMyStreet/Cobrand/Peterborough/Bulky.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough/Bulky.pm
@@ -39,6 +39,12 @@ a bulky waste collection
 
 sub max_bulky_collection_dates       {2}
 
+=item * Max length of location details text is 250 characters
+
+=cut
+
+sub bulky_location_max_length {250}
+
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
```diff
@@ -302,6 +302,18 @@ sub validate {
         }
     }
 
+    if ( $self->current_page->name eq 'location' ) {
```
dracos, line 305: I think this could be done as a `validate_method` on the `location` field? Just to keep it together; would have the same effect.

perllib/FixMyStreet/Cobrand/Peterborough/Bulky.pm
```diff
@@ -39,6 +39,12 @@ a bulky waste collection
 
 sub max_bulky_collection_dates       {2}
 
+=item * Max length of location details text is 250 characters
+
+=cut
+
+sub location_max_length {250}
```
dracos, line 46: I'd call it bulky_location_max_length to be clear.

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
dracos: In fact, if you move it to the `update_field_list` of the `location` page - you can see the special case things happening there already - I think all you need to do is set `$fields->{location}{maxlength}` and then the error handling is done for you, plus the client side will have the maxlength set so people can't go over anyway.

dracos (changes requested):
I think this can be moved to a different place to have the same effect with more assistance from existing code :)

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
nephila-nacrea: If by client side you mean the JS picking up on the error before the form is submitted, that does not happen - is it meant to?

dracos: Nothing JS involved, it sets `maxlength` on the HTML, which browsers prevent typing longer than in the box. I've just tried it locally, seems to work fine, maxlength=250 is present on the textarea and I can't enter more than 250 characters in the box.

dracos (approved):
Yeah, that works :)

PR #5213, opened 2024-10-09
Added GOVUK radio buttons on National Highways report message

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4549

Currently we use the segmented-control(pill) component, unfortunately it does not cover cases where the address of the report is quite long. This causes horizontal overflow and UI wise both buttons look quite uneven, because one is quite long while the other one says "Somewhere else".

With this change I think is also more obvious that they are options and not buttons. For example I tried stacking both "buttons" vertically so the first button being too long is not a problem anymore, unfortunately I think in terms of UX is a bit confusing because they look just like `.btn` and `btn--primary` so it just look like styled links, where one has more emphasis than the other one, instead of what they really are: option A which is active, and then we have option B.

Screenshot 2024-10-09 at 11 32 26

[skip changelog]


neprune (approved):
Nice, sorry for the delay on this. Looks good to me, radio buttons rather than stacked buttons definitely seems better.

lucascumsille:
Not a problem @neprune thank you for the help =)

PR #5217, opened 2024-10-09
[Bexley][WW] Display links for new container & removal requests

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/4570.

Displays links on bin days for requesting delivery and removal of waste containers.

Also implements a rough version of the first page of the request form.

[skip changelog]

### DIFF
--- a/t/app/controller/waste_bexley_container_requests.t
+++ b/t/app/controller/waste_bexley_container_requests.t
@@ -1,0 +1,451 @@
+use utf8;
+
+use FixMyStreet::Cobrand::Bexley;
+use FixMyStreet::TestMech;
+use Test::Deep;
+use Test::MockModule;
+use Test::MockObject;
+use Test::MockTime 'set_fixed_time';
+
+FixMyStreet::App->log->disable('info');
+END { FixMyStreet::App->log->enable('info'); }
+
+set_fixed_time('2024-03-31T01:00:00'); # March 31st, 02:00 BST
+
+my $addr_mock = Test::MockModule->new('BexleyAddresses');
+# We don't actually read from the file, so just put anything that is a valid path
+$addr_mock->mock( 'database_file', '/' );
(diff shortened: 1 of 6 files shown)


t/app/controller/waste_bexley_container_requests.t
chrismytton, line 1: With these tests I think we ideally want integration tests, i.e. ones that make GET requests to web endpoints, rather than unit tests like these that test the implementation of a individual methods.

Obviously a lot of stuff won't work properly at this point! But I think integration tests provide more comprehensive testing of the user interaction, and are generally a bit easier to maintain as they don't rely as much on the underlying implementation.

If you `git grep "subtest 'Request"` there are various existing tests that can hopefully be copied and tweaked a bit.

chrismytton (changes requested):
This looks great so far! The controller code all looks good, from what I can tell. Just one suggestion about the tests.

t/app/controller/waste_bexley_container_requests.t
nephila-nacrea: I am adding integration tests as part of https://github.com/mysociety/fixmystreet/pull/5219 . Not sure if it would make things complicated by trying to add the beginning of one here and rebasing 🤔 
I like having both these unit-type tests and the more complex integration tests. In this case, I wanted an easy way of checking that my data munging was correct for _set_request_containers(), as there's a lot going on in that one sub, and didn't want other code getting in the way.
What are your thoughts?

nephila-nacrea: Having said all that...I have a bug that would be better (thematically) to fix in this PR, so I think I'll try to add & push an integration test for that.

chrismytton: That all sounds sensible to me 🙂

chrismytton (approved):

PR #5218, opened 2024-10-14
[Oxfordshire] Include staff role in CSV export

davea:
For FD-4668

[skip changelog]


dracos (changes requested):
If you compare this with Northumberland, you'll see this is doing a bit too much in the pre-generated case, which already gets `staff_role` for free. So when `$csv->dbi` is set you only need the existing code; the fetching of csv_staff_users/csv_staff_roles and the setting of staff_role only needs to happen when dbi is not set. Otherwise looks fine

dracos (approved):
Thanks, sorry it's a bit fiddly for special things, perhaps there's a way to factor it out.

PR #5219, opened 2024-10-14
[Bexley][WW] Implement form for requesting new/replacement containers

nephila-nacrea:
NOTE: Currently on top of https://github.com/mysociety/fixmystreet/pull/5217 .

For https://github.com/mysociety/societyworks/issues/4575.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -791,6 +791,12 @@ sub request : Chained('property') : Args(0) {
             intro => 'request/intro.html',
             check_unique_id => 0,
             next => $next,
+            update_field_list => sub {
+                my $form = shift;
+                my $fields = $form->{c}->cobrand->call_hook(
+                    waste_request_form_update_field_list => $form ) // {};
+                return $fields;
+            },
         },
     ];
 
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -1235,16 +1235,16 @@ sub construct_bin_request_form {
 
     my $field_list = [];
 
-    my $request_type = $c->get_param('request-type');
+    my $request_type = $c->get_param('request_type');
 
     if ( $request_type eq 'delivery' ) {
         for my $container (
             @{ $c->stash->{property}{containers_for_delivery} } )
         {
             if ( $container->{subtypes} ) {
-                my $id = lc $container->{name} =~ s/ /-/gr;
+                my $id = $container->{name} =~ s/ /-/gr;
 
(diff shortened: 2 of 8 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -1236,10 +1237,12 @@ sub construct_bin_request_form {
         for my $container (
             @{ $c->stash->{property}{containers_for_delivery} } )
         {
+            my $id;
```
dracos, line 1240: Doesn't matter, but can't see why ID was pulled out of the if here, it's never used after.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -1273,10 +1274,28 @@ sub construct_bin_request_form {
                     type         => 'Checkbox',
                     label        => $container->{name},
                     option_label => $container->{description},
+                    tags => { toggle => "form-quantity-$id-row" },
```
dracos, line 1275: Should this only be included if the max is > 1 (I don't think the JS toggle will care, though, so doesn't matter)

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -859,6 +859,36 @@ sub request : Chained('property') : Args(0) {
             intro => 'request/intro.html',
             check_unique_id => 0,
             next => $next,
+            update_field_list => sub {
```
dracos, line 862: I see why this is here, and works fine, but I guess ideally we'd want it not to be, given it's Bexley specific :)
Wonder how best to do that. I guess we have the cobrand available in $form->{c} here (I think), so you could move this to a cobrand call_hook `munge_waste_request_page` or something like that, and then call it from here?

dracos (changes requested):
This looks good :)

perllib/FixMyStreet/App/Controller/Waste.pm
nephila-nacrea: Yeah, this wasn't going to be its permanent place, I need to refactor some things still but wanted to get something working :-) Yeah, I was thinking a call_hook would do the job in some way.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
nephila-nacrea: Yeah, I'm happy to leave this as is.

dracos (commented):
Thanks, all looks good :)

nephila-nacrea:
> Thanks, all looks good :)

@dracos, Thanks, just added one more small commit 🙏

dracos (approved):

PR #5226, opened 2024-10-22
Reduce some translation DB calls, and related

dracos:
Switch name to get_column('name') in a few places, as name does auto-translation lookup and we don't need this in these places, saves 2x(num contacts)+a few body calls (all quick but still). We only look for one row in moderation so request that (don't think this actually makes any material difference).
On the other side, realised there was some hard-coded waste category hiding that wasn't applying to some councils, so add that, didn't seem worth a PR on its own.
And  I wanted the test suite to be clean again. [skip changelog]


MorayMySoc (approved):
All looks good to me

PR #5227, opened 2024-10-23
Added aria-label to input#pc

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4561
Replaces: https://github.com/mysociety/fixmystreet/pull/5207

Because the Geolocate button comes after the submit button, assistive device users might not realise it even exists. So, we have added an aria-label to input#pc in the front and alert page that provides more context about the input postcode and a geolocate button at the end of the form.

[skip changelog]

### DIFF
--- a/templates/web/base/alert/index.html
+++ b/templates/web/base/alert/index.html
@@ -16,7 +16,7 @@
       [% INCLUDE 'errors.html' %]
     [% END %]
     <div>
-      <input type="text" id="pc" name="pc" value="[% pc | html %]" aria-describedby="pc-hint [% IF location_error %]email-error[% END %]" required>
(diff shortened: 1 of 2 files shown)


lucascumsille:
@dracos thanks for the feedback regarding
>We currently have aria-describedby of pc-hint and "search-help pc-error" if there was a lookup error. Perhaps this could be added to or changed to cover this as well. Perhaps: "Add a postcode or address to search for here, or there is a geolocate button at the end of this form."?

I did some testing with VoiceOver, this is what happened:

1. On loading the page will read the `aria-label` with the new info about the geolocate button at the end of the form.
2. I added a wrong name to trigger an error.
3. The content of input#pc gets announced again, also the aria-label.
4. Then it reads the content of p#pc-error and div#search-help.

Because on step 3 the geolocate button gets announced again I didn't include anything in p#pc-error or div#search-help. 

-------

At the end I went with this approach instead of the one in the [previous PR](https://github.com/mysociety/fixmystreet/pull/5207), because not only the UX is probably more friendly, the code is also more simple.

dracos (approved):
Thanks, looks/sounds good :)

PR #5229, opened 2024-10-23
Convert cobrand data to column on body

dracos:
Along with a couple of small things to prevent double lookups [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -419,11 +419,7 @@ sub visible_states_remove {
 sub public_asset_id {
     my $self = shift;
 
-    my $current_cobrand = $self->result_source->schema->cobrand;
-
-    my $cobrand_for_problem = $current_cobrand->call_hook(
-        get_body_handler_for_problem => $self );
-
+    my $cobrand_for_problem = $self->body_handler;
     return unless $cobrand_for_problem;
 
     # Should be of the form:
@@ -1043,8 +1039,7 @@ sub updates_sent_to_body {
     return unless $self->send_method_used && $self->send_method_used =~ /Open311/;
 
     # Some bodies only send updates *to* FMS, they don't receive updates.
-    my $cobrand = $self->get_cobrand_logged;
-    my $handler = $cobrand->call_hook(get_body_handler_for_problem => $self);
+    my $handler = $self->body_handler;
     my $ret = $handler && $handler->call_hook(updates_sent_to_body => $self);
     return $ret if defined $ret;
 
@@ -1219,6 +1214,17 @@ has get_cobrand_logged => (
         my $self = shift;
         my $cobrand_class = FixMyStreet::Cobrand->get_class_for_moniker( $self->cobrand );
         return $cobrand_class->new;
+    },
+);
+
+has body_handler => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my $self = shift;
+        my $cobrand = $self->get_cobrand_logged;
+        my $handler = $cobrand->call_hook(get_body_handler_for_problem => $self);
+        return $handler;
     },
 );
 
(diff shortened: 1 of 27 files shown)


perllib/FixMyStreet/TestMech.pm
```diff
@@ -653,6 +653,7 @@ sub create_body_ok {
     my ( $self, $area_id, $name, $params, $extra ) = @_;
 
     $params->{name} = $name;
+    $params->{cobrand} = delete $extra->{cobrand} if $extra->{cobrand};
```
davea, line 656: Is this solely so all the `create_body_ok` lines across all the `.t` files don't need updating or does it do something else? If not, perhaps updating all those lines would be a better approach. I think they all have at least an empty `{}` anyway so hopefully it'd just be the removal of a bunch of `}, {` instances.

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1222,6 +1217,17 @@ has get_cobrand_logged => (
     },
 );
 
+has body_handler => (
```
davea, line 1220: Would be good to have a brief comment explaining what this returns.

davea (changes requested):
When running tests with `-j10` this is causing problems, seemingly when multiple `.t` files try and create bodies with the same `cobrand` value which isn't permitted. Guess the transaction isolation doesn't extend to indexes in Postgres?

```
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st execute failed: ERROR:  deadlock detected
DETAIL:  Process 60372 waits for ShareLock on transaction 742; blocked by process 60377.
Process 60377 waits for ShareLock on transaction 737; blocked by process 60372.
HINT:  See server log for query details.
CONTEXT:  while inserting index tuple (0,17) in relation "body_cobrand_idx" [for Statement "INSERT INTO "body" ( "cobrand", "name") VALUES ( ?, ? ) RETURNING "id"" with ParamValues: 1='oxfordshire', 2='Oxfordshire County Council'] at /Users/davea/Code/fixmystreet/fixmystreet/perllib/FixMyStreet/TestMech.pm line 659
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 1.
t/app/controller/admin/defecttypes.t ......... Dubious, test returned 255 (wstat 65280, 0xff00)
All 1 subtests passed
```

That aside, a couple of small things I spotted.

perllib/FixMyStreet/TestMech.pm
dracos: No, was just so I didn't have to update all the files :) Yeah, guess so, I imagine in most cases that's all the extra was for.

dracos:
Unique indexes have to block if changed in more than one transaction, yeah, because otherwise one transaction could create something another one could conflict with, which you wouldn't know until the first committed or rolled back. That's fine, though, it just means a wait for the rollback; the issue here is then the first transaction **also** creates something that waits on the second to commit or roll back. So they're stuck for ever (which postgresql spots and kills one at random).

In this particular case you mention (and I get the same locally), the issue is that defecttypes.t creates a Bromley body, then an Oxfordshire body; whilst permissions.t creates an Oxfordshire body, then a Bromley body. So the timings generally end up meaning they deadlock each waiting for the other to commit/roll back.

We could make it non-unique, relying on the code to keep it unique (which is what the current situation is with extra, so no loss there); or we could fix the deadlocks in the code (in this case, swapping the body creation order in permissions.t to match - so there's never a chance of a deadlock. But permissions.t would still wait for defecttypes.t to finish, so possibly slower. And I don't know how many opposite order things there are in the tests.

davea (approved):
✨

dracos:
Note to prevent errors during a live deploy, I'm not going to let run the schema file as is, because after the first server, the cobrand extra would all have gone missing and everything would break until the other servers deployed.

So I will run the contents of https://github.com/mysociety/fixmystreet/blob/02685b3915a3420a7ee89eb1ec3ac34c0866143d/db/schema_0089-add-body-cobrand-column.sql but *without* the removal line, so that the data is still there for any old code looking for it there. Then once everything's deployed we can remove that manually.

PR #5230, opened 2024-10-24
Reduce body name translation db lookups

dracos:
[skip changelog]
I did some in f241299b1 but seemed to have missed quite a few! We don't need to check translations for all the overlapping asset stuff etc, we want to check against the actual db name.
This should cut down on the number of translation DB lookups when e.g. fetching /report/new/ajax pages.

I also try and fetch the translations first off, so that no more DB queries are needed anyway. (That's probably a better way of doing it I thought of after, but might as well leave the others, clearer for the comparison.)


MorayMySoc (approved):
Looks good to me

PR #5232, opened 2024-10-25
[Merton][WW] Allow transfer of ggw subscription

MorayMySoc:
When a Green Garden Waste subscriber in Merton moves to another address in Merton, they can transfer the remainder of their ggw subscription to the new address

1) As long as the new address does not already have a sub 
2) As long as there is a ggw on their previous property, and it is not in the renewal period
3) They have brought their bin(s) with them

This is a staff only feature.

https://github.com/mysociety/societyworks/issues/4551

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -18,6 +18,7 @@ use FixMyStreet::App::Form::Waste::Garden::Modify;
 use FixMyStreet::App::Form::Waste::Garden::Cancel;
 use FixMyStreet::App::Form::Waste::Garden::Renew;
 use FixMyStreet::App::Form::Waste::Garden::Sacks::Purchase;
+use FixMyStreet::App::Form::Waste::Garden::Transfer;
 use Memcached;
 use JSON::MaybeXS;
 
@@ -846,6 +847,54 @@ sub process_request_data : Private {
     return 1;
 }
 
+sub process_garden_transfer : Private {
+    my ($self, $c, $form) = @_;
+    my $data = $form->saved_data;
+
+    # Get the current subscription for the old address
+    my $old_property_id = $data->{previous_ggw_address}->{value};
+    #$c->forward('get_original_sub', ['', $old_property_id]);
+
+    my $base = {};
+    $base->{name} = $c->get_param('name');
+    $base->{email} = $c->get_param('email');
+    $base->{phone} = $c->get_param('phone');
+
+    # Cancel the old subscription
+    my $cancel = { %$base };
+    $cancel->{category} = 'Cancel Garden Subscription';
+    $cancel->{title} = 'Garden Subscription - Cancel';
+    $cancel->{address} = $data->{previous_ggw_address}->{label};
+    my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
+    my $end_date_field = $c->cobrand->call_hook(alternative_backend_field_names => 'Subscription_End_Date') || 'Subscription_End_Date';
+    $c->set_param($end_date_field, $now->ymd);
+    $c->set_param('property_id', $old_property_id);
+    $c->set_param('uprn', $data->{transfer_old_ggw_sub}{transfer_uprn});
+    $c->set_param('transferred_to', $c->stash->{property}->{uprn});
+    $c->forward('setup_garden_sub_params', [ $cancel, undef ]);
+    $c->forward('add_report', [ $cancel ]) or return;
+    $c->stash->{report}->confirm;
+    $c->stash->{report}->update;
+
+    # Create a report for it for the new address
+    my $new = { %$base };
+    $new->{category} = 'Garden Subscription';
+    $new->{title} = 'Garden Subscription - New';
+    $new->{bins_wanted} = $data->{transfer_old_ggw_sub}->{transfer_bin_number};
+    $new->{transfer_bin_type} = $data->{transfer_old_ggw_sub}->{transfer_bin_type};
+
+    my $expiry = $data->{transfer_old_ggw_sub}->{subscription_enddate};
+    $expiry = DateTime::Format::W3CDTF->parse_datetime($expiry);
+    $c->set_param($end_date_field, $expiry->ymd);
+    $c->set_param('property_id', '');
+    $c->set_param('uprn', '');
+    $c->set_param('transferred_from', $data->{transfer_old_ggw_sub}{transfer_uprn});
+    $c->forward('setup_garden_sub_params', [ $new, $c->stash->{garden_subs}->{New} ]);
+    $c->forward('add_report', [ $new ]) or return;
+    $c->stash->{report}->confirm;
+    $c->stash->{report}->update;
+}
+
 sub group_reports {
     my ($c, @reports) = @_;
     my $report = shift @reports;
@@ -1280,6 +1329,15 @@ sub garden_renew : Chained('garden_setup') : Args(0) {
     $c->forward('form');
 }
 
+sub garden_transfer : Chained('garden_setup') : Args(0) {
+    my ($self, $c) = @_;
+
+    $c->detach( '/page_error_403_access_denied', [] ) unless $c->stash->{is_staff};
+
+    $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Garden::Transfer';
+    $c->forward('form');
+}
+
 sub process_garden_cancellation : Private {
     my ($self, $c, $form) = @_;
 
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/Cobrand/Merton/Waste.pm
```diff
@@ -376,6 +376,34 @@ sub waste_post_report_creation {
...
+    my $old_services = $self->{api_serviceunits};
+
+    # Reset things as this is still the property we're actually on
+    $self->look_up_property($self->{c}->stash->{property}{id});
+
+    my ($old_garden) = grep { $_->{ServiceId} eq '409' } @$old_services;
+    $old_garden->{transfer_uprn} = $details->{uprn};
+
+    $error = 'no_previous' unless $error || ($old_garden && $old_garden->{ServiceTasks});
+    my $subscription_enddate = $old_garden->{ServiceTasks}->{ServiceTask}->{ServiceTaskSchedules}->{ServiceTaskSchedule}[0]->{EndDate}->{DateTime} unless $error;
```
dracos, line 400: You cannot assume this structure, Echo has some strange data in it with multiple services overlapping or what have you. Probably not the case for garden I hope, but better safe and to be consistent with other uses. I think you can call `garden_current_service_from_service_units($old_services)` to get the current service task, and then call `_parse_schedules(that servicetask)` to get something containing `->{end_date}` that matches what is displayed to the user.

t/app/controller/waste_merton_garden.t
```diff
@@ -1164,37 +1164,140 @@ FixMyStreet::override_config {
...
+        }});
+
+        $mech->submit_form_ok({ with_fields => {
+            addresses => '12345'
+        }});
+        $mech->content_contains('This should be the old address, not the new one',
+        'Can not transfer from current address');
+
+        $echo->mock('GetServiceUnitsForObject', sub {
+            if ($_[1] == '12345') {
```
dracos, line 1231: This always returns sacks :)

t/app/controller/waste_merton_garden.t
```diff
@@ -1164,37 +1164,140 @@ FixMyStreet::override_config {
...
+            }
+        });
+        $mech->submit_form_ok({ with_fields => {
+            addresses => '11345'
+        }});
+        $mech->content_contains('There is no garden subscription at this address',
+        'Can not transfer from property that does not have a ggw subscription');
+
+        $echo->mock('GetServiceUnitsForObject', sub {
+            if ($_[1] == '12345') {
```
dracos, line 1244: This always returns a bin

t/app/controller/waste_merton_garden.t
```diff
@@ -1164,37 +1164,140 @@ FixMyStreet::override_config {
...
+        );
+        my ($new_report, $cancel_report) = $recent_reports->all;
+        is $new_report->user->name, $p->user->name, 'User name on report';
+        is $new_report->user->email, $p->user->email, 'User email on report';
+        is $new_report->title, 'Garden Subscription - New', 'New report title correct';
+        is $new_report->detail, "Garden Subscription\n\n2 Example Street, Merton,", 'New report detail correct';
+        is $new_report->get_extra_field_value('uprn'), '1000000002', 'Correct uprn on new report';
+        is $new_report->get_extra_field_value('property_id'), '12345', 'Correct property id on new report';
+        is $new_report->get_extra_field_value('current_containers'), 1, 'Correct bin containers amount set on new report';
+        #is $new_report->get_extra_field_value('End_Date'), '2021-03-30', 'Subscription time transferred';
```
dracos, line 1291: Why is this line commented out?

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -835,6 +836,57 @@ sub process_request_data : Private {
...
+    $c->forward('setup_garden_sub_params', [ $cancel, undef ]);
+    $c->forward('add_report', [ $cancel ]) or return;
+    $c->stash->{report}->confirm;
+    $c->stash->{report}->update;
+
+    # Create a report for it for the new address
+    my $new = { %$base };
+    $new->{category} = 'Garden Subscription';
+    $new->{title} = 'Garden Subscription - New';
+    my $old_subscription_bin_data = $data->{transfer_old_ggw_sub}->{ServiceTasks}->{ServiceTask}->{Data}->{ExtensibleDatum}[0]->{ChildData}->{ExtensibleDatum};
```
dracos, line 871: Similar to before, should use the existing code to get the service task (in case there's more than one, this wouldn't work as {ServiceTasks} would be an array). And then you'd need to use force_arrayref in the Data as well, because you don't know if that will be array or single either.

perllib/FixMyStreet/Cobrand/Merton/Waste.pm
dracos: Also, if we look this up here, can we store it on $old_garden to not need to look it up again, like you've done with transfer_uprn

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -835,6 +836,57 @@ sub process_request_data : Private {
...
+    for my $bins (@{$old_subscription_bin_data}) {
+        if ($bins->{DatatypeName} eq 'Quantity') {
+            $new->{current_bins} = $bins->{Value};
+        } elsif ($bins->{DatatypeName} eq 'Container Type') {
+            $new->{current_type} = $bins->{Value};
+        }
+    }
+    #my $expiry = $data->{transfer_old_ggw_sub}->{ServiceTasks}->{ServiceTask}->{ServiceTaskSchedules}->{ServiceTaskSchedule}[0]->{EndDate}->{DateTime};
```
dracos, line 879: Yes, so store this before and look it up here, set it on the field (which is `End Date` also, yes) - will need adding to the test's category extra data fields list when set up, then it should be present in your test.

dracos (changes requested):
This is looking good :) I don't think any of my comments are major, and I think the end date should hook up fine for the new sub.

perllib/FixMyStreet/App/Form/Waste/Garden/Transfer.pm
```diff
@@ -0,0 +1,162 @@
...
+    lazy => 1,
+    default => sub {
+        my $self = shift;
+        my $c = $self->{c};
+
+        my $p = $c->model('DB::Problem')->search({
+        category => 'Garden Subscription',
+        title => ['Garden Subscription - New', 'Garden Subscription - Renew'],
+        extra => { '@>' => encode_json({ "_fields" => [ { name => "property_id", value => ($self->saved_data->{previous_ggw_address}[0]->{value}) } ] }) },
+        state => { '!=' => 'hidden' }
```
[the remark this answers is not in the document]
MorayMySoc: I got this from `perllib/FixMyStreet/App/Controller/Waste.pm` Not clear how it's setup with open_states? Is there an example?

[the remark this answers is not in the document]
dracos: Sorry, wasn't clear, but yes, a grep for that would have given you something like `state => [ FixMyStreet::DB::Result::Problem->open_states ],`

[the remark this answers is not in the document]
MorayMySoc: Thanks - I grepped, but then put curly braces around it! :-)

perllib/FixMyStreet/Cobrand/Merton/Waste.pm
```diff
@@ -390,12 +390,20 @@ sub check_ggw_transfer_applicable {
...
-    my $subscription_enddate = $old_garden->{ServiceTasks}->{ServiceTask}->{ServiceTaskSchedules}->{ServiceTaskSchedule}[0]->{EndDate}->{DateTime};
+    my $servicetask = $self->garden_current_service_from_service_units($old_services);
 
+    return { error => 'no_previous' } unless $servicetask;
+
+    my $subscription_enddate = _parse_schedules($servicetask)->{end_date};
     return { error => 'due_soon' } if ($subscription_enddate && $self->waste_sub_due($subscription_enddate));
 
+    my $old_subscription_bin_data = Integrations::Echo::force_arrayref($servicetask->{Data}, 'ExtensibleDatum');
+    my ($bin_number) = grep { $_->{DatatypeName} eq 'Quantity'} @{$old_subscription_bin_data->[0]->{ChildData}->{ExtensibleDatum}};
```
dracos, line 401: You need to loop through `@$old_subscription_bin_data` here, I think - you do not know that the garden information will be in the first entry.

dracos (changes requested):
Great, think you've covered everything I asked - just a couple of knock-on things from that.

dracos (changes requested):
Almost, but... the childdata might be a hash or an array, I'm afraid! See the double loop for the same type of data at https://github.com/mysociety/fixmystreet/blob/master/perllib/FixMyStreet/Roles/Cobrand/Echo.pm#L415

dracos (approved):

dracos (approved):

MorayMySoc:
open311 changes also in 4551-merton-ggw-transfers on servers

dracos (approved):
Thanks!

PR #5239, opened 2024-10-31
[Bexley][WW] Adds message about what containers not collected

MorayMySoc:
https://github.com/mysociety/societyworks/issues/4587


templates/web/bexley/waste/container_removal_intro.html
```diff
@@ -0,0 +1,2 @@
+<p>We are unable to collect kitchen caddies. Please dispose of unwanted kitchen caddies by taking them to a Reuse and Recycling Centre or by placing them in your household waste bin</p>
```
nephila-nacrea, line 1: I think you need full stops at the very end of both of these \<p> blocks.
Otherwise, looks good, and thanks for picking this up!

nephila-nacrea (changes requested):
You'll want to merge to `bexley-ww-container-requests` once approved. You may need to do some wrangling your end as here there seem to be commits from one of my branches (that I have since merged).

PR #5243, opened 2024-11-04
[Waste] Bulky items list editing improvements

davea:
- Shrinks sidebar
 - Removes per-item price field if feature not enabled
 - Uses entire page width for container

For FD-4697

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
@@ -125,8 +125,6 @@ sub bulky_items : Chained('body') {
     $c->stash->{item_list} = $cfg->{item_list} || [];
 
     my $cobrand = $c->stash->{body}->get_cobrand_handler;
-    $c->stash->{available_features} =
-        $cobrand->call_hook('bulky_available_feature_types') if $cobrand;
     $c->stash->{per_item_pricing_property_types} =
         $cobrand->call_hook('bulky_per_item_pricing_property_types');
 
@@ -144,7 +142,6 @@ sub bulky_items : Chained('body') {
             }
             my $item = {
                 bartec_id => $c->get_param("bartec_id[$i]"),
-                category => $c->get_param("category[$i]"),
                 name => $c->get_param("name[$i]"),
                 message => $c->get_param("message[$i]"),
                 price => $c->get_param("price[$i]"),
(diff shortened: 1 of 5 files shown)


dracos (commented):
Looks good. As far as I'm aware, the available_feature type dropdown was never used/worked, so I think that can be dropped as well, replaced with the small text field the others all have, that'd save more space as well. And the 'category' is not used by anything at all, either; might not need the shrunken sidebar :)

web/cobrands/sass/_admin.scss
```diff
@@ -343,6 +343,15 @@ textarea.code {
     max-width: initial;
 }
 
+// Shrink the sidebar to make more room for the main content
+.container--sidebar .sticky-sidebar {
```
dracos, line 347: Don't follow this fixup, sorry - this changes the sidebar everywhere, rather than just the bulky items page? Might have misunderstood.

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -125,8 +125,6 @@ sub bulky_items : Chained('body') {
     $c->stash->{item_list} = $cfg->{item_list} || [];
 
     my $cobrand = $c->stash->{body}->get_cobrand_handler;
-    $c->stash->{available_features} =
-        $cobrand->call_hook('bulky_available_feature_types') if $cobrand;
```
dracos, line 129: You can remove the function from Peterborough/Bulky.pm as well.

dracos (changes requested):
One extra thing can be removed, and one question about a change

web/cobrands/sass/_admin.scss
davea: Yep, that's right. Figured it didn't hurt and made a little more room for main content on all admin pages. Do you think that's OK?

dracos: It's not just the admin though - it's all pages :)

davea: Oops, I thought `_admin.scss` was scoped to admin pages. Will update.

dracos (changes requested):
Sorry to be annoying, but this is added in the base stylesheet - so would affect the width 100% at mobile widths. Now it's in .admin I don't think you need the !important, but you do need to restrict the parent one to desktop widths only, hopefully something like this works, do check:
```
.admin {
    ...
    .container--sidebar {
        .sticky-sidebar aside {
            width: 11em;
        }
        @media only screen and (min-width: 48em) {
            .sticky-sidebar {
                width: 11em;
            }
       }
   }
}
```

dracos (approved):

PR #5247, opened 2024-11-08
[Open311] Skip latest data identical updates.

dracos:
If we're told the update is only the latest data, so we can't trust the timestamp, compare it against the last system user comment, and skip if it matches the same text and problem state. Move the assigned user code earlier so that it will be run.

[FMS end corresponding with https://github.com/mysociety/open311-adapter/pull/371 ]

I did wonder about changing the update_id at the open311 end to be a hash of state+text (which it looks like we do do in another backend or two), so it would match once and then not again, but I thought comparing against the last system update received was better if something moved into the same state twice (and perhaps those other backends could be switched to do this as well - unless I've missed something about checking the last update!)

For https://github.com/mysociety/societyworks/issues/3710

### DIFF
--- a/perllib/Open311/UpdatesBase.pm
+++ b/perllib/Open311/UpdatesBase.pm
@@ -164,6 +164,8 @@ sub _process_update {
     my $open311 = $self->current_open311;
     my $body = $self->current_body;
 
+    $self->_handle_assigned_user($request, $p);
+
     my $state = $open311->map_state( $request->{status} );
     my $old_state = $p->state;
     my $external_status_code = $request->{external_status_code} || '';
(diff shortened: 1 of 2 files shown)


perllib/Open311/UpdatesBase.pm
```diff
@@ -177,6 +179,18 @@ sub _process_update {
         $email_text = $request->{email_text};
     };
 
+    if ($request->{latest_data_only}) {
```
davea, line 182: I believe this should mention `extras` otherwise it'll never match?

```suggestion
    if ($request->{extras}{latest_data_only}) {
```

davea (changes requested):
Unless I've misunderstood I don't think this will work as-is. And some tests would be good too.

perllib/Open311/UpdatesBase.pm
dracos: Tests certainly would be good, then, wouldn't they!

davea (approved):
✅

PR #5248, opened 2024-11-11
[Bexley][WW] Container requests fixes part 2

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/35109031/todolists/7983073165.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Request.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Request.pm
@@ -65,11 +65,8 @@ sub validate {
         $any = 1 if $_->name =~ /^container-/ && ($_->value || $self->saved_data->{$_->name});
         # Kingston special case for change-size-first-page
         $any = 1 if $_->name eq 'how_many_exchange' && ($_->value || $self->saved_data->{$_->name});
-        # Bexley special case for household_size first page
-        $any = 1 if $_->name eq 'household_size' && ($_->value || $self->saved_data->{$_->name});
-        # Bexley special case for 'parent' containers (wheelie bins with size options)
-        $any = 1 if $_->name =~ /^parent-/ && ($_->value || $self->saved_data->{$_->name});
     }
+
     $self->add_form_error('Please specify what you need')
         unless $any;
 
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -1303,6 +1303,14 @@ sub construct_bin_request_form {
             ? 'request_removal'
             : ( $above_shop ? 'letterbox_location' : 'request_reason' );
 
+        unshift @$removal_field_list, (
+            no_removal => {
+                type         => 'Checkbox',
+                label        => 'None',
+                option_label => 'I do not need any bins to be removed',
+            }
+        ) if $include_removal;
+
         $page_list = [
             request => {
                 intro => 'container_delivery_intro.html',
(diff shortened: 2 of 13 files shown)


nephila-nacrea:
> Think only thing on this that is new/not already in is the [Add container location option to request form](https://github.com/mysociety/fixmystreet/pull/5248/commits/f97703d93574144ba5f8b0d63b5ce34199645cc9) commit

Yeah, this was more of a WIP PR and I need to do a bit more work to make sure that doesn't clash with the letterbox location stuff for above-shop properties.

perllib/FixMyStreet/App/Form/Waste/Request.pm
```diff
@@ -59,19 +59,20 @@ has_field submit => (
...
 
-    foreach ($self->all_fields) {
-        $any = 1 if $_->name =~ /^container-/ && ($_->value || $self->saved_data->{$_->name});
-        # Kingston special case for change-size-first-page
-        $any = 1 if $_->name eq 'how_many_exchange' && ($_->value || $self->saved_data->{$_->name});
-        # Bexley special case for household_size first page
-        $any = 1 if $_->name eq 'household_size' && ($_->value || $self->saved_data->{$_->name});
-        # Bexley special case for 'parent' containers (wheelie bins with size options)
-        $any = 1 if $_->name =~ /^parent-/ && ($_->value || $self->saved_data->{$_->name});
+    # Bexley has its own validation method for requests
```
nephila-nacrea, line 63: I feel there's a better way to do this, but I didn't know how to skip this validate() method without also skipping the one in Form/Waste.pm.

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -1303,6 +1303,14 @@ sub construct_bin_request_form {
             ? 'request_removal'
             : ( $above_shop ? 'letterbox_location' : 'request_reason' );
 
+        unshift @$removal_field_list, (
+            no_removal => {
+                type         => 'Checkbox',
+                label        => 'None',
+                option_label => 'I do not need any bins to be removed',
+            }
+        ) if $include_removal;
```
dracos, line 1312: Just noting this for myself - this is here rather than in _construct_bin_request_form_removal because you only want it when it's used after the request step, not when they've asked for removal directly.

perllib/FixMyStreet/App/Form/Waste/Request.pm
dracos: In your Bexley specific validate() you could call the 'grandparent' validate function directly rather than using next->method, in order to skip this one, which hardcodes the relationship but we know that it's fixed anyway.

dracos (approved):

perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
nephila-nacrea: Yeah 👍

perllib/FixMyStreet/App/Form/Waste/Request.pm
nephila-nacrea: Would that just be doing something like 'Waste::Form->validate()'? Or is there some kind of syntax using SUPER?

dracos: You would want to call it as `...Form::Waste::validate($self)`. If you called it as Form::Waste->validate() then the class would be passed to it, not the instance, and it wouldn't have the data to look up anything.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -126,10 +126,14 @@ sub send_alert_type {
             # this might throw up the odd false positive but only in cases where the
             # state has changed and there was already update text
             if ($row->{item_problem_state} && $last_problem_state ne $row->{item_problem_state}) {
-                my $cobrand_name = $report->cobrand_name_for_state($cobrand);
-                my $state = FixMyStreet::DB->resultset("State")->display($row->{item_problem_state}, 1, $cobrand_name);
+                my $update = '';
```
nephila-nacrea, line 129: This feels a bit janky to me, I guess it always works OK if Bexley have a block of text for the update. Otherwise, no email is sent.

dracos (approved):

PR #5251, opened 2024-11-11
[Merton] Add larger bin process flow.

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4594
[skip changelog]

### DIFF
--- a/templates/web/merton/waste/_confirmation_after.html
+++ b/templates/web/merton/waste/_confirmation_after.html
@@ -1,0 +1,9 @@
+[% IF first_page == 'medical_condition' %]
+    <p>
+        We will consider your request. We reserve the right to check the information given.
+    </p>
+
+    <p>
+        We will swap your bin for a larger one within 2 weeks, or if your request was unsuccessful we will contact you to let you know.
+    </p>
+[% END %]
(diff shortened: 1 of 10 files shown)


templates/web/merton/waste/_confirmation_after.html
```diff
@@ -0,0 +1,9 @@
+[% IF first_page == 'medical_condition' %]
+    <p>
+        We will consider your request. We reserve the right to check the information given.
+    </p>
+
+    <p>
+        We will swap your bin for a larger one within 2 weeks, or if your request was unsuccessful we will contact you to let you know.
```
MorayMySoc, line 7: I wonder if the above information should be reflected in the email?

MorayMySoc (approved):
This seems to work well for me - not sure if the email should adapt to say same as confirmation page is all that crossed my mind

templates/web/merton/waste/_confirmation_after.html
dracos: not sure how to tell it's one in the email, actually, hmm, used `first_page` here. I guess I could check for the extra data being "2::1" but that seems a bit fragile.

MorayMySoc: It could wait until it's specifically requested. Was just a thought, though the person will be contacted if it's not approved anyway.

PR #5252, opened 2024-11-12
[Merton] Bin not being returned correctly and waste spillage

MorayMySoc:
Part 1 of MRT167 Set up remaining processes in WW

Allow residents to select specified bin types to report a problem that bins haven't been returned correctly to the correct  location or have been left open.

These reports allow a request for the bin collectors to return and rectify the issue along with a notes field

Waste spillage can be reported just with a notes field.

To make live: Check category details on servers with same branch name are live and merge this

https://github.com/mysociety/societyworks/issues/4593

[skip changelog]

### DIFF
--- a/t/app/controller/waste_merton.t
+++ b/t/app/controller/waste_merton.t
@@ -66,6 +66,23 @@ create_contact({ category => 'Request additional collection', email => 'addition
     { code => 'service_id', required => 1, automated => 'hidden_field' },
     { code => 'fixmystreet_id', required => 1, automated => 'hidden_field' },
 );
+create_contact({ category => 'Bin not returned', email => '1570' }, 'Waste',
+    { code => 'Report_Type', required => 1, datatype => 'singlevaluelist',
+    values => [
+        { 'name' => 'Bin Position', 'key' => '1'},
+        { 'name' => 'Lid Not Closed', 'key' => '2'}
+    ] },
+    { code => 'Crew_Required_to_Return?', required => 1, datatype => 'singlevaluelist',
+        values => [
+        { 'name' => 'Yes', 'key' => '1'},
+        { 'name' => 'No', 'key' => '0'}
+    ] },
+    { code => 'Notes', description => 'Details', required => 0, datatype => 'text' },
+);
+create_contact({ category => 'Waste spillage', email => '1545' }, 'Waste',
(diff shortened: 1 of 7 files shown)


templates/web/merton/waste/enquiry-problem.html
```diff
@@ -0,0 +1,16 @@
+[%
+
+f = [];
+service_id = c.req.params.service_id;
+service = services.$service_id;
```
dracos, line 5: These values aren't used; think these come from Bromley version, but e.g. Borsetshire's is a bit simpler, you don't need to construct an `f`, you could only have:
```
[% PROCESS waste/enquiry.html
    title = 'Report a problem'
    field_options = [
        { value = 'Bin not returned', label = 'Bin not returned correctly' },
        { value = 'Waste spillage', label = 'Waste spillage' },
    ]
%]
```

t/app/controller/waste_merton.t
```diff
@@ -515,6 +532,36 @@ FixMyStreet::override_config {
         $e->mock('GetEventsForObject', sub { [] }); # reset
     };
 
+    subtest 'test report a problem' => sub {
```
dracos, line 535: Probably worth a little test for the Waste spillage as well.

templates/web/base/waste/services.html
```diff
@@ -28,6 +28,9 @@
     [% IF is_staff AND NOT unit.orange_bag %]
         <a href="[% c.uri_for_action('waste/report', [ property.id ]) %]?additional=1&amp;service-[% unit.service_id %]=1" class="waste-service-link waste-service-descriptor">Request an additional [% unit.service_name FILTER lower %] collection</a>
     [% END %]
+    [% IF c.cobrand.waste_report_problem(unit.service_id) %]
```
dracos, line 31: I think perhaps for now we just do NOT unit.orange_bag, no need for special sub-check. It's unclear from 'recycling' in their original message if that includes paper and card, but I don't see why it wouldn't, they could go missing also, might as well have everything to begin with.

dracos (changes requested):
Looks good, just a few small things :)

MorayMySoc:
Forgot to say, config is in same branch on servers...

dracos (approved):

PR #5256, opened 2024-11-13
[Surrey] Add text under happy-or-not section

chrismytton:
For FD-4776

image


davea (changes requested):
Looking good, a couple of small things to improve display, but primarily I don't think this makes sense on every page where the HoN widget appears. The front page, for example, where the user hasn't made a report yet.

Is there a way to only include/display this text on the problem confirmation page?

PR #5259, opened 2024-11-18
cpanfile: Use url instead mirror for Geography::NationalGrid

thibaultduponchelle:
## Description
`mirror =>` is failing (for `carton install`, not sure about `carton install --deployment`) with recent `carton`/`cpanm` and I doubt it works with old vendored carton neither.

I propose to replace per `url =>` which is [documented](https://metacpan.org/pod/Carton#Specifying-a-CPAN-distribution) (`mirror` is documented but only with usage of `dist` along).  

What do you think?

Please check the following:

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -89,7 +89,7 @@ requires 'File::Find';
 requires 'File::Path';
 requires 'Geo::OLC';
 requires 'Geography::NationalGrid',
-    mirror => 'https://cpan.metacpan.org/';
+    url => 'https://cpan.metacpan.org/authors/id/P/PK/PKENT/Geography-NationalGrid-1.6.tar.gz';
 requires 'Getopt::Long', '2.52';
 requires 'Getopt::Long::Descriptive', '0.105';
 requires 'HTML::Entities';


dracos:
I've just run our vendored carton on a fresh install with `--deployment`, and without `--deployment` and Geograph::NationalGrid installed in both cases, at least - I think it might use the snapshot information if present regardless?
(Separately, it looks like there's an issue with XML::LibXML needing upgrading).

Happy to change the line if it doesn't work with more recent versions though, of course, as long as it doesn't break our vendored version.

thibaultduponchelle:
Yes it's using the `cpanfile.snapshot` even without `--deployment`

What gives you a clean run `./vendor/bin/carton install` (possibly with `bin/` in PATH for cpanm) after removing `local/` and `cpanfile.snapshot`?

dracos:
So, I tested something wrongly yesterday somehow.
With `mirror`, no `local/`, but with cpanfile.snapshot, our vendored carton installs Geography::NationalGrid okay; with `url` it fails unless you have recent-enough versions perl-wide of IO::Socket::SSL and Net::SSleay, which our GitHub actions do not. I could add them there, and also deal with anyone locally who doesn't have recent enough versions, plus other fallout...
Or I'm just going to vendor in Geography::NationalGrid into our local repository and remove it from the cpanfile entirely, which then solves this issue.

PR #5266, opened 2024-11-21
[Peterborough] Link to report waste service only in on place

MorayMySoc:
Link should only be under individual containers and not in the more_services block

https://mysocietysupport.freshdesk.com/a/tickets/4774

[skip changelog]


chrismytton (approved):
Think the title of this PR should say "Peterborough" instead of "Merton", but otherwise all looks good.

PR #5267, opened 2024-11-21
[Surrey] [Hart] Continue button should not be visible when disabled

MorayMySoc:
Reported that users think the button is broken and don't read the message above for an explanation of why they can't proceed

https://mysocietysupport.freshdesk.com/a/tickets/4768
https://mysocietysupport.freshdesk.com/a/tickets/4818

[skip changelog]


dracos (approved):
I have a weird feeling there's some

dracos (approved):

PR #5268, opened 2024-11-22
[Northumberland] Add display_name

dracos:
and consolidate the templates so don't need to remember to add it [skip changelog]


chrismytton (approved):
Thanks! ✨

PR #5270, opened 2024-11-22
[Surrey] Add External ID to CSV export

chrismytton:
For FD-4770

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Surrey.pm
+++ b/perllib/FixMyStreet/Cobrand/Surrey.pm
@@ -200,6 +200,7 @@ sub dashboard_export_problems_add_columns {
 
     $csv->add_csv_columns(
         alerts_count => "Subscribers",
+        external_id  => "External ID",
     );
 
     my $alerts_lookup = $csv->dbi ? undef : $self->csv_update_alerts;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Surrey.pm
```diff
@@ -200,18 +200,18 @@ sub dashboard_export_problems_add_columns {
...
         my $report = shift;
 
-        if ($alerts_lookup) {
-            return { alerts_count => ($alerts_lookup->{$report->id} || 0) };
-        } else {
-            return { alerts_count => ($report->{alerts_count} || 0) };
-        }
+        return {
+            alerts_count => ($alerts_lookup ? $alerts_lookup->{$report->id} || 0 : $report->{alerts_count} || 0),
+            external_id  => $csv->dbi ? $report->{external_id} : $report->external_id,
```
chrismytton, line 213: Is this right? I notice that Northamptonshire just does `return if $csv->dbi;`, but couldn't work our why 😕

dracos: The Script::CSVExport gets all the columns on a report by default in its hashref, but the non-pre-gen only gets a subset. So you don't need to look up anything that's a column of the DB when $csv->dbi is set, you could if $alerts_lookup is set add external_id there, and if not set return only alerts_count as it does currently.

dracos (changes requested):
Works fine, but yes, from your comment I think you can leave it more like it was before, only adding a line to csv_extra_data in the alerts_lookup being set section.

dracos (approved):
Thanks :)

Issue #5271, opened 2024-11-23
Postgres Error log with new installation


jaykeerti:
**Describe the bug**
I ran a fresh new installation of FMS using the docker deployment method
https://fixmystreet.org/install/docker-deploy/

**To Reproduce**

1. docker ps
```
CONTAINER ID   IMAGE                            COMMAND                  CREATED        STATUS                  PORTS                                     NAMES
c9843d90370a   nginx:1.15.3                     "nginx -g 'daemon of…"   39 hours ago   Up 39 hours             0.0.0.0:8000->80/tcp, [::]:8000->80/tcp   fixmystreet-nginx-1
8ee734fc6e07   fixmystreet/fixmystreet:stable   "/usr/local/bin/boot…"   39 hours ago   Up 11 hours             9000/tcp                                  fixmystreet-fixmystreet-1
a4157fb3cc58   fixmystreet/postgres:9.6-en_GB   "docker-entrypoint.s…"   39 hours ago   Up 39 hours (healthy)   5432/tcp                                  fixmystreet-postgres-1
fd0a725f80ba   memcached:1.5.10                 "docker-entrypoint.s…"   39 hours ago   Up 39 hours             11211/tcp                                 fixmystreet-memcached-1
```
2. docker logs <YOUR OWN CONTAINER ID>
```
STATEMENT:  WITH me AS MATERIALIZED ( SELECT * FROM problem_find_nearby( $1, $2, $3 ) ) (SELECT "me"."problem_id", "me"."distance" FROM "me"  JOIN "problem" "problem" ON "problem"."id" = "me"."problem_id" WHERE ( ( "confirmed" >= current_timestamp-'6 months'::interval AND ( "problem"."state" = $4 OR "problem"."state" = $5 OR "problem"."state" = $6 OR "problem"."state" = $7 OR "problem"."state" = $8 OR "problem"."state" = $9 OR "problem"."state" = $10 OR "problem"."state" = $11 OR "problem"."state" = $12 OR "problem"."state" = $13 OR "problem"."state" = $14 OR "problem"."state" = $15 OR "problem"."state" = $16 OR "problem"."state" = $17 ) AND "problem_id" NOT IN ( $18 ) ) ) ORDER BY "distance", "created" DESC LIMIT $19)
ERROR:  syntax error at or near "MATERIALIZED" at character 12
[...]

PR #5272, opened 2024-11-25
[Docker] Update third party containers.

dracos:
Is switching to an inlne dockerfile like this okay, do you think? Saves us having to maintain/update a postgres container separately, but I didn't know if there was a downside.

Fixes https://github.com/mysociety/fixmystreet/issues/5271

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   nginx:
-    image: nginx:1.15.3
+    image: nginx:1.27.2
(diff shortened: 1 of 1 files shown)


sagepe:
I don't see a reason why not to do this. I just bootstrapped a completely clean installation from this branch successfully, so it looks fine to me.

sagepe (approved):
As noted in a previous comment, I think this is fine.

PR #5273, opened 2024-11-25
[Bexley] Use image with white background

MorayMySoc:
We removed the purple in the email as the transparent logo didn't read well.

Bexley said they had sent a log with a white background so put that in instead.

https://mysocietysupport.freshdesk.com/a/tickets/4769

[skip changelog]


dracos (changes requested):
Looks good, but the file is much larger, could we shrink it to around the same size?

PR #5275, opened 2024-11-26
[Bexley] Add customised map pins

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/4521 https://github.com/mysociety/societyworks/issues/4626

[skip changelog]


dracos (changes requested):
Not sure what package was used to change the colours, but it's made them very jaggedy in doing so - compare e.g. https://github.com/mysociety/fixmystreet/blob/dc1ba7b689138aa5a6efa0755d61fe0eeeda244d/web/cobrands/bexley/images/pin-aqua-big.png vs https://github.com/mysociety/fixmystreet/blob/master/web/i/pin-yellow-big.png
So that needs sorting somehow, I'm afraid

perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -99,6 +99,27 @@ ID.
...
+}
+
+sub pin_colour {
+    my ( $self, $p ) = @_;
+
+    return 'aqua' if $p->state eq 'investigating';
+    return 'orange' if $p->state eq 'action_scheduled';
+    return 'green' if $p->is_fixed;
+    return 'spring' if $p->is_closed;
+    return 'grape' if $p->state eq 'not_council_responsibility';
```
dracos, line 119: This line will never be reached because it'll match on the is_closed line above.

dracos (changes requested):
Thanks, that looks better - one bug in the logic though!

dracos (approved):

PR #5278, opened 2024-11-27
[Bromley] Disable 'Street Cleansing' categories when selection is in a park.

neprune:
[skip changelog] https://github.com/mysociety/societyworks/issues/4635


neprune:
Requirement has changed to disable the categories and display a message when the selected point is in a park, rather than remove them.

PR #5279, opened 2024-11-29
Better logging of page_error

davea:
This adds more logging to page_error, including the HTTP status code, the request IP address, an internal message to show the reason for the error, and a random string to help identify the error in the logs. This string is also shown on the error page to the user, to improve the chance of it being included in any screenshots from users.

image

image

For https://github.com/mysociety/societyworks/issues/4628

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Releases
 
+* Unreleased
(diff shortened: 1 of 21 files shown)


MorayMySoc (approved):
This looks good to me.

PR #5282, opened 2024-12-02
Bexley WW container requests: post go-live amendments

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/4520.

- https://3.basecamp.com/4020879/buckets/35109031/todos/8072413022 - add clinical waste link
- https://3.basecamp.com/4020879/buckets/35109031/todos/8057915745 - text changes for enquiry form (I also made some additional fixes because otherwise Bexley are likely to raise those separately later on)

[skip changelog]

### DIFF
--- a/templates/email/bexley/waste/_email_data.html
+++ b/templates/email/bexley/waste/_email_data.html
@@ -3,7 +3,7 @@ is_missed_collection = report.category == 'Report missed collection';
 is_delivery_request = report.category == 'Request new container';
 is_removal_request = report.category == 'Request container removal';
 is_container_request = is_delivery_request || is_removal_request;
-is_request_enquiry = report.category == 'Request container enquiry';
+is_request_enquiry = report.category == 'Replacement bin enquiry';
 is_no_collections_missed_enquiry
     =  report.category == 'Business or organisation missed collection enquiry'
     || report.category == 'Missed collection enquiry';
--- a/templates/email/bexley/waste/other-reported.html
+++ b/templates/email/bexley/waste/other-reported.html
@@ -17,7 +17,7 @@ INCLUDE '_email_top.html';
   [% ELSIF is_container_request %]
     <h1 style="[% h1_style %]">Thank you for requesting replacement bins</h1>
   [% ELSIF is_no_collections_missed_enquiry OR is_request_enquiry %]
-    <h1 style="[% h1_style %]">Thank you for your [% report.category %]</h1>
+    <h1 style="[% h1_style %]">Thank you for your [% report.category | lower %]</h1>
   [% ELSE %]
     <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
   [% END %]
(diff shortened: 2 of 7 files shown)


dracos (commented):
Looks fine; another option would be to rename the category, presumably, which would be a smaller change? But all the categories do match at present, I guess.

nephila-nacrea:
> Looks fine; another option would be to rename the category, presumably, which would be a smaller change? But all the categories do match at present, I guess.

I wasn't sure I should be changing the category name because I'd thought other cobrands used the same name, but I think I may be mistaken on that. Can have another look and see if I can do that as a simpler change.

templates/email/bexley/waste/other-reported.html
```diff
@@ -16,7 +16,9 @@
     <h1 style="[% h1_style %]">Thank you for reporting a missed collection</h1>
   [% ELSIF is_container_request %]
     <h1 style="[% h1_style %]">Thank you for requesting replacement bins</h1>
-  [% ELSIF is_no_collections_missed_enquiry OR is_request_enquiry %]
+  [% ELSIF is_request_enquiry %]
+    <h1 style="[% h1_style %]">Thank you for your replacement bin enquiry</h1>
```
dracos, line 20: This file change isn't needed any more as the category now works as you'd want.

templates/email/bexley/waste/other-reported.txt
```diff
@@ -1,9 +1,9 @@
 [% PROCESS 'waste/_email_data.html' ~%]
-Subject: [% IF is_missed_collection %]Thank you for reporting a missed collection[% ELSIF is_container_request %]Thank you for requesting a replacement bin[% ELSE %]Thank you for your [% report.category %][% END %]
+Subject: [% IF is_missed_collection %]Thank you for reporting a missed collection[% ELSIF is_container_request %]Thank you for requesting a replacement bin[% ELSIF is_request_enquiry %]Thank you for your replacement bin enquiry[% ELSE %]Thank you for your [% report.category %][% END %]
```
dracos, line 2: Same here for the changes in this file, the ELSE clause now works (if you add a `| lower` to it which seems nice anyway).

templates/email/bexley/waste/_email_data.html
```diff
@@ -3,7 +3,7 @@
 is_delivery_request = report.category == 'Request new container';
 is_removal_request = report.category == 'Request container removal';
 is_container_request = is_delivery_request || is_removal_request;
-is_request_enquiry = report.category == 'Request container enquiry';
+is_request_enquiry = report.category == 'Replacement bin enquiry';
```
dracos, line 6: We'll need to rename all the existing rows in the database when this change goes through for consistency.

dracos (commented):
Looks good, I think the email template changes can be reduced further now

templates/email/bexley/waste/_email_data.html
nephila-nacrea: I thought this would be the simpler change 😭 
But more seriously - I would need help with that, I don't think I've run live DB changes before.
Something like
```
update problem set category = 'Replacement bin enquiry' where category = 'Request container enquiry';
```
?

dracos (approved):
Thanks :) It'll be fine :) The easiest steps to take are probably 1) Create new Replacement bin enquiry category that's identical to the Request container enquiry one. 2) deploy site - can start using new category. 3) mark old category as deleted. 4) update database for old entries. Could rename the category, but this way prevents any edge case of code deployed and category not changed yet (or vice-versa), I guess

nephila-nacrea:
To do for live:
- [] Create 'Replacement bin enquiry' category
- [] Deploy branch
- [] Mark 'Request container enquiry' category as deleted
- [] Update DB for old entries: `update problem set category = 'Replacement bin enquiry' where category = 'Request container enquiry';`

PR #5283, opened 2024-12-02
[Bromley] Disable 'Street Cleansin' categories when selection is in park.

neprune:
[skip changelog]

For https://github.com/mysociety/societyworks/issues/4635.


dracos (approved):
Thanks for adding a JS test :-)
(Perhaps we should improve that road_not_found second function thing at some point, I think it's do with TfL being available as well or similar, not sure how best to deal with it.)

PR #5284, opened 2024-12-03
Improve speed of admin update email search

dracos:
And NH all reports page


MorayMySoc (approved):
Does the database search change need a changelog change or maybe it's too minor

PR #5285, opened 2024-12-03
[Bexley] Fix for assisted collections flag

chrismytton:
This was being set incorrectly as Whitespace returns inaccurate information from its GetSiteContracts endpoint. So instead set it from $property->{has_assisted} as we do elsewhere.

For FD-4777


nephila-nacrea (approved):
Yay!

PR #5289, opened 2024-12-04
[N/land] Pass category updates to/from Alloy

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4648.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -248,6 +248,8 @@ extra_metadata) to Alloy, as an 'extra_details' attribute.
 We pass the name and email address of the user assigned to the report (the
 user who has shortlisted the report).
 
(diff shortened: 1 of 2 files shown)


dracos (commented):
Looks good; perhaps a quick check that group is one of the $contact->group would make sense, though if they don't match, not sure what we would do, so perhaps doesn't really matter!

nephila-nacrea:
> Looks good; perhaps a quick check that group is one of the $contact->group would make sense, though if they don't match, not sure what we would do, so perhaps doesn't really matter!

@dracos So I had actually faffed with this a bit, then realised it might not work or be hard to correct if, on the Alloy end, someone edited the category, forgot to change the group, but later edited the group. It would probably come in as two separate edits our end, but then one or neither would apply because of the group-must-match-category checks.

dracos (approved):

PR #5291, opened 2024-12-05
[Bromley] Any streetlight on a Bromley asset should be 'in Bromley'

MorayMySoc:
Set a hidden field when the asset is selected and the report is made so that we can recognise it was on an streetlight belonging to Bromley regardless of the id

- [x] To deploy, add 'fms_layer_owner' field to "Lamp Column Damaged", "Light blocked by vegetation", "Light Not Working", and "Light On All Day"
- [x] Also make assets/bromley.yml changes for actions, live
- [ ] Merge deploy branch

https://mysocietysupport.freshdesk.com/a/tickets/4815

[skip changelog]

### DIFF
--- a/.cypress/cypress/integration/bromley.js
+++ b/.cypress/cypress/integration/bromley.js
@@ -45,6 +45,24 @@ describe('Bromley cobrand', function() {
     cy.get('.mobile-map-banner').should('be.visible');
   });
 
+  it('sets extra cobrand owner field when streetlight asset selected', function() {
+    cy.pickCategory('Street Lighting and Road Signs');
+    cy.nextPageReporting();
+    cy.pickSubcategory('Street Lighting and Road Signs', 'Lamp Column Damaged');
+    cy.wait('@lights');
+    cy.nextPageReporting();
+    cy.get('.mobile-map-banner').should('be.visible');
+    cy.get('#form_fms_layer_owner').should('have.value', '');
+    cy.visit('http://bromley.localhost:3001/report/new?longitude=0.022775&latitude=51.398387');
+    cy.wait('@report-ajax');
+    cy.get('#mob_ok').click();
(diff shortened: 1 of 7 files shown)


MorayMySoc:
@dracos There is an issue with selecting 'Continue' on the cypress test - think it's me not getting the test right rather than the code, but I can come back to that if it's not obvious to you. But wanted this progressing and visible and to check it's otherwise ok

.cypress/cypress/integration/bromley.js
```diff
@@ -45,6 +45,24 @@ describe('Bromley cobrand', function() {
...
+    cy.pickCategory('Street Lighting and Road Signs');
+    cy.nextPageReporting();
+    cy.pickSubcategory('Street Lighting and Road Signs', 'Lamp Column Damaged');
+    cy.wait('@lights');
+    cy.nextPageReporting();
+    cy.get('.mobile-map-banner').should('be.visible');
+    cy.get('#form_fms_layer_owner').should('have.value', '');
+    cy.visit('http://bromley.localhost:3001/report/new?longitude=0.022775&latitude=51.398387');
+    cy.wait('@report-ajax');
+    cy.contains('Bromley');
```
dracos: You were missing a `cy.get('#mob_ok').click();` here (you can see that's present as the last line of the beforeEach).

.cypress/cypress/integration/bromley.js
```diff
@@ -45,6 +45,24 @@ describe('Bromley cobrand', function() {
...
+    cy.nextPageReporting();
```
dracos: Then this should come after the next line, once you've picked the category, you move to the next page.

dracos (commented):
Seems to work well, was just a missing line in the test as you suspected

dracos (approved):

PR #5292, opened 2024-12-05
[Zurich] Fix setting of body in inactive category.

dracos:
The check for a private category needed to be done in contacts_to_bodies as well as process_report.
[skip changelog] Fixes FD-4869


MorayMySoc (approved):
Looks good to me

PR #5295, opened 2024-12-10
[Gloucester] Added cobrand front-end

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4610

It includes:

- Logos
- Fonts
- Emails
- Colours
- Navigation bar
- favicon
- Site message

### Desktop
https://github.com/user-attachments/assets/a78ac860-9ea9-43b1-be57-be5e59c79f99

https://github.com/user-attachments/assets/a07e75e3-4054-4221-bc8a-50334265c3d5

https://github.com/user-attachments/assets/679e955b-cb33-4aa7-9aa1-0e1491dac8bd

### Mobile

https://github.com/user-attachments/assets/5db11f6a-f82c-424a-a5ca-04eaa6f19fc6

### Emails
Screenshot 2024-12-10 at 10 42 38

Screenshot 2024-12-10 at 10 43 10

[skip changelog]


lucascumsille:
Hi @chrismytton just in case I deleted most of the files that are not yet necessary, but I didn't delete the following:

- `assets,js`
- `tracking_code.html`
- `footer_extra_js.html`

Those ones have the same info as Gloucestershire

chrismytton:
@lucascumsille Thanks! Will have a proper look at this next week and get it onto staging 🙂

chrismytton (approved):

chrismytton:
I've deployed this to staging so now waiting to see if the client has any feedback.

perllib/FixMyStreet/Cobrand/Gloucester.pm
```diff
@@ -0,0 +1,103 @@
...
+
+=cut
+
+sub admin_user_domain { 'gloucester.gov.uk' }
+
+=item * Allows anonymous reporting
+
+=cut
+
+sub allow_anonymous_reports { 'button' }
```
dracos, line 43: We shouldn't have this function, but have all categories ticked in the admin - then the staff can decide which categories are anonymous rather than it being forced on by this function.

templates/web/gloucester/footer_extra_js.html
```diff
@@ -0,0 +1 @@
+[% PROCESS 'footer_extra_js_base.html' highways=0 roadworks=0 validation=0 cobrand_js=0 %]
```
dracos, line 1: Just noting as mentioned in Slack, I think three of these should probably be on?

PR #5298, opened 2024-12-11
Prevent Open311 updates from being sent out of order

nephila-nacrea:
Updates are selected in random order but some cobrands/APIs (e.g. Northumberland/Alloy) do not know how to handle updates in order their end, so an earlier update may overwrite a later one.

This change checks that an update is not sent until earlier updates for that problem are sent.

Fixes https://github.com/mysociety/societyworks/issues/4640.

[skip changelog]

### DIFF
--- a/perllib/Open311/PostServiceRequestUpdates.pm
+++ b/perllib/Open311/PostServiceRequestUpdates.pm
@@ -182,6 +182,30 @@ sub process_update {
         return;
     }
 
+    # Comments are ordered randomly.
+    # Some cobrands/APIs do not handle ordering by age their end (e.g.
+    # Northumberland + Alloy) so we skip comment for now if an older unsent
+    # one exists for the problem. Otherwise an older update may overwrite a
+    # newer one in Alloy etc.
+    my $formatter = FixMyStreet::DB->schema->storage->datetime_parser;
+    my $unsent_comment_for_problem
+        = $problem->comments->search(
+            {
+                state => 'confirmed',
+                send_state => 'unprocessed',
+                confirmed => { '<' =>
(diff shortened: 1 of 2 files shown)


perllib/Open311/PostServiceRequestUpdates.pm
```diff
@@ -182,6 +182,24 @@ sub process_update {
...
     }
 
+    # Comments are ordered randomly.
+    # Some cobrands/APIs do not handle ordering by age their end (e.g.
+    # Northumberland + Alloy) so we skip comment for now if an older unsent
+    # one exists for the problem. Otherwise an older update may overwrite a
+    # newer one in Alloy etc.
+    my @unsent_comments_for_problem
+        = $problem->comments->search(
+            { state => 'confirmed', send_state => 'unprocessed' }
```
dracos, line 192: If you here, include a parameter of confirmed < the $comment->confirmed, then if you get any result you can skip, no need to fetch them all and then check.

dracos (changes requested):
Think you can simplify the check for an earlier unprocessed comment :)

dracos (changes requested):
Looks good, so just one more small further step - you can get rid of the order_by (as you don't care any more), I'd add an `id => { "!=" => $comment->id }` to be on the safe side (not sure if there might be some rounding issues with the date formatting), add `{ rows => 1 }` as a second parameter to search() (because you only care about getting one result) and then add `->single` after the search, putting the result in a scalar, not an array (so you get that one result or undef). :)

dracos (approved):
Great, thanks :)

PR #5302, opened 2024-12-16
[FMS] Make sure /reports link for Kingston work

nephila-nacrea:
Rather than directing back to the dashboard page.

We already had checks for Bexley & Greenwich, whose full names start with 'London|Royal Borough of'; we needed a check for Kingston upon Thames as well.

This is a minimal fix; I feel there's improvements to be made elsewhere or other bugs that may crop up (e.g. should perllib/FixMyStreet/Cobrand/Kingston.pm have its council_name updated to be 'Royal Borough of...'?) But I do not know how much behaviour that would affect down the line.

Fixes https://github.com/mysociety/societyworks/issues/4609.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UK.pm
+++ b/perllib/FixMyStreet/Cobrand/UK.pm
@@ -127,6 +127,12 @@ sub short_name {
     return $name;
 }
 
+sub is_london_or_royal {
+    my ( $self, $short_name ) = @_;
+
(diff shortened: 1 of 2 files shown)


dracos (approved):
Aha, thanks :) I think council_name is only used for some internal displays and FMS related (and Kingston is only waste), so this should hopefully be all that's needed

PR #5303, opened 2024-12-16
[Bromley] Display nudges for Streets categories in parks and vice-versa.

neprune:
[skip changelog]

For https://github.com/mysociety/societyworks/issues/4635.


web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -304,6 +304,41 @@ fixmystreet.assets.bromley.unset_asset_owner = function() {
...
+        'Street Lighting and Road Signs'
+    ];
+    var selected = fixmystreet.reporting.selectedCategory();
+    if (streets_category_groups.includes(selected.group)) {
+        var $msg = $('<div class="box-warning"></div>');
+        $msg.append(
+            "We’ve noticed that you’ve selected a Streets category but that your map pin is located within" +
+            " a park, are the location and category correct? You can change category by going back or" +
+            " relocate your map pin by clicking and dragging.");
+        $msg.prependTo('.js-post-category-messages');
```
dracos, line 322: On the example coords you provide, if you select e.g. Road, Pavement... click Continue and then a subcategory, the message appears twice (or one more time each category change).

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -304,6 +304,41 @@ fixmystreet.assets.bromley.unset_asset_owner = function() {
...
+    } else {
+        $('.js-post-category-messages').empty();
```
dracos, line 324: Conversely, if I click Parks and then dog bin, the message about "For our parks..." picking a dog bin no longer appears, as this is removing it immediately after it's added.

dracos (changes requested):
Good set up, but the message display needs some work - i think you need to have it act more like the general stuff does, give it a unique ID, use `$p.appendTo('.js-reporting-page--active .js-post-category-messages');` like check_zoom_message_visiblity does, and then removal can remove the unique ID and not worry about anything else.

PR #5304, opened 2024-12-18
[Peterborough] Check cutoff time in bulky slots.

dracos:
If we are coming back after failed payment or similar, make sure we're still within the time limit. [skip changelog]


davea (changes requested):
The tests still pass if I comment out the change in `Bulky.pm` - are they testing what you think they are?

dracos:
They are, but only in that I didn't actually add a test for the change :) Only the changes needed to make the tests continue to pass (they should never have passed in the real world, as the time in the test was fixed to mid August but it was showing dates in early August). Have added a new test testing 5th is no longer shown when date set to 7th.

davea (approved):

PR #5306, opened 2024-12-20
[Central Beds] Rename rights of way asset layer

chrismytton:
The "Rights of way" category is now called "Public Rights of way", so update the asset code to match.

See also commit `74b3a1473` in the servers repo.

For FD-4937


MorayMySoc (approved):
Looks good to me

PR #5307, opened 2024-12-23
[Brent][Waste] Update bin pricing text from 1st Jan 25.

neprune:
[skip changelog]

For https://mysocietysupport.freshdesk.com/a/tickets/4939.


neprune:
Screenshot 2024-12-23 at 11 19 51

dracos (approved):
Looks good :) There is one issue (copied from their original) which is that London phone numbers are `020 8... ....` not 0208... so if that could be fixed before merging, thanks :)

PR #5321, opened 2025-01-06
[Waste] Ignore any other bodies returned.

dracos:
The Merton body covers Kingston’s area due to an out-of-area park which is Merton’s responsibility. We do not care about this for waste. Fallout from parks deploy; this caused the cron emails from send-waste this morning as the Merton-only script picked up a Kingston report assigned to both. [skip changelog]


MorayMySoc (approved):
This seems good to me

PR #5322, opened 2025-01-10
[Bexley] GGW bit of early refactoring

dracos:
This moves some bits around to supply some defaults to make it easier to set up a new instance, move stuff out of Echo or Brent specific code so it can be used by others, and similar. [skip changelog]
The branch did originally have a 'first bin different cost' commit in, but I think that needs more work, so starting with these hopefully more straightforward tweaks.

With these changes (and that first bin commit), the basic Bexley GGW subscription flow (apart from no "DD bit cheaper") is this commit: 906db7608 - which is then only intro template, tests, a Bexley specific file to include SCP/Paye and specify the payment bits, and tiny other bits.


templates/web/base/waste/garden/subscribe_details_bin_desc.html
```diff
@@ -0,0 +1,38 @@
...
+      [% ELSIF cobrand.moniker == 'kingston' %]
+        Each garden waste bin holds 240 litres of waste. We collect them every 2 weeks.
+      [% ELSIF cobrand.moniker == 'merton' %]
+        Our garden waste bins are brown. They have a 140L or 240L
+        capacity, which is about the same size as a small or standard wheelie bin.
+      [% ELSIF cobrand.moniker == 'sutton' %]
+        Each garden waste bin holds 240 litres of waste. We collect them every 2 weeks.
+      [% ELSE %]
+        Each garden waste bin holds 240 litres of waste. We collect them every 2 weeks.
+      [% END %]
```
chrismytton, line 34: Could this if statement be tidied up by calling a method in the cobrand modules? Or even pull from general.yml? Same for the banner image above (though possibly trickier with Merton's special case).

(And same question for the subscribe_existing_ template below)

templates/web/base/waste/garden/subscribe_details_bin_desc.html
```diff
@@ -0,0 +1,38 @@
...
+    [% ELSIF cobrand.moniker == 'bromley' %]
+      <img src="/i/waste-containers/bin-black-brown-lid-recycling.png" alt="">
+    [% ELSIF cobrand.moniker == 'kingston' %]
+      <img src="/i/waste-containers/bin-grey-green-lid-recycling.png" alt="">
+    [% ELSIF cobrand.moniker == 'merton' %]
+      [% image = PROCESS "waste/garden/wheelie.svg" %]
+      <span style="--primary-color: #8B5E3D;">[% image.replace('<svg', '<svg style="height:auto;max-width:100%;"') | safe %]</span>
+    [% ELSIF cobrand.moniker == 'sutton' %]
+      <img src="/i/waste-containers/bin-green-brown-lid.png" alt="">
+    [% ELSE %]
```
chrismytton, line 14: Should there be some kind of fallback here, perhaps even an HTML comment explaining that there's no banner image?

chrismytton (approved):
Thanks for this, big step in the right direction 🙂 

Left a couple of questions about the templates, but nothing urgent.

templates/web/base/waste/garden/subscribe_details_bin_desc.html
dracos: I wanted it work okay by default without showing something 'wrong', though I guess it could show any bin image and then it'd hopefully be spotted as part of testing, same as the "???" in the fallback text part, perhaps should just show a black bin as I doubt garden bins are ever that.

templates/web/base/waste/garden/subscribe_details_bin_desc.html
dracos: I guess the only downside of config is that it means you have to set that up in order for the cobrand to work okay, makes it a bit harder to have the equivalent of what's being shown, iyswim - and for something like this is very unlikely to change. I think I prefer output text in the templates rather than the code in general. As this commit was doing, it was in separate per-cobrand template files, which was cleaner (though every one had to repeat the little boilerplate bits) but also harder to see that each one was a little bit customised only for a line or two of text - I thought this way was overall to see that each cobrand has its own tiny custom output.

PR #5324, opened 2025-01-13
[Waste] Factor out garden costs to single package.

dracos:
Moves all garden costs/discount stuff to its own package which can then be queried for costs.
And adds a 'different price for first bin'
[skip changelog]


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1175,8 +1176,10 @@ sub garden_setup : Chained('property') : PathPart('') : CaptureArgs(0) {
 
     $c->detach('property_redirect') if $c->stash->{waste_features}->{garden_disabled};
 
-    $c->stash->{per_new_bin_cost} = $c->cobrand->_get_cost('ggw_new_bin_cost');
-    $c->stash->{per_new_bin_first_cost} = $c->cobrand->_get_cost('ggw_new_bin_first_cost') || $c->stash->{per_new_bin_cost};
+    $c->stash->{garden_costs} = WasteWorks::Costs->new({
+        cobrand => $c->cobrand,
+        service => $c->cobrand->garden_current_subscription,
```
MorayMySoc, line 1181: Not sure if `service` would be tidier being set up inside the package as already available from setting the cobrand - or used directly inside from the cobrand attribute without having an actual `service` attribute.

MorayMySoc (approved):
I've left a couple of incidental comments that are just very minor queries of style, but this seems to work according to the tests. As you say, direct tests on the new package would help confirm its reliability.

perllib/WasteWorks/Costs.pm
```diff
@@ -0,0 +1,165 @@
...
+has renewal_type => ( is => 'lazy', isa => Enum['current', 'subscription_end'] );
+has has_pro_rata_modify => ( is => 'lazy', isa => Bool );
+
+sub _build_renewal_type {
+    my $self = shift;
+    return 'subscription_end' if $self->cobrand->moniker eq 'kingston' || $self->cobrand->moniker eq 'bromley';
+    return 'current';
+}
+
+sub _build_has_pro_rata_modify { $_[0]->cobrand->moniker eq 'bromley' }
```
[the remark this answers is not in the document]
dracos: Yeah, it was to make the one-liners less wordy/boilerplate-y; the use of `shift` was only so I could then pass the remaining arguments to the `_renewal` function (so couldn't use `$_[0]`), happy to get rid of that and leave only the $_[0] ones, will add a comment as well.

perllib/FixMyStreet/App/Controller/Waste.pm
dracos: Yeah, I'm not sure how this will go with future changes for new backend - it might actually be better to pass in the end date directly as that's all that's being used, but as you say, for now it might as well just get it from the cobrand and we can change again in future if need be.

dracos:
Moved service inside Costs.pm; then made some small additional changes - moved _get_costs into the Costs.pm as well, small wording changes, put all the cobrand bits together. Then noticed end date could be taken from the service as well mostly, so moved that out of some calls. And added a tests file. Probably worth a re-review, check my tests look as you'd expect :)

PR #5327, opened 2025-01-15
[Brent] [Waste] General waste bin signposting text changes.

neprune:
[skip changelog]

https://mysocietysupport.freshdesk.com/a/tickets/5028


dracos (changes requested):
For accessibility reasons, could we change the last paragraph they asked us to change - "form" is not good link text (it may be read out independently). I would suggest that the whole paragraph should be a link to the form and the text should be the call to action, eg "Request a new or replacement rubbish bin".
Looks like a corresponding test needs updating as well.

dracos (approved):

neprune:
Changes made so new text goes live from the 27th.

dracos (approved):

PR #5328, opened 2025-01-15
[Brent] [Waste] Add notice for food caddy delays.

neprune:
[skip changelog] https://mysocietysupport.freshdesk.com/a/tickets/5031
food caddy notice


dracos (changes requested):
"caddies" is the plural, and you want "due to stock issues", is all

dracos (approved):

PR #5329, opened 2025-01-15
[Street Manager] Only activate asset layer when starting report

davea:
Previously all map pans on /around were hitting the Street Manager API unnecessarily.

WIP - /report/new is broken at the very least, probably others.

[skip changelog]

For https://github.com/mysociety/societyworks/issues/4728


dracos:
@davea I've added empty asset_category so `relevant()` doesn't error at the start and never passes (so tests should be okay now), moved your code into a roadworks function, added setting always_visible in ther eso it (should?) stay on, and added an onload check if the page is new, to activate it then. Let me know if it works for you.

davea:
Lovely, thanks. Looks good here.

PR #5337, opened 2025-01-21
Add 'open-not-in-progress' updates allowed setting.

neprune:
[skip changelog] For FD-4990.


sallytay:
To note NH don't need this at the moment but may in the future

PR #5339, opened 2025-01-24
[Sutton] Only allow bulky if service unit present.

dracos:
Asked for at https://3.basecamp.com/4020879/buckets/40355775/todos/8248864814#__recording_8252677936
[skip changelog]


MorayMySoc (approved):
Looks good to me

PR #5340, opened 2025-01-27
[Bexley] Add payment fallback script

dracos:
For https://github.com/mysociety/societyworks/issues/4681 [skip changelog]

Doesn't actually include any Bexley specific tests/stuff, that should be added when payments are added in some other work. This is being used by Merton (and can be used by any SCP usage, which Bexley is).


dracos:
I've just pushed an extra commit on this to include apnReference support (as left by PAYE payments).

chrismytton (approved):

neprune:
Servers branch is `bexley-fallback-payment-cron`.

PR #5343, opened 2025-01-27
[Bristol] Add layers for allowed reports

MorayMySoc:
Setup for two asset layers to have either selected to allow reports in Bristol on relevant categories.

https://github.com/mysociety/societyworks/issues/4692

Configuration in same branch on the servers

[skip changelog]


templates/web/bristol/report/new/roads_message.html
```diff
@@ -0,0 +1,8 @@
+<div id="bristol-not-owned" class="hidden js-responsibility-message js-roads-bristol">
```
dracos, line 1: Could this (and other template change) be a `no_asset_message` in the configuration, or is there a reason that didn't work?

dracos:
@MorayMySoc have pushed commit to check staff, and store both flags separately

PR #5346, opened 2025-01-28
[Bexley] [Waste] WIP show existing subscription info

davea:
[skip changelog]

For https://github.com/mysociety/societyworks/issues/4666


perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -30,14 +35,27 @@ sub garden_current_subscription {
...
     return undef unless $uprn;
 
-    my $is_free = $self->agile->IsAddressFree($uprn);
-    return undef if $is_free->{IsFree} eq 'True';
+    my $results = $self->agile->CustomerSearch($uprn);
+    ::Dwarn($results);
+    return undef unless $results && $results->{Customers};
+    my $customer = $results->{Customers}[0];
+    return undef unless $customer && $customer->{ServiceContracts};
+    my $contract = $results->{ServiceContracts}[0];
```
nephila-nacrea, line 43: I think this needs to be $customer->{ServiceContracts}[0].

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -30,14 +35,27 @@ sub garden_current_subscription {
     my $uprn = $self->{c}->stash->{property}{uprn};
     return undef unless $uprn;
 
-    my $is_free = $self->agile->IsAddressFree($uprn);
-    return undef if $is_free->{IsFree} eq 'True';
+    my $results = $self->agile->CustomerSearch($uprn);
```
nephila-nacrea, line 38: I think we mentioned doing a check against the current subscription that is in the DB and making sure the customer email & external_id match the Customer and ServiceContract returned from Agile.
I'm not sure if it would be better to do this match as part of your PR, as strictly speaking it's not necessary to get cancellation working (am also worried about me making too many changes on my PR that then mess with your work here). I think get_original_sub() in Controller/Waste.pm needs updating so it searches on Bexley's `uprn` field as well as `property_id`.

perllib/Integrations/Agile.pm
```diff
@@ -65,4 +65,14 @@ sub IsAddressFree {
     );
 }
 
+sub CustomerSearch {
```
nephila-nacrea, line 68: Thanks for this 🙏  I've copied it over to my cancellation PR.

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -30,14 +35,27 @@ sub garden_current_subscription {
...
+    return undef unless $customer && $customer->{ServiceContracts};
+    my $contract = $results->{ServiceContracts}[0];
+    return unless $contract;
 
     # Agile says there is a subscription; now get service data from
     # Whitespace
     my $services = $self->{c}->stash->{services};
-    map { my $srv = $services->{$_}; return $srv if $srv }
-        @{ $self->garden_service_ids };
+    for my $id (@{ $self->garden_service_ids }) {
```
nephila-nacrea, line 49: Not sure if it's something that needs solving right now, but I am not sure of the best way to handle the Whitespace/Agile split; e.g. if we have Agile data for the contract but not Whitespace. Perhaps how we're doing it now will suffice 🤔 

In my cancellation PR I have extended the `agile_only` bit below to include a couple more fields like `customer_external_ref` and `end_date` (I *think* this can be used as the 'DueDate' required by the cancel call, but I'll need to check that with Luis). I guess I should add those to the Whitespace $srv data here as well.

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -29,19 +31,19 @@ sub garden_service_ids {
...
+    my ($self, $uprn) = @_;
 
-    my $current = $self->{c}->stash->{property}{garden_current_subscription};
-    return $current if $current;
+    my $sub = {
+        row => undef,
 
-    my $uprn = $self->{c}->stash->{property}{uprn};
-    return undef unless $uprn;
+        email => undef,
```
nephila-nacrea, line 40: Do you ever populate email?

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -50,26 +52,61 @@ sub garden_current_subscription {
...
+
+    my $parser = DateTime::Format::Strptime->new( pattern => '%d/%m/%Y %H:%M' );
+    $sub->{end_date} = $parser->parse_datetime( $contract->{EndDate} );
+
+    $sub->{customer_external_ref} = $customer->{CustomerExternalReference};
+
+    $sub->{bins_count} = $contract->{WasteContainerQuantity};
+
+    my $c = $self->{c};
+    my $p = $c->model('DB::Problem')->search({
```
nephila-nacrea, line 69: I think we may need to change `get_original_sub` in Controller/Waste as well (or special-case it so it calls this subroutine for Bexley, to get the 'row' bit).
The current Waste code for other cobrands makes use of both `get_original_sub` and `garden_current_subscription`, looks like the former is for DB lookup and the latter for getting the garden service from the list of services...? I don't know if we should be keeping this distinction here.

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -50,26 +52,61 @@ sub garden_current_subscription {
...
-            $srv->{customer_external_ref}
-                = $customer->{CustomerExternalReference};
-            $srv->{end_date} = $end_date;
+        if ( my $srv = $service_ids->{$_} ) {
+            my $sub = $self->lookup_subscription_for_uprn($uprn);
+            $srv->{customer_external_ref} = $sub->{customer_external_ref};
+            $srv->{end_date} = $sub->{end_date};
+            $srv->{garden_bins} = $sub->{bins_count};
+            $srv->{garden_cost} = $sub->{cost};
+            $self->{c}->stash->{orig_sub} = $sub->{row};
```
nephila-nacrea, line 100: Should this be set in `lookup_subscription_for_uprn`, at the `$sub->{row} = $p if $p;` bit?

nephila-nacrea (changes requested):

nephila-nacrea (changes requested):

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -32,55 +34,127 @@ sub garden_service_ids {
...
+    my $parser = DateTime::Format::Strptime->new( pattern => '%d/%m/%Y %H:%M' );
+    $sub->{end_date} = $parser->parse_datetime( $contract->{EndDate} );
+
+    $sub->{customer_external_ref} = $customer->{CustomerExternalReference};
+
+    $sub->{bins_count} = $contract->{WasteContainerQuantity};
+
+    my $c = $self->{c};
+    my $p = $c->model('DB::Problem')->search({
+        category => 'Garden Subscription',
```
nephila-nacrea, line 87: So looking at this orig_sub stuff again - I don't think we need to be fetching the DB data here (it doesn't look like $sub->{row} is used in garden_current_subscription below, is it used anywhere else?). I think we need to change the existing `get_original_sub` in Controller/Waste.pm, and rely on that to set `$c->stash->{orig_sub}`.

It's worth noting that `get_original_sub` a) does a check for title (so no cancellation reports are fetched) and b) does not do an external_id check. So having a separate lookup for Bexley here that is a bit different is possibly breaking assumptions further down the line 🤔

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1416,11 +1416,17 @@ sub get_current_payment_method : Private {
...
+    my $extra = { '@>' => encode_json({ "_fields" => [ { name => "property_id", value => $c->stash->{property}{id} } ] }) };
+    if ($c->cobrand->moniker eq 'bexley') {
+        # Bexley doesn't store property_id
+        $extra = { '@>' => encode_json({ "_fields" => [ { name => "uprn", value => $c->stash->{property}{uprn} } ] }) };
+    }
+
     my $p = $c->model('DB::Problem')->search({
         category => 'Garden Subscription',
         title => ['Garden Subscription - New', 'Garden Subscription - Renew'],
         # XXX Bexley does not store a property_id
```
nephila-nacrea, line 1428: You can remove this comment.

nephila-nacrea (changes requested):
Just some small cleanup.

PR #5350, opened 2025-01-29
Only update 'lastupdate' when the update is new.

neprune:
https://github.com/mysociety/societyworks/issues/4748

[skip changelog]


perllib/Open311/UpdatesBase.pm
```diff
@@ -281,7 +281,9 @@ sub _process_update {
     # insertion. We also then need a clone, otherwise the setting of lastupdate
     # will *also* reshift comment->created's time zone to TIME_ZONE.
     my $created = $comment->created->set_time_zone(FixMyStreet->local_time_zone);
-    $p->lastupdate($created->clone);
+    if ($created > $p->lastupdate) {
```
dracos, line 284: I think this works even given the time zone statement above, because both things by here should have a time zone for comparison purpose, I think.

dracos (approved):

PR #5353, opened 2025-02-05
[Lincolnshire] Add staff user and role to dashboard CSV export

chrismytton:
Extends the dashboard CSV export to include additional columns for staff users and their roles when a report is contributed by a staff member.

For FD-5038

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Lincolnshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Lincolnshire.pm
@@ -243,4 +243,30 @@ sub open311_get_user {
     return $user;
 }
 
+sub dashboard_export_problems_add_columns {
+    my ($self, $csv) = @_;
+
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
You need to add lincolnshire to `perllib/FixMyStreet/Script/CSVExport.pm` so that the pre-gen CSV includes the staff roles in the output (and staff_user as well I guess if you're including that, though they didn't ask for that?).
https://github.com/mysociety/fixmystreet/commit/6ad241c0d160c7df3ef25db05ddb174f3bbf1742 was last time this was done - you can see there the test also includes going back a day to test the pre-gen version of the CSV as well, might be worth including here?

christofferedwards:
[FD5038](https://mysocietysupport.freshdesk.com/a/tickets/5038) ticket was overdue so just dropped a holding message.

PR #5354, opened 2025-02-05
Improve meaningful sequence in report/around page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4560

Replaces: https://github.com/mysociety/fixmystreet/pull/5208

Unlike the previous [PR#5208](https://github.com/mysociety/fixmystreet/pull/5208) I thought that a better approach to include the icons would be to use HTML templates instead of copy and pasting the SVG. [See commit here](https://github.com/mysociety/fixmystreet/pull/5354/commits/204fb2926dc461b82a685e6a1d65d9a86913c59a)

### Preview 

https://github.com/user-attachments/assets/5fcf7689-dc7a-4b15-8ea2-f32d0b3df265

### Preview Mobile
https://github.com/user-attachments/assets/093b5109-7b1e-4bc0-9260-e215fc804a07

https://github.com/mysociety/fixmystreet/pull/5208#discussion_r1846514980
- [ ] `templates/web/base/report/display_tools_mobile.html` is quite


lucascumsille:
A couple of notes:

>This change shouldn't be made - the js-feed is still in key-tools on any page it is used (not a report page), and needs to moved from there to sub_map_links.
And I'm afraid that sub_map_links uses a background image, so you then end up with two RSS feed icons being shown.
https://github.com/mysociety/fixmystreet/pull/5208#discussion_r1846522253

I solved this issue in this PR. See here:

Screenshot 2025-02-05 at 10 18 22

>Don't think this ID is used?
https://github.com/mysociety/fixmystreet/pull/5208#discussion_r1846505053

Assuming we will use the skip report tools link, then yes, the link uses it the `href`

lucascumsille:
@dracos I made the changes requested in #5208 so I split the commits into 4:

>1.A commit to switch from external to inline SVGs, no other changes (that would also involve e.g. fixing the issue where the Feed link is moved to sub_map_links).
2.A commit to add the close button to the small drawer (and anything else related to that, but nothing more).
3.The skip button on around page if needed.
4.Changes to have a separate key-tools for desktop than mobile (fixing the issue with placement, not working for non-staff, etc)

lizettal:
@dracos I think this is still with you for review. Could you please review this before you go on leave so we can try and get this wrapped up in the coming sprint.

web/cobrands/sass/_base.scss
```diff
@@ -1032,7 +1001,43 @@ html.mobile.js-nav-open #js-menu-open-modal {
...
+    a, button {
+      border-radius:4px;
+      border: 1px solid #ccc;
+    }
+  }
+}
+
+#key-tools {
+  html.js & {
+    display: none;
```
dracos, line 1018: Why are the desktop buttons only hidden if JS is enabled? It means both sets of buttons are shown if JS isn't working - is there an issue with always hiding them, does something go wrong?

templates/web/base/report/display_tools_mobile.html
```diff
@@ -0,0 +1,43 @@
+<div class="shadow-wrap">
+    <ul id="key-tools-mobile" class="js-key-tools">
+      [% IF c.user_exists OR NOT problem.non_public %]
+        [% IF c.cobrand.users_can_hide AND relevant_staff_user %]
+        <li><form method="post" action="/report/[% problem.id %]/delete" id="remove-from-site-form">
+            <input type="hidden" name="token" value="[% csrf_token %]">
+            <button type="submit" id="key-tool-report-abuse" class="has-inline-svg has-inline-svg--column-wrap" data-confirm="[% loc('Are you sure?') %]" name="remove_from_site">
+              <svg role="presentation" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="1.5em" height="1.5em" viewBox="0 0 24 24"><path fill="none" d="M0 0h24v24H0z"/><path fill="currentColor" d="M12.866 3l9.526 16.5a1 1 0 0 1-.866 1.5H2.474a1 1 0 0 1-.866-1.5L11.134 3a1 1 0 0 1 1.732 0zM11 16v2h2v-2h-2zm0-7v5h2V9h-2z"/></svg>
```
dracos, line 8: This template could use the same icons/ templates I think? The contents of the SVGs look the same.

dracos: (Then, if all that's different then is the order of the text and the SVG, can flex order fix that? Then they could share more of the same template for the list items themselves.)

dracos (changes requested):
Looks good, thanks for tidying and making all those changes :) I think the display_tools_mobile template can include the other SVG templates you changed the main template to use, and I had one question about when the buttons are hidden on mobile.
The browser tests fail a lot, I assume due to where things have moved to so that will need looking at, I can take a look at some of them if time later today.

templates/web/base/report/display_tools_mobile.html
```diff
@@ -0,0 +1,43 @@
+<div class="shadow-wrap">
```
dracos, line 1: Sorry, one more thing - this file has some IDs in that are duplicating the ones in display_tools.html. I see you fixed the ones that you were then using in the JS, but the others will need fixing too - e.g. key-tool-report-abuse (not used at all, can be removed), key-tool-problems-nearby (does have some uses), and remove-from-site-form (not needed at all?)

dracos:
I've hopefully patched the commit so that the Cypress tests are passing again (they were getting confused now there were two key tools blocks in the source). So just the small issues above :)

lucascumsille:
@dracos on my latest commits I tackled all the issues above:

> (Then, if all that's different then is the order of the text and the SVG, can flex order fix that? Then they could share more of the same template for the list items themselves.)
 
I ended up getting rid of the `display_tool_mobile.html` considering that is pretty similar to `display_tool`. `flex-direction` did the trick =)

The above would also fix:

>  IDs in that are duplicating
> This template could use the same icons/ templates
> desktop buttons only hidden if JS is enabled

Let me know if you are happy with those changes =)

dracos (approved):
Oh great, that's much nicer/simpler, isn't it! :)
I've rebased and added a line to the changelog, and tidied up the cypress tests (as they don't need changing now there's only one key tools again).

dracos:
Sorry, not sure how I missed this, or if it got introduced later on this has completely changed the layout of the update on a report page. The update form is now appearing above all the updates, rather than at the bottom as it used to be? I don't think this was mentioned or a desired change of this PR, which is now much more confusing in terms of getting a form for an update before you've seen the previous updates. It's changed the reading order, rather than made the source order match the visual order.

PR #5358, opened 2025-02-07
Fix HTML in site messages

chrismytton:
If we detect that a site message contains HTML then don't use the html_para filter to avoid adding extra <br> elements which mess up spacing.

For FD-5103

## Before

image

## After

image

### DIFF
--- a/templates/web/base/front/site-message.html
+++ b/templates/web/base/front/site-message.html
@@ -1,7 +1,7 @@
 [% site_message = c.cobrand.site_message %]
 [% IF site_message && !rendered_site_message %]
 <div class="site-message">
-  [% site_message | html_para %]
+  [% site_message | html_para_conditional %]
 </div>
 [% END %]
 [% rendered_site_message = 1 %]
--- a/perllib/FixMyStreet/Template.pm
+++ b/perllib/FixMyStreet/Template.pm
@@ -156,6 +156,21 @@ sub html_paragraph_email_factory : FilterFactory('html_para_email') {
     }
 }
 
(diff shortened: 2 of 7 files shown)


perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -1432,6 +1432,7 @@ sub site_message {
         my $ooh = $self->ooh_times($body);
         $msg = $ooh_msg if $ooh->active;
     }
+    $msg =~ s/\r\n/\n/g;
```
dracos, line 1435: What's this bit for? To make the test simpler?

templates/web/base/front/site-message.html
```diff
@@ -1,7 +1,11 @@
 [% site_message = c.cobrand.site_message %]
 [% IF site_message && !rendered_site_message %]
 <div class="site-message">
+[% IF site_message.match('<[a-z][\s\S]*>', 'i') %]
```
dracos, line 4: There are some other places this would be good - `report/new/form_after_heading.html` for the reporting message, base/peterborough `waste/header.html` for waste, and `templates/web/bathnes/around/_postcode_form_post.html`. Perhaps a new Filter in `Template.pm` that does this check and then calls html_paragraph or not, so it's hidden from the templates, could just have `site_message | html_para_if_no_html` or something like that?

dracos (changes requested):
Works well, only suggesting might be good to hide the logic within a filter, and then it can be used everywhere a site message could be displayed?

PR #5360, opened 2025-02-07
[Bristol][Alloy] Adds putting usrn into Alloy reports

MorayMySoc:
Where an Alloy report is sending to Bristol over Open311 it must have a usrn present to identify the street it is on so the report can be parented to the correct road and given the correct locality.

https://github.com/mysociety/societyworks/issues/4743

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bristol.pm
+++ b/perllib/FixMyStreet/Cobrand/Bristol.pm
@@ -177,6 +177,40 @@ sub open311_config_updates {
     $params->{multi_photos} = 1;
 }
 
+=head2 open311_update_missing_data
+
+All reports sent to Alloy should have a USRN set so the street parent
+can be found and the locality can be looked up as well.
+
+The USRN may be set by the roads asset layer, but staff can report anywhere
+so are not restricted to the road layer and anyone can be make reports on
+specific Bristol owned properties that don't have a USRN
+
+=cut
+
+sub lookup_site_code_config {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Bristol.pm
```diff
@@ -177,6 +177,40 @@ sub open311_config_updates {
...
+            url => "https://$host/proxy/bristol/wfs/",
+            typename => "COD_LSG",
+            property => "USRN",
+            srsname => "urn:ogc:def:crs:EPSG::27700",
+            outputformat => "GML3",
+            accept_feature => sub { 1 },
+        };
+        my ($x, $y) = $row->local_coords;
+        my $features = $self->_fetch_features($cfg, $x, $y, 1) || [];
+        my $usrn = $features->[0]
```
dracos, line 207: You don't know the order results are returned, and you're asking for everything in a 400m-squared box, I'm afraid. This should call `_nearest_feature` I think, which does that for you? e.g. `my $usrn = _nearest_feature($cfg, $x, $y, $features);` - oh, is the issue that's JSON output only? Hmm, well, it needs fixing in some way.
Ooh, geojson outputformat works with version = 2.0.0, so using that is probably quickest/easiest way to simplify this.

dracos (changes requested):
One issue with the ordering of results is all, I think

PR #5363, opened 2025-02-10
[Hounslow] Rename body to "Hounslow Highways"

chrismytton:
Rename the **Hounslow Borough Council** body to **Hounslow Highways**, seeing as that's what it's publicly called.

This should also fix an issue where a report could go to multiple bodies but it wasn't previously listing all those bodies to the user.

For FD-5117

## Note to deployer

The following need to be done when deploying this change

- [ ] Update body name in the admin from `Hounslow Borough Council` to `Hounslow Highways`
  - [x] On staging
  - [ ] On live
- [ ] In the servers repo change `Hounslow Borough Council` to `Hounslow Highways` in `vhosts/fixmystreet/assets/hounslow.yml.include` and redeploy (will need to be done with a conditional on staging first)
  - [x] On staging
  - [ ] On live

### DIFF
--- a/.cypress/cypress/integration/hounslow.js
+++ b/.cypress/cypress/integration/hounslow.js
@@ -27,9 +27,9 @@ describe('private categories', function() {
     cy.get('#map_box').click(290, 307);
     cy.wait('@report-ajax');
     cy.pickCategory('Potholes');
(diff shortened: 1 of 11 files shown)


dracos (approved):
Good to get rid of bit of old special code!

PR #5369, opened 2025-02-13
[Bexley] Access paysuite integration

chrismytton:
Work in progress - add form for gathering user's direct debit details and pass off to Access PaySuite.

For https://github.com/mysociety/societyworks/issues/4612

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -401,6 +401,17 @@ sub direct_debit : Path('dd') : Args(0) {
     $c->detach;
 }
 
+sub direct_debit_internal : Private {
+    my ($self, $c) = @_;
+
+    $c->forward('populate_dd_details');
+    $c->cobrand->call_hook('waste_setup_direct_debit');
+    $c->stash->{title} = "Direct Debit mandate";
+    $c->stash->{message} = "Your Direct Debit has been set up successfully.";
+    $c->stash->{template} = 'waste/dd_complete.html';
+    $c->detach;
+}
+
 # we process direct debit payments when they happen so this page
 # is only for setting expectations.
 sub direct_debit_complete : Path('dd_complete') : Args(0) {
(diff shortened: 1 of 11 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -390,6 +391,34 @@ sub populate_dd_details : Private {
...
+                    $form->add_form_error('Failed to setup direct debit');
+                }
+            }
+        }
+
+        $c->stash->{form} = $form;
+    }
+
+    $c->stash->{template} = 'waste/dd.html';
+    $c->detach;
```
chrismytton, line 421: I'm trying to display an extra form to gather the users details instead of redirecting them off to a portal like we do for other places, but it's not working as it should.

I think when the POST request is made for the Direct Debit details form it's picked up as a re-submission of the garden form, and so goes back to the page with the terms and conditions checkbox again, instead of submitting the Direct Debit.

I tried to alter the existing garden form to add a new page for the direct debit, but I couldn't get that working, I think in part because there are a lot of assumptions that the summary page will be the final page.

I must be missing a much easier way to do this?

perllib/FixMyStreet/App/Form/Waste/BankDetails.pm
```diff
@@ -0,0 +1,156 @@
...
+# Create a dedicated page for entering bank details.
+has_page bank_details => (
+    title => 'Enter Your Bank Details',
+    template => 'waste/bank_details.html',
+    fields => ['account_holder', 'account_number', 'sort_code', 'reference', 'amount', 'report_id', 'submit'],
+    next => sub {
+        my $form = shift;
+        return $form->wizard_finished('process_bank_details');
+    },
+);
```
chrismytton, line 148: Not sure if this is even needed?

chrismytton (commented):
Really struggling with this change. Kept thinking if I could just get a day of focus I'd make a breakthrough and then I could make some progress, but I just keep hitting a wall with my understanding of the WW form wizard and how to add the extra form needed for these details.

Issue #5348, opened 2025-01-28
Email Accessibility - Contrast too low


edent:
**Describe the bug**
Current email colours fail WCAG guidance on text contrast.

**To Reproduce**
1. Receive an email with an update.
2. Try to read the text at the top and bottom (e.g. " This email was sent automatically, from an unmonitored email account. Please do not reply to it."
4. Hurt your eyes

**Expected behavior**
The current text colour is #4E646B; the background is #42494C. This is an accessibility failure - https://www.siegemedia.com/contrast-ratio#%234E646B-on-%2342494C

**Screenshots**
![Image](https://github.com/user-attachments/assets/1d4e4640-36d0-4e3b-abac-45c4e15419ac)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 134.0.2 (64-bit)

**Additional context**

The problem occurs at

https://github.com/mysociety/fixmystreet/blob/ed9a79956a076f80acbac7c33ccf4b2e00b1df8e/templates/email/default/_email_settings.html#L23

Changing it to `color_blue_pale` will fix the contrast issue.

dracos:
Thanks; perhaps we should actually drop the line entirely, it's only repeating the heading just below. I wonder if there's circumstances where it's saying something more useful, we can check.

PR #5372, opened 2025-02-17
Accessibility improvements to hint_style for emails

lucascumsille:
Fixes: https://github.com/mysociety/fixmystreet/issues/5348

Currently, the email's top(email_summary) and bottom(email_footer) parts don't necessarily pass the contrast test. The text colour uses body_text_color, and the background colour for the top and bottom sections is not set, so the hint text is above whatever the body_background_color is. Looking at some cobrands, I can see that the problem is not only with FMS. The solution on this commit gives hint_style a background that uses color_white as a variable, and because the text is using the body_text_color, it is less likely the contrast test will fail. The scenario where this could happen is when body_text_color is using a light colour, which, so far, I haven't seen a cobrand doing this.

https://github.com/user-attachments/assets/565bfa12-72b8-49cd-ab54-fe040785e8a8

### Important:

- [ ] If the solution above is accepted then we need to update the `email_top` and `email_bottom` templates for: Hackney, Bathnes and Fixamingata
- [ ] Regarding getting rid of `email_summary` it does seem like it could be done. However looking at the `email_summary` content across cobrands it seems like it doesn't mimic the main title of the email 100% of the time. Reason why I didn't remove it.

I added an extra commit that corrects the extra space use by "This email was sent from a staging site."
#### Before
Screenshot 2025-02-17 at 14 30 46

#### After
Screenshot 2025-02-17 at 14 30 05

[skip changelog]

### DIFF
--- a/templates/email/bathnes/_email_top.html
+++ b/templates/email/bathnes/_email_top.html
@@ -66,11 +66,7 @@
 [% END %]
   <table [% wrapper_table | safe %] style="[% wrapper_style %]">
     <tr>
-      <th class="spacer-cell"></th>
-      <th width="[% wrapper_max_width %]" style="[% td_style %][% hint_style %]" class="hint">
-        [% email_summary %]
-      </th>
-      <th class="spacer-cell"></th>
+      <td height="24">&nbsp;</td>
     </tr>
   </table>
   <table [% wrapper_table | safe %] style="[% wrapper_style %]">
--- a/templates/email/bathnes/archive.html
+++ b/templates/email/bathnes/archive.html
@@ -1,7 +1,5 @@
 [%
 
-email_summary = "Your reports on " _ site_name;
-
 PROCESS '_email_settings.html';
 
 INCLUDE '_email_top.html';
--- a/templates/email/bexley/alert-update.html
+++ b/templates/email/bexley/alert-update.html
@@ -3,12 +3,6 @@
 PROCESS 'waste/_email_data.html';
 
 title = report.title | html;
-email_summary
-  = is_missed_collection
-  ? "New updates on your missed collection report"
-  : is_container_request
-  ? 'New updates on your replacement bin request'
-  : "New updates on “" _ title _ "”";
 email_columns = 2;
 
 PROCESS '_email_settings.html';
--- a/templates/email/bexley/waste/other-reported.html
+++ b/templates/email/bexley/waste/other-reported.html
@@ -1,6 +1,5 @@
 [%
 
-email_summary = "Your reference number is " _ report.id;
 email_columns = 2;
 
 PROCESS 'waste/_email_data.html';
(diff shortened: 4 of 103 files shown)


dracos:
> I added an extra commit that corrects the extra space use by "This email was sent from a staging site."

Aww, I thought that was deliberate to have it as a banner pinned on top of the email :) (Don't mind it going, though, if you prefer it this way.)

I am tempted to say we try and get rid of email_summary - happy to look at any cases where they don't match up with the title, but in general it seems quite repetitive, and making the contrast work makes it much too prominent to me, like, why is there this message above the logo was my first thought :)

lucascumsille:
> > I added an extra commit that corrects the extra space use by "This email was sent from a staging site."
> 
> Aww, I thought that was deliberate to have it as a banner pinned on top of the email :) (Don't mind it going, though, if you prefer it this way.)
> 
> I am tempted to say we try and get rid of email_summary - happy to look at any cases where they don't match up with the title, but in general it seems quite repetitive, and making the contrast work makes it much too prominent to me, like, why is there this message above the logo was my first thought :)

If we can remove that would be great, because it is repetitive, but I didn't remove it because there are some case, I'm thinking about Bucks, we might not want it remove. We could also give it the same treatment as `_email_bottom` and only display if the template exist.

dracos:
> If we can remove that would be great, because it is repetitive, but I didn't remove it because there are some case, I'm thinking about Bucks, we might not want it remove.

Which one are you thinking about in particular? alert-update is the same, and the others all seem to match the default templates so I don't think were changed specifically for Bucks or anything.

lucascumsille:
> > If we can remove that would be great, because it is repetitive, but I didn't remove it because there are some case, I'm thinking about Bucks, we might not want it remove.
> 
> Which one are you thinking about in particular? alert-update is the same, and the others all seem to match the default templates so I don't think were changed specifically for Bucks or anything.

I'm thinking about the `alert_update` one where:
`email_summary = "New updates on " _ category _ " report";` which is different from the default version which is `email_summary = "New updates on “" _ title _ "”";` not sure if we made this change because Bucks requested, but if that is the case I wonder if we then should add the `_ category _ ` bit somewhere else if we are going to delete the top banner.

dracos:
The bucks template *does* also put the different one in its h1 :)

lucascumsille:
@dracos I missed that one =) I think that was the only case unless you can think of any other case where `email_summary` might be relevant then happy to drop it.

lucascumsille:
@dracos I added 4 fixups:

- [Welsh](https://github.com/mysociety/fixmystreet/pull/5372/commits/9ca0f3c5950d46e167d80978e45c112a06620a78) Removed the `email_summary` from the welsh emails.
- [Fixamingata](https://github.com/mysociety/fixmystreet/pull/5372/commits/15007cc845b8b8aa3e650e2b98fbe24715c4802c) Removed `email_summary` from Fixamingata.  
- [This commit](https://github.com/mysociety/fixmystreet/pull/5372/commits/f2b875f7303f7536d37c4911d51a255ed9c4e7ca) removed `email_summary` everywhere else.

The only places where `email_summary` is still present is in WasteWorks

PR #5380, opened 2025-02-24
[Sutton] Report problem functionality

dracos:
[skip changelog] https://github.com/mysociety/societyworks/issues/4727

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1100,6 +1100,20 @@ sub enquiry : Chained('property') : Args(0) {
             next;
         }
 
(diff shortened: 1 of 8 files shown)


MorayMySoc (approved):
Looks good

dracos:
Photo upload wasn't working, and the test didn't check the photo had been saved. Issues were:
* form didn't have multipart/form-data - I've added a `has_file_upload` helper to the Page class which the template can check, so hopefully will work for any future things like that
* Had to not show 'fileid' in summary enquiry template, as otherwise photo now shown twice (now that it was working)
* The photo isn't attached to `photo` on the report without a change to the `add_report` - but I decided to call the enquiry photo `location_photo` (as it's called in bulky) as then it's already caught by the code for that.

PR #5383, opened 2025-02-24
Improvements to extra data in submit email

dracos:
Removes some duplicate output from NH emails, uses better output for extra questions for Bristol flytipping (and anywhere else extra data is included in emails).
[skip changelog]


MorayMySoc (approved):
This looks good to me

PR #5385, opened 2025-02-25
[Bexley][WW] Validate DD bank details before submitting them

davea:
The Access PaySuite DD API does not validate bank details when creating new contracts, so we first pass them to a totally separate Access PaySuite 'bankchecker' API to make sure they're good.

For https://github.com/mysociety/societyworks/issues/4823

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden/Bexley.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden/Bexley.pm
@@ -1,6 +1,12 @@
 package FixMyStreet::App::Form::Waste::Garden::Bexley;
 
 use utf8;
+use LWP::UserAgent;
+use JSON::MaybeXS;
+use URI;
+use Try::Tiny;
+use Data::Dumper;
+
 use HTML::FormHandler::Moose;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Form/Waste/Garden/Bexley.pm
```diff
@@ -14,4 +20,102 @@ has_page bank_details => (
...
+sub _validate_bank_details {
+    my ($self, $sort_code, $account_number ) = @_;
+
+    # don't bother calling the API if we don't have both values
+    return 1 unless $sort_code->value && $account_number->value;
+
+    my $cfg = $self->{c}->cobrand->feature('payment_gateway');
+
+    # fail validation if not configured - we don't want to set up invalid DDs.
+    return 1 unless $cfg && $cfg->{validator_url};
```
chrismytton, line 55: This didn't seem to work in my testing, it was allowing any values through. Do you need to call `add_form_error` if the URL is missing, perhaps?

chrismytton (approved):
Just one minor observation about the missing config case, but otherwise seems to work well!

PR #5386, opened 2025-02-25
[Zurich][FMS] Add Script for anonymising reports by date

MorayMySoc:
Add a script to anonymise reports that are before n years ago.

Includes testing in Zurich as they have requested the feature and a script for calling the anonymisation script for use in a cron job.

https://mysocietysupport.freshdesk.com/a/tickets/5143

[skip changelog]

### DIFF
--- a/t/cobrand/zurich.t
+++ b/t/cobrand/zurich.t
@@ -1,6 +1,7 @@
 # TODO
 # Overdue alerts
 
+use Test::Output;
 use DateTime;
 use Email::MIME;
 use File::Temp;
@@ -11,6 +12,7 @@ use Path::Tiny;
 use t::Mock::MapItZurich;
 use FixMyStreet::TestMech;
 use FixMyStreet::Script::Reports;
+use FixMyStreet::Script::Inactive;
 my $mech = FixMyStreet::TestMech->new;
 
 FixMyStreet::App->log->disable('info');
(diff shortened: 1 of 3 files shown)


dracos (changes requested):
So I *think* this script is mostly covered by the existing process-inactive-reports script - the only change that looks to be needed (I may have missed something) is adding support for the date range to be based on 'created' rather than 'lastupdate' - but I think perhaps leaving it on lastupdate (confirming that's okay with Zurich) would be fine also?
Good to keep the extra tests and see if they still work with the existing script, if that's the case, certainly.

bin/fixmystreet.com/anonymise_reports_pre_date
```diff
@@ -0,0 +1,34 @@
...
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use FixMyStreet::Script::AnonymiseReportsPreDate;
+use Getopt::Long::Descriptive;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['body|b=i',     "int id of body to operate on (required)", { required => 1 } ],
```
dracos, line 24: Also Zurich has many bodies, so we don't want to operate on only one at a time.

t/cobrand/zurich.t
```diff
@@ -1179,6 +1182,76 @@ subtest 'CSV export includes lastupdate for problem' => sub {
...
+        ok($updated->user->id != $anon->id, "Report over two years old not anonymised without commit");
+    };
+
+    $opts{commit} = 1;
+    stdout_is { FixMyStreet::Script::AnonymiseReportsPreDate::anonymise(\%opts); }
+    "2 reports have been anonymised/n", "Report count of anonymised reports on commit";
+
+    for my $report (@old_reports) {
+        my ($updated) = FixMyStreet::DB->resultset("Problem")->search( { id => $report->id } );
+        ok($updated->user->id == $anon->id, "Report over two years old assigned to anonymous user");
```
dracos, line 1247: Please use is/isnt rather than ok with an ==/!=, it works better.

t/cobrand/zurich.t
```diff
@@ -1179,6 +1182,76 @@ subtest 'CSV export includes lastupdate for problem' => sub {
...
+        my ($updated) = FixMyStreet::DB->resultset("Problem")->search( { id => $report->id } );
+        ok($updated->user->id != $anon->id, "Report over two years old not anonymised without commit");
+    };
+
+    $opts{commit} = 1;
+    stdout_is { FixMyStreet::Script::AnonymiseReportsPreDate::anonymise(\%opts); }
+    "2 reports have been anonymised/n", "Report count of anonymised reports on commit";
+
+    for my $report (@old_reports) {
+        my ($updated) = FixMyStreet::DB->resultset("Problem")->search( { id => $report->id } );
```
dracos, line 1246: can use discard_changes here

t/cobrand/zurich.t
```diff
@@ -1179,6 +1181,67 @@ subtest 'CSV export includes lastupdate for problem' => sub {
...
+        state => 'all',
+        verbose => 1,
+        'dry-run' => 1,
+    );
+
+    my $user_id = $old_reports[0]->user_id;
+    my ($report1, $report2) = ($old_reports[0]->id, $old_reports[1]->id);
+
+    stdout_is { FixMyStreet::Script::Inactive->new(%opts)->reports; }
+    "DRY RUN\nAnonymizing problem #$report1\nAnonymizing problem #$report2\n", "Non verbose count of reports to be anonymised";
```
MorayMySoc, line 1224: Only this is 'Verbose count' now, I think

t/cobrand/zurich.t
```diff
@@ -1179,6 +1181,67 @@ subtest 'CSV export includes lastupdate for problem' => sub {
...
+
+    for my $report (@old_reports) {
+        $report->discard_changes;
+        is $report->user->id, $user_id, "Report over two years old not anonymised without commit";
+    };
+
+    $opts{'dry-run'} = 0;
+    stdout_is { FixMyStreet::Script::Inactive->new(%opts)->reports; }
+    "Anonymizing problem #$report1\nAnonymizing problem #$report2\n", "Non verbose count of reports to be anonymised";
+
```
MorayMySoc, line 1234: Oh, and this reports anonymised, not 'to be anon..'

PR #5387, opened 2025-02-26
Fix for value labels in email where value is 0.

dracos:
The IF check in the template was skipping fields with a value of "0". [skip changelog]


MorayMySoc (approved):
Test runs locally so if the tests don't throw anything else up, the changes look good

PR #5392, opened 2025-03-04
Limit the max-width for sign-in form on mobile

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4078

# Preview:
https://github.com/user-attachments/assets/c7d71368-7133-4bdb-869d-4ab98f76764b

### DIFF
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -462,6 +462,7 @@ a.platform-logo {
 body.authpage {
   .content {
     width: 27em;
+    max-width: 27em;
     margin-#{$left}: auto;
     margin-#{$right}: auto;
   }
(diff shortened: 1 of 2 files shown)


web/cobrands/sass/_layout.scss
```diff
@@ -462,6 +462,7 @@ a.platform-logo {
 body.authpage {
   .content {
     width: 27em;
+    max-width: 27em;
```
dracos, line 465: Can we drop the width line then; don't think it needs both?

dracos (changes requested):
One query about the width is all :)

web/cobrands/sass/_layout.scss
lucascumsille: @dracos I just did =), let me know if you are happy for me to merge

PR #5393, opened 2025-03-04
[SLWP] Deal with multiple slot lists.

dracos:
[skip changelog] For https://github.com/mysociety/societyworks/issues/4828 (everything except getting the two slots into Echo at the open311 end) - nothing within front end etc should change, this is about sending items through in order to get two slot lists, dealing with two slot lists in the hidden reservation IDs.


MorayMySoc (approved):
Looks good to me

PR #5395, opened 2025-03-05
[Bexley][WW] Add direct debit cancellation support

chrismytton:
When a user cancels their garden waste subscription, call Access PaySuite's "archive contract" method to cancel the Direct Debit.

For https://3.basecamp.com/4020879/buckets/40373795/todos/8374072726


dracos (approved):
Assuming the API call is correct, looks good :)

PR #5402, opened 2025-03-07
[Bexley] For pending reports, look for confirmed.

dracos:
Just had this thought this morning; DD reports are confirmed immediately, so we don't need to check for unconfirmed reports, which might confuse with unpaid credit card reports. [skip changelog]

Wasn't entirely sure if should only do this for cancelled, the known issue, not new/renew, but I thought this was slightly better in case there's a problem sending to Agile so we don't have the RENEWALDUE from there.


chrismytton (approved):
I think this should be fine for new and renew, as well as cancelled. 👍

PR #5407, opened 2025-03-07
[FMS][BANES] Create Script to send all Confirm reports to Passthrough endpoint

MorayMySoc:
After reports or comments have been sent to Confirm through the standard send-reports script, we want to then send them on to the Passthrough endpoint.

Reports and comments should keep their Confirm identifications as they will continue to receive updates from Confirm.

https://github.com/mysociety/societyworks/issues/4831

[skip changelog]

### DIFF
--- a/bin/banes/send_confirm_to_passthrough
+++ b/bin/banes/send_confirm_to_passthrough
@@ -1,0 +1,22 @@
+#!/usr/bin/env perl
+#
+# This script sends waste reports additionally to BANES's
+# Open311 endpoint. They have already been sent to Confirm.
+
+use v5.14;
+use warnings;
+
+BEGIN {    # set all the paths to the perl code
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use FixMyStreet::Script::BANES::PassthroughConfirm;
+
+my $send = FixMyStreet::Script::BANES::PassthroughConfirm->new;
+
+
+$send->send_reports;
+$send->send_comments;
--- a/perllib/FixMyStreet/Script/BANES/PassthroughConfirm.pm
+++ b/perllib/FixMyStreet/Script/BANES/PassthroughConfirm.pm
@@ -1,0 +1,114 @@
+package FixMyStreet::Script::BANES::PassthroughConfirm;
+
+use Moo;
+use FixMyStreet::DB;
+use FixMyStreet::Queue::Item::Report;
+use FixMyStreet::SendReport::Open311;
+use Open311;
+
+has body => (
+    is => 'ro',
+    default => sub { FixMyStreet::DB->resultset('Body')->find( { name => 'Bath and North East Somerset Council' } ) or die $! }
+);
+
+sub send_reports {
+    my ($self) = @_;
+
+    my $problems = $self->_problems;
+
+    while (my $row = $problems->next) {
+        my $item = FixMyStreet::Queue::Item::Report->new( report => $row );
+        FixMyStreet::DB->schema->cobrand($item->cobrand);
+        my $confirm_id = $row->external_id;
+        $item->cobrand->set_lang_and_domain($row->lang, 1);
+        $item->_create_vars;
+        $item->h->{sending_to_banes_passthrough} = 1;
+        my $sender_info = {
+            method => 'Open311',
+            config => $self->body,
+        };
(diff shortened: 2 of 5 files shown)


bin/banes/send_confirm_to_passthrough
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env perl
+#
+# This script sends waste reports additionally to Merton's
```
dracos, line 3: BANES's Confirm :)

perllib/FixMyStreet/SendReport/Open311.pm
```diff
@@ -77,6 +77,7 @@ sub send {
     my $open311 = Open311->new( %open311_params );
 
     my $skip = $cobrand->call_hook(open311_pre_send => $row, $open311);
+
```
dracos, line 80: Don't need this newline adding

dracos (approved):
Two meta things is all, thanks :)

MorayMySoc:
Fix up to add in code for sending the service_code with service request updates and allowing setting a service_code when initialising Open311 and honouring that so that we can assign the required generated email address to the service_code on updates.

t/script/banes/sendtopassthrough.t
```diff
@@ -51,7 +51,7 @@ FixMyStreet::override_config {
     $c = CGI::Simple->new($req->content);
 
     is $c->{service_request_id}[0] eq '248', 1, "Passthrough service_request_id is set as the report's passthrough id";
-
+    is $c->{service_code}[0] eq 'passthrough-confirm_graffiti@example.org', 1, "service_code given email address to send to passthrough";
```
dracos, line 54: Again, please can we have `is A, B` rather than `is A eq B, 1` because that produces nicer output. Wherever this happens

dracos (changes requested):
Couple of small comments.
Feels a bit awkward, but can't think of anything much better. Only potential thought was to not change Open311Multi, not pass through $self->service_code to the hook, but just if $self->service_code is present, set $params->{service_code} directly around where the hook is called instead (given that's basically what you're saying setting that means) - a tiny bit simpler?

MorayMySoc:
70b8514cbf2e651a6c41a3597f6275c4f2a20183 is restoring `open311_munge_update_params`

42cbb6870bfa6f0eb5c2ed792c54f138267ff86d is using the customer reference to set service_request_id as discussed

MorayMySoc:
This last fixup is intended to assume a regular Multi open311 backend

perllib/FixMyStreet/Script/BANES/PassthroughConfirm.pm
```diff
@@ -0,0 +1,115 @@
...
+        $item->_create_vars;
+        $item->h->{sending_to_banes_passthrough} = 1;
+        my $sender_info = {
+            method => 'Open311',
+            config => $self->body,
+        };
+        my $reporter = FixMyStreet::SendReport::Open311->new;
+        $reporter->add_body( $self->body, $sender_info->{config} );
+        $item->_set_reporters([$reporter]);
+        my $service_code = 'Passthrough-' . $row->contact->email . '@example.org';
```
dracos, line 33: So if I've thought it through right, I think you can now remove the `@example.org` here (and no need for the if statement in your open311-adapter change to strip it off again, because the code is now using the `Passthrough-` prefix and that's all it needs?

perllib/FixMyStreet/Script/BANES/PassthroughConfirm.pm
```diff
@@ -0,0 +1,115 @@
...
+sub send_comments {
+    my ($self) = @_;
+    my $comments = $self->_comments;
+    while (my $row = $comments->next) {
+        my $problem = $row->problem;
+        my $confirm_id = $row->external_id;
+        my $cobrand = $problem->get_cobrand_logged;
+        FixMyStreet::DB->schema->cobrand($cobrand);
+        $cobrand->set_lang_and_domain($problem->lang, 1);
+        my $service_code = 'Passthrough-' . $row->problem->contact->email . '@example.org';
```
dracos, line 57: Ditto

dracos (changes requested):
Given history, please check this but I think you can now drop the 'having to make it look like an email' bit. Looks good, thanks for doing that :)

dracos (approved):
Hope I am right in what I said this time 🤣

MorayMySoc:
Fix up over issue with duplicate Passthrough prefix caused by switching to using Multi in a more standard way

dracos (approved):

PR #5409, opened 2025-03-11
[Bexley][WW] Customize paye.net narrative field for GGW

chrismytton:
- Add support for custom narratives in the Paye integration
- Add method to Bexley garden module to generate a custom narrative
- Update Paye role to use the custom narrative method if available

For https://3.basecamp.com/4020879/buckets/40373795/todos/8407638567

### DIFF
--- a/t/app/controller/waste_bexley_garden.t
+++ b/t/app/controller/waste_bexley_garden.t
@@ -1653,6 +1653,33 @@ FixMyStreet::override_config {
         is $report->get_extra_metadata('direct_debit_reference'), 'APIRTM-DEFGHIJ1KL', 'Correct payer reference';
         is $report->state, 'confirmed', 'Report is confirmed';
     };
+
+    subtest 'Staff garden waste subscription uses paye.net with custom narrative' => sub {
+        $sent_params = {};
+        $call_params = {};
+        default_mocks();
+        $mech->log_in_ok($staff_user->email);
+
+        $mech->get_ok('/waste/10001/garden');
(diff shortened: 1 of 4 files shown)


t/app/controller/waste_bexley_garden.t
```diff
@@ -1642,6 +1642,19 @@ FixMyStreet::override_config {
...
+
+        my $report = FixMyStreet::DB->resultset("Problem")->first;
+        $report->update_extra_field({ name => 'uprn', value => '12345678' });
+
+        my $report_id = $report->id;
+        is $cobrand->waste_get_paye_narrative($report),
+            "Garden Waste Service Payment - Reference: $report_id Contract: 12345678",
+            'Narrative includes reference and contract details';
+    };
+
```
chrismytton, line 1657: Not sure if this unit test is enough, or if I should write a full integration test for the redirect URL code?

dracos: Interestingly, the test file does already have code to catch call_params and sent_params for any Paye integration calls. I guess having a test that is only a staff member creating a subscription would be a good test to have anyway, not just for this bit? Brent has a staff-subscribe-to-sacks test that e.g. has `is $mech->res->previous->header('Location'), 'http://paye.example.org/...` and checks `$sent_params->items{0}...` so I don't think you'd need to do much more than the steps-to-subscribe for a staff member, as it looks like the stuff to catch the low-level stuff is already present, just not used.

dracos (approved):
Woo :)

PR #5411, opened 2025-03-11
[BANES][FMS] Allows protection of open311 passthrough contacts

MorayMySoc:
All passthrough contacts are manually added as the passthrough does not return a service list.

We extend protection from open311 changes to stop the removal of categories that don't have a corresponding service.

https://github.com/mysociety/societyworks/issues/4746

[skip changelog]


dracos (approved):
Thanks, good documentation.
Perhaps this should be the default...

PR #5419, opened 2025-03-13
[SLWP] Make a container removal event if decrease in garden bins

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4826 [skip changelog]


MorayMySoc (approved):
This looks good to me

PR #5421, opened 2025-03-14
[Bexley] Add Content-Length header to DD requests

chrismytton:
This ensures that HTTP requests made to the Access PaySuite API have the correct Content-Length header set, which it seems to require, for POST requests at least.

Spotted this when trying to archive contracts manually from the CLI.

### DIFF
--- a/perllib/Integrations/AccessPaySuite.pm
+++ b/perllib/Integrations/AccessPaySuite.pm
@@ -93,7 +93,6 @@ sub headers {
 
     return {
         'Accept' => 'application/json',
-        'Content-Type' => 'application/x-www-form-urlencoded',
         'User-Agent' => 'WasteWorks by SocietyWorks (swtech@societyworks.org)',
         'ApiKey' => $self->config->{api_key},
     };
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
I'd suggest using HTTP::Request::Common (which this file includes but doesn't use), which does this for you.

`GET/DELETE $url, %{$self->headers}` for GET/DELETE
`POST/PUT $url, $data, %{$self->headers}` for POST/PUT.
You wouldn't need Content-Type or Content-Length in `headers()`, and it does the building of data for POST for you.

perllib/Integrations/AccessPaySuite.pm
```diff
@@ -128,13 +127,28 @@ sub build_request_url {
...
+            Content => $content,
+            %$headers
+        );
+    } elsif ($method eq 'GET') {
+        return HTTP::Request::Common::GET($url, %$headers);
+    } elsif ($method eq 'DELETE') {
+        return HTTP::Request::Common::DELETE($url, %$headers);
+    } else {
+        die "Unsupported method: $method";
+    }
```
dracos, line 151: I think this can be reduced to:
```suggestion
    return HTTP::Request::Common->can($method)->($url,
        $method eq 'POST' || $method eq 'PUT' ? $data : (),
        %$headers
    );
```
(if you supply $data as the second argument for post/put it will build the form data for you)

dracos (changes requested):
Looks good, one suggestion to only have the one method call :)

dracos (approved):
Ah, great, thanks :)

PR #5422, opened 2025-03-14
[Bexley][WW] Send UPRN in additionalReference field for DDs

chrismytton:
When creating a Direct Debit in Access PaySuite Bexley want the property UPRN to go in the additionalReference field to make it easier for them to match things up.

For [Additional reference field in DDCMS](https://3.basecamp.com/4020879/buckets/40373795/todos/8415278536)


dracos (approved):
Assuming that's definitely what they want :)

PR #5423, opened 2025-03-14
[Lincs] Enable category display_name

davea:
Required for FD-5227.

[skip changelog]


dracos (approved):

dracos:
This never got included, they don't seem to mind Obstruction has the brackets after it, going to close unless turns up again.

PR #5434, opened 2025-03-24
[Merton] Don’t retry if no Echo ID returned.

dracos:
Probably should have a test

### DIFF
--- a/perllib/FixMyStreet/Script/Merton/SendWaste.pm
+++ b/perllib/FixMyStreet/Script/Merton/SendWaste.pm
@@ -104,15 +104,18 @@ sub set_echo_id {
         my $cfg = $cobrand->feature('echo');
         my $echo = Integrations::Echo->new(%$cfg);
         my $event = $echo->GetEvent($problem->external_id);
-        return unless $event && $event->{Id};
-
         my $db = FixMyStreet::DB->schema->storage;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Script/Merton/SendWaste.pm
```diff
@@ -104,15 +104,17 @@ sub set_echo_id {
...
-                name => 'echo_id',
-                value => $event->{Id},
-            });
+            if ($event && $event->{Id}) {
+                $row2->update_extra_field({
+                    name => 'echo_id',
+                    value => $event->{Id},
+                });
+            } else {
+                $row2->set_extra_metadata( no_echo => 2 );
```
davea, line 116: Perhaps a comment here to explain what `2` means, for future greppers?

davea (approved):
Test probably worthwhile, and small suggestion to add a comment. But not blockers for merging given nature of the fix.

perllib/FixMyStreet/Script/Merton/SendWaste.pm
dracos: Turns out '1' wasn't actually used, that was code that hadn't been removed when Merton wanted garden transfer done entirely differently, so will just switch to more normal one.

PR #5436, opened 2025-03-25
[Bexley] If no Agile, but Whitespace, remove Whitespace.

dracos:
[skip changelog] For FD-5282

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
@@ -266,24 +266,14 @@ sub bin_services_for_address {
     $property->{red_tags} = $property_logs;
     $property->{service_updates} = $street_logs;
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Waste.pm
```diff
@@ -481,8 +463,26 @@ sub bin_services_for_address {
...
+
+    # To begin with we assume the property is eligible to sign up to GGW...
+    $property->{garden_signup_eligible} = 1;
+    # unless it's got a parent property,
+    # or no services of its own
+    # or it already has a subscription in Agile
+    # or it has sacks
+    if (   $property->{parent_property}
+        || !@site_services_filtered
+        || $property->{garden_current_subscription}
```
chrismytton, line 481: I feel like `$property->{has_garden_subscription}` would be more semantically correct here, though it obviously doesn't make any material difference!

chrismytton (approved):

PR #5438, opened 2025-03-27
[Bexley] Don't show form on double submission

chrismytton:
If a user refreshes the page after submitting the form we detect this and show them an error.

I also tried adding a check for an existing sub in Agile before creating the Direct Debit, but this doesn't work in the case of a renewal where a user is switching from Credit/Debit Card to Direct Debit, in which case there will already be a sub in Agile. Might be possible to do something there, but was a bit fiddly, and we're already checking for an existing sub in the main flow, so this is a bit of an edge case.

[Related Slack thread](https://mysociety.slack.com/archives/C07H1MFDNRG/p1742908679431989)

image

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Form.pm
+++ b/perllib/FixMyStreet/App/Controller/Form.pm
@@ -87,10 +87,15 @@ sub form : Private {
 
     $form->process unless $form->processed;
 
-    # If we have sent a confirmation email, that function will have
-    # set a template that we need to show
-    $c->stash->{template} = $c->stash->{override_template} || $form->template || $self->index_template
-        unless $c->stash->{sent_confirmation_message};
+    # If the form has the already_submitted_error flag set, show the already_submitted template
+    if ($form->already_submitted_error) {
+        $c->stash->{template} = 'waste/already_submitted.html';
+    } else {
+        # If we have sent a confirmation email, that function will have
+        # set a template that we need to show
+        $c->stash->{template} = $c->stash->{override_template} || $form->template || $self->index_template
+            unless $c->stash->{sent_confirmation_message};
+    }
     $c->stash->{form} = $form;
 }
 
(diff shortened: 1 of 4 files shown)


templates/web/base/waste/already_submitted.html
```diff
@@ -0,0 +1,19 @@
...
+    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+      Important
+    </h2>
+  </div>
+  <div class="govuk-notification-banner__content">
+    <p class="govuk-notification-banner__heading">
+      You have already submitted this form.
+    </p>
+    <p class="govuk-body">To avoid duplicate submissions, this form cannot be resubmitted. If you need to make changes or have questions about your submission, please contact us.</p>
+    <a href="/" class="govuk-button">Return to homepage</a>
```
dracos, line 15: Could this be to /waste

perllib/FixMyStreet/App/Controller/Form.pm
```diff
@@ -87,10 +87,15 @@ sub form : Private {
 
     $form->process unless $form->processed;
 
-    # If we have sent a confirmation email, that function will have
-    # set a template that we need to show
-    $c->stash->{template} = $c->stash->{override_template} || $form->template || $self->index_template
-        unless $c->stash->{sent_confirmation_message};
+    # If the form has the already_submitted_error flag set, show the already_submitted template
+    if ($form->can('already_submitted_error') && $form->already_submitted_error) {
```
dracos, line 91: Do we need the can part here? I think every form used here is a Wizard subclass, so would be good to know we always have it (and put it in if there's somewhere added that lacks it).

dracos (changes requested):
Looks good, couple of tiny comments

dracos (approved):
Thanks!

PR #5440, opened 2025-03-28
[Bexley] Default email address for Direct Debit when not provided

chrismytton:
Staff aren't required to enter an email address, so set a default in that case as the email is required by Access PaySuite for setting up the Direct Debit.

For [this Basecamp todo](https://3.basecamp.com/4020879/buckets/40373795/todos/8471147856).


dracos (approved):
Thanks

PR #5456, opened 2025-04-07
[Peterborough][WW] Update wording when can't report missed collection.

neprune:
https://mysocietysupport.freshdesk.com/a/tickets/5273

[skip changelog]


dracos (approved):
Is fine, I wonder if it'd be bit more effective if it wasn't /all/ in bold, but probably not worth pushing back on.

PR #5464, opened 2025-04-09
Sutton escalations (scheduled and bulky missed)

dracos:
For https://github.com/mysociety/societyworks/issues/4802 https://github.com/mysociety/societyworks/issues/4803 https://github.com/mysociety/societyworks/issues/4800 - two of the parts.
[skip changelog]

Hope this makes sense; I think then the container request changes should hopefully be more similar to the scheduled missed than the bulky - need to see if there's an open container request event after the required timeframe, no existing escalation, and add a "Failure to deliver" option/category to the "Report a problem" flow.


neprune (approved):
Nice, reads really clearly.

PR #5466, opened 2025-04-10
[Bristol] Send Dott reports to a different email address

davea:
For FD-4905

To test locally you will need:

```yaml
COBRAND_FEATURES:
  dott_email:
    bristol: foo@bar.example.org
```

[skip changelog]


chrismytton (approved):
🚲

PR #5469, opened 2025-04-16
[NorthNorthants] Added one.network message

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4883

Screenshot 2025-04-16 at 07 21 22

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Root.pm
+++ b/perllib/FixMyStreet/App/Controller/Root.pm
@@ -122,7 +122,7 @@ sub page_error_404_not_found : Private {
 sub page_error_410_gone : Private {
     my ( $self, $c, $error_msg ) = @_;
     $c->stash->{template}  = 'index.html';
-    $c->stash->{error} = $error_msg;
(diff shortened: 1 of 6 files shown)


lucascumsille:
@chrismytton I haven't sent this to the client yet, because I wanted you opinion regarding my approach here:

Instead of duplicating the `postcode_form.html` which I don't think we have done so far and probably it wouldn't be great for maintenance I ended up creating a new template called `front-main-message`

I tried adding the `bodyclass` conditional, so it only applies for the homepage, but for it doesn't work and 
`[% IF bodyclass.match('frontpage') %][% TRY %][% INCLUDE "front/front-main-message.html" %][% CATCH file %][% END %][% END %]`

Let me know if in the meantime I can put this on staging and ask the client if they are happy with it.

Another option would have been to put the message in `postcode_form_geolocation`, but it doesn't seem like the right approach considering the name/use of that file.

lucascumsille:
@dracos I tried and it works well, thank you =)

PR #5471, opened 2025-04-20
[Bexley] Fix child properies with services

chrismytton:
Correctly handle addresses with a parent shell record. Previously, the panel was hidden for all addresses with a parent, under the assumption that services are tied to the parent record (as is the case with communal collections).

However, some properties have a parent record that is unused, with all services (including kerbside collections) linked to the child record. In these cases, the panel should be visible.

## Todo

- [ ] Fix tests

For FD-5332


dracos (approved):
Looks good, I assume the edge case of a property with services but no filtered services shouldn't really ever happen and them subscribing to garden waste would probably be okay then anyway :)

PR #5374, opened 2025-02-18
[Bexley][WW] Allow modification of subscriptions

davea:
For https://github.com/mysociety/societyworks/issues/4677

[skip changelog]

PR #5472, opened 2025-04-20
[Bexley] GGW amend DD amount

chrismytton:
Adds code for taking a one-off payment to cover any in-year difference in cost, as well as code for changing the amount of the original contract for future payments.

For https://github.com/mysociety/societyworks/issues/4796

Part of https://github.com/mysociety/fixmystreet/pull/5374

## Notes/Todo

The live API doesn't have the "No Frequency - AD-HOC Payments" schedule, it's only in test, which you can see with the following calls:

```
$ curl -L -g -X GET 'https://playpen.accesspaysuite.com/api/v3/client/APIRTM/schedules' -H "apiKey: $APS_TEST_API_KEY"
$ curl -L -g -X GET 'https://ddcms.accesspaysuite.com/api/v3/client/LBBEX/schedules' -H "apiKey: $APS_LIVE_API_KEY"
```

You can also get the schedule ID you need for the `dd_adhoc_schedule_id` config setting from that playpen call.

- [x] Fix tests (there seems to be an issue with the DD discount not being applied currently)
- [x] Test things are working with an existing contract in the APS test API
- [x] Add error checking after creating the adhoc contract in `one_off_payment`

(Added 08/07/25:)

- [ ] Cost displayed on bin days page needs to be internally calculated, not fetched from Agile as the latter is likely to be incorrect

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1308,6 +1308,7 @@ sub garden_modify : Chained('garden_setup') : Args(0) {
         };
 
         $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Garden::Modify';
+        $c->stash->{current_payment_method} = $payment_method;
     }
 
     $c->stash->{first_page} = 'intro';
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -485,7 +485,6 @@ sub direct_debit_modify : Private {
 
     my $ref = $c->stash->{orig_sub}->get_extra_metadata('payerReference');
     $p->set_extra_metadata(payerReference => $ref);
-    $p->confirm if $c->cobrand->moniker eq 'bexley';
     $p->update;
     $c->cobrand->call_hook('waste_report_extra_dd_data');
 
@@ -517,6 +516,11 @@ sub direct_debit_modify : Private {
         amount => sprintf('%.2f', $total / 100),
         orig_sub => $c->stash->{orig_sub},
     } );
+
+    if ( $c->cobrand->moniker eq 'bexley' ) {
+        $p->confirm;
+        $p->update;
+    }
 }
 
 sub direct_debit_cancel_sub : Private {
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -485,6 +485,7 @@ sub direct_debit_modify : Private {
 
     my $ref = $c->stash->{orig_sub}->get_extra_metadata('payerReference');
     $p->set_extra_metadata(payerReference => $ref);
+    $p->confirm if $c->cobrand->moniker eq 'bexley';
```
nephila-nacrea: Am wondering if the DD report should be confirmed further down, after we're sure the APS calls have been successful.

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -496,6 +497,8 @@ sub direct_debit_modify : Private {
 
     my $i = $c->cobrand->get_dd_integration;
 
+    # TODO Display error messages to user
```
davea, line 499: Does this need to be done as part of this PR or is that future work?

perllib/FixMyStreet/App/Controller/Waste.pm
davea: Calling `confirm` would mean the report gets sent to Whitespace - something we probably don't want to happen if we know there were DD errors. Looks like we only confirm new GGWs after APS was successful, so guess the same pattern should apply here?

perllib/Integrations/AccessPaySuite.pm
```diff
@@ -249,4 +253,106 @@ sub cancel_plan {
...
+    } elsif (ref $resp eq 'HASH' && keys %$resp) {
+        # Assuming a successful response might return some data, but 1 indicates success
+        return 1;
+    } elsif (ref $resp eq 'HASH' && !keys %$resp) {
+        # Handle cases where success might be an empty hash {}
+        return 1;
+    } else {
+        # Handle unexpected response format or potential success cases not returning a hash
+        # Log or handle as appropriate, returning 1 for presumed success if no error indicated
+        warn "Unexpected response format from AccessPaySuite adhoc payment: " . Dumper($resp);
```
davea: This should be logged in the same way as other messages.

davea (changes requested):
Looking good but I think some small changes needed.

perllib/FixMyStreet/App/Controller/Waste.pm
nephila-nacrea: Future work (maybe) - I didn't want to affect other cobrands in this PR, but was surprised by the lack of error handling/messaging iirc.

davea (approved):

PR #5474, opened 2025-04-22
[Merton][Echo] Amend Bulky collection on same date

MorayMySoc:
When amending a bulky booking but keeping the same collection date, don't cancel the report.

Use an update on the report to send the data to Open311 and do the update to same booking.

https://github.com/mysociety/societyworks/issues/4732

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -328,7 +328,27 @@ sub process_bulky_amend : Private {
 
     my $p = $c->stash->{amending_booking};
 
-    if ($c->cobrand->bulky_cancel_by_update) {
+    # In this case we don't want to go through the process of cancelling the original
+    # report as we don't want to risk losing the booking slot
+    if ( $c->cobrand->call_hook('update_report_for_same_day_amend', $p, $data) ) {
+        $p->create_related( moderation_original_data => {
+            title => $p->title,
+            detail => $p->detail,
+            photo => $p->photo,
+            anonymous => $p->anonymous,
+            category => $p->category,
+            extra => $p->extra,
+        });
+        $c->forward( '/admin/log_edit', [ $c->cobrand->body->comment_user->id, 'problem', 'edit' ] );
+        $p->photo('');
+        amend_extra_data($c, $p, $data);
+        $p->update;
+        $p->add_to_comments( {
+                text => 'Amend ' . $p->category,
+                user => $c->cobrand->body->comment_user,
+                problem_state => $p->state,
+        });
+    } elsif ($c->cobrand->bulky_cancel_by_update) {
         # In this case we want to update the event to mark it as cancelled,
         # then create a new event with the amended booking data from the form
         my $update = add_cancellation_update($c, $p, 'delayed');
--- a/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
@@ -1302,6 +1302,94 @@ sub per_photo_size_limit_for_report_in_bytes {
     return min($max_size_per_image, $max_size_per_image_from_total);
 };
 
+=head2 update_report_for_same_day_amend
+
+This Echo only hook is used to check
+if a Bulky collection that has been changed
+has maintained the same date
+
+=cut
+
+sub update_report_for_same_day_amend {
+        my ($self, $p, $data) = @_;
+
+        my $collection_date = $p->get_extra_field_value('Collection_Date');
+        return $data->{chosen_date} =~ /^$collection_date;/
+};
+
+=head2 waste_munge_bulky_amend
+
+If a Bulky collection has been amended, but the
+date of collection remains the same, we need
+to reset the data in the original report as
+that will continue to reflect the actual
+booking
+
+=cut
+
+sub waste_munge_bulky_amend {
+    my ($self, $p, $data) = @_;
+
+    for my $extra (keys %{$p->get_extra_metadata}) {
+        if ($extra =~ /item_/) {
+            $p->unset_extra_metadata($extra);
+        }
+    }
+    $self->waste_munge_bulky_data($data);
+
+    $self->{c}->stash->{report} = $p;
+    $self->save_item_names_to_report($data);
+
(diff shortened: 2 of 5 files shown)


perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -328,7 +328,27 @@ sub process_bulky_amend : Private {
...
+        $p->create_related( moderation_original_data => {
+            title => $p->title,
+            detail => $p->detail,
+            photo => $p->photo,
+            anonymous => $p->anonymous,
+            category => $p->category,
+            extra => $p->extra,
+        });
+        $c->forward( '/admin/log_edit', [ $c->cobrand->body->comment_user->id, 'problem', 'edit' ] );
+        $p->photo('');
```
dracos, line 343: Is this line needed? Looks like amend_extra_data always sets `$p->photo` anyway.

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1302,6 +1302,94 @@ sub per_photo_size_limit_for_report_in_bytes {
...
+        } elsif ($field->{name} eq 'Bulky_Collection_Notes') {
+            $field->{value} = $data->{'extra_Bulky_Collection_Bulky_Notes'} || '';
+        }  elsif ($field->{name} eq 'TEM_-_Bulky_Collection_Description') {
+            $field->{value} = $data->{'extra_TEM_-_Bulky_Collection_Description'} || '';
+        } elsif ($field->{name} eq 'TEM_-_Bulky_Collection_Item') {
+            $field->{value} = $data->{'extra_TEM_-_Bulky_Collection_Item'} || '';
+        }
+    };
+    $p->set_extra_fields(@fields);
+    $p->update;
```
dracos, line 1358: Is this line needed, as you call `$p->update` soon after anyway after amend_extra_data

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1302,6 +1302,94 @@ sub per_photo_size_limit_for_report_in_bytes {
...
+This Echo only hook is used to check
+if a Bulky collection that has been changed
+has maintained the same date
+
+=cut
+
+sub update_report_for_same_day_amend {
+        my ($self, $p, $data) = @_;
+
+        my $collection_date = $p->get_extra_field_value('Collection_Date');
```
dracos, line 1316: Note that this can be different per cobrand - Kingston/Sutton it is now `Collection_Date_-_Bulky_Items` (might need to also default back to this for old reports though, sorry).

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -328,7 +328,27 @@ sub process_bulky_amend : Private {
...
+    if ( $c->cobrand->call_hook('update_report_for_same_day_amend', $p, $data) ) {
+        $p->create_related( moderation_original_data => {
+            title => $p->title,
+            detail => $p->detail,
+            photo => $p->photo,
+            anonymous => $p->anonymous,
+            category => $p->category,
+            extra => $p->extra,
+        });
+        $c->forward( '/admin/log_edit', [ $c->cobrand->body->comment_user->id, 'problem', 'edit' ] );
```
dracos, line 342: I don't think we need this line - the Peterborough else case a few lines below doesn't have it, and the moderation entry is logging the change. Unless you think Peterborough should have it as well?

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -328,7 +328,27 @@ sub process_bulky_amend : Private {
...
+        $p->photo('');
+        amend_extra_data($c, $p, $data);
+        $p->update;
+        $p->add_to_comments( {
+                text => 'Amend ' . $p->category,
+                user => $c->cobrand->body->comment_user,
+                problem_state => $p->state,
```
dracos, line 349: I wonder if we want to set a flag on this rather than rely on precise comment text.

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1302,6 +1302,94 @@ sub per_photo_size_limit_for_report_in_bytes {
...
+            $p->unset_extra_metadata($extra);
+        }
+    }
+    $self->waste_munge_bulky_data($data);
+
+    $self->{c}->stash->{report} = $p;
+    $self->save_item_names_to_report($data);
+
+    my @fields = @{$p->get_extra_fields};
+    for my $field (@fields) {
```
dracos, line 1344: Rather than this way round, could this loop over all the fields perhaps be instead:
```
foreach (qw(Bulky_Collection_Bulky_Items Bulky_Collection_Notes
    TEM_-_Bulky_Collection_Description TEM_-_Bulky_Collection_Item Exact_Location)) {
    if ($data->{"extra_$_"}) {
        $p->update_extra_field({ name => $_, value => $data->{"extra_$_"} });
    }
}
```
Does that work okay?

dracos: Could even put the cobrand specific fields in a cobrand function and then call it for the list here and in your open_311_bulky_update_same_date perhaps

dracos (changes requested):
This all looks good, made a few small comments.
It is missing one thing however - the fact that if the current date of the booking isn't included in the list of slot dates returned by Echo then it isn't shown to the user and they can't keep the same date. We need to make sure the current chosen date is always included in the list of returned dates for an amendment, so they can choose not to change it.

dracos (commented):
Oh, sorry, just thought of another thing, sigh - if someone amends a booking and adds items that brings it above the lower payment threshold, they have to pay more. The way the current cancellation works is it creates unconfirmed cancellation update, and new report, then awaits payment confirmation before confirming the cancellation and the payment on the new report.

With this new way, hmm. We can in theory create the unconfirmed comment the same way, so it won't update Echo until payment confirmed (and we'll have to update the payment amount then paid on the report as well for any further amendments); but that still leaves the changing of the actual report with the new information. Do we... store the new data on the comment and then if payment is confirmed (or no payment necessary) copy the new data to the existing report at that point? Not sure I can think of an easier way.

PR #5476, opened 2025-04-24
Add script to send stuck report summary emails.

neprune:
Similar to the send-failure-summary one used internally but can be filtered to reports for a specific body and specific categories, and uses the cobrand URL, so better suited for flagging stuck reports externally.

Also supports sending details of unconfirmed reports.

For https://github.com/mysociety/societyworks/issues/4879.

`bexley-ggw-stuck-reports-cron` branch in servers repo has the changes to set-up the cron for Bexley. Verified with test script runs on staging site.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/SendStuckReportsSummary.pm
+++ b/perllib/FixMyStreet/Script/SendStuckReportsSummary.pm
@@ -1,0 +1,77 @@
+package FixMyStreet::Script::SendStuckReportsSummary;
+
+use v5.14;
+use warnings;
+
+use FixMyStreet::DB;
+use FixMyStreet::Email;
+use Lingua::EN::Inflect qw(PL_N PL_V WORDLIST);
+
+sub run {
+    my $params = shift;
+    my $cobrand = $params->{body}->get_cobrand_handler;
+
+    my @stuck_reports = FixMyStreet::DB->resultset('Problem')->to_body($params->{body}->id)->search({
+        category => $params->{categories},
+        send_state => 'unprocessed',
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Script/SendStuckReportsSummary.pm
```diff
@@ -0,0 +1,59 @@
...
+    my $category_count = scalar @{$params->{categories}};
+
+    my @unconfirmed_reports;
+    my $unconfirmed_reports_count;
+
+    if ($params->{unconfirmed}) {
+        @unconfirmed_reports = FixMyStreet::DB->resultset('Problem')->to_body($params->{body}->id)->search({
+            category => $params->{categories},
+            state => 'unconfirmed',
+        })->order_by('-created')->all;
```
dracos, line 30: Repeating the stuck reports each email is fine, because we want them to have to deal with them and get them processed in some manner, just like we get them. But the unconfirmed reports will not change in most cases (they'll be people who actually didn't pay, or never got that far, etc), so I think we only want to show these once per email. So we could set a flag on each one to say it's been included in this email?

dracos (commented):
Have hopefully pushed fix for the test failure.
I think we only want to output the unconfirmeds once, I can tweak that if you like

perllib/FixMyStreet/Script/SendStuckReportsSummary.pm
neprune: Yeah, makes sense. Appreciated if you can do the honours if you're around already and it's quick enough.

PR #5479, opened 2025-04-25
[Bristol] Add FMS reference to emails

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/5417

[skip changelog]


dracos (approved):
Looks good. Perhaps (not now) we should think about having a config value for whether to include this or not, or something else like that

PR #5480, opened 2025-04-28
[Merton] Remove textiles as a service

MorayMySoc:
Textiles service being removed at request of Council though batteries service will continue.

https://mysocietysupport.freshdesk.com/a/tickets/5433

[skip changelog]


davea (approved):
👍

PR #5483, opened 2025-04-29
[Peterborough] Fix closure update for graffiti categories.

dracos:
[skip changelog] Fixes FD-5438


dracos:
Dave did the original code, so I picked him, but dunno if he's available at the mo, so asked Moray as well. Whoever gets there first :)

MorayMySoc (approved):
Looks good to me

PR #5486, opened 2025-04-30
[Aberdeenshire] Confirm integration

davea:
[skip changelog]


templates/web/aberdeenshire/report/new/roads_message.html
```diff
@@ -0,0 +1,5 @@
+    <div id="js-not-a-road" class="hidden js-responsibility-message js-roads-aberdeenshire">
```
dracos, line 1: Just to note in case unaware, the `no_asset_message` config should let you do this sort of thing without needing to add any template now. (And if you are keeping the template , you'll need a .com version as well - but I'd try and use no_asset_message preferably.)

PR #5487, opened 2025-05-06
[Sutton] Fix kerbside bag image

dracos:
For https://mysocietysupport.freshdesk.com/a/tickets/5453 [skip changelog]

Bit more complicated than expected, but hopefully tidied up some code a bit whilst doing so (getting rid of waste_quantity_max which confused me as was used both for container request maximums but also garden subscription maximum, so split that out). Then stored the quantity information for all containers (is just info), and use that for showing the right image.

Asides - have Kingston work the same as Sutton looking for the container on the property; remove a bit of unused code from the old SLWP file (which is only Merton now).


MorayMySoc (approved):
Looks good to me

PR #5488, opened 2025-05-06
Add open/staff config to updates_disallowed_check

chrismytton:
This allows anyone to update an open report, but only staff can update a closed report.

Part of https://github.com/mysociety/societyworks/issues/4668 for Gloucester City.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UK.pm
+++ b/perllib/FixMyStreet/Cobrand/UK.pm
@@ -464,6 +464,15 @@ sub updates_disallowed {
     return $self->_updates_disallowed_check($cfg, $problem, $body_user);
 }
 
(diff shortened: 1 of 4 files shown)


dracos (changes requested):
Could you also update templates/web/fixmystreet-uk-councils/about/faq-en-gb.html (sounds like an extra part of the reporter-open IF) and templates/web/fixmystreet-uk-councils/admin/bodies/_updates_disallowed_hint.html to explain in the admin if a cobrand has this option, thanks

dracos (approved):
Thanks! I wonder where to document the "update these templates if you edit this" could best go - perhaps a bit of POD documentation before this function, if you wanted to just stick something in here while we're here?

PR #5489, opened 2025-05-06
Fixed icon sizing disparity for svg icons

lucascumsille:
- All icons are using the same size(1.5em)
- Chevrons are now inside a square(24px24 pixels) frame like the rest of the svg icons.

[skip changelog]


dracos (changes requested):
I think this is fine, but perhaps without the "Fixed sizing for 'get updates' icon" commit?

lucascumsille:
@dracos in my last commit I added a fixup that does the following:
- First it makes the chevron(all the different directions) to have a square frame, just like the rest of the other icons, so now they all have a square view box(24). I think this will prevent issues if we ever decide to add new icon. It's also a better practice because no matter the icon we will know it will always be a square.
- Having done the above it created a new issue, where the chevron and the RSS icon look "bigger". Technically they are not, if you inspect them you'll notice they are the same size as the rest. However because of the nature of the shape they use more space inside the square frame compared with the "share" and "warning icon", hence why they look bigger. I fixed the above by decreasing the actual icon inside the 24x24 frame so the density of all of them look similar(roughly).

dracos:
Should the all reports page 1.5rem be 1.5em?

All reports before:
image

All reports after (with list bleed through under the yellow border): 
image

Report page before:
image

Report page after:
image

Changing 1.5rem to 1.5em fixes the chevron on All Reports, I think that's my only issue with it now if that's okay with you? Happy for consistency changes, of course.

Although, if we're going for consistency, the RSS image on the /around map page is much smaller than that on either a report page or All reports, if we care about that :)

lucascumsille:
@dracos good point I missed that one =). With the last fixup I updated all the `1.5rem` to `1.5em` so now we have consistency across all pages with the icons.

templates/web/base/icons/chevron-down.html
```diff
@@ -1 +1 @@
-<svg role="presentation" aria-hidden="true" width="[% width || '1rem' %]" height="[% height || '1rem' %]" viewBox="0 0 16 13" xmlns="http://www.w3.org/2000/svg"><path d="M16 6.5V0L8 6.5L0 0V6.5L8 13L16 6.5Z" fill="currentColor"/></svg>
+<svg ole="presentation" aria-hidden="true" width="[% width || '1rem' %]" height="[% height || '1rem' %]" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M1 12V3L12 12L23 3V12L12 21L1 12Z" fill="currentColor"/></svg>
```
dracos, line 1: Sorry, missed this before, this now has two width and height, should the 24 ones be removed?

dracos (changes requested):

lucascumsille:
I added a fixup for the width and height error, but when rebasing I'll change the name of the [Fixed sizing for 'get updates' icon](https://github.com/mysociety/fixmystreet/pull/5489/commits/f1fe14cfc140044fcb4ecdc1b4edcfc1e1fba535) commit so it reflect all the change in the fixups

PR #5490, opened 2025-05-07
[Bristol] Add force resend between Alloy and Confirm

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/5416

[skip changelog]


dracos (approved):
Think this is fine; they'll need to be clearly informed as to what this means in terms of reports being resent automatically, and testing it out on staging

PR #5494, opened 2025-05-09
[National Highways] Fix issue with special email handling on .com.

dracos:
As NH is exempted from get_body_handler_for_problem in Cobrand::UK, its code for handling special emails is not automatically called. [skip changelog] Fixes failure to send #7545071.


MorayMySoc (approved):
Looks good to me

PR #5501, opened 2025-05-16
Config table

dracos:
This adds a new config database table, and moves three lists to it - two hard-coded in code (TLRN and Bucks parish MapIt IDs), and one from config file (extra parishes). It adds an admin interface to view/edit (as JSON) these entries. And puts the config admin pages in their own controller. Fixes https://github.com/mysociety/societyworks/issues/4840 [skip changelog]


perllib/FixMyStreet/App/Controller/Admin/Config.pm
```diff
@@ -0,0 +1,83 @@
...
+
+=cut
+
+=head2 index
+
+This admin page displays the overall configuration for the site.
+
+=cut
+
+sub index : Path( '' ) : Args(0) {
```
chrismytton, line 28: Why doesn't the index route seem to need the `page_error_403_access_denied` bit like the methods below?

chrismytton (approved):
Looks good! Think it will be useful to add a simple way to create config entries in the future, whether that's a CLI script or web UI, but easy enough to add in SQL for now.

PR #5506, opened 2025-05-19
Small client support things

dracos:
[skip changelog]
Fixes all of FD-5462 FD-5509 FD-5511 FD-5512


MorayMySoc (approved):
Think these are all good

PR #5512, opened 2025-05-29
[Gloucester] Increase map zoom to 5

chrismytton:
Fixes https://github.com/mysociety/societyworks/issues/4932


nephila-nacrea (approved):
This was a bit tricky to review but I think it looks OK... 😛

PR #5514, opened 2025-05-30
[National Highways] Adds Traffic Scotland road network handling

MorayMySoc:
Traffic Scotland road network is combined with the National Highways data in our asset layer.

As the behaviour is identical when falling on a road belonging to TS and HE, we can use the same logic as long as the code differentiates between which body it is dealing with.
    
Presumes we will add an '(TS)' for all categories that are Transport Scotland.
    
https://github.com/mysociety/societyworks/issues/3442


web/cobrands/highways/assets.js
```diff
@@ -4,22 +4,15 @@ if (!fixmystreet.maps) {
     return;
 }
 
-var defaults = {
-    http_wfs_url: "https://tilma.mysociety.org/mapserver/highways",
+var live = $(location).attr('href').search('www.fixmystreet.com') + 1;
```
dracos: we have a `fixmystreet.staging` variable that can used here (see e.g. roadworks.js)

davea:
On `master` in b2c38c8aea52abb89db883b1af97feb99122ae77.

PR #5518, opened 2025-06-04
Add new section to All Reports page

sallybracegirdle:
Had a go at creating a new section to signpost users to page about understanding report data. Really not sure I have the IF metrics correct.

Please check the following:

- [x] Has the code POD documentation been added or updated?
- [x] Whether this PR should include changes to any public documentation, or the FAQ;

Thanks for contributing!

### DIFF
--- a/templates/web/base/reports/index.html
+++ b/templates/web/base/reports/index.html
@@ -107,6 +107,9 @@
                 <tr><td>[% loc('Overall average') %]</td><td>[% tprintf(nget("%s day", "%s days", average), average) %]</td></tr>
             </tfoot>
         </table>
+      [% ELSIF NOT body AND c.cobrand.moniker == 'fixmystreet' %]
+        <h2 class="dashboard-subheading">[% loc('Understanding FixMyStreet report data') %]</h2>
+        <p>[% loc('Interested in FixMyStreet report data? <a href="/about/understanding-report-data">Learn about what it can and can’t tell you</a>') %]</p>
       [% END %]
     </div>
     <div class="dashboard-item dashboard-item--6">


templates/web/base/reports/index.html
```diff
@@ -109,6 +109,22 @@ <h2 class="dashboard-subheading">[% loc('Top 5 responsive councils') %]</h2>
         </table>
       [% END %]
     </div>
+    <div class="dashboard-item dashboard-item--6">
+      [% IF NOT body AND c.cobrand.moniker != 'fixmystreet' %]
+        <h2 class="dashboard-subheading">[% loc('Understanding FixMyStreet report data') %]</h2>
+        <p>[% loc('Interested in FixMyStreet report data? <a href='https://www.fixmystreet.com/about/understanding-report-data'> Learn about what it can and can’t tell you.</a>') %]</p>
+        <table class="dashboard-ranking-table">
```
dracos: You don't want this table back presumably :)

templates/web/base/reports/index.html
```diff
@@ -109,6 +109,22 @@ <h2 class="dashboard-subheading">[% loc('Top 5 responsive councils') %]</h2>
         </table>
       [% END %]
     </div>
+    <div class="dashboard-item dashboard-item--6">
+      [% IF NOT body AND c.cobrand.moniker != 'fixmystreet' %]
```
dracos: We want to show this on FMS so `==`, not `!=`.

templates/web/base/reports/index.html
```diff
@@ -109,6 +109,22 @@ <h2 class="dashboard-subheading">[% loc('Top 5 responsive councils') %]</h2>
         </table>
       [% END %]
     </div>
+    <div class="dashboard-item dashboard-item--6">
```
dracos: We don't want anther column, we want this inside the above `div` as it's the contents of that column we're changing, not an extra column.

dracos (changes requested):
In the right area :) I'll make these changes.

PR #5529, opened 2025-06-13
[OIDC] Use ‘Sign in’ rather than ‘Login’

davea:
[skip changelog]


dracos (changes requested):
There's one more in the title in templates/web/base/errors/generic.html (which is only used in OIDC context) while we're here, thanks!

dracos (approved):

PR #5533, opened 2025-06-16
[Waste] Small items with bulky

dracos:
Allow small items and bulky to coexist, so both can be used at once on the same cobrand [skip changelog] 
Fixes https://github.com/mysociety/societyworks/issues/4926

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
@@ -14,6 +14,8 @@ use HTML::FormHandler::Moose;
 extends 'FixMyStreet::App::Form::Waste';
 use FixMyStreet::Template::SafeString;
 
+has small_items => ( is => 'ro', default => 0 );
+
 has_page choose_date_earlier => (
     fields => [ 'continue', 'chosen_date', 'show_later_dates' ],
     title => 'Choose date for collection',
@@ -125,6 +127,9 @@ has_page summary => (
         return 0;
     },
     finished => sub {
+        if ($_[0]->small_items) {
+            return $_[0]->wizard_finished('process_small_items_data');
+        }
         if ($_[0]->c->stash->{amending_booking}) {
             return $_[0]->wizard_finished('process_bulky_amend');
         } else {
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
@@ -93,7 +93,7 @@ has_page summary => (
     update_field_list => sub {
         my ($form) = @_;
         my $data = $form->saved_data;
-        my $new = _renumber_items($data, $form->c->cobrand->bulky_items_maximum);
+        my $new = _renumber_items($data, $form->c->stash->{booking_maximum});
         %$data = %$new;
         return {};
     },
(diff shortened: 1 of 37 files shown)


perllib/Integrations/Echo/Events.pm
```diff
@@ -0,0 +1,115 @@
+package Integrations::Echo::Events;
```
davea: Nice bit of refactoring!

templates/web/brent/waste/_more_services_sidebar.html
```diff
@@ -15,7 +15,7 @@ <h3>More services</h3>
        <li><a href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?category=Assisted+collection+add&amp;service_id=262">Set up for assisted collection</a></li>
      -->
      [% END %]
-     [% IF property.show_bulky_waste AND NOT pending_bulky_collections %]
+     [% IF property.show_small_items AND NOT collections.small_items.pending %]
```
davea, line 18: Should we have a check for `small_items_multiple_bookings` here so the link is shown?

davea (approved):
One small thought but overall this looks good to me! 👍

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -145,18 +145,19 @@ sub waste_munge_bin_services_open_requests {
...
-    $self->$orig($events, $blocked_codes);
+    $self->$orig($type, $events, $blocked_codes);
 
     # Now check for any old open missed collections that can be escalated
 
     my $cfg = $self->feature('echo');
-    my $service_id = $cfg->{bulky_service_id};
-    my $escalations = $events->filter({ event_type => 3134, service => $service_id });
+    my $service_id = $cfg->{$type . '_service_id'} or return;
+    my $service_id_missed = $cfg->{$type . '_service_id_missed'};
```
dracos: This variable isn't used.

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -145,18 +145,19 @@ sub waste_munge_bin_services_open_requests {
     }
```
dracos, line 139: This (the code change, not the test) should be a fixup to the commit "[Waste] Separate bulky/small items Echo IDs and process." because that is the point at which the function was renamed/changed.

perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
```diff
@@ -117,7 +117,6 @@ Users of this role must supply the following:
 
 =cut
 
-requires 'bulky_allowed_property';
```
dracos, line 107: See why this had to be done, bit annoying as it was useful to make sure it got added, ah well, perhaps can rejig how it does it later.

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
```diff
@@ -213,7 +213,9 @@ sub _get_dates {
             selected => $existing_date && $existing_date eq $_->{date},
             }
             : undef
-        } @{ $c->stash->{booking_class}->find_available_slots($last_earlier_date) };
+        } $c->cobrand->moniker eq 'bexley'
+         ? @{ $c->stash->{booking_class}->find_available_slots($c->stash->{property}, $last_earlier_date) }
```
dracos: The booking class has the property on it as $self->property. So you don't need to call this differently for bexley and pass in the property, you need to change the bexley function to use the property on the class.

PR #5536, opened 2025-06-17
Cache mapit parishes in memcached

chrismytton:
Cached the list of parishes from mapit in memcached. This should speed up pages that need to display a list of parishes in a dropdown.

Part of https://github.com/mysociety/societyworks/issues/4834


dracos:
I've moved the cache generation to an offline script, then outputting/reading from a file, so that I could make a number more MapIt calls in order to provide 'nearby' information for parishes in the same parent with the same name. And so there wasn't ever a time when there'd be a slow waiting for mapit when cache expired. e.g.:

image

(still tiny bit of work to do on nicest ordering I see from that)

PR #5532, opened 2025-06-16
[Gloucester] Lookup Alloy code before sending over Open311

chrismytton:
Use the lookup site code functionality to get the relevant Alloy ID for the nearest feature if it wasn't selected during report creation.

For https://3.basecamp.com/4020879/buckets/38208405/todos/8726231700

PR #5541, opened 2025-06-19
[Gloucester] Use plots in place of sites.

neprune:
[skip changelog]


dracos:
If my comment on #5532 is correct, we won't need this PR, as it will take from the already-adjusted config.

PR #5544, opened 2025-06-20
[Brent] Direct fly-tipping based on location.

dracos:
The category used for the report depends on if the report lies within an estate or a park, so check the location and update appropriately when a report is made. [skip changelog] Fixes https://github.com/mysociety/societyworks/issues/4955


dracos:
Dunno why test is failing :-/ passes locally. could still be reviewed given that I guess

perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -554,6 +554,49 @@ sub dashboard_export_problems_add_columns {
...
+
+    my $host = FixMyStreet->config('STAGING_SITE') ? "tilma.staging.mysociety.org" : "tilma.mysociety.org";
+    my ($x, $y) = $problem->local_coords;
+    my $filter = "(<Filter xmlns:gml=\"http://www.opengis.net/gml\"><Intersects><PropertyName>geom</PropertyName><gml:Point srsName=\"27700\"><gml:coordinates>$x,$y</gml:coordinates></gml:Point></Intersects></Filter>)";
+    my $cfg = {
+        url => "https://$host/mapserver/brent",
+        srsname => "urn:ogc:def:crs:EPSG::27700",
+        typename => 'Parks_and_Open_Spaces,Housing',
+        outputformat => "GML3",
+        filter => $filter x 2,
```
chrismytton, line 592: Why does the filter need to be included twice?

dracos: You have to specify a filter for every `typename` you supply (so you can actually have different filters for each layer if you wanted). Here, we want the same 'contains' check for both layers we're querying, so we repeat the filter.

PR #5550, opened 2025-06-24
[Glos City] Do not display County 'dead animal' category

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/38208405/todos/8778506727 .

[skip changelog]


chrismytton (approved):
Longer-term it feels like county and district should have a shared dead animals category that routes based on…whatever the difference is. But this is a good stop-gap solution until then 🙂

PR #5552, opened 2025-06-25
[OIDC] Allow custom redirect parameter name when logging out

davea:
Aberdeenshire, for example, doesn’t use ‘post_logout_redirect_uri’

[skip changelog]


dracos (approved):
🌳

PR #5556, opened 2025-06-27
[Peterborough] Use private land layer for Abandoned vehicles category

chrismytton:
This will be shown when people try and make a report on land that Peterborough aren't responsible for.

I've refactored the asset layers so that there are templates for the PCC land and PCC leased land, then there's two layers each for Flytipping/Graffiti/Abandoned vehicles, one for owned and one for leased.

See also commit `d769fca1d7d3359eff4e9e7209515e815c82eb68` on the `FD-5595-peterborough-abandoned-vehicles-private-land-check` branch in /data/servers, which updates the asset layer config.

For FD-5595

image

Also tidied up the HTML for the graffiti category. HTML isn't supported in these templates because they're piped though `add_links | html_para`, so I've removed the tel:111 links and made the .com graffiti template match the Peterborough one.

Which should fix this issue:

Screenshot 2025-06-27 at 12 00 14 pm

### DIFF
--- a/.cypress/cypress/integration/peterborough.js
+++ b/.cypress/cypress/integration/peterborough.js
@@ -53,6 +53,10 @@ describe('new report form', function() {
     cy.wait('@report-ajax');
     cy.pickCategory('Non offensive graffiti');
     cy.get('.pre-button-messaging:visible').should('not.exist');
+    cy.visit('http://peterborough.localhost:3001/report/new?longitude=-0.242007&latitude=52.571903');
+    cy.wait('@report-ajax');
+    cy.pickCategory('Abandoned vehicles');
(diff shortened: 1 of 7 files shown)


chrismytton:
Have updated /data/servers to use an `if ($staging)` block, see commit `d769fca1d7d3359eff4e9e7209515e815c82eb68` on the `FD-5595-peterborough-abandoned-vehicles-private-land-check` branch.

dracos (approved):
Nice rejigging of the asset layers, makes it clear.

The FD ticket, I see he also asked for new additional conditional questions as well? Presumably that’s not part of this (at least, don’t think we can show a message for their Y to Q1), I guess you can say they can rejig the questions in the admin, within their limitations

PR #5558, opened 2025-06-30
[TfL] Add photo column to CSV export

chrismytton:
Adds a photo column to the CSV export that contains either 1 or 0 to indicate whether a report has a photo.

I'm not 100% sure I've done this correctly, I still haven't fully grokked the CSV pre-generation, so there may be more to do on this.

Fixes FD-5606


MorayMySoc (approved):
Great - thanks. I'll put this on staging

PR #5566, opened 2025-07-01
[Bexley] Bulky waste functionality

dracos:
Hopefully all makes sense, have tried to build it up from bare minimum.
The BulkyWaste Role already has some `requires` in which was quite helpful in getting starting, so we could look to extending that with the other functions that are required by the end, which seem to be: _bulky_date_to_dt, find_available_bulky_slots, check_bulky_slot_available, save_item_names_to_report (could be default?), waste_munge_bulky_data, bulky_nice_item_list (could be default?) and waste_reconstruct_bulky_data.

Fixes https://github.com/mysociety/societyworks/issues/4995
Fixes https://github.com/mysociety/societyworks/issues/4996
Fixes https://github.com/mysociety/societyworks/issues/4997
Fixes https://github.com/mysociety/societyworks/issues/5002
For https://github.com/mysociety/societyworks/issues/5003 (awaiting options)
For https://github.com/mysociety/societyworks/issues/5005 (awaiting text)
Fixes https://github.com/mysociety/societyworks/issues/5004
Fixes https://github.com/mysociety/societyworks/issues/5006
Fixes https://github.com/mysociety/societyworks/issues/5007
Fixes https://github.com/mysociety/societyworks/issues/5009
Fixes https://github.com/mysociety/societyworks/issues/5010
For https://github.com/mysociety/societyworks/issues/5021
For https://github.com/mysociety/societyworks/issues/5022

[skip changelog]

### DIFF
--- a/templates/web/base/waste/index.html
+++ b/templates/web/base/waste/index.html
@@ -25,7 +25,7 @@
                 <li>Order new or additional bins</li>
             [% END %]
             <li>Subscribe to garden waste collection service</li>
-            [% IF c.cobrand.call_hook('bulky_enabled') && (c.cobrand.moniker == 'sutton' || c.cobrand.moniker == 'kingston') %]
+            [% IF c.cobrand.call_hook('bulky_enabled') %]
               <li>Book a bulky waste collection</li>
             [% END %]
           </ul>
--- a/templates/web/bexley/waste/_more_services_sidebar.html
+++ b/templates/web/bexley/waste/_more_services_sidebar.html
@@ -25,6 +25,8 @@
 
   <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/clinical-waste-collections/get-clinical-waste-collected">Get clinical waste collected</a></li>
   <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/recycling/which-bin-should-i-use">Which bin should I use?</a></li>
-  <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/bulky-or-large-item-collections/arrange-bulky-or-large-item-collection">Arrange a bulky waste collection</a></li>
+  [% IF property.show_bulky_waste %]
+    <li><a href="[% c.uri_for_action('waste/bulky/index', [ property.id ]) %]">Arrange a bulky waste collection</a></li>
+  [% END %]
   <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/assisted-collections/help-putting-bins-out">Get help with putting your bins out</a></li>
 </ul>


perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
```diff
@@ -279,6 +279,12 @@ has_field tandc => (
...
         } elsif ($c->cobrand->moniker eq 'brent') {
             $label = 'I have read and agree to the <a href="' . $link . '" target="_blank">terms and conditions</a> and understand any additional items presented that do not meet the terms and conditions will not be collected';
+        } elsif ($c->cobrand->moniker eq 'bexley') {
+            $label = '&bull; I confirm that the submitted information is current and correct, and any misrepresentations could lead to a cancellation of the arranged service without refund.
+<br>&bull; I understand that collections can take place any time after 6am.
+<br>&bull; I understand that the arranged service cannot be edited or cancelled once payment has been submitted and refunds are not provided.
+<br>&bull; I understand that only the items added to the booking will be taken. Items must be accessible on the collection day and left in a neat and safe manner. Items cannot be left for collection on the public highway.
+<br>&bull; I understand I may be subject to further charges if proof of pension cannot be provided when requested.';
         } else {
```
MorayMySoc, line 298: i think this might need some formatting

image

MorayMySoc (approved):
This all seems good to me - just a comment on look, but maybe that's on my system

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
dracos: Meh, I see the bullet points are a bit off, but I'm not that fussed about T&Cs that I imagine no-one reads anyway :)

perllib/Integrations/SCP.pm
```diff
@@ -147,6 +147,7 @@ sub pay {
                 'scpbase:address' => ixhash(
                     'scpbase:address1' => $args->{address1},
                     'scpbase:address2' => $args->{address2},
+                    'scpbase:city' => $args->{address2}, # Town/city seems to end up in the address2 field
```
dracos: This means address2 is being passed in twice, in both 'address2' and 'city' - we should only pass it in once.
Looking at `populate_payment_details`, we get one of "address1, postcode", "address1, address2, postcode", "address1, address2, town, poscode" or "address1, address2, address3, town, postcode". So perhaps city here should be "town || address2" and address2 should be "address2 if town && address2"

nephila-nacrea: > Looking at populate_payment_details, we get one of "address1, postcode", "address1, address2, postcode", "address1, address2, town, poscode" or "address1, address2, address3, town, postcode".

How did you figure that out? It seems to be a series of shifts and pops to me; couldn't there e.g. be "address1, town, postcode"?

nephila-nacrea: (But have fixed it according to your suggestion.)

dracos: By "we get", I meant that what's would end up in the variables depending on the number of commas in the string, not that I'd know what the contents of each part would be. Yes, the actual address could be "address1, town, postcode" - but then town would end up in address2.

templates/web/base/waste/index.html
```diff
@@ -25,7 +25,7 @@ <h1>[% loc('Bins, rubbish and recycling') %]</h1>
                 <li>Order new or additional bins</li>
             [% END %]
             <li>Subscribe to garden waste collection service</li>
-            [% IF c.cobrand.call_hook('bulky_enabled') && (c.cobrand.moniker == 'sutton' || c.cobrand.moniker == 'kingston') %]
+            [% IF c.cobrand.call_hook('bulky_enabled') && (c.cobrand.moniker == 'sutton' || c.cobrand.moniker == 'kingston' || c.cobrand.moniker == 'bexley') %]
```
dracos: I think we can drop the moniker part of this entirely, not sure why it's there - if we offer bulky then we should show it?

templates/web/bexley/waste/_more_services_sidebar.html
```diff
@@ -25,6 +25,6 @@ <h3>More services</h3>
 
   <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/clinical-waste-collections/get-clinical-waste-collected">Get clinical waste collected</a></li>
   <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/recycling/which-bin-should-i-use">Which bin should I use?</a></li>
-  <li><a [% external_new_tab | safe %] href="https://www.bexley.gov.uk/services/rubbish-and-recycling/bulky-or-large-item-collections/arrange-bulky-or-large-item-collection">Arrange a bulky waste collection</a></li>
+  <li><a [% external_new_tab | safe %] href="[% c.uri_for_action('waste/bulky/index', [ property.id ]) %]">Arrange a bulky waste collection</a></li>
```
dracos: Could we a) get rid of external_new_tab as it no longer is external, and b) wrap this in a IF property.show_bulky_waste like it is in other sidebars

templates/web/base/waste/index.html
nephila-nacrea: Yeah, can't think why we wouldn't show it otherwise 🤔

PR #5567, opened 2025-07-02
[Central Bedfordshire] Adds cleaning csv column

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/5648

[skip changelog]


dracos (commented):
There are 4 categories in CBeds with an "Issue" extra field - three of them are "What is the issue?" so I assume that doesn't matter and they can filter; the fourth is graffiti and the question is actually if it's offensive or not, don't know if that would be confusing. It's probably fine, but you could check the category here if we think it matters? Also the cleaning category has an Other option, which then is a free text field, they don't say if they want to include that (they ask for a column, so I'd guess not and Other is fine, but could ask first off).

MorayMySoc:
Hmmm...I thought 'Issue' might be common, but didn't want to filter at our end in case they change the category name. I suppose I could filter to !Graffiti though.

I'll ask about the 'other' category

MorayMySoc:
They are happy without the text from the additional 'Other' field

t/cobrand/centralbedfordshire.t
```diff
@@ -232,16 +233,24 @@ subtest 'Dashboard CSV extra columns' => sub {
...
         MAPIT_URL => 'http://mapit.uk/',
         ALLOWED_COBRANDS => 'centralbedfordshire',
         PHOTO_STORAGE_OPTIONS => { UPLOAD_DIR => $UPLOAD_DIR },
     }, sub {
         $mech->get_ok('/dashboard?export=1');
     };
-    $mech->content_contains('"Site Used","Reported As",CRNo');
-    $mech->content_contains('centralbedfordshire,,' . $report->external_id . ',"Bin bags;Furniture"');
+    $mech->content_contains('"Site Used","Reported As",CRNo,"Fly-tipping types","Cleaning types"');
+    $mech->content_contains('centralbedfordshire,,' . $report->external_id . ',"Bin bags;Furniture",');
```
dracos: If this line is testing the next field is blank, guess it's not doing that, as there could be anything after the comma :) But don't think it matters.

PR #5570, opened 2025-07-03
[Sutton] Request to start ggw immediately if bins on premises

MorayMySoc:
If resident already has container, start 'New' subscription immediately

https://mysocietysupport.freshdesk.com/a/tickets/5650

[skip changelog]


dracos:
Looks like this has been merged as df585464c9

PR #5575, opened 2025-07-04
[Bulky] Improve the items page form

dracos:
No longer will it print 200 copies of the same div for Bromley (or similar for Bexley for current work), but use a template, hooray.
Fixes https://github.com/mysociety/societyworks/issues/4891
Fixes https://github.com/mysociety/societyworks/issues/5039
[skip changelog]


MorayMySoc (approved):
Looks good to me

PR #5581, opened 2025-07-10
[Aberdeenshire] Pass category changes both ways between Confirm/FMS

davea:
For https://github.com/mysociety/societyworks/issues/5043

Depends on open311-adapter running https://github.com/mysociety/open311-adapter/pull/429

[skip changelog]

### DIFF
--- a/perllib/Open311/UpdatesBase.pm
+++ b/perllib/Open311/UpdatesBase.pm
@@ -165,6 +165,7 @@ sub _process_update {
     my $body = $self->current_body;
 
     $self->_handle_assigned_user($request, $p);
+    $self->_handle_category_change($request, $p);
 
     my $state = $open311->map_state( $request->{status} );
     my $old_state = $p->state;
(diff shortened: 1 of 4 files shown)


perllib/Open311/UpdatesBase.pm
```diff
@@ -364,13 +365,49 @@ sub _handle_assigned_user {
...
-            my $contact
-                = $body->contacts->search( { category => $category } )->first;
-            $p->category($category) if $contact;
+            if (my $contact = $body->contacts->search( { category => $category } )->first) {
+                my $old = $p->category;
+                my $new = $contact->category;
+                if ($new ne $old) {
+                    $p->category($new);
+                    my $text = '*' . sprintf(_('Category changed from ‘%s’ to ‘%s’'), $old, $new) . '*';
+                    my $comment = $self->schema->resultset('Comment')->new(
```
dracos: Do you want to factor this and the main comment add around line 200ish into a function, they seem to be the same.

perllib/Open311/UpdatesBase.pm
```diff
@@ -388,17 +377,7 @@ sub _handle_category_change {
...
-                            problem => $p,
-                            user => $self->system_user,
-                            send_state => 'processed',
-                            text => $text,
-                            confirmed => $request->{comment_time},
-                            created => $request->{comment_time},
-                            private_email_text => $text,
-                        }
-                    );
+                    my $comment = $self->_comment_for_update($p, $text, $text, $request->{comment_time});
```
dracos: Now I'm seeing this - nothing to do with this, but I think we shouldn't be setting private_email_text if it's the same as text? That just leads to it not being in an update email because it thinks it's only for the problem reporter, if I understand the alert code correctly. So would be good to fix that if I'm right there.

PR #5583, opened 2025-07-11
[Shropshire][FMS] Add comment when report made private/public

MorayMySoc:
Request to have a view of when a report has been made private

https://mysocietysupport.freshdesk.com/a/tickets/5659

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -1655,6 +1655,24 @@ sub bulky_add_payment_confirmation_update {
     $self->cancel_update_alert($comment->id);
 }
 
+=head2 add_privacy_change_comment
+
+Called with a user object and an optional variable.
+
+Variable should exist when noting report made private
+otherwise will note report made public.
+
+=cut
+
+sub add_privacy_change_comment {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1655,6 +1655,24 @@ sub bulky_add_payment_confirmation_update {
...
+Variable should exist when noting report made private
+otherwise will note report made public.
+
+=cut
+
+sub add_privacy_change_comment {
+    my ($self, $user, $private) = @_;
+
+    $self->add_to_comments( {
+            text => $private ? 'Report made private' : 'Report made public',
```
dracos, line 1671: As suggested on the linked FD ticket, would this perhaps be better done by extending Moderation so that editing mark_private counts as a moderation the same as editing the title/detail/photo/anonymous etc does?

I'm not sure we want a public comment saying when a report has been made public (I guess made private doesn't matter so much, given it would then be private, but the other way round in the case of a mistake or what have you seems odd to be a public comment). It feels more like a moderation item than a comment, what do you think?

Another alternative would be to use log_edit, but that isn't exposed anywhere at present, and is more for internal checking (though no reason for that really), so would be more development.

PR #5584, opened 2025-07-14
[Gloucester] Text tweaks

dracos:
For https://3.basecamp.com/4020879/buckets/38208405/todos/8758827925
and https://3.basecamp.com/4020879/buckets/38208405/todos/8822257240

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Gloucester.pm
+++ b/perllib/FixMyStreet/Cobrand/Gloucester.pm
@@ -109,6 +109,21 @@ sub categories_restriction {
 
 =back
 
+=head2 post_report_report_problem_link
+
+Overrides the 'post-report' report another problem here button with one linking back to the front page, rather than the report view at the same location.
+
+=cut
+
+sub post_report_report_problem_link {
+    return {
+        uri => '/',
+        label => 'Report a problem',
+        attrs => 'class="report-a-problem-btn"',
+    };
+
+}
+
 =head2 pin_colour
 
 * Yellow if open/confirmed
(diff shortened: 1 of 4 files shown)


templates/web/aberdeenshire/report/new/after_photo.html
```diff
@@ -6,8 +6,3 @@
         <li id="js-photo-tips-avoid-personal-info">[% loc('Avoid personal information and vehicle number plates') %]</li>
     </ul>
 </div>
-<div class="description_tips hidden" id="js-photo-tips-bucks-vehicle-littering" aria-label="[% loc('Tips for perfect photos') %]">
```
nephila-nacrea, line 9: Why is there a change to an Ab/shire template here?

perllib/FixMyStreet/Cobrand/Gloucester.pm
```diff
@@ -109,6 +109,21 @@ sub categories_restriction {
 
 =back
 
+=head2 post_report_report_problem_link
```
nephila-nacrea, line 112: What does this do exactly? I'm not clear on what links it affects.

dracos: This is used (only) by the `templates/web/base/navigation/_report.html` template to override the link shown in the navbar at this point. Looks like it was done for BANES, I discovered this when looking to find out where the navbar "Report another problem here"  link was coming from.

templates/web/aberdeenshire/report/new/after_photo.html
dracos: This div is only used by Buckinghamshire (so on .com and Bucks), so doesn't need to be present here; it didn't seem worth a commit of its own but that would be clearer.

nephila-nacrea (approved):

PR #5585, opened 2025-07-15
[Gloucestershire] Remove roads_message.html

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/5067 .

- [ ] Needs assets config change on mysociety-servers to be live first - 99edd076

[skip changelog]


dracos (changes requested):
The js-not-a-prow section can be removed from the .com equivalent template as it was only used for Gloucestershire.

PR #5589, opened 2025-07-19
[Bexley] Check for cancellations in Agile

chrismytton:
Add a script to check Agile's LastCancelled API for recent cancellations and cancel the related Direct Debit in Access PaySuite.

## TODO

- [ ] Create a cancellation report in the database to cancellations that have already been processed (see TODO comments in code)
- [ ] Fix broken tests (for some reason I can't figure out the report isn't being found in the database via its UPRN)
- [ ] Test with real data once Bexley have done some test cancellations in Agile

Fixes https://github.com/mysociety/societyworks/issues/4799

### DIFF
--- a/t/script/bexley/cancel-garden-waste.t
+++ b/t/script/bexley/cancel-garden-waste.t
@@ -1,0 +1,148 @@
+use FixMyStreet::TestMech;
+use DateTime;
+use Test::Output;
+use Test::MockModule;
+use JSON::MaybeXS;
+
+use_ok 'FixMyStreet::Script::Bexley::CancelGardenWaste';
+
+my $mech = FixMyStreet::TestMech->new;
+my $area_id = 2494;
+my $body = $mech->create_body_ok($area_id, 'Bexley Council', {
+    cobrand => 'bexley',
+});
+
+# Create test contacts
+my $garden_contact = $mech->create_contact_ok(
+    category => 'Garden Subscription',
+    body_id => $body->id,
+    email => 'garden@example.com'
+);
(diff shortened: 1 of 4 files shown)


t/script/bexley/cancel-garden-waste.t
```diff
@@ -164,7 +163,7 @@ FixMyStreet::override_config {
                 { name => 'payment_method', value => 'direct_debit' },
                 { name => 'customer_external_ref', value => 'AGILE_CUSTOMER_REF' },
                 { name => 'direct_debit_reference', value => 'DD_REF_123' },
-                { name => 'reason', value => 'Cancelled on Agile end: Bins are bad' },
+                { name => 'reason', value => 'Cancelled on Agile end: No reason provided' },
```
chrismytton: I think it would be good to have a test covering the second parameter to `cancel_by_uprn` for when a reason is provided (e.g. `cancel_by_uprn($uprn, 'Customer moved house')`) and then checking that cancellation report has the correct reason.

chrismytton (commented):
I can only comment, because it's my PR!

Would be good to add the test I've suggested, but otherwise looks good!

t/script/bexley/cancel-garden-waste.t
nephila-nacrea: I will note that if you are calling the script with an individual UPRN, we don't pick up the reason (because that would require looking up the cancelled contract by UPRN and there didn't seem to be a straightforward way to do that). Unless you know of a way?

PR #5592, opened 2025-07-29
Improve tablet breakpoints on map page

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3784
Replaces: https://github.com/mysociety/fixmystreet/pull/4566/

# FMS
## /around
https://github.com/user-attachments/assets/d9e14f46-c752-49ef-a5b9-d44a6a58c25e

## /report/556
https://github.com/user-attachments/assets/f89eb283-21f9-4a5d-b900-f06572528a45

## /admin/report_edit/101
https://github.com/user-attachments/assets/c0b074b3-393f-486a-bfc7-33841ab96773

# Highways
## /around

https://github.com/user-attachments/assets/8863268d-c8b0-46d3-a949-e6bbaba9810f

## /admin/report_edit/101

https://github.com/user-attachments/assets/fd934dae-6e05-45ef-8d82-a472cc50e41f

[skip changelog]

### DIFF
--- a/web/cobrands/highwaysengland/_colours.scss
+++ b/web/cobrands/highwaysengland/_colours.scss
@@ -58,9 +58,19 @@ $front-main-background: $color-he-grey-2;
 
 // NH want the inspector form to be wider
 $mappage-actions-sidebar-width: 24em;
-$mappage-actions-width: 36em;
+$mappage-actions-sidebar-width--medium: 24em;
+$mappage-actions-sidebar-width--large: 24em;
+$mappage-actions-sidebar-width--xlarge: 24em;
+
+$mappage-sidebar-width: 24em;
 $mappage-sidebar-width--medium: 24em;
+$mappage-sidebar-width--large:  24em;
+$mappage-sidebar-width--xlarge: 24em;
+
+$mappage-actions-width: 36em;
 $mappage-actions-width--medium: 36em;
+$mappage-actions-width--large:  36em;
+$mappage-actions-width--xlarge: 36em;
 
 $mobile-sticky-sidebar-button-menu-image: "menu-white";
 
--- a/web/cobrands/sass/_layout.scss
+++ b/web/cobrands/sass/_layout.scss
@@ -2,13 +2,34 @@
 
 $layout_front_stats_color: $primary !default;
 $mappage-header-height: 4em !default; // eg: might want this to equal outer height of #site-header on normal pages
-$mappage-sidebar-width: 29em !default;
+
+$screen-size--medium: 63em !default;
+$screen-size--large:  79em !default;
+$screen-size--xlarge:  90em !default;
+
 $mappage-sidebar-padding: 1em;
-$mappage-notes-width: 15em;
-$mappage-actions-width: 25em !default;
-$mappage-actions-sidebar-width: $mappage-sidebar-width !default;
-$mappage-sidebar-width--medium: 24em !default;
-$mappage-actions-width--medium: 20em !default;
+
+$mappage-sidebar-width: 25em !default;
+$mappage-sidebar-width--medium:  27em !default;
+$mappage-sidebar-width--large:  29em !default;
+$mappage-sidebar-width--xlarge: 30em !default;
+
+$mappage-notes-width: 15em !default;
+$mappage-notes-width--medium: 17em !default;
+$mappage-notes-width--large: 19em !default;
+$mappage-notes-width--xlarge: 21em !default;
+
+$mappage-actions-width: 18em !default;
+$mappage-actions-width--medium:  20em !default;
+$mappage-actions-width--large:  25em !default;
+$mappage-actions-width--xlarge: 27em !default;
+
+// These variables are the same as $mappage-sidebar-width, but it controls the width when .with-action is present.
+$mappage-actions-sidebar-width: 23em !default;
+$mappage-actions-sidebar-width--medium: 25em !default;
+$mappage-actions-sidebar-width--large: $mappage-sidebar-width--large !default;
+$mappage-actions-sidebar-width--xlarge: $mappage-sidebar-width--xlarge !default;
+
 $header-top-border-width: 0.25em !default;
 $header-top-border: $header-top-border-width solid $primary !default;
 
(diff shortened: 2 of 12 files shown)


web/cobrands/sass/_govuk.scss
```diff
@@ -20,6 +20,9 @@
   font-size: 0.875rem;
   line-height: 150%;
   padding: flip(0.4em 1.5em 0.4em 0.6em, 0.4em 0.6em 0.4em 1.5em);
+  // Override the min and max inherited from GOVUK
```
dracos: I'm not sure what this is doing, given the commit message? The default is 11.5em/100%, so if it was only the 200px override that needed fixing, why are these changing too?

web/cobrands/sass/_report_list.scss
```diff
@@ -150,34 +150,24 @@ input.bulk-assign {
...
   padding: 1rem;
+  $grid-gap: 5%;
 
   .filter-group {
-    display: flex;
-    flex-direction: row;
-    gap: 1rem;
+    display: grid;
+    // Creates columns at least 180px wide, then all columns must be equally wide
+    grid-template-columns: repeat(auto-fill, minmax(180px, 1fr));
```
dracos, line 158: Do we want `auto-fit`? Otherwise if there's space for another 180px column they'll all shrink to that again, IIRC

web/cobrands/highwaysengland/_colours.scss
```diff
@@ -58,9 +58,19 @@ $front-main-background: $color-he-grey-2;
 
 // NH want the inspector form to be wider
 $mappage-actions-sidebar-width: 24em;
-$mappage-actions-width: 36em;
+$mappage-actions-sidebar-width--medium: 24em;
+$mappage-actions-sidebar-width--large: 24em;
+$mappage-actions-sidebar-width--large: 24em;
```
dracos: Is this one meant to be xlarge?
Although your videos looked okay?

web/cobrands/sass/_layout.scss
```diff
@@ -328,28 +426,67 @@ body.mappage.admin {
   // with the edge of .shadow-wrap, we give #side-report a fixed width,
   // and leave the other column to shrink to the remaining space.
   #side-report {
-    width: $mappage-sidebar-width;
```
dracos, line 331: I think for consistency this should still be present?
(Though I don't think anywhere on the site can have two_column_sidebar without with-actions, so it doesn't really matter.

web/cobrands/sass/_report_list.scss
lucascumsille: I tried using `auto-fit` at first and it works well when we have two elements per row. Unfortunately I run into some problems when the second row doesn't have the "show older reports", where the first child uses more space than it should. Adding`auto-fill` fixed that issue.

Screenshot 2025-08-12 at 09 22 25

web/cobrands/highwaysengland/_colours.scss
lucascumsille: I added a commit to add the `xlarge` =)

web/cobrands/sass/_layout.scss
lucascumsille: Let's go with consistency. I added the missing line.

lucascumsille:
Thanks for the feedback @dracos =)
Here is a list of the things I did with the commit fixups:

>Zurich doesn't display correctly, I'm afraid, I think that needs to remain at 27em for all possibilities, given its header and navigation (see pic), and I think its notes can always remain at 15em, they don't need to get wider.
I added a fixup for this.

> It seems odd to now have 20em sidebar width at 60em
I set it up at 25em now

> Was there a reason the defaults changed a bit for the normal view, e.g. a report page logged in as admin, currently at large width is 29/25, but in this PR is 28/22 at large or 30/23 at xlarge.

At the beginning I thought have it wider didn't make much sense with the examples I had because there wasn't that much information. But I just increased the size so it's wider considering all thereports with longer info and a narrow side bar will cause more vertical scrolling.

web/cobrands/sass/_layout.scss
```diff
@@ -2,13 +2,34 @@
...
+
+$screen-size--medium: 63em !default;
+$screen-size--large:  79em !default;
+$screen-size--xlarge:  90em !default;
+
 $mappage-sidebar-padding: 1em;
-$mappage-notes-width: 15em;
-$mappage-actions-width: 25em !default;
+
+$mappage-sidebar-width: 25em !default;
```
dracos, line 12: When there are actions, this does feel a tiny bit squashed at low widths.
Probably no-one really uses it like that, but perhaps 24em or 23em would be a happy medium?

dracos (changes requested):
One suggestion just to tweak the smallest (non-mobile) sidebar width a bit.
I can look at the broken JS tests, which are due to the fact they have to click on X,Y coordinate of the map, which has now moved...

lucascumsille:
@dracos thanks for all your help with this =). I have rebased to master and squashed the fixup commits. I added your commit to fix the test errors to this commit: [Improved responsiveness for #map_sidebar](https://github.com/mysociety/fixmystreet/pull/5592/commits/5a9c1239bea6a35204369e0a0aab66ff05669373)
Let me know if you are happy with having these 4 commits.

dracos (changes requested):
I tidied up the commit messages, but sorry, while doing that I've just spotted possibly one more thing - on an All Reports body page, the "Assign to..." dropdown menu (which is the only other place that uses govuk-fieldset-wrapper) now has a full width dropdown and the "Assign" button has wrapped to a new line, which looks okay I guess, but wondered if you'd realised/wanted to change that at all.

Before:

Image

After:

Image

Perhaps it should not be using govuk-fieldset-wrapper, or should be using it in a more consistent way with the other usage?

lucascumsille:
> I tidied up the commit messages, but sorry, while doing that I've just spotted possibly one more thing - on an All Reports body page, the "Assign to..." dropdown menu (which is the only other place that uses govuk-fieldset-wrapper) now has a full width dropdown and the "Assign" button has wrapped to a new line, which looks okay I guess, but wondered if you'd realised/wanted to change that at all.
> 
> Before:
> 
> Image
> After:
> 
> Image
> Perhaps it should not be using govuk-fieldset-wrapper, or should be using it in a more consistent way with the other usage?

I gave them the same treatment as the select items above. So I wrapped the select and button with a `filter-group` div so it aligns with the rest. Let me know what you think and if you'd like me to create a new class with a better naming technically these two are not filters, but I guess they are in way.

Screenshot 2025-08-27 at 07 22 36

PR #5598, opened 2025-08-06
[Open311] Skip deleted contacts updating category.

dracos:
[skip changelog] For FD-5817, should fix the Call Back/Other issue there


davea (approved):
🆒

PR #5600, opened 2025-08-07
[Northumberland] Add aerial mapping layer.

dracos:
Tidying up a little now Bing aerial has gone, but hopefully then only changing Northumberland and nothing else.
The tilma set up is already there, added to the mapcache config in servers (this relies on two files existing in /data/vhost/tilma.../layers/northumberland/ called watermark.png and aerial.xml which contains the actual endpoint and metadata etc).
The aerial.xml file was generated following the instructions at https://mapserver.org/de/mapcache/sources.html#wmts-sources
The watermark was generated by me personally.
[skip changelog]


davea (approved):
✈️📸🗺️

PR #5601, opened 2025-08-07
[Peterborough] Mandatory phone number for bulky.

dracos:
For FD-5763.
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste.pm
+++ b/perllib/FixMyStreet/App/Form/Waste.pm
@@ -67,6 +67,9 @@ sub validate {
     $email->add_error('Please provide an email address')
         unless $email->is_inactive || $email->value || ($is_staff_user && !$staff_provide_email);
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/App/Form/Waste.pm
```diff
@@ -67,6 +67,9 @@ sub validate {
     $email->add_error('Please provide an email address')
         unless $email->is_inactive || $email->value || ($is_staff_user && !$staff_provide_email);
 
+    $email->add_error('Please provide a phone number')
```
nephila-nacrea: Shouldn't the error be attached to the $phone field?

nephila-nacrea (changes requested):

perllib/FixMyStreet/App/Form/Waste.pm
dracos: whoops!

PR #5604, opened 2025-08-11
[UK] Questionnaire fixes

dracos:
This should stop questionnaires being sent out on fixed reports that don't allow reopening/updates (so not much point). Few different areas - the bit that checks the cobrand's `updates_allowed` feature (and also changing that to mark the report as no questionnaire rather than just skip it), adding checking `reopening_disallowed` (in case it's on a category basis) (and also changing that because AFAICT that's public only, not staff?) [skip changelog]
Raised as FD-5775

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/DDProcessor.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/DDProcessor.pm
@@ -331,6 +331,7 @@ sub _duplicate_waste_report {
         state => 'unconfirmed',
         non_public => 1,
         cobrand_data => 'waste',
+        send_questionnaire => 0,
     });
 
     my @extra = map { { name => $_, value => $extra->{$_} } } keys %$extra;


chrismytton (approved):

perllib/FixMyStreet/Script/Questionnaires.pm
```diff
@@ -45,20 +45,18 @@ sub send_questionnaires_period {
...
-        next unless $cobrand->send_questionnaires;
-
-        # If cobrand doesn't allow users to
-        # update don't send questionnaires
-        next if $cobrand->deny_updates_by_user($row);
-
         # Cobrands can also override sending per row if they wish
         my $cobrand_send = $cobrand->call_hook('send_questionnaire', $row) // 1;
+        my $cobrand_deny = $cobrand->deny_updates_by_user($row);
+        my $no_reopen = $cobrand->reopening_disallowed($row, $row->user);
```
dracos: No need for $row->user here, which dates from when I was trying to keep the user but not require `$c->user`, but then it turned out none of the user stuff was needed.

PR #5606, opened 2025-08-12
[Zurich] Admin interface for managing hierarchical attributes

chrismytton:
[edit: This PR now also incorporates #5631 and #5635]

Interface allows creation, editing and deletion of attributes associated with a division. These attributes will then be added to the admin report screen to allow admins to update these attributes before the reports is reassigned to a subdivision.

Fixes https://github.com/mysociety/societyworks/issues/4973
Fixes https://github.com/mysociety/societyworks/issues/4974
Fixes https://github.com/mysociety/societyworks/issues/4975

## Screenshot

image

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
@@ -151,6 +151,49 @@ sub edit : Chained('body') : PathPart('') : Args(0) {
     return 1;
 }
 
+=head2 attributes
+
+Controller for managing hierarchical attributes (Zurich only).
+
+=cut
+
+sub attributes : Chained('body') : PathPart('attributes') : Args(0) {
+    my ($self, $c) = @_;
+
+    # Only for Zurich cobrand
+    unless ($c->cobrand->moniker eq 'zurich') {
+        $c->detach('/page_error_404_not_found', []);
+    }
+
+    unless ($c->user->has_permission_to('category_edit', $c->stash->{body_id})) {
+        $c->forward('check_for_super_user');
+    }
+
+    $c->forward('/auth/get_csrf_token');
+
+    my $body = $c->stash->{body};
+    my $attributes = $body->get_extra_metadata('hierarchical_attributes') || {};
+
+    # Initialize with defaults if empty
+    if (!keys %$attributes) {
+        $attributes = $c->cobrand->get_default_hierarchical_attributes();
+    }
+
+    if ($c->req->method eq 'POST') {
+        $c->forward('/auth/check_csrf_token');
+        $c->forward('update_hierarchical_attributes', [$body, $attributes]);
+    }
+
+    $c->stash->{hierarchical_attributes} = $attributes;
+    $c->stash->{template} = 'admin/bodies/attributes.html';
+
+    if (defined $c->flash->{status_message}) {
+        $c->stash->{updated} = $c->flash->{status_message};
+    }
+
+    return 1;
+}
+
 =head2 add_category
 
 Controller for adding a new category.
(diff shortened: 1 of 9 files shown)


templates/web/zurich/admin/bodies/body.html
```diff
@@ -39,6 +39,14 @@ <h2>[% loc('Add new category') %]</h2>
 
 [% END %]
 
+[% IF admin_type == 'super' AND body.parent AND NOT body.parent.parent # A division %]
```
dracos: Could include this in the same IF above?

templates/web/zurich/admin/bodies/attributes.html
```diff
@@ -0,0 +1,129 @@
...
+                                                    <option value="[% parent_id %]"
+                                                        [% ' selected' IF entry.parent_id == parent_id %]>
+                                                        [% parent_entry.name | html %]
+                                                    </option>
+                                                [% END %]
+                                            </select>
+                                        </td>
+                                    [% END %]
+                                    <td>
+                                        <label>
```
dracos: Adding a class="inline" here makes each row take up a bit less space

templates/web/zurich/admin/bodies/attributes.html
```diff
@@ -0,0 +1,129 @@
...
+                                    </td>
+                                </tr>
+                            [% END %]
+                        </tbody>
+                    </table>
+                [% ELSE %]
+                    <p><em>[% loc('No entries yet') %]</em></p>
+                [% END %]
+
+                <div class="add-new">
```
dracos: Rather than this separate div, if you always show the table and its header above, and have an extra row at the bottom that consists of these form fields, I think that looks quite nice:

Image

Something bit tidier than this? :)

```diff
diff --git a/templates/web/zurich/admin/bodies/attributes.html b/templates/web/zurich/admin/bodies/attributes.html
@@ -31,7 +31,6 @@
         <div class="admin-box">
             <h3>[% level_name %]</h3>

-                [% IF level_entries.keys.size > 0 %]
                     <table class="admin-table" cellspacing="0" cellpadding="2" border="1">
                         <thead>
                             <tr>
@@ -44,6 +43,7 @@
                             </tr>
                         </thead>
                         <tbody>
+                [% IF level_entries.keys.size > 0 %]
                             [% FOREACH id IN level_entries.keys.sort %]
                                 [% entry = level_entries.$id %]
                                 <tr [% IF entry.deleted %]class="is-deleted"[% END %]>
@@ -73,7 +73,7 @@
                                         </td>
                                     [% END %]
                                     <td>
-                                        <label>
+                                        <label class="inline">
                                             <input type="checkbox" name="[% level_name %]_[% id %]_deleted" value="1"
                                                 [% ' checked' IF entry.deleted %]>
                                             [% loc('Deleted') %]
@@ -81,27 +81,19 @@
                                     </td>
                                 </tr>
                             [% END %]
-                        </tbody>
-                    </table>
-                [% ELSE %]
-                    <p><em>[% loc('No entries yet') %]</em></p>
                 [% END %]
-
-                <div class="add-new">
-                    <h4>[% tprintf(loc('Add new %s'), level_name) %]</h4>
-                    <p>
-                        <label for="new_[% level_name %]_name">[% loc('Name') %]:</label>
+                <tr>
+                                    <td>New:</td>
+                                    <td>
                         <input type="text" id="new_[% level_name %]_name" name="new_[% level_name %]_name"
                                 size="30" class="form-control"
                                 [% IF errors.item("new_${level_name}_name") %]style="border-color: red;"[% END %]>
                         [% IF errors.item("new_${level_name}_name") %]
                             <div class="error-msg">[% errors.item("new_${level_name}_name") %]</div>
                         [% END %]
-                    </p>
-
-                    [% IF level.parent %]
-                        <label for="new_[% level_name %]_parent">[% level.parent %]:</label>
-                        <p>
+                                    </td>
+                                    [% IF level.parent %]
+                                        <td>
                             <select id="new_[% level_name %]_parent" name="new_[% level_name %]_parent" class="form-control"
                                     [% IF errors.item("new_${level_name}_parent") %]style="border-color: red;"[% END %]>
                                 <option value="">-- [% loc('Select parent') %] --</option>
@@ -115,9 +107,13 @@
                             [% IF errors.item("new_${level_name}_parent") %]
                                 <div class="error-msg">[% errors.item("new_${level_name}_parent") %]</div>
                             [% END %]
-                        </p>
-                    [% END %]
-                </div>
+                                        </td>
+                                    [% END %]
+                                    <td>
+                                    </td>
+                                </tr>
+                        </tbody>
+                    </table>

                 <div class="form-actions">
                     <input type="submit" value="[% loc('Update attributes') %]" class="btn btn-primary">
```

dracos (changes requested):
Looks good, only some small display suggestions :)

dracos (approved):
Woo, thanks!

dracos (changes requested):
~~Sorry, only spotted this while reviewing the other PR. They said in the quote the name for the 'sub' tables is non unique, which I took to mean that you could have two entries in e.g. Objekt with the same name but in different Geschäftsbereich. I thought I had tested this here, but it doesn't seem to work so perhaps I didn't - the "already exists" check here should only be unique per 'parent' if there's a parent, I guess fixes it?~~ Done in fixup

perllib/FixMyStreet/App/Controller/Admin/Bodies.pm
```diff
@@ -618,6 +661,166 @@ sub update_translations : Private {
...
+                    next;
+                }
+
+                my $duplicate = 0;
+                foreach my $existing_id (keys %{$level->{entries}}) {
+                    my $existing_entry = $level->{entries}->{$existing_id};
+                    if (!$existing_entry->{deleted} && lc($existing_entry->{name}) eq lc($new_name)) {
+                        $errors{$name_param} = _('An entry with this name already exists');
+                        $duplicate = 1;
+                        last;
```
dracos: For info, you can put a label on the outer loop, `LEVEL: foreach my $level_name...` and then here you could put `next LEVEL` and not need the `$duplicate` variable.

chrismytton (commented):
Nice, thanks! 👍

dracos:
Unneeded

PR #5609, opened 2025-08-14
Update submit emails to account for category display names.

neprune:
[skip changelog]

Motivated by https://mysocietysupport.freshdesk.com/a/tickets/5860

### DIFF
--- a/templates/email/fixmystreet.com/submit.html
+++ b/templates/email/fixmystreet.com/submit.html
@@ -52,7 +52,7 @@ of a local problem that they believe might require your attention.</p>
 
     <p style="[% secondary_p_style %]"><strong>Report ID:</strong> [% report.id %]</p>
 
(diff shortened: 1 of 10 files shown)


templates/email/fixmystreet.com/submit.html
```diff
@@ -52,7 +52,7 @@ <h2 style="[% h2_style %]">[% report.title | html %]</h2>
 
     <p style="[% secondary_p_style %]"><strong>Report ID:</strong> [% report.id %]</p>
 
-    <p style="[% secondary_p_style %]"><strong>Category:</strong> [% report.category | html %]</p>
+    <p style="[% secondary_p_style %]"><strong>Category:</strong> [% report.category_display | html %]</p>
```
dracos: Don't need the `| html` any more (for long time now, it's automatic), didn't seem worth going through and removing them but you can do so when you're changing them anyway.

PR #5611, opened 2025-08-15
[Bulky] Repopulate items on Back, and have add work without JS

dracos:
[skip changelog]

### DIFF
--- a/web/js/waste.js
+++ b/web/js/waste.js
@@ -275,6 +275,19 @@ $(function() {
             display_band_pricing();
         });
 
+        $('#item-selection-form').on('submit', function(){
+            var items = [];
+            $('.govuk-select[name^="item_"] option:selected').each(function(i, e) {
+                var v = $(this).val();
+                if (v) {
+                    items.push(v);
+                }
+            });
+            try {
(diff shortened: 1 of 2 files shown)


web/js/waste.js
```diff
@@ -292,24 +303,39 @@ $(function() {
...
-                // to spot changes to the value
-                setTimeout(function() {
-                    $wrapper.find('.autocomplete__wrapper input').val(value);
-                }, 110);
+        if (!template) { return; } // Only bulky items page
+
+        var i;
+        // If on page load we have some stored items to re-show, as browser does not
+        var items = JSON.parse(sessionStorage.getItem('bulky_items'));
+        if (!items) { return; }
```
nephila-nacrea, line 315: Should this be an `items.length` check like below? What is the exact difference between checking for `items` and `items.length`?

dracos: If the key doesn't exist in storage, `getItem` will return null, which will become JS `null`. If you tried to call `.length` on null, you'd throw an error. Given we're the only ones who set this item, we can assume if it's defined at all, it'll be an array with a length (though I guess for completeness it's possible it could not be, but think it's okay).

PR #5612, opened 2025-08-15
[Sutton][Small Items] Add email notifications for booking

MorayMySoc:
Handles checking for small items/bulky in category for correct text in Sutton templates.

Creates new attribute for free small items collection, defaulting to non-free.

When sending reminders for booking, if skipping reminders until payment is received, then also check if the report itself is marked as free so reminder can be sent.

https://github.com/mysociety/societyworks/issues/4921

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Sutton.pm
+++ b/perllib/FixMyStreet/Cobrand/Sutton.pm
@@ -790,4 +790,6 @@ sub waste_munge_small_items_data {
     $self->waste_munge_bulky_data($data);
 }
 
+sub small_items_free_collection_available { 1 };
+
 1;
--- a/perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
@@ -50,6 +50,8 @@ sub small_items_enabled {
     return $cfg->{small_items_enabled};
 }
 
+sub small_items_free_collection_available { 0 };
+
 sub bulky_items_master_list { $_[0]->wasteworks_config->{item_list} || [] }
 sub small_items_master_list { $_[0]->wasteworks_config->{small_item_list} || [] }
 sub bulky_per_item_costs { $_[0]->wasteworks_config->{per_item_costs} }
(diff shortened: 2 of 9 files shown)


templates/email/kingston/waste/bulky-reminder.html
```diff
@@ -28,6 +28,13 @@
     [% END %]
   </p>
 
+  [% IF cobrand.moniker == 'sutton' AND report.title == 'Small items collection' %]
+  <p style="[% p_style %]">
+    Items to be collected:
+    [% INCLUDE 'waste/_bulky_list.txt' %]
```
dracos: This should be .html not .txt.

@jacquelinebylau It's a bit annoying that Sutton do not have item lists in their reminder emails for bulky collections - presumably they specifically asked for them not to be included, given it is in there by default - but we included them in the small items technical spec.

perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
```diff
@@ -203,7 +205,10 @@ sub bulky_total_cost {
     my ($self, $data) = @_;
     my $c = $self->{c};
 
-    if ($self->bulky_free_collection_available) {
+    if (
+        (!$c->stash->{small_items} && $self->bulky_free_collection_available)
+        || ($c->stash->{small_items} && $self->small_items_free_collection_available)
+    ) {
         $data->{extra_CHARGEABLE} = 'FREE';
```
dracos: This is only like this because Peterborough had a field named this for sending to Bartec for free collections.
Could we just have the reminders function check the category for now?
I think in future if Sutton introduce charging, it'd be easier to check if there's `payment` on the row rather than having to check a different field.

MorayMySoc:
694099afa5b010268f101696c12b861e8a9b6b9d is the rebase fixup

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -803,4 +803,6 @@ sub waste_munge_small_items_data {
     $self->waste_munge_bulky_data($data);
 }
 
+sub small_items_free_collection_available { 1 };
```
dracos: Don't need this any more (and if you did, you'd need it in Brent as well :) )

dracos (changes requested):
One more fn to be removed is all :)

dracos (approved):

PR #5616, opened 2025-08-18
[Bexley] Improve Agile API error handling for garden waste

chrismytton:
Differentiate between 400/404 error, which are expected when there's no subscription for a property, and other errors which indicate an unexpected problem.

When an unexpected problem occurs show the user a helpful message and remove garden bins from the listing page, as they won't have the correct data available.

Also adds some debugging info to the superuser debugging panel.

Related to FD-5877.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
@@ -48,9 +48,13 @@ sub lookup_subscription_for_uprn {
     my ( $customer, $contract );
 
     my $results = $self->agile->CustomerSearch($uprn);
-    # 'error' will usually be 404, maybe 400; we can't guarantee there aren't
-    # other possible values
-    return if $results->{error};
+    if ($results->{error}) {
+        # Unexpected error - API problem
+        return {
+            error => $results->{error},
+            error_message => $results->{error_message},
+        };
+    }
 
     # find the first 'ACTIVATED' Customer with an 'ACTIVE'/'PRECONTRACT' contract
     my $customers = $results->{Customers} || [];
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -48,9 +48,20 @@ sub lookup_subscription_for_uprn {
     my ( $customer, $contract );
 
     my $results = $self->agile->CustomerSearch($uprn);
-    # 'error' will usually be 404, maybe 400; we can't guarantee there aren't
-    # other possible values
-    return if $results->{error};
+    # 404 and 400 are expected responses indicating no subscription found
+    # Other errors (like 503) indicate API problems
+    if ($results->{error}) {
+        if ($results->{error} eq '404' || $results->{error} eq '400') {
```
dracos: I think if I read https://github.com/mysociety/societyworks/issues/4882 right we should treat 400 the same as unexpected error.

chrismytton: :+1: I tried a sample of 100 UPRNs and none of them returned a 404 error, so I think treating both 400 and 404 as an unexpected error probably makes sense. Have added a fixup commit with that change.

dracos: Yeah, I think from looking at the history, 404 was returned for a property that we had in our list of properties but Agile was missing due to their list being a bit out of date. In that situation, presumably there's no point letting someone sign up, because Agile wouldn't recognise the UPRN anyway when it tried to be submitted (I assume? Perhaps it would allow sign up. But hopefully doesn't happen anyway.)

dracos (approved):

PR #5620, opened 2025-08-19
Update README.md to remove fixmystreet.ie from examples

trevorparsons:
fixmystreet.ie unfortunately no longer belongs in the list of examples of fixmystreet implementations. 

The service was closed down two or three years ago, and the domain has since been squatted.

This is how the Irish Independent covered the closure:
https://www.independent.ie/regionals/dublin/dublin-news/website-closing-down-after-11-years-of-helping-solve-dublins-broken-paths-and-street-lights/41852882.html

### DIFF
--- a/docs/sites.md
+++ b/docs/sites.md
@@ -8,22 +8,6 @@ title: Sites running on the FixMyStreet platform
 FixMyStreet works everywhere! Take a look at how FixmyStreet is being
 used in cities and countries across the world.
 
-<h2>Americas</h2>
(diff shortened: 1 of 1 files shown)


dracos (approved):
Thanks for the PR, a shame the site is no longer running. Note that Irish Independent article is not about this site, but about fix**your**street which was a different service. I've updated the commit message, and taken the opportunity to update the sites page as well to bring that up to date.

PR #5621, opened 2025-08-20
[Sutton][Small Items] Adds tests for Small items cancellations

MorayMySoc:
https://github.com/mysociety/societyworks/issues/4922

[skip changelog]

### DIFF
--- a/t/app/controller/waste_sutton_small_items.t
+++ b/t/app/controller/waste_sutton_small_items.t
@@ -3,6 +3,7 @@ use Test::MockTime qw(:all);
 use FixMyStreet::TestMech;
 use Path::Tiny;
 use FixMyStreet::Script::Reports;
+use FixMyStreet::Script::Alerts;
 
 FixMyStreet::App->log->disable('info');
 END { FixMyStreet::App->log->enable('info'); }
@@ -57,6 +58,9 @@ create_contact(
 );
 
 my $user = $mech->create_user_ok('maryk@example.org', name => 'Test User');
+my $staff = $mech->create_user_ok('staff@example.org', name => 'Staff User', from_body => $sutton->id);
+$staff->user_body_permissions->create({ body => $sutton, permission_type => 'report_edit' });
+$sutton->update( { comment_user_id => $staff->id } );
 
 FixMyStreet::override_config {
     MAPIT_URL => 'http://mapit.uk/',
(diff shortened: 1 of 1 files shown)


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -738,6 +738,7 @@ sub bulky_allowed_property {
 
 sub bulky_collection_time { { hours => 6, minutes => 0 } }
 sub bulky_cancellation_cutoff_time { { hours => 6, minutes => 0, days_before => 0 } }
+sub small_items_cancellation_cutoff_time { { hours => 23, minutes => 59, days_before => 0 } }
```
dracos: I don't understand why this isn't the same as bulky? Collections start at 6 so no point in cancelling after that.

perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
```diff
@@ -589,7 +592,12 @@ sub bulky_nice_cancellation_cutoff_time {
 
 sub _bulky_cancellation_cutoff_date {
     my ($self, $collection_date) = @_;
-    return _bulky_time_object_to_datetime($collection_date, $self->bulky_cancellation_cutoff_time());
+
+    my $fn =
```
dracos: The unit tests are presumably failing because they call these low level functions as static methods on the class, rather than instances, so $self->{c} won't/can't exist. This does feels like a too low level place to be testing the bulky/small items, I realise code is shared but something right down here shouldn't be deciding between bulky/shared, I think, we'd have to move that decision higher in the calling chain and then pass in bulky/small items, I think.

Although given my other comment, I don't think this is needed at all - I think this should, until/unless we hear otherwise, have the same cutoff as bulky, I don't see why they would differ, and this could all be ignored for now and only have the test :)

dracos (changes requested):

t/app/controller/waste_sutton_small_items.t
```diff
@@ -329,8 +333,68 @@ FixMyStreet::override_config {
...
+                $mech->clear_emails_ok;
+            };
+            # Set report active again
+            $report->state('confirmed');
+            my $detail = $report->detail;
+            $detail =~ s/ \| Cancelled at user request//;
+            $report->detail($detail);
+            $report->update;
+            $comment->delete;
+            $comment->update;
```
dracos: Don't think you need an update after a delete, not sure what that does :)

dracos (approved):

PR #5623, opened 2025-08-21
"Add items for collection": layout and styling improvements

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5079

This PR adds:

- `govuk-button--secondary` to "add item" button
- `govuk-button--warning` to "Delete item"
- Deleted `margin-top` for "Delete item". `govuk-form-group` which is the previous element already has `margin-bottom`
- Added extra vertical margin to `hr` elements to improve distance between the items to collect. 

### Desktop
https://github.com/user-attachments/assets/33f7cb57-7653-4ab5-a0dc-5304a7852aea

### Mobile
https://github.com/user-attachments/assets/237da0bf-cbfa-4969-b184-781c3d13e8cf

[skip changelog]

### DIFF
--- a/templates/web/base/waste/bulky/items.html
+++ b/templates/web/base/waste/bulky/items.html
@@ -118,7 +118,7 @@ END; END %]
   [% IF c.cobrand.bulky_pricing_strategy.strategy == 'banded' %]
   <p id="band-pricing-info"></p>
   [% END %]
-  <button type="submit" id="add-new-item" class="btn-secondary govuk-!-margin-bottom-3" aria-label="Add item" name="goto-same-page" value="1">
+  <button type="submit" id="add-new-item" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-3" aria-label="Add item" name="goto-same-page" value="1">
     [% IF c.cobrand.moniker == 'kingston' OR c.cobrand.moniker == 'sutton' ~%]
       Add another item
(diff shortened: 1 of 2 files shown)


lucascumsille:
@dracos about what you mentioned yesterday:

>On their spacing suggestions, I agree, but it makes it more scroll-y. I was wondering if when “Add item” is tapped, it automatically scrolls down a bit to the new form so you don’t have to scroll

https://github.com/user-attachments/assets/76f5cd25-f730-461a-a133-2a14c1d3dd6f

We currently do that, I tested it on the staging site on my phone. However what I found was an issue with the "Add item" button because, depending on where you are in the page you might not see it. So I added a commit that makes the button `sticky` when is not disabled.

https://github.com/user-attachments/assets/bd041305-2abd-4bd3-905b-5f0d0bbf319e

Let me know what you think.

jacquelinebylau:
@dracos Would you be able to review this? It would be good to wrap it up this week :)

dracos (approved):
Thanks; I do agree we could make it nicer on mobile with something like this, so that'd be good :)

lucascumsille:
Thanks @dracos, I like your suggestion:

> (I'd perhaps suggest we make Add item and continue both in a sticky bar at the bottom, half and half, so people can add another item or easily continue when they're done

I'll see what I can do

PR #5624, opened 2025-08-21
[Sutton] Missed collections for small items

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4925 .

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -840,21 +840,30 @@ sub construct_bin_report_form {
         };
     }
 
-    # XXX Should we refactor bulky into the general service data (above)?
+    # XXX Should we refactor bulky & small items into the general service
+    # data (above)?
     # Plus side, gets the report missed stuff built in; minus side it
-    # doesn't have any next/last collection stuff which is assumed
+    # doesn't have any next/last collection stuff which is assumed.
     my $allow_report_bulky = 0;
-    foreach (values %{ $c->stash->{booked_missed} || {} }) {
-        $allow_report_bulky = $_ if $_->{report_allowed} && !$_->{report_open};
+    my $allow_report_small_items = 0;
+
+    foreach ( values %{ $c->stash->{booked_missed} || {} } ) {
+        if ( $_->{report_allowed} && !$_->{report_open} ) {
+            $_->{service_name} eq 'Small items'
+                ? $allow_report_small_items = $_
+                : $allow_report_bulky = $_;
+        }
     }
-    if ($allow_report_bulky) {
-        my $service_id = $allow_report_bulky->{service_id};
-        my $service_name = $allow_report_bulky->{service_name};
-        push @$field_list, "service-$service_id" => {
-            type => 'Checkbox',
-            label => "$service_name collection",
-            option_label => "$service_name collection",
-        };
+    for ( $allow_report_bulky, $allow_report_small_items ) {
+        if ($_) {
+            my $service_id = $_->{service_id};
+            my $service_name = $_->{service_name};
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -865,6 +874,11 @@ sub construct_bin_report_form {
 sub report : Chained('property') : Args(0) {
     my ($self, $c) = @_;
 
+    $c->stash->{original_booking_report}
+        = FixMyStreet::DB->resultset("Problem")
+        ->find( { id => $c->get_param('original_booking_id') } )
+        if $c->get_param('original_booking_id');
```
dracos, line 880: I guess doing this means the `$booking_report` look up in `waste_munge_report_data` in SLWP2 (and Brent, no need to do Merton as they're moving to SLWP2 shortly) isn't needed, it can use this stage variable?

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -119,17 +119,20 @@ sub waste_munge_report_form_fields {
...
+bulky or small items, where we ask for more information.
 
 =cut
 
 sub waste_report_form_first_next {
     my $self = shift;
     my $cfg = $self->feature('echo');
-    my $bulky_service_id = $cfg->{bulky_service_id};
+    my $bulky_or_small_service_id
+        = $cfg->{bulky_service_id} || $cfg->{small_items_service_id};
```
dracos: Sutton already has bulky, so this will always be bulky - so I think this won't do what you're after for a small items collection, because bulky won't be ticked? The reason the test works if because it's not setting bulky_service_id.

templates/web/base/waste/summary_report.html
```diff
@@ -1,5 +1,9 @@
 [%
-title = c.req.params.original_booking_id ? 'Submit missed bulky collection' : 'Submit missed bin report';
+title = original_booking_report.category == 'Bulky collection'
```
dracos, line 2: Nice

perllib/FixMyStreet/Roles/Cobrand/SLWP2.pm
```diff
@@ -317,12 +319,19 @@ sub waste_munge_report_data {
...
     }
+    if (   $cfg->{small_items_service_id}
+        && $id == $cfg->{small_items_service_id} )
+    {
+        $service = 'small items collection';
+    }
     if ($c->get_param('additional') && $c->stash->{is_staff}) {
         $data->{category} = 'Request additional collection';
-        $data->{title} = "Request additional $service collection";
+        $data->{title} = "Request additional $service";
```
dracos: Not sure this is needed - the normal service names don't have "collection" in them, so this is doing the right thing for an additional collection (which isn't bulky/small items).

perllib/FixMyStreet/App/Controller/Waste.pm
nephila-nacrea: True; that is where I copied the DB lookup from.

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
nephila-nacrea: Ah, testing manually it doesn't show the 'notes' page if I select the small items checkbox, gotcha.

perllib/FixMyStreet/Roles/Cobrand/SLWP2.pm
nephila-nacrea: Ah, I hadn't twigged that additional collections never included bulky/small.

dracos (approved):
Thanks :)

PR #5625, opened 2025-08-21
[Waste] PWA installation improvements

davea:
- Adds new `/about/waste-web-app` page with WW-tailored instructions
 - Updates web manifest used on property pages to set `start_url` to the property page
 - Adds link to new help page from property page and allows PWA installation for that property from there

For https://github.com/mysociety/societyworks/issues/4962
For https://github.com/mysociety/societyworks/issues/3993

### DIFF
--- a/templates/web/fixmystreet-uk-councils/about/web-app.html
+++ b/templates/web/fixmystreet-uk-councils/about/web-app.html
@@ -2,9 +2,9 @@
 
 <h1>[% c.cobrand.council_name %] reporting app</h1>
 
-<p>The [% c.cobrand.council_name %] reporting service is powered by FixMyStreet Pro, which is a progressive web app (PWA). This means that you can save this website to your mobile’s home screen and use it just like an app. See below for instructions on how to do this on both iOS and Android devices.</p>
+<p>The [% c.cobrand.council_name %] reporting service is powered by [% IF wasteworks %]WasteWorks[% ELSE %]FixMyStreet Pro[% END %], which is a progressive web app (PWA). This means that you can save this website to your mobile’s home screen and use it just like an app. See below for instructions on how to do this on both iOS and Android devices.</p>
 
-<p>We chose to develop a PWA instead of a dedicated app because it gives you the choice to use this service on whatever device you want, and ensures that you always benefit from the same user experience, including having access to offline reporting functionality. It also means that we can introduce new features and updates more quickly, while also keeping maintenance costs low due to only having to support one codebase.</p>
+<p>We chose to develop a PWA instead of a dedicated app because it gives you the choice to use this service on whatever device you want, and ensures that you always benefit from the same user experience[% UNLESS wasteworks %], including having access to offline reporting functionality[% END %]. It also means that we can introduce new features and updates more quickly, while also keeping maintenance costs low due to only having to support one codebase.</p>
 
 <h2>How to install [% site_name %] as an app</h2>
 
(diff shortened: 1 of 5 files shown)


templates/web/fixmystreet-uk-councils/about/web-app.html
```diff
@@ -31,9 +31,11 @@ <h2>How to install [% site_name %] as an app</h2>
 <h3>iOS (Apple iPhones/iPads)</h3>
 <img alt="" class="web-app-install__scrshot" srcset="/cobrands/fixmystreet-uk-councils/web-app/small/iphone-[% c.cobrand.moniker %]_framed.png, /cobrands/fixmystreet-uk-councils/web-app/large/iphone-[% c.cobrand.moniker %]_framed.png 2x" src="/cobrands/fixmystreet-uk-councils/web-app/large/iphone-[% c.cobrand.moniker %]_framed.png">
 <ol>
-    <li>Load [% c.uri_for('/') %] on your mobile browser (if you have iOS version 16.3 or less, the installation will only work via Safari)</li>
+    [% UNLESS wasteworks AND property_id %]
+        <li>Load [% IF wasteworks %]your property's bin days page[% ELSE %][% c.uri_for('/') %][% END %] on your mobile browser</li>
```
dracos: Could we have curly apostrophes/quotes please

templates/web/fixmystreet-uk-councils/about/web-app.html
```diff
@@ -31,9 +31,11 @@ <h2>How to install [% site_name %] as an app</h2>
...
+    [% END %]
     <li>
-        Hit the save/share button
+        Hit the save/share button[% IF wasteworks AND property_id %] on this page or your property's bin days page[% END %]
```
dracos: You've just said "Load your property’s bin days page" in the previous line, so does this line need "or your property’s bin days page"?

templates/web/fixmystreet-uk-councils/about/web-app.html
```diff
@@ -50,8 +52,10 @@ <h3>iOS (Apple iPhones/iPads)</h3>
 <h3>Android</h3>
 <img alt="" class="web-app-install__scrshot" srcset="/cobrands/fixmystreet-uk-councils/web-app/small/android-[% c.cobrand.moniker %]_framed.png, /cobrands/fixmystreet-uk-councils/web-app/large/android-[% c.cobrand.moniker %]_framed.png 2x" src="/cobrands/fixmystreet-uk-councils/web-app/large/android-[% c.cobrand.moniker %]_framed.png">
 <ol>
-    <li>Load [% c.uri_for('/') %] on your mobile browser</li>
-    <li>Press the ‘three dots’ icon to open the browser menu
+    [% UNLESS wasteworks AND property_id %]
+        <li>Load [% IF wasteworks %]your property's bin days page[% ELSE %][% c.uri_for('/') %][% END %] on your mobile browser</li>
+    [% END %]
+    <li>Press the ‘three dots’ icon to open the browser menu[% IF wasteworks AND property_id %] on this page or your property's bin days page[% END %]
```
dracos: ditto

dracos (commented):
Couple of tiny things in the text.
Couple of broader things to consider (might have been considered):
* Bexley/Peterborough have a "Change address" link by the address - do we want to add that everywhere, so people using this can more easily change the address should they ever need to? As otherwise I don't think there is any real way to get from a bin days page to the new lookup. Although, it does make the PWA feel more "property"-based.
* Similarly, if someone wants to change the stored property, as it's stored in the start URL presumably they have to delete and re-add the app. I guess if we're associating the PWA with a property, that sort of makes sense, and won't be very often.

templates/web/fixmystreet-uk-councils/about/web-app.html
davea: The previous line is in an `UNLESS` but this is in an `IF`, so they will never be shown at the same time. I think?

dracos: Oh, yep, missed that bit, sorry!

PR #5626, opened 2025-08-21
[Waste] Bulky/small items missed collection skips selection page

nephila-nacrea:
Reporting a bulky or small items collection as missed relies on having an 'original_booking_id' parameter. Normally, the missed collection form starts with a list of checkboxes so multiple container types/collections can be selected, but this does not make much sense for bulky/small, as the 'original_booking_id' implies that the flow is only dealing with one collection.

Closes https://github.com/mysociety/societyworks/issues/5116 .

[skip changelog]

### DIFF
--- a/t/app/controller/waste_merton_bulky.t
+++ b/t/app/controller/waste_merton_bulky.t
@@ -988,7 +988,7 @@ FixMyStreet::override_config {
         $mech->submit_form_ok( { form_number => 1 }, 'Go back to first page' );
         $mech->content_contains('About you');
         $mech->submit_form_ok({ form_number => 1 });
-        $mech->submit_form_ok( { form_number => 3 }, 'Submit summary' );
+        $mech->submit_form_ok( { form_number => 2 }, 'Submit summary' );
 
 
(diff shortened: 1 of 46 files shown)


t/app/controller/waste_merton_bulky.t
```diff
@@ -977,12 +977,19 @@ FixMyStreet::override_config {
...
+        $mech->submit_form_ok(
+            { with_fields => { 'service-413' => 1 } },
+            "Follow link for reporting a missed bulky collection"
+        );
+        $mech->content_contains('About you');
         $mech->submit_form_ok({ form_number => 1 });
+        $mech->content_contains('Submit missed bin report');
         #$mech->submit_form_ok({ with_fields => { extra_detail => "They left the mattress" } });
+        $mech->submit_form_ok( { form_number => 1 }, 'Go back to first page' );
+        $mech->content_contains('About you');
```
nephila-nacrea, line 989: Merton and Brent will now only have the 'about you' page to change answers for, so I ought to make it so only one 'change answers' button is shown on summary page.

PR #5629, opened 2025-08-26
[Sutton][Small Items] Adds amend for Small items for use by Sutton

MorayMySoc:
https://github.com/mysociety/societyworks/issues/5056

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -10,6 +10,7 @@ use FixMyStreet::App::Form::Waste::Bulky::Amend;
 use FixMyStreet::App::Form::Waste::Bulky::Cancel;
 use FixMyStreet::App::Form::Waste::SmallItems;
 use FixMyStreet::App::Form::Waste::SmallItems::Cancel;
+use FixMyStreet::App::Form::Waste::SmallItems::Amend;
 
 has feature => (
     is => 'ro',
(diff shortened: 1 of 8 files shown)


perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -214,11 +215,17 @@ sub index_booking : Private {
 sub amend : Chained('setup') : Args(1) {
     my ($self, $c, $id) = @_;
 
+    my $small = $c->stash->{small_items};
     $c->stash->{first_page} = 'intro';
-    $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Bulky::Amend';
+    $c->stash->{form_class} = $small
```
dracos: You already set fom_class in amend_small - so either you don't do that, or this changes to be a `||=` to only be set if not already set.

dracos (changes requested):
I think there's a couple other templates that might need adjusting - base/waste/bulky/summary.html and base/waste/bulky/_bin_days_list.html both have amend links/checks in, but think that's it beyond one thing

MorayMySoc:
I think summary.html looks like it takes care of itself by  processing _form_action.html, so I've just confirmed with a  test if I understood correctly. But  updated _bin_days_list.html

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -214,11 +215,17 @@ sub index_booking : Private {
 sub amend : Chained('setup') : Args(1) {
     my ($self, $c, $id) = @_;
 
+    my $small = $c->stash->{small_items};
     $c->stash->{first_page} = 'intro';
-    $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Bulky::Amend';
+    $c->stash->{form_class} ||= $small
```
dracos: You've already set it to SmallItems::Amend below if it's small items.
This can just be `$c->stash->{form_class} ||= 'FixMyStreet::App::Form::Waste::Bulky::Amend';`

dracos (changes requested):
> I think summary.html looks like it takes care of itself

There are three other uses of `waste/bulky/amend` in summary.html - two deciding whether to show user details or not (so also shouldn't be shown for amend_small I assume), and the third providing an "Amend this booking" link on a existing bulky page to the user - so won't work on a small items page I assume.

MorayMySoc:
Oh, yes, sorry! Hopefully made the right changes this time!

PR #5635, opened 2025-08-29
[Zurich] Add additional attributes to admin report edit form

chrismytton:
Adds a hierarchical attributes section to the admin report edit page for selecting the required attributes for a report based on the division.

Attributes are validated on the client and server side, and are required to be set for any report that's not in a submitted (Erfasst) state.

image

Initially, the second and third drop-downs are disabled until something has been selected from the first drop-down.

image

Also add a read-only view of the attributes to the subdivision manager's report edit view:

image

Fixes https://github.com/mysociety/societyworks/issues/4974

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -611,6 +611,19 @@ sub admin_report_edit {
 
     }
 
+    # Fetch hierarchical attributes for the current body (for all user types)
+    my $current_body = $c->model('DB::Body')->find($problem->bodies_str);
+    if ($current_body) {
+        my $hierarchical_attributes = $current_body->get_extra_metadata('hierarchical_attributes') || {};
+        if (!keys %$hierarchical_attributes) {
+            $hierarchical_attributes = $c->cobrand->get_default_hierarchical_attributes();
+        }
+        $c->stash->{hierarchical_attributes} = $hierarchical_attributes;
+
+        my $selected_attributes = $problem->get_extra_metadata('hierarchical_attributes') || {};
+        $c->stash->{selected_hierarchical_attributes} = $selected_attributes;
+    }
+
     # If super or dm check that the token is correct before proceeding
     if ( ($type eq 'super' || $type eq 'dm') && $c->get_param('submit') ) {
         $c->forward('/auth/check_csrf_token');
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -639,6 +652,36 @@ sub admin_report_edit {
...
+            }
+
+            if (%hierarchical_errors) {
+                $c->stash->{hierarchical_errors} = \%hierarchical_errors;
+                $c->stash->{status_message} = '<p class="message-error">' . _('Please select all hierarchical attributes before saving') . '</p>';
+                return $self->admin_report_edit_done;
+            }
+
+            $problem->set_extra_metadata('hierarchical_attributes', {
+                geschaftsbereich => $geschaftsbereich,
```
dracos: Why not use the same keys as in the admin data? That would make the template display bits easier to put in loops then.

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -639,6 +652,36 @@ sub admin_report_edit {
     # Problem updates upon submission
     if ( ($type eq 'super' || $type eq 'dm') && $c->get_param('submit') ) {
 
+        # Validate hierarchical attributes if problem state is not 'submitted'
```
dracos, line 661: Whilst it seems unlikely to be abused or anything, this doesn't actually validate that the things passed in are valid values, only that they exist.

templates/web/zurich/admin/report_edit-sdm.html
```diff
@@ -79,6 +79,24 @@
      </dd>
   [% END %]
 
+    [% IF problem.state != 'submitted' AND hierarchical_attributes AND selected_hierarchical_attributes %]
+        [% SET selected = selected_hierarchical_attributes %]
+        [% IF selected.geschaftsbereich OR selected.objekt OR selected.kategorie %]
+            <dt><span class="mock-label">[% loc('Hierarchical Attributes:') %]</span></dt>
+            <dd>
+                [% IF selected.geschaftsbereich AND hierarchical_attributes.Geschäftsbereich.entries.${selected.geschaftsbereich} %]
```
dracos: Similarly (to the main template), this feels like it could be a `[% FOR field in ['Geschäftsbereich', 'Objekt', 'Kategorie']` - if the data is stored with those three names as well, especially.

templates/web/zurich/admin/report_edit-sdm.html
```diff
@@ -79,6 +79,24 @@
...
+                    <strong>[% loc('Geschäftsbereich:') %]</strong> [% hierarchical_attributes.Geschäftsbereich.entries.${selected.geschaftsbereich}.name | html %]<br>
```
dracos: You don't need to `| html` escape parameters, it happens by default.

templates/web/zurich/admin/reports/edit.html
```diff
@@ -152,6 +152,71 @@
...
     </dd>
 
+    [% IF problem.state != 'submitted' AND hierarchical_attributes %]
+        <dt class="screen-only">
+            <span class="mock-label">[% loc('Hierarchical Attributes:') %]</span>
+        </dt>
+        <dd class="screen-only hierarchical-attributes">
+            [% SET selected = selected_hierarchical_attributes || {} %]
+
+            <div class="hierarchical-field">
```
dracos, line 163: I realise we need to keep the order, but these three divs feel like they could be a loop of
`FOR field in [ [ 'Geschäftsbereich', 'hierarchical_geschaftsbereich', 0 ], [ 'Objekt', 'hierarchical_objekt', 1 ], [ 'Kategorie', 'hierarchical_kategorie', 1 ] ] ...`

chrismytton (commented):
Thanks! Good to make the get_param bits more future-proof, and I hadn't thought/realised that the body could be an SDM not a division, so glad you caught that.

Can't approve as I'm the author, but 👍

PR #5639, opened 2025-09-03
[Northumberland] Add separate "Old ward" column to CSV export

chrismytton:
Previously the "Wards" column in the CSV export contained both old and current wards. This change splits this column into two columns, one for current wards, and one for old/historic wards.

For FD-5800

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Northumberland.pm
+++ b/perllib/FixMyStreet/Cobrand/Northumberland.pm
@@ -160,6 +160,28 @@ sub dashboard_export_problems_add_columns {
         nearest_address => 'Nearest address',
     );
 
+    # Insert Old ward column immediately after the Ward column
+    $csv->splice_csv_column('local_coords_x', 'old_ward', 'Old ward');
+
+    # MapIt children lookups:
+    #  - area_children()   -> current generation only (new wards)
+    #  - area_children(1)  -> all generations (old + new wards)
+    # We use these to split the CSV 'Ward' (current only) and 'Old ward'.
+    my $current_children = $csv->body ? $csv->body->area_children() : {};
+    my $all_children = $csv->body ? $csv->body->area_children(1) : {};
+
+    # Helper to compute Ward (current) and Old ward (historic) from a CSV areas string
+    my $compute_wards = sub {
+        my ($areas) = @_;
+        $areas ||= '';
+        my @ids = split ',', $areas;
+        my @current_ids = grep { $current_children->{$_} } @ids;
+        my @old_ids = grep { $all_children->{$_} && !$current_children->{$_} } @ids;
+        my $wards = join ', ', map { $current_children->{$_}->{name} } @current_ids;
+        my $old_ward = join ', ', map { $all_children->{$_}->{name} } @old_ids;
+        return ($wards, $old_ward);
+    };
+
     my $response_time = sub {
         my $hashref = shift;
         if (my $response = ($hashref->{fixed} || $hashref->{closed}) ) {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -160,6 +160,28 @@ sub dashboard_export_problems_add_columns {
...
+    my $current_children = $csv->body ? $csv->body->area_children() : {};
+    my $all_children = $csv->body ? $csv->body->area_children(1) : {};
+
+    # Helper to compute Ward (current) and Old ward (historic) from a CSV areas string
+    my $compute_wards = sub {
+        my ($areas) = @_;
+        $areas ||= '';
+        my @ids = split ',', $areas;
+        my @current_ids = grep { $current_children->{$_} } @ids;
+        my @old_ids = grep { $all_children->{$_} && !$current_children->{$_} } @ids;
```
dracos: To do this in one loop rather than two (given this is processing 000s of rows), and I think we can assume there's only one ward result, so perhaps something like this?
```
my ($wards, $old_ward);
for (@ids) {
    if ($current_children->{$_}) {
        $wards = $current_children->{$_}{name};
        last;
    } elsif ($all_children->{$_}) {
        $old_ward = $all_children->{$_}{name};
        last;
    }
}
```

perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -160,6 +160,28 @@ sub dashboard_export_problems_add_columns {
         nearest_address => 'Nearest address',
     );
 
+    # Insert Old ward column immediately after the Ward column
```
dracos: Can we remove the 'wards' column from csv_columns (so the code that finds both wards there, which is going to be ignored, doesn't run at all) and add two new columns (old_ward and current_ward?) here instead.

dracos (changes requested):
Looks good, two suggestions to make it more efficient.

perllib/FixMyStreet/Cobrand/Northumberland.pm
chrismytton: Have updated it to use a single loop, though code is slightly different as some rows (e.g. 4530478) have both old _and_ new ward listed, so need to handle that case.

perllib/FixMyStreet/Cobrand/Northumberland.pm
```diff
@@ -160,6 +160,44 @@ sub dashboard_export_problems_add_columns {
...
+    my @heads = @{ $csv->csv_headers };
+    for (my $i = 0; $i < @cols; $i++) {
+        if ($cols[$i] eq 'wards') {
+            splice @cols, $i, 1, ('current_ward', 'old_ward');
+            splice @heads, $i, 1, ('Current ward', 'Old ward');
+            last;
+        }
+    }
+    $csv->_set_csv_columns(\@cols);
+    $csv->_set_csv_headers(\@heads);
```
chrismytton: Is there a more concise way to do this?

perllib/FixMyStreet/Cobrand/Northumberland.pm
dracos: Oh right, yes, of course, thanks.

perllib/FixMyStreet/Cobrand/Northumberland.pm
dracos: Hmm sorry, thought we'd have done this before somewhere. That looks fine, though yes, you could make it more concise, by splicing in place, so ditching `@cols` and `@heads` and splicing directly on them, same as `splice_csv_column` is doing, something like:

```
    for (my $i = 0; $i < @{$csv->csv_columns}; $i++) {
        if ($self->csv_columns->[$i] eq 'wards') {
            splice @{$self->csv_columns}, $i, 1, ('current_ward', 'old_ward');
            splice @{$self->csv_headers}, $i, 1, ('Current ward', 'Old ward');
            last;
        }
    }
```

dracos (approved):
⚔️

PR #5641, opened 2025-09-04
[Southwark] Updated cobrand branding

dracos:
[skip changelog]


dracos (commented):
Can't approve PR I opened but looks fine, I can fix commit message duplication

dracos (commented):
Actually, looks like the PNGs could all be smaller

PR #5643, opened 2025-09-05
[MSS] Add an api for receiving updates from MSS

MorayMySoc:
Adds an api base and builds on that for MSS updates.

Adds tests for receiving updates.

https://github.com/mysociety/societyworks/issues/5027

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Api.pm
+++ b/perllib/FixMyStreet/App/Controller/Api.pm
@@ -1,0 +1,70 @@
+package FixMyStreet::App::Controller::Api;
+
+use Moose;
+use namespace::autoclean;
+use JSON::MaybeXS;
+use Try::Tiny;
+
+BEGIN { extends 'Catalyst::Controller' }
+
+has api_username => (isa => 'Str', is => 'rw', default => '');
+has api_password => (isa => 'Str', is => 'rw', default => '');
+has api_config => (isa => 'HashRef', is => 'rw', lazy => 1, default => sub {{}});
+has post_data => (isa => 'HashRef', is => 'rw', default => sub {{}});
+
+sub json_response : Private('/api/json_response') {
+    my ($self, $c, $code, $message) = @_;
+
+    my $body = {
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Api.pm
```diff
@@ -0,0 +1,70 @@
...
+        if (
+            ($self->api_password && $self->api_config->{password} && $self->api_username && $self->api_config->{username})
+            && ($self->api_password eq $self->api_config->{password} && $self->api_username eq $self->api_config->{username})
+        ) {
+            return;
+        }
+        $c->detach;
+    } catch {
+        $c->detach('/api/json_response', [401]);
+    }
```
chrismytton, line 39: Why is the try/catch needed here? Would it work if you just had a single call to `$c->detach('/api/json_response', [401]);` after the if statement?

perllib/FixMyStreet/App/Controller/Api/MSS/Validate/Update.pm
```diff
@@ -0,0 +1,53 @@
...
+around BUILDARGS => sub {
+    my ($orig, $class, $args) = @_;
+
+    die unless $args->{updated_datetime} =~ /\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d/;
+    die unless $args->{update_id};
+
+    return $class->$orig($args);
+};
+
+return 1;
```
chrismytton: Minor thing, but the standard in this codebase seems to be to omit the `return` at the end of the module and just having a bare `1;`

perllib/FixMyStreet/App/Controller/Api.pm
MorayMySoc: I *think* if something happens to the code, like a syntax error is introduced, and there is an uncaught error, it won't reach the detach and will continue processing - and that will mean the credentials are treated as accepted. So it was to secure against that.

PR #5644, opened 2025-09-08
[Bexley] Remember round instance ID through payment cancellation

dracos:
waste_reconstruct_bulky_data() is called on a bulky collection page, but also if payment is cancelled and we're showing the summary page again, for the same reason - so we need to get the reference out in there too, so it can be reused if they submit payment again. I don't mind the duplicate report creation in this case, because if they cancelled payment they could have changed anything else to do with the collection (date, items, etc), and this seems rare/easier than trying to treat it as an amendment of a confirmed collection.
(Echo we're fetching new reservation slots anyway in this case, which is why it hasn't come up there.)

Also don't ignore bulky reports in stuck reports alerting.
And an "AboutYou" tidy that was lying around.

For https://mysocietysupport.freshdesk.com/a/tickets/5941 [skip changelog]


perllib/Integrations/Whitespace/Booking.pm
```diff
@@ -67,7 +67,13 @@ sub check_slot_available {
         $property, undef, 'no_cache' );
 
     my ($slot) = grep { $_->{date} eq $collection_date } @$available_slots;
-    return $slot ? 1 : 0;
+    if ($slot) {
+        # We don't need new references like Echo, so no need to update chosen_date
+        # $args{form}->saved_data->{chosen_date} = $slot->{date} . ';' . $slot->{reference} . ';';
```
chrismytton, line 73: Is this commented out code just to show what Echo does, or can it be removed?

dracos: Yeah, it's the equivalent of what Echo does, and is what you'd do here if you did want to update it (but wouldn't change anything). Guess could be clearer

PR #5646, opened 2025-09-09
Add superuser category editing for display name cobrands

chrismytton:
When a cobrand is setup to use a display name its sometimes still useful to have a way to change the category name.

This change adds a form field for superusers to allow them to update the category name as well as the display name on these cobrands.

Should make tickets like FD-5943 possible to fix from the admin rather than having to use the DB console.

image


dracos (approved):
Yeah, was bit odd/confusing you could still edit them on .com, thanks

PR #5661, opened 2025-09-16
[Merton] (SLWP) Bulky same day amendment fix

dracos:
Fixes https://github.com/mysociety/societyworks/issues/4732 [skip changelog]

If a backend has no slots left, we cannot use the normal method of cancelling and creating a new booking (which we still have to do for date change amendments). Instead, add an amendment moderation entry, and store the changes in an amendment update for sending through to the backend (delayed confirmation if further payment is needed).

Tidied up some bits around the area to hopefully make this a bit clearer.

Writing open311-end now.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -399,30 +399,65 @@ sub process_bulky_amend : Private {
     my $p = $c->stash->{amending_booking};
 
     if ($c->cobrand->bulky_cancel_by_update) {
-        # In this case we want to update the event to mark it as cancelled,
-        # then create a new event with the amended booking data from the form
-        my $update = add_cancellation_update($c, $p, 'delayed');
 
-        if ($c->stash->{small_items}) {
-            $c->forward('process_small_items_data', [ $form ]) or return;
+        my $current_date = $c->cobrand->collection_date($p);
+        my ($date, $ref, $expiry) = split(";", $data->{chosen_date});
+        if ($c->cobrand->bulky_backend_can_amend_same_date && $current_date eq $date) {
+            # For the same date, we add an amendment update
+            $p->create_related( moderation_original_data => {
+                title => $p->title,
+                detail => $p->detail,
+                photo => $p->photo,
+                anonymous => $p->anonymous,
+                category => $p->category,
+                extra => $p->extra,
+            });
+            $c->stash->{report} = $p;
+            $c->cobrand->bulky_total_cost($data);
+            if ($c->stash->{payment}) {
+                my $update = add_amendment_update($c, $p, $data, 'delayed');
+                $p->unset_extra_metadata('payment_reference');
+                $p->set_extra_metadata(amendment_update => $update->id);
+                $p->update_extra_field({ name => 'payment', value => $c->stash->{payment} });
+                # XXX Store old payment reference and payment somewhere else?!
+                $p->update;
+                $c->forward('/waste/pay_process', [ 'bulky', $data->{payment_method}, $data ]);
+            } else {
+                my $update = add_amendment_update($c, $p, $data, 'immediate');
+                $p->waste_amend_extra_data($c->cobrand, $c->stash->{booking_maximum}, $data);
+                $c->cobrand->waste_amend_amendment_update($p, $update);
+                $update->update;
+                $p->update;
+                if ($c->cobrand->suppress_report_sent_email($p)) {
+                    $p->send_logged_email({ report => $p, cobrand => $c->cobrand }, 0, $c->cobrand);
+                }
+            }
         } else {
-            $c->forward('process_bulky_data', [ $form ]) or return;
-        }
+            # In this case we want to update the event to mark it as cancelled,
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -430,6 +465,27 @@ sub process_bulky_amend : Private {
     return 1;
 }
 
+sub add_amendment_update {
+    my ($c, $p, $data, $type) = @_;
+
+    my $update = $p->add_to_comments({
+        text => "Booking amended",
+        user => $c->cobrand->body->comment_user || $p->user,
```
nephila-nacrea, line 493: Is there ever a case where there isn't a body->comment_user?
At least, I know I get errors when e.g. I set up a cobrand locally and forget to set the comment user, so it seems to be something that other parts of our codebase assume exists.

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
```diff
@@ -108,7 +108,8 @@ has_page summary => (
 
         if ($c->stash->{amending_booking}) {
             my $current_date = $c->cobrand->collection_date($c->stash->{amending_booking});
-            return 1 if $current_date eq $form->saved_data->{chosen_date};
+            my ($date, $ref, $expiry) = split(";", $form->saved_data->{chosen_date});
```
nephila-nacrea, line 111: Was this a bug?

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1591,12 +1612,6 @@ sub waste_confirm_payment {
         }
     }
 
-    if ($cobrand->suppress_report_sent_email($self)) {
```
nephila-nacrea, line 1596: Why was this moved further up?

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -1583,6 +1583,27 @@ sub waste_confirm_payment {
...
+        $self->send_logged_email({ report => $self, cobrand => $cobrand }, 0, $cobrand);
+    }
+
+    if (my $amendment = $self->get_extra_metadata('amendment_update')) {
+        $amendment = FixMyStreet::DB->resultset("Comment")->find($amendment);
+        my $data = $amendment->get_extra_metadata('fms_extra_amend');
+        my $max_key = $self->category eq 'Small items collection' ? 'small_items_per_collection_max' : 'items_per_collection_max';
+        my $max = $cobrand->wasteworks_config->{$max_key} || 5;
+        $self->waste_amend_extra_data($cobrand, $max, $data);
+        $cobrand->waste_amend_amendment_update($self, $amendment);
```
nephila-nacrea, line 1600: Should this be a hook, if the only cobrand that implements this so far is SLWP2?

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
dracos: Yeah, generally, and actual site almost certainly, I think it's probably just to deal with any tests that haven't set it up right.

perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
dracos: Currently if we're amending an Echo (the only ones that stick extra data on the date like this, the others are just a date), you wouldn't want it to ever match, because presently you have to cancel/book with a new reservation slot.

perllib/FixMyStreet/DB/Result/Problem.pm
dracos: So that the email is sent if amendment_update is set (as that block then returns).

perllib/FixMyStreet/DB/Result/Problem.pm
dracos: We can never get here unless `bulky_backend_can_amend_same_date` was set, so we can be safe that it will be set up (and I guess if someone set that flag and forgot to create this function, we'd want it to error).

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -402,27 +402,9 @@ sub process_garden_new_or_renew : Private {
         $c->forward('/waste/process_request_data', [ $form, [ $c->stash->{report} ], 1 ]);
     }
 
-    my $payment_method = $data->{payment_method} || '';
-    if ( FixMyStreet->staging_flag('skip_waste_payment') ) {
-        $c->forward('/waste/pay_skip', []);
-    } elsif ($c->cobrand->waste_cheque_payments && $payment_method eq 'cheque') {
```
MorayMySoc: Does it matter that $c->cobrand->waste_cheque_payments is not covered in the new pay_process?

dracos: I think I thought that `options_payment_method` in `Billing.pm` could only output the cheque option if waste_cheque_payments was set, so it had already been checked by that.

PR #5664, opened 2025-09-17
[Aberdeenshire] Handle superseding defects

davea:
For https://github.com/mysociety/societyworks/issues/5138

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Aberdeenshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Aberdeenshire.pm
@@ -166,6 +166,74 @@ sub open311_munge_update_params {
 }
 
 
+=head2 open311_get_update_munging
+
+The incoming update might be for a defect which has superseded an existing one,
+so if that's the case we need to identify and close it.
+
+=cut
+
+sub open311_get_update_munging {
+    my ($self, $comment, $state, $request) = @_;
+
+    my $supersedes = $request->{extras}{supersedes};
+    return unless $supersedes && $supersedes =~ /^DEFECT_/;
+
+    $self->_supersede_report($comment->problem, $supersedes);
+}
+
+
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Aberdeenshire.pm
```diff
@@ -166,6 +166,83 @@ sub open311_munge_update_params {
...
+    # Migrate any alerts on the old report the new
+    my $alerts = $superseded->result_source->schema->resultset('Alert')->search({
+        alert_type => 'new_updates',
+        parameter => $superseded->id,
+        whendisabled => undef,
+        confirmed => { '!=', undef },
+    });
+
+    while (my $alert = $alerts->next) {
+        $alert->disable;
```
dracos: We could, I guess, update the alert to have a new parameter, rather than disable/recreate? Guess that loses a bit of technical history, but not sure that matters?

t/cobrand/aberdeenshire.t
```diff
@@ -176,6 +176,437 @@ FixMyStreet::override_config {
...
+                parameter => $superseded_report->id,
+                confirmed => 1,
+                cobrand => 'aberdeenshire',
+                lang => 'en-gb',
+                cobrand_data => '',
+                whensubscribed => DateTime->now->subtract(days => 1),
+            });
+
+            # Create a comment/update for the superseding report that indicates superseding
+            my $superseding_comment = $mech->create_comment_for_problem(
```
dracos: I guess to be a true copy of what is happening, the comment at this point in the flow isn't in the database yet (not that I think it matters, but I guess in theory something could try and look at an ID or something?).

dracos (changes requested):
Looks good, couple of small suggestions

PR #5670, opened 2025-09-19
Use container width for front page stats numbers.

dracos:
[skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -800,6 +800,7 @@ ul.error {
   margin: 0 auto;
   padding: 0 1em;
   max-width: $container-max-width;
+  container-type: inline-size;
 }
 
 // Use full width to reverse .container margins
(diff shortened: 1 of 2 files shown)


lucascumsille:
@dracos I added a fixup that allow us to keep using the `cqw`, but adds the `container-type: inline-size` rule to a more specific `div` instead of every single element using the class `.container`

This prevent the issue where the vertical margin spacing was collapsing and the mySoc badge was getting cropped in the upper half.

lucascumsille:
I didn't know about `clamp()` that is one nice trick. I like this solution a lot, but I would tweak the `preferred` value instead of `3.5VW` I would do `4.5VW`.

- Testing on desktop it will always take 2.5em, because anything `>3.5vw` is too big,
- But on mobile it will take the `preferred` value, however `3.5vw` looks a bit small

### 4.5VW
https://github.com/user-attachments/assets/9e88191a-058e-41ec-8cd7-d3f55d7369ba

### 3.5VW
Screenshot 2025-09-25 at 09 57 50

PR #5671, opened 2025-09-22
Persist send_method_used immediately during report sending

chrismytton:
Some senders call `discard_changes()` within their `send()` method, which results in the `send_method_used` change being overwritten rather than appended to. The end result of this is that in some cases updates aren't being sent via Open311 because the `send_method_used` ends up as `Email`.

This change ensures that `send_method_used` is persisted between sender calls, so in cases with multiple senders it correctly ends up as `Email,Open311` or similar.

Fixes https://github.com/mysociety/societyworks/issues/5146

Relates to FD-5945

### DIFF
--- a/perllib/FixMyStreet/Queue/Item/Report.pm
+++ b/perllib/FixMyStreet/Queue/Item/Report.pm
@@ -268,6 +268,9 @@ sub _send {
         my @body_ids = map { $_->id } @{ $sender->bodies };
         if ($sender->success) {
             $report->add_send_method($sender_name);
+            # Persist send_method_used immediately so later discard_changes() calls
+            # within other sender implementations cannot wipe previous values.
+            $report->update;
             push @remove_send_fail_body_ids, @body_ids;
         } else {
             push @add_send_fail_body_ids, @body_ids;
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Queue/Item/Report.pm
```diff
@@ -268,6 +268,9 @@ sub _send {
         my @body_ids = map { $_->id } @{ $sender->bodies };
         if ($sender->success) {
             $report->add_send_method($sender_name);
+            # Persist send_method_used immediately so later discard_changes() calls
+            # within other sender implementations cannot wipe previous values.
+            $report->update({ send_method_used => $report->send_method_used });
```
dracos: Would `$report->update;` not be enough here? The `send_method_used` has just been set in the `add_send_method` call, so you're setting it to the same thing.

dracos (changes requested):
Agree with the solution to update the report immediately, but don't think you need to re-set the field, unless I've missed something :)

perllib/FixMyStreet/Queue/Item/Report.pm
chrismytton: My intention was to _only_ update the `send_method_used` field at this point, even if other report fields had changed. I guess there shouldn't really be any other changes, given the senders are calling `discard_changes()`. I'll switch it to a plain `update()`.

dracos: Ah, so for info, update() will always update everything that's changed on the object, the parameters are just *extra* things to update on top.

chrismytton: Thanks, didn't realise that.

PR #5677, opened 2025-09-30
[TfL] Script to import pre-existing bus station reports from CSV

davea:
For https://github.com/mysociety/societyworks/issues/5148

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/assets.js
+++ b/web/cobrands/fixmystreet/assets.js
@@ -694,7 +694,8 @@ function construct_layer_class(options) {
 }
 
 function update_floating_button_messaging(layer, messaging) {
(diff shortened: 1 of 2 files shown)


web/cobrands/fixmystreet/assets.js
```diff
@@ -694,7 +694,8 @@ function construct_layer_class(options) {
 }
 
 function update_floating_button_messaging(layer, messaging) {
-    var id = 'js-responsibility-message-' + layer.id;
+    var uid = ( layer.controls[0] || {}).id || layer.id;
```
dracos, line 697: I think the first control on a layer is the HoverFeature, not the SelectFeature, fwiw. Don't think that matters, we only add controls that are shared in the same way at present, but just to note!

PR #5679, opened 2025-10-01
[Gloucester] Add a separate stopper message for GCH owned land

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/5908

[skip changelog]


web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -796,9 +796,14 @@ fixmystreet.assets.gloucester.watercourses_filter = function(feature) {
 };
 
 fixmystreet.assets.gloucester.street_or_plot_found = function(layer, asset) {
-    var selected_usrn = layer.selected_feature.attributes.itemId;
-    $('input[name="asset_resource_id"]').val(selected_usrn);
-    fixmystreet.assets.cobranduk.combined_road_found(layer, asset);
+    if (asset.attributes.owner && asset.attributes.owner[0] === '64a8227cebabca0376c5c0dd') {
+        layer.selected_feature = null;
+        fixmystreet.assets.cobranduk.combined_road_not_found(layer, "The land is owned by <strong>Gloucester City Homes</strong>. You will need to contact them directly regarding this matter. You can contact them via their <a href=https://www.gch.org.uk>website</a>");
```
dracos: Link needs updating as per your message :)

data/test-asset-layers.yml
```diff
@@ -676,4 +676,46 @@ westminster:
     asset_item: 'street nameplate'
     http_options:
       url: https://tilma.mysociety.org/resource-proxy/proxy.php?https://westminster.staging/46/query?
+gloucester:
+  - - name: alloy
+      http_options:
+        url: "tilma.staging/alloy/layer.php"
```
dracos: not sure if this matters, but cypress has a list of domains not to allow requests on, and tilma.staging.mysociety.org is in there, but this isn't. Rest of this file uses real staging URL so for consistency?

web/cobrands/fixmystreet-uk-councils/assets.js
dracos: Rather than changing combined_road_not_found, could we do it here, given if we've got here we know we want to call road_not_found etc? So directly here, override no_asset_message as you've done and call             fixmystreet.message_controller.road_not_found(key_layer, function() { return true; });


(I think that would do for here, but also note the message controller road_found does have a way (used for e.g. Bucks roads I think) to pass in a function for when a geometry is found but you don't want to allow all of them (as is the case here), and you can provide a special msg ID which would meant the no asset message could be set in configuration by making it a hash rather than a string. That works for Bucks but is more complicated here due to the double layer stuff.)

dracos (changes requested):
Looks good, and thanks for the new tests :) Couple of comments

web/cobrands/fixmystreet-uk-councils/assets.js
MorayMySoc: Sorry - irritating!

PR #5684, opened 2025-10-03
[Sutton][Small Items] Fix issue with not excepting small items from payment

MorayMySoc:
If a collection doesn't have a payment reference, skip adding payment comment.

Also fixes flow for small items to go through process_small_items_data rather than process_bulky_data during an amendment.

[skip changelog]


dracos (changes requested):
Looks good, but would like one more test of failure for the `process_` change you've made - I think all that's needed is to add to`t/app/controller/waste_sutton_small_items.t`  `band1_price => 3750, band1_max => 3` in the wasteworks_config set up and the tests will fail without your change (and Sutton does have those config items set).

PR #5693, opened 2025-10-10
[Bexley] Prevent duplicate garden subscription cancellations

chrismytton:
Added a check for recent cancellations before processing a new cancellation request. If a cancellation exists within the last hour, the system now shows a confirmation message without creating a duplicate entry.

Some of the calls to the backends were taking 30s or more, which meant if a second request came in during that time it could lead to two reports being created.

Fixes https://github.com/mysociety/societyworks/issues/4927

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Garden.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Garden.pm
@@ -13,6 +13,7 @@ use FixMyStreet::App::Form::Waste::Garden::Sacks::Purchase;
 use FixMyStreet::App::Form::Waste::Garden::Transfer;
 use WasteWorks::Costs;
 use Hash::Util qw(lock_hash);
+use JSON::MaybeXS;
 
 has feature => (
     is => 'ro',
@@ -228,6 +229,27 @@ sub process_garden_cancellation : Private {
         $data->{new_bins} = $bin_count * -1;
     }
     $c->forward('setup_garden_sub_params', [ $data, undef ]);
+
+    # Check for existing recent cancellation to prevent duplicates from concurrent requests
+    # This is Bexley-specific as Bexley auto-confirms cancellations
+    if ($c->cobrand->moniker eq 'bexley') {
+        my $uprn = $c->stash->{property}{uprn};
+        my $existing_cancel = $c->cobrand->problems->search({
+            category => 'Cancel Garden Subscription',
+            state => 'confirmed',  # Bexley auto-confirms cancellations
+            created => { '>=' => \"current_timestamp-'1 hour'::interval" },
+            # TODO: Update this to use the new `uprn` column once GH-5745 is merged.
+            extra => { '@>' => encode_json({ "_fields" => [ { name => "uprn", value => $uprn } ] }) }
+        })->first;
+
+        if ($existing_cancel) {
+            # Already cancelled recently - show confirmation without creating duplicate
+            $c->stash->{payment_method} = $existing_cancel->get_extra_field_value('payment_method');
+            $c->stash->{template} = 'waste/garden/cancel_confirmation.html';
+            $c->stash->{property_id} = $c->stash->{property}{id};
+            $c->detach;
+        }
+    }
 
     $c->forward('/waste/add_report', [ $data, 1 ]) or return;
 
--- a/t/app/controller/waste_bexley_garden.t
+++ b/t/app/controller/waste_bexley_garden.t
@@ -2198,6 +2198,7 @@ FixMyStreet::override_config {
         };
 
         subtest 'with Whitespace data' => sub {
+            $mech->delete_problems_for_body($body->id);
             $whitespace_mock->mock(
                 'GetSiteCollections',
                 sub {
@@ -2385,6 +2386,7 @@ FixMyStreet::override_config {
     };
 
     subtest 'Test direct debit cancellation' => sub {
+        $mech->delete_problems_for_body($body->id);
         $mech->clear_emails_ok;
 
         # Log in as a staff user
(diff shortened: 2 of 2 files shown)


perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -225,6 +225,30 @@ sub process_garden_cancellation : Private {
...
+        created => { '>=' => \"current_timestamp-'1 hour'::interval" },
+    })->all;
+
+    my $existing_cancel;
+    for my $cancel (@recent_cancels) {
+        if (($cancel->get_extra_field_value('uprn') || '') eq $uprn) {
+            $existing_cancel = $cancel;
+            last;
+        }
+    }
```
chrismytton: There's presumably a way to do this as part of the `->search` query above, but I couldn't seem to get it working.

t/app/controller/waste_bexley_garden.t
```diff
@@ -2117,6 +2119,68 @@ FixMyStreet::override_config {
...
+        }});
+        like $mech->text, qr/Your subscription has been cancelled/, 'Shows confirmation for second attempt';
+
+        # Verify no duplicate was created
+        my @all_cancellations = FixMyStreet::DB->resultset('Problem')->search({
+            category => 'Cancel Garden Subscription',
+            state => 'confirmed',
+        })->all;
+        # Filter to only those with UPRN 10001
+        @all_cancellations = grep { ($_->get_extra_field_value('uprn') || '') eq '10001' } @all_cancellations;
```
chrismytton: As above, think this should be possible in one step.

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -225,6 +225,33 @@ sub process_garden_cancellation : Private {
...
+        my $uprn = $c->stash->{property}{uprn};
+        my @recent_cancels = $c->cobrand->problems->search({
+            category => 'Cancel Garden Subscription',
+            state => 'confirmed',  # Bexley auto-confirms cancellations
+            created => { '>=' => \"current_timestamp-'1 hour'::interval" },
+        })->all;
+
+        my $existing_cancel;
+        for my $cancel (@recent_cancels) {
+            if (($cancel->get_extra_field_value('uprn') || '') eq $uprn) {
```
nephila-nacrea: Would it be simpler to fold the UPRN comparison into the DB search query (like we do for e.g. get_pending_subscription?)

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -225,6 +225,33 @@ sub process_garden_cancellation : Private {
...
+                $existing_cancel = $cancel;
+                last;
+            }
+        }
+
+        if ($existing_cancel) {
+            # Already cancelled recently - show confirmation without creating duplicate
+            $c->stash->{report} = $existing_cancel;
```
nephila-nacrea: Where is stash->{report} used?
(Want to make sure it can be a cancellation report and isn't always assumed to be new/renewal).

t/app/controller/waste_bexley_garden.t
nephila-nacrea: I'm not even sure the UPRN check is needed, as there shouldn't be reports for other UPRNs at this point (since you cleared the DB at the start of the subtest).
Though might be good to create a cancellation report for another UPRN, to make sure the UPRN filtering in the main code works.

chrismytton:
@nephila-nacrea I'm afraid I've mucked up the commits a bit so the first commit isn't actually the one that you reviewed, because of rebasing and me getting in a muddle. So you'll probably need to review it as a whole again, sorry. Hopefully it makes sense!

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -229,6 +230,26 @@ sub process_garden_cancellation : Private {
...
     $c->forward('setup_garden_sub_params', [ $data, undef ]);
 
+    # Check for existing recent cancellation to prevent duplicates from concurrent requests
+    # This is Bexley-specific as Bexley auto-confirms cancellations
+    if ($c->cobrand->moniker eq 'bexley') {
+        my $uprn = $c->stash->{property}{uprn};
+        my $existing_cancel = $c->cobrand->problems->search({
+            category => 'Cancel Garden Subscription',
+            state => 'confirmed',  # Bexley auto-confirms cancellations
+            created => { '>=' => \"current_timestamp-'1 hour'::interval" },
```
nephila-nacrea, line 240: I assume this is correct because it follows an existing pattern in our code, but do you know where this particular `::interval` syntax is documented?

dracos: https://www.postgresql.org/docs/18/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS is the documentation for casting

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -229,6 +230,27 @@ sub process_garden_cancellation : Private {
...
+            # TODO: Update this to use the new `uprn` column once GH-5745 is merged.
+            extra => { '@>' => encode_json({ "_fields" => [ { name => "uprn", value => $uprn } ] }) }
+        })->first;
+
+        if ($existing_cancel) {
+            # Already cancelled recently - show confirmation without creating duplicate
+            $c->stash->{payment_method} = $existing_cancel->get_extra_field_value('payment_method');
+            $c->stash->{template} = 'waste/garden/cancel_confirmation.html';
+            $c->stash->{property_id} = $c->stash->{property}{id};
+            $c->detach;
```
dracos, line 250: @chrismytton I think (though see below), this should perhaps `override_template` and return, for the same reasons as 25466fca50dbebe537eac341603f4bf9bd745dae and c04e23dd4e5ce90054b567a04764b39e908c398c - detaching means the unique ID wouldn't be cleared (just in case they somehow resubmit after an hour).
But then, if the unique ID *was* being cleared, how was the double submission happening in the first place?! So am confused as to what is happening here, might be something else going on :-/

PR #5698, opened 2025-10-16
[Bexley][GGW] Update pro-rata calculation for garden waste subscription

chrismytton:
Update the pro-rata logic in Costs.pm to include Bexley alongside Bromley. Add tests to check pro-rata amounts are correct for both credit card and direct debit.

In `perllib/FixMyStreet/App/Controller/Waste/Garden.pm`, `process_garden_modification` calculates the pro-rata cost by calling `$pro_rata = $costs->pro_rata_cost($new_bins);`. If you look at the `pro_rata_cost` method it's checking `if ($self->has_pro_rata_modify) {` before it does the pro-rata calculation. The `has_pro_rata_modify` attribute is set in the `_build_has_pro_rata_modify` method, which was only setting it to true for Bromley.

This all meant that Bexley was using the total for the year for the pro-rata cost, rather than the partial charge that we calculate.

I've added the config for this change onto the servers already.

Part of [this Basecamp todo](https://3.basecamp.com/4020879/buckets/43587866/card_tables/cards/9164987302).

### DIFF
--- a/t/app/controller/waste_bexley_garden.t
+++ b/t/app/controller/waste_bexley_garden.t
@@ -129,6 +129,8 @@ FixMyStreet::override_config {
             ggw_first_bin_discount => $ggw_first_bin_discount,
             ggw_cost_first => $ggw_cost_first,
             ggw_cost => $ggw_cost,
+            pro_rata_minimum => 0,
+            pro_rata_weekly => 106,
             cc_url => 'http://example.org/cc_submit',
             scpID => 1234,
             hmac_id => 1234,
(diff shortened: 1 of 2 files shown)


t/app/controller/waste_bexley_garden.t
```diff
@@ -889,6 +891,197 @@ FixMyStreet::override_config {
...
+                external_id => "Agile-$contract_id",
+                user => $user,
+            },
+        );
+        $new_sub_report->set_extra_fields(
+            { name => 'uprn', value => $uprn } );
+        $new_sub_report->update;
+        FixMyStreet::Script::Reports::send();
+
+        $agile_mock->mock( 'CustomerSearch', sub { {
```
nephila-nacrea, line 915: I don't know if some of this needs to be folded into the mock stuff that Matthew did. (I can always try that on Monday if need be).

nephila-nacrea (approved):

t/app/controller/waste_bexley_garden.t
chrismytton: Yeah, hopefully not too tricky!

chrismytton:
@nephila-nacrea This is now on staging. I haven't rebased it at all, so this is still based on an older version of the `bexley-ggw-pending-msg` branch and will need to be rebased/cherry-picked onto your new branch.

It's just commit 9c597a01908a2342c14fe7bbbe71dca5aa401842 that we're interested in.

chrismytton:
After discussing this in Slack it was decided to revert this as it hasn't been explicitly asked for and can introduce edge cases and bugs. Easier to just charge for a full year for a new bin.

PR #5658, opened 2025-09-11
[Bexley][WW] Add pages for GGW verifications (for cancel, modify, renew) [MERGED TO STAGING]

nephila-nacrea:
For https://github.com/mysociety/societyworks/milestone/131.

Closes https://github.com/mysociety/societyworks/issues/5139.
Closes https://github.com/mysociety/societyworks/issues/5124.
Closes https://github.com/mysociety/societyworks/issues/5122.

[skip changelog]

PR #5699, opened 2025-10-16
[Bexley][GGW] Amendments and verifications

nephila-nacrea:
A holding branch for the long-running amendments work plus the verifications work (and any in-between pieces that also need to be considered).

Contains PRs:
https://github.com/mysociety/fixmystreet/pull/5374
https://github.com/mysociety/fixmystreet/pull/5658
https://github.com/mysociety/fixmystreet/pull/5688
https://github.com/mysociety/fixmystreet/pull/5690

[skip changelog]

### DIFF
--- a/bin/bexley/cancel-garden-waste-from-agile
+++ b/bin/bexley/cancel-garden-waste-from-agile
@@ -1,0 +1,40 @@
+#!/usr/bin/env perl
+
+use v5.14;
+use warnings;
+
+BEGIN {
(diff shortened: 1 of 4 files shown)


dracos:
Please could we try and tidy things as we go along, not leave things until the end. eg I understand we changed the approach with the verifications to have 'shared' forms as discussed, so once that was done/approved, the previous approach should have been tidied/removed/dropped at that point

PR #5700, opened 2025-10-16
[SLWP] Include bulky items in CSV exports.

neprune:
For https://mysocietysupport.freshdesk.com/a/tickets/6149.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
@@ -10,6 +10,7 @@ package FixMyStreet::Roles::Cobrand::KingstonSutton;
 
 use Moo::Role;
 use Hash::Util qw(lock_hash);
+use List::Util qw(max);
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -195,6 +195,9 @@ sub dashboard_export_problems_add_columns {
 
     $csv->modify_csv_header( Detail => 'Address' );
 
+    my $config = $self->wasteworks_config || {};
+    my $max_items = $config->{items_per_collection_max} || 5;
```
dracos: Sutton has small collections as well as bulky, and stores the items in the same variables, and has a different maximum. I guess for ease you could take the higher of items_per_collection_max and small_items_per_collection_max (as it's only one different for Sutton); call the column "Item $_" rather than Bulky Item, and then this would cover both?

dracos (approved):
Nice, thanks

PR #5701, opened 2025-10-16
Oxfordshire show "Other" category in report meta line.

neprune:
[skip changelog]

For https://mysocietysupport.freshdesk.com/a/tickets/6148.

### DIFF
--- a/perllib/FixMyStreet/DB/Result/Problem.pm
+++ b/perllib/FixMyStreet/DB/Result/Problem.pm
@@ -773,23 +773,22 @@ sub meta_line {
     my $category = $problem->category_display;
     $category = $cobrand->call_hook(change_category_text => $category) || $category;
 
-    # Create display version with quotes for "Other" category
-    my $category_display = ($category eq 'Other') ? '"' . $category . '"' : $category;
-
     # Call a hook on the cobrand to check if the problem should be treated as anonymous
     my $anonymous = $cobrand->call_hook('is_problem_anonymous');
 
+    my $show_category = $category ne _('Other') || $cobrand->show_other_category_in_summary;
+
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -773,20 +773,23 @@ sub meta_line {
     my $category = $problem->category_display;
     $category = $cobrand->call_hook(change_category_text => $category) || $category;
 
+    # Create display version with quotes for "Other" category
+    my $category_display = ($category eq 'Other') ? '"' . $category . '"' : $category;
```
dracos: This needs _() around Other to match it being translated, I think?
Also would like curly quotes - but I'm not sure the quotes are needed (or at least, there are other categories where it doesn’t read particularly well as a sentence but no-one has minded so far :) )

perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -773,20 +773,23 @@ sub meta_line {
...
+
     # Call a hook on the cobrand to check if the problem should be treated as anonymous
     my $anonymous = $cobrand->call_hook('is_problem_anonymous');
 
     if ( $problem->anonymous || $anonymous ) {
-        if ( $problem->service_display && $category && $category ne _('Other') ) {
+        if ( $problem->service_display && $category && ($category ne _('Other') || $cobrand->show_other_category_in_summary) ) {
```
dracos: Perhaps factor this Other or flag check out first, as you then use it four times after all

dracos (approved):
Hmm, I think my approval for this got lost, I swear I remember doing it!
Thanks!

PR #5704, opened 2025-10-20
[Bexley] Form for requesting assisted collections

MorayMySoc:
Adds a form for requesting assisted collections. If the form is used by staff there is an extra notes field they can use.

An FMS report is created with the form details, with the exception of the staff notes.

Emails are sent to the member of the public with the information in of what has been requested and to the council to let them know about the request.

https://github.com/mysociety/societyworks/issues/5061
https://github.com/mysociety/societyworks/issues/5066

[skip changelog]


t/app/controller/waste_bexley.t
```diff
@@ -1236,6 +1236,87 @@ FixMyStreet::override_config {
             like $email_submit, qr{<td><p>UPRN</p></td> <td><p>$_->{id}</p></td>}
         };
     }
+
+    my $assisted_collection = $mech->create_contact_ok(
```
dracos: Probably best to put this up at the top, with the other category creation, it's sort of like "set up" and would be present for all the tests (just in case it had an effect on something else already being tested, I mean).

PR #5707, opened 2025-10-22
[Echo] Work out frequency ourselves.

dracos:
Rather than rely upon a task or schedule description. [skip changelog]

For FD-6114 (date display from that is fixed, this is about the frequency display which is still wrong)

Most file changes are removing now-unused ScheduleDescriptions; some tests had to have time changes so that the last/next were actually displaying the right things (rather than relying on the description as previously); some code no longer needs to check for 'fortnightly'; the substantive change is in `Echo.pm`, to store the last/next original dates, and pick the closest weeks from them. This will display the wrong day if the next schedule's original date was a different day, but not sure there's a lot we can do then.

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
@@ -461,15 +461,11 @@ sub _schedule_object {
 
 sub _parse_schedules {
     my $servicetask = shift;
-    my $desc_to_use = shift || 'schedule';
     my $schedules = Integrations::Echo::force_arrayref($servicetask->{ServiceTaskSchedules}, 'ServiceTaskSchedule');
 
     my $today = DateTime->now->set_time_zone(FixMyStreet->local_time_zone)->strftime("%F");
-    my ($min_next, $max_last, $description, $max_end_date);
-
-    if ($desc_to_use eq 'task') {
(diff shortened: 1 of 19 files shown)


perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -493,9 +489,24 @@ sub _parse_schedules {
...
-            $description = $schedule->{ScheduleDescription} if $desc_to_use eq 'schedule';
+            $next_orig = construct_bin_date($last->{OriginalScheduledDate});
         } elsif ($d && (!$max_last || $d > $max_last->{date})) {
             $max_last = _schedule_object($last, $d);
+            $last_orig = construct_bin_date($last->{OriginalScheduledDate});
+        }
+    }
+
+    my $description = '';
+    if ($next_orig && $last_orig) {
```
davea, line 502: Is there any possibility `$next_orig` will be <= `$last_orig` and we'd get some odd e.g. `every -2 weeks` situation?

davea (approved):

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
dracos: I think `delta_days` always returns a non-negative result, so should be okay

PR #5710, opened 2025-10-22
[Bristol] Override ignored USRNs when sending to Alloy

davea:
The report may get a USRN value from the frontend which is one that should be ignored when sending to Alloy. Now we replace it with the correct code if that's the case.

I think this means we can remove the asset layer config that stores the USRN:
```
usrn:
      attribute: USRN
      field: usrn
```

but we need to leave the actual asset layer in to ensure a valid location is selected.

[skip changelog]


dracos (approved):
Looks good; I'd probably still leave the USRN in the client, saves a double lookup in the normal case where there isn't any issue?

PR #5712, opened 2025-10-23
[Bexley][GGW] Add explanatory text for one-off DD payments

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/43587866/card_tables/cards/9202444583.

[skip changelog]

### DIFF
--- a/templates/web/base/waste/summary.html
+++ b/templates/web/base/waste/summary.html
@@ -78,7 +78,7 @@
 
 </dl>
 
-[% IF c.cobrand.moniker == 'bexley' && data.payment_method == 'direct_debit' %]
+[% IF c.cobrand.moniker == 'bexley' && type == 'amend' && data.payment_method == 'direct_debit' %]
   <div class="govuk-form-group">
     <fieldset class="govuk-fieldset">
       <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -863,17 +863,6 @@ sub waste_sub_overdue {
 
 # Garden waste
 
-sub waste_display_payment_method {
```
nephila-nacrea, line 866: Huh, just realised this is the same change Chris made in https://github.com/mysociety/fixmystreet/pull/5703/files.

templates/web/base/waste/summary.html
```diff
@@ -78,6 +78,18 @@ <h1 class="govuk-heading-xl">[% title %]</h1>
 
 </dl>
 
+[% IF c.cobrand.moniker == 'bexley' && data.payment_method == 'direct_debit' %]
```
dracos: I think this is being shown on new subscription pages as well, presumably needs a check it's only on a modify page.

dracos (changes requested):
One query about the summary page showing it for not just modifying.
And yes, the change in #5703 is covered by that :)

PR #5714, opened 2025-10-24
[Open311] Protect contacts with send_method from service list updates

davea:
Treat contacts that have a send_method set the same as if open311_protect is set.

[skip changelog]


davea:
Yes, that would be much simpler wouldn't it!

PR #5715, opened 2025-10-24
[Bexley] Staff to be able to approve or deny assisted collection requests

MorayMySoc:
Adds the form and controller and processor and templates for a form to allow staff to approve or not a request for assisted help.
    
In the case of approval an open311 report is created to be used for sending details to Whitespace.

Includes tests

https://github.com/mysociety/societyworks/issues/5064

[skip changelog]


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -796,6 +798,36 @@ sub construct_bin_problem_form {
...
+    $c->stash->{form_class} = 'FixMyStreet::App::Form::Waste::Assisted';
+    $c->forward('form');
+}
+
+sub process_assisted_data : Private {
+    my ($self, $c, $form) = @_;
+    my $data = $form->saved_data;
+
+    if ($data->{outcome_choice} eq 'Approve') {
+        $data->{category} = 'Request assisted collection approve';
```
dracos: For consistency, perhaps this could be "Assisted collection add" (is what it's called for Bromley and Merton).

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -796,6 +798,36 @@ sub construct_bin_problem_form {
...
+        $data->{name} = $c->cobrand->body->comment_user->name;
+        $data->{email} = $c->cobrand->body->comment_user->email;
+        $data->{title} = 'Confirm assisted collection';
+        $data->{detail} = 'Generated by approval form';
+        foreach my $id (grep { /^assisted_/ } keys %$data) {
+            $c->set_param($id, $data->{$id});
+        };
+        $c->forward('add_report', [ $data ]);
+    } else {
+        $c->cobrand->call_hook('_add_update_to_assisted_request', 'deny');
```
dracos: I don't think this needs to be a hook or be private, we want to always call it and for it to exist if we're here.

perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -184,6 +186,14 @@ sub waste_munge_enquiry_form_pages {
     }
 }
 
+sub waste_post_report_creation {
+    my ($self, $report) = @_;
+
+    if ($report->category eq 'Request assisted collection approve') {
+        $self->_add_update_to_assisted_request('approve');
```
dracos: This is fine, but would it be clearer if this were in `process_assisted_data` after add_report? You have the report at that point in `$c->stash->{report}` and that then keeps the update adding for approval and denial in the same block.

templates/web/base/admin/reports/_edit_waste.html
```diff
@@ -10,6 +10,11 @@
          [% uri = c.uri_for_email( '/report', problem.id ) %]
     <li><a href="[% uri %]" class="admin-offsite-link">[% loc('View report on site' )%]</a></li>
 
+[% IF problem.category == 'Request assisted collection' AND report.state == 'confirmed' %]
+[% SET response_link = cobrand.base_url _ '/waste/' _ problem.get_extra_field_value('uprn') _ '/assisted' %]
```
dracos: For bulky amendments, we have URL of the form /waste/PROPERTYID/bulky/amend/REPORTID which means we know which report is being amended.
Could we do the same here, so that we know which report is being approved and you can't go to "assist" on any property without there being a report?

perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -796,6 +798,36 @@ sub construct_bin_problem_form {
     return $field_list;
 }
 
+sub assisted : Chained('property') : Args(0) {
```
dracos: Have an Args(1) here, look up and check that the ID given is an open assisted collection request report. And then you have the report already at the end when it comes to adding the update.

perllib/FixMyStreet/Cobrand/Bexley.pm
```diff
@@ -525,4 +535,48 @@ sub skip_alert_state_changed_to {
...
+        return;
+    };
+
+    my $uprn = $self->{c}->stash->{property}{uprn};
+
+    my $request_report = FixMyStreet::DB->resultset('Problem')->search({
+        category => 'Request assisted collection',
+        state => 'confirmed',
+        extra => { '@>' => encode_json({ "_fields" => [ { name => 'uprn', value => $uprn } ] }) }
+    })->first;
```
dracos: No ordering here, this could (in theory) return the wrong report conceivably; see other comment on how I think we should pass in the ID of the report to the approval flow URL rather than look it up.

perllib/FixMyStreet/Cobrand/Bexley.pm
MorayMySoc: dbdedaa585deaff5323c4bd36e1572deccf43516

perllib/FixMyStreet/App/Controller/Waste.pm
MorayMySoc: 98df3cabef185ea6146c1ecb39c67c2ce2607da9

perllib/FixMyStreet/App/Controller/Waste.pm
MorayMySoc: c23f99d30e2c8afef193aae0d6091f2b64694ead
ddf3074bfff16e62ecd943986137d73eeefeb666

perllib/FixMyStreet/Cobrand/Bexley.pm
MorayMySoc: 2b4539c0d5c96a5e5171c38ec545ce0351dabc65

templates/web/base/admin/reports/_edit_waste.html
MorayMySoc: 2b4539c0d5c96a5e5171c38ec545ce0351dabc65

MorayMySoc:
Merged into https://github.com/mysociety/fixmystreet/tree/5061-form-for-bexley-assisted but didn't push so no-autoclose

PR #5716, opened 2025-10-27
[Bexley][GGW] Add T&Cs to cancellation form

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/5125. 
For https://3.basecamp.com/4020879/buckets/43587866/card_tables/cards/9216848737.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden/Cancel/Bexley.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden/Cancel/Bexley.pm
@@ -47,6 +47,14 @@ has_page verify_failed => ( verify_failed() );
 has_page confirm =>
     FixMyStreet::App::Form::Waste::Garden::Cancel::Shared::intro();
 
+has_field confirm => (
+    type => 'Checkbox',
+    option_label => 'I agree to the terms and conditions',
+    required => 1,
+    label => '',
+    order => 998,
+);
+
 has_field continue => (
     type => 'Submit',
     value => 'Continue',
--- a/templates/web/bexley/waste/garden/cancel.html
+++ b/templates/web/bexley/waste/garden/cancel.html
@@ -1,8 +1,51 @@
 [% BLOCK 'content' %]
-  <p>
-    Please be aware refunds will only be given if a subscription is cancelled within 14 days of the original sign up.
-  </p>
+  <details>
+    <summary class="govuk-fieldset__heading">Terms and conditions</summary>
 
+    <h3>Right to cancel</h3>
+
+    <p>
+    You have the right to cancel this contract within 14 calendar days without giving any reason.
+    </p>
+
+    <p>
+    To exercise the right to cancel, you must inform us either by:
+    </p>
+    <ol>
+      <li>Online form - complete the cancellation process by submitting your cancellation request below</li>
+      <li>Email - send an email to Bexley's Garden Waste department using the details included below</li>
+    </ol>
+    <p>
+    Please ensure that you include your full name, your address, a contact telephone number and an email address. Upon receipt of your cancellation request, we will communicate to you an acknowledgement of receipt of such a cancellation without delay.
+    </p>
+
+    <h3>Effects of cancellation, including reimbursement (if applicable)</h3>
+
+    <p>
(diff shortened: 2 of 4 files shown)


templates/web/bexley/waste/garden/cancel_tandc.html
```diff
@@ -0,0 +1,35 @@
+<details>
+<summary class="govuk-fieldset__heading">Terms and conditions</summary>
+
+<h3>Right to cancel</h3>
+
+You have the right to cancel this contract within 14 calendar days without giving any reason.
```
dracos: `<p>...</p>` rather than multiple `<br>`s

templates/web/bexley/waste/garden/cancel_tandc.html
```diff
@@ -0,0 +1,35 @@
...
+<br><br>
+To exercise the right to cancel, you must inform us either by:
+<ol>
+  <br>
```
dracos: No `<br>` here. If it looks odd, this isn't the way to fix it (but I'm not sure it matters).

templates/web/bexley/waste/garden/cancel_tandc.html
```diff
@@ -0,0 +1,35 @@
...
+  <li>Online form - complete the cancellation process by submitting your cancellation request below</li>
+  <li>Email - send an email to Bexley's Garden Waste department using the details included below</li>
+</ol>
+Please ensure that you include your full name, your address, a contact telephone number and an email address. Upon receipt of your cancellation request, we will communicate to you an acknowledgement of receipt of such a cancellation without delay.
+<br>
```
dracos: No `<br>` here, wrap the line above in a `<p>`

perllib/FixMyStreet/App/Form/Waste/Garden/Cancel/Bexley.pm
```diff
@@ -44,8 +44,23 @@ has_field reason_further_details => (
...
+    intro => 'garden/cancel_tandc.html',
+    title => 'Cancel your garden waste subscription',
+    fields => ['confirm_tandc', 'submit'],
+    finished => sub {
+        return $_[0]->wizard_finished('process_garden_cancellation');
+    },
+    next => 'done',
+);
+
+has_field confirm_tandc => (
```
dracos: Was there a reason this had to be changed from `confirm`? Guess doesn't matter.

perllib/FixMyStreet/App/Form/Waste/Garden/Cancel/Bexley.pm
```diff
@@ -44,8 +44,23 @@ has_field reason_further_details => (
 
 has_page verify_failed => ( verify_failed() );
 
-has_page confirm =>
-    FixMyStreet::App::Form::Waste::Garden::Cancel::Shared::intro();
+has_page confirm => (
+    intro => 'garden/cancel_tandc.html',
```
dracos: This could be the same template for Bexley, I don't see why it needs to be different?

nephila-nacrea: True, it is superseding the original cancel text.

perllib/FixMyStreet/App/Form/Waste/Garden/Cancel/Bexley.pm
nephila-nacrea: I think while I was rejigging the shared code, I found it less confusing for this to be a different name, but yeah, it didn't need to be different.

templates/web/bexley/waste/garden/cancel.html
```diff
@@ -1,8 +1,52 @@
...
+    <h3>Right to cancel</h3>
+
+    <p>
+    You have the right to cancel this contract within 14 calendar days without giving any reason.
+    </p>
+
+    <p>
+    To exercise the right to cancel, you must inform us either by:
+    <ol>
+      <br>
```
dracos: This one is still here, please remove :) 
Wouldn't have mentioned but also - `ol` and `p` are both block elements, you can't wrap one in another. So "To exercise the right to cancel, you must inform us either by:" can/should have an end `</p>` after it (though it's implicit), and you don't want a `</p>` after the `</ol>` as as that point there isn't one open.

nephila-nacrea: Ah I missed that br, thanks.

PR #5722, opened 2025-10-30
[Bexley] Check APN references for non-payment.

dracos:
They are done remotely, so can be checked like SCP payments.
Spotted in https://mysocietysupport.freshdesk.com/a/tickets/6181


chrismytton (approved):
Looks good! Does this fix apply to Brent as well, as they have a similar setup?

dracos:
Would do, but Brent don't have bulky, so doesn't matter.

PR #5723, opened 2025-10-31
[Merton] Extend filter for TfL River categories

davea:
Same as done for Brent - open up filter to anything beginning 'River '

For FD-6246

[skip changelog]


MorayMySoc (approved):
Might be worth just adding Camden to this if you don't mind as they've got the same thing.

PR #5726, opened 2025-11-03
[Hackney] Route tree reports based on asset owner attribute

chrismytton:
Add tree-aware email routing for Hackney that directs tree reports to different departments based on the tree asset owner (Housing, Highways, Parks, etc.). Introduces a generic split email parser that handles both the existing park/estate/other format and new tree owner formats.

Refactors _split_emails to use the new generic _parse_split_emails method while maintaining backwards compatibility. Updates validate_contact_email to support both routing formats.

Fixes https://github.com/mysociety/societyworks/issues/5163

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Hackney.pm
+++ b/perllib/FixMyStreet/Cobrand/Hackney.pm
@@ -203,6 +203,20 @@ sub get_body_sender {
             return { method => 'Email', contact => $contact };
         }
     }
+
+    # Tree assets are routed by owner if present, or fall back to 'other'
+    if (my $split_emails = $self->_parse_split_emails($contact->email)) {
+        my $owner = $problem->get_extra_field_value('owner');
+        my $owner_key = $owner ? $self->_normalize_owner_key($owner) : '';
+        my $to = $owner_key ? ($split_emails->{$owner_key} || $split_emails->{other}) : $split_emails->{other};
+        if ($to) {
+            $problem->set_extra_metadata(split_match => { $contact->email => $to });
+            if (is_valid_email($to)) {
+                return { method => 'Email', contact => $contact };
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Hackney.pm
```diff
@@ -203,6 +203,20 @@ sub get_body_sender {
             return { method => 'Email', contact => $contact };
         }
     }
+
+    # Tree assets are routed by owner if present, or fall back to 'other'
+    if (my $split_emails = $self->_parse_split_emails($contact->email)) {
+        my $owner = $problem->get_extra_field_value('owner');
+        my $owner_key = $owner ? $self->_normalize_owner_key($owner) : '';
+        my $to = $owner_key ? ($split_emails->{$owner_key} || $split_emails->{other}) : $split_emails->{other};
```
dracos: I think this line can just be `my $to = $split_emails->{$owner_key} || $split_emails->{other};` - if owner key is blank string, the lookup won't return anything anyway.

t/cobrand/hackney.t
```diff
@@ -557,4 +558,215 @@ subtest 'Dashboard CSV extra columns' => sub {
     };
 };
 
+# Test tree routing for Hackney based on asset owner attribute
+
+# Create tree categories with 'Trees' group using split email format
```
dracos: I would put these up at the top of the file - keeps all the setup together and matches more what the real thing would have, with all categories present all the time (just in case that throws up an issue somewhere else unexpected).

dracos (changes requested):
This all looks good, nice solution; couple of tiny suggestions is all

dracos (approved):

PR #5729, opened 2025-11-05
[WW] Have cobrands give garden due date, rather than no. of days.

neprune:
[skip changelog]

And also fix this date to the 1st of Jan for Bromley.
Closes https://github.com/mysociety/societyworks/issues/4703.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Brent.pm
+++ b/perllib/FixMyStreet/Cobrand/Brent.pm
@@ -1388,7 +1388,7 @@ sub waste_munge_request_form_data {
 
 =item * Cheque payments are not an option
 
-=item * Renewals can happen within 90 days so they are available from beginning of Jan
+=item * Renewals are available from beginning of Jan
 
 =cut
 
@@ -1401,8 +1401,12 @@ sub garden_service_name { 'Garden waste collection service' }
 sub garden_subscription_event_id { $EVENT_TYPE_IDS{garden} }
 sub garden_due_date {
     my ($self, $end_date) = @_;
-    return $end_date->subtract(days => 87);
-};
+    return DateTime->new(
+        year => $end_date->year,
+        month => 1,
+        day => 1,
+    );
+}
 
 sub waste_show_garden_modify {
     my ($self, $unit) = @_;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -335,18 +338,17 @@ sub garden_renew_as_new {
...
 
-    my $now = DateTime->now->set_time_zone( FixMyStreet->local_time_zone );
+    my $now = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
     my $sub_end = DateTime::Format::W3CDTF->parse_datetime($date);
-
-    my $diff = $now->delta_days($sub_end)->in_units('days');
-    return $diff <= $self->garden_due_days;
+    my $due_date = $self->garden_due_date($sub_end->truncate(to => 'day'));
+    return $now >= $due_date;
 }
```
neprune: Some artefacts here from copy and pasting the updated version in the mainline `Garden.pm` which I just decided to leave in.

dracos: Give the functions are identical, should we move it to Roles::Cobrand::Waste?

perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -750,7 +750,10 @@ sub garden_service_name { 'Green Garden Waste collection service' }
 sub garden_subscription_type_field { 'Subscription_Type' }
 sub garden_subscription_container_field { 'Subscription_Details_Container_Type' }
 sub garden_echo_container_name { 'LBB - GW Container' }
-sub garden_due_days { 48 }
+sub garden_due_date {
```
dracos, line 753: This is all good, but given it's only the one special case, it'd be nice perhaps to try and keep the simple way of specifying the renewal days, maybe? Not sure. One suggestion could be keep `garden_due_days` in all of them bar Brent, plus have a default `garden_due_date` in Roles::Cobrand::Waste that reads `garden_due_days` and calculates the date from that. Keeping the override in Brent to fix the renewal date there. Alternative suggestion could be change `waste_sub_due` to use `garden_due_days` if present, `garden_due_date` if not. Not sure any of these options is much different or worth it, though :)

neprune: Personally I feel the extra verbosity in this case is small enough that it's worth leaving as is for tracing through the logic, but I'm not bullish.

dracos: Yeah, no need to do more changes now :)

dracos (approved):
📅

PR #5730, opened 2025-11-06
[Waste] Always have quarter/three-quarter image/collections even on mobile

dracos:
Will need to run make_css. I don't think this should have any bad effects. Basically makes the quarter/three-quarter view always happen at any width rather than only >40em; and moves the services.html template down to a new row that is full width until $large-screen width when it becomes three-quarters like what's above. Also removed some inline styling that didn't seem needed; might be some tidying that could be done, but generally seems okay? [skip changelog]

Quick comparison before/after:
image image


lucascumsille:
I'll  use `grid` properties so we can make it look like this:

Screenshot 2025-11-26 at 09 33 30

PR #5732, opened 2025-11-07
Store change of non_public; show report admin logs on admin page

dracos:
This adds recording of the change of non_public to the admin_log table, and then makes the problem-related rows of that table accessible in the admin report edit page, putting the three types of update behind tabs.

image
image

### DIFF
--- a/templates/web/base/admin/reports/_edit_main.html
+++ b/templates/web/base/admin/reports/_edit_main.html
@@ -216,21 +216,83 @@ class="admin-offsite-link">[% problem.latitude %], [% problem.longitude %]</a>
 </div>
 
 [% moderation_history = problem.moderation_history %]
-[% IF moderation_history %]
-<div class="full-width" style="margin-bottom:-2em; padding-bottom: 2em">
-    <h2>[% loc('Moderation history') %]</h2>
-    [% last_history = problem %]
-    <ul>
-      [% FOR history IN moderation_history %]
-        <li><i>[% INCLUDE 'report/update/moderation_meta.html' %]</i>
-            [% INCLUDE 'report/update/moderation_diff.html' diff=history.compare_with(last_history) %]
-        </li>
-        [% last_history = history %]
-      [% END %]
-    </ul>
-</div>
-[% END %]
-
+[% admin_log = problem.admin_log_by_date_user %]
+[% SET show_tabs = (moderation_history AND admin_log) OR (moderation_history AND updates.size) OR (admin_log AND updates.size) %]
 <div class="full-width full-width--bottom">
-[% INCLUDE 'admin/list_updates.html' %]
+
+  [% IF show_tabs %]
+    <tabbed-element>
+      <ul>
+       [% IF updates.size %]
+        <li><a href="#section_updates">[% loc('Updates') %]</a></li>
+       [% END %]
+       [% IF moderation_history %]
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/DB/Result/AdminLog.pm
```diff
@@ -142,4 +142,20 @@ sub object {
     return $object;
 }
 
+my %action_map = (
```
davea: Opportunity here to make these translatable, maybe?

templates/web/base/admin/reports/_edit_main.html
```diff
@@ -216,21 +216,76 @@
...
+      </ul>
+  [% END %]
+
+    <section id="section_updates">
+        [% INCLUDE 'admin/list_updates.html' %]
+    </section>
+
+ [% IF moderation_history %]
+  <section id="section_moderation">
+        <h2>[% loc('Moderation history') %]</h2>
```
davea, line 244: Might be nice to omit the `h2` if `show_tabs` is set to save a bit of repetition?
image

(though might need some vertical padding too)
image

templates/web/base/admin/reports/_edit_main.html
```diff
@@ -216,21 +216,76 @@
...
+            </li>
+            [% last_history = history %]
+          [% END %]
+        </ul>
+  </section>
+    [% END %]
+
+  [% IF admin_log %]
+    <section id="section_log">
+        <h2>[% loc('Admin log') %]</h2>
```
davea, line 259: Ditto here
image

templates/web/base/admin/reports/_edit_main.html
```diff
@@ -216,21 +216,76 @@
...
+        [% FOREACH moment IN admin_log.keys.sort %]
+            [%- curdate = admin_log.$moment.date.strftime('%A, %e %B %Y') -%]
+            [%- IF date != curdate %]
+                [% '</dl>' IF date %]
+                <h3>[% curdate %]</h3>
+                <dl class="timeline">
+                [%- date = curdate -%]
+            [%- END -%]
+          [% FOREACH admin_user IN admin_log.$moment.users.keys %]
+            <dt>[% admin_log.$moment.date.hms %]<br>[% admin_user %]</dt>
```
davea: These ended up weirdly misaligned after a few entries:

image

Would it be nicer as a table, maybe (quick hack mockup)? Could even link name to user edit page
image

web/js/tabbed-interface.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    // Initially activate the first tab and reveal the first tab panel
+    this.tabs[0].removeAttribute('tabindex');
+    this.tabs[0].setAttribute('aria-selected', 'true');
+    this.panels[0].hidden = false;
+  }
+}
+
+window.customElements.define('tabbed-element', TabbedElement);
```
davea, line 94: Nice!

perllib/FixMyStreet/DB/Result/AdminLog.pm
dracos: Would have to be moved inside the function, to look up after the language is set; could do. Though on user log page it's currently displayed as eg "Added problem 24" where "problem" is the hard-coded object type, so that still wouldn't be translated.

davea (approved):
Lovely! 👍

PR #5734, opened 2025-11-11
Redirect after DD email

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/5000.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Garden.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Garden.pm
@@ -14,6 +14,7 @@ use FixMyStreet::App::Form::Waste::Garden::Transfer;
 use WasteWorks::Costs;
 use Hash::Util qw(lock_hash);
 use JSON::MaybeXS;
+use MIME::Base64;
 
 has feature => (
     is => 'ro',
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -573,6 +575,30 @@ sub direct_debit_complete : Path('/waste/dd_complete') : Args(0) {
         sent_confirm_id_ref => $c->stash->{report}->id,
     } );
 
+    my $params_encoded = encode_base64(
```
dracos: I don't think we want to pass all these variables in a URL, especially an email address.

I think what I would look to do is redirect to `$c->stash->{report}->confirmation_url($c)` (like the FMS reports do), and have an extra bit in that controller for a DD report showing the right template there. Then the templates here don't need changing.

nephila-nacrea: Have got something working, though the flash messages for 'message' and 'error' disappear if you reload. But I think the template could do with some improvements anyway, but not major/not worth dealing with now. Main thing is the email is not resent.

perllib/FixMyStreet/App/Controller/Waste/Garden.pm
```diff
@@ -566,14 +568,16 @@ sub direct_debit_complete : Path('/waste/dd_complete') : Args(0) {
...
 
-    $c->stash->{title} = "Direct Debit mandate";
-
     $c->send_email('waste/direct_debit_in_progress.txt', {
         to => [ [ $c->stash->{report}->user->email, $c->stash->{report}->name ] ],
         sent_confirm_id_ref => $c->stash->{report}->id,
     } );
 
-    $c->stash->{template} = 'waste/dd_complete.html';
+    for (qw/ message error /) {
```
dracos: Neither message nor error are ever set, as far as I can see, so this flash code isn't needed.

PR #5739, opened 2025-11-13
[Bromley] Chinbrook Meadows asset layer

chrismytton:
This adds the JS for a new asset layer for Bromley.

You'll need to add the following under Bromley's `asset_layers:` key in general.yml:

```yaml
- wfs_feature: "Chinbrook_Meadows"
  asset_type: 'area'
  always_visible: true
  road: true
  non_interactive: true
  stylemap: fixmystreet.assets.stylemap_invisible
  no_asset_message: The location you have selected is within Chinbrook Meadows, which is not accessible for public reporting.
  actions:
    found: fixmystreet.assets.bromley.chinbrook_found
    not_found: fixmystreet.assets.bromley.chinbrook_not_found
```

See also https://github.com/mysociety/fixmystreet.com/pull/76

For FD-6165

You can find Chinbrook Meadows just south-east of Grove Park station:

image

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/assets.js
+++ b/web/cobrands/fixmystreet-uk-councils/assets.js
@@ -325,6 +325,14 @@ fixmystreet.assets.bromley.add_park_message = function(layer) {
     layer.map_messaging.asset = layer.fixmystreet.no_asset_message;
 };
 
+fixmystreet.assets.bromley.chinbrook_found = function(layer) {
+    fixmystreet.message_controller.road_not_found(layer, function() {return true;});
+};
+
+fixmystreet.assets.bromley.chinbrook_not_found = function(layer) {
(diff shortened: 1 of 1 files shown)


web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -325,6 +325,14 @@ fixmystreet.assets.bromley.add_park_message = function(layer) {
     layer.map_messaging.asset = layer.fixmystreet.no_asset_message;
 };
 
+fixmystreet.assets.bromley.chinbrook_found = function(layer) {
+    fixmystreet.message_controller.road_not_found(layer, function() {return true;});
```
dracos, line 329: Matches others, perhaps we should open a separate ticket to make this return-true function not needed in some way, feels like this could be tidied somehow.

PR #5742, opened 2025-11-14
[WW][Sutton] Add flow for container request cancellations.

neprune:
[skip changelog]

for https://github.com/mysociety/societyworks/issues/4805

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -374,8 +374,21 @@ sub edit : Path('/admin/report_edit') : Args(1) {
         if ( $problem->state ne $old_state ) {
             $c->forward( '/admin/log_edit', [ $id, 'problem', 'state_change' ] );
 
+            my $updates = Open311::GetServiceRequestUpdates->new(
+                blank_updates_permitted => 1,
+            );
+
+            my ($description, $email_text);
+            foreach my $body (values %{$problem->bodies}) {
+                my $template = $problem->response_template_for($body, $problem->state, $old_state, '', '');
+                ($description, $email_text) = $updates->comment_text_for_request($template, {}, $problem);
+            }
+
+            my $text = join("\n\n", grep { $_ } ($c->stash->{update_text}, $description));
+
             $problem->add_to_comments( {
-                text => $c->stash->{update_text} || '',
+                text => $text,
+                $email_text ? (private_email_text => $email_text) : (),
                 user => $c->user->obj,
                 problem_state => $problem->state,
             } );
--- a/t/app/controller/waste_sutton_r.t
+++ b/t/app/controller/waste_sutton_r.t
@@ -1112,6 +1112,32 @@ FixMyStreet::override_config {
             };
         };
     }
+
+    subtest "Container request cancellation in admin" => sub {
+        my ($container_request_report) = $mech->create_problems_for_body(
+            1, $body->id,
+            'Container request', {
+                cobrand => 'sutton',
+                external_id => 'container-request',
+                cobrand_data => 'waste',
+                user => $user,
+                title => "Request replacement container",
+                category => "Request new container",
+            }
+        );
+        $container_request_report->update({ state => 'confirmed' });
+        $mech->log_in_ok($staff->email);
+        $mech->get_ok('/admin/report_edit/' . $container_request_report->id);
+        $mech->submit_form_ok({ with_fields => { state => 'cancelled' } });
+        $container_request_report->discard_changes;
+        is $container_request_report->state, 'cancelled';
+        my $latest_comment = $container_request_report->comments->search(
+                {},
+                { order_by => { -desc => 'id' } },
+        )->first;
+        is $latest_comment->text, "response template text", "cancel update uses response template";
+    };
+
 };
 
 sub get_report_from_redirect {


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -804,4 +804,34 @@ sub waste_munge_small_items_data {
...
+}
+
+=head2 waste_can_cancel_request
+
+Allows cancelling a request if the user is staff or the person that
+made the request, and the report is not already cancelled.
+
+=cut
+
+sub waste_can_cancel_request {
```
dracos: I'd probably have this as the default, rather than Sutton only, and add a feature flag check for something (`request_cancel_enabled` would match the bulky one), then it's easier to add anyone else in future.

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -804,4 +804,34 @@ sub waste_munge_small_items_data {
...
+=head2 waste_show_cancel_request
+
+Show the option to cancel a request for all users, if the report hasn't
+already been cancelled.
+
+=cut
+
+sub waste_show_cancel_request {
+    my ($self, $request_report) = @_;
+    return $request_report->state ne 'cancelled';
```
dracos: Same thing

perllib/FixMyStreet/App/Form/Waste/Request/Cancel.pm
```diff
@@ -0,0 +1,47 @@
...
+
+has_page intro => (
+    fields => ['confirm'],
+    finished => sub {
+        return $_[0]->wizard_finished('process_request_cancellation');
+    },
+    next => 'done',
+);
+
+has title => ( is => 'ro', 'isa' => 'Str', lazy => 1, builder => '_build_title' );
```
dracos, line 15: This is fine, as such a simple form - if you look at the garden forms and Form::Page::Waste (if you hadn't already) you can see how that inserts a service name into the title there that way, perhaps there's a neater way to bring these together, but don't think worth it for here.

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -286,7 +286,14 @@ sub bin_services_for_address {
         }
 
         # $containers should contain them all now (garden may only have been added just above here)
-        my $open_requests = { map { $_->{container} => $_ } $events->filter({ type => 'request', containers => $containers })->list };
+        my $open_requests = { map { $_->{container} => $_ } $events->filter(
+            {
+                type => 'request',
+                containers => $containers,
+                report_not_cancelled => 1,
```
dracos: Just checking, is this so we don't show reports that have been cancelled but aren't yet sent to Echo? Closed request will already be ignored because Events ignores closed request events by default. So I'm not sure if this is necessary, though I guess there could be this edge case.

dracos (changes requested):
Only real request is to make the Sutton specific things the default, with a feature flag; other two are just question/comments :)

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
neprune: Yeah exactly, I thought there might be a confusing period where the form is been submitted but Echo has not yet been updated. Have added a comment to spell this out more clearly.

perllib/FixMyStreet/Roles/Cobrand/Waste.pm
```diff
@@ -114,4 +114,46 @@ sub waste_sub_due {
...
+
+=cut
+
+sub waste_can_cancel_request {
+    my ($self, $request_report) = @_;
+
+    my $c = $self->{c};
+    return unless $c->stash->{waste_features}->{request_cancel_enabled};
+
+    return unless $request_report->state ne 'cancelled';
```
dracos: These lines could be replaced with `return unless $self->waste_show_cancel_request($request_report);` (might as well do the same check?)

dracos (approved):
Thanks :)

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -93,6 +93,7 @@ sub skip_alert_state_changed_to {
     my ( $self, $report ) = @_;
 
     return $report->category eq 'Small items collection';
+    return $report->category eq 'Request new container';
```
dracos: This will never be reached

neprune: 😭

perllib/FixMyStreet/App/Controller/Waste/Echo.pm
```diff
@@ -108,6 +108,15 @@ sub receive_echo_event_notification : Path('/waste/echo') : Args(0) {
                 $updates->suppress_alerts(1);
             }
 
+            # If it's a cancellation update on a container request
+            # that is already cancelled, don't send an alert as
+            # we would have already sent one out for this.
+            if ($p->category eq 'Request new container'
+                && $request->{status} eq 'cancelled'
+                && $p->state eq 'cancelled') {
+                $updates->suppress_alerts(1);
```
dracos, line 117: Just checking (I think we discussed this, but have forgotten) - this is just in case it's not already caught/hidden by the 'identical update' checking (now that we use the template)? Do we want to only suppress the alert, or if we're here, skip the update entirely?

neprune: Yeah on staging the update was still getting through, I guess because the template was not picked up so text wasn't matching, but I thought rather than figure that out better to just be explicit anyway. The reason I went for suppression rather than skipping was in case we wanted to keep a log on the report of when we get confirmation Echo that it was actually cancelled.

dracos: Oh yes, the template isn't being picked up because there *isn't* a state change (so it's not even looking); I wonder if it should be in that case, but that seems like a bigger thing.

Issue #5662, opened 2025-09-17
Remove Twitter from side menu


MyfanwyNixon:
There's a link to Twitter in the RH side menu of https://fixmystreet.org/overview/ but we don't use it any more. Thanks!

PR #5744, opened 2025-11-15
Remove twitter from doc

max246:
Please check the following:

- [x] Whether this PR should include changes to any public documentation, or the FAQ;

https://github.com/mysociety/fixmystreet/issues/5662

Thanks for contributing!


max246:
> Thanks for looking at this for us, looks good. We are currently using Bluesky, so it'd be good to add some links to them instead, but not suggesting you have to do that, just noting that's something we'd want to do.

If you have the icon and the link. I am happy to update the pr.
I just picked something simple to put my dive into the code and understand how it is all setup, but also try to help with some tickets.

PR #5745, opened 2025-11-17
[Waste] Promote UPRN to column

dracos:
This is split into two so can be deployed with zero downtime. The first commit adds a `uprn` column, and additionally stores the UPRN there on new waste reports. This commit can be deployed on a db that has the new column (with no data), then the data copied across for existing reports while live. Then once that's all in sync, the second commit can then stop saving the data in extra field/using that extra field anywhere. (Then the extra field data could be removed on existing reports, I guess, but no harm in keeping it.)

This should speed up all waste queries looking for a particular property (which is most/all of them).

[skip changelog]

Double request given it touches a lot of stuff!


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -121,6 +115,14 @@ __PACKAGE__->add_columns(
...
   "defect_type_id",
   { data_type => "integer", is_foreign_key => 1, is_nullable => 1 },
+  "extra",
+  { data_type => "jsonb", is_nullable => 1 },
+  "geocode",
+  { data_type => "jsonb", is_nullable => 1 },
+  "send_state",
+  { data_type => "text", default_value => "unprocessed", is_nullable => 0 },
+  "uprn",
+  { data_type => "varchar", is_nullable => 1, size => 12 },
```
davea: Is there any performance penalty if we instead used `text` here rather than limiting it to 12 chars? I'm just thinking ahead to the day we/a reuser tries to put something else in this field and it crashes :)

perllib/Open311.pm
```diff
@@ -201,6 +201,11 @@ sub _populate_service_request_params {
         $params->{deviceid} = $problem->service;
     }
 
+    # Always send through UPRN as an attribute if set on the report
+    if ($problem->uprn) {
```
davea, line 205: This seems like a change in behaviour, is it going to cause issues for Open311 endpoints that aren't ours?

db/schema.sql
```diff
@@ -199,6 +199,7 @@ create table problem (
     postcode text not null,
     latitude double precision not null,
     longitude double precision not null,
+    uprn varchar(12),
```
davea: Is there any performance penalty if we instead used `text` here rather than limiting it to 12 chars? I'm just thinking ahead to the day we/a reuser tries to put something else in this field and it crashes :)

perllib/Open311.pm
dracos: I don't think so; only waste will be setting the column, and there are no endpoints that aren't ours there.

db/schema.sql
dracos: There's no difference no, apart from postgresql doing a free length check for us. I guess I thought it was a bit clearer, given a UPRN can be at most 12 characters, but doesn't really matter, no.

PR #5747, opened 2025-11-20
[Kingston][Sutton] Remove fms emergency messaging admin

MorayMySoc:
Remove the fields from the emergency/site messaging admin that don't have a use for WasteWorks only cobrands.

https://mysocietysupport.freshdesk.com/a/tickets/6263

[skip changelog]

### DIFF
--- a/t/app/controller/admin/sitemessage.t
+++ b/t/app/controller/admin/sitemessage.t
@@ -8,7 +8,8 @@ my $bexley = $mech->create_body_ok(2494, 'Bexley Council', { cobrand => 'bexley'
 $mech->create_contact_ok(body_id => $bexley->id, category => 'Damaged road', email => "ROAD");
 my $body = $mech->create_body_ok(2237, 'Oxfordshire County Council', { cobrand => 'oxfordshire' });
 my $user = $mech->create_user_ok('user@example.com', name => 'Test User', from_body => $body);
-
+my $sutton = $mech->create_body_ok(2498, 'Sutton Borough Council', { cobrand => 'sutton' });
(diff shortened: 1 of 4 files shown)


t/app/controller/admin/sitemessage.t
```diff
@@ -59,6 +60,23 @@ FixMyStreet::override_config {
...
+    subtest "Sutton don't have Report or Homepage message form elements" => sub     {
+        $sutton_user->user_body_permissions->create({
+            body => $sutton,
+            permission_type => 'emergency_message_edit',
+        });
+        $mech->log_in_ok( $sutton_user->email );
+        $mech->get_ok('/admin/sitemessage');
+        $mech->content_lacks('Reporting page');
+        $mech->content_lacks('Homepage');
+        $mech->content_contains('Waste message');
```
davea, line 76: Would be good to test form submission and resulting DB changes too.

templates/web/base/admin/sitemessage/edit.html
```diff
@@ -11,6 +11,8 @@
 }
 </style>
 
+[% IF body_cobrand.moniker != 'kingston' AND body_cobrand.moniker != 'sutton' %]
+
```
davea, line 15: This change removes the `<form>` element from the page, so you can't save the message :) Might need to rejig things a bit there!

Also, I wonder if it's worth factoring out the logic here so there's nothing cobrand-specific in the template - maybe add something like `sub wasteworks_only { 1 }` to `Cobrand::KingstonSutton`, which would also make it easy to extend this change to future WW-only cobrands.

MorayMySoc: ooops!

PR #5750, opened 2025-11-21
[Dumfries] Further cobrand setup and Alloy integration

davea:
More cobrand configuration (e.g. privacy policy URL, turning off questionnaires) and Open311 configuration (e.g. site code lookup, parsing incoming `extras` and setting inspection date and disallowing updates based on that)

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Auth/Social.pm
+++ b/perllib/FixMyStreet/App/Controller/Auth/Social.pm
@@ -352,7 +352,7 @@ sub oidc_callback: Path('/auth/OIDC') : Args(0) {
     $c->session->{oauth}{id_token} = $id_token->token_string;
 
     # Cobrands can use different fields for name and email
-    my ($name, $email) = $c->cobrand->call_hook(user_from_oidc => $id_token->payload, $access_token);
+    my ($name, $email, $phone) = $c->cobrand->call_hook(user_from_oidc => $id_token->payload, $access_token);
     $name = '' if $name && $name !~ /\w/;
 
     # There's a chance that a user may have multiple OIDC logins, so build a namespaced uid to prevent collisions
@@ -365,7 +365,7 @@ sub oidc_callback: Path('/auth/OIDC') : Args(0) {
     # TfL would like to set the user's roles when logging in through oidc
     my $roles = $id_token->payload->{roles} || ();
 
-    $c->forward('oauth_success', [ 'oidc', $uid, $name, $email, $extra, $roles ]);
+    $c->forward('oauth_success', [ 'oidc', $uid, $name, $email, $extra, $roles, $phone ]);
 }
 
 # Just a wrapper around random_token to make mocking easier.
@@ -389,7 +389,7 @@ sub oauth_failure : Private {
 }
 
 sub oauth_success : Private {
-    my ($self, $c, $type, $uid, $name, $email, $extra, $roles) = @_;
+    my ($self, $c, $type, $uid, $name, $email, $extra, $roles, $phone) = @_;
 
     my $user;
     if ($email) {
@@ -416,6 +416,7 @@ sub oauth_success : Private {
             $user->add_oidc_id($uid);
         }
         $user->name($name) if $name;
+        $user->phone($phone) if $phone;
         if ($extra) {
             $user->extra({
                 %{ $user->get_extra() },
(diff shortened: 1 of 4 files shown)


templates/web/dumfries/report/new/roads_message.html
```diff
@@ -0,0 +1,5 @@
+<div id="js-not-a-road" class="hidden js-responsibility-message">
```
dracos: Ideally we don't add this for new cobrands, but use `no_asset_message` instead solely in the config - is that okay here? See e.g. vhosts/fixmystreet/assets/aberdeenshire.yml.include for an example.

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -564,7 +564,7 @@ sub check_recaptcha {
 
 sub public_holidays {
     my $self = shift;
-    my $nation = shift || 'england-and-wales';
+    my $nation = shift || ($self->can('is_scotland') && $self->is_scotland ? 'scotland' : 'england-and-wales');
```
dracos: Shall we add a default is_scotland 0 to UK.pm? Guess this is only use, so doesn't matter much at this point.

perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -62,6 +63,19 @@ sub open311_get_update_munging {
             $text, $request );
         $comment->private_email_text($text);
     }
+
+    # If the update includes a latest_inspection_time, store it on the problem
+    # If the value is 'NOT COMPLETE', unset the metadata
+    if (exists $request->{extras}{latest_inspection_time}) {
```
dracos: Don't need exists here, you don't want to do this if it's there but undefined (and presumably it can't be that).

perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -151,4 +165,60 @@ sub _fetch_features_url {
 sub contact_extra_fields { [ 'display_name' ] }
 
 
+=head2 _updates_disallowed_check
+
+Updates are only allowed on reports in a closed state (closed, duplicate, etc.),
```
dracos: Presumably superseded somewhere, but your comment at https://github.com/mysociety/societyworks/issues/5252#issuecomment-3768559755 also says if the problem is in an open state.

perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -151,4 +165,60 @@ sub _fetch_features_url {
...
+        return 1; # If we can't parse the date, disallow updates
+    }
+
+    my $cutoff = DateTime->now(time_zone => FixMyStreet->local_time_zone)->subtract(days => 14);
+    if ($inspection_dt > $cutoff) {
+        return 1;
+    }
+
+    # Only staff or the original reporter can leave updates
+    unless ($staff || $reporter) {
```
dracos: Can this check come sooner before the date checking?

templates/web/dumfries/report/new/roads_message.html
davea: Sorry, we did discuss this earlier didn't we! Yes, no reason why we can't do the same as Aberdeenshire here.

perllib/FixMyStreet/Cobrand/UK.pm
davea: I was worried about the political tensions of something like that ;) But I suppose it does make sense from a code point of view.

perllib/FixMyStreet/Cobrand/Dumfries.pm
davea: Ooh good catch, in fact the comment on the github ticket was correct and the code wasn't! Latest discussion was [here](https://3.basecamp.com/4020879/buckets/44203298/card_tables/cards/9637620417) which confirms updates only allowed on open reports. Thanks for spotting.

perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -198,13 +197,16 @@ sub _updates_disallowed_check {
     my $superuser = $c->user_exists && $c->user->is_superuser;
     my $staff = $body_user || $superuser;
```
dracos: Don't need this line any more, then, if staff can't leave updates at all.

PR #5752, opened 2025-11-24
[Bexley] Add contract ID lookup for legacy subscriptions

chrismytton:
Add `BexleyContracts` module and `make-bexley-contract-db` script to handle UPRN to contract ID lookups for garden waste subscriptions created before the WasteWorks integration.

First checks the report for a matching DD contract ID, indicating it was created by WW, and if found uses that, otherwise looks up legacy contract IDs in this database and attempts to cancel all of those. Also works for the cancellation script that runs nightly.

Fixes https://github.com/mysociety/societyworks/issues/5247

For [this Basecamp thread](https://3.basecamp.com/4020879/buckets/43587866/card_tables/cards/9251295180).

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste.pm
@@ -1155,6 +1155,11 @@ sub get_current_payment_method : Private {
         $payment_method = $c->stash->{orig_sub}->get_extra_field_value('payment_method');
     }
 
+    # Allow cobrand to override payment method detection for legacy subscriptions
+    if (!$payment_method) {
+        $payment_method = $c->cobrand->call_hook('waste_get_current_payment_method' => $c->stash->{orig_sub});
+    }
+
     return $payment_method || 'credit_card';
 
 }
--- a/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
@@ -294,6 +294,35 @@ sub waste_get_legacy_contract_ids {
     return $contract_ids;
 }
 
+=head2 waste_get_current_payment_method
+
+For legacy pre-WasteWorks subscriptions without stored payment_method,
+check if there are legacy contracts in the BexleyContracts database for this UPRN.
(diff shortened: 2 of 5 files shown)


bin/bexley/make-bexley-contract-db
```diff
@@ -0,0 +1,130 @@
...
+
+$db->begin_work;
+
+while (my $row = $csv->getline_hr($fh)) {
+    my $uprn = $row->{UPRN};
+    my $contract_id = $row->{'Contract ID'};
+    my $bank_ref = $row->{'Bank Reference'};
+
+    # Skip rows with missing UPRN
+    if (!defined $uprn || $uprn eq '') {
```
dracos: I think `if (!$uprn)` would do here? There's never going to be a "0" UPRN.

perllib/BexleyContracts.pm
```diff
@@ -0,0 +1,77 @@
...
+
+    my $db = connect_db() or return [];
+
+    my $contracts = $db->selectall_arrayref(
+        <<"SQL",
+  SELECT contract_id
+    FROM contracts
+   WHERE uprn = ?
+SQL
+        {},
```
dracos: Does this need to be { Slice => {} } in order to work? See BexleyAddresses. Otherwise it won't be returning a hashref of each row, and will all be blank.
Not caught by tests because this is being mocked out to return a hashref.

chrismytton: Ahhh, I did have the Slice bit in there (because I cargo-culted it from BexleyAddresses) but I figured I didn't need it because the tests passed without it…forgot about the mock though 🤦

chrismytton:
I think there's a slightly more fundamental flaw which I've uncovered while investigating FD-6296 which is that when a cancellation report is created and we don't have the original report we're defaulting the payment method to `credit_card`, which means this logic is skipped entirely (also not caught by the tests…). Think the fix is to do the lookup earlier in the process and default to DD if there are any matching contracts in the legacy database.

dracos (approved):
Good catch, hopefully that will deal with any more of these.
Can we check it would have dealt with the ones that raised this - can we manually pass in something to see if it then runs the cancellation?

PR #5753, opened 2025-11-26
[Hackney] Add canals and rivertrust layer

MorayMySoc:
Configuration in same branch name on servers

Branch name says camden - but it's actually a change for Hackney

https://mysocietysupport.freshdesk.com/a/tickets/6279

[skip changelog]


chrismytton (approved):
For the servers config we got rid of `all_categories` recently as it was always tracking `always_visible`, so that line can be removed. (I only discovered this recently when Matthew left a similar comment on one of my asset layer changes!)

PR #5756, opened 2025-11-27
[Admin] Fix tabs when report has no updates

davea:
Odd behaviour and JS errors in console:

tabs

[skip changelog]


dracos (approved):
Aha, thanks!

PR #5762, opened 2025-12-04
[Kingston] Rebranding 2025

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5242

**Preview:**

https://github.com/user-attachments/assets/e128512d-1f06-424e-ab58-520b5403c3bc

[skip changelog]


web/cobrands/kingston/images/kingston-logo.png
dracos, line 1: Is this not a bit big if it's only going to be shown at 92x112?

templates/web/kingston/waste/site-name.html
```diff
@@ -0,0 +1 @@
+Bins, rubbish and recycling | Royal Borough of Kingston upon Thames
```
dracos, line 1: Hmm, this is used as the name in the https://kingston.staging.societyworks.org/about/waste-web-app page, which now reads a bit oddly! Guess we can put an override in templates/web/fixmystreet-uk-councils/about/waste-web-app.html to change it for kingston is probably easiest?

templates/web/kingston/header_site.html
```diff
@@ -0,0 +1,15 @@
+<header id="site-header" role="banner">
+    <div class="container">
+        [% INCLUDE 'header_logo.html' %]
+        <label id="nav-link" for="main-nav-btn">
+            <span class="nav-link-text nav-closed-text">[% loc('Menu') %] </span>
```
dracos, line 5: Do you think this should apply to everything, not just kingston?

lucascumsille: We could do it, not all users might know the icon's meaning, so it would be an improvement. I'll create a ticket for that though, because some cobrands might have overlapping issues between wide logos, report button and now the menu.

templates/web/kingston/waste/site-name.html
lucascumsille: Thanks I just Implemented the solution above. Let me know if you'd like any changes and also if we should keep that commit separate or squash it into the main commit.

PR #5763, opened 2025-12-04
Refactor front_stats_data to use static JSON files

davea:
Splits the front_stats_data method into two:
- store_front_stats_data: calculates stats and writes to JSON file
- front_stats_data: reads stats from JSON file

Adds bin/generate-front-stats script to generate stats for all cobrands. This allows stats to be generated once (e.g. via cron) rather than on page load.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Default.pm
+++ b/perllib/FixMyStreet/Cobrand/Default.pm
@@ -10,6 +10,8 @@ use FixMyStreet::Geocode::Address;
 use FixMyStreet::OutOfHours;
 use DateTime;
 use JSON::MaybeXS;
+use Path::Tiny;
+use Try::Tiny;
 use List::MoreUtils 'none';
 use URI;
 use Digest::MD5 qw(md5_hex);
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -445,9 +447,46 @@ sub front_stats_data {
...
+
+        # Don't show stats that are more than 48 hours old.
+        if (my $generated = delete $stats->{_generated}) {
+            my $age = time() - $generated;
+            return undef if $age > 172800;
+        }
+
+        return $stats;
+    } catch {
+        return undef;
```
dracos: This doesn't do what you think, you can't return from inside a catch.
I don't think a catch is needed, as we're not doing anything if it goes wrong.
The `return undef`s could just be `return` inside the try (which again only return to outside the try block), as the return value isn't being captured.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -445,9 +447,46 @@ sub front_stats_data {
...
+    };
 }
```
dracos, line 495: Could we fall back to doing the lookup live if the file doesn't exist? I realise this could potentially lead to an issue, but seems low risk, we can generate the files first, plus then we don't break all the other users of FixMyStreet who don't need this to run like this.

perllib/FixMyStreet/Cobrand/Default.pm
```diff
@@ -445,9 +447,46 @@ sub front_stats_data {
         updates => $updates,
         new     => $new,
         recency => $recency,
+        _generated => time(),
     };
```
dracos: So I guess having a third function that does the actual stats generation would make that easier.

perllib/FixMyStreet/Cobrand/Default.pm
davea: Do you think it'd be worth putting the memcached stuff back in (behind a flag somehow) in that case?

perllib/FixMyStreet/Cobrand/Default.pm
davea: Oops, good catch (😉). I added the `try`/`catch` later on after thinking it'd be worth making the failure mode a bit less crashy.

perllib/FixMyStreet/Cobrand/Default.pm
dracos: Hmm, guess it can't hurt really, can it, if that's straightforward. Wouldn't need a flag, doesn't really matter if we also store it in memcache for no reason, it won't get looked up in there

t/cobrand/hounslow.t
```diff
@@ -1,3 +1,6 @@
+use Test::MockModule;
```
dracos: No longer need to add these to this file.

dracos (approved):
Thanks :)

davea:
> entries that are in ALLOWED_COBRANDS more than once are processed more than once

Ah, yes, should probably have done a uniqueness check in the loop! Having tested locally, at least the second iteration is super fast because it's coming from memcached rather than being calculated again :)

PR #5767, opened 2025-12-08
[Waste] WasteWorks specific offline fallback page

struan:
Adds a waste specific offline fallback page at `/offline/waste_fallback` and updates the service worker to display this as a fallback if the path includes `/waste`.

Screenshot 2025-12-09 at 11 54 12

Fixes [#societyworks/4655](https://github.com/mysociety/societyworks/issues/4655)

### DIFF
--- a/templates/web/base/offline/fallback.html
+++ b/templates/web/base/offline/fallback.html
@@ -4,7 +4,7 @@
 <h1>[% loc('You are currently offline') %]</h1>
 
 <p>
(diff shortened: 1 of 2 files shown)


templates/web/base/offline/service_worker.html
```diff
@@ -43,6 +45,11 @@
       return;
   }
 
+  let offline_page = "[% offline_html %]";
+  if (RegExp('/waste').test(url)) {
```
dracos, line 49: Didn't realise you could call RegExp without `new`! TIL.

struan:
Pushed an update to the fallback templates language.

PR #5768, opened 2025-12-09
[Kingston] add flow for reporting bins not returned or contents spilled

struan:
This is mostly based on the [Sutton code](https://github.com/mysociety/fixmystreet/commit/0cc3311cf5) with a few Kingston specific changes to add checkbox extra details to the bin not returned, plus an extra message for FAS properties for spillages.   

Fixes mysociety/societyworks#5245
Fixes mysociety/societyworks#5246

[skip changelog]


templates/email/default/waste/other-reported.txt
```diff
@@ -55,6 +55,14 @@ On delivery day - You do not need to be home for our crew to deliver the bin.
 Collecting your old bin - If you've requested a collection of your old bin, we'll schedule a specific date for our
 crew to remove it. We aim for this to be the same day as we deliver your new bin, but it could be a different day.
 
+[% ELSIF cobrand.moniker == 'kingston' AND report.category == 'Waste spillage' %]
```
dracos: Any change you make to a .txt email template you'll probably have to make the same change to the .html version as well

dracos: Also could factor it together with the Sutton ones in this file as they're identical - but I guess might change 🤷

templates/web/kingston/waste/enquiry-problem.html
```diff
@@ -0,0 +1,29 @@
+[%
```
dracos: If this is identical to Sutton, do we want to make it a symlink rather than a copy? We've done that in a couple of other places.

templates/web/base/waste/services.html
```diff
@@ -68,6 +68,19 @@
     [% END %]
 [% END %]
 
+[% IF c.cobrand.moniker == 'kingston' AND (property.domestic_refuse_bin OR communal_property OR fas_property) %]
+    [% IF unit.escalations.missed_open %]
```
dracos: They don't have escalations (yet, so guess can leave this in, but just for clarity)

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -623,6 +623,130 @@ sub admin_fee_cost {
     return $cost;
 }
 
+
+=head2 waste_munge_enquiry_form_pages
```
dracos: This all seems pretty much identical to the Sutton code, so could we instead move the Sutton code into Roles/Cobrand/KingstonSutton.pm and deal with any differences (if any) in it?

dracos (changes requested):
Looks good, some refactoring mostly I think :)

perllib/FixMyStreet/Cobrand/Kingston.pm
struan: I've left `waste_munge_enquiry_data` in the specific Cobrand files as it's different enough at the moment that it makes more sense to split it.

I've also done a small amount of input sanitising to avoid casing issues which in theory aren't really necessary as we should control everything but I figured aren't harmful either.

templates/email/default/waste/other-reported.txt
```diff
@@ -48,7 +48,8 @@ We apologise for any inconvenience this may have caused and appreciate your feed
 Please note that we are unable to return to correct the problem, unless you are on the Assisted Collection service and you have indicated you require additional help.
 [% END %]
 
-[% ELSIF cobrand.moniker == 'kingston' AND report.category == 'Request new container' %]
+[% ELSIF cobrand.moniker == 'kingston'%]
```
dracos, line 68: There's one more kingston check unneeded in the second sub-IF.

templates/email/default/waste/other-reported.html
```diff
@@ -64,16 +64,32 @@ <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
...
-  <p style="[% p_style %]">
-    <b>On delivery day</b><br>
-    You do not need to be home for our crew to deliver the bin.
-  </p>
-  <p style="[% p_style %]">
-    <b>Collecting your old bin</b>
-    If you&apos;ve requested a collection of your old bin, we'll schedule a specific date for our crew
-    to remove it. We aim for this to be the same day as we deliver your new bin, but it could be a different day.
-  </p>
+[% ELSIF cobrand.moniker == 'kingston'%]
```
dracos, line 74: Almost, but due to this we've lost the "We will arrange another collection as soon as possible over the next 2 working days." output for Kingston in the two missed categories, that needs getting back in.

templates/web/kingston/waste/enquiry-problem.html
```diff
@@ -13,10 +13,7 @@
 END;
 
 IF service.report_allowed AND property.domestic_refuse_bin;
-    SET field = {
```
dracos: If this was meant to make the file a symlink, I don't think it's worked, dunno if there's some git thing going on here.

templates/email/default/waste/other-reported.txt
dracos: And same issue with missed collections as the HTML

templates/web/kingston/waste/enquiry-problem.html
dracos: (Don't think this has worked)

dracos (changes requested):
Tiny issue with the email template to resolve, symlink thing I'm sure we can fix.

dracos (approved):
Woo; still got that superfluous `ELSIF cobrand.moniker == 'kingston'` in the text email but doesn't matter if it's left in :)

MorayMySoc:
Rebased onto the Sutton 'Report a problem' page work and fixup is to align the process with that

templates/web/sutton/waste/enquiry-problem.html
```diff
@@ -94,7 +110,10 @@
     f.push(field);
 
     SET field = { value = 'Waste spillage', label = 'Report a spillage by the crew' };
-    IF service.report_allowed AND property.domestic_refuse_bin;
+    IF c.cobrand.moniker == 'kingston' AND fas_property;
+        field.hint = 'Follow the link in the introduction to request a street clean';
```
dracos: Not sure I'd bother with the hint.
(It's possible from the KGS272 internal comparison doc that they're happy for FAS to continue here, unclear if that changes the previous request...)

templates/web/sutton/waste/enquiry-problem.html
```diff
@@ -94,7 +110,10 @@
...
+        field.disabled = 1;
+    ELSIF service.report_allowed AND property.domestic_refuse_bin;
```
dracos: I don't think the IF domestic should be here, sorry, missed that in previous PR, otherwise this just means it isn't checking for 'already reported a waste spillage' for communal/FAS properties.

templates/web/sutton/waste/enquiry-problem.html
```diff
@@ -33,11 +33,27 @@
     [% ELSIF service.report_locked_out %]
         <p>The crew have closed your collection task as not collected[%
             IF service.report_locked_out_reason %]: [% service.report_locked_out_reason %][% END %].
-    [% ELSIF service.report_allowed %]
+    [% ELSIF c.cobrand.moniker == 'sutton' AND service.report_allowed %]
         <p>Our records show this collection took place
         on [% date.format(service.last.date) | replace('~~~', service.last.ordinal) %].
         If it has not been collected, you can report it as missed and we will return within 2 working days to collect it.
         </p>
+    [% ELSIF c.cobrand.moniker == 'kingston' AND (property.domestic_refuse_bin OR communal_property OR fas_property)%]
```
dracos: Few things on this section:
a) Think "property.domestic_refuse_bin OR communal_property OR fas_property" covers all properties? :)
b) Link for Kingston has a typo in 'street'
c) I don't think all this text should be here. It can be shown after they pick the category, as an extra data notice on the category (which is what currently happens on live, see the live Sutton "Waste spillage" category for example), and this can then be simplified and not repeat the date part.
d) The text here isn't what's given in the final document https://docs.google.com/document/d/193SMUDL-JOlm8qi0_XXgsdHO_OWBgQ0E58naNsw8cWE/edit?tab=t.0 under "Spillage by Crew" - so that text would be what's then added as an extra data notice on each category.
e) Then as per above doc, once that's done, I think https://github.com/mysociety/societyworks/issues/5575 then adds on nicely, in that the template needs updating to stay the same for Sutton but to always show the Bin not returned/Spillage options for Kingston.

So this bit I'd have something like
```
    [% ELSIF service.report_allowed %]
        <p>Our records show this collection took place
        on [% date.format(service.last.date) | replace('~~~', service.last.ordinal) %].
        If it has not been collected, you can report it as missed and we will return within 2 working days to collect it.
        </p>
        [% IF cobrand.moniker == 'kingston' AND fas_property %]
            <p>If a spillage on a public highway occurred during a [% unit.service_name FILTER lower %] collection, please report <a href="https://kingston-self.achieveservice.com/service/Request_a_street_clean_in_Kingston">a street cleaning issue</a>.</p>
       [% END %]
```

dracos (changes requested):

PR #5774, opened 2025-12-11
[Internal] Replace ids with urls in send-failures log

MorayMySoc:
Preliminary effort to make log message more practical.

https://github.com/mysociety/societyworks/issues/4984

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Script/Reports.pm
+++ b/perllib/FixMyStreet/Script/Reports.pm
@@ -129,10 +129,10 @@ sub end_summary_failures {
     })->order_by('-confirmed');
     my %bodies;
     while (my $row = $unsent->next) {
-        my $base_url = FixMyStreet->config('BASE_URL');
         my $key =  join ', ', @{ $row->body_names };
         $bodies{$key} ||= [];
-        push @{ $bodies{$key} }, $row->id;
+        my $base_url = get_base($self, $row);
+        push @{ $bodies{$key} }, $base_url . "/admin/report_edit/" . $row->id;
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Script/Reports.pm
```diff
@@ -132,7 +132,7 @@ sub end_summary_failures {
         my $base_url = FixMyStreet->config('BASE_URL');
         my $key =  join ', ', @{ $row->body_names };
         $bodies{$key} ||= [];
-        push @{ $bodies{$key} }, $row->id;
+        push @{ $bodies{$key} }, "$base_url/admin/report_edit/" . $row->id;
```
dracos: For waste, or TfL reports (where the BASE_URL won't work and only give a 404, dunno if there are other cases), could this know that and provide a working link?

perllib/FixMyStreet/Script/Reports.pm
```diff
@@ -154,4 +158,15 @@ sub log {
     STDERR->print("[fmsd] $msg\n");
 }
 
+sub get_base {
+    my ($self, $row) = @_;
+
+    my $base;
+    if ($row->contact->body->name eq 'TfL' || ($row->contact->extra && $row->contact->get_extra_metadata('type') eq 'waste')) {
```
dracos: As all waste reports are logged on their own cobrand, and TfL ones logged on .com are still viewable there, I think this could be something like:

`my $base = ($row->cobrand eq 'tfl' || $row->cobrand_data eq 'waste') ? $row->get_cobrand_logged->base_url : FixMyStreet->config('BASE_URL');`

PR #5775, opened 2025-12-11
[Bucks] Automatically set reports about flags to private

nephila-nacrea:
For https://mysocietysupport.freshdesk.com/a/tickets/6339.

For deployment:
- [ ] Set `is_flag` question on 'Unauthorised signs' category, on
    - [x] staging
    - [ ] live 

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
@@ -383,13 +383,20 @@ sub open311_contact_meta_override {
 sub report_new_munge_before_insert {
     my ($self, $report) = @_;
 
-    return unless $report->category eq 'Flytipping';
     return unless $self->{c}->stash->{report}->to_body_named('Buckinghamshire');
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Buckinghamshire.pm
```diff
@@ -383,13 +383,20 @@ sub open311_contact_meta_override {
...
     my ($self, $report) = @_;
 
-    return unless $report->category eq 'Flytipping';
     return unless $self->{c}->stash->{report}->to_body_named('Buckinghamshire');
 
-    my $placement = $self->{c}->get_param('road-placement');
-    return unless $placement && $placement eq 'off-road';
+    if ( $report->category eq 'Unauthorised signs' ) {
+        # If they are reporting flags, automatically mark as private
+        my $is_flag = $report->get_extra_field_value('is_flag');
```
dracos: I'd probably add a default here just in case the field doesn't exist to prevent an uninitialized comparison warning in the next line; will add that.

PR #5776, opened 2025-12-12
[CentralBeds] Fix updated tree categories.

neprune:
[skip changelog]

There are now many tree related categories under a "Trees" group, rather than one "Trees" category. Update the road requirement check to reflect this. Also add a custom relevance function so layer applies both to these new categories and the existing "Communal Trees" category under the "Housing" group.

Closes https://github.com/mysociety/societyworks/issues/5298.

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/assets.js
+++ b/web/cobrands/fixmystreet-uk-councils/assets.js
@@ -657,13 +657,12 @@ var centralbeds_types = [
 
 function cb_should_not_require_road() {
     // Ensure the user can select anywhere on the map if they want to
-    // make a report in the "Trees", "Fly Tipping" or "Public Rights of way" categories.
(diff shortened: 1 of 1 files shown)


web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -708,6 +707,11 @@ fixmystreet.assets.centralbedfordshire.not_found = function(layer) {
     }
 };
 
+fixmystreet.assets.centralbedfordshire.trees_relevant = function(options) {
+    return options.category === "Communal Trees" ||
```
dracos: I think you could have asset_group: ['Trees', 'Communal Trees'] in the config - as that checks both group and caetgory, and not need this function, but perhaps having the function is clearer...

neprune: Went for dropping the function in favour of the config approach

PR #5778, opened 2025-12-18
[Shropshire] Add parishes to dashboard filters

dracos:
And I guess Oxfordshire will get it as well! [skip changelog] For FD-6399.

Also fixed a fixture running bug, as recently added category to the base list didn't have any fixture titles.

Also fixes the display of the filters on the heatmap which looked a bit ropey:
Before:
image
After:
image


davea (approved):
🗺️ 🆗

PR #5782, opened 2025-12-18
[Bromley] [DD] Don’t warn about already processed hidden reports

davea:
Prevents warnings about subscriptions which were manually marked as hidden after the DD payment was successfully processed.

[skip changelog]


dracos (approved):
Thanks :)

PR #5784, opened 2025-12-18
[Central Beds] Update cobrand for migration to Aurora.

neprune:
Removing logic that we should no longer need and generally tidying up. Relates to https://github.com/mysociety/societyworks/issues/5288 and https://github.com/mysociety/societyworks/issues/5289.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
@@ -145,6 +145,12 @@ sub open311_config {
     $params->{upload_files} = 1;
 }
 
+sub open311_munge_update_params {
+    my ($self, $params, $comment, $body) = @_;
+    my $contact = $comment->problem->contact;
+    $params->{service_code} = $contact->email;
+}
+
 =head2 should_skip_sending_update
 
 Do not try and send updates to the Jadu backend.


perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -218,10 +207,6 @@ sub open311_extra_data_include {
     }
 
     my $open311_only = [
-        { name => 'title',
```
dracos, line 223: Both these are used by the Jadu open311-adapter code?

perllib/FixMyStreet/Cobrand/CentralBedfordshire.pm
```diff
@@ -146,20 +147,9 @@ sub open311_config {
...
 }
 
-sub open311_munge_update_params {
-    my ($self, $params, $comment, $body) = @_;
-
-    # TODO: This is the same as Bexley - could be factored into its own Role.
-    $params->{service_request_id_ext} = $comment->problem->id;
-
-    my $contact = $comment->problem->contact;
-    $params->{service_code} = $contact->email;
```
dracos, line 156: We need to include the service code, so that open311-adapter Multi knows which backend to send it to.

dracos (changes requested):
I think we need to keep some things that have been removed, unless I've misunderstood.

dracos (approved):
Thanks :)

PR #5786, opened 2025-12-18
Allow superusers to change body of failed reports

davea:
Replaces the body name on the admin report edit page with a dropdown, allowing superusers to change the destination body for a failed report.

This is intended primarily for the situation where a report has been associated with the wrong body when created (e.g. some NH reports on FMS.com) and needs manually changing to the correct body.

Deliberately limited in scope, the dropdown only appears for reports that:

 - haven't yet been sent,
 - have failed to send at least once,
 - are for a single body, and
 - are at a location that has another body they can be asssigned to.

Any changes made are added to the admin log for the report.

dropdown2

[skip changelog]


MorayMySoc (approved):
Looks good to me and very handy!

PR #5787, opened 2025-12-18
[Kingston] [WW] Escalations and target dates

struan:
Fixes mysociety/societyworks#5257

Update Kingson Echo integration to support escalations and target dates.

The escalations code is much the same as Sutton, so the code for that has been hoisted up into the shared role, with a few configuration functions to cover differences in timelines.

This also adds target dates for Kingston to enable specific dates to be displayed on the bin page for things like container delivers, missed bin resolutions etc rather than "within n days".

[skip changelog]


dracos:
(Question aside not to do with these changes for @sallytay here - as Kingston asked us to add the specific dates here (which we'd refused to do for Sutton previously when they tried to just shoehorn it in), do we want to have this for Sutton now as well given we've done the work - I'm guessing so?)

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -595,6 +601,92 @@ sub waste_munge_request_data {
...
+
+=cut
+
+sub waste_target_days {
+    {
+        container_escalation => 5,
+        missed => 2,
+        missed_escalation => 1,
+        missed_bulky => 2,
+        missed_bulky_escalation => 1,
```
chrismytton: Should this be 2 days, as per your comment?

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -863,6 +723,48 @@ sub waste_munge_enquiry_data {
...
+
+=cut
+
+sub waste_target_days {
+    {
+        container_escalation => 5,
+        missed => 2,
+        missed_escalation => 1,
+        missed_bulky => 2,
+        missed_bulky_escalation => 1,
```
chrismytton: And here, should this be 2 days?

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -169,6 +169,170 @@ sub waste_report_form_first_next {
     };
 }
 
+=head2 Escalations
+
+Sutton has custom behaviour to allow escalation of unresolved missed collections
+or container requests.
```
chrismytton: Might be worth updating this comment, to avoid confusing future readers.

chrismytton (changes requested):
Looks good! Just a minor question about SLAs and one comment update suggestion.

chrismytton (approved):
Looks good, thanks!

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -169,6 +169,170 @@ sub waste_report_form_first_next {
...
+            my $start = $wd->add_days($missed_event->{date}, $self->waste_escalation_window->{bulky_start})->set_hour($self->waste_day_end_hour);
+            my $end = $wd->add_days($start, $self->waste_escalation_window->{bulky_length});
+            if ($now >= $start && $now < $end) {
+                $missed->{$guid}{escalations}{missed} = $missed_event;
+            }
+        } elsif ($open_escalation) {
+            if ($self->waste_target_days && $self->waste_target_days->{missed_bulky_escalation}) {
+                $missed_event->{target} = $wd->add_days($escalation_event->{date}, $self->waste_target_days->{missed_bulky_escalation})->set_hour($self->waste_day_end_hour);
+            }
+            $missed->{$guid}{escalations}{missed_open} = $missed_event;
```
dracos: I don't understand what's happening here. Above we've already set missed_open to the relevant open escalation event, but here we're overwriting it with the original missed collection event?

dracos: I've changed the target setting couple of lines above to in escalation_event and removed this line. Can discuss if I've misunderstood, but it fixes a failing test in my current work

PR #5793, opened 2025-12-23
Waste nicer mobile view

lucascumsille:
Fixes part of: https://github.com/mysociety/fixmystreet/pull/5730

This PR Implements the changes mentioned in [here](https://github.com/mysociety/fixmystreet/pull/5730#issuecomment-3580439098):

@dracos  In ended up having to make some changes to the nesting/structure of the HTML, because just overriding the govuk classes for their grid/layout system wasn't going to be enough to accomplish the layout we wanted.

Just in case the second commit is entirely optional and we can drop it.

### Preview

https://github.com/user-attachments/assets/5d91c0ee-686c-4338-909c-78178e42e626

https://github.com/user-attachments/assets/9e663d2e-c8c1-435a-951d-a238130df290

https://github.com/user-attachments/assets/dde23d40-ed0f-43dd-8f74-73fd99999e63

https://github.com/user-attachments/assets/1b214289-b42b-4875-a816-97963732641b

[skip changelog]

### DIFF
--- a/templates/web/base/waste/bin_days.html
+++ b/templates/web/base/waste/bin_days.html
@@ -13,7 +13,7 @@
 
 <div class="waste__collections">
   [% IF c.cobrand.moniker != 'peterborough' %]
-    <h2 class="govuk-heading-l govuk-!-margin-bottom-2">Your collections</h2>
+    <h2 class="govuk-heading-l govuk-!-margin-bottom-2 govuk-!-margin-top-8">Your collections</h2>
     [% IF c.cobrand.moniker == 'brent' && service_data.size %]
       [% FOR unit IN service_data %]
         [% IF unit.service_id == 262 ~%]
@@ -25,7 +25,7 @@
     [% END %]
   [% ELSE %]
     <div class="govuk-!-margin-bottom-8">
-      <h2 class="govuk-heading-l govuk-!-margin-bottom-2">Your collections</h2>
+      <h2 class="govuk-heading-l govuk-!-margin-bottom-2 govuk-!-margin-top-8">Your collections</h2>
       [% IF service_data.size %]
         [% TRY %][% PROCESS waste/_calendar.html %][% CATCH file %]
           <a href="[% c.uri_for_action('waste/calendar', [ property.id ]) %]">Add to your calendar</a>
--- a/web/cobrands/sass/_waste.scss
+++ b/web/cobrands/sass/_waste.scss
@@ -192,8 +192,10 @@ body.waste {
 
 .waste-service-name.govuk-heading-m {
   margin-top: 1.5em;
-  border-bottom: 1px solid #808080;
   padding-bottom: 0.5em;
+  @media (min-width: $large-screen) {
+    border-bottom: 1px solid #808080;
+  }
 }
 
 .waste-service-wrapper {


templates/web/bexley/waste/services_extra.html
```diff
@@ -1,13 +1,12 @@
-<div class="govuk-grid-row" id="in-cab-logs">
   [% IF property.garden_signup_eligible %]
   <h3 class="govuk-heading-m waste-service-name">
     Brown Wheelie Bin
   </h3>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-one-quarter text-centered">
+    <div class="govuk-grid-column govuk-grid-column-always-one-quarter text-centered">
```
dracos: I also changed bexley and bromley services_extra - should these presumably also be changed to match your changes?

templates/web/base/waste/bin_days.html
```diff
@@ -3,289 +3,314 @@
...
-          <img src="[% image.src %].png" srcset="[% image.src %].png 1x, [% image.src %]@2x.png 2x" alt="[% image.alt %]" class="waste-service-image">
-        [% ELSE  %]
-          <img src="[% image %].png" srcset="[% image %].png 1x, [% image %]@2x.png 2x" alt="" class="waste-service-image">
-        [% END %]
-        </div>
-        <div class="govuk-grid-column-three-quarters">
-      [% ELSE %]
-        <div class="govuk-grid-column-full">
-         [% IF c.cobrand.moniker == 'peterborough' %]
-          <h3 id="[% unit.service_name %]" class="govuk-heading-m waste-service-name">[% unit.service_name %]</h3>
```
dracos, line 93: (easier to see with whitespace changes ignored) This is lost for a service in Peterborough that doesn't have an image, there's now no heading at all; presumably is okay to leave this in?

templates/web/bexley/waste/services_extra.html
dracos: I've pushed a fixup commit that I think matches the changes if you'd like to check

templates/web/base/waste/bin_days.html
dracos: Have added fe86ccb0f0 to put this back, if you could check that also

templates/web/bexley/waste/services_extra.html
dracos: https://github.com/mysociety/fixmystreet/pull/5793/commits/97337b9fbb951dee87bd8a6cc9ba4c9a9269d6a5 is the commit involved

templates/web/base/waste/bin_days.html
lucascumsille: It works on my local.

templates/web/bexley/waste/services_extra.html
lucascumsille: This is also working for me

lucascumsille:
@dracos I added on [last commit](https://github.com/mysociety/fixmystreet/pull/5793/commits/5722fe747f803fdc43a622c54a61bce6b6749d4e) to remove some margin and improve the alignment on mobile and desktop.

web/cobrands/sass/_waste.scss
```diff
@@ -50,6 +50,24 @@ body.waste {
     }
 }
 
+.govuk-grid-column {
```
dracos: We can drop these now, no longer using them

web/cobrands/sass/_waste.scss
```diff
@@ -72,11 +90,49 @@ body.waste {
   }
 }
 
+.waste-service-grid {
+  display: grid;
+  column-gap: 1rem;
+  row-gap: 0;
+
+  grid-template-columns: 25% 75%;
```
dracos: I think this should be `1fr 3fr` - with percentages, the column-gap is being added, so the width ends up more than 100%, you can see this on Bromley where it gets close to the sidebar, and on mobile the lines hit the edge.

web/cobrands/sass/_waste.scss
```diff
@@ -72,11 +90,49 @@ body.waste {
...
+    grid-template-areas:
+      "name name"
+      "image description";
+
+    &--service-name {
+      align-self: start;
+    }
+
+    &--service-image {
+      align-self: start;
```
dracos, line 109: Should this have `text-align: center` maybe, so the images all look in a line (looking at Bromley)

dracos (changes requested):
A few general queries.

The loss of the border bottom on the h3s - was a bit weird to me on desktop, perhaps it should only be dropped on mobile, maybe? Dunno.

web/cobrands/sass/_waste.scss
lucascumsille: Good one, I ended up using `justify-self:center`

PR #5794, opened 2025-12-23
[Gloucestershire] Use per-category anonymous reporting setting

davea:
The Gloucestershire cobrand was allowing anonymous reports for Gloucester City categories, which should not be permitted since Gloucester's cobrand does not enable anonymous reporting.

This change removes the cobrand-level anonymous reporting setting in favour of setting it on individual categories within the body.

## Notes for deployer

- [ ] Before deploying go through Gloucestershire's categories and set them as `anonymous_allowed` (or do it in the DB)

For FD-6454

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Gloucestershire.pm
+++ b/perllib/FixMyStreet/Cobrand/Gloucestershire.pm
@@ -51,12 +51,6 @@ sub council_url { 'gloucestershire' }
 
 sub admin_user_domain { 'gloucestershire.gov.uk' }
 
-=item * Allows anonymous reporting
-
-=cut
-
-sub allow_anonymous_reports { 'button' }
-
 =item * Gloucestershire use their own privacy policy
 
 =cut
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Gloucestershire.pm
```diff
@@ -51,11 +51,32 @@ sub council_url { 'gloucestershire' }
 
 sub admin_user_domain { 'gloucestershire.gov.uk' }
 
-=item * Allows anonymous reporting
+=item * Allows anonymous reporting (but not for Gloucester City categories)
 
 =cut
 
-sub allow_anonymous_reports { 'button' }
+sub allow_anonymous_reports {
```
dracos: This function is the same as Default's now, apart from it doesn't have the anonymous_allowed check. So perhaps here instead we could remove this function entirely and tick every category for Gloucestershire to have the same effect without special code?

dracos (commented):
Is okay, but as discussed after checking it perhaps it is easier to remove the allow_anonymous_reports function here and set the admin tickboxes instead, then that works without needing special behaviour.

perllib/FixMyStreet/Cobrand/Gloucestershire.pm
chrismytton: Have force-pushed a commit to remove this fn and update the tests.

PR #5800, opened 2026-01-07
[Second PR] Allow group to be updated alongside category on report edit/inspect forms

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4639.

New PR for `handle-group-with-category-selection` as the original (https://github.com/mysociety/fixmystreet/pull/5777) was merged but changes had to be reverted due to a bug.

- [x] Must be deployed after bug fix here: https://github.com/mysociety/fixmystreet/pull/5803

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Reports.pm
@@ -222,49 +222,6 @@ sub edit : Path('/admin/report_edit') : Args(1) {
     }
 
     $c->stash->{problem} = $problem;
-    if ( $problem->extra ) {
-        my @fields;
-        if ( my $fields = $problem->get_extra_fields ) {
-            for my $field ( @{$fields} ) {
-                my $name = $field->{description} ?
-                    "$field->{description} ($field->{name})" :
-                    "$field->{name}";
-                push @fields, { name => $name, val => $field->{value} };
-            }
-        }
-        my $extra = $problem->get_extra_metadata;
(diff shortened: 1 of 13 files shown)


perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -409,6 +409,7 @@ sub edit_category : Private {
     my ($group_changed, $category_changed);
     my $category_display;
     my $group_old = $problem->get_extra_metadata('group') // '';
+    $group_old = $group_old->[0] if ref $group_old eq 'ARRAY';
```
davea: I suppose it's very unlikely to happen, but if `$group_old` was `[]` then after this line it'd be `undef`. Not sure if that's worth handling?

nephila-nacrea:
> I don't think the group on the problem should ever be an array, the fact it ever was is the underlying bug. (I think it's when an Open311 update is pulled in and includes a `<group>`, I think that might always be setting an array.) If we can fix that as part of this, given we now know of it, that would be good. And then we could manually fix the DB entries that aren't a group, and then this fix wouldn't be needed (though could leave it just in case, but I think we should be able to assume the group on a report is only a string - there are probably other places that are assuming this as well, dunno if any other bugs that we haven't noticed due to it).

@dracos Yeah, I'll have a look at Open311. (Suspect I may have introduced a bug with work I did on N/land.) Would you be able to find which DB entries on live have arrayref groups? I guess we can start fixing the data at any point.

PR #5801, opened 2026-01-07
[Oxfordshire] Allow staff and reporter updates on specific categories.

neprune:
Applies even when the category has the 'no updates' flag set. For https://mysocietysupport.freshdesk.com/a/tickets/6467.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Oxfordshire.pm
@@ -604,9 +604,8 @@ sub get_ward_type {
     }
 }
 
-around updates_disallowed => sub {
-    my ($orig, $self, $problem) = @_;
-    my $result = $self->$orig($problem);
+sub updates_disallowed_override {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Cobrand/Oxfordshire.pm
```diff
@@ -604,4 +604,21 @@ sub get_ward_type {
...
+    my $result = $self->$orig($problem);
+
+    # Allow staff and reporter updates on specific categories.
+    if (($problem->category eq 'Other' ||
+        $problem->category eq 'Other lighting issue' ||
+        $problem->category eq 'Unauthorised objects on street lighting assets' ||
+        $problem->category eq 'Unauthorised objects on other highway infrastructure')
+        && $result) {
+        my $c = $self->{c};
+        return !($c->user->from_body || $problem->user_id == $c->user->id);
```
dracos: If they're logged out, $c->user is undefined, so this needs a $c->user_exists check as well or it will error.

PR #5804, opened 2026-01-12
[WW][Merton] Add note to mixed recycling container requests.

neprune:
For https://mysocietysupport.freshdesk.com/a/tickets/6479.

Screenshot 2026-01-12 at 14 03 35

[skip changelog]

### DIFF
--- a/t/app/controller/waste_merton.t
+++ b/t/app/controller/waste_merton.t
@@ -730,6 +730,23 @@ FixMyStreet::override_config {
         $mech->content_contains('is set up for assisted collection');
         $e->mock('GetServiceUnitsForObject', sub { $bin_data });
     };
+
+    FixMyStreet::Script::Reports::send();
+    $mech->clear_emails_ok;
+
+    subtest 'test special text for mixed recycling container requests' => sub {
+        $mech->get_ok('/waste/12345/request');
+        $mech->content_contains('Currently out of stock.');
+        $mech->submit_form_ok({ with_fields => { 'container-12' => 1, 'quantity-12' => 1 } });
+        $mech->submit_form_ok({ with_fields => { 'request_reason' => 'new_build', 'request_reason_text' => 'Large household' }});
+        $mech->submit_form_ok({ with_fields => { name => 'Bob Marge', email => $user->email }});
+        $mech->submit_form_ok({ with_fields => { process => 'summary' } });
+        $mech->content_contains('Currently out of stock.');
+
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1320,6 +1320,7 @@ sub add_report : Private {
         $report->set_extra_metadata($k => $data->{$_});
     }
 
+    $report->set_extra_metadata(confirmation_text => $c->stash->{confirmation_text});
```
dracos: Feels a bit odd to store this permanently on the report?
Is there a reason to do so? Aren't we showing the confirmation page immediately, so `confirmation_text` should be on the stash in the confirmation template.

t/app/controller/waste_merton.t
```diff
@@ -730,6 +730,14 @@ FixMyStreet::override_config {
         $mech->content_contains('is set up for assisted collection');
         $e->mock('GetServiceUnitsForObject', sub { $bin_data });
     };
+    subtest 'test confirmation page for mixed recycling container request includes special text' => sub {
+        $mech->get_ok('/waste/12345/request');
```
dracos, line 738: Could this test it's shown on this page as well?

perllib/FixMyStreet/App/Controller/Waste.pm
neprune: Makes sense, my bad.

templates/email/default/waste/other-reported.html
```diff
@@ -90,6 +90,8 @@ <h1 style="[% h1_style %]">Your report has been&nbsp;logged</h1>
   Your report about the problem with your bin collection has been made to the council.
   The crew will be advised. Please let us know if it happens again.
   </p>
+[% ELSIF cobrand.moniker == 'merton' AND report.category == 'Request new container' AND cobrand.service_name_override({ServiceId => report.get_extra_field_value('service_id')}) == 'Mixed recycling' %]
```
dracos: I checked the ticket again and I think this is only for the green recycling boxes (going by what they say about where to show the message?), not all mixed recycling services (not that I think any other containers should appear, that'd be a purple sack or a communal bin(!), but just in case it might confuse?).
So the checks in Merton/Waste.pm can check the service ID directly against `$SERVICE_IDS{domestic_mixed}`, and the email templates here can check that the service ID is the specific required number (I don't think we have any lookups in the templates at present).

PR #5805, opened 2026-01-13
[Image report first] Provisory styling fixes

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5342

I also included some spacing fixes, for all cobrands so the current spacing doesn't change with the image report first.

### DIFF
--- a/web/cobrands/aberdeenshire/layout.scss
+++ b/web/cobrands/aberdeenshire/layout.scss
@@ -54,7 +54,7 @@
 }
 
 .postcode-form-box {
-  margin-top: 2rem;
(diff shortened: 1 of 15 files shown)


lucascumsille:
Thanks @dracos it looks good to me. Regarding the commits how would you like to handle this. Should I merge all the cobrands into one?

dracos:
I'll squash and put on top/with the base commit (and then `photo-first-frontend-cobrands` can be the branch for any changes needed when it's turned on for the cobrands - I've left any changes you've made for that there but they might not be needed now perhaps).

PR #5806, opened 2026-01-14
[Third PR 🐵] Allow group to be updated alongside category on report edit/inspect forms

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4639.

Another PR for `handle-group-with-category-selection` as yet another bug, where response templates were not being picked up in JS.
Previous two: https://github.com/mysociety/fixmystreet/pull/5777, https://github.com/mysociety/fixmystreet/pull/5800.

### DIFF
--- a/templates/web/base/admin/report-category.html
+++ b/templates/web/base/admin/report-category.html
@@ -2,7 +2,7 @@
 [% IF cat.is_disabled AND NOT cobrand.staff_can_assign_reports_to_disabled_categories %]
     [% SET disable = 1 %]
 [% END %]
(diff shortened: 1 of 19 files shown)


dracos (changes requested):
Not a full review, sorry, but one thing I've noticed on staging with this is that some reports no longer have the category preselected in the inspector form, e.g. https://staging.fixmystreet.com/report/36732 (which I think is a report pulled in from the back end). I guess reports pulled in might only set the category, not the group, so I think we need this to still select a possible option rather than nothing, which would be confusing.

templates/web/base/admin/report-category.html
dracos, line 12: This value hasn't been updated to cope with this change, which means if a category in a group is showing here (because it's a Waste, or has since been deleted, or something), the group will be lost upon saving, I think? It needs to include the existing group here too.

dracos (changes requested):

PR #5807, opened 2026-01-14
[TfL] Add licence application forms

chrismytton:
As part of the application forms project this adds the foundations for the new licencing forms, and implements the first form, scaffolding, along with the shared fields etc for them. Also refactors some bits of the claims form around file uploads to be shared between the forms.

In order to test this locally you'll need to enable it in the config:

```
COBRAND_FEATURES:
  # Other fields…
  licencing_forms:
    tfl: 1
```

I think you might need to create the category as well, but I was able to create reports without the category existing, so not sure if I've done something wrong there?

I haven't done anything about the PDF generation (https://github.com/mysociety/societyworks/issues/5343) yet, this is just the form itself.

Fixes https://github.com/mysociety/societyworks/issues/5304

### DIFF
--- a/perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
+++ b/perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
@@ -14,6 +14,8 @@ sub tandc_link { 'https://content.tfl.gov.uk/scaffold-guidance-notes-and-terms-c
 
 sub next_after_applicant { 'dimensions' }
 
+sub num_steps { 16 }
+
 # ==========================================================================
 # Introduction / Before you get started
 # ==========================================================================
@@ -34,6 +36,7 @@ has_field start => (
 # Location (fields from Fields::Location role)
 # ==========================================================================
 has_page location => (
+    step_number => 1,
     fields => ['building_name_number', 'street_name', 'borough', 'postcode', 'continue'],
     title => 'Location of the scaffold',
     intro => 'location.html',
@@ -48,6 +51,7 @@ has_page location => (
 # Scaffold dimensions
 # ==========================================================================
 has_page dimensions => (
+    step_number => 4,
     fields => ['scaffold_height', 'scaffold_length', 'scaffold_width', 'continue'],
     title => 'Scaffold dimensions',
     next => 'activity',
@@ -105,6 +109,7 @@ has_field scaffold_width => (
 # Scaffold activity
 # ==========================================================================
 has_page activity => (
+    step_number => 5,
     fields => ['activity', 'continue'],
     title => 'Purpose of the scaffold',
     next => 'site_pedestrian_space',
@@ -125,6 +130,7 @@ has_field activity => (
 # Split into one question per page sometimes for better UX with long labels
 # ==========================================================================
 has_page site_pedestrian_space => (
+    step_number => 6,
     fields => ['site_adequate_space', 'footway_incursion', 'continue'],
     title => 'Pedestrian space',
     next => 'site_carriageway_distance',
@@ -153,6 +159,7 @@ has_field footway_incursion => (
 
 # ==========================================================================
 has_page site_carriageway_distance => (
+    step_number => 7,
     fields => ['carriageway_incursion', 'site_within_450mm', 'continue'],
     title => 'Carriageway impact',
     next => 'site_infrastructure',
@@ -180,6 +187,7 @@ has_field site_within_450mm => (
 
 # ==========================================================================
 has_page site_infrastructure => (
+    step_number => 8,
     fields => ['site_obstruct_infrastructure', 'site_trees_nearby', 'site_tfl_structures', 'continue'],
     title => 'Street infrastructure',
     next => 'site_protection',
@@ -228,6 +236,7 @@ has_field site_tfl_structures => (
 
 # ==========================================================================
 has_page site_protection => (
+    step_number => 9,
     fields => ['site_protection_fan', 'site_foundations_surveyed', 'continue'],
     title => 'Scaffold installation',
     next => 'site_hoarding',
@@ -260,6 +269,7 @@ has_field site_foundations_surveyed => (
 
 # ==========================================================================
 has_page site_hoarding => (
+    step_number => 10,
     fields => ['site_hoarding_attached', 'continue'],
     title => 'Hoarding',
     next => 'type',
@@ -283,6 +293,7 @@ has_field site_hoarding_attached => (
 # Scaffold type
 # ==========================================================================
 has_page type => (
+    step_number => 11,
     fields => ['scaffold_type', 'scaffold_configured', 'continue'],
     title => 'Type of scaffold',
     intro => 'scaffold/type.html',
@@ -334,6 +345,7 @@ has_field scaffold_configured => (
 # Fields from Fields::TemporaryProhibition role
 # ==========================================================================
 has_page have_you_considered => (
+    step_number => 12,
     fields => [
         'parking_dispensation',
         'parking_bay_suspension',
@@ -349,6 +361,7 @@ has_page have_you_considered => (
 );
 
 has_page terms => (
+    step_number => 13,
     fields => [
         'terms_accepted',
         'continue'
@@ -369,6 +382,7 @@ my $upload_fields = [
     'continue'
 ];
 has_page uploads => (
+    step_number => 14,
     fields => $upload_fields,
     title => 'Upload required documents',
     intro => 'uploads.html',
--- a/perllib/FixMyStreet/App/Form/Wizard.pm
+++ b/perllib/FixMyStreet/App/Form/Wizard.pm
@@ -49,6 +49,7 @@ has current_page => ( is => 'ro', lazy => 1,
         intro_template => 'intro',
         title => 'title',
         template => 'template',
+        step_number => 'step_number',
     }
 );
 
--- a/perllib/FixMyStreet/App/Form/Licence/Banner.pm
+++ b/perllib/FixMyStreet/App/Form/Licence/Banner.pm
@@ -14,6 +14,8 @@ sub tandc_link { 'https://content.tfl.gov.uk/banners-guidance-notes-and-terms-co
 
 sub next_after_contractor { 'type' }
 
+sub num_steps { 14 }
+
 # ==========================================================================
 # Introduction / Before you get started
 # ==========================================================================
@@ -34,6 +36,7 @@ has_field start => (
 # Location (fields from Fields::Location role)
 # ==========================================================================
 has_page location_1 => (
+    step_number => 1,
     fields => ['building_name_number', 'street_name', 'borough', 'postcode', 'add_another', 'continue'],
     title => 'Location of the banner/s',
     intro => 'location.html',
@@ -52,6 +55,7 @@ for my $page (2..5) {
         push @$fields, 'add_another';
     }
     has_page "location_$page" => (
+        step_number => 1,
         fields => $fields,
         title => "Location of the Column Attachments ($page)",
         intro => 'location.html',
@@ -80,6 +84,7 @@ has_field 'add_another' => (
 # Banner activity
 # ==========================================================================
 has_page type => (
+    step_number => 5,
     fields => ['banner_type', 'banner_content', 'continue'],
     title => 'Banner details',
     next => 'installation',
@@ -110,6 +115,7 @@ has_field banner_content => (
 
 # ==========================================================================
 has_page installation => (
+    step_number => 6,
     fields => ['banner_installation', 'continue'],
     title => 'Banner installation',
     next => 'site_pedestrian_space',
@@ -130,6 +136,7 @@ has_field banner_installation => (
 # Site Specific Information (banner-specific questions)
 # ==========================================================================
 has_page site_pedestrian_space => (
+    step_number => 7,
     fields => [ 'site_adequate_space', 'footway_incursion', 'footway_headroom',  'continue' ],
     title => 'Pedestrian space',
     next => 'site_carriageway_distance',
@@ -170,6 +177,7 @@ has_field footway_headroom => (
 
 # ==========================================================================
 has_page site_carriageway_distance => (
+    step_number => 8,
     fields => ['carriageway_incursion', 'carriageway_headroom', 'continue'],
     title => 'Carriageway space',
     next => 'site_infrastructure',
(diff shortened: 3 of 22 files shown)


web/cobrands/tfl/licence.js
```diff
@@ -0,0 +1,26 @@
+// Licence forms: disable contractor fields when "same as applicant" is checked
```
dracos, line 1: Would be good, to eventually wean off jQuery, if we could use vanilla JS for little things like this.

As this is at the end of the page, we don't need the $( loop; getElementById for the checkbox, and so on - disabled is easier, as it is then only `.disabled = disabled`.

perllib/FixMyStreet/App/Controller/Licence.pm
```diff
@@ -0,0 +1,158 @@
+package FixMyStreet::App::Controller::Licence;
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'FixMyStreet::App::Controller::Form' }
+
+use utf8;
+
+# Discover and load all form classes in FixMyStreet::App::Form::Licence::*
```
dracos: Seems like a nice way to do it :)

perllib/FixMyStreet/App/Controller/Licence.pm
```diff
@@ -0,0 +1,158 @@
...
+        && $c->user->from_body
+        && $data->{email}
+        && $c->user->email ne $data->{email};
+
+    # Find or create user
+    my $user = $c->user_exists
+        ? $c->user->obj
+        : $c->model('DB::User')->find_or_new({ email => $data->{email} });
+    $user->name($data->{name}) if $data->{name};
+    $user->phone($data->{phone}) if $data->{phone};
```
dracos, line 55: I think this may be a problem in Claims as well; if the user already exists, its name/phone are not updated (which I think is right, because otherwise someone could override it only by specifying their email, if there's no confirmation). It looks like in Waste we're now setting a `$report->set_extra_metadata(phone => $c->stash->{phone});` in these cases, maybe? Bit unclear, and hopefully won't matter, but thought worth noting to have a think.

perllib/FixMyStreet/App/Controller/Licence.pm
```diff
@@ -0,0 +1,158 @@
...
+        for my $stage (@{ $form->fields_for_display }) {
+            next if $stage->{hide};
+            for my $field (@{ $stage->{fields} }) {
+                next if $field->{hide};
+                $detail .= "$field->{desc}: " . $field->{pretty} . "\n";
+            }
+        }
+    }
+
+    my $category = "Licence: $name";
```
dracos: Whilst it should be fine, I'd rather avoid a colon in a category name if possible :) 
Also it's sometimes used "in a sentence" as in "Reported in the Foo category" (which I realise doesn't make much sense here, but "Scaffolding licence" would work better

perllib/FixMyStreet/App/Controller/Licence.pm
```diff
@@ -0,0 +1,158 @@
...
+
+    my %shared = (
```
dracos: Not sure what's shared here, this is only used once (guess it does match claims :) ).

perllib/FixMyStreet/App/Controller/Licence.pm
```diff
@@ -0,0 +1,158 @@
...
+    # Find or create user
+    my $user = $c->user_exists
+        ? $c->user->obj
+        : $c->model('DB::User')->find_or_new({ email => $data->{email} });
+    $user->name($data->{name}) if $data->{name};
+    $user->phone($data->{phone}) if $data->{phone};
+
+    # Build detail string from form fields
+    my $detail = "";
+    if ($form->can('fields_for_display')) {
```
dracos, line 59: I think this section needs a little bit of work - with Claims there's no overlapping questions, but here eg. there's no way to tell the answers for applicant and contractor apart in this big wodge of Q: A. Would be good to split each page/section with a blank line, as well.

perllib/FixMyStreet/App/Form/Claims.pm
```diff
@@ -17,11 +13,10 @@ has finished_action => ( is => 'ro' );
 
 has '+is_html5' => ( default => 1 );
 
-has upload_dir => ( is => 'ro', default => sub {
```
dracos: Rather than three similar upload_dir functions, you could have the one (in Wizard.pm) that calls $_[0]->upload_subdir instead of "uploads", then each class/subclass only has to set that to claims_files/licence_files etc.

perllib/FixMyStreet/App/Form/Wizard.pm
```diff
@@ -283,6 +315,85 @@ sub update_photo {
     }
 }
 
+=head2 File upload methods
```
dracos, line 334: Thanks for adding docs

perllib/FixMyStreet/App/Form/Licence.pm
```diff
@@ -0,0 +1,125 @@
...
+                value  => 'raw_value',
+                hide   => 0/1,  # optional
+            },
+            ...
+        ]
+    }
+
+=cut
+
+sub fields_for_display {
```
dracos: Could this and format for display be shared with Claims? They seem pretty almost identical.
With other change for upload_dir, that would mean this file isn't needed at all then.

templates/web/base/licence/have_you_considered.html
```diff
@@ -0,0 +1,13 @@
+<p class="govuk-inset-text">If the answer to VIII or IX is 'yes' then you will also require a Temporary Traffic Regulation Order (TTRO) or a Temporary Consolidated Suspension Request (TCSR). There are additional lead times required to process these.</p>
+
+<p>You may also need to consider:</p>
+<ul>
+    <li><a href="https://tfl.gov.uk/modes/driving/red-routes/dispensations" target="_blank" rel="noopener">Dispensations</a></li>
+    <li><a href="https://tfl.gov.uk/modes/buses/bus-stop-suspensions" target="_blank" rel="noopener">Bus stop suspensions</a></li>
```
dracos: Dunno if these have come from them, but a number of the links are 404

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+    required => 1,
+    tags => {
+        hint => 'For example, "no carriageway incursion" or "temporary lane closure for installation"',
+    },
+);
+
+# ==========================================================================
+# Page 7: Site Specific Information (scaffold-specific questions)
+# ==========================================================================
+has_page site_specific => (
```
dracos: Question 1 is missing

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+    ],
+    title => 'Site specific information',
+    intro => 'scaffold/site_specific.html',
+    next => 'have_you_considered',
+);
+
+has_field site_within_450mm => (
+    type => 'Select',
+    widget => 'RadioGroup',
+    label => 'II. Will any part of the scaffold be within 450mm of the edge of carriageway?',
```
dracos: Think we should drop the Roman numerals. Could number the site specific qns 1-7 to keep the help text the same. (Or ideally - not necessarily for now - the message could show with JS if the answers are as the message says?)

perllib/FixMyStreet/App/Form/Licence/Fields/TemporaryProhibition.pm
```diff
@@ -0,0 +1,50 @@
...
+parking_bay_suspension, road_closure_required, terms_accepted
+
+These fields are identical across all 16 licence types.
+
+=cut
+
+has_field parking_bay_suspension => (
+    type => 'Select',
+    widget => 'RadioGroup',
+    label => 'VIII. Will a parking, loading, disabled and/or motorcycle parking bay need to be suspended?',
```
dracos: ...and 8/9 should be 1/2 again if needed, as it’s a new page (but really, don’t think numbers are needed here, help text can be tweaked instead and no need to number this page)

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+    next => 'summary',
+);
+
+has_field payment_transaction_id => (
+    type => 'Text',
+    label => 'Transaction ID',
+    tags => {
+        hint => 'Enter the transaction ID from your payment',
+    },
+    required => 1,
```
dracos: Was transaction ID meant to be optional, or required?

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
dracos: As mentioned, given they’ve mentioned mobile use, I wonder if some of the pages with many questions should be split up to more one-per-page - the “details of” (H/L/W on one, then rest each one at a time, apart from perhaps both incursions?), and “site specific” (each one separate?) pages specifically.

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+    next => 'applicant',
+);
+
+# ==========================================================================
+# Page 4: About You (Applicant)
+# Fields: organisation, address, phone_24h from Fields::Applicant role
+# Fields: name, email, phone from AboutYou role
+# ==========================================================================
+has_page applicant => (
+    fields => [
```
dracos: The field order on the about applicant page is wrong. This came up in waste with bexley verifications, perhaps now is an opportunity to fix it so the order is always that specified in fields ; at present it’s done on order and then somehow on MRO import order - which makes including things in multiple places harder than need be (but for now, using manual order numbering could provide a fix if necessary here).

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+    tags => {
+        hint => 'Enter the transaction ID from your payment',
+    },
+    required => 1,
+);
+
+# ==========================================================================
+# Page 11: Summary
+# ==========================================================================
+has_page summary => (
```
dracos: The summary page should hide the empty applicant rows if you’ve clicked “Yes” to it being the same (think can do this with hide tag functions), to tidy that up a bit.

perllib/FixMyStreet/App/Controller/Licence.pm
dracos: Also as mentioned don't need the contractor empty rows

web/cobrands/sass/_licencing_forms.scss
```diff
@@ -0,0 +1,67 @@
+body.formflow {
```
dracos, line 1: Could we have a shared CSS, as this is the same as claims, and I don't think it's going to differ a lot

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
```diff
@@ -0,0 +1,459 @@
...
+extends 'FixMyStreet::App::Form::Wizard';
+use utf8;
+
+# Core licence functionality (upload handling, display methods)
+with 'FixMyStreet::App::Form::Licence';
+
+# Standard user fields (name, email, phone)
+with 'FixMyStreet::App::Form::AboutYou';
+
+# Shared field roles
```
dracos: This all looks good way of laying it out :)

web/cobrands/tfl/licence.js
chrismytton: af270c89346ddd3c1821b9b3b92a080be1d3bdce

perllib/FixMyStreet/App/Controller/Licence.pm
chrismytton: b5489f098ee34af853fd7b9a2fe10569c04ca272

perllib/FixMyStreet/App/Controller/Licence.pm
chrismytton: ff4b36b93e4c641b8ed1b86949f085e47eed1549

templates/web/base/licence/have_you_considered.html
chrismytton: b5489f098ee34af853fd7b9a2fe10569c04ca272

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
chrismytton: Question 1 added in 23368e1377365c2787c32ef574084f40e882b050

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
chrismytton: 23368e1377365c2787c32ef574084f40e882b050 drops the roman numerals. Agree that using JS would be the ideal solution.

perllib/FixMyStreet/App/Form/Licence/Fields/TemporaryProhibition.pm
chrismytton: 23368e1377365c2787c32ef574084f40e882b050

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
chrismytton: Optional! 23368e1377365c2787c32ef574084f40e882b050

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
chrismytton: 23368e1377365c2787c32ef574084f40e882b050

perllib/FixMyStreet/App/Form/Licence/Scaffold.pm
chrismytton: Split things up into separate pages in 2fe140b33d6220c84036a70d19d765c65a68cd56

perllib/FixMyStreet/App/Controller/Licence.pm
chrismytton: 4544af132cbb6599a9c22040e263d380ee97651b adds section headers and blank lines.

chrismytton:
Have put this on staging and setup a "Scaffold licence" category.

For now I've used "Red claim " as the category type, but we should make this more generic, have opened https://github.com/mysociety/societyworks/issues/5348 to track that.

chrismytton:
@dracos Your commits all look good!

I've pushed two fixup commits, one to fix the curly quotes in the Mobile Apparatus form, and one to add token-auth to the PDF link. Can you review those bits?

chrismytton:
Pushed another fixup to add a download attribute to the PDF download link.

PR #5810, opened 2026-01-15
Make sure price shown in pounds

dracos:
And a missing bit of info not being shown in the admin.
[skip changelog]
FD-6465


MorayMySoc (approved):
Looks good to me

PR #5813, opened 2026-01-16
FD6151: Kingston WW: allow non-actionable missed collection reports after initial window

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/5262.

Also requires config changes in `mysociety-servers/vhosts/open311-adapter/conf_council-kingston_echo.yml.ugly` (commit 720d363cf - has been pushed).

[skip changelog]


templates/web/base/waste/summary_enquiry.html
```diff
@@ -1,11 +1,22 @@
 [%
-title = 'Submit enquiry';
-thing = 'enquiry';
-summary_title = data.category;
-step1 = 'enquiry';
+title = non_actionable ? 'Submit missed bin report' : 'Submit enquiry';
```
dracos: I'd probably do this as a `IF non_actionable` / `ELSE` block, given everything is different, and in case something else comes along in future.

templates/web/base/waste/_service_missed.html
```diff
@@ -9,6 +9,9 @@
     <input type="hidden" name="service-[% unit.service_id %]" value="1">
     <input type="submit" value="Report a [% unit.service_name FILTER lower %] collection as missed" class="waste-service-descriptor waste-service-link">
   </form>
+[% ELSIF unit.report_allowed_non_actionable %]
+  [% # TODO Can we make this a post so params are not visible in URL? Does mean making sure params are passed between form pages. %]
```
dracos: Currently it reads category/service_id into the form in the About You page. I don't think there's an issue with the parameters being in the URL?

dracos (changes requested):
One tiny thing is all from code point of view.

I think it'd be slightly simpler to have its own category, then you wouldn't need to have the `non_actionable` parameter anywhere, you could check the category instead (don't think it would be a big change). If they do other general enquiry things in future would be easier to keep them separate for reporting purposes. Let me know if there's a reason you didn't do that, we can chat about it :)

templates/web/base/waste/summary_enquiry.html
```diff
@@ -1,11 +1,34 @@
...
+    summary_title = data.category;
+  %]
+[% END %]
 [%
-title = 'Submit enquiry';
-thing = 'enquiry';
-summary_title = data.category;
-step1 = 'enquiry';
+  step1 = first_page;
 %]
```
dracos: If you have a big block of mostly `[%` type stuff, I find it a bit easier to have just the one block, just means you need a few semicolons, but have less whitespace in the output, so something like:
```
[%
data = form.saved_data;
non_actionable = ( data.category == 'Report out-of-time missed collection' );
IF non_actionable;
    title = 'Submit missed bin report';
    thing = 'missed collection';
    summary_title = 'Missed collection';
ELSE;
    title = 'Submit enquiry';
    thing = 'enquiry';
    summary_title = data.category;
END;

step1 = first_page;
%]

File: perllib/FixMyStreet/Cobrand/Kingston.pm [hunk 4]
```diff
@@ -638,6 +638,13 @@ sub waste_munge_enquiry_data {
         $data->{extra_Notes} = "Originally Echo Event #$echo";
         $data->{extra_original_ref} = $ww;
         $data->{extra_container_request_guid} = $guid;
+    } elsif ($data->{category} eq 'Report out-of-time missed collection') {
+        $data->{title}
+            = 'Report missed '
+            . $self->service_name_override( { ServiceId => $data->{service_id} } )
+            . ' (non-actionable)';
```
dracos: As we've got its own category now, I think this bracket bit can be left off (which means you don't need to then worry about stripping it in the email).

dracos (changes requested):
Thanks for doing that change; only couple of tiny things now :)

dracos (approved):
Woo, thanks :) 🦾

perllib/FixMyStreet/Cobrand/Kingston.pm
```diff
@@ -629,7 +629,14 @@ sub waste_munge_enquiry_form_pages {
 
     return unless $category eq 'Report out-of-time missed collection';
 
-    $pages->[1]{intro} = 'enquiry-non-actionable-intro.html';
+    $c->stash->{first_page} = 'intro';
+    unshift @$pages, intro => {
```
nephila-nacrea: This seems to be adding a second 'intro' page to the list of pages, which I don't think we want. Were the changes not working in the previous commit?

nephila-nacrea: I.e. I get the following when I dump `$pages`:

```
[                                                                                                                                 
  'intro',                                                                                                                                
  {                                                                                                                                       
    'fields' => [                                                                                                                         
      'continue'                                                                                                                          
    ],                                                                                                                                    
    'intro' => 'enquiry-non-actionable-intro.html',                                                                                       
    'next' => 'about_you',                                                                                                                
    'title' => 'Reporting a missed collection outside the 48 hour reporting window'
  },                                                                                                                                      
  'enquiry',                                                                                                                              
  {                                                                                                                                       
    'fields' => [                                                                                                                         
      'continue'
    ],                                                                                                                                        'intro' => 'enquiry-intro.html',                                                                                                      
    'next' => 'about_you',                                           
    'title' => 'Reporting a missed collection outside the 48 hour reporting window'
  }                                                                  
]
```

PR #5816, opened 2026-01-20
[Kingston][Sutton][WW] Amend small items warning text

chrismytton:
There's no charge for small items collection, so update the text to reflect that.

For FD-6510

### DIFF
--- a/templates/web/base/waste/bulky/summary.html
+++ b/templates/web/base/waste/bulky/summary.html
@@ -45,7 +45,7 @@ SET data = form.saved_data;
           <div class="govuk-warning-text error">
             <div class="govuk-warning-text__content">
               Please carefully read all the details for your booking including the items as you
-              won't be able to amend the booking once you have paid for the service.
+              won't be able to amend the booking once submitted.
             </div>
           </div>
         [% ELSE %]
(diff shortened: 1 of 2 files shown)


templates/web/base/waste/bulky/summary.html
```diff
@@ -41,7 +41,14 @@ <h2 class="govuk-heading-l govuk-!-margin-bottom-5">Booking Summary</h2>
   [% UNLESS problem %]
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        [% IF cobrand.moniker == 'sutton' || cobrand.moniker == 'kingston' %]
+        [% IF cobrand.moniker == 'sutton' AND small_items %]
```
MorayMySoc: I think this applies to Kingston too as further down (125) in a k/s block we exclude mentioning charging for both if it's small items...

chrismytton: Yeah, though the Freshdesk ticket is "Sutton small items collection - text change", so I wonder if its worth just doing Sutton for now and then asking if they intended it to cover Kingston as well?

dracos: Only Sutton have small items at present, though we can probably assume Kingston will want to at some point in future, so I'd probably have this apply to both now anyway, yeah.

MorayMySoc (approved):

PR #5827, opened 2026-01-28
[Sutton] Add message to request list

MorayMySoc:
Where a container has already been requested and option is disabled, add message saying so.

https://mysocietysupport.freshdesk.com/a/tickets/6441

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Sutton.pm
+++ b/perllib/FixMyStreet/Cobrand/Sutton.pm
@@ -507,7 +507,7 @@ sub waste_munge_request_form_fields {
             value => $id,
             label => $self->{c}->stash->{containers}->{$id},
             disabled => $value->{disabled},
-            $hint ? (hint => $hint) : (),
+            hint => $value->{option_hint} || $hint, # In progress overrides
         };
         if ($cost && $change_cost && $cost == $change_cost) {
             push @replace_options, $data;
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -502,7 +502,14 @@ sub waste_munge_request_form_fields {
         my $id = $1;
 
         my ($cost, $hint) = $self->request_cost($id, $c->stash->{quantities});
-
+	if ($value->{disabled}) {
+	    my $text = 'Unavailable to request as order in progress';
+	    if ($hint) {
+		$hint .= "\n$text";
+	    } else {
+		$hint = $text;
```
chrismytton: Indentation needs tweaking here, looks like this is using tabs instead of spaces.

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -502,7 +502,14 @@ sub waste_munge_request_form_fields {
...
+	    }
+	};
```
chrismytton: Not wrong, but I think a semicolon on an if statement is a bit unconventional 🙂

chrismytton (approved):
Yeah you're right, does make sense to do it for all of them 👍

PR #5833, opened 2026-01-30
[Admin] Allow superusers to add new database config entries

davea:
Add functionality to create new config table entries from the admin interface at /admin/config

admin config

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Config.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Config.pm
@@ -41,6 +41,33 @@ sub index : Path( '' ) : Args(0) {
         my $db = FixMyStreet::DB->schema->storage;
         my $txn_guard = $db->txn_scope_guard;
 
+        # Handle adding new config entry
+        if (my $new_key = $c->get_param('new-config-key')) {
+            $new_key =~ s/^\s+|\s+$//g;  # Trim whitespace
+            my $new_value = $c->get_param('new-config-value');
+
+            if ($new_key && $new_value) {
+                if (FixMyStreet::DB->resultset("Config")->get($new_key)) {
+                    $c->stash->{errors}->{new} =
+                        sprintf(_("Configuration key '%s' already exists"), $new_key);
+                    $c->stash->{new_config_key} = $new_key;
+                    $c->stash->{new_config_value} = $new_value;
+                } else {
+                    try {
+                        FixMyStreet::DB->resultset("Config")->set($new_key, $json->decode($new_value));
+                    } catch {
+                        my $e = $_;
+                        $e =~ s/ at \/.*$//; # trim the filename/lineno
+                        $c->stash->{errors}->{new} =
+                            sprintf(_("Not a valid JSON string: %s"), $e);
+                        $c->stash->{new_config_key} = $new_key;
+                        $c->stash->{new_config_value} = $new_value;
+                    };
+                }
+            }
+        }
+
+        # Handle updating existing config entries
         foreach my $entry (@db_config) {
             if (my $cfg = $c->get_param("db-config-" . $entry->key)) {
                 try {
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/App/Controller/Admin/Config.pm
```diff
@@ -41,13 +41,40 @@ sub index : Path( '' ) : Args(0) {
...
+
+        # Handle updating existing config entries
         foreach my $entry (@db_config) {
             if (my $cfg = $c->get_param("db-config-" . $entry->key)) {
                 try {
                     $entry->update({ value => $json->decode($cfg) });
                 } catch {
                     my $e = $_;
-                    $e =~ s/ at \/.*$//; # trim the filename/lineno
+                    $e =~ s/ at \/.*$//;
```
dracos: Why remove this comment?

perllib/FixMyStreet/App/Controller/Admin/Config.pm
```diff
@@ -41,13 +41,40 @@ sub index : Path( '' ) : Args(0) {
...
+                    $c->stash->{new_config_key} = $new_key;
+                    $c->stash->{new_config_value} = $new_value;
+                } else {
+                    try {
+                        FixMyStreet::DB->resultset("Config")->set($new_key, $json->decode($new_value));
+                    } catch {
+                        my $e = $_;
+                        $e =~ s/ at \/.*$//;
+                        $c->stash->{errors}->{new} =
+                            sprintf(_("Error adding configuration: %s"), $e);
```
dracos: Could we use the same error message as line 79 to reduce translations? Not like it's going to be shown much/ at all.

templates/web/base/admin/config/index.html
```diff
@@ -177,14 +175,27 @@ <h2 id="db">Database site configuration</h2>
...
-        <input type="hidden" name="token" value="[% csrf_token %]">
-        <input type="submit" class="btn" value="Update">
-    </form>
-  [% END %]
 [% END %]
+    <tr>
+        <td>
+            <input type="text" name="new-config-key"
+                   placeholder="new_config_key" size="30"
+                   value="[% new_config_key | html %]">
```
dracos: No need for `| html`, it's the default.

templates/web/base/admin/config/index.html
```diff
@@ -177,14 +175,27 @@ <h2 id="db">Database site configuration</h2>
...
+            <textarea name="new-config-value" rows="3" cols="60"
+                   placeholder='e.g. true or ["value1","value2"]' class="code">[% new_config_value | html %]</textarea>
+            [% IF errors.new %]
+            <div class="form-error">[% errors.new %]</div>
+            [% END %]
+        </td>
+    </tr>
+</table>
+<input type="hidden" name="token" value="[% csrf_token %]">
+<input type="submit" class="btn" value="Update">
```
dracos: Needs a `loc()`

t/app/controller/admin/config.t
```diff
@@ -0,0 +1,155 @@
+use FixMyStreet;
+BEGIN { FixMyStreet->test_mode(1); }
```
dracos: Including FixMyStreet::TestMech as you do automatically includes lines 1, 2, and 5.

t/app/controller/admin/config.t
```diff
@@ -0,0 +1,155 @@
...
+
+use FixMyStreet::TestMech;
+use Test::More;
+use Capture::Tiny 'capture_stderr';
```
dracos, line 2: Thanks for adding some tests :)
I guess we don't want this warning being output at all, rather than suppressing it in the test? Won't happen often/at all though. Perhaps the whole POST could be wrapped in a try, then if $c->stash->{errors} wasn't empty it would die/raise an exception, so that the guard worked in the way it's expecting. But that seems a bit pointless.

templates/web/base/admin/config/index.html
davea: One day I'll remember :)

perllib/FixMyStreet/App/Controller/Admin/Config.pm
davea: No good reason!

t/app/controller/admin/config.t
davea: 🤷 Seemed simpler to just reduce noise output from test runs.

dracos (approved):
Thanks, looks good!

PR #5834, opened 2026-02-02
[TFL] Builder's skip form.

neprune:
[skip changelog] 🦹‍♂️.


dracos (commented):
Thanks :) Used a curly apostrophe; added a missing "Explain why materials cannot be loaded directly into a vehicle" (and removed text from intro)

PR #5838, opened 2026-02-06
[Bristol] Re-enable category search

chrismytton:
For FD-6588


MorayMySoc (approved):
Great - thanks!

dracos:
Not wanted unless it can be 'start' only (tree is the example), and/or could reorder other than alphabetical

PR #5841, opened 2026-02-10
[Open311] Strip EXIF metadata from photos imported via media_url

davea:
Photos fetched from Open311 `media_url` elements were stored with their original EXIF metadata intact, unlike user-uploaded photos which are stripped. This could expose GPS coordinates, camera info, and other metadata publicly.

We now process imported photos through the same ImageMagick pipeline used for user uploads, with a fallback to the original blob if processing fails.


dracos (approved):
Looks good - though now I'm very confused because the default (non coverage) tests *do* install the imagemagick package, so I don't know why they failed and the coverage ones succeeded in your first review, surely it should have been all or nothing :-/

PR #5844, opened 2026-02-11
[Bromley][Waste] Handle common DD reconcile cases in script.

neprune:
https://github.com/mysociety/societyworks/issues/4806

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/DDProcessor.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/DDProcessor.pm
@@ -126,46 +126,8 @@ sub waste_reconcile_direct_debits {
         # create a renewal record using the original subscription as a basis.
         if ( $type eq $self->waste_subscription_types->{Renew} ) {
             $self->log("is a renewal");
-            my $p;
-            # loop over all matching records and pick the most recent new sub or renewal
-            # record. This is where we get the details of the renewal from. There should
-            # always be one of these for an automatic DD renewal. If there isn't then
-            # something has gone wrong and we need to error.
-            while ( my $cur = $rs->next ) {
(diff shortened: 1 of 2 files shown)


t/cobrand/bromley_waste.t
```diff
@@ -1518,6 +1518,7 @@ subtest 'check direct debit reconcilliation' => sub {
         "looking at potential match " . $renewal_nothing_in_echo->id . " with state confirmed\n",
         "is a matching new report\n",
         "no matching service to renew for $renewal_nothing_in_echo_ref\n",
+        "done looking at payment $renewal_nothing_in_echo_ref\n",
```
neprune: Because the new method doesn't uses 'next RECORD' skip but instead just exits with the payment unhandled.

perllib/FixMyStreet/Roles/Cobrand/DDProcessor.pm
```diff
@@ -191,10 +159,16 @@ sub waste_reconcile_direct_debits {
...
             }
-        }
 
-        unless ( $handled ) {
-            $self->log("no matching record found for $category payment with id " . $payment->payer);
+            unless ( $handled ) {
+                $self->log("no matching record found for $category payment with id " . $payment->payer);
+            }
+
+            unless ( $handled ) {
```
dracos: Could be only one unless block

dracos (approved):
One tiny thing is all, and doesn't actually matter.

I guess next step would be doing the same the other way round (for --force-new) and have it check it's new. Then I wonder if we can *actually* combine the renewal/new handling so that it's not needed at all and we can only have one loop through that finds a 'relevant' report (regardless if DD is new or renew) and does something, but could that cause a problem somehow. You'd hope not, because there shouldn't be the 'other' thing in either case...

PR #5845, opened 2026-02-12
Parishes frontend

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/4834

General:
- [x] Increased content container padding
- [x] Split pages into sections
- [x] Improved heading hierarchy
- [x] Moved the fee to its own paragraph. Mainly because I find it easier to scan; it's also the first thing you see before you start completing the form, so it's not mixed in brackets with the rest of the text.
- [x] Added a "Change your previous answers" below H1 titles.
- [x] Added an override to Dropzone for parishes, so the preview thumbnail is visible at its original ratio, allowing the user to see exactly what they are uploading.
- [x] Added Progress form bar component to each page.

Intro Page:
- [x] Added USP sections
- [x] Reduced `max-width` for input fields and Dropzone area

Pick Category Page:
- [x] Changed page title so it includes the parish name.
- [x] Moved category selection higher on the page
- [x] Added more headings and split the page into sections to improve hierarchy and make the page easier to scan.
- [x] Differentiated the help section so it's clearer what its purpose is.
- [x] Added optional styling in case we decide to move ahead with GOV.UK styling for the "Remove category" button.
- [x] Moved the high-tier council categories to be next to the category selection to increase the chances that users will read them through before start choosing their category.
- [x] Added a script that checks which categories are being added by the user. Then it compares them with the high-tier council list on the left and strikethrough any `li` element that matches the user's input.

Summary:
- [x] Increased padding for content container
- [x] Split the summary into GOV.UK Summary Cards
- [x] Removed bottom margin from "Continue to Payment" button

Subscription Confirmed:
- [x] Added extra `h3` titles so the page is easier to scan
- [x] WIP: Replace video
- [x] Added promotional material

**Preview:**

Step 1:

https://github.com/user-attachments/assets/a7d007ae-2c53-4dd4-8578-b9eb28ae57a6

Step 2:

https://github.com/user-attachments/assets/fab3ca58-cfa1-4837-83f0-2b50853d3777

Screenshot 2026-02-24 at 09 36 34

Step 3:

https://github.com/user-attachments/assets/96dd3264-d70b-40bd-86e2-7e8976555392

Step 4:

https://github.com/user-attachments/assets/0f07ed35-a27c-4867-93b8-1a85cb8d5093

## Promotional material

Screenshot 2026-02-24 at 09 47 55

## Pending:

- [x] Replace video
- [x] Decide whether the buttons "add another category" and "Remove category" will be added the classes `govuk-button', 'govuk-button--secondary` and `govuk-button', 'govuk-button--secondary` in the backend. Or we will include the CSS GOVUK styling to the classes `.add_element` and `.rm_element`. Alternatively we can also doing via JS.

[skip changelog]

### DIFF
--- a/templates/web/fixmystreet.com/parishes/pay_complete.html
+++ b/templates/web/fixmystreet.com/parishes/pay_complete.html
@@ -1,4 +1,8 @@
+[% SET bodyclass = 'parish-page' %]
 [% INCLUDE 'header.html' title = 'Payment complete' %]
+
+<!-- <link rel="stylesheet" href="[% version('/vendor/lite-youtube-embed/lite-yt-embed.css') %]">
+<script src="[% version('/vendor/lite-youtube-embed/lite-yt-embed.min.js') %]" type="module"></script> -->
 
 <div class="govuk-panel govuk-panel--confirmation">
   <h1 class="govuk-panel__title">
@@ -11,34 +15,37 @@
   </div>
 </div>
 
-<p>
-  A confirmation email has been sent to the address you provided. Please allow a few minutes for this to appear in your inbox, and be sure to check your junk mail box in case it ends up in there.
+<div class="text-content" style="margin-top: 2em">
+  <p>A confirmation email has been sent to the address you provided. Please allow a few minutes for this to appear in your inbox, and be sure to check your junk mail box in case it ends up in there.</p>
 
-<h2>What now?</h2>
+  <h2>What now?</h2>
 
-<p>
-You will now receive any reports made via FixMyStreet within the boundary and categories selected.
+  <p>You will now receive any reports made via FixMyStreet within the boundary and categories selected.</p>
 
-<p>
-First, we recommend watching the below video to help you get to grips with using FixMyStreet for Parish &amp; Town Councils.
+  <!-- <p>First, we recommend watching the below video to help you get to grips with using FixMyStreet for Parish &amp; Town Councils.</p> -->
 
-<p>
-[Video]
+  <!-- <lite-youtube videoid="S8XhhDKvcmQ">
+    <a href="https://youtube.com/watch?v=S8XhhDKvcmQ" class="lty-playbtn" title="Play Video">
+      <span class="lyt-visually-hidden">Play Video: Parishes</span>
+    </a>
+  </lite-youtube> -->
 
-<p>
-Please refer to our FAQs for guidance and information about how to use and get the best out of your subscription.
+  <h3>Manage your account</h3>
 
-<p>
-<a href="/about/faq-parishes">Read our FAQs</a>
+  <p>Please <a href="/about/faq-parishes">refer to our FAQs</a> for guidance and information about how to manage your council’s settings.</p>
 
-<p>
-Finally, spread the word! Below are some assets to help you promote your FixMyStreet service within your local area.
+  <h3>Let everyone know</h3>
 
-<ul>
-    <li> A poster you can print out and physically distribute in the local area (noticeboards, libraries etc)
-    <li> An image to share on your social media channels
-    <li> An email signature banner to signpost people to the service
-    <li> A website badge perhaps?
-</ul>
+  <p>Finally, spread the word! Below are some assets to help you promote your FixMyStreet service within your local area.</p>
+
+  <ul>
+    <li>Four posters/flyers you can print out and physically distribute in the local area (noticeboards, libraries etc)</li>
+    <li>An image to share on your social media channels</li>
+    <li>An email signature banner to signpost people to the service</li>
+    <li>A badge you can add to your parish website</li>
+  </ul>
+
+  <a href="/cobrands/fixmystreet.com/posters/parishes-pack.zip" data-goodielink="parishes-pack" class="govuk-button govuk-button--secondary">Download promotional materials</a>
+</div>
 
 [% INCLUDE footer.html %]
--- a/templates/web/fixmystreet.com/parishes/summary.html
+++ b/templates/web/fixmystreet.com/parishes/summary.html
@@ -1,4 +1,4 @@
-[% SET bodyclass = 'formflow' %]
+[% SET bodyclass = 'formflow parish-page' %]
 
 [% USE date(format = '%A, %-d~~~ %B') ~%]
 [% PROCESS 'govuk/fields.html' ~%]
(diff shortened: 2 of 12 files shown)


zarino (approved):
@lucascumsille I’ve added a fixup commit with a few improvements to the categories page:

- Moved map above the intro text, to more naturally fall into the top-left to bottom-left reading order.
- Prevent drag/scroll events in the map, so it’s easier to drag/scroll past. The map is really just a visual aid, to show the rough outline/position of the parish. We’re not expecting people to closely interact with it, so the existing Zoom buttons are probably enough.
- Dropped the text explaining why the map is there and how to go back if you need to change it – on balance I think it’s better to simplify text where possible, as there’s already so much on this page, and how common is it _really_ going to be that someone heavily involved in the parish will pick the _wrong_ parish name at the previous step?
- Matching headings above the two form columns, making it clearer that one side is your parish’s categories, and the other side is the higher tier authority’s categories.
- Swap the order of the two columns, so the (primary) interactive elements are all on the left, in the natural reading order, and the (secondary) higher tier categories list is on the right, out of the way.
- More polished display of category name inputs and their associated "Remove" buttons, wrapped in individual coloured containers, to make the heirarchy of the buttons a bit clearer, and also add some visual weight, to counter the higher tier categories list beside it.
- Use a `button` rather than a `div` for the "Add another category" button.
- More polished presentation of the higher tier categories, as a zebra-striped list.

Screenshot 2026-02-25 at 14 45 37

@dracos You’ll spot my commit adds `govuk-button` classes to the `add_category` and `categories.rm_category` buttons, but, as mentioned in Slack yesterday, the form handler is automatically appending a `btn` class to these, causing an awkward mixture of styles. Given the govuk-style forms are a major focus of these pages, I _think_ it makes sense to standardise on that style of button, rather than the fixmystreet style – do you agree? If so, when you’re reviewing this PR, could you also subclass these fields, to prevent those extra classes from being added? (Perhaps, while you’re in there, you might have your new subclass apply the correct `govuk-button` styles _by default_ too…)

web/cobrands/sass/_base.scss
```diff
@@ -3492,6 +3493,146 @@ $site-message-border: 1px solid #525252 !default;
   display: none;
 }
 
+.parish-page {
```
dracos: Do we want to have a separate parishes CSS file? Given it's only applying to this section.
I couldn't decide if that was better or not.

templates/web/fixmystreet.com/about/posters.html
```diff
@@ -26,6 +34,8 @@
             <li><a href="#web">Website Badge</a></li>
             <li><a href="#print">Printed Materials</a></li>
             <li><a href="#premadeadverts">Pre-made Adverts</a></li>
+            <li><a href="#fms-parishes">Promotional material for parishes</a></li>
+            <li><a href="#fms-parishes">Website badges for parishes</a></li>
```
dracos: ID should be fms-parishes-badges here

templates/web/fixmystreet.com/parishes/pay_complete.html
```diff
@@ -23,14 +24,26 @@ <h2>What now?</h2>
 First, we recommend watching the below video to help you get to grips with using FixMyStreet for Parish &amp; Town Councils.
 
 <p>
-[Video]
 
+<div class="video-wrapper">
+  <iframe
```
dracos: Can we use lite-youtube-embed, already used by Oxfordshire?

(Noting here as well video needs replacing)

templates/web/fixmystreet.com/parishes/summary.html
```diff
@@ -7,55 +7,79 @@
...
-  <div class="govuk-summary-list__row">
-    <dt class="govuk-summary-list__key">
-      [% summary_title %]
-    </dt>
-    <dd class="govuk-summary-list__value">
-    </dd>
-    <dd class="govuk-summary-list__actions">
-    </dd>
-  </div>
+<form method="post" style="display:inline">
```
dracos: Not sure the display:inline does anything here?

templates/web/fixmystreet.com/parishes/summary.html
```diff
@@ -7,55 +7,79 @@
...
+            <div class="progress-form__circle" aria-hidden="true"></div>
+            <div class="progress-form__label">Payment</div>
+        </li>
+    </ol>
+</nav>
+
+<div class="claims__summary">
+[% FOR page IN form.fields_for_display %]
+    [% NEXT IF page.hide %]
+    <div class="govuk-summary-card">
```
dracos, line 45: Maybe something like this would be good to do on the TfL licence summary page as well...

web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1090,13 +1090,14 @@ $.extend(fixmystreet.set_up, {
...
         var prevFile;
+        var isParishPage = document.body.classList.contains('parish-page');
         var photodrop = new Dropzone($dropzone[0], {
             url: '/photo/upload',
             paramName: 'photo',
             maxFiles: max_photos,
             addRemoveLinks: true,
-            thumbnailHeight: 150,
-            thumbnailWidth: 150,
+            thumbnailHeight: isParishPage ? null : 150,
```
dracos, line 1108: I think setting *both* of these to null means no thumbnail is made at all, and it displays the full size picture tiny. I think if you leave one as 150, it should resize but keep ratio as desired

(I pushed fixup for this)

web/cobrands/sass/_base.scss
zarino: As in a separate `<link rel="stylesheet" href="parishes-base.css">` and `<link rel="stylesheet" href="parishes-layout.css">` on parish pages? Could do. Might be a bit messy though as those separate files would also need to import the `$primary` variable, which is currently handled via the `base.scss` and `layout.scss` parent files in the cobrands. Hmmm.

Does the parishes UI only ever exist on the fixmystreet.com cobrand? Or are visitors on other cobrands able to access it? If the former, then I guess the parish styling could all at least be shuffled off into the `fixmystreet.com` cobrand, to prevent it bloating all the other cobrands’ CSS files…?

Or did you just mean that we should separate all the parish styling into `_parishes-{base,layout}.scss` partials that can then be imported into `_{base,layout}.scss` to just tidy those files up a bit?

templates/web/fixmystreet.com/parishes/pay_complete.html
zarino: Good catch! I’d not seen the `lite-youtube-embed` component. :+1:

And yes, we need to replace the video.

web/cobrands/sass/_base.scss
dracos: Yeah, think I meant a separate parishes.scss (no need for base/layout, could just have media query in the one file), included only by the parishes pages. You can import _colours from anywhere, that should be fine, it's only variables.

It's only on .com cobrand, yeah, so yeah, in .com cobrand CSS would be another way, but still increases those a bit just for that. But yeah, that'd probably be fine also.

templates/web/fixmystreet.com/parishes/pay_complete.html
zarino: After chatting this through with @sallybracegirdle we’ve decided to hide the video for now.

web/cobrands/sass/_base.scss
zarino: Ok @dracos I’ve split the Parishes styles out into a separate `parishes.css` that’s only loaded (by `fixmystreet.com/header_extra.html`) when the body class contains `parish-page` 👍

PR #5848, opened 2026-02-18
[Bexley][WW] Add sharps eligibility check and sidebar link

chrismytton:
As per [this Basecamp message](https://3.basecamp.com/4020879/buckets/45093048/todos/9572660043#__recording_9593110483) properties with BLPU code prefixes of RD, RH, RI, RE & CE are eligible for sharps collections. Also adds sidebar link for sharps booking form.

Part of https://github.com/mysociety/societyworks/issues/5371
Fixes https://github.com/mysociety/societyworks/issues/5372

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Bulky.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Bulky.pm
@@ -255,10 +255,8 @@ sub bulky_refund_collection {
 
 sub sharps_allowed_property {
     my ( $self, $property ) = @_;
-
-    # XXX Implement checks
-
-    return 1;
+    my $class = $property->{class} || '';
+    return $self->sharps_enabled && $class =~ /^(RD|RH|RI|RE|CE)/;
 }
 
 sub waste_munge_sharps_data {
--- a/templates/web/bexley/waste/_more_services_sidebar.html
+++ b/templates/web/bexley/waste/_more_services_sidebar.html
@@ -28,6 +28,9 @@
   [% IF property.show_bulky_waste %]
     <li><a href="[% c.uri_for_action('waste/bulky/index', [ property.id ]) %]">Arrange a bulky waste collection</a></li>
   [% END %]
+  [% IF property.show_sharps %]
+    <li><a href="[% c.uri_for_action('waste/bulky/index_sharps', [ property.id ]) %]">Arrange a sharps collection</a></li>
+  [% END %]
   <li><a [% external_new_tab | safe %] href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?category=Request+assisted+collection">Get help with putting your bins out</a></li>
 </ul>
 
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Bulky.pm
```diff
@@ -255,10 +255,8 @@ sub bulky_refund_collection {
 
 sub sharps_allowed_property {
     my ( $self, $property ) = @_;
-
-    # XXX Implement checks
-
-    return 1;
+    my $class = $property->{class} || '';
+    return $self->sharps_enabled && $class =~ /^(RD|RH|RI|RE|CE)/ ? 1 : 0;
```
nephila-nacrea: I don't think you need the `1 : 0` bit, as the `&&` check is going to return a true or false value anyway.

templates/web/bexley/waste/_more_services_sidebar.html
```diff
@@ -28,6 +28,9 @@ <h3>More services</h3>
   [% IF property.show_bulky_waste %]
     <li><a href="[% c.uri_for_action('waste/bulky/index', [ property.id ]) %]">Arrange a bulky waste collection</a></li>
   [% END %]
+  [% IF property.show_sharps %]
+    <li><a href="[% c.uri_for_action('waste/bulky/index_sharps', [ property.id ]) %]">Arrange a sharps collection</a></li>
+  [% END %]
   <li><a [% external_new_tab | safe %] href="[% c.uri_for_action('waste/enquiry', [ property.id ]) %]?category=Request+assisted+collection">Get help with putting your bins out</a></li>
```
nephila-nacrea, line 34: Matthew mentions in https://github.com/mysociety/societyworks/issues/5372: 'not a separate section of the page like Bulky', so I wonder if we should remove the `property.show_sharps` check I added in `bin_days.html`.
But then again, we may want a sharps section for when user wants to view their report.

chrismytton: The tech spec says:

> Upcoming sharps bookings will be shown to the user (if signed in) on their property’s “Your bin days” page on WasteWorks, including a reminder of what is being collected.

So yeah, I think we'll only show that section when there's an upcoming booking.

perllib/FixMyStreet/Cobrand/Bexley/Bulky.pm
chrismytton: Ah, turns out the reason `bulky_allowed_property` has this is to prevent a warning such as the one below when running the tests.

```
Odd number of elements in anonymous hash at perllib/FixMyStreet/Cobrand/Bexley/Waste.pm line 283.
```

This occurs because `sharps_allowed_property` is called in list context in the hash constructor, and when there are no regex matches it returns `()` which causes uneven number of hash elements and therefore the above warning. I'll push a fixup to the main branch to add it back in.

nephila-nacrea: Ah good point. Sneaky behaviour!

PR #5850, opened 2026-02-18
[Peterborough] Update flytipping submit handling.

dracos:
Send all reports to the backend, whether witnessed or not. Change the subject line for the email also sent, plus send an additional email if it was witnessed.
[skip changelog]
For https://github.com/mysociety/societyworks/issues/5312


neprune (approved):
Looks good to me. As an aside, I was going to ask about whether we should have a base template for the submit email but I guess it would be super messy in trying to account for each variant so copying is probably better.

dracos:
Yeah, I wondered that too, perhaps something for the future somehow.

dracos:
I've added a second commit to send through updates (and added tests now!), if you could give that a look, ta.

neprune (approved):
LGTM

PR #5855, opened 2026-02-23
[Bexley][WW] Support dynamic service IDs for collection slots

chrismytton:
The `GetCollectionSlots` method now accepts a `service_id` parameter instead of hardcoding it to 78. This allows different waste types (such as sharps collections) to use their appropriate service IDs.

Part of https://github.com/mysociety/societyworks/issues/5373

### DIFF
--- a/t/app/controller/waste_bexley_sharps.t
+++ b/t/app/controller/waste_bexley_sharps.t
@@ -59,6 +59,12 @@ FixMyStreet::override_config {
         $mech->submit_form_ok;
 
         $mech->content_contains('About you');
+        my $got_service_id;
+        $whitespace_mock->mock('GetCollectionSlots', sub {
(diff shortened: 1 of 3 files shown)


t/app/controller/waste_bexley_sharps.t
```diff
@@ -67,6 +71,8 @@ FixMyStreet::override_config {
                 }
             }
         );
+        $whitespace_mock->mock('GetCollectionSlots', sub { $slots_default });
```
nephila-nacrea: I am not sure what this second mock is doing - the tests pass if I remove it.

chrismytton: Think that was just resetting the mock, but you're right its not needed, have removed it in a fixup.

PR #5858, opened 2026-02-24
[Bexley][WW] Add reminder emails for sharps bookings

chrismytton:
Setup reminder emails for sharps bookings. This uses the existing mechanism that Bulky and Small items uses, just adds extra conditionals to handle sharps.

Fixes https://github.com/mysociety/societyworks/issues/5408

### DIFF
--- a/templates/email/default/waste/bulky-reminder.html
+++ b/templates/email/default/waste/bulky-reminder.html
@@ -29,10 +29,14 @@ cancel_url = cobrand.base_url _ '/waste/' _ property_id_uri _ '/' _ bulky_cancel
     [% END %]
   </p>
 
+  [% IF report.category == 'Sharps collection' %]
+    [% INCLUDE 'waste/_sharps_list.html' %]
+  [% ELSE %]
   <p style="[% p_style %]">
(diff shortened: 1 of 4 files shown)


templates/email/default/waste/bulky-reminder.html
```diff
@@ -51,7 +55,7 @@
     [% IF cobrand.moniker == 'peterborough' %]
       If you wish to cancel your booking, please call 01733 74 74 74.
     [% END %]
-  [% ELSIF cobrand.moniker != 'peterborough' AND cobrand.moniker != 'bexley' %]
+  [% ELSIF cobrand.moniker != 'peterborough' AND (cobrand.moniker != 'bexley' OR report.category == 'Sharps collection') %]
```
nephila-nacrea: Oh, it's occurred to me that the cancel link isn't shown for Bexley because you cannot cancel if there's only a day left? (Same for text email below). So I think you can leave it as the old behaviour.

chrismytton: Yep, good point! Have pushed a fixup to remove.

PR #5859, opened 2026-02-25
[Brent][WW] Implement new refuse container referral logic.

neprune:
[skip changelog] 

For https://github.com/mysociety/societyworks/issues/5376.


perllib/FixMyStreet/Cobrand/Brent.pm
```diff
@@ -1558,10 +1563,7 @@ sub waste_garden_mod_params {
 sub waste_post_report_creation {
     my ($self, $report, $data) = @_;
 
-    if (
-        $report->title =~ /Request new General rubbish bin \(grey bin\)/
-        && ($data->{request_reason} eq 'extra' || $data->{request_reason} eq 'damaged')
-        ) {
+    if ($report->title =~ /Request new General rubbish bin \(grey bin\)/) {
```
dracos, line 1596: I was worried we'd forgotten about `new_build` but then I saw that wasn't shown for grey bins, phew! :)

dracos (approved):
This all looks good, thanks for breaking it down so cleanly :)

I've pushed two extra commits on top if you'd like to check over those - one just moves the intro text a page earlier, as it seemed to make more sense there now every flow goes through this; and the second one switches the questions to be one per page with radio buttons instead (this means the validate isn't needed as that's now handled by skipping that page if 0 bins is selected).

dracos (approved):
Looks good :)
I've tidied up the commits a bit, moved some field name changes earlier so don't need the separate name change commit, and squashed a few others together

nephila-nacrea:
> Looks good :) I've tidied up the commits a bit, moved some field name changes earlier so don't need the separate name change commit, and squashed a few others together

@dracos Ah thanks a lot! I didn't get a notification of your comment or approval, presumably because I hadn't actually assigned myself to this ticket, but did see your comment on https://github.com/mysociety/societyworks/issues/5376.

I'll get this live.

PR #5860, opened 2026-02-25
[Bexley][WW] Sharps cancellation

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/5371 / https://github.com/mysociety/societyworks/issues/5394.

No open311-adapter changes needed as the existing CancelWorksheet logic suffices.

[skip changelog]

### DIFF
--- a/templates/web/base/waste/bin_days.html
+++ b/templates/web/base/waste/bin_days.html
@@ -305,7 +305,6 @@
           </div>
 
           <div class="waste-service-grid--service-description">
-            [% # XXX Handle sharps %]
             [% PROCESS 'waste/bulky/_bin_days_list.html' type='sharps' %]
           </div>
         </div>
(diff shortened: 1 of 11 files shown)


templates/web/base/waste/bin_days.html
```diff
@@ -294,7 +294,8 @@ <h3 id="bulky-waste" class="govuk-heading-m waste-service-name">Small items</h3>
         </div>
         [% END %]
 
-      [% IF property.show_sharps %]
+      [% # We only show sharps section when a user signed in %]
+      [% IF property.show_sharps && c.user_exists %]
```
chrismytton: This prevents the entire sharps section from being displayed to a signed out user, including the "Book a collection" link.

I think the cancel button should handle the sign-in for that controller action when needed. Looks like there are some calls with `c.user_exists` within `templates/web/base/waste/bulky/_bin_days_list.html` so that seems like the place for more targeted auth restrictions, if needed.

PR #5862, opened 2026-02-25
[Bexley][WW] Intro & (interim) T&Cs for sharps

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/5371 / https://github.com/mysociety/societyworks/issues/5374.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Bulky/Shared.pm
@@ -314,7 +314,16 @@ has_field tandc => (
         } elsif ($c->cobrand->moniker eq 'brent') {
             $label = 'I have read and agree to the <a href="' . $link . '" target="_blank">terms and conditions</a> and understand any additional items presented that do not meet the terms and conditions will not be collected';
         } elsif ($c->cobrand->moniker eq 'bexley') {
-            $label = << 'HERE';
+            if ( $c->stash->{sharps} ) {
+                $label = << 'HERE';
+&bull; I understand that collections can take place any time after 6am on the chosen collection day. Items must be accessible at the location given and left in a neat and safe manner. Items cannot be left for collection on the public highway.
+<br>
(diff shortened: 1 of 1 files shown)


templates/web/base/waste/sharps/intro.html
```diff
@@ -1,4 +1,14 @@
...
+  <li>request the delivery of sharps boxes</li>
+  <li>request the collection of full-up sharps boxes</li>
+</ul>
 
-[% # XXX %]
-<h2>TODO</h2>
+<h3>Before you start</h3>
+You need to provide the following information:
+<ul>
+  <li>number of clinical waste bags that need to be collected or delivered</li>
```
chrismytton: Seems odd that this is referring to clinical waste bags. Feels like it should say "number of sharps boxes that…" for consistency?

templates/web/base/waste/sharps/intro.html
```diff
@@ -1,4 +1,14 @@
-<h3>Before you start your booking</h3>
+<h3>Request a sharps box delivery or collection</h3>
+Use this service to:
```
chrismytton: I think this should be wrapped in a `<p>` tag so it picks up the standard paragraph styles.

templates/web/base/waste/sharps/intro.html
```diff
@@ -1,4 +1,14 @@
...
+<ul>
+  <li>request the delivery of sharps boxes</li>
+  <li>request the collection of full-up sharps boxes</li>
+</ul>
 
-[% # XXX %]
-<h2>TODO</h2>
+<h3>Before you start</h3>
+You need to provide the following information:
```
chrismytton: And here, I think this should be wrapped in a `<p>` tag so it picks up the standard paragraph styles.

chrismytton (changes requested):

templates/web/base/waste/sharps/intro.html
nephila-nacrea: Yes I'd noticed that and then forgot to change it!

PR #5866, opened 2026-02-26
[Sutton] Per container pricing

MorayMySoc:
Allow different containers to have different cost/admin charges

https://github.com/mysociety/societyworks/issues/5429

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Sutton.pm
+++ b/perllib/FixMyStreet/Cobrand/Sutton.pm
@@ -83,6 +83,19 @@ my %CONTAINERS = (
 );
 lock_hash(%CONTAINERS);
 
+=head2 CONTAINER_SWAPS_FROM
+
+This is a map for special case pricing when changing to the container size in the key from
+the container sizes in the value. If one of these swaps is detected then the change price
+of `request_change_cost_${value}_${key}` will be used instead of the usual `change_cost_$key`.
+
+=cut
+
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -711,7 +712,8 @@ Calculate how much, if anything, a request for a container should be.
 sub request_cost {
     my ($self, $id, $containers) = @_;
     my $costs = WasteWorks::Costs->new({ cobrand => $self });
-    if (my $cost = $costs->get_cost('request_change_cost')) {
+    my %id_to_name = reverse %CONTAINERS;
+    if (my $cost = $costs->get_cost('request_change_cost_' . $id_to_name{$id})) {
```
struan: might be nice to set `$container_name` once and use it here and below

t/app/controller/waste_sutton_r.t
```diff
@@ -117,8 +117,13 @@ FixMyStreet::override_config {
         echo => { sutton => { bulky_service_id => 960 }},
         payment_gateway => { sutton => {
             cc_url => 'http://example.com',
-            request_replace_cost => 500,
-            request_change_cost => 1500,
+            request_change_cost_refuse_140 => 1500,
```
struan: I wonder if we could make these not all the same and then update the tests just to have some "yup, there are different prices" confidence

PR #5867, opened 2026-02-26
Optional PostGIS support

dracos:
Fixes https://github.com/mysociety/societyworks/issues/5433
Firstly upgrade GH actions to Ubuntu 24 because that makes it easier to install postgis.
Then add a normal index on problem.confirmed, there are a number of queries that filter/order by confirmed, I'm not sure why this wasn't already present TBH!
Then a46b943508f8a07222ba6d9916fd5a5201aeeb73 is the main part - this changes one Perl version GH action to run with postgis installed (using a `WITH_POSTGIS` environment variable) and the fixture change to postgis functions.

This doesn't remove any existing code - if the `FMS_DB_POSTGIS` variable is set, then the email alerts script Alerts.pm and the Nearby ResultSet (used by RSS feeds, nearby duplicate reports, etc) run different new code to make the same queries, but using PostGIS functions (and assuming that a PostGIS index has been created).

The clever part is that in order to search by distance but sort by reverse date, we spatially index both the lat/lon and the confirmed timestamp, letting us then find reports around a point ordered by how far they are away from the far future.

For an email alert query:
```
bci=# explain analyze select problem.id, problem.state,problem.confirmed, problem.cobrand
    from problem_find_nearby(51.526802035, -0.104179447, 2.7) nearby, problem
            where
                nearby.problem_id = problem.id
            and problem.state not in ('unconfirmed', 'hidden', 'partial')
            and problem.non_public = 'f'
            and extract(epoch from problem.confirmed) >= extract(epoch from current_timestamp - '7 days'::interval)
            and (select whenqueued from alert_sent where alert_sent.alert_id = 6832 and alert_sent.parameter = problem.id::text) is null
            order by confirmed desc;

 Sort  (cost=268168.12..268168.12 rows=1 width=34) (actual time=1211.121..1211.129 rows=24 loops=1)
   Sort Key: problem.confirmed DESC
   Sort Method: quicksort  Memory: 26kB
   ->  Nested Loop  (cost=121285.05..268168.11 rows=1 width=34) (actual time=717.118..1211.072 rows=24 loops=1)
         ->  Sort  (cost=121284.62..121306.63 rows=8804 width=12) (actual time=685.591..704.326 rows=69341 loops=1)
               Sort Key: (('6372.8'::double precision * acos((trunc(((('0.7828992734514981'::double precision * sin(radians(problem_1.latitude))) + (('0.6221484771572751'::double precision * cos(radians(problem_1.latitude))) * cos(radians(('-0.104179447'::double precision - problem_1.longitude))))))::numeric, 14))::double precision))) DESC
               Sort Method: quicksort  Memory: 6323kB
               ->  Bitmap Heap Scan on problem problem_1  (cost=22961.85..120707.78 rows=8804 width=12) (actual time=124.550..656.618 rows=69341 loops=1)
                     Recheck Cond: ((radians(latitude) > '0.8988886729126736'::double precision) AND (radians(latitude) < '0.8997360241554554'::double precision) AND (radians(longitude) > '-0.002499262353894671'::double precision) AND (radians(longitude) < '-0.0011372863722191866'::double precision))
                     Filter: (('6372.8'::double precision * acos((trunc(((('0.7828992734514981'::double precision * sin(radians(latitude))) + (('0.6221484771572751'::double precision * cos(radians(latitude))) * cos(radians(('-0.104179447'::double precision - longitude))))))::numeric, 14))::double precision)) < '2.7'::double precision)
                     Rows Removed by Filter: 25092
                     Heap Blocks: exact=87210
                     ->  Bitmap Index Scan on problem_radians_latitude_longitude_idx  (cost=0.00..22959.65 rows=26412 width=0) (actual time=107.642..107.643 rows=95030 loops=1)
                           Index Cond: ((radians(latitude) > '0.8988886729126736'::double precision) AND (radians(latitude) < '0.8997360241554554'::double precision) AND (radians(longitude) > '-0.002499262353894671'::double precision) AND (radians(longitude) < '-0.0011372863722191866'::double precision))
         ->  Index Scan using problem_pkey on problem  (cost=0.43..16.67 rows=1 width=34) (actual time=0.007..0.007 rows=0 loops=69341)
               Index Cond: (id = problem_1.id)
               Filter: ((NOT non_public) AND (state <> ALL ('{unconfirmed,hidden,partial}'::text[])) AND (date_part('epoch'::text, confirmed) >= date_part('epoch'::text, (CURRENT_TIMESTAMP - '7 days'::interval))) AND ((SubPlan 1) IS NULL))
               Rows Removed by Filter: 1
               SubPlan 1
                 ->  Index Scan using alert_sent_alert_id_parameter_idx on alert_sent  (cost=0.57..8.59 rows=1 width=8) (actual time=0.011..0.011 rows=1 loops=136)
                       Index Cond: ((alert_id = 6832) AND (parameter = (problem.id)::text))
 Planning Time: 1.590 ms
 Execution Time: 1211.224 ms
(23 rows)
```

Changes to

```
bci=# explain analyze select problem.id, problem.state,problem.confirmed, problem.cobrand
    from problem
            where
                ST_DWithin(
                    ST_SetSRID(ST_Point(longitude, latitude), 4326)::geography,
                    ST_SetSRID(ST_Point(-0.104179447, 51.526802035), 4326)::geography,
                    2700)
            and problem.state not in ('unconfirmed', 'hidden', 'partial')
            and problem.non_public = 'f'
            and extract(epoch from problem.confirmed) >= extract(epoch from current_timestamp - '7 days'::interval)
            and (select whenqueued from alert_sent where alert_sent.alert_id = 6832 and alert_sent.parameter = problem.id::text) is null
            order by (extract(epoch from problem.confirmed) <-> extract(epoch from '3000-01-01'::date));

 Index Scan using problem_location_gist_idx on problem  (cost=0.55..5067.22 rows=1 width=42) (actual time=1.010..9.909 rows=24 loops=1)
   Index Cond: (((st_setsrid(st_point(longitude, latitude), 4326))::geography && _st_expand('0101000020E610000005C6C71581ABBABF48E5C33F6EC34940'::geography, '2700'::double precision)) AND (date_part('epoch'::text, confirmed) >= date_part('epoch'::text, (CURRENT_TIMESTAMP - '7 days'::interval))))
   Order By: (date_part('epoch'::text, confirmed) <-> '32503680000'::double precision)
   Filter: ((NOT non_public) AND (state <> ALL ('{unconfirmed,hidden,partial}'::text[])) AND ((SubPlan 1) IS NULL) AND st_dwithin((st_setsrid(st_point(longitude, latitude), 4326))::geography, '0101000020E610000005C6C71581ABBABF48E5C33F6EC34940'::geography, '2700'::double precision, true))
   Rows Removed by Filter: 317
   SubPlan 1
     ->  Index Scan using alert_sent_alert_id_parameter_idx on alert_sent  (cost=0.57..8.59 rows=1 width=8) (actual time=0.008..0.008 rows=0 loops=320)
           Index Cond: ((alert_id = 6832) AND (parameter = (problem.id)::text))
 Planning Time: 0.848 ms
 Execution Time: 10.052 ms
(10 rows)
```

Similar for other queries you can think of.

### DIFF
--- a/perllib/FixMyStreet/Script/Alerts.pm
+++ b/perllib/FixMyStreet/Script/Alerts.pm
@@ -170,7 +170,7 @@ sub send_alert_type {
             _extra_new_update_data($row, \%data, $cobrand, $schema);
         # this is ward and council problems
         } else {
-            _extra_new_area_data($row, $ref);
+            _extra_new_area_data($row, $ref, 'send_alert_type');
         }
         push @{$data{data}}, $row;
 
@@ -248,9 +248,9 @@ sub _extra_new_update_data {
 }
 
 sub _extra_new_area_data {
-    my ($row, $ref) = @_;
+    my ($row, $ref, $caller) = @_;
 
-    if ( $ref =~ /ward|council/ ) {
+    if ( $ref =~ /ward|council|local_problems/ ) {
         my $nearest_st = FixMyStreet::Geocode::Address->new($row->{geocode})->for_alert;
         $row->{nearest} = $nearest_st;
     }
@@ -258,6 +258,8 @@ sub _extra_new_area_data {
     my $dt = $parser->parse_timestamp( $row->{confirmed} );
     FixMyStreet->set_time_zone($dt);
     $row->{confirmed} = $dt;
+    $row->{confirmed_str} = $dt->strftime('%Y-%m-%d %H:%M:%S')
+        if $caller eq 'send_local_find_nearby';
 
     # Hack in the image for the non-object reports
     my $photo = $row->{photo};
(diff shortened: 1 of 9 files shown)


perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -275,6 +277,62 @@ sub _extra_new_area_data {
...
+            where
+                ST_DWithin(ST_SetSRID(ST_Point(longitude, latitude), 4326)::geography, ST_SetSRID(ST_Point(?, ?), 4326)::geography, ?)
+            and problem.state in ($states)
+            and problem.non_public = 'f'
+            and extract(epoch from problem.confirmed) >= ?
+            and extract(epoch from problem.confirmed) >= extract(epoch from current_timestamp - '7 days'::interval)
+            and (select whenqueued from alert_sent where alert_sent.alert_id = ? and alert_sent.parameter = problem.id::text) is null
+            and user_id != ?
+            order by (extract(epoch from problem.confirmed) <-> extract(epoch from '3000-01-01'::date))";
+        $q = FixMyStreet::DB->schema->storage->dbh->prepare($q);
```
chrismytton: Doubt it makes much difference but this could be moved outside the while loop so the prepare only runs once.

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -248,16 +248,18 @@ sub _extra_new_update_data {
...
+    if ( $ref =~ /ward|council|local_problems/ ) {
         my $nearest_st = FixMyStreet::Geocode::Address->new($row->{geocode})->for_alert;
         $row->{nearest} = $nearest_st;
     }
 
     my $dt = $parser->parse_timestamp( $row->{confirmed} );
     FixMyStreet->set_time_zone($dt);
     $row->{confirmed} = $dt;
+    $row->{confirmed_str} = $dt->strftime('%Y-%m-%d %H:%M:%S')
+        if $caller eq 'send_local_find_nearby';
```
chrismytton, line 262: What's the reason for setting this conditionally?

chrismytton (commented):
Very impressive speedup!! All looks good, couple of comments but nothing blocking.

Might be good to add a changelog entry, as the failing check suggests?

perllib/FixMyStreet/Script/Alerts.pm
dracos: `confirmed_str` is only used by the `find_nearby` branch (because the new way does it in the query itself). Probably doesn't make much difference either.

perllib/FixMyStreet/Script/Alerts.pm
dracos: Ooh yes, ta

PR #5868, opened 2026-02-27
[Rutland][Confirm] Add Roles for Confirm sending

MorayMySoc:
Adds the roles for sending to Confirm and to a Multi backend.

Engineers previous open311_extra_data_include to work with Confirm role's open311_extra_data_include in case of changes to Confirm or Salesforce though unlikely.

https://github.com/mysociety/societyworks/issues/5403

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Rutland.pm
+++ b/perllib/FixMyStreet/Cobrand/Rutland.pm
@@ -13,6 +13,12 @@ Rutland is a unitary authority, with a Salesforce backend.
 package FixMyStreet::Cobrand::Rutland;
 use base 'FixMyStreet::Cobrand::UKCouncils';
 
+use Moo;
+with 'FixMyStreet::Roles::ConfirmOpen311';
+with 'FixMyStreet::Roles::ConfirmValidation';
+with 'FixMyStreet::Roles::Cobrand::OpenUSRN';
+with 'FixMyStreet::Roles::Open311Multi';
+
 use strict;
 use warnings;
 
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Rutland.pm
```diff
@@ -13,6 +13,12 @@ Rutland is a unitary authority, with a Salesforce backend.
 package FixMyStreet::Cobrand::Rutland;
 use base 'FixMyStreet::Cobrand::UKCouncils';
 
+use Moo;
+with 'FixMyStreet::Roles::ConfirmOpen311';
+with 'FixMyStreet::Roles::ConfirmValidation';
```
dracos, line 18: This brings in a report_validation which is then being overwritten by the one present in this file. So that needs adjusting or replacing - I think easiest solution is to remove the report_validation from Rutland.pm and setting max_title_length to 254 and might have to do something about the 40-max name, though, perhaps another `around` would be enough.

perllib/FixMyStreet/Cobrand/Rutland.pm
```diff
@@ -51,20 +57,25 @@ sub open311_config {
...
-    my ($self, $row, $h) = @_;
-
-    return [
-        { name => 'external_id', value => $row->id },
-        { name => 'title', value => $row->title },
-        { name => 'description', value => $row->detail },
-        $h->{closest_address} ? { name => 'closest_address', value => "$h->{closest_address}" } : (),
-    ];
-}
+around open311_extra_data_include => sub {
```
dracos, line 75: This is fine, though as salesforce one is superset of Confim, looks like, you could call $orig always, and then if not Confirm add on external_id and closest_address as an alternative.

dracos (changes requested):
Bit of change needed with report_validation and one possible elsewhere is all :)

PR #5870, opened 2026-03-02
[Sutton][Waste] enable different prices for bulky collections containing POPs

struan:
Update the waste admin to enable charging different amounts for bulky collections containing items that may have POPs.
Update the pricing strategies etc to take account of this when calculating and displaying prices to the resident.

This assumes that only the banded and standard pricing strategies are affected by this so doesn't update the other two.

[skip changelog]

Fixes mysociety/societyworks#5428

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Waste.pm
@@ -76,12 +76,15 @@ sub edit : Chained('body') : PathPart('') : Args(0) {
             $new_cfg = $c->stash->{body}->get_extra_metadata("wasteworks_config", {});
             my %keys = (
                 per_item_costs => 'bool',
+                pop_costs => 'bool',
                 per_item_min_collection_price => 'int',
                 base_price => 'int',
+                base_pop_price => 'int',
                 daily_slots => 'int',
                 items_per_collection_max => 'int',
                 small_items_per_collection_max => 'int',
                 band1_price => 'int',
+                band1_pop_price => 'int',
                 band1_max => 'int',
                 free_mode => 'bool',
                 food_bags_disabled => 'sel',
@@ -144,6 +147,7 @@ sub booked_items : Private {
         $cobrand->call_hook('bulky_per_item_pricing_property_types');
 
     $c->stash->{item_points} = $cobrand->call_hook('bulky_points_per_item_pricing');
+    $c->stash->{pop_costs} = $cobrand->call_hook('bulky_pop_item_costs');
 
     if ($c->req->method eq 'POST') {
         $c->forward('/auth/check_csrf_token');
@@ -166,6 +170,9 @@ sub booked_items : Private {
             };
             if ($c->stash->{item_points}) {
                 $item->{points} = $c->get_param("points[$i]");
+            }
+            if ($c->stash->{pop_costs}) {
+                $item->{contains_pops} = $c->get_param("contains_pops[$i]");
             }
 
             foreach my $property_type (@{$c->{stash}->{per_item_pricing_property_types}}) {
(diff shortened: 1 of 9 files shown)


templates/web/base/admin/waste/edit.html
```diff
@@ -33,6 +33,14 @@ <h3>Bulky Collections</h3>
         value="[% base_price %]">
     </div>
 
+    [% IF pop_costs %]
+    <div>
+    <label for="waste_base_pop_price">Price for a collection with POPs(in <strong>pence</strong>)</label>
```
davea: Missing a space in `POPs(in`

perllib/FixMyStreet/App/Controller/Admin/Waste.pm
```diff
@@ -167,6 +171,10 @@ sub booked_items : Private {
             if ($c->stash->{item_points}) {
                 $item->{points} = $c->get_param("points[$i]");
             }
+            if ($c->stash->{pop_costs}) {
+                $item->{contains_pops} = $c->get_param("contains_pops[$i]");
+                $c->log->info($i . ": ".  $c->get_param("contains_pops[$i]"));
```
davea: Debug logging left in?

perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
```diff
@@ -152,7 +153,18 @@ sub bulky_pricing_strategy {
         $out = { strategy => 'per_item', min => $min_collection_price };
     } elsif (my $band1_price = $self->wasteworks_config->{band1_price}) {
         my $max = $self->{c}->stash->{booking_maximum};
-        $out = { strategy => 'banded', bands => [ { max => $band1_max, price => $band1_price }, { max => $max, price => $base_price } ] };
+        if ($self->wasteworks_config->{pop_costs}) {
+            $out = { strategy => 'banded_pop', bands =>
```
davea, line 163: These new `banded_pop`/`pop_costs` strategies aren't handled in `updateTotal` in `waste.js`, so on the 'Add items for collection' page the incorrect price is shown at the bottom:

Image

templates/web/base/admin/waste/bulky_items.html
```diff
@@ -64,6 +70,8 @@
                     <th>Points</th>
                 [% ELSIF per_item_costs %]
                     <th>Price (pence)</th>
+                [% ELSIF pop_costs %]
```
davea: If both `per_item_costs` and `pop_costs` are enabled the table header is incorrect:

Image

perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm
davea: Similarly the `[% IF c.cobrand.bulky_pricing_strategy.strategy == 'banded' %]` line in `waste/bulky/items.html` won't show the pricing bands if strategy is `banded_pop`.

templates/web/base/admin/waste/bulky_items.html
struan: tbh I'm not sure we should allow both of these to be checked at the same time as it doesn't really make sense. If you have per item pricing then you don't need POP costs. I'll fix it in the meantime.

templates/web/base/waste/bulky/intro.html
```diff
@@ -60,15 +60,27 @@ <h2>Before you book</h2>
...
+        SET band1_pop_price = cfg.band1_pop_price / 100;
     %]
     <li>The price depends on how many items you would like collected:
         <ul class="bulky-intro-pricing">
             <li class="bulky-intro-pricing">
                 1 to [% band1_items %] items cost £[% pounds(band1_price) %]
             </li>
+            [% IF cfg.pop_costs %]
+            <li class="bulky-intro-pricing">
+                1–[% band1_items %] items containing POPs = £[% pounds(band1_pop_price) %]
```
dracos: I think this could be confusing, as could imply the cost is only for "items containing POPs" rather than it being the cost of 1-4 items if any of them contains POPs. I'm going to change to "1-4 items cost £N (£M if any item contains POPs)"

PR #5871, opened 2026-03-03
[Surrey] Add auto-close script for Parking illegally reports

davea:
For FD-6672

[skip changelog]

### DIFF
--- a/bin/surrey/auto-close-reports
+++ b/bin/surrey/auto-close-reports
@@ -1,0 +1,43 @@
+#!/usr/bin/env perl
+#
+# Close reports in 'Parking illegally' category older than a number of days
+
+use strict;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use CronFns;
+use Getopt::Long::Descriptive;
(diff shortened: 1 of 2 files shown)


bin/surrey/auto-close-reports
```diff
@@ -0,0 +1,43 @@
...
+use FixMyStreet::Script::UK::AutoClose;
+
+my $site = CronFns::site(FixMyStreet->config('BASE_URL'));
+CronFns::language($site);
+
+my $default_text = 'Thank you for this report. It has been sent to our team of enforcement officers who will have reviewed the case and taken action where they were able to do so. This report has now been closed automatically.';
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['commit!', "actually close reports and send emails. Omitting this flag will do a dry-run"],
```
MorayMySoc, line 26: Not sure this exclamation mark is supposed to be here

davea: [It means](https://metacpan.org/pod/Getopt::Long#A-little-bit-less-simple-options) you can call it with `--no-commit` if you want to be really explicit about your intentions:

```
./bin/surrey/auto-close-reports --help
auto-close-reports [-h] [long options...]
        --[no-]commit       actually close reports and send emails. Omitting
                            this flag will do a dry-run
```

(I just copied it from Enfield :))

PR #5872, opened 2026-03-03
[Bromley] Different domestic/trade minimum pricing

dracos:
Fixes https://github.com/mysociety/societyworks/issues/5422
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
+++ b/perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
@@ -299,6 +299,8 @@ sub view : Private {
     $c->stash->{property} = {
         id => $p->waste_property_id,
         address => $p->get_extra_metadata('property_address'),
+        # For Bromley, different refund depending on if property was domestic/trade, stored at time of booking
+        pricing_property_type => $p->get_extra_field_value('property_type') || 'Domestic',
     };
 
     my $items_extra;
--- a/t/app/controller/waste_bromley_bulky.t
+++ b/t/app/controller/waste_bromley_bulky.t
@@ -12,6 +12,7 @@ my $mech = FixMyStreet::TestMech->new;
 my $sample_file = path(__FILE__)->parent->child("sample.jpg");
 my $sample_file_2 = path(__FILE__)->parent->child("sample2.jpg");
 my $minimum_charge = 500;
+my $minimum_charge_trade = 2700;
 
 my $user = $mech->create_user_ok('bob@example.org');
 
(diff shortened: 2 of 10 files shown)


perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
```diff
@@ -271,6 +271,8 @@ sub view : Private {
     $c->stash->{property} = {
         id => $p->waste_property_id,
         address => $p->get_extra_metadata('property_address'),
+        # For Bromley, different refund depending on if property was domestic/trade, stored at time of booking
+        pricing_property_type => $p->get_extra_field_value('property_type'),
```
davea: Does this matter for bookings that have already happened and don't have that extra field set? Should we explicitly default to Domestic here if we don't know it?

Also, I don't think it's this line exactly, but something is now causing a `Use of uninitialized value in concatenation (.) or string at [...]/perllib/FixMyStreet/Roles/Cobrand/BulkyWaste.pm line 759.` warning when running the `"Cancel booking when staff marks as payment failed"` subtest.

templates/web/bromley/waste/bulky/cancel_intro.html
```diff
@@ -1,7 +1,7 @@
 [% USE pounds = format('%.2f') %]
 [% partial = c.cobrand.bulky_refund_would_be_partial(cancelling_booking) %]
 [% IF partial %]
-  [% minimum_charge_amount = c.cobrand.bulky_minimum_charge %]
+  [% minimum_charge_amount = c.cobrand.bulky_per_item_min_collection_price %]
   [% minimum_charge = pounds(minimum_charge_amount / 100) %]
   [% partial_explainer = "Cancellations within 24 hours of collection are only eligibile to be refunded the amount paid above the minimum charge £" _ minimum_charge _ "." %]
```
davea: Not part of this PR but just spotted the typo on `eligibile`.

t/app/controller/waste_bromley_bulky.t
```diff
@@ -631,23 +633,26 @@ FixMyStreet::override_config {
...
         $echo->mock('GetPointAddress', sub { return { %$address,
             PointAddressType => { Id => 151, Name => 'Commercial - Agricultural' }
         } });
-        test_prices('£20.00', '£20.00');
+        test_prices('£27.00', '£27.00', '£40.00');
         $echo->mock('GetPointAddress', sub { $address });
-        test_prices('£10.00', '£10.00');
-    };
-
-    subtest 'Minimum charged enforced' => sub {
```
davea, line 645: Is the removal of this test deliberate?

dracos: Yeah, it's covered by the one above now, because the minimum trade cost is higher than the items cost.

perllib/FixMyStreet/App/Controller/Waste/Bulky.pm
dracos: It's staff CSC payment failure trying to calculate a refund price (shouldn't be getting there because payment failure means there was no payment, added check to refund calculation to bail out then).

PR #5873, opened 2026-03-03
[Lincolnshire] Include external reference in all customer emails from .com

davea:
When a Lincolnshire report is made via fixmystreet.com, the Confirm external reference number was only shown in the initial confirm_report_sent email, not update/alert emails.

For FD-6674

NB this deliberately only affects Lincs reports, but perhaps we could/ should use can_display_external_id instead so it would apply to all cobrands that show external IDs.

[skip changelog]

### DIFF
--- a/templates/email/fixmystreet.com/_council_reference.html
+++ b/templates/email/fixmystreet.com/_council_reference.html
@@ -1,5 +1,6 @@
-[% IF sent_confirm_id_ref ~%]
-<p style="[% p_style %]">The report's reference number is <strong>[% sent_confirm_id_ref %]</strong>.
+[% ref = sent_confirm_id_ref || (problem.to_body_named('Lincolnshire') AND problem.external_id) %]
+[% IF ref ~%]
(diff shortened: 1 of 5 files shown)


templates/email/fixmystreet.com/_council_reference.html
```diff
@@ -1,5 +1,6 @@
-[% IF sent_confirm_id_ref ~%]
-<p style="[% p_style %]">The report's reference number is <strong>[% sent_confirm_id_ref %]</strong>.
+[% ref = sent_confirm_id_ref || (problem.to_body_named('Lincolnshire') AND problem.external_id) %]
+[% IF ref ~%]
+<p style="[% p_style %]">The report's reference number is <strong>[% sent_confirm_id_ref || problem.external_id %]</strong>.
```
chrismytton, line 3: Could this line and the other similar ones below print out `ref` instead of repeating the `sent_confirm_id_ref || problem.external_id` logic?

PR #5874, opened 2026-03-04
[Sutton][WW] Add premium food caddy

struan:
~~This change is mostly to make sure the fields on the form remain in the correct order because we can no longer rely on different pricing to differentiate between replace and change options so just make that explicit in the code~~

This also enables a second type of kitchen food caddy

[skip changelog]

Fixes mysociety/societyworks#5426


perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -723,8 +726,8 @@ sub request_cost {
         }
     }
     if (my $cost = $costs->get_cost('request_replace_cost_' . $id_to_name{$id})) {
-        foreach ($CONTAINERS{refuse_140}, $CONTAINERS{refuse_240}, $CONTAINERS{refuse_360}, $CONTAINERS{paper_240}) {
-            if ($id == $_ && $containers->{$_}) {
+        foreach ($CONTAINERS{refuse_140}, $CONTAINERS{refuse_240}, $CONTAINERS{refuse_360}, $CONTAINERS{paper_240}, $CONTAINERS{food_indoor_premium}) {
+            if ($id == $CONTAINERS{food_indoor_premium} || ( $id == $_ && $containers->{$_} )) {
```
chrismytton: I don't think `$CONTAINERS{food_indoor_premium}` in the foreach list is needed?

When `$id` is 45, the `$id == $CONTAINERS{food_indoor_premium}` check is true on every iteration, so the function always returns on the first one.

perllib/FixMyStreet/Roles/Cobrand/SLWP2.pm
```diff
@@ -340,7 +341,15 @@ sub waste_service_containers {
...
     my $food_indoor_key = $self->moniker eq 'merton' ? 'food_indoor_7' : 'food_indoor_5';
-    if ($service_name =~ /Food/ && !$self->{c}->stash->{quantities}->{$CONTAINERS{$food_indoor_key}}) {
+    my $food_indoor_premium_key = 'food_indoor_10';
+    if ($service_name =~ /Food/ && !(
+            $self->{c}->stash->{quantities}->{$CONTAINERS{$food_indoor_key}}
+            || $self->{c}->stash->{quantities}->{$CONTAINERS{$food_indoor_premium_key}}
+        ) ) {
+        if ($self->moniker eq 'sutton') {
+                $request_max->{$CONTAINERS{$food_indoor_premium_key}} = 1;
+                push @$containers, $CONTAINERS{$food_indoor_premium_key}; # Premium food waste bin (kitchen)
```
chrismytton: These two lines are over-indented by one level.

chrismytton (approved):
Looks good! Couple of minor non-blocking comments.

perllib/FixMyStreet/Cobrand/Sutton.pm
dracos: Noting this has been changed to not have the foreach loop at all, as we can now rely on the config existing to decide if something is chargeable or not.

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -713,24 +737,33 @@ sub request_cost {
...
+    # increasing container size does not cost the same as decreasing so put in a mechanism
+    # to allow specific prices for moving from one container size to another with a fallback
+    # to the generic change price for a container size
+    if (my $froms = $CONTAINER_SWAPS_FROM{$cost_id}) {
+        for my $from_container (@$froms) {
+            $cost_id = $from_container . '_' . $cost_id if $containers->{$CONTAINERS{$from_container}};
+        }
+    }
+    if (my $cost = $costs->get_cost('request_change_cost_' . $cost_id)) {
+        foreach ($CONTAINERS{refuse_140}, $CONTAINERS{refuse_240}, $CONTAINERS{paper_240}, $CONTAINERS{paper_360}) {
```
dracos: As with the replace, do we need this foreach any more, can we instead rely on the config existing to know that we want to charge?

dracos: (pushed fixup for this matching 781a60d, think that makes sense)

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -83,6 +84,20 @@ my %CONTAINERS = (
...
+This is a map for special case pricing when changing to the container size in the key from
+the container sizes in the value. If one of these swaps is detected then the change price
+of `request_change_cost_${value}_${key}` will be used instead of the usual `change_cost_$key`.
+
+=cut
+
+my %CONTAINER_SWAPS_FROM = (
+    refuse_140 => ['refuse_240'],
+    refuse_240 => ['refuse_360'],
+    paper_240 => ['paper_360'],
```
dracos, line 98: Fine to leave this in in case things change in future, but noting that I don't think the paper needs this behaviour, because:
140->240 is replace cost (as you have to do this if you have a 140L)
240->240 is replace cost
240->360 is change cost
360->240 is replace cost.
So all "change to 240" is the one cost and thus could be done with standard config, iyswim.
But they haven't confirmed this yet have they, so will see.

So new config is, just to check:
      request_change_cost_paper_360: 3000
      request_change_cost_paper_360_paper_240: 1500
      request_replace_cost_paper_360: 1500

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -75,6 +75,7 @@ my %CONTAINERS = (
     paper_1100 => 32,
     paper_bag => 34,
     food_indoor => 43,
+    food_indoor_premium => 1981,
```
dracos: I think this should be 85, for the Container Request Container Type CodeFile.

perllib/FixMyStreet/Cobrand/Sutton.pm
```diff
@@ -451,6 +453,7 @@ sub waste_containers {
         $CONTAINERS{paper_240} => 'Paper and Cardboard Green Wheelie Bin (240L)',
         $CONTAINERS{paper_140} => 'Paper and Cardboard Green Wheelie Bin (140L)',
         $CONTAINERS{food_indoor} => 'Small Kitchen Food Waste Caddy (7L)',
+        $CONTAINERS{food_indoor_premium} => 'Premium Kitchen Food Waste Caddy (10L)',
```
dracos: The description in Echo is "Premium Food Caddy 23L" - is it 10L or 23L?

perllib/FixMyStreet/Roles/Cobrand/SLWP2.pm
```diff
@@ -128,6 +128,8 @@ my %CONTAINERS = (
     paper_140 => 26,
     food_indoor_5 => 43,
     food_indoor_7 => 44,
+    food_indoor_10 => 45,
+    food_indoor_premium => 1981,
```
dracos: Ditto, plus no need for 10 line any more I assume

perllib/FixMyStreet/Cobrand/Sutton.pm
dracos: Confirmed this should be 23L

dracos (changes requested):

PR #5875, opened 2026-03-05
[Dumfries] Custom response template lookups

davea:
Dumfries external status codes from Alloy are colon-separated values (status:outcome:priority). This adds support for wildcard matching against these values in response templates, allowing a single template to match multiple incoming codes from Open311 updates.

For example a response template whose `external_status_code` was `abc:*:*` would be used for incoming updates with external status codes of `abc:123:456`, `abc::789`, `abc:def:`, etc.

Also adds a nicer UI to the response template editor in the Dumfries admin which means you can configure how/what the template matches without having to see the 24-char Alloy item IDs.

image

(The values in these dropdowns come from a `DB::Config` entry, this is set up on staging so you can copy that JSON blob locally)

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Templates.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Templates.pm
@@ -107,6 +107,16 @@ sub edit : Path : Args(2) {
             $ext_code ||= $c->get_param('external_status_code');
             $template->external_status_code($ext_code);
 
+            # Allow cobrands to validate the external_status_code format
+            # Use the body's cobrand handler, not the page cobrand
+            my $body_cobrand = $c->stash->{body}->get_cobrand_handler;
+            if ($body_cobrand && (my $error = $body_cobrand->call_hook(
+                validate_response_template_external_status_code => $ext_code
+            ))) {
+                $c->stash->{errors} ||= {};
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/DB/Result/Problem.pm
```diff
@@ -936,10 +936,28 @@ sub response_template_for {
...
+                    'me.state' => '',
+                    'me.external_status_code' => { '!=' => '' },  # Must have a pattern
+                    # Use -bool with a literal SQL ref for the regex match
+                    -bool => \[$regex_match->{sql}, @{$regex_match->{bind}}],
+                };
+                $order = { order_by => \$regex_match->{order} };
+            } else {
+                # The double comma option here dates from Echo updates with only a
+                # resolution code ID being stored in the database as "CODE,,". Once
+                # they're all updated to only be CODE, this can be removed.
```
davea, line 957: Can this now be removed?

dracos: I don't remember the details behind this without more checking, but there are still 36 rows in both live and staging of the form `ID,,` from Bromley, Kingston, Sutton. I don't know if they can just be updated, I assume probably? (But then, why not done at the time?)

PR #5876, opened 2026-03-05
[TfL] Form refactoring

dracos:
Just bits that fell out of TfL work to make that easier. Pulling out first as also can be used by other bits in progress.


MorayMySoc (approved):
This seems good to me

PR #5877, opened 2026-03-09
[Dumfries] Ignore any already seen photo in update

dracos:
[skip changelog]

### DIFF
--- a/t/cobrand/dumfries.t
+++ b/t/cobrand/dumfries.t
@@ -289,6 +289,38 @@ FixMyStreet::override_config {
             'Inspection time not cleared when not in extras';
     };
 
+    subtest 'photo already seen not included again' => sub {
+        my $request = {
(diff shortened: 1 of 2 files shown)


t/cobrand/dumfries.t
```diff
@@ -289,6 +289,34 @@ FixMyStreet::override_config {
...
+            state      => 'confirmed',
+            confirmed  => DateTime->now,
+            photo => '74e3362283b6ef0c48686fb0e161da4043bbcc97',
+        });
+
+        my $request = {
+            extras => {},
+        };
+
+        $cobrand->open311_get_update_munging($comment, 'investigating', $request);
```
davea, line 319: Wonder if it's worth running this on `$existing` too just to prove that photo does get processed the first time. But maybe that's overly cautious.

PR #5879, opened 2026-03-09
[Shropshire] Add ‘Ward’/‘Parish’ suffix to area names on /reports and dashboard

davea:
For FD-6717

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Shropshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Shropshire.pm
@@ -259,6 +259,20 @@ sub add_parish_wards {
         %$areas,
         %$extra_areas
     );
+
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Cobrand/Shropshire.pm
```diff
@@ -259,6 +259,20 @@ sub add_parish_wards {
         %$areas,
         %$extra_areas
     );
+
+    if ($self->{c}->action =~ /^dashboard\/(index|heatmap)/) {
```
davea, line 263: Would have been nicer to do this without the action check and with no template changes, but it would have meant /reports and RSS URLs would have changed also.

chrismytton (approved):

PR #5886, opened 2026-03-12
Ignore pin clicks during pinch-to-zoom

chrismytton:
The `marker_click` method was being triggered after a pinch-to-zoom if one of the fingers happened to interact with a marker.

I tested this on my iPhone while plugged into my laptop with Safari's debugger open. With a breakpoint set on the `marker_click` method you can see that when it gets triggered while pinching to zoom the event has a `.touches` array but with only one PointerEvent, but that PointerEvent _also_ has a `.touches` property and that one has two entries, corresponding to two fingers on the screen. So when we detect that condition we ignore it and return from `marker_click`.

Fixes FD-6625

### DIFF
--- a/web/js/map-OpenLayers.js
+++ b/web/js/map-OpenLayers.js
@@ -576,6 +576,14 @@ $.extend(fixmystreet.utils, {
     }
 
     function marker_click(feature, evt) {
+        // When pinching to zoom, OpenLayers fires clickFeature as the first finger lands
+        // on a pin while touches are still active. For a normal tap, the click fires on
+        // touchend when no touches remain, so evt.touches[0] is absent. Suppress
+        // navigation whenever any touch is still held down.
+        if (evt && evt.touches && evt.touches[0]) {
+            return;
+        }
+
         $(fixmystreet).trigger('maps:marker_click', feature);
 
         var problem_id = feature.attributes.id;
--- a/.cypress/cypress/integration/regressions.js
+++ b/.cypress/cypress/integration/regressions.js
@@ -96,4 +96,57 @@ describe('Regression tests', function() {
       cy.get('#form_start_date').should('have.value', '2019-01-01');
     });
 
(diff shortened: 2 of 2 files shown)


web/js/map-OpenLayers.js
```diff
@@ -576,6 +576,14 @@ $.extend(fixmystreet.utils, {
     }
 
     function marker_click(feature, evt) {
+        // When pinching to zoom, OpenLayers fires clickFeature as the first finger lands
+        // on a pin. The evt has a single PointerEvent in evt.touches[0], and on iOS Safari
+        // PointerEvent.touches holds the full live touch count — two touches indicates a
+        // pinch, so suppress navigation.
+        if (evt && evt.touches && evt.touches[0] && evt.touches[0].touches && evt.touches[0].touches.length > 1) {
```
dracos: Can we go simpler than this - if `evt.touches[0]` exists at all, that means a finger is still being held down somewhere on the screen, so we presumably don't want to do anything at that point?

For my reference, while checking this, I came up with the following to explain the current behaviour:

| 1st&nbsp;touch | 2nd&nbsp;touch | 1st off | 2nd off | Outcome |
| - | - | - | - | - |
| Pin | Not pin | Not pin | Pin | The first touchstart `registers` the pin in handle, so the first touchend triggers a clickout (which does nothing, we're not actually selecting/unselecting anything here) and the second touchend then does nothing |
| Not pin | Pin | Not pin | Pin | The second touchstart registers nothing (because it's now multi-touch), the second touchend registers as a click but it realises it's way too far away from the first touchstart |
| Pin | Not pin | Pin | Not pin | The first touchend triggers as a click |
| Not pin | Pin | Pin | Not pin | The first touchend triggers as a click |

chrismytton: Ah yep, that's a much nicer fix. Have pushed a fixup commit.

PR #5887, opened 2026-03-13
[Surrey] Defer report sending when GIS server is down

chrismytton:
When Surrey's GIS server is unavailable, defer sending the report rather than sending it without the necessary location data. Introduces a `DEFER:<reason>` return value for `open311_pre_send` hooks.

Ideally want to get this live ahead of this coming Wednesday 18th as there's some expected downtime then.

For FD-6749

### DIFF
--- a/perllib/FixMyStreet/Cobrand/UKCouncils.pm
+++ b/perllib/FixMyStreet/Cobrand/UKCouncils.pm
@@ -6,6 +6,7 @@ use Carp;
 use List::Util qw(min max);
 use URI::Escape;
 use LWP::Simple;
+use LWP::UserAgent;
 use URI;
 use Try::Tiny;
 use JSON::MaybeXS;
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/UKCouncils.pm
```diff
@@ -560,11 +561,17 @@ sub _fetch_features {
     }
 
     my $uri = $self->_fetch_features_url($cfg);
-    my $response = get($uri) or return;
+    my $ua = LWP::UserAgent->new;
```
davea: We should maybe set a short-ish timeout here (default is 180s) if we're expecting downtime to prevent the send daemon getting bogged down.

chrismytton: Good shout. Have pushed a fixup commit with a 30 second timeout, guess we could go shorter but I know some GIS systems can be slow to respond…

PR #5890, opened 2026-03-17
[Open311] Allow importing of specific service requests

davea:
Adds a new --report-ids option to bin/fetch that accepts a comma-separated list of service request IDs to import.

For https://github.com/mysociety/societyworks/issues/5271


bin/fetch
```diff
@@ -33,19 +33,30 @@ my ($opts, $usage) = describe_options(
...
+}
+
 my %params = (
     verbose => $opts->verbose,
     bodies => $opts->body || [],
     bodies_exclude => $opts->exclude || [],
 );
 
+if ($opts->report_ids) {
+    $params{report_ids} = [ split /,/, $opts->report_ids ];
```
chrismytton, line 57: Is it worth doing `split /\s*,\s*/` here, to handle strings with spaces in like `1, 2, 3, 4`?

PR #5896, opened 2026-03-20
Small admin tweaks and test fix

dracos:
Fixes https://github.com/mysociety/societyworks/issues/5370
Fixes https://github.com/mysociety/societyworks/issues/5101
Fixes flaky Cypress test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
         - Allow per-category timespan for closing inactive reports to updates. #5649
         - Dashboard: toggle visibility of deleted categories and categories with no reports
         - Allow search of non-integer external IDs.
+        - List categories that prevent reporting.
     - Development improvements
         - More logging when `page_error` is called, to aid troubleshooting. #5279
         - Docker changes needed for systemd to work. #5257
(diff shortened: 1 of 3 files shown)


templates/web/base/admin/reports/index.html
```diff
@@ -4,6 +4,7 @@
 <form method="get" action="[% c.uri_for_action('/admin/reports/index') %]" enctype="application/x-www-form-urlencoded" accept-charset="utf-8">
     <p><label for="search">[% loc('Search:') %]</label>
         <input class="form-control" type="text" name="search"  size="30" id="search" value="[% searched | html %]">
+        {% loc('e.g. word in report, email, <code>ref:[external id]</code>') %]</p>
```
chrismytton: Typo: `{%` should be a `[%`

perllib/FixMyStreet/App/Controller/Admin/Reports.pm
```diff
@@ -80,7 +80,7 @@ sub index : Path {
             $query->{'-or'} = [
                 'me.areas' => { like => "%,$1,%" }
             ];
-        } elsif ($search =~ /^ref:(\d+)$/) {
+        } elsif ($search =~ /^\s*ref:\s*(.*?)\s*$/) {
```
chrismytton: Minor thing, but I think this will now match `ref:` without an argument. Is it worth using `(.+?)` to require at least one character, like the numeric match did?

PR #5897, opened 2026-03-20
[Rutland] Recategorise report from Confirm to Salesforce

MorayMySoc:
Specific status codes from Confirm should resend a report to Salesforce.

This is currently by recategorising to one specified Salesforce category

https://github.com/mysociety/societyworks/issues/5420

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Rutland.pm
+++ b/perllib/FixMyStreet/Cobrand/Rutland.pm
@@ -88,6 +88,26 @@ around open311_extra_data_include => sub {
     return $data;
 };
 
+=head2 open311_get_update_munging
+
+This is used to perform a recategorisation of a report from Confirm to Salesforce
+when a specific update is received.
+
+=cut
+
+sub open311_get_update_munging {
+    my ($self, $comment, $state, $request) = @_;
+    my $problem = $comment->problem;
+
+    my @recategorise_ids = ('1600', '1610', '1611', '1612', '1620', '1621');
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Rutland.pm
```diff
@@ -88,6 +88,25 @@ around open311_extra_data_include => sub {
...
+This is used to perform a recategorisation of a report from Confirm to Salesforce
+when a specific update is received.
+
+=cut
+
+sub open311_get_update_munging {
+    my ($self, $comment, $state, $request) = @_;
+    my $problem = $comment->problem;
+
+    if ($state eq 'referred to salesforce') {
```
dracos: I assume we should be using an external status code here, rather than a state?

dracos (changes requested):
I assumed it would be an external status code we act on here, have I misunderstood.

t/cobrand/rutland.t
```diff
@@ -225,50 +225,66 @@ FixMyStreet::override_config {
...
+        is $report2->category, 'Drains', 'Category remains the same';
 
         FixMyStreet::Script::Reports::send();
         $report->discard_changes;
+        $report2->discard_changes;
 
         is $report->external_id, 248, 'External ID updated after resend';
+        is $report2->external_id, 34567, 'External ID remains with Confirm';
         my $req = Open311->test_req_used;
         my $c = CGI::Simple->new($req->content);
```
MorayMySoc: Sorry, these two lines aren't needed now I don't think so going to remove them

PR #5898, opened 2026-03-20
Add missing Confirm councils to client-side validation rules

chrismytton:
Had a stuck Aberdeenshire report that had two phone numbers on the user, but Confirm only wants one.

There is server-side validation for long phone numbers, but it's skipped for reports created by staff members on behalf of someone else, which was the case for the stuck report.

https://github.com/mysociety/fixmystreet/blob/10d3d6ad0a3e2cf85f8613a36057fa976252e8d3/perllib/FixMyStreet/App/Controller/Report/New.pm#L1477-L1481

Fix is to add in the missing client side validation. Have done this for the other councils that were obviously missing this.

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
+++ b/web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
@@ -64,13 +64,12 @@ body_validation_rules = {
           maxlength: 100
         },
     },
-    'North Northamptonshire Council': confirm_validation_rules,
-    'Northamptonshire Highways': {
+    'North Northamptonshire Council': $.extend({
         title: {
           required: true,
           maxlength: 120
         }
-    },
+    }, confirm_validation_rules),
     'Oxfordshire County Council': {
         detail: {
           required: true,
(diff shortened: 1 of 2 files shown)


web/cobrands/fixmystreet-uk-councils/council_validation_rules.js
```diff
@@ -62,6 +64,7 @@ body_validation_rules = {
           maxlength: 100
         },
     },
+    'North Northamptonshire Council': confirm_validation_rules,
```
dracos: North/West are Kier, not Confirm - so should they have the same as the old "Northamptonshire Highways"?

chrismytton: I used Confirm based on this comment in NorthNorthants.pm (the same comment appears in West as well):

https://github.com/mysociety/fixmystreet/blob/10d3d6ad0a3e2cf85f8613a36057fa976252e8d3/perllib/FixMyStreet/Cobrand/NorthNorthants.pm#L19-L24

But you're right, that doesn't handle the max title length restriction from the old setup. I should probably add a `sub max_title_length` like Peterborough has for the server side and then use `$.extend` in the client JS like Peterborough.

PR #5907, opened 2026-03-26
Govukpay integration

m7xedhoward:
This adds GOV.UK Pay as an alternative payment gateway alongside the existing SCP (Capita) and Adelante integrations. GOV.UK Pay is free for government and local authority use, making it attractive for councils that don't already have an SCP or Adelante contract.

 Yes, Pod and .md included.

No Public front end changes

Nothing is Cobrand-specific altho tests are based on cobrands

Test files created based upon existing tests.

The GOVUKPay role is only activated when a cobrand explicitly does with 'FixMyStreet::Roles::Cobrand::GOVUKPay' and provides govukpay_api_key in COBRAND_FEATURES.

Not applicable

No

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
@@ -1,0 +1,236 @@
+=head1 NAME
+
+FixMyStreet::Roles::Cobrand::GOVUKPay - GOV.UK Pay payment gateway role
+
+=head1 SYNOPSIS
+
+In your cobrand:
+
+    package FixMyStreet::Cobrand::YourCobrand;
+    use Moo;
+    with 'FixMyStreet::Roles::Cobrand::GOVUKPay';
+
+    sub waste_cc_payment_reference {
+        my ($self, $p) = @_;
+        return 'REF-' . $p->id;
+    }
+
+=head1 DESCRIPTION
+
+A Moo::Role providing GOV.UK Pay integration for FixMyStreet waste/garden
+subscriptions and similar payment flows.  Follows the same interface as
+L<FixMyStreet::Roles::Cobrand::SCP> and L<FixMyStreet::Roles::Cobrand::Adelante>.
+
+Configuration is read from C<< $cobrand->feature('payment_gateway') >>
+and should include:
+
+    COBRAND_FEATURES:
+      payment_gateway:
+        yourcobrand:
+          govukpay_api_key: 'your-api-key'
+          govukpay_api_url: 'https://publicapi.payments.service.gov.uk'
+          govukpay_description_prefix: 'FixMyStreet'  # optional
+
+=cut
+
+package FixMyStreet::Roles::Cobrand::GOVUKPay;
+
+use Moo::Role;
+use Try::Tiny;
+use Integrations::GOVUKPay;
+
+# Cobrands consuming this role must provide a method that returns a
+# payment reference string for a given Problem.
+requires 'waste_cc_payment_reference';
+
+=head2 waste_cc_has_redirect
+
+Returns 1 — GOV.UK Pay always redirects the payer to their hosted page.
+
+=cut
+
+sub waste_cc_has_redirect { 1 }
+
+=head2 _govukpay_config
+
+Returns the GOV.UK Pay subset of the payment_gateway config.
+
+=cut
+
+sub _govukpay_config {
+    my $self = shift;
+    my $cfg = $self->feature('payment_gateway') || {};
+    return {
+        api_key   => $cfg->{govukpay_api_key},
+        api_url   => $cfg->{govukpay_api_url} || 'https://publicapi.payments.service.gov.uk',
+        log_ident => $cfg->{log_ident} || 'govukpay',
+    };
+}
+
+=head2 _govukpay_client
+
+Returns a configured L<Integrations::GOVUKPay> instance.
+
+=cut
+
+sub _govukpay_client {
+    my $self = shift;
+    return Integrations::GOVUKPay->new({
+        config => $self->_govukpay_config,
+    });
+}
+
+=head2 waste_cc_get_redirect_url($c, $back)
+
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
```diff
@@ -0,0 +1,236 @@
...
+    $p->update_extra_metadata(redirect_id => $redirect_id);
+
+    my $reference = $self->waste_cc_payment_reference($p);
+    my $cfg       = $self->feature('payment_gateway') || {};
+    my $prefix    = $cfg->{govukpay_description_prefix} || 'FixMyStreet';
+    my $description = "$prefix: " . $p->title;
+
+    # Build return / back URLs
+    my $return_url;
+    if ($p->category eq 'Bulky collection') {
```
dracos: This $return_url is the same in both cases

perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
```diff
@@ -0,0 +1,236 @@
...
+=cut
+
+sub cc_check_payment_and_update {
+    my ($self, $reference, $p) = @_;
+
+    my ($error, $data) = $self->cc_check_payment_status($reference);
+    if ($data) {
+        $p->update_extra_metadata(
+            govukpay_status  => $data->{state}{status},
+            payment_reference => $data->{payment_id},
```
dracos: This level of code doesn't set payment_reference, calling things end up setting it with the passed-back value.

perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
```diff
@@ -0,0 +1,236 @@
...
+
+=cut
+
+sub cc_check_payment_and_update {
+    my ($self, $reference, $p) = @_;
+
+    my ($error, $data) = $self->cc_check_payment_status($reference);
+    if ($data) {
+        $p->update_extra_metadata(
+            govukpay_status  => $data->{state}{status},
```
dracos: This will always be 'success' so I don't think needs to be stored.

perllib/FixMyStreet/Script/Waste/CheckPayments.pm
```diff
@@ -23,7 +23,9 @@ sub check_payments {
     while (my $row = $problems->next) {
         $cobrand->set_lang_and_domain($row->lang, 1);
         my ($error, $reference);
-        if (my $scp = $row->get_extra_metadata('scpReference')) {
+        if (my $govukpay_id = $row->get_extra_metadata('govukpay_id')) {
```
dracos: There's also a loop like this in BulkyWaste.pm.
Though note we use `scpReference` for historic reasons, Adelante also stores its redirect reference in scpReference, for example, so I don't see an issue with also using scpReference in this code, and then this change here wouldn't be needed.

perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
```diff
@@ -0,0 +1,236 @@
...
+status from GOV.UK Pay, updates the report, and returns the payment
+reference on success or C<undef> on failure (with C<$c->stash->{error}>
+set).
+
+=cut
+
+sub waste_cc_check_payment_status {
+    my ($self, $c, $p) = @_;
+
+    my $govukpay_id = $p->get_extra_metadata('govukpay_id');
```
dracos: See note elsewhere that we could perhaps use scpReference here/elsewhere.

dracos (commented):
Hi, Thanks for looking at this! Is this for any use in particular, or just to add the general functionality?

Do you know how to use git rebase? For example, your "Add tests for Integrations::GOVUKPay" commit could be rebased/squashed with the commit that added the code, so that the code+tests for that code are in the same commit. And similar for other test-only commits. And then the two "fix" commits can presumably be squashed into the commits they are fixing. Bu that can be done at the end/by us, no biggie.

Could you explain the two fix commits a bit more?
"Replace HTTP::Request::Common::POST with explicit HTTP::Request->new to avoid header mangling when Content_Type/Content params are mixed with custom headers." - this shouldn't be needed, as far as I'm aware, what was going wrong?
And the other fix commit says "Add parentheses to JSON::MaybeXS::true() and JSON::MaybeXS::false() calls in both test files to avoid compilation errors." but the code isn't adding brackets, it's changing them to scalar references, which doesn't feel right; what compilation error were you getting before?

The docs are written like they're a commit message, rather than docs (e.g. "The `CheckPayments` cron script has been updated" doesn't make sense for someone reading this after the fact, as documentation).

I've made a few other small comments. I don't have access to a test GOV.UK Pay account to try it out, but in general this looks pretty good.Thanks again for your contribution

m7xedhoward:
Hi,

Thanks for the thoughtful review and comments.


This PR is just to add general functionality and as a learning opportunity of both git and the codebase. 

When I used HTTP::Request::Common::POST the header that was getting sent was garbled, I don't know why. 

The other issue was caused by originally having

Bareword "JSON::MaybeXS::true" not allowed while "strict subs" in use at t/app/controller/waste_govukpay.t line 152.

but changed to scalar references instead for simplicity, should have amended the commit message too

I've made changes to the md doc and I think I've done the rebase correctly.

perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
m7xedhoward: changed to scpReference and in the tests

perllib/FixMyStreet/Script/Waste/CheckPayments.pm
m7xedhoward: made the changes in govukpay.pm, No changes to this file are included now

perllib/FixMyStreet/Roles/Cobrand/GOVUKPay.pm
m7xedhoward: removed the conditional

PR #5918, opened 2026-04-09
[TfL] Change banner to banners in URL and copy

chrismytton:
I've left the licence name as "Banner licence" because "Banners licence" doesn't seem right and "Licence application for banners" is a bit wordy. Will see what the client says. Also means the category name can stay the same.

## Notes for deployer

- [x] Update payment links in config before deploying
- [x] Create new Banners licence category on staging and live sites

For https://3.basecamp.com/4020879/buckets/44968646/card_tables/cards/9741452332


MorayMySoc (approved):
Looks good to me!

PR #5919, opened 2026-04-10
[Bexley] Adds message to say cancellation not necessary

MorayMySoc:
Unless a resident actively renews their ggw subscription it will cancel.

Let them know so they can avoid unnecessary worry and contacting the council to ask.

https://mysocietysupport.freshdesk.com/a/tickets/6449

[skip changelog]


davea (approved):
👍

PR #5922, opened 2026-04-16
[Westminster] Remove cobrand

chrismytton:
For https://github.com/mysociety/societyworks/issues/5456


templates/web/westminster/report/new/_category_extra_field_notice.html
```diff
@@ -1,3 +0,0 @@
-[% IF meta.datatype_description AND NOT hide_descriptions %]
```
dracos, line 1: Do we want to remove the INCLUDE report/new/_category_extra_field_notice.html line from category_extras_fields as well, as this was the only use (and not entirely sure what it was doing!).

dracos (changes requested):
Looks good; only one suggestion to get rid of a hook that's not otherwise used

dracos (approved):

PR #5923, opened 2026-04-16
[Bexley] Ignore existing unconfirmed garden subs.

dracos:
Direct debit setup is done at our end, so we know that unconfirmed entries have not been completed and can be ignored when looking for an existing subscription.

I think this is enough, the ticket also mentions changing the renewal page check to match the bin day page, but I don't think that's necessary beyond that at this point.

Fixes https://github.com/mysociety/societyworks/issues/5518
For https://mysocietysupport.freshdesk.com/a/tickets/6827

[skip changelog]


chrismytton (approved):
Thanks!

PR #5924, opened 2026-04-16
Fix setting of options on multiple questions.

dracos:
The template wasn't being created correctly, so if you added multiple extra data questions with dropdown options, only one set of options was being set/overwritten.
Fixes https://github.com/mysociety/societyworks/issues/4829


MorayMySoc (approved):
Looks good

PR #5929, opened 2026-04-16
[Zurich] Store response template used

dracos:
Fixes mysociety/societyworks#5465
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Zurich.pm
+++ b/perllib/FixMyStreet/Cobrand/Zurich.pm
@@ -1279,6 +1279,10 @@ sub admin_stats {
 sub export_as_csv {
     my ($self, $c, $params) = @_;
 
+    my %templates = map { $_->id => $_->title } FixMyStreet::DB->resultset("ResponseTemplate")->search(
+        undef, { columns => ['id', 'title'] },
+    )->all;
+
     my $reporting = FixMyStreet::Reporting->new(
         objects_rs => $c->model('DB::Problem')->search_rs(
             $params,
@@ -1308,7 +1312,7 @@ sub export_as_csv {
             'External Body', 'Time Spent', 'Title', 'Detail',
             'Media URL', 'Interface Used', 'Council Response',
             'Strasse', 'Mast-Nr.', 'Haus-Nr.', 'Hydranten-Nr.',
-            'Interne meldung',
+            'Interne meldung', 'Template',
         ],
         csv_columns => [
             'id', 'created', 'whensent', 'lastupdate', 'local_coords_x',
@@ -1317,7 +1321,7 @@ sub export_as_csv {
             'body_name', 'sum_time_spent', 'title', 'detail',
             'media_url', 'service', 'public_response',
             'strasse', 'mast_nr',' haus_nr', 'hydranten_nr',
-            'interne_meldung',
+            'interne_meldung', 'template',
         ],
         csv_extra_data => sub {
             my $report = shift;
@@ -1345,6 +1349,9 @@ sub export_as_csv {
                 ? $c->cobrand->base_url . $report->photos->[$photo_to_display-1]->{url}
                 : '';
 
+            my $template = $report->get_extra_metadata('template_used') || '';
+            $template = $templates{$template} if $template;
+
             return {
                 whensent => $report->whensent,
                 lastupdate => $report->lastupdate,
@@ -1364,6 +1371,7 @@ sub export_as_csv {
                 haus_nr => $extras{'haus_nr'} || '',
                 hydranten_nr => $extras{'hydranten_nr'} || '',
                 interne_meldung => $report->non_public,
+                template => $template,
             };
         },
         filename => 'stats',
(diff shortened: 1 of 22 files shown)


templates/web/zurich/admin/templates/edit.html
```diff
@@ -27,15 +27,14 @@ <h3> [% IF rt.id %]
         <strong>[% loc('Text:') %] </strong>
         <textarea name="text" class="form-control required">[% rt.text |html %]</textarea>
     </p>
+    <p>
+        <label for="state">[% loc('Flag as deleted') %] </label>
+        <input type="checkbox" name="state" id="state" value="inactive">
```
davea: When editing an inactive template this box isn't ticked, making it easy to accidentally activate a template again.

perllib/FixMyStreet/Cobrand/Zurich.pm
```diff
@@ -690,6 +690,9 @@ sub admin_report_edit {
         # update the public update from DM
         if (my $update = $c->get_param('status_update')) {
             $problem->set_extra_metadata(public_response => $update);
+            if (my $template_id = $c->get_param('response_template_status_update')) {
+                $problem->set_extra_metadata(template_used => $template_id);
```
davea, line 694: Should we unset `template_used` if leaving a subsequent update without a template? Otherwise it would show that a template had been used when it hadn't necessarily.

templates/web/base/admin/response_templates_select.html
```diff
@@ -1,8 +1,11 @@
 [% IF problem.response_templates %]
-    <select id="templates_for_[% for %]" class="form-control js-template-name" data-for="[% for %]" name="response_template">
+    <select id="templates_for_[% for %]" class="form-control js-template-name" data-for="[% for %]" name="response_template_[% for %]">
         <option value="">[% loc('--Choose a template--') %]</option>
       [% FOR t IN problem.response_templates %]
-        <option value="[% t.text | html %]" data-problem-state="[% t.state | html %]"> [% t.title | html %] </option>
+        [% NEXT IF t.state == 'inactive' # Zurich special addition %]
```
davea: Worth including a Zürich cobrand check to prevent surprises in the future maybe?

web/cobrands/zurich/js.js
```diff
@@ -30,7 +30,7 @@ $(function() {
 
     $('.js-template-name').on('change', function() {
         var $this = $(this);
-        $('#' + $this.data('for')).val($this.val());
+        $('#' + $this.data('for')).val($this.data('text'));
```
davea: This doesn't work because `this` is the `<select>` element, not the `<option>` with the `data-text`.
```suggestion
        var $opt = $(this.selectedOptions[0]);
        $('#' + $this.data('for')).val($opt.data('text'));
```

web/cobrands/fixmystreet/staff.js
```diff
@@ -398,7 +398,7 @@ fixmystreet.staff_set_up = {
         var $this = $(this);
         var $input = $('#' + $this.data('for'));
         if (!$input.data('dirty')) {
-            $input.val($this.val());
+            $input.val($this.data('text'));
```
davea: Same here

davea (changes requested):

davea (approved):

dracos (commented):
This breaks the response templates in the inspector form, as that sets `templates_by_category` for updating the `<select>` when the category is changed in the form, and that isn't switched to using ID.

Going to try and make 'inactive' an option for everyone, not just Zurich.

dracos:
@davea I've hopefully fixed the issue with the inspector form in 15a8d27883 and then in 8895c58c97 I've generalised the inactive thing so any cobrand can do it, not just Zurich (and hopefully scattered the use in all the relevant places), if you could take a look, thanks.

templates/web/base/admin/templates/edit.html
```diff
@@ -80,12 +80,24 @@
       </label>
     </p>
 
+    <div class="admin-hint">
+      <p>
+        [% loc('If ticked, this template will not be used.') %]
```
davea: This wording is a little unclear - will not be used by who/what/where? Maybe something like
```suggestion
        [% loc('If ticked, this template will not be used for incoming Open311 updates, nor available for selection by staff when adding updates manually.') %]
```

davea (approved):
Works well, just one small suggestion for a wording change is all

templates/web/base/admin/response_templates_select.html
```diff
@@ -1,8 +1,11 @@
 [% IF problem.response_templates %]
-    <select id="templates_for_[% for %]" class="form-control js-template-name" data-for="[% for %]" name="response_template">
+    <select id="templates_for_[% for %]" class="form-control js-template-name" data-for="[% for %]" name="response_template_[% for %]">
         <option value="">[% loc('--Choose a template--') %]</option>
       [% FOR t IN problem.response_templates %]
-        <option value="[% t.text | html %]" data-problem-state="[% t.state | html %]"> [% t.title | html %] </option>
+        [% NEXT IF t.deleted %]
```
dracos: This line isn't needed, now that problem.response_templates queries for deleted=0

PR #5931, opened 2026-04-17
[Sutton] Premium food containers

MorayMySoc:
Two commits - one for text change https://mysocietysupport.freshdesk.com/a/tickets/6902

Other, replace svg with png - caddy not showing ie number 77 SM6 0DE

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Kingston.pm
+++ b/perllib/FixMyStreet/Cobrand/Kingston.pm
@@ -109,6 +109,8 @@ sub waste_check_staff_payment_permissions {
 
 =item * Garden subscriptions are 5 bins maximum
 
+=item * 4 days before a new garden subscription starts if already have a container
+
 =back
 
 =cut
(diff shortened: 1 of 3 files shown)


MorayMySoc:
(Sorry - forgot to add the png - should be there now)

dracos (changes requested):
The other SVGs for that property are appearing fine, as far as I can see, so I don't think the issue is necessarily because it's a SVG rather than a PNG.

Also this won't fix the error because the 2x image still isn't present.
Looks like (looking at the bin days template) you'd need to either supply:
* `{ type => 'png1', src => '$base/caddy-brown-premium', alt => 'Brown premium food caddy' }` if you only have the one PNG and you want PNG;
* `svg_container_bin('Brown premium food caddy', 'caddy-brown-premium', '#8B5E3D')` to keep the SVG and have that work (though it's not actually colour changing, so 3rd argument not doing anything here.

MorayMySoc:
Accidentality merged the first fixup, but gone with svg and added a title in a fixup which seems necessary for alt text

dracos (approved):
Aha, yep, that should do it, thanks :)

PR #5932, opened 2026-04-20
[Rutland] hard code feed population estimate to max population

struan:
The population of Rutland is less than the 200k we use for the gaze call so hard code the rough county population in to the template

[skip changelog]

Screenshot 2026-04-20 at 11 39 02

Fixes mysociety/societyworks#5524

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Alert.pm
+++ b/perllib/FixMyStreet/App/Controller/Alert.pm
@@ -479,7 +479,10 @@ sub determine_location : Private {
         $c->go('index');
     }
 
-    my $dist = FixMyStreet::Gaze::get_radius_containing_population($c->stash->{latitude}, $c->stash->{longitude});
+    my $population = $c->cobrand->feature('alerts_population');
+
+    my $dist = FixMyStreet::Gaze::get_radius_containing_population($c->stash->{latitude}, $c->stash->{longitude}, $population);
+    $c->stash->{population_count} = $c->cobrand->feature('alerts_population');
     $c->stash->{population_radius} = $dist;
 
     return 1;
--- a/perllib/FixMyStreet/App/Controller/Alert.pm
+++ b/perllib/FixMyStreet/App/Controller/Alert.pm
@@ -262,7 +262,7 @@ sub set_local_alert_options : Private {
     {
         $type = 'local_problems';
         push @params, $2, $1; # Note alert parameters are lon,lat
-        my $distance = $c->forward('get_distance');
+        my $distance = $c->forward('get_distance', [$1, $2]);
         push @params, $distance if $distance;
     }
     return unless $type;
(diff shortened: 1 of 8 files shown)


dracos (changes requested):
We'd need to actually change the number that it then looks up? :)
200_000 is in Gaze.pm

Could we make it a cobrand variable, so we can change it for others in future; not sure how easy that is.

dracos:
Also made more complex because it's used by alert sending if a distance wasn't recorded, not sure if we'll need to deal with that if the default is now changing

perllib/FixMyStreet/App/Controller/Alert.pm
```diff
@@ -479,7 +479,10 @@ sub determine_location : Private {
         $c->go('index');
     }
 
-    my $dist = FixMyStreet::Gaze::get_radius_containing_population($c->stash->{latitude}, $c->stash->{longitude});
+    my $population = $c->cobrand->feature('alerts_population') || 200000;
```
dracos: Could we not have the `||` so then it defaults inside the Gaze function, easier to change only in one place (plus then population_count line can just set it to $population)

templates/web/base/alert/_list.html
```diff
@@ -37,7 +38,12 @@ <h3>[% loc('Problems in an area') %]</h3>
           [% ELSE %]
             [% tprintf( loc('Problems within %.1fkm of %s'), population_radius, name_of_location ) %]
           [% END %]
+          [% IF population_count;
+            pop_count = population_count | format_number;
+            tprintf('covers roughly %s people', pop_count ) %]
+          [% ELSE %]
             [% loc('(covers roughly 200,000 people)') %]
```
dracos, line 45: Do we need the ELSE any more, if the default is going to be this?

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -407,10 +407,12 @@ sub _alert_iterator {
             next unless $cobrand->email_host;
             next if $alert->is_from_abuser;
 
+            my $population = $cobrand->feature('alerts_population') || 200000;
```
dracos: Ditto here with the default, I guess.

This still has the issue I mentioned - this changes any existing alert (that is lacking a distance parameter) to now have a different default from what they signed up for. I can't think of a good solution apart from to update every alert to have the current Gaze distance instead of null, and always store it, in case this default changes again in future? Either that or we have to record dates for `alerts_population` and use the right one dependent on alert creation date.

dracos (changes requested):

templates/web/base/alert/_list.html
struan: my reasoning here was this is only going to apply on cobrands and this means we keep the translated string

perllib/FixMyStreet/Script/Alerts.pm
struan: I think I've resolved this by always setting a distance now and not using the cobrand override in alerts script because then the old ones will always use the 200k number if there isn't a distance and new ones will have a distance which I think solves both sides?

perllib/FixMyStreet/App/Controller/Alert.pm
```diff
@@ -262,7 +262,10 @@ sub set_local_alert_options : Private {
     {
         $type = 'local_problems';
         push @params, $2, $1; # Note alert parameters are lon,lat
-        my $distance = $c->forward('get_distance');
+        # need these for get_distance population fallback
+        $c->stash->{latitude} = $1;
+        $c->stash->{longitude} = $2;
+        my $distance = $c->forward('get_distance', [1]);
```
dracos: Good solution - but I think we want this in the RSS feed as well presumably (the other call to get_distance in this file), so that doesn't potentially change in future either?
So might as well not have the flag to get_distance and have it always look up the radius if needed (and you could pass in lat/lon as params rather than stash maybe?).

dracos (changes requested):
Only the one more step for RSS feeds I think

dracos (approved):
Thanks!
Guess there'll be a slight up-tick in Gaze calls on the front end alert pages (though guess might be cached hopefully), but should be fine

PR #5934, opened 2026-04-20
[Bexley][WW] Use external_id instead of UPRN when matching Agile cancellations

chrismytton:
When checking Agile cancellations use the `external_id` instead of UPRN when matching to WW. Moves the legacy contract check/cancel step so that it always runs regardless of whether a WW report was found or not.

Changes the cancellation reports to have an `agile_reference` in their `extra`, which does mean that when it next runs it will pick up cancellations that it might have already processed, but I think it will just try cancelling and fail, so shouldn't be an issue, will just create a few duplicate cancellation reports in WW (that don't get sent anywhere).

I've also updated the output to use `print` more liberally, so that errors are properly surfaced in the cron email.

Note: this doesn't handle tracking cancel failures, I'm doing that separately as part of https://github.com/mysociety/societyworks/issues/5515

For FD-6862

### DIFF
--- a/bin/bexley/cancel-garden-waste-from-agile
+++ b/bin/bexley/cancel-garden-waste-from-agile
@@ -18,6 +18,7 @@ my ($opts, $usage) = describe_options(
     '%c %o',
     ['days=i', 'How many days to check for cancellations (max 7)', { default => 7 } ],
     ['uprn=s', 'A specific UPRN to cancel'],
+    ['reference=s', 'Agile reference to cancel'],
     ['verbose|v', 'Print verbose output'],
     ['help|h', "print usage message and exit" ],
 );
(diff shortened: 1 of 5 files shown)


dracos (commented):
Think this looks fine - the only question I could think of was would it be possible for an Agile contract reference to get 'reenabled' in future - so the $report would still find the most recent, okay, but $cancellation_report would then find something? Or will that not happen. I guess as well as storing agile_reference, could it store the $report ID it's linked to, so we know which report caused which cancellation? Then actually it wouldn't need to store agile_reference then, because you could work that out from the linked report should you need to.

chrismytton:
@dracos Ahh yep, storing the original report ID on the cancellation is a much nicer approach, means there's a proper link rather than just an inferred one. Thanks, will update.

chrismytton:
Have added two fixup commits:

- Switch from agile_reference to original_report_id in extra on the cancellation report
- Include DD customer ID in the cancellation report's extra, as this is useful for querying a customer's active contracts

dracos (approved):
Woo, thanks.
Given your message about initial run, " I think it will just try cancelling and fail" - do we want to have a manual run without it doing anything and mark the reports somehow otherwise, to prevent this happening? I'd rather not create a bunch of cancellation reports that then have to be manually closed/explained to Bexley if they notice/etc.
Perhaps we can manually run LastCancelled and check? Or perhaps it's not worth it.
Just worried about the potential outcome, I guess.

chrismytton:
@dracos Yeah, good point. I might be able to update the existing cancellation reports with the correct `original_report_id` directly in the database, so that it picks those up when it runs rather than creating new ones.

I'll look at doing that before running it manually, will dry-run locally first.

PR #5935, opened 2026-04-20
[Central Bedfordshire] Updated `#site-logo`

lucascumsille:
- Removed "FixMyStreet" text from non-map pages
- On map pages replaces FixMyStreet" text with council logo

Preview here:
Screenshot 2026-04-20 at 15 50 24

Screenshot 2026-04-20 at 15 50 43

Screenshot 2026-04-20 at 15 50 04

[skip changelog]


templates/web/centralbedfordshire/header_logo.html
```diff
@@ -0,0 +1,2 @@
+<a href="https://www.centralbedfordshire.gov.uk/" id="site-logo" title="Home">Central Bedfordshire FixMyStreet</a>
```
dracos: Should be "Council", not "FixMyStreet" :)

web/cobrands/centralbedfordshire/base.scss
```diff
@@ -63,14 +68,21 @@ a {
     font-weight: bold;
 }
 
+#site-header {
```
dracos, line 71: I don't think this block is needed? The colour is already fff inherited, and the min-height doesn't do anything, because the logo makes it that high anyway?

dracos (changes requested):
Couple of small changes is all I think :)

web/cobrands/centralbedfordshire/base.scss
lucascumsille: Unfortunately we need the `min-height: 4rem` because the `#site-logo` has `display:none` so this happens without defining a height:

https://github.com/user-attachments/assets/5c2f2ee6-0fcc-4e8d-9fcc-b53c604ac6fc

I added a fixup for the background color though

templates/web/centralbedfordshire/header_logo.html
lucascumsille: I just added a fixup up for that as well =)

dracos (approved):

PR #5937, opened 2026-04-22
[Bexley] Ask for user details if nothing in Agile.

dracos:
If they pass customer ref verification, but have no email/phone or missing first/last name, ask the user for those details so that we have them for the report to be created. Otherwise a staff user can create an anonymous entry with their details, or a user can get an error about missing details.
[skip changelog]
Hopefully for https://mysocietysupport.freshdesk.com/a/tickets/6922 and https://mysocietysupport.freshdesk.com/a/tickets/6924


chrismytton (approved):
Looks good, thanks!

MorayMySoc (approved):

PR #5941, opened 2026-04-24
Include filter category/group in new report links

dracos:
Also consolidate all the banner templates, they seemed similar enough.

For FD-6925.


davea (approved):
Nice fix and tidy up. 👍

PR #5944, opened 2026-04-27
[Bexley][WW] Fixes for Agile cancellation script

chrismytton:
Now matches by Agile ID or Reference to pick up both new and renew reports. Removes the cancellation report tracking, that's not actually needed here because those reports aren't being sent anywhere, so we can just cancel the direct debit directly. Also adds some extra tracking for failed reports so we can see the error, how many times they failed, and when they last failed.

### DIFF
--- a/perllib/FixMyStreet/Script/Bexley/CancelGardenWaste.pm
+++ b/perllib/FixMyStreet/Script/Bexley/CancelGardenWaste.pm
@@ -49,8 +49,8 @@ sub cancel_contract {
     my ( $self, $contract ) = @_;
 
     my $uprn = $contract->{UPRN};
+    my $id = $contract->{Id};
     my $reference = $contract->{Reference};
-    my $external_id = "Agile-$reference";
     my $reason = $contract->{Reason};
 
     $self->_vprint("Attempting to cancel contract $reference (UPRN: $uprn)");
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Script/Bexley/CancelGardenWaste.pm
```diff
@@ -192,8 +108,21 @@ sub _cancel_direct_debit {
     my $resp = $i->cancel_plan({ report => $original_report });
 
     if ( ref $resp eq 'HASH' && $resp->{error} ) {
+        my $failures = $original_report->get_extra_metadata('direct_debit_cancellation_failures') // 0;
+        $failures++;
+        $original_report->set_extra_metadata(
+            direct_debit_cancellation_failures       => $failures,
+            direct_debit_cancellation_error          => $resp->{error},
+            direct_debit_cancellation_last_failed_at => FixMyStreet->set_time_zone(DateTime->now)->iso8601,
```
dracos: It looks like in other Bexley code we use `DateTime->now->set_time_zone( FixMyStreet->local_time_zone ), but can keep this if you prefer

dracos (approved):
Nice, glad we worked that out :) One comment is all

PR #5945, opened 2026-04-27
Ability to have MapIt-based paragraphs in notices.

dracos:
`<p data-area-ids="ID">...</p>` will only be displayed if the point is covered by the area with that MapIt ID.

`<p data-area-invert="ID">...</p>` will not display if the point is covered by the area with that MapIt ID.

[skip changelog] For FD-6620
Fixes https://github.com/mysociety/societyworks/issues/5430


t/cobrand/gloucester.t
```diff
@@ -51,6 +51,20 @@ my $flytipping = $mech->create_contact_ok(
     email    => 'Regular_fly-tipping_(not_witnessed_and_no_evidence_likely)',
 );
 
+my $grass = $mech->create_contact_ok(
+    body_id => $body->id,
+    category => 'Grass',
+    email => 'grass@example.org',
+    extra => { _fields => [
+        { code => 'notice', variable => 'false', description => '<p>Paragraph for everyone.</p>
```
nephila-nacrea, line 59: Would this mean we add messaging on behalf of the client, or they learn what area IDs they need to provide?

.cypress/cypress/integration/gloucester.js
```diff
@@ -73,6 +73,7 @@ describe('Gloucester cobrand', function(){
             cy.route('**gloucester.assets*adopted_streets*', 'fixture:gloucester_empty.json').as('gloucester_streets');
 
             cy.visit('http://gloucester.localhost:3001/report/new?longitude=-2.2458&latitude=51.86506');
+            cy.pickCategory('Broken glass');
```
nephila-nacrea, line 76: Why has this been added - was the test faulty?

perllib/FixMyStreet/App/Controller/Report/New.pm
```diff
@@ -395,6 +395,15 @@ sub disable_form_message : Private {
 
     my %out;
 
+    # Ward based message changing
+    my $areas = '(?:\d+,)*(?:' . (join '|',
+        sort keys %{$c->stash->{all_areas_mapit}}) . ')[,"]';
+    foreach (@{$c->stash->{category_extras}->{$c->stash->{category}}}) {
+        next unless ($_->{variable} || '') eq 'false';
+        $_->{description} =~ s/<p[^>]*data-area-ids="(?!$areas)\d+(?:,\d+)*"[^>]*>.*?<\/p>//gs;
```
nephila-nacrea: Is there a module or anything that can read from HTML tags for us?

nephila-nacrea (commented):
A few questions, though I have reproduced the test cases locally so that all seems OK 👍 The regex is really complicated 😵‍💫

t/cobrand/gloucester.t
dracos: Yeah, I assumed at least to begin with we'd probably do it for them, though depends how capable they are I guess :)

.cypress/cypress/integration/gloucester.js
dracos: Before, there was only one category, so it was being auto-selected and thus this wasn't needed, but as I've added other categories now, it is. Took me a while to work out this was the reason :)

perllib/FixMyStreet/App/Controller/Report/New.pm
dracos: Yeah, we already have e.g. HTML::Tree/HTML::TreeBuilder but it seemed a bit overkill to parse it for this :)
Perhaps I should have written the regex with `x`:
```
/
<p                  # Start of a p tag
[^>]*               # Anything at the start of the p tag
data-area-ids="     # Start of this attribute
(?!$areas)          # Zero-width negative lookahead - what follows after this must NOT match $areas
\d+(?:,\d+)*        # One or more integers separated by commas
"                   # End quote mark of attribute
[^>]*               # Anything up to the end of the p tag
>                   # End of p tag
.*?                 # Anything inside the p
<\/p>               # Closing p tag
/
```
So that's basically saying match any paragraph with a data-area-ids attribute of comma-separated integers that **doesn't** match $areas.

And then `$areas` is:
```
(?:\d+,)*           # Zero or more integers followed by a comma
(?:123|456|789)     # Any of the IDs at the point clicked
[,"]                # A comma or the end quote mark
```
So `$areas` matches an area ID we care about somewhere inside this comma separated list.
So the final regex will remove any line that has a data-area-ids p that doesn't match an ID we care about.

nephila-nacrea: I probably would have tried it with HTML::Tree* stuff (have mucked around a bit and have an idea of how it might work), because big regexes like this are hard to parse! Would you be happy to rewrite it as the `x` stuff above at least?
Granted, it's a fairly small part of the code / not particularly dangerous (I don't think?) so would be OK to approve anyway.

dracos: Yep, I'll comment it all. Here's the other reason why I don't want to use HTML::TreeBuilder (given this will run quite a few times, in quite a common bit of code):

```
               Rate HTML::Tree      regex
HTML::Tree   3864/s         --       -99%
regex      392991/s     10069%         --
```

nephila-nacrea: 😭 oh that's awful speeds

PR #5952, opened 2026-04-28
[Bug] /around page state dropdown

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/5470 ; spotted by Dumfries but was a wider issue, where body and filter_states were not defined for /around page, but only for /reports page.

[skip changelog]

### DIFF
--- a/templates/web/base/reports/_list-filter-status.html
+++ b/templates/web/base/reports/_list-filter-status.html
@@ -1,3 +1,8 @@
+[% # 'body' and 'filter_states' already set for /reports but not for /around %]
+[%
+  DEFAULT body = c.cobrand.body;
+  DEFAULT filter_states = c.cobrand.state_groups_inspect;
+%]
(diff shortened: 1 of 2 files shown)


templates/web/base/reports/_list-filter-status.html
```diff
@@ -1,3 +1,8 @@
+[% # 'body' and 'filter_states' already set for /reports but not for /around %]
+[%
+  SET body = body || c.cobrand.body;
+  SET filter_states = filter_states || c.cobrand.state_groups_inspect;
```
davea: You could use `DEFAULT` ([docs](https://template-toolkit.org/docs/manual/Directives.html#section_DEFAULT)) here which is a little cleaner:

```suggestion
  DEFAULT body = c.cobrand.body;
  DEFAULT filter_states = c.cobrand.state_groups_inspect;
```

davea (approved):
One small suggestion, feel free to ignore! (Also I think the base branch is wrong/doesn't need to branch from there, so I only reviewed the second commit in this PR)

templates/web/base/reports/_list-filter-status.html
nephila-nacrea: Huh, today I learned!

PR #5953, opened 2026-04-28
[Westminster] Add script to anonymise users

chrismytton:
This roughly follows the anonymisation patterns from `perllib/FixMyStreet/Script/Inactive.pm`. 

It anonymises any users who has only made a report on the Westminster cobrand, and it anonymises reports along with their comments and alerts.

One thing to check, the problem step is based on the `bodies_str`, which will match across all cobrands, is that right? This means that even if a person made a report to Westminster on .com before the Westminster cobrand existed that report will disappear from their account?

Fixes https://github.com/mysociety/societyworks/issues/5456

### DIFF
--- a/bin/one-off-anonymize-westminster
+++ b/bin/one-off-anonymize-westminster
@@ -1,0 +1,143 @@
+#!/usr/bin/env perl
+#
+# Anonymises personal data left over from the Westminster cobrand
+# off-boarding.
+#
+# Users who only ever reported on the Westminster cobrand have their
+# accounts anonymised in place. Reports sent to Westminster council
+# (whether made on the Westminster cobrand or routed via .com /
+# neighbouring cobrands) are reassigned to the shared anonymous user,
+# along with their comments and alerts.
+
+use strict;
+use warnings;
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../setenv.pl";
+}
+
+use Getopt::Long;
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
Chat about alerts in Slack.
So this anonymizes any user who only used the westminster cobrand for reports (and we don' need to worry about updates as was staff only), that looks okay.
But for reports - this anonymizes too much, I think? We don't need to anonymize reports made on .com at all, they can stay, we only need to anonymize reports made on the Westminster cobrand.

bin/one-off-anonymize-westminster
```diff
@@ -0,0 +1,142 @@
...
+        next;
+    }
+
+    $users_anonymised++;
+    say "Anonymising account for user #" . $user->id if $verbose;
+    $user->anonymize_account unless $dry_run;
+}
+
+my $problems = FixMyStreet::DB->resultset("Problem")->search({
+    cobrand => 'westminster',
```
dracos: Presumably should be $cobrand to match the rest.

chrismytton:
@dracos I've pushed two fixups, one to only search for cobrand=westminster problems, and the other to disable the Westminster alerts.

dracos (approved):
One hardcoded westminster is all, given you'd done all the others as a variable :)
What's the dry-run output?

chrismytton:
Dry-run output:

```
Westminster-only users anonymised: 1185
Multi-cobrand users kept:          312
Westminster reports anonymised:    43760
Westminster alerts disabled:       399
```

That matches what I'd expect, querying the database manually. Going to go ahead and run it on live.

PR #5954, opened 2026-04-29
[Southwark] No public photos in certain categories

dracos:
Previously allow_photo_display has only been used by Zurich, which has other special templates etc, so there were places it wasn't being checked when it should be. Hopefully add all/most of those (email templates, list views the main two), and then add a function to Southwark to hide photos to the public on certain categories (both by that mechanism and a special recent_photos for alert page).

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Southwark.pm
+++ b/perllib/FixMyStreet/Cobrand/Southwark.pm
@@ -17,6 +17,7 @@ use strict;
 use warnings;
 
 use List::Util qw(any);
+use Scalar::Util qw(blessed);
 use Moo;
 
 =head2 Defaults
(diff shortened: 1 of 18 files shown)


perllib/FixMyStreet/Cobrand/Southwark.pm
```diff
@@ -83,6 +84,49 @@ sub pin_colour {
 
 sub path_to_pin_icons { '/i/pins/whole-shadow-cone-spot/' }
 
+sub allow_photo_display {
+    my ($self, $object_or_hash, $idx) = @_;
+
+    #return 1; # Make all the tests fail first
```
chrismytton: Think this can be deleted?

perllib/FixMyStreet/Cobrand/Southwark.pm
```diff
@@ -83,6 +84,49 @@ sub pin_colour {
...
+
+    if ($self->{c} && $self->{c}->user_exists) {
+        my $user = $self->{c}->user;
+        return 1 if $user->from_body || $user->is_superuser;
```
chrismytton: Not sure it actually matters in practice, but does this mean that theoretically non-Southwark staff members could see hidden photos?

perllib/FixMyStreet/Cobrand/Southwark.pm
```diff
@@ -83,6 +84,49 @@ sub pin_colour {
 
 sub path_to_pin_icons { '/i/pins/whole-shadow-cone-spot/' }
 
+sub allow_photo_display {
```
chrismytton, line 87: I think this will only apply to the Southwark cobrand, right? If that's the case does it matter if these photos appear on the .com cobrand, or do we need something there as well for Southwark?

chrismytton (changes requested):

perllib/FixMyStreet/Cobrand/Southwark.pm
dracos: Yeah, I think Southwark cobrand only is okay, is all they've asked about, and it would seem to get much more complex to try and do "Southwark reports on .com" (where we don't say we'd remove any photos like this anyway).

perllib/FixMyStreet/Cobrand/Southwark.pm
dracos: Yep!

perllib/FixMyStreet/Cobrand/Southwark.pm
dracos: Hmm, yes, I guess so - guess I could change it to belongs_to_body($self->body) to make it only southwark staff.

chrismytton (approved):
Looks good!

PR #5956, opened 2026-04-30
[Dumfries] Only show password button for users with FMS password

MorayMySoc:
If a user already logged in by OIDC don't show FMS password set/change button

https://github.com/mysociety/societyworks/issues/5498

[skip changelog]

### DIFF
--- a/templates/web/base/my/my.html
+++ b/templates/web/base/my/my.html
@@ -105,11 +105,10 @@
 
 </ul>
 </form>
-
 <p class="my-account-buttons">
   [% IF c.session.oauth.change_password_uri %]
     <a class="btn change-password has-inline-svg" href="[% c.session.oauth.change_password_uri | html %]">[% INCLUDE 'icons/password.html' width='1em' height='1em' %][% loc('Change password') %]</a>
-  [% ELSE %]
+  [% ELSIF !(c.cobrand.moniker == 'dumfries' AND c.session.oauth.logout_redirect_uri) %]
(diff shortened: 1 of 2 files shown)


MorayMySoc:
Not sure whether it will be clear which password you're changing still

templates/web/base/my/my.html
```diff
@@ -109,7 +109,7 @@ <h1>[% loc('Your account') %]</h1>
 <p class="my-account-buttons">
   [% IF c.session.oauth.change_password_uri %]
     <a class="btn change-password has-inline-svg" href="[% c.session.oauth.change_password_uri | html %]">[% INCLUDE 'icons/password.html' width='1em' height='1em' %][% loc('Change password') %]</a>
-  [% ELSE %]
+  [% ELSIF !(c.cobrand.moniker == 'dumfries' AND !c.user.password)  %]
```
davea: This also hides the 'Set password' button from users who have logged in via email. Probably a small number of people, but still worth maintaining existing the functionality.

davea (changes requested):
It does what it says on the tin, but also breaks some existing functionality unfortunately.

davea (changes requested):
Still something not quite right with the logic, the 'Set password' button still not visible.

Steps to reproduce:

 - open an incognito/private window
 - Sign in to the Dumfries cobrand with a brand new email address, with the 'email me a link to sign in' option
 - Click link in email
 - View account page, 'Set password' button is missing

Image

davea (approved):

PR #5957, opened 2026-04-30
[Northumberland] add DVLA lookup for abandoned vehicles

struan:
Adds DVLA lookup to Northumberland.

As the only have a single text field for the relevant categories this adds in the option to put a single summary with all the DVLA information in a single field.

[skip changelog]

Fixes mysociety/societyworks#5497

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/dvla.js
+++ b/web/cobrands/fixmystreet-uk-councils/dvla.js
@@ -171,6 +171,10 @@ function dvla_lookup(e) {
                     if (field) {
                         field.value = data[name];
                     }
+                    field = document.querySelector('textarea[name*="' + fields[name] + '"]');
+                    if (field) {
+                        field.value = data[name];
+                    }
                 }
             });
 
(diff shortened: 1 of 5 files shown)


davea (approved):

web/cobrands/fixmystreet-uk-councils/dvla.js
```diff
@@ -85,7 +93,7 @@ function dvla_lookup(e) {
     const yesno = document.querySelector('input[name=dvla_reg_have]:checked');
 
     if (!yesno) return;
-    if (!yesno.value) {
+    if (!yesno.value && !fields.summary) {
```
dracos: Do we want a positive fields.reg check here?

struan: something like this?

```
if (!yesno.value && fields.reg) {
```

dracos: Yeah - given that's what then's being set inside.

PR #5965, opened 2026-05-06
[Bexley][WW] Track DD cancellation, amend and one-off failures

chrismytton:
Failures are recorded under a single `direct_debit_errors` metadata hash, cleared when a subsequent operation succeeds. Amend and one-off failures also stash the args needed to retry. Adds `bin/bexley/check-dd-errors` to list outstanding failures with a per-op retry recipe.

Fixes https://github.com/mysociety/societyworks/issues/5515

## Deploy notes

- [x] Add `bin/bexley/check-dd-errors` to crontab so it runs regularly

### DIFF
--- a/bin/bexley/check-dd-errors
+++ b/bin/bexley/check-dd-errors
@@ -1,0 +1,95 @@
+#!/usr/bin/env perl
+
+use v5.14;
+use warnings;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use DateTime;
+use Getopt::Long::Descriptive;
+use FixMyStreet;
+use FixMyStreet::Cobrand;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['days=i', 'only show failures with last_failed_at within this many days', { default => 10 }],
+    ['all',   'show all failures regardless of age'],
+    ['help|h', 'print usage message and exit'],
+);
+$usage->die if $opts->help;
+
+my $cutoff;
+unless ($opts->all) {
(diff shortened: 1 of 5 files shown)


dracos:
GitHub PR comments are currently down :-/

1. I'm not sure what `die` will do where it's still present - nothing is catching it, so I assume the user will get an internal error or something like that. And then perhaps try again (which could make the problem worse; not sure?). Should we just return the same as if it's succeeded, assuming it's now going to be handled elsewhere manually. But then why has no-one got in touch about an error when this has been tried, if that's what's happened? :-/
2. Similarly/related, if an ad-hoc payment fails when increasing a subscription, presumably we do still want to go on and try and amend the plan?
3. If there was a blip calling the DD backend, then the script looks okay, assuming it worked the next time. But if the issue is with the lack of a direct_debit_contract_id (which seems more likely?), then that's not resolvable and this script will keep sending the same thing each time. Should we set a flag in the script to not alert on an issue more than once (or to be same as Bromley DD, perhaps with last-failure in past 10 days or something like that, so we do get it more than once but not for ever...).

dracos (changes requested):
^

chrismytton:
@dracos Thanks, all good points!

I've changed `amend_plan` and `one_off_payment` to return undef on failure or 1 on success instead of dying. fe742382312aa424a9c9310e06820f967b1c0d64.

For the check script I've gone with the Bromley approach of only showing things from the past 10 days, with flags to show different number of days or show all errors regardless of age. 9cda976f3e02491ebf01095706e522c254453a9c

dracos (approved):
Thanks :) Guess the only issue I can see is the `retry` if there's a missing contract ID (which I'm assuming/hoping will be the main/only reason it happens?) won't be able to do anything, but can deal with that if/when it arises.

PR #5966, opened 2026-05-06
No user password check

dracos:
And upgrade a couple of modules.
CC @sagepe for interest


neprune (approved):
LGTM. I guess the other option would have been to set a time floor that is always slower than a genuine lookup, and have both branches normalised to that time, but probably would have to have been set much slower to ensure the 'user exist' branch was never took longer.

PR #5967, opened 2026-05-07
WEBP support, error handling, ward messaging bugfix

dracos:
After Camden had some problems uploading a photo, it turned out the photo was WEBP but named .jpg (so was uploaded as image/jpeg). Also the error they got was yucky, trying to display the HTML source of the SW error page. So - change it to 400 error (makes more sense anyway, should stop HTML error source display); add WEBP as allowed (might as well, seems to be used a bit more by phones and things); look at file header over MIME type for what the file actually is.

And also allow admin code in the Open311 descriptions so that my data-area-ids thing for Gloucester isn't immediately overwritten by the syncing script.

### DIFF
--- a/perllib/FixMyStreet/App/Model/PhotoSet.pm
+++ b/perllib/FixMyStreet/App/Model/PhotoSet.pm
@@ -155,18 +155,16 @@ has ids => ( #  Arrayref of $fileid tuples (always, so post upload/raw data proc
                 # get the photo into a variable
                 my $photo_blob = eval {
                     my $filename = $upload->tempname;
-                    my $out;
-                    if ($type eq 'jpeg' && can_run('jhead')) {
+                    my ($w, $h, $err) = Image::Size::imgsize($filename);
+                    die _("Please upload an image only") . "\n" if !defined $w || $err !~ /JPG|GIF|PNG|TIF|WEBP/;
+                    if ($err eq 'JPG' && can_run('jhead')) {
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Model/PhotoSet.pm
```diff
@@ -138,7 +138,7 @@ has ids => ( #  Arrayref of $fileid tuples (always, so post upload/raw data proc
                 my $upload = $part;
                 my $ct = $upload->type;
                 $ct =~ s/x-citrix-//; # Thanks, Citrix
-                my ($type) = $ct =~ m{image/(jpeg|pjpeg|gif|tiff|png)};
+                my ($type) = $ct =~ m{image/(jpeg|pjpeg|gif|tiff|png|webp)};
```
davea, line 141: Wonder if it's worth adding `webp` to the URL pattern in `F::A::Controller::Photo::index` as well? Unlikely, but if we hit the `catch` on L194 we'd store the original .webp which would then 404 when requested.

davea (approved):
One small suggestion, but overall looks good 👍

perllib/FixMyStreet/App/Model/PhotoSet.pm
dracos: Yeah, guess so; wonder how much that's actually happened. (Looks like we switched to trying to convert everything to JPEGs in 2023: https://github.com/mysociety/fixmystreet/pull/4778 )

PR #5968, opened 2026-05-11
[Dumfries] Trigger response template based on old vs new external status codes

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/5475.

Allows adding of an old external status code for response templates, so a specific template can be triggered if (e.g.) an Alloy update changes a report from 'Job Raised' to 'Reported'.

[skip changelog]


nephila-nacrea:
Different approach taken in https://github.com/mysociety/fixmystreet/pull/6000.

PR #5969, opened 2026-05-12
Use lower case instead of title case in faq titles

lucascumsille:
Fixes part of: https://github.com/mysociety/societyworks/issues/5541
Reference number: 3

This is part of the feedback received from Canal River Trust. Because it's a small change I figured is better to change it on the base template instead of creating a new template override.

[skip changelog]

### DIFF
--- a/t/app/controller/about.t
+++ b/t/app/controller/about.t
@@ -12,7 +12,7 @@ ok( my $mech = FixMyStreet::TestMech->new, 'Created mech object' );
 
 # check that we can get the page
 $mech->get_ok('/faq');
-$mech->content_like(qr{Frequently Asked Questions ::\s+FixMyStreet});
+$mech->content_like(qr{Frequently asked questions ::\s+FixMyStreet});
 $mech->content_contains('html class="no-js" lang="en-gb"');
 
 $mech->get_ok('/privacy');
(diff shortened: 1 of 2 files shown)


lucascumsille:
I know there is a good bunch of cobrands that also have their own template override. Let me know if you think we should me make it part of this PR.

dracos:
That BDA document doesn't mention title case? In fact the document uses title case for its headings!
eg **1. Dyslexia Friendly Text.** https://blogs.cardiff.ac.uk/LTAcademy/wp-content/uploads/sites/286/2025/10/British-Dyslexia-Association-Style-Guide.pdf another BDA doc that uses title case for its headings.

Though if it's easier to change for everyone than do it just for CRT, then that's fine, I don't mind the change :) Just thought it odd to say it was to do with dyslexia.

lucascumsille:
I understood it from this bit:
> Headings and Emphasis
> For headings, use larger font size in bold, lower case.

dracos:
I guess I took that to mean not all caps, and if it did, then why did they write `Headings and Emphasis`...! :)

lucascumsille:
Thanks @dracos =)
I amended the message of the PR/commit so it reflects why we made the change.

davea (approved):

PR #5970, opened 2026-05-12
FMS006 view dashboard permission

struan:
This adds the `view_dashboard` permission and requires this in order to view the dashboard.

Fixes mysociety/societyworks#5545

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Dashboard.pm
+++ b/perllib/FixMyStreet/App/Controller/Dashboard.pm
@@ -81,7 +81,10 @@ sub check_page_allowed : Private {
         } else {
             return $cobrand_body;
         }
-    } elsif ($c->user->from_body && (!$cobrand_body || $cobrand_body->id == $c->user->from_body->id)) {
+    } elsif (
+        $c->user->from_body
+        && (!$cobrand_body || $c->user->has_permission_to('view_dashboard', $cobrand_body->id))
+    ) {
         $body = $c->user->from_body;
 
         my @extra_bodies = $c->cobrand->call_hook('dashboard_extra_bodies');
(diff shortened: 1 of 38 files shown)


perllib/FixMyStreet/App/Controller/Dashboard.pm
```diff
@@ -81,7 +81,11 @@ sub check_page_allowed : Private {
         } else {
             return $cobrand_body;
         }
-    } elsif ($c->user->from_body && (!$cobrand_body || $cobrand_body->id == $c->user->from_body->id)) {
+    } elsif (
+        $c->user->from_body
+        && (!$cobrand_body || $cobrand_body->id == $c->user->from_body->id)
+        && $c->user->has_body_permission_to('view_dashboard')
```
dracos: I think these lines could be 
```
  $c->user->from_body
        && (!$cobrand_body || $c->user->has_permission_to('view_dashboard', $cobrand_body->id)
```
(I guess the no-cobrand-body check is to allow international installs that might not have a cobrand body set up like that)

dracos (changes requested):
Looks good, one potential simplifying of the check is all I think.

dracos (approved):
〰️

struan:
I've realised that the one flaw in this is that if you have a mixed deploy where some things have cobrand bodies and some do not then this will by default grant access to the dashboard for cobrands without a body to anyone who has access to view the dashboard for a body. E.g. the fms.com dashboard for all council users.

Although that does seem to be the existing behaviour too?

dracos:
Only superusers can log in to .com, no from_body users can, that's the default for any cobrand (see `admin_allow_user`). So as long as that's set up right, I think that's okay?

dracos:
Presumably last commit should be fixup of first commit, not the K&S one

PR #5971, opened 2026-05-13
[Bexley] Use historic DD list for DD amendments

dracos:
[skip changelog] For FD-6962 and FD-6992. Fixes https://github.com/mysociety/societyworks/issues/5544

Hopefully all makes sense and doesn't change any existing behaviour. I did some refactoring to hopefully make it a bit clearer where things are getting the DD reference from and try and separate out Pay360 from AccessPay bits cleanly. (Also as discussed record failure on the amendment report itself, might not have an orig_sub now.)

Also assumes Agile's payment method is correct in terms of checking the subscription payment method, as discussed.

Also, for showing of 'pending' message more correctly, FD-6992, spots DD status explanations when checking for existing on bin day page.


perllib/FixMyStreet/App/Controller/Waste.pm
```diff
@@ -1230,19 +1230,19 @@ sub process_enquiry_data : Private {
...
     my ($self, $c) = @_;
 
-    my $payment_method;
-
-    if ($c->stash->{orig_sub}) {
-        $payment_method = $c->stash->{orig_sub}->get_extra_field_value('payment_method');
+    if ($c->cobrand->can('garden_current_subscription')) {
+        my $service = $c->cobrand->garden_current_subscription;
+        # See if the backend returned the method e.g. Agile
+        return $service->{payment_method} if $service->{payment_method};
```
chrismytton: Looks like `garden_current_subscription` can return `undef` for various reasons, so think this needs to check that `$service` is defined before calling `->{payment_method}` on it.

chrismytton (changes requested):
This is great, thanks! Just spotted one thing is all.

chrismytton (approved):

PR #5972, opened 2026-05-13
[Shropshire] Prevent user in Councillor role leaving updates

MorayMySoc:
Adds a call hook to check if an individual user is not authorised to leave an update. This precedes/overrides general settings of which category of user is allowed to leave updates on reports.

For Shropshire this is any user in the Councillor role

https://github.com/mysociety/societyworks/issues/5546

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Shropshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Shropshire.pm
@@ -120,6 +120,22 @@ sub staff_ignore_form_disable_form {
         && $c->user->belongs_to_body( $self->body->id );
 }
 
+=head2 updates_disallowed_for_user
+
+Users in the Councillor role should not be able to update
+reports even though they are body members
+
+=cut
+
+sub updates_disallowed_for_user {
+    my ($self, $user, $from_body) = @_;
+
+    return unless $from_body;
+
+    my $role = $user->obj->roles->search({ name => "Shropshire Councillor" })->count;
+    return $role;
+}
+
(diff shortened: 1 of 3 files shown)


MorayMySoc:
Fixup is just that there seems to be no need to trouble the database unless we have to

perllib/FixMyStreet/Cobrand/Shropshire.pm
```diff
@@ -120,6 +120,24 @@ sub staff_ignore_form_disable_form {
...
+
+=cut
+
+sub updates_disallowed_for_user {
+    my ($self, $user, $from_body) = @_;
+
+    return unless $from_body;
+    my $role = FixMyStreet::DB->resultset("Role")->search({
+       'name' => 'Shropshire Councillor'
+    }, { 'join' => 'user_roles', 'where' => { 'user_id' => $user } });
```
dracos: If you passed in $c->user rather than $c->user->id to this function, I think this could be written as: `my $role = $user->roles->search({ name => "Shropshire Councillor" })->count;`
and the return could be just `return $role` (don't think you can be in same role more than once but no harm).

dracos (changes requested):
Looking good, only bit of simplifying of the Shropshire function I think

perllib/FixMyStreet/Cobrand/Shropshire.pm
MorayMySoc: As discussed, added ->obj

dracos (approved):

PR #5974, opened 2026-05-13
FMS006 scripts to update permissions for roles and staff users

struan:
Adds a couple of utility scripts for adding/removing permissions from roles or staff users.

One of the allows adding/removing permissions to all roles, a particular council bodies roles. This can be further limited to roles that have a particular permission for cases where a permission has been split.

The second does the same but for staff users without a role.

[skip changelog]

Fixes mysociety/societyworks#5550

### DIFF
--- a/bin/fixmystreet.com/update_council_users_without_role
+++ b/bin/fixmystreet.com/update_council_users_without_role
@@ -1,0 +1,37 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all staff without a role.
+
+use strict;
+use warnings;
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use FixMyStreet::Script::UpdateCouncilUsersWithoutRole;
+use Getopt::Long::Descriptive;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['commit', 'whether to commit changes to the database'],
+    ['permissions=s', 'comma seperated list of permissions to add/delete', { required => 1 }],
+    ['limit' => 'hidden' => { one_of => [
+        ['council=s', 'name of council to update'],
+        ['all', 'update users for all councils'],
+    ], required => 1 }],
+    ['mode' => 'hidden' => { one_of => [
+        ['add', 'add permissions to council\'s users'],
+        ['remove', 'remove permissions from council\'s users'],
+    ], required => 1 }],
+    ['with_permission=s', 'only update users that already have this permission'],
+    ['verbose', 'output a list of everything done'],
+    ['help|h', 'print usage message and exit']
+);
+$usage->die if $opts->help;
+
+FixMyStreet::Script::UpdateCouncilUsersWithoutRole::update($opts);
--- a/bin/fixmystreet.com/update_roles
+++ b/bin/fixmystreet.com/update_roles
@@ -1,0 +1,34 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all roles, or optionally to roles belonging to a body.
+
+use warnings;
+use v5.14;
+
+BEGIN {
+    use File::Basename qw(dirname);
+    use File::Spec;
+    my $d = dirname(File::Spec->rel2abs($0));
+    require "$d/../../setenv.pl";
+}
+
+use FixMyStreet::Script::UpdateRoles;
+
+use Getopt::Long::Descriptive;
+
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['commit', 'whether to commit changes to the database'],
+    ['permissions=s', 'comma seperated list of permissions to add/delete', { required => 1 }],
+    ['council=s', 'name of council to update (updates all roles if omitted)'],
+    ['mode' => 'hidden' => { one_of => [
+        ['add', 'add permissions to roles'],
+        ['remove', 'remove permissions from roles'],
+    ], required => 1 }],
+    ['with_permission=s', 'only update roles that already have this permission'],
+    ['verbose', 'output a list of everything done'],
+    ['help|h', 'print usage message and exit']
+);
+$usage->die if $opts->help;
+
+FixMyStreet::Script::UpdateRoles::update($opts);
--- a/perllib/FixMyStreet/Script/UpdateCouncilUsersWithoutRole.pm
+++ b/perllib/FixMyStreet/Script/UpdateCouncilUsersWithoutRole.pm
@@ -1,0 +1,85 @@
+package FixMyStreet::Script::UpdateCouncilUsersWithoutRole;
+
+use v5.14;
+use warnings;
+
+use FixMyStreet::DB;
+
+sub update {
+    my $opts = shift;
+
+    if (!$opts->{commit}) {
+        say "*** DRY RUN ***";
+    }
+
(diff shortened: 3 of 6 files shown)


bin/fixmystreet.com/update_council_users_without_role
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all staff without a role.
+
+use strict;
```
dracos, line 5: use strict is automatic when v5.14 is present, fwiw

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
...
+    });
+}
+my %permissions_list = map { Utils::trim_text($_) => 1 } split(',', $opts->permissions);
+
+for my $role ( $roles->all ) {
+    my $permissions = $role->permissions;
+
+    if ( $opts->mode eq 'remove') {
+        next unless $permissions;
+        $permissions = grep { !$permissions_list{$_} } $permissions;
```
nephila-nacrea: If I run (e.g.) `bin/fixmystreet.com/update_roles --permissions=category_edit --remove`
I get
`Can't use string ("1") as an ARRAY ref while "strict refs" in use at bin/fixmystreet.com/update_roles line 83.`
I think you need to deref the arrayrefs here to `@$permissions`.

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
...
+my ($opts, $usage) = describe_options(
+    '%c %o',
+    ['commit', 'whether to commit changes to the database'],
+    ['permissions=s', 'comma seperated list of permissions to add/delete', { required => 1 }],
+    ['council=s', 'name of council to update (updates all roles if omitted)'],
+    ['mode' => 'hidden' => { one_of => [
+        ['add', 'add permissions to roles'],
+        ['remove', 'remove permissions from roles'],
+    ], required => 1 }],
+    ['with_permission=s', 'only update roles that already have this permission'],
```
nephila-nacrea, line 28: Would we ever want to search using more than one permission here?

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
...
+
+if (!$opts->commit) {
+    say "*** DRY RUN ***";
+}
+
+my $body = FixMyStreet::DB->resultset("Body")->find({ name => $opts->council});
+
+my $roles;
+if ($body) {
+    $roles = FixMyStreet::DB->resultset("Role")->search({ from_body => $body->id });
```
nephila-nacrea: `from_body` needs to be `body_id`.

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
...
+    my $permissions = $role->permissions;
+
+    if ( $opts->mode eq 'remove') {
+        next unless $permissions;
+        $permissions = grep { !$permissions_list{$_} } $permissions;
+    } elsif ( $opts->mode eq 'add' ) {
+        my %existing = map { $_ => 1 } $permissions;
+        my @permissions_to_add = grep { !$existing{$_} } keys %permissions_list;
+        next unless @permissions_to_add;
+        say join ",", @$permissions;
```
nephila-nacrea: Do you need the 'say's here? Or otherwise, should you add similar for the remove block above?

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
...
+        $permissions = [@$permissions, @permissions_to_add];
+        say join ",", @$permissions;
+    }
+    if ($opts->commit) {
+        $role->permissions = $permissions;
```
nephila-nacrea: This should be $role->permissions($permissions). Otherwise I get a `Can't modify non-lvalue subroutine call of &FixMyStreet::DB::Result::Role::permissions` error.

nephila-nacrea: I also get weird behaviour - at least for `remove` - where the role permissions don't get updated in the DB despite the correct arrayref being passed in here. (I dump out the data here and it looks correct, but if I check my database, nothing has changed).
It may be related to pass-by-reference behaviour from the `my $permission = $role->permissions` being set above. I found if I made a separate `@new_permissions` array (not ref), things updated correctly 🤔

nephila-nacrea (changes requested):
This is for the update_roles script; I will look at the other script next.
[Later:] Matthew beat me to it 😁 but I gave update_council_users_without_role a go with various inputs and it works fine 👍

bin/fixmystreet.com/update_council_users_without_role
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all staff without a role.
```
dracos, line 3: There are a few staff with both roles and permissions, not sure how, or if it matters, probably not.

bin/fixmystreet.com/update_council_users_without_role
```diff
@@ -0,0 +1,101 @@
...
+    ['with_permission=s', 'only update users that already have this permission'],
+    ['help|h', 'print usage message and exit']
+);
+$usage->die if $opts->help;
+
+if (!$opts->commit) {
+    say "*** DRY RUN ***";
+}
+
+my $users = FixMyStreet::DB->resultset("User")->search({
```
dracos: Would be good for the body of this script to be in perllib/FixMyStreet/Scripts and then it can be tested more easily (and should have some tests).

bin/fixmystreet.com/update_council_users_without_role
```diff
@@ -0,0 +1,101 @@
...
+        say STDERR "Could not find " . $opts->council;
+        exit 1;
+    }
+}
+
+if ($users->count) {
+    my @permissions_list = map { Utils::trim_text($_) } split(',', $opts->permissions);
+
+    for my $user ( $users->all ) {
+        if ($opts->with_permission) {
```
dracos: Feels like we could have fetched only users with this permission in the original query, rather than check it per user, but perhaps it's awkward.

bin/fixmystreet.com/update_roles
dracos: Yes, I guess as the reference hasn't changed, it doesn't realise anything has changed.

bin/fixmystreet.com/update_roles
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env perl
+#
+# script to add/remove a list of permissions to all roles, or optionally to roles belonging to a body.
```
dracos, line 3: Moving to Script/tests for this would be good as well, would catch the things already raised.

dracos (changes requested):
Main thing would be some tests to catch the small issues, overall looks fine

bin/fixmystreet.com/update_council_users_without_role
struan: I did wonder about that but at least in this case decided that they would mostly be caught by adding the permission to existing roles

bin/fixmystreet.com/update_roles
struan: I was writing this sort of for the current two cases and decided this would complicated it a lot as is it roles with all or any of the permissions?

dracos (approved):

dracos:
I've merged this now - but I've only just spotted there is a `bin/fixmystreet.com/update_council_user_permissions` script that @struan added in 2019. So that should be removed or checked to see if new script does the same/more, or what :)

PR #5586, opened 2025-07-16
WIP add pin colour key to map controls

davea:
Needs more thought for mobile/accessibility.

pin colours

https://github.com/mysociety/fixmystreet/issues/1004

PR #5977, opened 2026-05-14
Show pin legends on map

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5466

**TODOs**
- [x] Desktop version needs to be hidden on mobile.
- [x] Needs a mobile solution. For now, I think I'll try adding a drawer inside the sidebar. It shouldn't use much space.
- [x] For the around page, because we have more space I have included another map item into the `.map_links`
- [x] Make it work on (landscape) tablet devices, where there’s no hover and no keyboard
- [x] Once the backend development is ready. We need some testing specially on tablet, where some cobrands might have less available space for the panel.

**TODOs Developer**
The only that comes to mind is:
- [x] Needs back-end work so the legend panel uses the pins and legends from their respective cobrands.

### Notes: 
- At the beginning I put the drawer on the right side with the other controllers on desktop, but it seemed a bit spaced limited considering the list of pins can get quite long.
- Happy to change the `pin-legend-wrapper--preview` for something else, for now it seem like a good approach for now. Considering that we have cobrands that use other pins, those previews could be problematic to implement.

## Desktop

https://github.com/user-attachments/assets/e2f59936-294e-454f-85f5-5cc1f365d3a9

### Keyboard navigation

https://github.com/user-attachments/assets/494fa76e-fa43-4e25-944d-4450c7e50b81

### Reports page

https://github.com/user-attachments/assets/f1d7ec7f-ac90-4f39-820c-3588410609cb

### Around page

https://github.com/user-attachments/assets/39440b52-f50c-4ce7-b029-41f04fa587bb


lucascumsille:
@dracos and @davea let me know if you have any thoughts about this approach, happy to make any changes. I'll try to have something for mobile before the user group.

lucascumsille:
@dracos I just added some version for mobile(/around and /reports pages)

davea:
The design of this supersedes #5586, but that PR's approach to configuring pin colours per-cobrand might be worth using.

lucascumsille:
For tablet I need to add a click for the desktop version.

lucascumsille:
@davea I left this PR ready for review, but it's not really ready for merge. It needs all the back-end changes so we use the pins and legends from each cobrand.

davea (approved):
This looks great! Dev work to add per-cobrand colours being tracked here: https://github.com/mysociety/societyworks/issues/5595.

lucascumsille:
@nephila-nacrea I added 3 commit fixup that solved some issues with cobrands that have bigger font sizes. They also solve some overlapping.

lucascumsille:
@nephila-nacrea let me know if after the backend changes have been added you want me to run another test.

lucascumsille:
@nephila-nacrea I rebased with master, but I just added two commit fixups, because when testing it I noticed that generally was fine...except for Aberdeenshire. So I did some layout changes. 
I tested on mobile and desktop and seems to be working, but I would appreciate one final pair of eyes.

If you are happy with it happy to be merged.

nephila-nacrea:
@lucascumsille Thanks, the horizontal lines look good!
I need to make sure tests are passing and fix if need be; think it's too late to deploy today so I'll have to try on Monday (FYI @lizettal ).

zarino (commented):
@nephila-nacrea and the other devs asked me to take a look at this PR, as it’s been fairly long-running and they were a bit worried it had started to lose its way.

As [mentioned originally](https://github.com/mysociety/societyworks/issues/5466) I’m still a bit worried that this new pull-out drawer feels _literally_ tacked-on to the existing interface, which was _already_ confusingly busy. But I admit, I can’t think of a simpler treatment, without re-architecting the entire page. 🤷 

It could be added to the stack of semi-transparent black controls down the right-hand side of the map. This would put it near to other “map” related stuff, including the hide/show pins button, which is conceptually nice. But I have no idea what one small white-on-black icon we’d use to connote “pin legend”. I’m also wary of hiding the list away – the more hidden it becomes, the less likely anyone will reveal it, rendering all of this work useless.

We could instead avoid a key entirely, and somehow show the status of each report when you hover over its pin. But then that wouldn’t automatically work on touchscreen devices (unless we force them to tap once to show status, tap again to open report). And you could only see the statuses one at a time. But I guess do people really _need_ to see all the statuses and their icons in a list, or do they just want to see “what does the colour of this particular pin mean”? 🤷 

Anyway, if we assume the current pull-out drawer arrangement is the best we can do, then…

Image

I agree with @dracos’ point (in Slack) that, if we’re going to use coloured pin icons in the button that reveals the legend, then they need to be icons that _actually_ appear on that cobrand’s map. Perhaps it could pick the icons for that cobrand’s first three categories, alphabetically, or something? (And, if there are are only two categories, then it only shows two pins in the button?)

---

Image

A very minor point, but, if we’re going to have dividers between the legend items, could we use an `li + li`-style selector and `border-top` so that the divider only appears _between_ the items and not after the final one? And could it use the same grey as the other horizontal divider line, in the sidebar, right next to it? (Maybe all of these lines should use the same SCSS colour variable? 🤔)

---

Image

I also wonder whether users will be confused by different statuses appearing in the map key versus the “Report status” filter dropdown? This is particularly noticeable on cobrands like Aberdeenshire where there are many statuses:

Image

Will they wonder why they can’t filter by “In progress” or “Investigating”? Will they be surprised when they filter to “Open” and it shows those “In progress” and “Investigating” reports? 🤷

lucascumsille:
> And could it use the same grey as the other horizontal divider line, in the sidebar, right next to it? (Maybe all of these lines should use the same SCSS colour variable? 🤔)

I added a commit so the multi select and and report filters use the existing `$form-control-border-color` variable, instead of static colours.

>  A very minor point, but, if we’re going to have dividers between the legend items, could we use an li + li-style selector and border-top so that the divider only appears between the items and not after the final one?

I just added a commit fixup for that.

lucascumsille:
> I agree with @dracos’ point (in Slack) that, if we’re going to use coloured pin icons in the button that reveals the legend, then they need to be icons that actually appear on that cobrand’s map. Perhaps it could pick the icons for that cobrand’s first three categories, alphabetically, or something? (And, if there are are only two categories, then it only shows two pins in the button?)

I added a [commit fixup](https://github.com/mysociety/fixmystreet/pull/5977/commits/edeb7bf3149498b3cd3e2c7585998511edf3a76c) for this as well.

Screenshot 2026-09-07 at 07 47 07
Screenshot 2026-09-07 at 07 47 19
Screenshot 2026-09-07 at 07 47 29

lucascumsille:
> I also wonder whether users will be confused by different statuses appearing in the map key versus the “Report status” filter dropdown? This is particularly noticeable on cobrands like Aberdeenshire where there are many statuses. Will they wonder why they can’t filter by “In progress” or “Investigating”? Will they be surprised when they filter to “Open” and it shows those “In progress” and “Investigating” reports? 🤷

If we thought it was something really important for the users, then we should include those subcategories in the filters. As an alternative and not sure if from a dev point of view is easy to do. As I understand "Investigating", "Planned", "Internal referral", etc. They are all subcategories of "Open" so we could create a list within a list item to make it more clear.

Screenshot 2026-09-07 at 07 56 26
Screenshot 2026-09-07 at 07 58 13

zarino:
Thanks for the most recent UI fixes @lucascumsille! I think let’s hold off on the visual nesting of subcategories for now, and see whether anyone actually reports it as a source of confusion once the feature’s gone live :+1:

So, from a UI point of view, I think this PR is good to go.

But, from a UX point of view, I think there’s still an unresolved conceptual problem that needs resolving before we can merge and deploy – namely that the pin labels we display often conflate report states with responsible bodies, and even when they don’t, the wording of labels containing the word “other” is pretty confusing.

This is a confusion that has always existed in the various cobrands, but has never really been noticed, because we’ve never before listed all of the pin colours/icons, and their labels, all in one place. Now we will be, and it needs cleaning up.

As far as I can tell, the `/around` page key can take one of three different shapes, depending on the cobrand (ignoring the fourth case, where the key is hidden because it only contains a single pin colour):

1. **Report states only** (~14 cobrands including Aberdeenshire, B&NES, Bexley, Bristol, Gloucester, etc) – each pin is labelled with a state like “Confirmed“, “Fixed”, “Closed” or “Not responsible”. From context, “Other” pretty clearly means ‘other state’.
1. **Responsibility only** (~15 cobrands including Camden, Dumfries, Enfield, Hackney, fixmystreet.com, and Cycling UK) – each pin is labelled with some form of ownership or responsibility, eg: “Other” (meaning reported by the public), and “Other council/authority”. For fixmystreet.com and Cycling UK, the two labels are “Responsibility of TfL” and “Other”. North and West Northants add a third “Defect” colour, which means that the issue was reported by staff rather than the public.
1. **Report states + responsibilities** (Bromley, Bucks, Hart, Lincolnshire, Oxfordshire, Rutland, Surrey) – a mixture of labels for states (eg: “Open“, “Fixed”, “Closed”) and a label for “Other council/authority” or “Not responsible”. Confusingly, in many of these cobrands, _the same pin colour/icon_ is used for _both_ “closed”, “not responsible“, and/or “other council/authority”, but fixing that is probably beyond the scope of this PR!

@dracos has suggested that we hide the key on cobrands where there are only two pin colours, as a way to limit the exposure of the confusing “Other” / “Other authority” duo. But I think, if there are two colours on the map, we should show a key explaining them. Otherwise we’re only doing half a job.

I would suggest that we need to standardise the labels, to make clear when “other” is about a state or a body.

That means, for the ~14 cobrands in my first list item, nothing needs to change. (Unless we want to be really belt-and-braces and change “Other” to “Other state”. I don’t think we need to.)

For the ~13 cobrands in my second list item, I’d suggest:

| Before | After |
| --- | --- |
| Defect | Reported by this council |
| Other | Reported to this council |
| Other council/authority | Reported to another council/authority |

For fixmystreet.com and Cycling UK, I’d suggest:

| Before | After |
| --- | --- |
| Other | Reported to the council or another authority |
| Responsibility of TfL | Reported to TfL |

For the cobrands in my third list item, we’ll need to vary the approach based on the logic that each cobrand uses in its `sub pin_colour`. As a general rule of thumb, though, I’d encourage using “Other states” instead of just plain “Other”, and “Reported to another council/authority” instead of just “Other council/authority”, to avoid that confusion between the two types of ‘other’ as much as possible.

The fixes we make for the `/around` page should also apply cleanly to the `/reports` page on most cobrands, where (as far as I can tell) the “Other council/authority” pin colour is removed – meaning in some cases, the number of pin colours drops to 1 and the key is hidden entirely, but in all the other cases, the key just gets one label shorter, but _I think_ all the remaining labels will continue to make sense without.

(I think the exceptions to this will be Rutland, Surrey, and Lincolnshire, where their `pin_colour_key` configurations mean nothing is removed and will sometimes even reference “other council/authority” even though no pins belonging to other councils/authorities are displayed on `/reports`; and fixmystreet.com and Cycling UK, where I think TfL pins continue to show on `/reports`.)

I see @nephila-nacrea started [a branch off this one](https://github.com/mysociety/fixmystreet/pull/6105), for some additional code fixes, but so far it only contains a debug script. I’ll leave it up to the developers as to whether this final adjustment of the labels happens in [that branch](https://github.com/mysociety/fixmystreet/pull/6105), or [this one](https://github.com/mysociety/fixmystreet/pull/5977).

PR #5985, opened 2026-05-18
[Bexley] Spot amendment failures, allow cancellation after subscription end

dracos:
[skip changelog]
For FD-7016 (amend failures) and FD-7008 (allowing cancel after sub end) and FD-6992 (storing verified method).

### DIFF
--- a/bin/bexley/check-dd-errors
+++ b/bin/bexley/check-dd-errors
@@ -46,8 +46,9 @@ my @kinds = (
         retry => sub {
             my ($p, $info) = @_;
             my $amount = $info->{context}{amount} or return;
+            my $ref = $info->{context}{ref} or return;
             my $id = $p->id;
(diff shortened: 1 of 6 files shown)


bin/bexley/check-dd-errors
```diff
@@ -46,8 +46,9 @@ my @kinds = (
         retry => sub {
             my ($p, $info) = @_;
             my $amount = $info->{context}{amount} or return;
+            my $ref = $info->{context}{ref} or return;
             my $id = $p->id;
-            return qq{bin/cron-wrapper perl -MFixMyStreet::DB -MFixMyStreet::Cobrand -e 'my \$r = FixMyStreet::DB->resultset("Problem")->find($id); my \$i = FixMyStreet::Cobrand->get_class_for_moniker("bexley")->new->get_dd_integration; \$i->amend_plan({ orig_sub => \$r, amount => "$amount" });'};
+            return qq{bin/cron-wrapper perl -MFixMyStreet::DB -MFixMyStreet::Cobrand -e 'my \$r = FixMyStreet::DB->resultset("Problem")->find($id); my \$i = FixMyStreet::Cobrand->get_class_for_moniker("bexley")->new->get_dd_integration; \$i->amend_plan({ dd_reference => $ref, report => \$r, amount => "$amount" });'};
```
chrismytton: I think `$ref` needs double quotes around it, like `$amount` has. Same on line 62 below.

chrismytton (changes requested):
Looks good, just one issue with the retry command syntax.

PR #5988, opened 2026-05-20
[FMS] Hide selected subcategory on load

MorayMySoc:
Flow goes to the subcategory page next
so duplicates selecting of subcategory

https://github.com/mysociety/societyworks/issues/4738

[skip changelog]


dracos (approved):
Thanks for working this out :)

dracos:
Only thing, do we want to have it up with the other classList changes?

PR #5989, opened 2026-05-20
[Dumfries] Remove unused statuses from dropdowns

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/5470 (Basecamp: https://3.basecamp.com/4020879/buckets/44203298/card_tables/cards/9683807634#__recording_9910044687).

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Dumfries.pm
+++ b/perllib/FixMyStreet/Cobrand/Dumfries.pm
@@ -307,11 +307,19 @@ sub is_scotland { 1 }
 
 sub state_groups_inspect {
     my $rs = FixMyStreet::DB->resultset("State");
+
+    my @open = grep { $_ !~ /^(action scheduled|for triage)$/ }
+        FixMyStreet::DB::Result::Problem->open_states;
+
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -307,11 +307,19 @@ sub is_scotland { 1 }
 
 sub state_groups_inspect {
     my $rs = FixMyStreet::DB->resultset("State");
+
+    my @open = grep { $_ !~ /^(action scheduled|for triage|twin)$/ }
```
dracos: twin doesn't exist except on staging, so probably best not to have this here at present.
(And then an exemption for Dumfries on staging would be added to the issues/freshdesk/5202-northumberland-twin-state branch to cover it there instead)

dracos (changes requested):
One meta issue with 'twin' is all :)

dracos (approved):
Thanks :)
I rebased/merged the old staging 'twin' branch and added twin to this file there for the merge, so staging should start passing again now.

PR #5991, opened 2026-05-20
Form saved data tweaks

dracos:
Tidy up some Digest calls, sign any JSON fields, and fix a few internal message fields I spotted on the way. [skip changelog]


neprune (approved):
👌

PR #5995, opened 2026-05-26
[Kingston][WW] Dispute closed cases

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/5554.

Will need Granicus `dispute_url` in configs for staging and live.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
@@ -239,11 +239,10 @@ around booked_check_missed_collection => sub {
             my $open_dispute = $self->_check_for_open_disputes($disputes, $missed_event->{report}->external_id);
 
             my $within_window = $self->_check_date_within_dispute_window(
-                $missed_event->{date}
+                $missed_event->{date}, 'is_bulky',
             );
             my $resolution_valid = $self->waste_check_can_raise_dispute(
-                $missed_event->{service_id},
-                $missed_event->{ResolutionCodeId}
+                type => 'missed_collection_report',
             );
 
             if ($open_dispute){
@@ -257,11 +256,10 @@ around booked_check_missed_collection => sub {
         } elsif ($locked_out) {
             my $open_dispute = $self->_check_for_open_disputes($disputes, $guid);
             my $within_window = $self->_check_date_within_dispute_window(
-                $missed->{$guid}{report_locked_out_date}
+                $missed->{$guid}{report_locked_out_date}, 'is_bulky',
             );
             my $resolution_valid = $self->waste_check_can_raise_dispute(
-                $missed->{$guid}{service_id},
-                $missed->{$guid}{report_locked_out_reason}
+                resolution_key => $missed->{$guid}{resolution_id},
             );
             if ($open_dispute) {
                 $missed->{$guid}{dispute}{open} = 1;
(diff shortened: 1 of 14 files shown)


templates/email/default/_email_comment_list.txt
```diff
@@ -11,6 +11,8 @@
 [% email_sanitize_text(row, private_email ? "item_private_email_text" : undef) -%]
 [% IF cobrand.moniker == 'sutton' AND (report.category == 'Report missed collection' OR report.category == 'Complaint against time') AND (row.item_problem_state == 'closed' OR row.item_problem_state == 'unable to fix') -%]
 To report a problem with this missed collection visit your bin page
+[% ELSIF cobrand.moniker == 'kingston' AND (report.category == 'Report missed collection' OR report.category == 'Complaint against time') AND (row.item_problem_state == 'closed' OR row.item_problem_state == 'unable to fix' OR row.item_problem_state == 'fixed') %]
```
dracos: Just checking - why is Kingston fixed as well, but Sutton not? Should they just be the same?

templates/email/default/_email_comment_list.html
```diff
@@ -21,16 +21,31 @@
...
+          </p>
+        [% END %]
 
-      <p>
-      <a href="[% cobrand.base_url %]/waste/[% report.get_extra_field_value('property_id') %]/enquiry?category=Missed+collection+dispute&service_id=[% report.get_extra_field_value('service_id') %]&original_booking_id=[% report.id %]">Report a problem with this missed collection</a>
-      </p>
+        <p>
+          <a href="[% cobrand.base_url %]/waste/[% report.get_extra_field_value('property_id') %]/enquiry?category=Missed+collection+dispute&service_id=[% report.get_extra_field_value('service_id') %]&original_booking_id=[% report.id %]">Report a problem with this missed collection</a>
+        </p>
+      [% ELSIF cobrand.moniker == 'kingston' AND (report.category == 'Report missed collection' OR report.category == 'Complaint against time') AND (update.item_problem_state == 'closed' OR update.item_problem_state == 'unable to fix' OR update.item_problem_state == 'fixed') %]
```
dracos: Should this be Sutton as well? Even if it's always true, think will be easier if kingston/sutton match if possible (but again, possibly missed something regarding 'fixed')

templates/email/default/waste/_bulky_email_comment_end.html
```diff
@@ -4,20 +4,41 @@
...
+    [% ELSE %]
+        <p>
+        Our crews reported your bulky waste collection was not made for the reason above.
+        </p>
+    [% END %]
+
+    <p>
+        <a href="[% cobrand.base_url %]/waste/[% property_id %]/enquiry?category=Missed+collection+dispute&service_id=[% service_id %]&original_booking_id=[% report.id %]">Report a problem with this missed collection</a>
+    </p>
+[% ELSIF cobrand.moniker == 'kingston' %]
```
dracos: Ditto

perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -423,6 +424,9 @@ sub _alert_iterator {
 sub _send_aggregated_alert(%) {
     my %data = @_;
 
+    $data{item_extra}
```
dracos: Does this change mean that the two decodings in Template.pm needs to be removed?
(Not sure why that hasn't failed somewhere, though!)

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -298,15 +307,44 @@ sub _check_for_open_disputes {
...
+}
 
 sub munge_bin_services_for_address {
     my ($self, $rows) = @_;
 
-    # Escalations
+    # store all the missed events which we need for checking details
+    # of incomplete missed collection re-attempts when clicking through
+    # from a link in an alert email
+    my $events = $self->_parse_events($self->{api_events});
```
dracos: Probably no bother, but I wonder if we could avoid parsing the events again, given that's already happened. I guess that's why they were in the `booked_check_missed_collection` but then they were being checked twice I assume (bulky and small items)? Probably fine to leave it for now.

dracos (changes requested):
All looks good :) Some small queries and possible changes.

templates/email/default/_email_comment_list.txt
nephila-nacrea: I don't think Sutton had a 'fixed' check originally and I focused on the Kingston requirements, but yes, I think both need it.

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
nephila-nacrea: I put them here because iirc I also needed them for standard missed collections, not just bulky/small. Would be nice to refactor/tidy later.

nephila-nacrea: Refactored/fixed another section and turns out this is no longer needed anyway.

nephila-nacrea: Ugh never mind, it's actually needed for some weirdness with Sutton.

dracos (approved):
Think this all looks good - want to check if a couple of the Sutton fixes in here e.g. the service_id in the email, should be being included as fixup on the Sutton commit rather than in here?  (We can fix that at a later point if so)

nephila-nacrea:
@dracos 

> Think this all looks good - want to check if a couple of the Sutton fixes in here e.g. the service_id in the email, should be being included as fixup on the Sutton commit rather than in here? (We can fix that at a later point if so)

Yeah, that should be included in the Sutton changes. (What were the other Sutton fixes? I'd need to look through things again.)

dracos:
I guess the problem state being fixed, or the `booked_missed` `Original_Event_ID` bit, that kind of thing

PR #5998, opened 2026-05-27
[TfL] Ignore special type contacts in reporting.

dracos:
For staff as well as non-staff now. FD-7050
Fixes https://github.com/mysociety/societyworks/issues/5557


dracos:
(feel free to put live while I'm not here if it looks okay)

neprune (approved):

PR #5999, opened 2026-05-28
Add scripts for bulk staff management.

neprune:
[skip changelog]

Motivated by https://mysocietysupport.freshdesk.com/a/tickets/7034.


neprune:
As noted by @davea, there is already a web flow for batch staff add that I missed. Converting to draft while I integrate with that instead.

PR #6000, opened 2026-05-29
[Dumfries] Handle Cancelled Job response templates for Open311 updates

davea:
FIxes https://github.com/mysociety/societyworks/issues/5475

This adds a cobrand hook which allows a specific response template to be chosen according to the problem's old/new status. For Dumfries that hook looks for a response template whose title begins 'Cancelled Job' and uses that when a problem state goes from 'planned' to 'confirmed'. Also adds validation for these response templates when edited via the admin - can't be marked as auto response, can't have an external status, must be for 'confirmed' state.

(alternative approach to https://github.com/mysociety/fixmystreet/pull/5968)

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Templates.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Templates.pm
@@ -126,6 +126,14 @@ sub edit : Path : Args(2) {
             }
 
             $template->auto_response( $c->get_param('auto_response') && ( $template->state || $template->external_status_code ) ? 1 : 0 );
+
+            if ($body_cobrand) {
+                my $errors = $body_cobrand->call_hook(validate_response_template => $template);
+                if ($errors && ref $errors eq 'HASH') {
+                    $c->stash->{errors}{$_} = $errors->{$_} for keys %$errors;
+                }
+            }
+
             if ($template->auto_response) {
                 my @check_contact_ids = @new_contact_ids;
                 # If the new template has not specific categories (i.e. it
--- a/perllib/FixMyStreet/Cobrand/Dumfries.pm
+++ b/perllib/FixMyStreet/Cobrand/Dumfries.pm
@@ -323,6 +323,59 @@ sub state_groups_inspect {
     ]
 }
 
+=head2 response_template_override
+
+When an update moves a report from planned to confirmed, use the first
+matching response template whose title starts with "Cancelled Job".
+
+=cut
+
+sub response_template_override {
+    my ($self, $problem, $state, $old_state) = @_;
+
+    return unless $old_state eq 'planned' && $state eq 'confirmed';
+
+    return $problem->response_templates->search({
+        'me.title' => { -like => 'Cancelled Job%' },
+    }, {
+        order_by => 'me.title',
+    })->first;
+}
+
+
+=head2 validate_response_template
+
+Performs some custom validation for templates whose title start with 'Cancelled Job':
+can't be auto-response; can't be for an external status code; must be for 'confirmed' state.
+
+=cut
+
+sub validate_response_template {
+    my ($self, $template) = @_;
+
+    return unless ($template->title || '') =~ /^Cancelled Job/;
+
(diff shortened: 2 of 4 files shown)


perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -323,6 +323,57 @@ sub state_groups_inspect {
...
+    })->first;
+}
+
+
+=head2 validate_response_template
+
+
+Performs some custom validation for templates whose title start with 'Cancelled Job':
+can't be auto-response; can't be for an external status code; must be for 'confirmed' state.
+=cut
```
nephila-nacrea, line 351: Would look better to move one of the empty lines down to be just above `=cut`.

perllib/FixMyStreet/App/Controller/Admin/Templates.pm
```diff
@@ -126,6 +126,15 @@ sub edit : Path : Args(2) {
             }
 
             $template->auto_response( $c->get_param('auto_response') && ( $template->state || $template->external_status_code ) ? 1 : 0 );
+
+            if ($body_cobrand) {
+                my $errors = $body_cobrand->call_hook(validate_response_template => $template);
+                if ($errors && ref $errors eq 'HASH') {
+                    $c->stash->{errors} ||= {};
+                    %{ $c->stash->{errors} } = (%{ $c->stash->{errors} }, %$errors);
```
nephila-nacrea: I wonder if this would read better as something like
`$c->stash->{errors}{$_} = $errors->{$_} for keys %$errors;`

t/cobrand/dumfries.t
```diff
@@ -759,6 +761,172 @@ subtest 'admin template external_status_code validation' => sub {
...
+            state => 'confirmed',
+            external_status_code => '',
+        });
+
+        my $cancelled_job_template = FixMyStreet::DB->resultset('ResponseTemplate')->create({
+            body_id => $body->id,
+            title => 'Cancelled Job - Planned to Confirmed',
+            text => 'Cancelled job template text',
+            auto_response => 0,
+            state => '',
```
nephila-nacrea: Should the template have either state or external_status_code?

t/cobrand/dumfries.t
```diff
@@ -759,6 +761,172 @@ subtest 'admin template external_status_code validation' => sub {
...
+            $mech->submit_form_ok({ with_fields => {
+                title => 'Cancelled Job - Reserved Template',
+                text => 'Template text',
+                state => 'confirmed',
+                auto_response => 'on',
+            } });
+
+            is $mech->uri->path, '/admin/templates/' . $body->id . '/new',
+                'Not redirected when reserved template is set as auto-response';
+            $mech->content_contains('reserved for the cancelled job feature',
```
nephila-nacrea: Might be better to specifically test for the 'cannot be set as auto-response' bit.

perllib/FixMyStreet/Cobrand/Dumfries.pm
```diff
@@ -323,6 +323,57 @@ sub state_groups_inspect {
...
+sub validate_response_template {
+    my ($self, $template) = @_;
+
+    return unless ($template->title || '') =~ /^Cancelled Job/;
+
+    my %errors;
+
+    if (($template->state || '') ne 'confirmed') {
+        $errors{state} =
+            'Templates whose title starts with "Cancelled Job" are reserved for cancelled job handling and must use the Open state.';
```
nephila-nacrea: You could put the initial part of the error message into a shared variable, as it is used by all the error messages.

nephila-nacrea (changes requested):

perllib/FixMyStreet/App/Controller/Admin/Templates.pm
davea: That autovivification gets me every time :)

nephila-nacrea (approved):

PR #6003, opened 2026-06-01
[TfL] Treat multiple location licences the same.

dracos:
Like column attachments, banners and temporary traffic signs will now also lock the street name / borough.
Factor this out to a shared role. [skip changelog] FD-7071


neprune (approved):
Nice 🙏

PR #6004, opened 2026-06-01
[pboro] Apply private land check to street cleansing and grounds maint.

neprune:
For https://mysocietysupport.freshdesk.com/a/tickets/7067.

[skip changelog]

### DIFF
--- a/data/test-asset-layers.yml
+++ b/data/test-asset-layers.yml
@@ -455,24 +455,38 @@ peterborough:
 
   - asset_category: ['General fly tipping', 'Hazardous fly tipping' ]
     template: 'pcc_property_combined'
-    message_template: '#js-environment-message'
+    no_asset_msg_id: '#js-environment-message'
   - asset_category: ['General fly tipping', 'Hazardous fly tipping' ]
     template: 'pcc_property_leased_out'
-    message_template: '#js-environment-message'
+    no_asset_msg_id: '#js-environment-message'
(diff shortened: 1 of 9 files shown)


data/test-asset-layers.yml
```diff
@@ -474,6 +474,20 @@ peterborough:
     template: 'pcc_property_leased_out'
     message_template: '#js-abandoned-vehicles-message'
 
+  - asset_group: ['Street cleansing']
+    template: 'pcc_property_combined'
+    message_template: '#js-street-cleansing-message'
```
dracos: I guess for consistency with what's actually in place on the server, it's using `no_asset_msg_id` there, not `message_template` (which looks like that was for to display a message, but still allow reporting?). So actually, these new ones could be done without template additions but by using `no_asset_message` to put the message in, which would save on the template changes being needed (though I see how that's consistent with the ones already in place, so perhaps not worth the change there)

dracos (changes requested):
One comment about not using `message_template` given that's not what it's actually using

data/test-asset-layers.yml
neprune: Didn't migrate to `no_asset_message` but wish I'd spotted that earlier, alas. Did update the test cases to use `no_asset_msg_id` though.

dracos (approved):
Thanks

PR #6007, opened 2026-06-02
[Bexley] Treat inactive DD Agile subs same as card.

dracos:
[skip changelog] For FD-7057

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
+++ b/perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
@@ -84,19 +84,6 @@ sub lookup_subscription_for_uprn {
         $customer = $_;
         for ( @$contracts ) {
             next unless $_->{UPRN} == $uprn;
-
-            # NB Options for ServiceContractStatus seem to be
-            # ACTIVE/NOACTIVE/PRECONTRACT/RENEWALDUE.
-            # We only do a check for ServiceContractStatus if payment type
-            # is direct debit, as we otherwise want to continue offering a
-            # renewal option to contracts that were cancelled or expired in
-            # the past 3 months.
-            #
-            # TODO Can we always rely on the first payment type?
-            next
-                if $_->{ServiceContractStatus} eq 'NOACTIVE'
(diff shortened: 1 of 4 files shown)


perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
```diff
@@ -114,7 +101,10 @@ sub lookup_subscription_for_uprn {
 
     $sub->{payment_method} = $payment_method;
 
-    if ($payment_method eq 'direct_debit') {
+    # NB Options for ServiceContractStatus seem to be
```
nephila-nacrea, line 104: I wonder if we had the original logic in place due to this line in [the tech spec](https://docs.google.com/document/d/1dzthN_7PRcIBuJSeV3EUs29HQZuLL6Zj6K0vRpbeRG4/edit?tab=t.0#bookmark=id.mg01w6habibe):

> Renewing a subscription section applies only to card payment users.

Judging from Rebecca's issue on https://mysocietysupport.freshdesk.com/a/tickets/7057, however, it looks like they do want to handle the case where an early-cancelled DD can be renewed as CC 🤷 

I suppose we can put this fix in place and see what else crops up.

t/app/controller/waste_bexley_garden.t
nephila-nacrea: Not massively important, but it looks like customer_status/CustomertStatus is no longer used. Might be good in the long run to get rid of it in tests to prevent confusion; I kept thinking it was something important.

nephila-nacrea (approved):

perllib/FixMyStreet/Cobrand/Bexley/Garden.pm
dracos: I think "Renewing a subscription section applies only to card payment users." was meant to be referring to the fact active direct debits renew themselves.

neprune (approved):
Looks good. Can see the tests confirm this doesn't open the door to cancellations / amendments in the rendered page, and I guess we don't need to worry about a backend guard here anyway.

PR #6008, opened 2026-06-03
[Bexley] Check first letter of first name only.

dracos:
For FD-7088.
[skip changelog]


nephila-nacrea (approved):
That seems straightforward enough 🤞 Thanks!

PR #6012, opened 2026-06-05
[Canal] Remove categories from cobrands

MorayMySoc:
Whilst developing we don't want the categories showing on cobrands as causing confusion to other councils

https://github.com/mysociety/societyworks/issues/5549

[skip changelog]


davea (approved):
⛵ 🆗

PR #6013, opened 2026-06-10
No-JS map changes

dracos:
Fixes a bug in a redirect from /report/new to /around where it wasn't remembering the show_old_reports setting.
Replaces the non-JS map control links with buttons that should do the same thing.
And add a file check to allow easy override to OSM mapping should that be needed in future.
[skip changelog]


neprune (approved):
🤖 🗑️

PR #6014, opened 2026-06-12
[FMS] Repeated report submission generating multiple reports

MorayMySoc:
Some browsers keep submitting the report leading to multiple accidental reports made by the user.

This happens although the continue button is disabled.

Try testing the submitted bool before sending and also setting it to discard any further submissions.

It is already reset to 'false' on validation error.

https://github.com/mysociety/societyworks/issues/5506

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -552,7 +552,12 @@ $.extend(fixmystreet.set_up, {
         },
         submitHandler: function(form) {
             $('input[type=submit][data-disable],button[data-disable]', form).prop("disabled", true);
-            form.submit();
+            if (!submitted) {
+                submitted = true;
+                setTimeout(function() {
+                    form.submit();
+                }, 0);
+            }
         },
         // make sure we can see the error message when we focus on invalid elements
         showErrors: function( errorMap, errorList ) {


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -552,7 +552,10 @@ $.extend(fixmystreet.set_up, {
         },
         submitHandler: function(form) {
             $('input[type=submit][data-disable],button[data-disable]', form).prop("disabled", true);
-            form.submit();
+            if (!submitted) {
+                submitted = true;
+                form.submit();
```
davea: I added a `sleep 10` to `Report::New::report_new` to simulate a slow connection, and noticed that in Safari (on both iOS and macOS) the page would freeze before the button had been visibly disabled. 
Wrapping it in a `setTimeout` allows the browser to finish any pending DOM changes before doing the form submit, and, in my testing at least, meant the button was always reliably shown as disabled before the report form was submitted:

```suggestion
                setTimeout(function() {
                    form.submit();
                }, 0);
```

(prior art: https://github.com/mysociety/fixmystreet/commit/b654b6ea555cf2f1ebd6e57b237265b4b9d4e9b9#diff-52731772b0030538ca88d9d8d1f612094bf4d8a2eeeabebd8c7c5c4644957412R170)

davea (approved):
Looks good, one small suggestion to improve rendering in Safari.

dracos:
I'm afraid I've had to roll this back because it broke the SSO sign in form and the inspector form. Both use the button clicked to decide upon what to do, which works fine, but my assumption is that moving this into a timeout meant the button click was lost in the subsequent frame form.submit() - that's the only thing I can think of, anyway.

PR #6016, opened 2026-06-15
[Gloucester] Add staff user to CSV export.

dracos:
FD-7156. [skip changelog]


chrismytton (approved):
Looks good!

As an aside, it looks like the `staff_user` and `staff_roles` keys of `FixMyStreet::Script::CSVExport::$EXTRAS` are both redundant since e1f87b41b2, so I think can be deleted?

dracos:
Yeah, I guess perhaps I thought that could be documentation, but it'll just get out of date (like now).

PR #6017, opened 2026-06-15
Cypress upgrade

dracos:


.cypress/cypress/e2e/photos.cy.js
```diff
@@ -1,24 +1,22 @@
...
     cy.get('#map_box').click(200, 200);
     cy.wait('@report-ajax');
     cy.pickCategory('Flyposting');
     cy.nextPageReporting();
-  });
-  it('uploads a photo', function() {
-    cy.server();
-    cy.route('POST', '/photo/upload').as('photo-upload');
+  //});
+  //it('uploads a photo', function() {
```
struan: some debug comments here?

struan (approved):
few extraneous commented out lines but otherwise works for me and makes sense.

.cypress/cypress/e2e/photos.cy.js
dracos: Hmm, well, the existing test has it split into 3, but that didn't work any longer because it resets each `it`. 
I'll just change them to normal comments for now.

PR #6018, opened 2026-06-16
Make confirmation text when deleting role more explicit.

neprune:
[skip changelog]


dracos (approved):
Thanks!

dracos:
Was merged as f844270597

PR #6021, opened 2026-06-18
[K/S] Container disputes

dracos:
For https://github.com/mysociety/societyworks/issues/5574 [skip changelog]

Hopefully all makes sense. Not particularly happy with the passing of the container request ID through the category, but can't see any easier way to do it with this set up, and works okay for what it's doing at present.


MorayMySoc (approved):
Ok, this seems good to me.

PR #6022, opened 2026-06-18
[Kingston] Non actionable spillage/not returned reports

MorayMySoc:
If we are outside of the standard reporting an issue timeframe, Kingston would like reports to still be allowed for waste spillage issues and reporting containers as not returned.

These will not be actioned by the council, but so they can monitor the amount of issues raised for service improvement.

https://github.com/mysociety/societyworks/issues/5575

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
@@ -451,6 +451,12 @@ sub _enquiry_nice_title {
         $category = 'Issue with delivery';
     } elsif ($category eq 'Report out-of-time missed collection') {
         $category = 'Reporting a missed collection outside the reporting window';
+    } elsif ($category eq 'Report out-of-time spillage') {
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -502,6 +508,10 @@ sub waste_munge_enquiry_data {
             . $self->service_name_override( { ServiceId => $data->{service_id} } );
 
         $data->{extra_Notes} = 'Non-actionable missed collection report';
+    } elsif ($data->{category} eq 'Report out-of-time spillage') {
+        $data->{extra_Notes} .= "/n/nNon-actionable spillage report";
```
dracos: Think you mean `\n` :)
Also probably want this first, to explain what it is with the user provided stuff after?

dracos (changes requested):
One change/bug to how the notes is stored

PR #6025, opened 2026-06-24
Removed `text-transform: uppercase` from Big green banner

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5533

### Preview here:
[Archive.zip](https://github.com/user-attachments/files/29284465/Archive.zip)

[skip changelog]

### DIFF
--- a/web/cobrands/bathnes/layout.scss
+++ b/web/cobrands/bathnes/layout.scss
@@ -83,10 +83,6 @@
 
 // Map page
 
-.big-green-banner {
-    text-transform: none;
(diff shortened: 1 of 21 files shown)


lucascumsille:
@chrismytton I left the commits separate, but I think once the PR is ready I could just squash all into one, unless you think is fine as it is.

chrismytton (approved):
Thanks! I think its ok to leave them as separate commits, makes it clear what's happening at each stage.

PR #6026, opened 2026-06-24
[Dumfries] Handle detailed_information from O311 updates

nephila-nacrea:
This was previously tied to _handle_assigned_user (for Northumberland).

Part of fix for https://github.com/mysociety/societyworks/issues/5477.

Corresponding O311 changes: https://github.com/mysociety/open311-adapter/pull/513.

Needs O311 config change for staging and live.

[skip changelog]


nephila-nacrea:
Change not really needed, though it would be good to get `detailed_information` out from under the `assigned_user` umbrella.

PR #6027, opened 2026-06-25
[Greenwich] Add DVLA lookup for Greenwich

MorayMySoc:
https://github.com/mysociety/societyworks/issues/5560

[skip changelog]

### DIFF
--- a/web/cobrands/fixmystreet-uk-councils/dvla.js
+++ b/web/cobrands/fixmystreet-uk-councils/dvla.js
@@ -32,6 +32,15 @@ const FIELDS = {
             'yes': 'Y',
             'no': 'N'
         }
+    },
+    'greenwich': {
+        'block': true,
+        'categories': [
+            'Abandoned vehicles'
+        ],
+        'reg': 'vehicle_registration',
+        'make': 'vehicle_make',
+        'colour': 'vehicle_colour',
     }
 };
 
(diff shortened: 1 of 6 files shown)


struan (approved):
looks good.

MorayMySoc:
Hi @struan - turns out cars under 3 years old are exempt from MOT and it returns 'No details held by DVLA'. But figure it's better to just work against 'Not valid' if that makes sense https://developer-portal.driver-vehicle-licensing.api.gov.uk/apis/vehicle-enquiry-service/v1.2.0-vehicle-enquiry-service.json

web/cobrands/fixmystreet-uk-councils/dvla.js
```diff
@@ -44,7 +44,7 @@ const TYPES = {
 const REASONS = {
     'greenwich': {
         fn: function (data) {
-            if ( data.taxStatus == 'SORN' || (data.taxStatus == 'Taxed' && data.motStatus == 'Valid')) {
+            if ( data.taxStatus == 'SORN' || (data.taxStatus == 'Taxed' && data.motStatus != 'Not valid')) {
```
struan, line 83: I guess maybe a comment here to explain the 3 year thing? Should also update the generic case further down the file too.

Do we know what "No results returned" means? I can't tell from the documentation why that would happen.

struan (changes requested):
One small request for a comment, plus to reflect the 3 year thing in the main MOT status check.

web/cobrands/fixmystreet-uk-councils/dvla.js
MorayMySoc: I don't really know - it looks maybe like it's vehicles that may never have been registered somewhere in some way - like old cars which are exempt, but never expected to be on the system

struan (approved):
looks good

PR #6028, opened 2026-06-26
[Central Bedfordshire] Adds site name

MorayMySoc:
Just doing a PR in case there could be any repercussions I haven't seen though seems to be consistent with most Cobrands

https://mysocietysupport.freshdesk.com/a/tickets/7223

[skip changelog]


chrismytton (approved):
Thanks!

PR #6030, opened 2026-06-26
[Northumberland][FMS] Adds new cookie banner

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/7207

[skip changelog]

### DIFF
--- a/templates/web/northumberland/tracking_code.html
+++ b/templates/web/northumberland/tracking_code.html
@@ -1,46 +1,4 @@
 [% IF c.config.BASE_URL == "https://www.fixmystreet.com" %]
-<script src="https://cc.cdn.civiccomputing.com/9/cookieControl-9.x.min.js" type="text/javascript"></script>
-<script>
-    var config = {
-        apiKey: 'ce75ba48a905669e1736e11d267d26ce2c39d8ea',
-        product: 'PRO',
-        initialState: 'top',
-        theme: 'light',
-        settingsStyle: 'button',
-        toggleType: 'checkbox',
-        text: {
-            notifyTitle: 'Tell us whether you accept cookies',
-            notifyDescription: 'We use cookies to collect information about how you use fix.northumberland.gov.uk. We use this information to make the website work as well as possible and improve our services.',
-            accept: 'Accept all cookies',
-            reject: 'Decline all but essential cookies',
-            settings: 'Set preferences',
-            acceptSettings: 'Accept all cookies',
-            rejectSettings: 'Decline all but essential cookies',
-            title: 'Cookies on fix.northumberland.gov.uk',
-            intro: 'Cookies are files saved on your phone, tablet or computer when you visit a website. We use cookies to store information about how you use the fix.northumberland.gov.uk website, such as the pages you visit.',
-        },
-        optionalCookies: [
-            {
-                name: 'analytics',
-                label: 'Cookies that measure website use',
-                description: 'We use Google Analytics to measure how you use the website so we can improve it based on user needs. We do not allow Google to use or share the data about how you use this site.',
-                cookies: ['_ga', '_gid', '_gat', '__utma', '__utmt', '__utmb', '__utmc', '__utmz', '__utmv'],
-                onAccept: function(){
-                    window.dataLayer = window.dataLayer || [];
-                    function gtag(){dataLayer.push(arguments);}
-                    gtag('js', new Date());
-                    gtag('config', 'G-FSE5G4JX5L');
-                },
-                onRevoke: function (){
-                    //Disable Google Analytics
-                    window['ga-disable-G-FSE5G4JX5L'] = true;
-                    //End Google Analytics
-                }
-            }
-         ]
-    };
-    CookieControl.load( config );
-</script>
-<!-- Google tag (gtag.js) -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=G-FSE5G4JX5L"></script>
-[% END %]
+<link rel="stylesheet" href="[% version('/cobrands/northumberland/vendor/_tracking_code.css') %]">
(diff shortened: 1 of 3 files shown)


davea (changes requested):
Functionally this looks fine. However including the JS/CSS in this template means they'll be included in every single response - quite a lot of bandwidth!

Please put the contents of the third `<script>` tag (starting L105) and the `<style>` tag (L946) into their own vendored files which are then referenced from this HTML file. Because the Silktide JS code uses the `?.` operator which might not be available in all browsers we aim to support, please use `<script type="module">` when including it as this should stop older browsers breaking the page.

Design-wise the button being present on every single page is a bit offensive, especially on mobile where it overlaps with a lot of our UI.

Image

Can we hide this button everywhere except the homepage, or similar?

MorayMySoc:
Ok, I've separated the vendor js out and kept the gtag code in the main script of the template - https://silktide.com/consent-manager/docs/google-consent-mode/ (that was in before too, but just for info).

L738 of ```_tracking_code.js``` has the hiding of the cookie image when not on the front page for the reasons of space outlined.

This appears to work now, but I wouldn't take my js cookie skills over templates as read!

templates/web/northumberland/tracking_code.html
```diff
@@ -1,46 +1,106 @@
 [% IF c.config.BASE_URL == "https://www.fixmystreet.com" %]
-<script src="https://cc.cdn.civiccomputing.com/9/cookieControl-9.x.min.js" type="text/javascript"></script>
+<link rel="stylesheet" href="[% version('/cobrands/northumberland/vendor/_tracking_code.css') %]">
+<script type='module' src=[% version('/cobrands/northumberland/vendor/_tracking_code.js') %]></script>
```
dracos, line 3: Sorry, I think the module suggestion might have come from me. If this was instead `async`, then I think *all* the JS could be moved into this file, which I assume would be preferable. Unless there was some reason it is staying in the HTML.

MorayMySoc: I've moved all the js into __tracking_code.js_ but if I change to `<script async src=...` which is as it looks to be meant to be initiated, then nothing happens. But leaving it as `type='module'` it does all seem to work from the vendor file.

davea: `<script type='module' async src=` seems to work OK for me in Safari and Chrome on macOS. What browser are you on?

MorayMySoc: I took @dracos saying 'instead' to mean replace `type='module'` so I meant async alone didn't work, but it still all works in one file without async. I'm not clear what I'm trying to achieve here, so I don't know what matters

dracos: https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6 might be of interest.
It works for me whichever way, I don't think for this purpose it should matter much, so not sure why it wouldn't be working one way and not another.

web/cobrands/northumberland/vendor/_tracking_code.js
```diff
@@ -0,0 +1,950 @@
...
+          document.body.style.position = 'fixed';
+          document.body.style.width = '100%';
+      }
+
+      allowBodyScroll() {
+          document.body.style.overflow = '';
+          document.body.style.position = '';
+          document.body.style.width = '';
+      }
+  }
```
dracos, line 782: The indent from here on is too much - this line shouldn't be indented, then everything below it can be less as well.

templates/web/northumberland/tracking_code.html
dracos: Also let's consider what's actually happening here. I could see a race condition whereby the listener for using their config is set up *after* the document is already ready, if async is present, in which case it never fires, so probably best not to use async.

dracos: Happy to jump on a call, you should understand how this is working in general.

web/cobrands/northumberland/vendor/_tracking_code.js
```diff
@@ -0,0 +1,950 @@
...
+          }
+
+          document.head.appendChild(script);
+      }
+
+      window.silktideCookieBannerManager.initCookieBanner = initCookieBanner;
+      window.silktideCookieBannerManager.updateCookieBannerConfig = updateCookieBannerConfig;
+      window.silktideCookieBannerManager.injectScript = injectScript;
+
+      if (document.readyState === 'loading') {
```
dracos: This block isn't needed, because of what's happening below (which is calling update config which then calls init itself, no need to call it twice).

web/cobrands/northumberland/vendor/_tracking_code.js
```diff
@@ -0,0 +1,950 @@
...
+          dataLayer.push(arguments);
+      }
+
+      document.addEventListener("DOMContentLoaded", function () {
+
+          if (typeof silktideCookieBannerManager === 'undefined') {
+              console.error("Silktide Cookie Banner Manager not found.");
+              return;
+          }
+          window.dataLayer = window.dataLayer || [];
```
dracos: Copy of lines just above, not needed twice.

web/cobrands/northumberland/vendor/_tracking_code.js
```diff
@@ -0,0 +1,950 @@
...
+
+          function gtag() {
+              dataLayer.push(arguments);
+          }
+
+          gtag('consent', 'default', {
```
dracos: This block isn't present in their files as far as I can see?

MorayMySoc: I was getting gtag undefined at one point and the silktide website had code for using with google. But removing this now seems to work so must have been some other issue.

dracos (approved):
Thanks, looks good :)

davea (approved):

PR #6031, opened 2026-06-30
[Echo] Always shrink photos

chrismytton:
We've had a couple of reports failing to send to Echo because the photos were too big. These were FMS reports, so weren't caught by this method. Removing this line means that this will shrink photos for any Echo report, rather than just Bulky Waste ones.


dracos (approved):
Should hopefully deal with any like this!

PR #6034, opened 2026-07-01
[Sutton] Update garden sack ID

chrismytton:
Similar to the change we did for Merton in 3f7c579b9 / FD-6858.

We had a couple of stuck reports come up earlier in the week and they went through after changing this, so make it permanent to prevent future stuck reports.

### DIFF
--- a/t/app/controller/waste_sutton_garden.t
+++ b/t/app/controller/waste_sutton_garden.t
@@ -187,7 +187,7 @@ sub garden_waste_bin_with_refuse_sacks {
 sub _garden_waste_service_units {
     my ($bin_count, $type) = @_;
 
-    my $bin_type_id = $type eq 'sack' ? 1928 : 1915;
+    my $bin_type_id = $type eq 'sack' ? 1910 : 1915;
 
     return [ {
         Id => 1002,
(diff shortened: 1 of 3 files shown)


t/app/controller/waste_sutton_garden.t
```diff
@@ -829,7 +829,7 @@ FixMyStreet::override_config {
 
         my ( $token, $new_report, $report_id ) = get_report_from_redirect( $sent_params->{returnUrl} );
 
-        check_extra_data_pre_confirm($new_report, bin_type => 1928, quantity => 11, new_bins => 11);
+        check_extra_data_pre_confirm($new_report, bin_type => 1910, quantity => 11, new_bins => 11);
```
dracos, line 832: There's another 1928 on line 190

dracos (changes requested):
One other ID is all :)

t/app/controller/waste_sutton_garden.t
chrismytton: Oops! Pushed a fixup commit for that.

dracos (approved):
(Not sure it's used given tests pass, but might as well :) ) thanks

PR #6038, opened 2026-07-03
[Zurich] Add report title to DM email

MorayMySoc:
https://mysocietysupport.freshdesk.com/a/tickets/7218

### DIFF
--- a/t/cobrand/zurich.t
+++ b/t/cobrand/zurich.t
@@ -320,6 +320,12 @@ subtest "Auto select category and input description from url" => sub {
     $mech->content_contains('Klicken Sie auf den Link im');
     ok (my $report = FixMyStreet::DB->resultset("Problem")->find({ title => 'Test 1234 Äï'}), "Report created with text from url");
     is $report->category, 'Cat1', "Report in correct category from url";
+    send_reports_for_zurich();
+    my @emails = $mech->get_email;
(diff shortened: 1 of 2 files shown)


t/cobrand/zurich.t
```diff
@@ -320,6 +320,12 @@ subtest "Auto select category and input description from url" => sub {
     $mech->content_contains('Klicken Sie auf den Link im');
     ok (my $report = FixMyStreet::DB->resultset("Problem")->find({ title => 'Test 1234 Äï'}), "Report created with text from url");
     is $report->category, 'Cat1', "Report in correct category from url";
+    send_reports_for_zurich();
+    my @emails = $mech->get_email;
+    my ($DMemail) = grep { $_->body =~ m#https://www.zurich/report/1# } @emails;
+    like $DMemail->as_string, qr/Title: Test Test 1 for 2/;
```
chrismytton: Looks like the tests are failing because the email output has `Title: Test Test 1 for 313` instead of `Title: Test Test 1 for 2`, so I think replacing the 2 with `\d+` or similar should fix it.

chrismytton (changes requested):

t/cobrand/zurich.t
MorayMySoc: Thank you - always forget about that!

chrismytton (approved):
Looks good!

PR #6041, opened 2026-07-08
[Waste] Make sure next collection contains data.

dracos:
It is possible for next to be auto-vivified (e.g. by Merton’s calendar loop in munge_bin_services_for_address), so check it has a size in the templates. For FD-7274


MorayMySoc (approved):
Great - thanks for picking that up

PR #6042, opened 2026-07-10
[Banes] Adds report_nearby_filter

MorayMySoc:
Adds a report_nearby_filter to filter params
when not coming through the around page and
having access to around_nearby_filter.

Used here to extend suggestion options to 10 for
inspectors and only offer open reports.

https://mysocietysupport.freshdesk.com/a/tickets/7161

[skip changelog]


davea (approved):
Looks good! A test would be good to prevent future breakage, maybe, but 👍

PR #6043, opened 2026-07-13
Dudley frontend

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5596 [skip changelog]

⚠️ Notes:
- Dudley cobrand uses `Proxima nova`. This is a paid font so I replaced it with `nunito-sans` pretty close and free. But it's worth double checking with them in case they want to be our clients.
- No footer for this showcase, because is usually a replica, so it won't be hard for them to imagine how it will look like. If they ask, we can include the `.bgd-skyline` infographic that goes on the top of their footer.

Screenshot 2026-07-13 at 12 06 58

### Preview
Screenshot 2026-07-13 at 12 02 01
Screenshot 2026-07-13 at 12 01 37
Screenshot 2026-07-13 at 12 01 21


dracos:
My commits here are to add skeleton cobrand, using OpenUSRN to get the nearest road in case if that's needed (assuming it will be), and then a little hack to go with the https://github.com/mysociety/open311-adapter/pull/521 - it sounds like we're going to have to poll each open report in Symology for updates; we already have the code in `bin/open311-update-reports` which does exactly this (the old-style Open311 way of updating), but it assumes it's fetching /reports/ and getting their status, whereas we can actually fetch history/updates and return them /like/ reports but with the update ID present. I think this should work. @chrismytton You okay to review my bits of this? Thanks.

perllib/FixMyStreet/Cobrand/Dudley.pm
```diff
@@ -0,0 +1,76 @@
+=head1 NAME
+
+FixMyStreet::Cobrand::Dudley - code specific to the Dudley cobrand
+
+=head1 SYNOPSIS
+
+Whatever is a metropolitan district authority, with a Symology backend.
```
chrismytton: Is this supposed to say "Dudley" instead of "Whatever" or similar?

templates/web/dudley/footer_extra_js.html
```diff
@@ -0,0 +1 @@
+[% PROCESS 'footer_extra_js_base.html' highways=1 roadworks=1 validation=1 cobrand_js=1 %]
```
chrismytton: Doesn't look like there's any cobrand JS yet, so think it can be cobrand_js=0.

chrismytton (approved):
Couple of minor comments, but no blockers.

lucascumsille:
@dracos Just in case I added a fixup commit that add the footer to the cobrand and another one for the site message.

PR #6045, opened 2026-07-16
New style for badge component

lucascumsille:
Fixes: Brent and Dudley provide feedback about how the current style for the badge component can be confused with a button.

- Decreased padding.
- Make the badge more similar to the "tags" from GOVUK.
- By default the `border-color` and background of the `badge` have the same colour to give a highlighted text effect.
- Replaced some of the attributes with variables so we can avoid overrides.
- For the font-size; I wanted  a rule to make sure we don't go below 14px for cobrands that use the classic 16px body size. And for cobrand that use >16px the badge doesn't look too big.

### Preview
Screenshot 2026-07-16 at 10 46 51
Screenshot 2026-07-16 at 10 47 04
Screenshot 2026-07-16 at 10 47 26
Screenshot 2026-07-16 at 11 19 38

[Archive 2.zip](https://github.com/user-attachments/files/30212436/Archive.2.zip)
[Archive.zip](https://github.com/user-attachments/files/30212438/Archive.zip)

⚠️ Considering the start the report with an image feature has already been around a few months, not sure if the "New" badge is needed anymore, so we can drop this component. I don't think we use it anywhere else.

[skip changelog]

### DIFF
--- a/web/cobrands/sass/_base.scss
+++ b/web/cobrands/sass/_base.scss
@@ -74,6 +74,15 @@ $footer-main-nav-color: #fff !default;
 // Invisble when there is not enough space
 $is-mobile-menu-visible: true !default;
 
+// Badge variables
+$badge-background-color: mix($primary, #fff, 20%) !default;
+$badge-border-color: $badge-background-color !default;
+$badge-padding: 2px 5px !default;
+$badge-border-radius: 0px !default;
+// Ensures the font-size doesn't go down below 14px for accessibility purposes and at the same time we can reuse the component in other elements like headings.
+$badge-font-size: clamp(14px,0.75em, 0.75em) !default;
+
+
 @import "_mixins";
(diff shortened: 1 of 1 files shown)


lucascumsille:
If approved get rid of the custom code for `badge` for Dudley and CRT

dracos (commented):
Looks good - but yes, been there since the start of the year, so perhaps we just drop it? :) Not sure if anyone else would have an opinion

lucascumsille:
Superseded by https://github.com/mysociety/fixmystreet/pull/6121

lucascumsille:
> Looks good - but yes, been there since the start of the year, so perhaps we just drop it? :) Not sure if anyone else would have an opinion

Thanks Matthew I ended up replacing this PR with this [new one](https://github.com/mysociety/fixmystreet/pull/6121)

PR #6046, opened 2026-07-16
Fix option alias clash in bin/fetch

chrismytton:
Both `--end` and `--exclude` had the alias `-e`, which meant you might try to set an end date with `-e` but it would be interpreted as a body name to exclude.

Fix by changing the alias for `--exclude` to `-x`. Changed this instead of end because it was added later.


neprune (approved):
👌

PR #6052, opened 2026-07-22
Implement pin_colour_key for cobrands

nephila-nacrea:
Closes https://github.com/mysociety/societyworks/issues/5595.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Aberdeenshire.pm
+++ b/perllib/FixMyStreet/Cobrand/Aberdeenshire.pm
@@ -295,6 +295,17 @@ sub pin_colour {
     return 'red';
 }
 
+sub pin_colour_key {
+    [
+        [ 'red', 'Open / action scheduled' ],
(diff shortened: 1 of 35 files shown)


lizettal:
@davea are you able to review this today so we can look to get this work wrapped up?

davea (changes requested):
Overall this is looking good! I think a few little tweaks are needed though:

<hr>

For cobrands that only display a single pin colour (Camden, Dumfries, Hackney, Merton, Peterborough, South Kesteven, though all ultimately derive from `UKCouncils::pin_colour_key`) we should probably hide the map pin legend UI entirely as it doesn't really add anything useful to the display:

Image

<hr>

Aberdeenshire's labels are quite unwieldy, though appreciate their colours are quite hard to describe succinctly! Maybe line breaks instead of `/`, and a smaller font size might work?

Image

<hr>

National Highways/Peabody don't seem to have anything in the popup UI

Image

Image

davea (approved):

PR #6055, opened 2026-07-28
Adding official and not official boxes to report updates

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/3797

### Notes
- I used the wording proposed by Brent, but we dropped the phrase `Any official response from %s will be marked as such.`
- The PR is currently marked as "Draft" because there is still some dev work to do.

### DEV Work
There are two commits marked as WIP:

#### `_email_comment_list.html`
- [ ] Please check the conditionals in `_email_comment_list.html` at the moment they are placeholders. And there is an extra  `ELSE` to cover the official response scenario.
- [ ] Please check if `replace('%s', report.body)` would be the right approach.

#### `_update_state.html`
I left some comments in the file.
- [ ] We need to implement the conditional when the response is anonymous and also made by a non-staff member.
- [ ] Implement a conditional when the response has been made by the authority in charge.

### Preview Emails
Screenshot 2026-07-28 at 05 38 47
Screenshot 2026-07-27 at 13 03 18

### Preview Front-end
Screenshot 2026-08-19 at 09 59 03

I tested them by adding the components temporarily in the front page. They all look the same the only difference at the moment is the black each cobrand uses for`$primary_b`

[Archive.zip](https://github.com/user-attachments/files/31215932/Archive.zip)

[skip changelog]


lucascumsille:
Hi @zarino before sending it to a developer, I wanted to check with you. Let me know if you think we should do any tweak =)

zarino:
@lucascumsille well it’s visually prominent, so that’s good!

I like the idea of reusing the Twitter/Facebook “verified” badge. Despite the recent changes at X, hopefully people still associate it with “official” sources.

My instinct is we could shorten the wording for the non-staff version, to make it more likely that people will actually read it, rather than skipping over it. But we can always tweak the wording later, with input from clients / account managers.

Two questions:

 - Will the pale yellow and pale blue colours be available in all cobrands’ emails? Or do any cobrands override the colours? (Just thinking, if a council overrides the pale blue to something else, we might need a way to easily include a colour-matched `official-icon.png`.)
 - If we go ahead with this styling in the emails, do you think we could/should introduce the same styling on updates in the web interface, for consistency?

lucascumsille:
> Will the pale yellow and pale blue colours be available in all cobrands’ emails? Or do any cobrands override the colours? (Just thinking, if a council overrides the pale blue to something else, we might need a way to easily include a colour-matched official-icon.png.)

I amended both commits so the style of each box + the icons can be overriden if a client wishes to.

> If we go ahead with this styling in the emails, do you think we could/should introduce the same styling on updates in the web interface, for consistency?

Absolutely, we can do that.

> My instinct is we could shorten the wording for the non-staff version, to make it more likely that people will actually read it, rather than skipping over it. But we can always tweak the wording later, with input from clients / account managers.

I'm happy to drop the second part `Any official response from %s will be marked as such.` it does seem like an overkill. Let me know what you think.

Let me know if after these amendments you are happy for me to pass it to develpment.

zarino:
> I amended both commits so the style of each box + the icons can be overriden if a client wishes to.

👍 

> Absolutely, we can do that.

Great, could you add it as part of this PR? Don’t know whether the variety of cobrand styles will add any complications, hopefully not?

> I'm happy to drop the second part `Any official response from %s will be marked as such.` it does seem like an overkill. Let me know what you think.

Yep! I think just the `Please be aware this update was posted anonymously by another FixMyStreet user, not by [council name].` is a good default, and then if any FMS Pro customers require something different, we can always override it for them.

PR #6057, opened 2026-07-28
[Kingston/Sutton] Bulky waste fixes

struan:
Allow reporting of open bulky/small items after 6pm on day of collection even if they are still open.

Tweak the not presented message to remove Bin.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/Echo.pm
@@ -1064,9 +1064,13 @@ sub booked_check_missed_collection {
             }
         }
 
+        # set a Kingston/Sutton specific flag (hence hardcoding time) if the current time
+        # is before the end of the possible collection window. Used to determine if a
+        # collection can be reported as missed
         my $today = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
-        if ($event->{date} > $today) {
-            $row->{bulky_before_collection_date} = 1;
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -173,6 +173,31 @@ sub waste_report_form_first_next {
     };
 }
 
+=head2 bulky_open_overdue
```
dracos: I guess as this is now same as Merton (I think?), it could move to SLWP2.pm? But meh, is fine.

perllib/FixMyStreet/Roles/Cobrand/Echo.pm
```diff
@@ -1065,8 +1065,9 @@ sub booked_check_missed_collection {
         }
 
         my $today = DateTime->now->set_time_zone(FixMyStreet->local_time_zone);
-        if ($event->{date} > $today) {
-            $row->{bulky_before_collection_date} = 1;
+        my $event_is_today = $today->clone->truncate(to => 'day') == $event->{date}->clone->truncate(to => 'day');
+        if ($event->{date} > $today || ($event_is_today && $today->hour < 18)) {
```
dracos, line 1072: Perhaps just a comment saying this flag is only used for K/S (so that's why okay to hard code 18).

dracos (commented):
Couple of tiny suggestions is all

dracos (approved):
Thanks :)

PR #6058, opened 2026-07-28
Ensure user email addresses are always stored lowercase

davea:
Normalises user email addresses to lowercase, preventing duplicate users being created.

For mysociety/societyworks#5594


davea:
PR https://github.com/mysociety/fixmystreet/pull/6065 will be done first as that fixes the immediate problem.

PR #6059, opened 2026-07-28
[Sutton][WW] Disputes: completed missed collection text

nephila-nacrea:
For update on https://app.basecamp.com/4020879/buckets/45711452/card_tables/cards/10082629714#__recording_10124155093.

[skip changelog]

### DIFF
--- a/templates/web/sutton/waste/enquiry-intro.html
+++ b/templates/web/sutton/waste/enquiry-intro.html
@@ -22,7 +22,7 @@ ELSIF booked_missed.$guid;
     SET booked = 1;
     SET service = booked_missed.$guid; # Original bulky/small booking
     SET service_name = service.service_name;
-    SET resolution = service.report_locked_out_reason;
+    SET resolution = service.report_locked_out_reason || service.report_open.resolution_reason;
     SET missed = service.dispute.missed_event; # May have a missed collection report against it
 END -%]
 
--- a/templates/web/sutton/waste/enquiry-intro.html
+++ b/templates/web/sutton/waste/enquiry-intro.html
@@ -7,19 +7,23 @@ SET booking_id = c.req.params.original_booking_id;
 IF services.$service_id;
     SET service = services.$service_id;
     SET service_name = service.service_name;
-    SET resolution = c.cobrand.resolution_text(service.last.resolution_id) || service.report_open.resolution_reason;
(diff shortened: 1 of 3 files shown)


templates/web/sutton/waste/enquiry-intro.html
```diff
@@ -30,21 +34,38 @@
...
 [%- IF category == 'Missed collection dispute' %]
-    [%- IF service_name %]
-    <p>Our crews reported that your [% service_name %] collection was not made:</p>
+    [% IF missed && missed.completed %]
+        [%
+            SET type = service_name == 'Bulky waste'
+                ? 'bulky'
+                : service_name == 'Small items'
+                    ? 'small'
+                    : '';
```
struan, line 43: rather than do this ternary here and an if below maybe just do the if once here and set the issue description and then use it below?

templates/web/sutton/waste/enquiry-intro.html
```diff
@@ -30,21 +34,38 @@
...
+            If you are disputing the reasons for this missed collection please use this form. We will investigate using our onboard 360 degree video cameras and other photo evidence before making a decision.
+        </p>
     [% END %]
 
-    <p>
-    If you are disputing the reasons for this missed collection please use this form. We will investigate using our onboard 360 degree video cameras and other photo evidence before making a decision.
-    </p>
 
     [%- SET update = cobrand.waste_bulky_resolution_photo_update( booking_id) %]
     [%- IF update %]
```
struan: I know this isn't part of this but this will be there if there is a photo and the collection has been marked as completed in which case the "This photo provides the evidence" text will be wrong. I'm not sure if it's fine because there's only ever a photo of there's a problem but maybe this should go inside the ELSE above?

struan (approved):
This looks like it fixes the reported issue. I've pushed a commit to fix a problem for incomplete missed collections.

The code tidy thing is a nice to have.

The photo question I think is technically a bug but I'm not sure if it's ever triggered in practice.

templates/web/sutton/waste/enquiry-intro.html
nephila-nacrea: I am not sure that is simplifying things, but maybe I have misunderstood (or just a difference in preference)?

PR #6061, opened 2026-07-29
Sutton bulky email template fixes

struan:
Corrects both the link and the text in the html updates.

[skip changelog]


dracos (approved):
Looks good, "from from" in the commit message
But this presumably all needs tidying anyway

PR #6065, opened 2026-07-31
Ensure OIDC email addresses are always stored lowercase

davea:
Normalises OIDC email addresses to lowercase, preventing duplicate users being created. Does the same for Lincs-created users imported via Open311 too.

For mysociety/societyworks#5594

### DIFF
--- a/t/app/controller/auth_social.t
+++ b/t/app/controller/auth_social.t
@@ -259,6 +259,13 @@ for my $state ( 'refused', 'no email', 'existing UID', 'okay' ) {
                 }
             }
 
+            if ($state eq 'okay' && $test->{type} eq 'oidc') {
+                # IdP might give us a mixed-case email; check that we store, and
+                # can look it up, lowercase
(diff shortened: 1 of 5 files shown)


t/app/controller/auth_social.t
```diff
@@ -259,6 +259,13 @@ for my $state ( 'refused', 'no email', 'existing UID', 'okay' ) {
                 }
             }
 
+            if ($state eq 'okay' && $test->{type} eq 'oidc') {
+                # IdP might give us a mixed-case email; check that we store, and
+                # can look it up, lowercase
+                my $user = FixMyStreet::DB->resultset('User')->find( { email => $test->{email} } );
+                is $user && $user->email, $test->{email}, 'User email stored lowercase';
```
dracos: Don't need the first `$user && ` here, it's always set (and it'd die anyway if it wasn't without it).

dracos (changes requested):
One unnecessary bit in the test is all :)

dracos (approved):
⏬

PR #6066, opened 2026-07-31
[UK] Left-anchor category search again.

dracos:
It gives too many false positives such as tree -> street or sign -> missing.
Add a non-letter search as well, which then still lets 'fly tip' match to 'flytip'.

Fixes https://github.com/mysociety/societyworks/issues/5604
[skip changelog]


MorayMySoc (approved):
Looks good

PR #6067, opened 2026-08-05
[CRT] Handle tunnel & tunnel portal assets

nephila-nacrea:
Asset config in 876ee81e7, fixup 6448446f1 in `canal-trust-tunnel-assets` branch in mysociety-servers.

[skip changelog]

### DIFF
--- a/templates/web/canalrivertrust/footer_extra_js.html
+++ b/templates/web/canalrivertrust/footer_extra_js.html
@@ -1,6 +1,7 @@
 [%
 IF bodyclass.match('mappage');
     scripts.push( version('/cobrands/fixmystreet/assets.js') );
+    scripts.push( version('/cobrands/fixmystreet-uk-councils/assets.js') );
     scripts.push( version('/cobrands/canalrivertrust/assets.js') );
     scripts.push( version('/js/asset_layers.js') );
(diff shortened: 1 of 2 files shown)


web/cobrands/canalrivertrust/assets.js
```diff
@@ -37,6 +37,45 @@ fixmystreet.assets.add(defaults, {
     }
 });
 
+fixmystreet.assets.tunnel_stylemap = new OpenLayers.StyleMap({
```
dracos: We're going to need to show these on .com as well, so I think these should be in fixmystreet-uk-councils/assets.js, not here (under fixmystreet.assets.crt... presumably). That's where the shared functions like this have ended up; this CRT specific file is for assets only on the CRT cobrand. You'll also then need the new JS file adding to the footer_extra_js.html template so that it's included.

dracos (changes requested):
Looks good, but think should be in a different file.

The server commit will then need updating with different function locations, but also looks good.

dracos (approved):

dracos:
The server commits haven't been pushed fwiw - I've merged the client side code, so will leave you to push/add that next week :)

PR #6068, opened 2026-08-05
[Open311] Fix same code update with template vars.

dracos:
If we get an update that is the same state/external status code as the previous, we still might want to include it if there are variables in the response template, because the variables may be different.

For Aberdeenshire FD-7242 though would also affect Dumfries. Hopefully shouldn't change any current behaviour [skip changelog]

### DIFF
--- a/perllib/Open311/UpdatesBase.pm
+++ b/perllib/Open311/UpdatesBase.pm
@@ -188,7 +188,7 @@ sub _process_update {
     my $old_ext_code = $p->get_extra_metadata('external_status_code') || '';
     my $ext_code_changed = $external_status_code && $external_status_code ne $old_ext_code;
     # So if there are no variables involved, and no state/code change, remove the template
-    if ($template && !$state_changed && !$ext_code_changed && $template !~ /\{\{/) {
+    if ($template && !$state_changed && !$ext_code_changed && $template->text !~ /\{\{/) {
         $template = undef;
     }
 
--- a/perllib/Open311/UpdatesBase.pm
+++ b/perllib/Open311/UpdatesBase.pm
@@ -164,6 +164,7 @@ sub _process_update {
     my ($self, $request, $p) = @_;
     my $open311 = $self->current_open311;
     my $body = $self->current_body;
+    my $cobrand = $body->get_cobrand_handler;
 
     $self->_handle_assigned_user($request, $p);
     $self->_handle_category_change($request, $p);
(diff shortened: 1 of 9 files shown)


perllib/Open311/UpdatesBase.pm
```diff
@@ -173,11 +174,25 @@ sub _process_update {
...
+    # variable replacement in comment_text_for_request could create different
+    # output for the same external status code. There shouldn't be an update
+    # with the same variables+code, but if there is we can fall back on the
+    # text comparison later.
+    my $state_changed = $state ne $old_state
+        && !( $p->is_fixed && FixMyStreet::DB::Result::Problem->fixed_states()->{$state} );
+    my $old_ext_code = $p->get_extra_metadata('external_status_code') || '';
+    my $ext_code_changed = $external_status_code && $external_status_code ne $old_ext_code;
+    # So if there are no variables involved, and no state/code change, remove the template
+    if ($template && !$state_changed && !$ext_code_changed && $template !~ /\{\{/) {
```
davea: I think this is missing `->text` - it's currently comparing something like `FixMyStreet::DB::Result::ResponseTemplate=HASH(0x...)`, which won't contain `{{`.

```suggestion
    if ($template && !$state_changed && !$ext_code_changed && $template->text !~ /\{\{/) {
```

davea (changes requested):
There's a sneaky bug in there which means this doesn't actually work :)

If you add some extra tests to L609 of `getservicerequestupdates.t` it illustrates the problem:

```perl
    $aber_problem->comments->delete;
    # another update taht only changes the extra data/response variables and leaves state/external status alone
    $local_requests_xml = setup_xml($aber_problem->external_id, $aber_problem->id, 'ACTION_SCHEDULED', '',
        '<jobStartDate>2026-01-20T09:00:00</jobStartDate><targetDate>2026-02-28T10:30:00</targetDate>');
    $local_requests_xml =~ s#</request_update>#<external_status_code>39</external_status_code></request_update>#;
    Open311->_inject_response('/servicerequestupdates.xml', $local_requests_xml);
        ALLOWED_COBRANDS => 'aberdeenshire',
    }, sub {
        $update->process_body;
    };
    $aber_problem->discard_changes;
    $c = $aber_problem->comments->first;
    is $aber_problem->comments->count, 1, 'comment count';
    is $c->state, 'confirmed', 'comment is visible';
    is $c->text, "Job start date: 20/01/2026\nTarget date: 28/02/2026", 'new template vars used';
```

perllib/Open311/UpdatesBase.pm
davea: One other thought (though don't think it matters in practice as no templates do this?) if template has `{{vars}}` in its `email_text` but not the normal `text` this would skip the update. Not sure how best to handle this without having multiple identical updates shown on the web page though.

dracos: Yeah, hopefully seems pretty unlikely? (Only Aberdeenshire/Dumfries do this at all at present.)

dracos:
Thanks, have pushed fix (with your test), and also fixed the warnings by putting in a new commit to treat private email text as empty string/the same everywhere.

davea (approved):

PR #6069, opened 2026-08-06
[CRT] Reviewed branch

nephila-nacrea:
[skip changelog]


dracos:
This PR doesn't include e.g. 728837f52ae076c1cad122a7b436e86998e8775d to turn on display_name which is already on staging and would be lost on a merge without checking. It looks like there's a `canal-trust-add-display-name` branch that was never pushed (cc @MorayMySoc). It also I'm afraid doesn't appear to contain a number of design changes that are already present, so I'm not sure the right commits were taken to be part of this. It looks like 3 commits were made directly to the staging branch, which shouldn't be done whenever possible (cc @lucascumsille). We need to keep clear which PRs/commits are where, otherwise we won't know where we are :)

I have for now cherry-picked 2096270f3d 0cc53013f1 544ce8992f (three design commits, presumably to be a fixup to e1f9e5474c in the end?) and c56548ff11 (adding display_name), and staging now merges much more cleanly.

This also fixes that staging still contained the unfinished SSO commit that lower-cased name - please try to remember to get updated/fixed branches on to staging when they're done, so it's also not forgotten.

lucascumsille:
@dracos Those commits are correct I had made them originally in this branch(`canal-trust-frontend-feedback`). I'm currently waiting for the client to see if they are happy with those changes.

PR #6071, opened 2026-08-06
Fix state filtering on /around for staff users

davea:
Logged-in staff would see incorrect pins on the map depending on the state filters they'd selected, as `body` wasn't being set on the stash which meant the staff user check in `stash_report_filter_status` wasn't passing.

For FD-7319.

[skip changelog]


MorayMySoc (approved):
Looks good

PR #6072, opened 2026-08-07
[Bromley] Always send email on open311 updates to passthrough.

neprune:
[skip changelog]

The endpoint fails if no email is provided.

We're seeing this with a couple updates.

Note this is regardless of what `email_alerts_requested` is set to.


dracos (approved):
I wonder if using `always_send_email` would have been okay, I assume the issue might just not have come up on reports? But having it separate is fine too :)

PR #6073, opened 2026-08-07
Export/import tests and couple of fixes

dracos:
Discovered after the Gloucestershire export/import this morning.
I noticed it had exported deleted templates without that information beforehand, so that was fine, I could delete them from the export, but I did not see that it was not importing send method on new categories, which meant any Email-devolved new categories got deleted by the Open311 sync until I fixed them manually.


neprune (approved):
LGTM. Appreciate the clean breakdown.

PR #6074, opened 2026-08-07
[Merton] Maintain discount of old free booking.

dracos:
A booking made before 6th August was made as a normal booking with a price of £0. In the new system (which could be free or discounted) a booking gets flagged as discounted if a discount applies. If an old, free booking is amended in the new system, we need to behave like it had been discounted. [skip changelog] FD-7087

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Merton/Waste.pm
+++ b/perllib/FixMyStreet/Cobrand/Merton/Waste.pm
@@ -699,8 +699,7 @@ sub bulky_pricing_strategy {
     my $previous = $c->stash->{amending_booking};
     my $discount;
     if ($previous) {
-        my $discounted = $previous->get_extra_field_value('discounted') || '';
-        $discount = 1 if $discounted eq 'yes';
+        $discount = 1 if $self->bulky_previous_was_discounted($previous);
     } else {
         my $data = $c->stash->{form}->saved_data;
         $discount = $self->bulky_discount_available_by_date($data);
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Merton/Waste.pm
```diff
@@ -750,6 +749,18 @@ sub bulky_minimum_cost {
...
 
+# If we are amending a booking, we need to maintain the discount - which is if
+# the booking was marked as discounted, or it has never had any payment on it
+# (so was made under a non-discouned system with a price of £0)
+sub bulky_previous_was_discounted {
+    my ($self, $previous) = @_;
+    my $discounted = $previous->get_extra_field_value('discounted') || '';
+    return 1 if $discounted eq 'yes';
+    my $already_paid = $self->get_total_paid($previous);
+    return 1 if !$already_paid;
```
neprune, line 760: nit: My understanding is in this scenario we assume the booking to have been paid and we are checking if the payment cost was zero, rather than if the report has been paid, so perhaps reads better as something like:
```
my $payment_amount = $self->get_total_paid($previous);
return 1 if $payment_amount == 0;
```

neprune (approved):

perllib/FixMyStreet/Cobrand/Merton/Waste.pm
dracos: Ah yes, by that variable name I meant "what amount has already been paid", so yours is clearer

dracos:
Turns out we can't do this because some of the £0 bookings should not have been made (people ordering twice or three times in the period when it was possible).

PR #6075, opened 2026-08-10
[pboro] Don't send censorship check mail for staff flytip reports.

neprune:
For https://mysocietysupport.freshdesk.com/a/tickets/7398.

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -372,13 +372,16 @@ sub open311_post_send {
     my $emails = $self->feature('open311_email') || {};
     my %flytipping_cats = map { $_ => 1 } @{ $self->_flytipping_categories };
     if ( $emails->{flytipping} && $flytipping_cats{$row->category} ) {
-        my $dest = [ [ $emails->{flytipping}, "Environmental Services" ] ];
-        $self->_send_extra_flytipping_email(
-            $row,
-            $h,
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -372,13 +372,16 @@ sub open311_post_send {
...
     my %flytipping_cats = map { $_ => 1 } @{ $self->_flytipping_categories };
     if ( $emails->{flytipping} && $flytipping_cats{$row->category} ) {
-        my $dest = [ [ $emails->{flytipping}, "Environmental Services" ] ];
-        $self->_send_extra_flytipping_email(
-            $row,
-            $h,
-            dest => $dest,
-            send_context => { flytipping_extra => 1 },
-        );
+        if (!$row->user->belongs_to_body($self->body->id)) {
```
dracos: This would still send the email for a report made by staff but "as another user" (which then uses that other user as the user on the report). I think you want to check `$row->get_extra_metadata("contributed_by")` which is set whenever a report is made by a staff user.

dracos (changes requested):

perllib/FixMyStreet/Cobrand/Peterborough.pm
neprune: Aha, thanks

dracos (approved):
Ta!

PR #6080, opened 2026-08-11
[Bromley] Fix open311 config to always send email on passthrough open311 updates.

neprune:
🙃 [skip changelog]


dracos (approved):
Looks good - so the original test https://github.com/mysociety/fixmystreet/pull/6072/changes#diff-c069d0cc4f4b4dc8ff1a2b572714883d8856c5d63abfb47a23a8544613a4e5e9R409 passed because it directly mocks the config being passed in, right.

PR #6083, opened 2026-08-12
[Kingston/Sutton] display correct failed bulky collection text on enquiry problem

struan:
[skip changelog]


dracos (approved):
Happy to approve, though seems bit odd to have two almost identical sentences:
"The crew have closed your collection task as 'not collected' because: resolution"
and:
"The crew have closed your collection task as not collected: resolution"
Feels like we could go with one of those in both branches?

PR #6091, opened 2026-08-18
[UK] Geocoding fallbacks

dracos:
Fixes some of https://github.com/mysociety/societyworks/issues/5662 (though something that did distance checks on all the results would be more comprehensive and presumably cover this result as well).
Fixes Peterborough thing noticed, and lets postcodes in strings that otherwise failed to geocode to be used.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Peterborough.pm
+++ b/perllib/FixMyStreet/Cobrand/Peterborough.pm
@@ -77,7 +77,7 @@ sub disambiguate_location {
         %{ $self->SUPER::disambiguate_location() },
         centre => '52.6085234396978,-0.253091266573947',
         bounds => [ 52.5060949603654, -0.497663559599628, 52.6752139533306, -0.0127696975457487 ],
-        result_only_if => 'City of Peterborough',
+        result_only_if => 'City of Peterborough|Peterborough, Cambridgeshire',
         result_strip => ', City of Peterborough, (East of England|Cambridgeshire and Peterborough), England',
     };
 }
--- a/perllib/FixMyStreet/Cobrand/UK.pm
+++ b/perllib/FixMyStreet/Cobrand/UK.pm
@@ -115,6 +115,29 @@ sub geocode_postcode {
     return {};
 }
 
+=head2 geocode_postcode_fallback
+
+If we get no geocoding response, and the string contains what could be a
+postcode, let's try that.
+
+=cut
+
+sub geocode_postcode_fallback {
+    my ( $self, $s ) = @_;
+    $s = uc($s);
+    if (my ($pc) = $s =~ m# (( [A-Z]{1,2}\d{1,2} | [A-Z]{1,2}\d[A-Z] ) \s* \d[A-Z][A-Z]) #x) {
+        if (mySociety::PostcodeUtil::is_valid_postcode($pc)) {
+            my $location = mySociety::MaPit::call('postcode', $pc);
+            return {} if $location->{error} || !$location->{coordsyst};
+            return {
+                latitude  => $location->{wgs84_lat},
+                longitude => $location->{wgs84_lon},
+            };
+        }
+    }
+    return {};
+}
+
 sub short_name {
     my $self = shift;
     my ($area) = @_;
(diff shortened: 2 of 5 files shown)


perllib/FixMyStreet/Cobrand/Peterborough.pm
```diff
@@ -77,7 +77,7 @@ sub disambiguate_location {
         %{ $self->SUPER::disambiguate_location() },
         centre => '52.6085234396978,-0.253091266573947',
         bounds => [ 52.5060949603654, -0.497663559599628, 52.6752139533306, -0.0127696975457487 ],
-        result_only_if => 'City of Peterborough',
+        result_only_if => 'City of Peterborough|Peterborough, Cambridgeshire',
```
zarino, line 80: I note that searching for “peterborough” on https://report.peterborough.gov.uk doesn’t work. Should it?

zarino (approved):
I can’t comment on the Perl code. But it looks like these are three nice little improvements to some of the edge-cases I identified earlier this week :+1:

perllib/FixMyStreet/Cobrand/Peterborough.pm
dracos: Yes, that's what this is to fix? :)

perllib/FixMyStreet/Geocode/OSM.pm
```diff
@@ -57,10 +57,17 @@ sub string {
     }
 
     my ( $error, @valid_locations, $latitude, $longitude, $address );
+    my %same_bar_last_letter;
     foreach (@$js) {
         next if $params->{result_only_if} && $_->{display_name} !~ /$params->{result_only_if}/;
         $_->{display_name} =~ s/$params->{result_strip}//g if $params->{result_strip};
 
+        # Ignore results same apart from last postcode letter
+        $_->{display_name} =~ s/, United Kingdom//;
```
chrismytton: I think this needs to be anchored to the end of the string, to avoid problems if "United Kingdom" appears elsewhere in the display_name.

perllib/FixMyStreet/Geocode/OSM.pm
```diff
@@ -57,10 +57,17 @@ sub string {
...
+        my $no_last_letter = substr $_->{display_name}, 0, -1;
+        next if $same_bar_last_letter{$no_last_letter};
+        $same_bar_last_letter{$no_last_letter} = 1;
+
```
chrismytton, line 73: It's possible for the `geocoder_munge_results` hook below to change the `display_name`, e.g. Hackney does. Is it worth moving this code after the `next unless $_->{display_name};` line below, so were deduping on the final results the user will see?

perllib/FixMyStreet/Cobrand/UK.pm
```diff
@@ -115,6 +115,27 @@ sub geocode_postcode {
...
+If we get no geocoding response, and the string contains what could be a
+postcode, let's try that.
+
+=cut
+
+sub geocode_postcode_fallback {
+    my ( $self, $s ) = @_;
+    $s = uc($s);
+    if (my ($pc) = $s =~ m# (( [A-Z]{1,2}\d{1,2} | [A-Z]{1,2}\d[A-Z] ) \s* \d[A-Z][A-Z]) #x) {
+        my $location = mySociety::MaPit::call('postcode', $pc);
```
chrismytton: Is it worth adding a `mySociety::PostcodeUtil::is_valid_postcode` before calling mapit?

chrismytton (changes requested):

perllib/FixMyStreet/Geocode/OSM.pm
dracos: hmm, interesting, that "blank display_name" line isn't actually needed any more - it was from before `result_only_if` was added which now does the same thing, but is presumably left in in case of a special case in future. But yes, if it was, it would make sense for this to be after it, though not sure it would make any actual difference.

perllib/FixMyStreet/Cobrand/UK.pm
dracos: I guess if it's extracted something that's almost like it, might as well, yeah!

chrismytton (approved):
🌐

PR #6093, opened 2026-08-18
[Kingston/Sutton] text for closed disputes on enquiry problem page

struan:
[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
@@ -449,7 +449,11 @@ sub _setup_missed_collection_disputes_for_service {
             $row->{dispute}{too_late} = 1;
         }
     } elsif ($missed_event && $dispute_event) {
-        $row->{dispute}{open} = $dispute_event;
+        if ($dispute_event->{closed}) {
+            $row->{dispute}{closed} = $dispute_event;
+        } else {
+            $row->{dispute}{open} = $dispute_event;
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -448,7 +448,7 @@ sub _setup_missed_collection_disputes_for_service {
             $row->{dispute}{missed_event} = $missed_event;
             $row->{dispute}{too_late} = 1;
         }
-    } elsif ($dispute_event) {
+    } elsif ($missed_event && $dispute_event) {
```
dracos, line 451: Is this a bug fixup for a previous commit, should it always have been checking this do you think?

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -448,7 +448,7 @@ sub _setup_missed_collection_disputes_for_service {
...
         $row->{dispute}{open} = $dispute_event;
```
dracos: Should this also be checking for closed state of the dispute event?

dracos (approved):
Sure it's fine, just trying to make sure I understand it!

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
struan: Yes. Otherwise this gets triggered if there's a dispute on something that isn't a missed collection. I guess to be properly robust it should be checking if the dispute is for the missed event but I think in practice that's likely to always be the case?

perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
struan: Hmmm, yes, it probably should.

PR #6094, opened 2026-08-19
[Kingston/Sutton] Prevent WW site from being indexed by search engines

chrismytton:
For FD-7409

### DIFF
--- a/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
+++ b/perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
@@ -34,6 +34,12 @@ sub send_questionnaires { 0 }
 
 sub abuse_reports_only { 1 }
 
+=item * Search engines are asked not to index the site
+
+=cut
+
+sub robots_noindex { 1 }
+
 =item * Only waste reports are shown on the cobrand
 
 =cut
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/Roles/Cobrand/KingstonSutton.pm
```diff
@@ -34,6 +34,12 @@ sub send_questionnaires { 0 }
 
 sub abuse_reports_only { 1 }
 
+=item * Search engines are asked not to index the site
+
+=cut
+
+sub noindex { 1 }
```
dracos: Could this perhaps be robots_noindex so it's clear what index it's referring to?

templates/web/base/common_header_tags.html
```diff
@@ -50,10 +50,10 @@
 }
 </script>
 
-[% IF robots %]
-    <meta name="robots" content="[% robots %]">
-[% ELSIF c.config.STAGING_SITE %]
+[% IF c.config.STAGING_SITE OR c.cobrand.call_hook('noindex') %]
```
dracos: Up to you, as a template falls back to false if it errors, I think this can be c.cobrand.noindex (or robots_noindex), no need for the call_hook wrapper (but guess it makes it clear it's a hook).

dracos (approved):
Looks good, would probably rename the function

chrismytton:
Thanks, have applied both suggestions.

PR #6096, opened 2026-08-19
[CRT] Canal trust extra data

MorayMySoc:
Adds extra data for attributes going to open311

Removes old code affecting the calls made to open311 that are not required any more

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/CanalRiverTrust.pm
+++ b/perllib/FixMyStreet/Cobrand/CanalRiverTrust.pm
@@ -43,6 +43,7 @@ sub contact_name { FixMyStreet::Cobrand::UKCouncils::contact_name($_[0]) }
 sub contact_email { FixMyStreet::Cobrand::UKCouncils::contact_email($_[0]) }
 sub users_staff_admin { FixMyStreet::Cobrand::UKCouncils::users_staff_admin($_[0]) }
 sub admin_allow_user { FixMyStreet::Cobrand::UKCouncils::admin_allow_user($_[0], $_[1]) }
+sub open311_extra_data { FixMyStreet::Cobrand::UKCouncils::open311_extra_data($_[0], $_[1], $_[2]) }
 
 sub enter_postcode_text { 'Enter a location, bridge number or postcode' }
 sub report_a_problem_label { 'Report' }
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/CanalRiverTrust.pm
```diff
@@ -133,4 +134,23 @@ sub admin_contact_validate_category {
     return "(CRT)" eq substr($category, -5) ? "" : "Category must end with (CRT).";
 }
 
+sub open311_extra_data_include {
+    my ($self, $row, $h) = @_;
+
+    my $open311_only = [
+        { name => 'fixmystreet_id',
```
dracos: I don't think this one is needed, it should be added by the `id_field` bit of `perllib/FixMyStreet/SendReport/Open311.pm`.

perllib/FixMyStreet/Cobrand/CanalRiverTrust.pm
```diff
@@ -133,4 +134,23 @@ sub admin_contact_validate_category {
...
+          value => $row->id },
+        { name => 'report_url',
+          value => $h->{url} },
+        { name => 'title',
+          value => $row->title },
+        { name => 'description',
+          value => $row->detail },
+        { name => 'nearest_address',
```
dracos: ~~Hmm, if this was closest_address, then that would also be automatically picked up by SendReport/Open311.pm in that same part. But perhaps closest_address is not same as nearest_address in some way?~~

Edit - only if it's an extra data field, and it is different in that it is just the address, not surrounds

dracos (changes requested):
Couple of queries about auto-set options is all

perllib/FixMyStreet/Cobrand/CanalRiverTrust.pm
MorayMySoc: I just suggested nearest_address as that seemed to be what the field required as it was called location. It's possible it will actually be taken from an asset layer. Not clear yet. I'll take this out and update open311 to look for closest_address

MorayMySoc: Ok, but does 'closest_address' require the category to have a field added which I didn't do for nearest_address?

dracos: ~~No, you don't need to have fields for anything that's only part of the Open311 transmission. That's only for stuff to be stored as part of the creation of the report, or in update_missing_data to add some data we want to store that is missing. extra_data_include and all that is only ephemeral.~~

This is incorrect, the auto-filled ones (easting/northing/closest_address/fixmystreet_id) do need to be extra data fields

MorayMySoc: Removed nearest_address as now not needed - location for SugraCRM will be from and extra field set by an asset and fixmystreet_id is created by the default Serivce on the open311 side

dracos (approved):

PR #6097, opened 2026-08-20
[Brent] Update request referral message

chrismytton:
Update emails and confirmation template for request referral message.

For FD-7412


dracos (approved):
🏤

PR #6086, opened 2026-08-13
[CRT] Front-end feeback v3

lucascumsille:
Fixes: https://github.com/mysociety/societyworks/issues/5656

```
1. Removed strapline. Now Desktop and mobile use straplineless logo.
3. Added `#1e29f5` to `.item-list--wards a, .item-list--reports a`
4. `.big-green-banner` moved down and change background color to `background: #ecf0f2 (Brand Towpath Tint 80%)`. Also removed uppercase.
5. Added `background: #ecf0f2 (Brand Towpath Tint 80%)` to `.pre-button-messaging, .box-warning, .extra-category-questions`
6a. Removed `#report-cta` button under `360px`
6b. `.dz-clickable .dz-message u` breaks into two lines when screen is below 360px.
6c. `#key-tools a` breaks into two lines when screen is below 360px.
7. `.report-list-filters` Changed labels from grey to white.
8b.  Added branding to KM form
8c. `.alerts__cta-box` uses `background: #ecf0f2 (Brand Towpath Tint 80%)`
10. Hides `#report-cta` in homepage
11a. Removes active underline in navbar
11b. Report button styled as the rest of links
11c. Renamed "Report a Problem" to "Report"
11d. Move report link to the top right.
11e. "Back to main site" is first element
12. Removed hover state from `.item-list__item`, `.nav-menu` and `.dropzone`
13. Added focus(yellow and black) state to inputs and links.
14b. Removed grey bg from `form-box ` in authpage
14c. Sign in form has been centered.
17. Links in sidebar use `#1e29f5`
```

### Pending for developers:

- [x] (Sign in page) Replace "Sign in with Microsoft Entra" with " Sign in with MyTrust account".
- [ ] Slide 4(Pint 3) Disable ward feature - as the Trust has no wards. Also if we can change the wording, "Get updates of problems in this council" considering CRT is not a council.

[skip changelog]

PR #6098, opened 2026-08-21
[CRT] Ampserand in body name encoded

MorayMySoc:
XML fails if the '&' in the body name has not been encoded so encode any occurrences of '&' before
serving the feed.

We should show reports list for rss feed, but not wards as that doesn't make sense for CRT

Fixes: https://github.com/mysociety/fixmystreet/pull/6086#issuecomment-5325941955

[skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Rss.pm
+++ b/perllib/FixMyStreet/App/Controller/Rss.pm
@@ -329,9 +329,13 @@ sub add_parameters : Private {
     my $uri = $c->uri_for( '/' . $c->req->path );
     $uri .= '?type=' . $c->stash->{ward_code} if $c->stash->{ward_code} && $c->stash->{ward};
 
+    my $qs = $c->stash->{qs} || '';
+    $uri =~ s/&/&amp;/g;
+    $qs =~ s/&/&amp;/g;
+
     $c->stash->{rss}->channel(
         title       => encode_entities($title),
-        link        => $c->uri_for($link) . ($c->stash->{qs} || ''),
+        link        => $c->uri_for($link) . $qs,
         description => encode_entities($desc),
         language    => 'en-gb',
         atom => { link => {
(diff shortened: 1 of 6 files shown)


perllib/FixMyStreet/App/Controller/Rss.pm
```diff
@@ -388,6 +392,7 @@ sub xsl : Path {
     };
     my $body = $c->view('Email')->render($c, 'xsl.xsl', $vars);
 
+    $body =~ s/&/&amp;/g;
```
dracos: This can't be right, as it will double encode any & character that has already been correctly encoded.
The actual issue here is that the xsl.xsl email template sets email_footer to site_name and is printed as safe, which means it isn't being escaped. I think the best solution would be to stop making email_footer safe (so it gets escaped) and any usage that needs it not to be escaped fixed a different way (It think that's only the .com submit.html template, looks like?)

perllib/FixMyStreet/App/Controller/Rss.pm
```diff
@@ -329,9 +329,13 @@ sub add_parameters : Private {
     my $uri = $c->uri_for( '/' . $c->req->path );
     $uri .= '?type=' . $c->stash->{ward_code} if $c->stash->{ward_code} && $c->stash->{ward};
 
+    my $qs = $c->stash->{qs};
```
dracos: If this had the || '' on it, then the if bit two lines later wouldn't be needed, nor the || '' in the setting of link

dracos (changes requested):
One clean up, one thing needs doing differently.
Also, I think the 'fixing & in RSS' should be a separate commit from "changing canal template to put back a link" - and in the end, we should tidy up the reviewed branch so that it was never removed in the first place, which is much easier if the changes are split up more.

perllib/FixMyStreet/App/Controller/Rss.pm
MorayMySoc: I'm not sure how you worked out that submit.html was required, but I've assumed the 'cy' version needs to be marked 'safe' as well.

dracos: Yep, looks like it. That looked to be the only use of `email_footer` that had HTML in it that I could see, anyway.

dracos (approved):
Think that should do it, yep!

PR #6102, opened 2026-08-21
[UK] Move admin user search to code.

dracos:
The update to PostgreSQL 15 has severely degraded the admin user search for cobrands. The non-limited search is fine, so move the filtering to the correct output more to the code side of things. [skip changelog]


chrismytton (approved):
Looks good! Hopefully has the desired effect on live.

PR #6106, opened 2026-08-25
Don't include unconfirmed alerts in CSV count

chrismytton:
The SQL for the alerts count was checking for `confirmed IS NOT NULL`, but the `confirmed` column of the `alert` table is not nullable, so this would always be true.

Fix it to actually check for `confirmed = 1` so that only confirmed alerts are included in the CSV export.

Spotted while replying to FD-7427


MorayMySoc (approved):
Looks good.

PR #6107, opened 2026-08-25
[Kingston/Sutton] handle closed disputes for bulky collections

struan:
[skip changelog]


dracos (commented):
What about the clause just above, when there is $missed_event and it's closed? Could that have a dispute that has been closed as well?

dracos (approved):
thanks, hopefully that's got it

PR #6112, opened 2026-08-25
Restrict superuser role admin to the cobrand's body

chrismytton:
When being viewed by a superuser, change the roles list, the role filter on the user list, and the body field on the role form to use $c->cobrand->body where there is one.

This fixes the confusing long list of all roles when viewing on a cobrand as a superuser.

Means that roles belonging to other bodies are no longer editable from a cobrand's admin.

### DIFF
--- a/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
+++ b/perllib/FixMyStreet/App/Controller/Admin/Roles.pm
@@ -11,13 +11,19 @@ sub auto :Private {
 
     my $user = $c->user;
     if ($user->is_superuser) {
-        $c->stash(rs => $c->model('DB::Role')->search_rs({}, {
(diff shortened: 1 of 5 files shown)


dracos (changes requested):
Nice, thanks. For perfection, it'd be good if the optgroup on the user role dropdown and the headings in roles/index.html weren't output when there's now only one option? Guess checking c.cobrand.body is easiest, or could set a special flag?

chrismytton:
Have hidden the optgroup and headings when there's a body. Also fixed a confusing bug where no optgroups were being shown at all for me locally, which turned out to be because I had only one language listed in LANGUAGES, so Translate::_translate wasn't using msgstr from the query.

dracos (approved):
Woo, good spot

PR #6115, opened 2026-08-26
[CRT] Asset messaging

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/5624.

Fixes various issues with asset messaging; the behaviour should now consistently be:

- names shown for aqueducts, bridges, locks, tunnels, tunnel portals
- no names/data shown for 'WSF facility' assets such as elsans
- an asset (ignoring the underlying canal asset, which must always be selected) must always be selected unless the category is `Boating etiquette`, `Electricity bollard`, `Navigation conditions`, or `Towpath`
- messaging disappears/changes when appropriate (e.g. when selecting a tunnel then selecting another category)

Corresponding config changes in mysociety-servers `canal-trust-asset-messaging`.

[skip changelog]


data/test-asset-layers.yml
```diff
@@ -276,6 +276,90 @@ camden:
...
+      class: OpenLayers.Layer.VectorAssetMove
+      max_resolution: 1.75
+      actions:
+        asset_found: fixmystreet.assets.canalrivertrust.asset_found
+        asset_not_found: fixmystreet.assets.canalrivertrust.asset_not_found
+      feature_code: sap_description
+    - name: web_csf_facilities
+      template: arcgis_spot
+      http_options:
+        url: https://services.arcgis.com/DknzyjEEie5tEW0u/arcgis/rest/services/Web_CSF_Facilities/FeatureServer/0/query
```
dracos, line 308: If we're using real URLs in this, then do make sure the domain is included in the .cypress/cypress.config.js `blockHosts` list, just to make sure we don't make any stray external requests if there's a fixture missing or something (not saying there is a fixture missing, but better to be safe)

web/cobrands/fixmystreet-uk-councils/assets.js
```diff
@@ -720,14 +718,11 @@ fixmystreet.assets.canalrivertrust.tunnel_found = function(layer, feature) {
...
-        // This may never be reached, as selecting a spot asset implies
-        // a road has also been found
-        return;
-    }
+    fixmystreet.message_controller.tunnel_not_found(layer);
 
-    fixmystreet.assets.named_select_action_not_found.call(layer);
+    // Force deletion of 'You have selected tunnel...' message as it does
+    // not otherwise happen if e.g. another category selected
+    layer.map_messaging.asset = layer.fixmystreet.no_asset_message;
```
chrismytton: This works, but I guess could break if a `no_asset_message` was ever added to the tunnels asset layer. Think it can be changed to `delete layer.map_messaging.asset` which would still behave the same, and would guard against future config changes causing problems.

chrismytton (approved):
One small thing, which I can fix before merging.

PR #6117, opened 2026-08-27
[CRT] Handle category names

MorayMySoc:
Category names are distinguished by having the pattern:

Category Name (CRT: group name)

CRT is for matching Canal and River Trust categories as a group for excluding or including where relevant.

The group name allows the categories to be unique so that stats can be reported on categories that are in multiple groups

[skip changelog]


dracos (approved):
Makes sense, perhaps we can think of a better way of doing this in future

PR #6118, opened 2026-08-28
[Greenwich] Add header to all Open311 requests.

dracos:
FD-7424 [skip changelog]
I thought what about if we had the situation where one endpoint was used in more than one place and we needed to send different headers, which did seem unlikely, but given using fixmystreet_body matches other uses in Open311.pm, I thought might as well set that consistently throughout and then can use that to set a cobrand-specific extra headers function. Perhaps this shouldn't have cobrand specific stuff in at all, but any other way seemed overly complicated.

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Greenwich.pm
+++ b/perllib/FixMyStreet/Cobrand/Greenwich.pm
@@ -180,4 +180,10 @@ sub open311_update_missing_data {
     }
 }
 
+sub open311_extra_headers {
+    my ($self, $req) = @_;
+    my $key = $self->feature('open311_api_key_header');
+    $req->header( Ocp_Apim_Subscription_Key => $key ) if $key;
+}
+
 1;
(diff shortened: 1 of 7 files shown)


perllib/FixMyStreet/Cobrand/Greenwich.pm
```diff
@@ -180,4 +180,10 @@ sub open311_update_missing_data {
     }
 }
 
+sub open311_extra_headers {
+    my ($self, $req) = @_;
+    my $key = $self->feature('open311_token');
```
chrismytton: There's already a feature named `open311_token` in `Open311::Updates::receive`, which is an _inbound_ shared secret. This should probably have a different name to avoid confusion.

chrismytton (changes requested):
Yeah hadn't thought of multiple cobrands using the same endpoint.

So yep, I think this approach is fine. Just one comment.

perllib/FixMyStreet/Cobrand/Greenwich.pm
dracos: TBH I did that deliberately to save having to come up with a new config name 🤣  But yes, probably should do.
Annoying that the admin has an API key variable already!

PR #6120, opened 2026-08-28
[Sutton] Correct IDs, premium food caddy upgrade.

dracos:
If they don't have one, assume we are Removing Containers a normal food caddy (46), and Add to Service a new premium food caddy. [skip changelog] FD-7472
(Also rename it in code because we called it indoor and it's outdoor)


struan (approved):
Looks good

PR #6122, opened 2026-09-01
[Bexley] Allow 0 existing bins in garden renewal.

dracos:
This was made editable in fc090deefe, but sometimes there will be zero bins currently. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/App/Form/Waste/Garden/Renew/Shared.pm
+++ b/perllib/FixMyStreet/App/Form/Waste/Garden/Renew/Shared.pm
@@ -52,7 +52,7 @@ sub intro {
             my $bin_count = $c->get_param('bins_wanted') || $form_data->{bins_wanted} || $data->{bins} || 1;
             my $new_bins = $bin_count - $current_bins;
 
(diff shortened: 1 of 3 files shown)


perllib/FixMyStreet/App/Form/Waste/Garden/Renew/Shared.pm
```diff
@@ -70,9 +70,19 @@ sub intro {
 
             $form_data->{_direct_debit_internal} = 1 if $c->cobrand->direct_debit_collection_method eq 'internal';
 
+            my $current_min = 1;
+            $current_min = 0 if $c->cobrand->moniker eq 'bexley'; # Bexley can say they have no bins when renewing
```
chrismytton: Wonder if this should be a hook similar to the ones on lines 55-56 above, rather than hardcoding Bexley here?

chrismytton (approved):
Looks good! One comment but not a blocker.

perllib/FixMyStreet/App/Form/Waste/Garden/Renew/Shared.pm
dracos: Yeah, I actually made it part of the same renewal edit current hook, given it's actually about the same thing.

PR #6125, opened 2026-09-02
Convert key tools to form actions

dracos:
This makes the key tools bar on report/around pages form submissions, as they can be considered the start of a flow, sort of. This shouldn't change the visual display at all.

The font change (or something like it anyway) was needed for the "Back to all reports" to stay in the same font as currently. I put the font change in a separate commit for discussion - as it says in the commit message, it looks like the majority of form inputs are already overriding to `inherit` anyway (and most cobrands have the same body/meta font anyway anyway), so if we did want a different meta font it seemed better to keep that for 'meta line' things (like the timestamp under a report in a list) and try and have all inputs use the same font. (e.g. reporting form seems to all end up as 'inherit' in the end, but the contact form page is not). Some bits could presumably be simplified on top of this but haven't looked at that.

### DIFF
--- a/web/cobrands/fixmystreet/fixmystreet.js
+++ b/web/cobrands/fixmystreet/fixmystreet.js
@@ -111,7 +111,8 @@ function isR2L() {
                 if (!$this.addClass('hover').data('setup')) {
                     // Optionally fill $drawer with HTML from an AJAX data source
                     if (ajax) {
-                        var href = $this.attr('href') + ';ajax=1';
+                        var href = this.form.action || $this.attr('href');
+                        href = href + ';ajax=1';
                         var margin = isR2L() ? 'margin-left' : 'margin-right';
                         var $ajax_result = $('<div>').appendTo($drawer);
                         $ajax_result.html('<p style="text-align:center">Loading</p>');
@@ -1848,7 +1849,7 @@ $.extend(fixmystreet.set_up, {
     });
 
     $('#map_sidebar').on('click', '.js-back-to-report-list', function(e) {
-        var report_list_url = $(this).attr('href');
+        var report_list_url = this.form.action;
         var map_state = around_map_state;
         var set_map_state = true;
         fixmystreet.back_to_reports_list(e, report_list_url, map_state, set_map_state);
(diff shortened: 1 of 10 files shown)


web/cobrands/fixmystreet/fixmystreet.js
```diff
@@ -1410,9 +1411,9 @@ $.extend(fixmystreet.set_up, {
 
     if ($('.mobile').length) {
         // Make sure we end up with one Get updates link
-        if ($('#key-tools a.js-feed').length) {
-            $('#sub_map_links a.js-feed').remove();
-            $('#key-tools a.js-feed').appendTo('#sub_map_links');
+        if ($('#key-tools .js-feed').length) {
```
lucascumsille, line 1414: @dracos I removed the `a` tag from here considering now is a `button`

web/cobrands/sass/_base.scss
```diff
@@ -2247,7 +2247,7 @@ html.js #map .noscript {
   bottom: 0;
   margin: 0;
 
-  a {
+  a, button {
```
lucascumsille, line 2257: @dracos So it uses the same space as anchors.

web/cobrands/sass/_base.scss
```diff
@@ -2288,6 +2288,15 @@ html.js #map .noscript {
     @extend %sub-map-link-icon;
     background-position: -96px 0;
   }
+
+  #key-tool-around-updates {
```
lucascumsille, line 2300: Now that is a button, it needed some styling reset.

web/cobrands/sass/_base.scss
```diff
@@ -1848,6 +1848,7 @@ input.final-submit {
   border: none;
   border-bottom: 0.2em solid #eee;
   color: #666;
+  width: calc(100% + 2em);
```
lucascumsille, line 1857: It wasn't using the whole width

lucascumsille:
@dracos This all look good. I added some commits, let me know if you need me to change anything, otherwise I'm happy for you to squash and merge. 

Just in case I tested in the following pages(mobile and desktop):

- /reports/[cobrand here]?zoom=6&lat=51.543&lon=-0.15912
- /report/666
- around page

lucascumsille (approved):

PR #6128, opened 2026-09-03
[Sutton] Changes for update emails

dracos:
Remove some bits from email template, and spot justified data field to split to two different states. [skip changelog]


struan (approved):
👍

PR #6129, opened 2026-09-04
Use users full text index for admin report email search

chrismytton:
This should hopefully speed up searching for reports by email address. It uses the same full-text search that the user search is already using.

Note, this means we're now matching tokens rather than substrings. The code only runs on if it's a valid email address, so foo@bar.org still finds foo@bar.org.uk, but will no longer find xfoo@bar.org.


chrismytton:
Just had a bit more of a look at the user search (which already uses this) and realised that searching for foo@bar.org will also match foo+test@bar.org etc, which might not be what's wanted for a report search where an exact match is probably what you're after, hmm.

PR #6135, opened 2026-09-07
[Bromley] Include extra data fields in CSV export.

dracos:
Fixes FD-7520. [skip changelog]

### DIFF
--- a/perllib/FixMyStreet/Cobrand/Bromley.pm
+++ b/perllib/FixMyStreet/Cobrand/Bromley.pm
@@ -1116,30 +1116,51 @@ sub waste_cc_payment_sale_ref {
 sub dashboard_export_problems_add_columns {
     my ($self, $csv) = @_;
 
+    my @contacts = $csv->body->contacts->order_by('category')->all;
+    my %extra_columns;
+    foreach my $contact (@contacts) {
+        foreach (@{$contact->get_metadata_for_storage}) {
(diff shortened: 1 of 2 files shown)


perllib/FixMyStreet/Cobrand/Bromley.pm
```diff
@@ -1116,30 +1116,51 @@ sub waste_cc_payment_sale_ref {
...
+    my @contacts = $csv->body->contacts->order_by('category')->all;
+    my %extra_columns;
+    foreach my $contact (@contacts) {
+        foreach (@{$contact->get_metadata_for_storage}) {
+            $extra_columns{$_->{code}} = $_->{description} || $_->{code};
+        }
+    }
+
+    my @extra_columns = sort keys %extra_columns;
+    @extra_columns = map { $_ => $extra_columns{$_} } @extra_columns;
```
neprune, line 1128: I hadn't clocked before that Perl lets you keep completely separate variables sharing an identifier, if they're a different type.

Maybe an argument for just having a separate var name but appreciate this might be a standard idiom I just wasn't up to speed with.

neprune (approved):

perllib/FixMyStreet/Cobrand/Bromley.pm
dracos: I always consider the $/@/% to be part of the name, I guess :) But yes, I'd only ever use it for something like this, where it's basically an equivalent thing.

PR #6136, opened 2026-09-08
Keep National Highways answer during form re-render

chrismytton:
Signing in with a password mid-report re-renders the page, losing the answer to the "on this road, or somewhere else?" question and rebuilding it defaulting to "On the road". That re-routed the report to National Highways and hid the council category the user had picked.

Render the submitted do_not_send back into the form, and have the JS read it on load to decide which radio to check.

Fixture changes add the missing data-valuealone values to match the live response.

Fixes https://github.com/mysociety/societyworks/issues/5334


templates/web/base/report/new/form_report.html
```diff
@@ -128,7 +128,7 @@ <h3>[% loc('Photo uploaded successfully') %]</h3>
 [% END %]
 
 <input type="hidden" id="single_body_only" name="single_body_only" value="">
-<input type="hidden" id="do_not_send" name="do_not_send" value="">
+<input type="hidden" id="do_not_send" name="do_not_send" value="[% c.get_param('do_not_send') | html %]">
```
dracos, line 131: No need for `| html`, it's the default.

dracos (approved):
One unneeded filter is all, thanks :)

PR #6140, opened 2026-09-09
[Hackney] Allow updates left on split destination reports to be sent

davea:
The unparsed service_code from the contact was being passed to Open311 adapter and being rejected; now we parse it when sending updates and send the correct value. If the original report was emailed we just mark the update as skipped.

[skip changelog]


dracos (approved):
Woo, thanks

PR #6143, opened 2026-09-17
Use transaction for adding comments in archive-old-enquiries

davea:
This prevents a race condition with the send-alerts script which would occasionally cause alerts to stop being sent due to duplicate rows in the subquery.

(I couldn't figure out a sensible way to test this, because the tests already run inside transactions, so lining up a send-alerts run to overlap with archiving old enquiries didn't seem possible)

[skip changelog]


dracos (approved):
Yeah, tricky to test. Think it should be okay :)

PR #6144, opened 2026-09-17
[Bulky] Record user who amended/cancelled booking

davea:
Sets the user of the update created when a bulky booking is amended or cancelled to the logged in user who made the change.

The send_alert_to_reporter extra field is necessary to ensure Sutton cancellation emails get sent when the original reporter is the one doing the cancellation. (I'm not sure if this is feasible due to the changes it makes to an already slow query, however...)

For FD-7538

[skip changelog]


perllib/FixMyStreet/Script/Alerts.pm
```diff
@@ -72,7 +72,7 @@ sub send_alert_type {
         and alert_type='$ref' and whendisabled is null and $item_table.confirmed >= whensubscribed
         and $item_table.confirmed >= current_timestamp - '7 days'::interval
          and (select whenqueued from alert_sent where alert_sent.alert_id = alert.id and alert_sent.parameter = $item_table.id::text) is null
-        and $item_table.user_id <> alert.user_id
+        and ($item_table.user_id <> alert.user_id or $item_table.extra @> '{\"send_alert_to_reporter\":1}')
```
dracos: Yeah, I'd be quite worried about adding this :-/
It looks like Bromley is currently sending two emails on customer cancellation - one in the bulky_send_cancellation_confirmation hook and then the update email is going through as well?
So perhaps the better solution is to make the Bromley email more generic/for everyone, and always cancel the alert on the cancellation update, so then this wouldn't be needed.

Or if that wouldn't work for some reason I haven't thought of, could remove this whole line from the query and in code do the same check

dracos (changes requested):

perllib/FixMyStreet/Script/Alerts.pm
davea: I decided the simplest thing to actually solve the problem mentioned in the FD ticket is what I mentioned on our call the other day - just setting `contributed_by` on the comments and then exposing that in the admin UI. Hope that's OK!

dracos: Happy to surface contributed_by in updates, given we do already record that - though in the normal case, staff leaving update on behalf of someone or what have you, seems a bit odd to have it in the list rather than on the update edit page, as with reports? But perhaps that doesn't happen often.
But currently it's only ever been a 'staff' flag, never had normal users setting it before; I can't see anywhere obvious that that does matter, but does seem a bit silly to be recording two users for one update when we know we only need one. My first point, would involve changing `$p->cancel_update_alert($update->id) if $type eq 'delayed';` to remove the type check if we're always cancelling it, and moving the Bromley hook to UKCouncils or core code. I think that's all? And that then fixes the double email bug I noticed as well.

davea: Yeah those are valid points! I've rejigged it to do the right thing, I think, mind taking another look?

dracos (approved):
Thanks, looks good :)

